PR #4916, opened 2022-11-11
Enable nullable checks in PublicAPI tracking files

zivkan:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/1999

Regression? no

## Description

* Set `#nullable enable` in all `PublicAPI.*.txt` files
* Use code fixes to annotate all APIs whose nullability is unknown (lines starting with `~`)
* NoWarn RS0041 (raised when there's at least one line with `~`)
* Opt all the `build\Shared\*.cs` files into nullable checks
* Improve the `ship-public-apis` tool, so it minimizes churn on files
 
## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


build/Shared/TypeExtensions.cs
```diff
@@ -1,4 +1,7 @@
 #if IS_NET40_CLIENT
+
```
nkolev92: We probably don't even need this file anymore do we?

build/Shared/TypeInfo.cs
```diff
@@ -1,4 +1,7 @@
 #if IS_NET40_CLIENT
+
```
nkolev92: Same applies here, we don't multi target net40 anymore.

src/NuGet.Core/NuGet.CommandLine.XPlat/PublicAPI.Shipped.txt
```diff
@@ -0,0 +1 @@
+#nullable enable
```
nkolev92: I don't imagine we need this do we?

nkolev92 (commented):
I had no clue you can enable nullability public API tracking files. 

Can you explain how do consumers of APIs benefit from this?

zivkan:
> Can you explain how do consumers of APIs benefit from this?

For the same reason why we all benefit from nullable checks when the BCL was annotated with attributes about when APIs might or will not return null. APIs that never return null don't need to be checked for null, and the only way the analyzer can make the decision when calling APIs in other assemblies, is when those assemblies have appropriate annotations.

Customers won't see the PublicAPI.*.txt files directly. Instead, consider [this](https://github.com/NuGet/NuGet.Client/pull/4912/commits/99ecae66a1aadcd01d46ec9f045c85bb0abedc60#diff-2ae20246c40c717c164cb40b9858455a72a6bc262b23e4d0a14fc7824fc1d860R43-R48) from my other PR. When enabling nullable checks on a file with public APIs, the public API analyzer will raise a message (that we elevate to an error) if the public api tracking file hasn't been flagged as nullable enabled.

This PR modified NuGet.Protocol's PublicAPI.Shipped.txt files with 1,251 lines changed, but NuGet.Protocol has different APIs per TFM, so times 3 that's about 3,753 lines just to enable nullable checking.  If I did that in my other PR, it would make that PR much harder to review.

So, by doing a one-time migration of all the public API tracking files to enable nullable checking, it means all future PRs that actually start annotating public APIs with nullable information will be smaller.

Also since we just shipped 6.4.0, and its public API changes were just merged into the dev branch, it minimizes risk of future merge conflicts if we do this change now.

edit: also, let's not forget that our own code (VS code, and code that ships in the .NET SDK) is also a consumer of the public APIs, and therefore we can reduce risk of NullReferenceExceptions by progressively enabling nullable checks throughout the entire repo.

nkolev92:
You described what nullability is, but that wasn't really my question. 

How does editing txt files help anyone? 
I didn't think these files get distributed in any way.
Wouldn't you need to declare the nullability on the APIs themselves?

zivkan:
Sorry, I misunderstood.

The Public API analyzer tracks nullability in the txt files, as changing nullability (particularly making something that previously never returned null to start returning null, or inputs which previously accepted null to stop accepting null) is a breaking change. https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md#nullable-reference-type-support

So, we can't enable nullability in our own code and use the public api analyzer, without having nullability annotations in our txt file as well.

nkolev92 (dismissed):
Thanks for the clarification.

I'm gonna trust you did this correctly, lol. 

We can ofc tweak things later if they are incorrect. It's not a hard rule.

nkolev92 (approved):

PR #4930, opened 2022-11-16
Fix Apex tests running locally

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2011

## Description

* Update Apex tests to use `SimpleTestPathContext` instead of using current directory
image
* Update OptProf tests to have any empty Directory.Build.props/targets when running the tests

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/IVsServicesTestCase.cs
```diff
@@ -103,8 +97,11 @@ public async Task SimpleInstallFromIVsInstaller_PackageSourceMapping_WithSingleF
             await CommonUtility.CreatePackageInSourceAsync(privateRepositoryPath, packageName, packageVersion);
 
             // Create nuget.config with Package source mapping filtering rules before project is created.
-            CommonUtility.CreateConfigurationFile(Path.Combine(mainDirectory, "NuGet.config"), $@"<?xml version=""1.0"" encoding=""utf-8""?>
+            CommonUtility.CreateConfigurationFile(simpleTestPathContext.NuGetConfig, $@"<?xml version=""1.0"" encoding=""utf-8""?>
```
nkolev92: You probably don't need this if you are using simple test path context.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/IVsServicesTestCase.cs
```diff
@@ -87,14 +86,9 @@ public void IVSPathContextProvider2_WithEmptySolution_WhenTryCreateUserWideConte
...
-            EnsureVisualStudioHost();
-            var dte = VisualStudio.Dte;
-            string mainDirectory = CommonUtility.CreateSolutionDirectory(Directory.GetCurrentDirectory());
-            string solutionDirectory = Path.Combine(mainDirectory, "Solution");
-            var solutionService = VisualStudio.Get<SolutionService>();
+            using var simpleTestPathContext = new SimpleTestPathContext();
             var nugetTestService = GetNuGetTestService();
-
-            var privateRepositoryPath = Path.Combine(mainDirectory, "PrivateRepository");
+            var privateRepositoryPath = Path.Combine(simpleTestPathContext.SolutionRoot, "PrivateRepository");
```
nkolev92: Similar to this, you don't need it if you're using the SimpleTestPathContext. It should cover that for oyu.

nkolev92 (commented):
I think we can simplify the test cases even more by just using the defaults that SimpleTestPathContext provides.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/IVsServicesTestCase.cs
jeffkl: I tried to leave the test mostly alone without refactoring it too much.  I was worried about messing them up so I tried to re-use the code and use the `SimpleTestPathContext`.  Should I really get in there and re-factor these tests?

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/IVsServicesTestCase.cs
jeffkl: The code is setting up package source mappings in the config by just writing a static XML string to disk.  Do we have code in the SimpleTestPathContext/Settings stuff to add package source mappings?

zivkan: In this case it is needed, because it sets up package source mapping. SimpleTestPathContext sets up a nuget.config with a test specific global packages folder and package source, but I'm not aware of a way to set up package source mapping via APIs.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/IVsServicesTestCase.cs
zivkan: `SimpleTestPathContext` already provides a package source: https://github.com/NuGet/NuGet.Client/blob/2e6da26086302f2c0153e026816e9b1797a0a586/test/TestUtilities/Test.Utility/SimpleTestSetup/SimpleTestPathContext.cs#L42

So, the test could use that, instead of the `PrivateRepository` directory.

```suggestion
            var privateRepositoryPath = simpleTestPathContext.PackageSource;
```

The following `Directory.CreateDirectory` might not be needed either. I think that SimpleTestPathContext creates it (even when it's not needed, yay slow tests!).

zivkan (dismissed):

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/IVsServicesTestCase.cs
nkolev92: https://github.com/NuGet/NuGet.Client/blob/20be995d651808feb5cd48f30cb97197e497377c/test/TestUtilities/Test.Utility/SimpleTestSetup/SimpleTestSettingsContext.cs#L100 can technically edit it. 
We can always add APIs in there that updates it. 
I don't think we should be using write to file.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/IVsServicesTestCase.cs
nkolev92: I think we should fix the simple test path context, but if you don't want to do it in this PR it's ok, as long as we have a future tracking one.

I don't expect the change to be super complicated and likely to just work on the first try.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/IVsServicesTestCase.cs
jeffkl: Okay I added a new API to add package source mappings in the test utility

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/IVsServicesTestCase.cs
jeffkl: I have refactored everything to clean things up and reduce code.  I also found some tests that were pulling newtonsoft.json from the network so now they use test packages instead.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/Apex/NuGetUIProjectTestExtension.cs
```diff
@@ -26,7 +26,7 @@ public NuGetUIProjectTestExtension(ApexTestUIProject project, ITestLogger logger
             _logger = logger;
         }
 
-        public bool SeachPackgeFromUI(string searchText)
+        public bool SearchPackgeFromUI(string searchText)
```
nkolev92: You fixed one typo, but not the other :D 

`SearchPack*a*ge`
I'd add a suggestion but you need to chnage the callers too so that wouldn't help.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/IVsServicesTestCase.cs
```diff
@@ -14,6 +14,14 @@ namespace NuGet.Tests.Apex
 {
     public class IVsServicesTestCase : SharedVisualStudioHostTestClass, IClassFixture<VisualStudioHostFixtureFactory>
     {
+        private const string TestPackageName = "Contoso.A";
+        private const string TestPackageVersionV1 = "1.0.0";
+        private const string TestPackageVersionV2 = "2.0.0";
+        private const string PrimarySourceName = "source";
+        private const string SecondarySourceName = "SecondarySource";
+
+        private readonly SimpleTestPathContext _pathContext = new SimpleTestPathContext();
```
nkolev92, line 23: Wouldn't this share the same pathContext across tests?
How does this get disposed?

test/NuGet.Tests.Apex/NuGet.Tests.Apex/Apex/NuGetUIProjectTestExtension.cs
jeffkl: Haha how did I not see that!?

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/IVsServicesTestCase.cs
jeffkl: xunit news up an instance of this class for each test and then xunit calls `Dispose()` for the instance since it implements `IDisposable` which calls `_pathContext.Dispose()`

https://xunit.net/docs/shared-context

nkolev92: Makes sense. I mixed it up with the Collection behavior.

nkolev92 (approved):
👏

zivkan (approved):

PR #4931, opened 2022-11-16
Remove Common and FileSystem submodules

### LINKED ISSUE #2345: NuGet.Commands embedded copy of FileSystemGlobbing is now public
Because `preprocess` code is not available in RC2, the code that "internalizes" the copy of FileSystemGlobbing that NuGet.Commands embeds isn't in effect. As a result, the classes are public and conflict with the real FileSystemGlobbing. I've worked around it in CLI, but it should be fixed.

/cc @emgarten @yishaigalatzer @davidfowl

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Home/issues/2345

## Description

* Remove submodules
* Reference packages
* ILMerge new assemblies into `NuGet.exe`
* Include new assembles in our VSIX but don't have to provide base paths for NGen since these assemblies ship in Visual Studio (https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/435192)
* Update logic for patching .NET CLI to work when testing with new assemblies that have yet to be in the .NET SDK
* Update ILMerge logic to include everything in the output folder of `NuGet.CommandLine`
* Sort everything in our `.swr` file and clean it up.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


erdembayar:
Could you be able to run RPS test just in case? Last time it was blocked for VS insertion because of RPS test failure.

src/NuGet.Clients/NuGet.VisualStudio.Client/.vsixignore
```diff
@@ -4,6 +4,9 @@ Ben.Demystifier.dll
 MessagePack.Annotations.dll
 MessagePack.dll
 Microsoft.Bcl.AsyncInterfaces.dll
+Microsoft.Extensions.FileProviders.Abstractions.dll
+Microsoft.Extensions.FileSystemGlobbing.dll
+Microsoft.Extensions.Primitives.dll
```
zivkan: Does msbuild restore on the command line still work?   VS's PrivateAssemblies aren't used by MSBuild.exe.

jeffkl: That's a good question, do any of our end-to-end tests cover that scenario?

zivkan: No. You'll need to create a VM, either install the CI build's bootstrapped VS, or just "normal VS" and download the NuGet vsix and install it, then manually test it.

It probably would be a good idea to add a test that runs msbuild.exe at some point. Thinking about how to make such a test work, especially locally where I have many versions of VS installed side-by-side, gives me a headache.

nkolev92: I thought we had MSBuild end to end tests, https://github.com/NuGet/NuGet.Client/blob/dev/test/NuGet.Core.FuncTests/Msbuild.Integration.Test/MsbuildRestoreTaskTests.cs.

build/common.ps1
```diff
@@ -227,6 +227,7 @@ Function Install-DotnetCLI {
     # However, Invoke-BuildStep checks if any error happened, ever. Hence we need to run dotnet-install
     # in a different process, to avoid treating their handled errors as build errors.
     & powershell $DotNetInstall -Runtime dotnet -Channel 3.1 -InstallDir $CLIRoot -NoPath
+    & powershell $DotNetInstall -Runtime dotnet -Channel 5.0 -InstallDir $CLIRoot -NoPath
```
nkolev92, line 230: Why do we need this?

src/NuGet.Clients/NuGet.VisualStudio.Client/.vsixignore
nkolev92: Didn't the assemblies get moved to the `PublicAssemblies` recently?

src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.csproj
```diff
@@ -90,7 +90,7 @@
       <PathToMergedNuGetPack>$(OutputPath)ilmerge\NuGet.Build.Tasks.Pack.dll</PathToMergedNuGetPack>
     </PropertyGroup>
     <ItemGroup>
-      <BuildArtifacts Include="$(OutputPath)*.dll" Exclude="$(OutputPath)*Pack.dll" />
+      <BuildArtifacts Include="$(OutputPath)publish\*.dll" Exclude="$(OutputPath)publish\*Pack.dll" />
```
nkolev92, line 93: how was this working before? Why do we need this change?

src/NuGet.Clients/NuGet.VisualStudio.Client/.vsixignore
jeffkl: It looks like the functional tests use the locally built task assembly which is next to the new dependencies so it "works".  I'm testing with the VSIX and will report back.

jeffkl: 🤦‍♂️ 
```
D:\temp\ClassLibrary2>msbuild /t:restore /clp:v=m;summary;forcenoalign
MSBuild version 17.5.0-preview-22564-01+d14b74d6f for .NET Framework
D:\VS-IntPreview\Common7\IDE\CommonExtensions\Microsoft\NuGet\NuGet.targets(132,5): error : Could not load file or assembly 'Microsoft.Extensions.FileProviders.Abstractions, Version=6.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' or one of its dependencies. The system cannot find the file specified. [D:\temp\ClassLibrary2\ClassLibrary2.csproj]

Build FAILED.

D:\VS-IntPreview\Common7\IDE\CommonExtensions\Microsoft\NuGet\NuGet.targets(132,5): error : Could not load file or assembly 'Microsoft.Extensions.FileProviders.Abstractions, Version=6.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' or one of its dependencies. The system cannot find the file specified. [D:\temp\ClassLibrary2\ClassLibrary2.csproj]
    0 Warning(s)
    1 Error(s)

Time Elapsed 00:00:01.62
```

zivkan: Ask me how I knew that MSBuild doesn't load assemblies from PublicAssemblies or PrivateAssemblies. Go on, I dare you 😁 

Yeah, unfortunately we have to make our MSBuild tasks entirely "self contained". (or in our msbuild task code, I guess we could subscribe to the app domain assembly load event, and redirect specific assemblies individually? Though, that might break restore on mono 🤷‍♂️).

I think if we make sure the versions of the dependencies we use are the same, or lower, than in public/private assemblies, then VS's binding redirects should load it from there, so we shouldn't get multiple copies of the same assembly in memory, unless we use a higher version of the assemblies.

But I never looked into why the previous attempt that this fix failed RPS, so I don't know what to look for in this PR to avoid a repeat of the previous attempt.

jeffkl: It was a great catch.  I'm starting to think our build task should be ILMerge'd and in a different directory while all of the actual assemblies our VSIX needs should be in the same old spot.  It feels like we're swimming upstream with our task assembly in here.

zivkan: ilmerge is abandonware. we should not use it for anything. One reason is that it doesn't support newer features like "portable" pdbs or "public signing", which causes additional technical debt on us. I will actively resist any attempt to increase use of ilmerge. The only time I want to read the name ilmerge is in PRs that remove it from our repo. Unfortunately, I see no alternative for nuget.exe, but in my opinion nothing else should use it.

Yes, maybe we need to split our VS extension and msbuild tasks into different (sub)directories (but no need for ilmerge), although I'm not sure that'll actually help us in this situation, unless we also duplicate all the assemblies that are shared, but then our VSIX size will bloat.

I can't see an ideal/perfect solution 😞

test/TestUtilities/Test.Utility/TestDotnetCLiUtility.cs
```diff
@@ -192,36 +195,39 @@ private static void UpdateCliWithLatestNuGetAssemblies(string cliDirectory)
 
         private static void CopyRestoreArtifacts(string artifactsDirectory, string pathToSdkInCli, string configuration)
         {
-            const string restoreProjectName = "NuGet.Build.Tasks";
-            const string restoreTargetsName = "NuGet.targets";
-            const string restoreTargetsExtName = "NuGet.RestoreEx.targets";
+            var fileExtensions = new[] { ".dll", ".pdb", ".targets" };
 
-            var sdkDependencies = new List<string> { restoreProjectName, "NuGet.Versioning", "NuGet.Protocol", "NuGet.ProjectModel", "NuGet.Packaging", "NuGet.LibraryModel", "NuGet.Frameworks", "NuGet.DependencyResolver.Core", "NuGet.Configuration", "NuGet.Common", "NuGet.Commands", "NuGet.CommandLine.XPlat", "NuGet.Credentials", "NuGet.Build.Tasks.Console" };
+            var sdkDependencies = new List<string> { "NuGet.Build.Tasks.Console", "NuGet.CommandLine.XPlat" };
```
heng-liu, line 200: May I know why are the dependencies reduced into two? Is that because the two assemblies are now ilmerged?
If so, in the real .NET SDK insertion scenario, are the copied NuGet assemblies also get reduced? Thanks!

test/TestUtilities/Test.Utility/TestDotnetCLiUtility.cs
```diff
@@ -192,36 +195,39 @@ private static void UpdateCliWithLatestNuGetAssemblies(string cliDirectory)
...
+                        file.CopyTo(Path.Combine(pathToSdkInCli, file.Name), overwrite: true);
+                    }
+                }
+
+                // Update NuGet.CommandLine.XPlat.runtimeconfig.json so that our dependencies are loaded correctly
+                FileInfo runtimeConfigJsonFile = new FileInfo(Path.Combine(pathToSdkInCli, $"{projectName}.runtimeconfig.json"));
+
+                JObject jObject = runtimeConfigJsonFile.Exists ? JsonUtility.LoadJson(runtimeConfigJsonFile.OpenText()) : new JObject();
+
+                jObject["runtimeOptions"]["configProperties"]["Microsoft.NETCore.DotNetHostPolicy.SetAppPaths"] = true;
```
heng-liu, line 222: May I know what's this flag for? Is there any document about it? Thanks!

src/NuGet.Clients/NuGet.VisualStudio.Client/.vsixignore
jeffkl: Do you know why our Pack task is ILMerged?  I'd love to know more history about that one.

One way to get NuGet.exe to not be ILMerged is to actually compile the sources directly into it which would be a fair bit of work but totally possible.  Another way would be to have it be a self-extracting ZIP that then runs the expanded EXE with the DLLs next to it.  

The only other way I can think of to avoid shipping these is to update `MSBuild.exe.config` to find these assemblies in the shared location.

nkolev92: > Do you know why our Pack task is ILMerged? I'd love to know more history about that one.

At some point it was causing assembly loading issues as the pack tasks don't have to match what the restore tasks carry, as pack can be installed as a separate SDK/package. 

That was so long ago, that I haven't found much details on GH, so at this point I don't remember the exact scenario so I can't really retest. 

Thinking through the scenario, I think we might be safe to stop ilmerging, but that's just a potentially risky change that hasn't been prioritized.

jeffkl:
> Could you be able to run RPS test just in case? Last time it was blocked for VS insertion because of RPS test failure.

I am doing this now and if it passes I'll reply to all the other comments

jeffkl:
RPS passed: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/436688

build/common.ps1
jeffkl: I started to not be able to run unit tests locally because .NET 5 is removed by Visual Studio 2022 17.4 since its EOL.  Our unit tests are generally built for `netcoreapp3.1`, `net5.0` and `net472` so I figured we might as well install the `net5.0` runtime since we're already installing `netcoreapp3.0`.  Should I do this in different change?

src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.csproj
jeffkl: Yeah I'm not even sure how I found this.  For `netstandard2.0`, here is the difference between `$(OutputPath)` and `$(OutputPath)\publish`:

```diff
+Microsoft.Extensions.FileProviders.Abstractions.dll
+Microsoft.Extensions.FileSystemGlobbing.dll
+Microsoft.Extensions.Primitives.dll
+Newtonsoft.Json.dll
NuGet.Build.Tasks.Pack.dll
NuGet.Commands.dll
NuGet.Common.dll
NuGet.Configuration.dll
NuGet.Credentials.dll
NuGet.DependencyResolver.Core.dll
NuGet.Frameworks.dll
NuGet.LibraryModel.dll
NuGet.Packaging.dll
NuGet.ProjectModel.dll
NuGet.Protocol.dll
NuGet.Versioning.dll
+System.Buffers.dll
+System.Formats.Asn1.dll
+System.Memory.dll
+System.Numerics.Vectors.dll
+System.Runtime.CompilerServices.Unsafe.dll
+System.Security.Cryptography.Cng.dll
+System.Security.Cryptography.Pkcs.dll
+System.Security.Cryptography.ProtectedData.dll
```

For `net472` folders are the same.  I'm not sure how it was working before, do you want me to dive more into it and figure out what's going on in the shipped .NET SDK?

test/TestUtilities/Test.Utility/TestDotnetCLiUtility.cs
jeffkl: I'm not 100% sure but I'm assuming its because our projects used to be "legacy" C# projects where project references were not transitive.   This meant that if you wanted to collect the outputs of a project and its dependencies, you sometimes needed to get them from each output folder.  However, with SDK-style projects, everything is transitive so the output folder for these two projects contains everything we need.

We are currently inserting three packages/projects into the .NET SDK:

[NuGet.CommandLine.XPlat](https://github.com/dotnet/sdk/blob/main/src/Layout/tool_nuget/tool_nuget.csproj#L10)
[NuGet.Build.Tasks](https://github.com/dotnet/sdk/blob/main/src/Layout/redist/redist.csproj#L26)
[NuGet.Build.Tasks.Console](https://github.com/dotnet/sdk/blob/main/src/Layout/redist/redist.csproj#L27)

Technically `NuGet.Build.Tasks.Console` references `NuGet.Build.Tasks` so the union of the locally built outputs of `NuGet.CommandLine.XPlat` and `NuGet.Build.Tasks.Console` contain what we need to patch an SDK.

This logic does not currently patch the NuGet-based MSBuild project SDK resolver though, perhaps I should add that?  The layout is slightly different so it would need to be special.

[Microsoft.Build.NuGetSdkResolver](https://github.com/dotnet/sdk/blob/main/src/Layout/redist/redist.csproj#L29)

test/TestUtilities/Test.Utility/TestDotnetCLiUtility.cs
jeffkl: It took me a while to find it, I don't think the .NET folks want people knowing about it or using it.  We are only allowed to use it for testing so we're okay though.  I added our team to an email thread about how I found this and what it does.

build/common.ps1
nkolev92: Doesn't that only install the runtime for the test SDK.
As in, locally I expected your project to be using the machine SDK and runtime.

src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.csproj
nkolev92: I'm sure tbh. I'm not sure I have a good grasp on the severity of this.

test/TestUtilities/Test.Utility/TestDotnetCLiUtility.cs
nkolev92: > I'm not 100% sure but I'm assuming its because our projects used to be "legacy" C# projects where project references were not transitive. This meant that if you wanted to collect the outputs of a project and its dependencies, you sometimes needed to get them from each output folder. However, with SDK-style projects, everything is transitive so the output folder for these two projects contains everything we need.

Those projects were never legacy, as they were among the first multi targeted projects in production :D 

I think it's just how the copying was implemented, instead of 2 entry points plus all their dependencies, all assemblies were enumerated.

nkolev92: > This logic does not currently patch the NuGet-based MSBuild project SDK resolver though, perhaps I should add that? The layout is slightly different so it would need to be special.

I think it'd be great if we patch the SDK resolver too and add a test or 2 if at all possible, but that can be a future change.

heng-liu: Got it~ Thanks for the explanation!

build/common.ps1
zivkan: argh 🤦 

https://github.com/NuGet/NuGet.Client/blob/340d447dfc6eaa0cb5d8c25df69fe7f235e37d62/build/build.proj#L389
https://github.com/NuGet/NuGet.Client/blob/340d447dfc6eaa0cb5d8c25df69fe7f235e37d62/build/common.project.props#L34

src/NuGet.Clients/NuGet.CommandLine/NuGet.CommandLine.csproj
```diff
@@ -120,14 +120,6 @@
     <Exec Command="$(IlmergeCommand)" ContinueOnError="false" />
   </Target>
 
-  <Target Name="NuGetLinkVSSDKEmbeddableAssemblies" AfterTargets="ResolveReferences" BeforeTargets="FindReferenceAssembliesForReferences">
-    <ItemGroup>
-      <ReferencePath Condition="'%(FileName)' == 'Microsoft.VisualStudio.Setup.Configuration.Interop'">
-        <EmbedInteropTypes>true</EmbedInteropTypes>
```
zivkan, line 126: I feel like removing this might be risky. I'm not even sure what nuget.exe feature would be using Microsoft.VisualStduio.Setup.Configuration.Interop. But since you added `ExcludeAssets="runtime"`, then that should prevent it from being copied to `bin\`, which means our ILMerge config won't ilmerge it, and since this is removing the embed interop types, then there will be a runtime file not found exception when it needs to load the assembly.

jeffkl: Sorry I completely forgot to call out why this is okay.  I updated us to a newer version of [Microsoft.VisualStudio.Setup.Configuration.Interop](https://github.com/NuGet/NuGet.Client/pull/4931/files#diff-5baf5f9e448ad54ab25a091adee0da05d4d228481c9200518fcb1b53a65d4156R56) because newer versions of the package injects this logic for us:

image

I had remembered to double check but forgot to post anything here, thanks for keeping me honest 😄 

image

build/common.ps1
jeffkl: Ah yes that would explain it.  I'm fine with reverting this for this PR, its up to you.

zivkan (approved):

PR #3421, opened 2020-05-28
Add support for multiple config files in the %APPDATA%\NuGet\config directory

nkolev92:
## Bug

Fixes: https://github.com/NuGet/Home/issues/9394
Regression: No  
* Last working version:   
* How are we preventing it in future:   

## Fix

Details: 

Add the ability to detect additional configs in the user specific directory similar to the machine wide one as described in the issue. 

Given that these are additional configs, they are merged with lower priority than the default user specific config. The rest are merged deterministically in alphabetical order because they are not expected to confict. 
The determinism allows an out in case something really goes wrong. 

There is one concern here, which is what happens if someone already had a config in that folder, but NuGet was never reading them before. 
The answer is we read it. This is a NuGet specific folder so it's extremely unlikely that additional/unrelated configs are found there.

## Testing/Validation

Tests Added: Yes  
Reason for not adding tests:  
Validation:

PR #4939, opened 2022-11-21
Reload config in VS when config files change

### LINKED ISSUE #1538: Reload Visual Studio package sources when nuget.config is modified manually
Visual Studio should reload nuget.config settings after the config files have been manually modified outside of Visual Studio. Today sources are only reloaded when the Tools > Options dialog is used.

This could be achieved by re-initializing settings before package restores and before package operations. Note: SourceRepositories should be kept between reloads to avoid losing cached data such as index.json.

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/1538

## Description

Modify `VsSettings` to have a "user config watcher" and "solution config watcher".

The "user config watcher" watches the user-default nuget.config file, and also `config\*.config` for any relevant changes.

The "solution config watcher" watches solution nuget.config file, and all directories up to the drive root for nuget.config changes.

The "user config watcher" is initialized when VsSettings is constructed, whereas the solution config watcher updates every time the solution directory changes.

reload-NuGet-Config

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Utility/FileWatchers/UserConfigFileWatcher.cs
```diff
@@ -0,0 +1,102 @@
...
+            _userSettingsWatcher.Created += OnFileSystemEvent;
+            _userSettingsWatcher.Changed += OnFileSystemEvent;
+            _userSettingsWatcher.Deleted += OnFileSystemEvent;
+            _userSettingsWatcher.Renamed += OnFileSystemEvent;
+            _userSettingsWatcher.EnableRaisingEvents = true;
+
+            _configDirectoryWatcher = new FileSystemWatcher(configDirectory);
+            _configDirectoryWatcher.Created += OnFileSystemEvent;
+            _configDirectoryWatcher.Changed += OnFileSystemEvent;
+            _configDirectoryWatcher.Deleted += OnFileSystemEvent;
```
erdembayar: Should we monitor config directory given it couldn't be renamed or deleted because of lock file presence?

zivkan: This is monitoring files in the directory, rather than the directory itself (see the tests), but I notice now that I forgot to add a `*.config` filter on the watcher 🤦

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/VSSettings.cs
```diff
@@ -30,33 +34,46 @@ private ISettings SolutionSettings
...
+        public event EventHandler? SettingsChanged;
 
         public VSSettings(ISolutionManager solutionManager)
             : this(solutionManager, machineWideSettings: null)
         {
         }
 
         [ImportingConstructor]
-        public VSSettings(ISolutionManager solutionManager, IMachineWideSettings machineWideSettings)
+        public VSSettings(ISolutionManager solutionManager, IMachineWideSettings? machineWideSettings)
```
nkolev92: Why is only IMachineWideSettings being flagged for nullable? 
Looks like the analyzers are wrong here.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Utility/FileWatchers/SolutionConfigFileWatcher.cs
```diff
@@ -0,0 +1,77 @@
...
+#nullable enable
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using NuGet.Configuration;
+
+namespace NuGet.PackageManagement.VisualStudio.Utility.FileWatchers
+{
+    internal class SolutionConfigFileWatcher : IFileWatcher
```
nkolev92, line 13: I'm slightly concerned about the fact that we have to reimplement the logic for config reading. 

Can we just watch the files we loaded instead? 

I understand that won't handle new file creation, but maybe that's a worthwhile trade off?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/VSSettings.cs
```diff
@@ -30,33 +34,46 @@ private ISettings SolutionSettings
...
+        }
+
+        public VSSettings(ISolutionManager solutionManager, IMachineWideSettings? machineWideSettings, IFileWatcherFactory fileWatcherFactory)
         {
             SolutionManager = solutionManager ?? throw new ArgumentNullException(nameof(solutionManager));
             MachineWideSettings = machineWideSettings;
             SolutionManager.SolutionOpening += OnSolutionOpenedOrClosed;
             SolutionManager.SolutionClosed += OnSolutionOpenedOrClosed;
+
+            _fileWatcherFactory = fileWatcherFactory ?? throw new ArgumentNullException(nameof(fileWatcherFactory));
```
nkolev92, line 64: I imagine we're ignoring the machine wide config on purpose?

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Utility/FileWatchers/SolutionConfigFileWatcherTests.cs
```diff
@@ -0,0 +1,170 @@
...
+using System.Threading;
+using FluentAssertions;
+using NuGet.Configuration;
+using NuGet.PackageManagement.VisualStudio.Utility.FileWatchers;
+using NuGet.Test.Utility;
+using Xunit;
+
+namespace NuGet.PackageManagement.VisualStudio.Test.Utility.FileWatchers
+{
+    public class SolutionConfigFileWatcherTests
```
nkolev92, line 17: Great tests 👏

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/VSSettings.cs
zivkan: No, the current code designed like this:

https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/VSSettings.cs#L43-L52

Given there were no tests for this class before this PR, and zero references to the one parameter constructor, I can't figure out why it was designed like this. I can change it.

nkolev92: Oh now it makes sense. 

Given that it's not an importing constructor, I'm fairly confident it's only being used in tests so changing it would be cheap/free. Up to you whether you make that change.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/VSSettings.cs
```diff
@@ -76,31 +93,58 @@ private bool ResetSolutionSettingsIfNeeded()
...
-                                MessageHelper.ShowErrorMessage(Common.ExceptionUtilities.DisplayMessage(ex), Strings.ConfigErrorDialogBoxTitle);
-                            }
-
-                            return settings;
-
-                        }, NuGetUIThreadHelper.JoinableTaskFactory));
+                var oldSolutionConfigFileWatcher = _solutionConfigFileWatcher;
+                var newSolutionConfigFileWatcher = root == null ? null : _fileWatcherFactory.CreateSolutionConfigFileWatcher(root);
+
+                var exchanged = Interlocked.CompareExchange(ref _solutionConfigFileWatcher, newSolutionConfigFileWatcher, oldSolutionConfigFileWatcher);
```
donnie-msft: use explicit types

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Utility/FileWatchers/SolutionConfigFileWatcherTests.cs
```diff
@@ -0,0 +1,170 @@
...
+            }
+            string slnPath = Path.Combine(testDirectory.Path, "solution.sln");
+            string packagesConfigPath = Path.Combine(testDirectory.Path, "packages.config");
+            string nugetConfigBakPath = Path.Combine(testDirectory.Path, Settings.DefaultSettingsFileName + ".bak");
+
+            // Act
+            File.WriteAllText(slnPath, string.Empty);
+            File.WriteAllText(packagesConfigPath, string.Empty);
+            File.WriteAllText(nugetConfigBakPath, string.Empty);
+            // filesystem watchers are not real time, so we need to give a litle time for all the async file IO to happen
```
donnie-msft:
```suggestion
            // File system watchers are not real time, so we need to give a little time for all the async file IO to happen.
```
(Applies to multiple copies of this comment)

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Utility/FileWatchers/SolutionConfigFileWatcherTests.cs
```diff
@@ -0,0 +1,170 @@
...
+            obtained.Should().BeFalse();
+        }
+
+        [Fact]
+        public void FileChanged_FileChangedAfterDispose_NotificationNotReceived()
+        {
+            // Arrange
+            using TestDirectory testDirectory = TestDirectory.Create();
+            ManualResetEventSlim mre = new(initialState: false, spinCount: 0);
+            using (SolutionConfigFileWatcher target = new(testDirectory.Path))
```
donnie-msft, line 158: Isn't this kind of the Act for this particular test?

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Utility/FileWatchers/UserConfigFileWatcherTests.cs
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        // Simulate two instances of Visual Studio being open at the same time by having two instances of the class at the same time.
+        [Fact]
+        public void ctor_TwoInstances_DoesNotFail()
+        {
+            // Arrange
+            using TestDirectory testDirectory = TestDirectory.Create();
+            var configDirectory = Path.Combine(testDirectory, "config");
+            using UserConfigFileWatcher target1 = new(testDirectory.Path);
```
donnie-msft, line 188:
```suggestion
            // Act & Assert
            using UserConfigFileWatcher target1 = new(testDirectory.Path);
```

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Utility/FileWatchers/SolutionConfigFileWatcherTests.cs
```diff
@@ -0,0 +1,170 @@
...
+            {
+                target.FileChanged += (s, e) => mre.Set();
+            }
+            string slnPath = Path.Combine(testDirectory.Path, "solution.sln");
+            string packagesConfigPath = Path.Combine(testDirectory.Path, "packages.config");
+            string nugetConfigBakPath = Path.Combine(testDirectory.Path, Settings.DefaultSettingsFileName + ".bak");
+
+            // Act
+            File.WriteAllText(slnPath, string.Empty);
```
donnie-msft, line 139: I had a hard time mentally parsing why this one is using `string.Empty`. I checked `WriteAllText` and it doesn't explain what an empty content argument actually does. Is it setting metadata like modified date and that's how this test works?
If it's that subtle, can you just comment that you're just using the fact that empty content is considered a modification, or somehow explain this so it isn't something a reader has to figure out when they look at this test?

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Utility/FileWatchers/SolutionConfigFileWatcherTests.cs
```diff
@@ -0,0 +1,170 @@
...
+            // Arrange
+            using TestDirectory testDirectory = TestDirectory.Create();
+            string configPath = Path.Combine(testDirectory.Path, Settings.DefaultSettingsFileName);
+            File.WriteAllText(configPath, string.Empty);
+            ManualResetEventSlim mre = new(initialState: false, spinCount: 0);
+            using SolutionConfigFileWatcher target = new(testDirectory.Path);
+            target.FileChanged += (s, e) => mre.Set();
+
+            // Act
+            mre.Wait(0).Should().BeFalse();
```
donnie-msft: nit: Seems like an Arrange to me. (same applies to multiple copies of this line)

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/VSSettingsTests.cs
```diff
@@ -0,0 +1,74 @@
...
+using Xunit;
+
+namespace NuGet.PackageManagement.VisualStudio.Test
+{
+    [Collection(MockedVS.Collection)]
+    public class VSSettingsTests : MockedVSCollectionTests
+    {
+        public VSSettingsTests(GlobalServiceProvider globalServiceProvider) : base(globalServiceProvider)
+        {
+            globalServiceProvider.Reset();
```
donnie-msft: `base` already does this

donnie-msft (commented):
Nice PR and feature! 🎉

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Utility/FileWatchers/SolutionConfigFileWatcher.cs
zivkan: I understand the concern. If in some point in the future we change where config files are loaded from, then the VS file watchers need to be updated as well, otherwise there'll be a bug. 

When I worked on this implementation, I made the tradeoff that removing paths from being loaded would be a breaking change, and therefore incredibly unlikely. Also, I'm only aware of a single time when a new path was added (`UserSettingsPath\config\*.config`).  Basically, it's a kind of "public API", and we try hard not to change public APIs much. This allowed me to keep the implementation simple, while still "powerful" (like checking for file creation, and minimizing FileWatcher new instances & disposal).

I'm also wondering if customers creating new nuget.config files in the solution directory is actually uncommon. When I need to test a customer scenario, particularly for non-SDK projects, I create a solution in VS, open the terminal to run `dotnet new nugetconfig`, then immediate reload the solution so it gets loaded.  I can imagine customers who are starting new solutions also benefit from the solution directory file creation watching.

If you feel particularly strongly, please let me know. Otherwise I'd like to hear from other people about their thoughts.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/VSSettings.cs
zivkan: Resolving as a duplicate of https://github.com/NuGet/NuGet.Client/pull/4939#discussion_r1028711850

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Utility/FileWatchers/SolutionConfigFileWatcher.cs
nkolev92: I guess an alternative would be to only watch the solution folder for new files. 

@NuGet/nuget-client 
Any thoughts on this discussion?

If no one else shares my concerns, we can keep the current iteration.

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Utility/FileWatchers/SolutionConfigFileWatcherTests.cs
zivkan: 🤷‍♂️ I don't think there's any obvious correct/wrong answer. Especially since the test is validating that something does not happen

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Utility/FileWatchers/SolutionConfigFileWatcherTests.cs
zivkan: > I checked WriteAllText and it doesn't explain what an empty content argument actually does.

It either creates an empty file, or if the file already exists, it deletes its contents, so it becomes an empty file. If the file is already empty, then I guess that it will just update the last modified timestamp. I'm assuming that operating systems/file systems aren't smart enough to know that a file was "modified" with the same contents it already had and therefore doesn't need to notify file system watchers.

However, since none of these files already exist in the temp directory created for the test (`TestDirectory.Create()`), in this specific scenario, it's creating empty files.

It sounds like the misunderstanding is that you believed the sln file already existed, hence this line is modifying it. But I'm not sure why you'd think the sln file already exists.  We have a different helper `SimpleTestPathContext`, which creates multiple directories, and sets up a `nuget.config` file to use them as the global packages folder, a local package source, etc. But even that helper doesn't create a solution file.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Utility/FileWatchers/SolutionConfigFileWatcher.cs
kartheekp-ms: > I guess an alternative would be to only watch the solution folder for new files.

+1 for this suggestion. Looking at the comments in https://github.com/NuGet/Home/issues/1538, I get the impression that that customers are mainly looking for `NuGet.Config file under the solution root`. For example, [`branch switch, git pull scenario`](https://github.com/NuGet/Home/issues/1538#issuecomment-425571925) Andrew pointed out in the issue. Chances of customers modifying repository level NuGet.Config file is higher than modifying the hierarchy of NuGet.Config files.

In the [docs](https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file), we have made clear that `Add a nuget.config file in the root of your project repository. This is considered a best practice as it promotes repeatability and ensures that different users have the same NuGet configuration. You may need to configure clear elements to ensure no user or machine specific configuration is applied.`

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Utility/FileWatchers/SolutionConfigFileWatcherTests.cs
donnie-msft: I see now. Thx!

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Utility/FileWatchers/UserConfigFileWatcher.cs
erdembayar: What kind of config files are we expecting here? Currently on my local `configDirectory = C:\Users\xxxx\AppData\Roaming\NuGet\config` is empty. Do we need to watch any config file here?
Do mean to watch userSettingsDirectory instead?

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Utility/FileWatchers/SolutionConfigFileWatcherTests.cs
erdembayar: Could `wait` be 1 instead of 0 to avoid random runtime flakiness just in case?

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Utility/FileWatchers/UserConfigFileWatcherTests.cs
```diff
@@ -0,0 +1,192 @@
...
+        [Fact]
+        public void FileChanged_WhenFileCreatedInConfigSubdirectory_NotificationReceived()
+        {
+            // Arrange
+            using TestDirectory testDirectory = TestDirectory.Create();
+            var configDirectory = Path.Combine(testDirectory, "config");
+            using UserConfigFileWatcher target = new(testDirectory.Path);
+            ManualResetEventSlim mre = new(initialState: false, spinCount: 0);
+            target.FileChanged += (s, e) => mre.Set();
+            string configPath = Path.Combine(configDirectory, "contoso.config");
```
erdembayar, line 111: I still don't understand why we need config directory in user config file watcher given by default it points to NuGetFolderPath.UserSettingsDirectory.

zivkan: In NuGet 5.7, user `config\*.config` support was added. Have a look at:
* https://learn.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#additional-user-wide-configuration
* https://github.com/NuGet/docs.microsoft.com-nuget/commit/fcf9f7d3b4f5768ca26b5f840e330b22b21e799b
* https://github.com/NuGet/NuGet.Client/pull/3421
* https://github.com/NuGet/Home/issues/9394

However, due to this thread: https://github.com/NuGet/NuGet.Client/pull/4939#discussion_r1030766127, it seems people don't want user config to be watched 🤷‍♂️

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Utility/FileWatchers/SolutionConfigFileWatcher.cs
zivkan: Ok, I've changed it.

Keep in mind that if a customer has two instances of VS open, and they use VS to change package sources, which gets persisted in the user-config directory (rather than a solution nuget.config), this means that the second instance of VS is not going to automatically get the changes, and the customer will still need to reload the solution.

Also, I can't remember where I saw this directory layout, but I've seen at least one repo that puts the solution file in the repo `src\` directory, but the `nuget.config` is still in the repo root. When branch switching that repo, this PR will no longer detect the `nuget.config` file change.

It's in an isolated commit, so if we change our mind, we can revert that commit to get the "full" behaviour back, assuming we're ok with "duplicating" file location logic.

nkolev92: > Also, I can't remember where I saw this directory layout, but I've seen at least one repo that puts the solution file in the repo src\ directory, but the nuget.config is still in the repo root. When branch switching that repo, this PR will no longer detect the nuget.config file change.

This makes me change my mind about it :D 

Is there a perf hit from watching multiple directories? If not, then our concerns are moot. 

I'd love for this to be a topic we don't have to revisit :)

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Utility/FileWatchers/SolutionConfigFileWatcher.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            var dotNuGetPath = Path.Combine(solutionDirectory, NuGetConstants.NuGetSolutionSettingsFolder);
+            if (Directory.Exists(dotNuGetPath))
+            {
+                _dotNuGetWatcher = Create(dotNuGetPath);
+            }
+
+            FileSystemWatcher Create(string path)
+            {
+                var watcher = new FileSystemWatcher(path, Settings.DefaultSettingsFileName);
```
erdembayar, line 38: What happens here if solution dir doesn't have any nuget.config file at all?

erdembayar, line 38: I can see we have [NuGetDefaults.Config](https://learn.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#nuget-defaults-file) file too. Do we need to watch it too?

kartheekp-ms: The file watcher also raises an event when a solution level NuGet.Config file is created so that settings can be refreshed.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Utility/FileWatchers/UserConfigFileWatcher.cs
zivkan: same question & answer as https://github.com/NuGet/NuGet.Client/pull/4939#discussion_r1033228718.  Resolving this comment too.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Utility/FileWatchers/SolutionConfigFileWatcher.cs
zivkan: Yes, that's right. In fact, the subscription to created events is on the very next line: 
https://github.com/NuGet/NuGet.Client/blob/e386036500c00560b2fe377b99caee61fa9cad89/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Utility/FileWatchers/SolutionConfigFileWatcher.cs#L35

zivkan: `NuGetDefaults.Config` is not in a user-writable location, it needs admin access to modify. So, there are no actions in Visual Studio that will modify the file (branch change, or using the package source or package source mapping options windows). Similarly, `dotnet nuget` and `nuget.exe` commands won't edit the file unless the customer both elevates to admin and also uses `--config-file` to explicitly tell nuget to modify that specific file.

So, the scenarios where the file will be modified are very unlikely (frequency of nearly zero), and I believe customers are more accepting that programs need to be restarted when modifying machine wide state.

Although having written all this, I assume that `NuGetDefaults.config` is the preferred way for admins to force certain compliance via group policy. Therefore, in theory a group policy update could update the file while VS is running. On the other hand, NuGet allows users to override any NuGetDefaults.config with their own user-writable nuget.config, so the group policy compliance argument falls flat, in my opinion.

What do others think?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Utility/FileWatchers/SolutionConfigFileWatcher.cs
zivkan: > Is there a perf hit from watching multiple directories?

The only perf issue related to file system watchers I'm aware of are:

1. being too slow to process events, causing the watcher's queue to overflow, and then events are lost
2. multiple events coming in causing the event handler to run multiple times (need to "debounce" the event).

However, in my opinion neither is a concern for us, because:
1. all we do is set up a new lazy, which is fairly fast. Honestly, it might be just as fast as adding a message to a concurrent queue, so I'm not sure we could meaningfully make the handler faster.
2. our handler is idempotent
3. the probability of our watch filter triggering a large number of events is low
4. regardless of which filenames trigger the event, the action is always the same. We don't do action 1 for some files and action 2 for different files.

I know that on Linux (not sure about Mac), there is a limit to the number of file system watchers allowed (fs.inotify.max_user_watches), but I've never heard of such a case with Windows, nor can I quickly find any evidence of any limits on Windows. And the code in this PR only runs on Windows. A long time ago, Windows did have a low max of total handles that could be opened. I don't know if that limit was significantly increased to the point of irrelevance when Windows became 32-bit or 64-bit, or maybe the limit was independent of the OS's bit-ness. But since VS2022 is 64-bit only, I'm confident we don't have to worry about handle count.

So, all this to say that I'm not aware of any perf issues with many file system watchers that rarely trigger any events.

nkolev92: Thanks for the details. 

I then think we can move back to the "do it all" approach. 
I don't have concerns with that.

erdembayar (dismissed):
LGTM. :shipit:

nkolev92 (dismissed):

zivkan:
The failing test exposes a nasty deadlock. It's not exactly a realistic user scenario, but it's also not impossible. What it requires is for a FileSystemWatcher event to cause VsSettings to reload settings, but for the load to fail (either because the nuget.config is invalid XML, or the file is just in use), so VsSettings will show a MessageBox and therefore prevent VsSettings' AsyncLazy from finishing evaluation, hence FileSystemWatcher's callback is still on the call stack, which means that it's internal `lock(this)` is still locked. Somehow, in our test programatically from a background thread, something else calls VS API to close the solution, so that `VsSettings.ResetSolutionSettingsIfNeeded` is called, and tries to dispose `oldSolutionConfigFileWatcher`. FileSystemWatcher's `Dispose` also wants to `lock(this)`. 

While the FileSystemWatcher callback is on a thread pool thread (or maybe even an IO completion thread), it switches to the UI thread (and blocks) until the MessageBox is closed. Since solution close callback is also on the UI thread and this deadlocks, the UI thread can't handle clicking the Ok button on the MessageBox in order to complete VsSetting's AsyncLazy and unblock everything.

In case that's not very clear, here's a significantly shortened summary of the callstack:

```text
FileSystemWatcher.OnCreated
VsSettings.OnConfigFileChanged
PackageSourceProvider.OnPackageSourcesChanged
PMC.OnPackageSourcesChanged
PackageSourceProvider.LoadPackageSources
VsSettings.GetSettings
AsyncLazy<>.GetValueAsync
VsSettings.ResetSolutionSettings.AnonymousMethod
VsShellUtilities.ShowMessageBox
HwndWrapper.WndProc/Dispatcher.WndProcHook
VSSolutionHelper.CloseSolutionAsync
Solution.NotifyOnAfterCloseSolution
VSSolutionManager.OnAfterClosing
VSSettings.OnSolutionOpenedOrClosed
VSSettings.ResetSolutionSettingsIfNeeded
FileSystemWatcher.Dispose
```

Both `FileSystemWatcher.OnCreated` and `FileSystemWatcher.Dispose` try to `lock(this)`, so there's a deadlock. 

As an aside, if this test was in Apex, or anything else that runs via vstest.exe and has a test timeout, the VS testing infrastructure could have recorded a screen recording, and possibly memory dump, which would have made this much easier and quicker to debug. Our customer powershell based E2E tests made this a lot harder.

test/EndToEnd/tests/ServicesTest.ps1
```diff
@@ -881,10 +881,6 @@ function Test-CreateVsPathContextWithConfiguration {
 	$settingFileContent -f $userPackageFolder, $fallbackPackageFolderA, $fallbackPackageFolderB `
 	    | Out-File -Encoding "UTF8" $settingFile
 
-	SaveAs-Solution($solutionFile)
```
erdembayar, line 884: Is it ok to remove this part? It could be useful to catch other bugs in the future.

zivkan: What kind of bugs do you have in mind? Do we already have other tests that test it?  [There are other tests that still use `SaveAs-Solution`](https://github.com/search?q=repo%3ANuGet%2FNuGet.Client%20SaveAs-Solution&type=code).

The name of this test is "CreateVsPathContextWithConfiguration". There's nothing in the name, or feature, of the test name that suggests it needs to close and then re-open the solution.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/VSSettings.cs
```diff
@@ -76,29 +88,90 @@ private bool ResetSolutionSettingsIfNeeded()
...
-                            ISettings settings = null;
                             try
                             {
-                                settings = Settings.LoadDefaultSettings(root, configFileName: null, machineWideSettings: MachineWideSettings);
+                                loadedSettings = Settings.LoadDefaultSettings(solutionDirectory, configFileName: null, machineWideSettings: MachineWideSettings);
                             }
+                            // If any config files are in use, retry after a short delay
+                            // Would be nice if there was a better way to detect file busy: https://github.com/dotnet/runtime/issues/79643
                             catch (NuGetConfigurationException ex)
+                                when (ex.InnerException?.GetType() == typeof(IOException)
```
dominoFire, line 148: Do you want to catch IOException derived classes?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/VSSettings.cs
```diff
@@ -76,29 +88,90 @@ private bool ResetSolutionSettingsIfNeeded()
...
+                                && (uint)ex.InnerException.HResult == WindowsSharingViolationHResult
```
dominoFire, line 149: Why do you need to cast to unit? Can you save the cast by comparing against an `int` and defining `WindowsSharingViolationHResult` as a `const int` ?

```suggestion
                                && ex.InnerException.HResult == WindowsSharingViolationHResult
```

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Utility/FileWatchers/UserConfigFileWatcher.cs
```diff
@@ -0,0 +1,102 @@
...
+
+namespace NuGet.PackageManagement.VisualStudio.Utility.FileWatchers
+{
+    /// <summary>Watches a user settings directory to detect changes to nuget.config and config/*.config.</summary>
+    internal class UserConfigFileWatcher : IFileWatcher
+    {
+        private bool _disposed;
+        private FileSystemWatcher _userSettingsWatcher;
+        private FileSystemWatcher _configDirectoryWatcher;
+        private FileStream _lockFile;
```
kartheekp-ms: These 3 variables can be read-only, I think.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Utility/FileWatchers/UserConfigFileWatcher.cs
```diff
@@ -0,0 +1,102 @@
...
+                Directory.CreateDirectory(configDirectory);
+            }
+            _lockFile = new FileStream(lockFile, FileMode.OpenOrCreate, FileAccess.Read, FileShare.ReadWrite);
+
+            _userSettingsWatcher = new FileSystemWatcher(userSettingsDirectory, Settings.DefaultSettingsFileName);
+            _userSettingsWatcher.Created += OnFileSystemEvent;
+            _userSettingsWatcher.Changed += OnFileSystemEvent;
+            _userSettingsWatcher.Deleted += OnFileSystemEvent;
+            _userSettingsWatcher.Renamed += OnFileSystemEvent;
+            _userSettingsWatcher.EnableRaisingEvents = true;
```
kartheekp-ms: Please consider extracting adding/removing delegate handlers' logic into separate methods.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/VSSettings.cs
```diff
@@ -30,33 +34,41 @@ private ISettings SolutionSettings
...
 
-        private ISolutionManager SolutionManager { get; set; }
+        private ISolutionManager SolutionManager { get; }
 
-        private IMachineWideSettings MachineWideSettings { get; set; }
+        private IMachineWideSettings MachineWideSettings { get; }
 
-        public event EventHandler SettingsChanged;
+        private IFileWatcherFactory _fileWatcherFactory;
+        private IFileWatcher _userConfigFileWatcher;
```
kartheekp-ms: These 2 variables can be read-only, I think.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Utility/FileWatchers/SolutionConfigFileWatcher.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            _watchers.Clear();
+
+            GC.SuppressFinalize(this);
+        }
+
+        private void OnFileSystemEvent(object sender, FileSystemEventArgs e)
+        {
+            FileChanged?.Invoke(this, EventArgs.Empty);
+        }
```
kartheekp-ms, line 75: I think we are ignoring `FileSystemEventArgs e` argument here because settings are loaded again irrespective of the file path. I think we can make use of the argument value to let the user know that NuGet has reloaded the settings because of change to [`e.FullPath`](https://learn.microsoft.com/en-us/dotnet/api/system.io.filesystemeventargs.fullpath?view=net-7.0) file. I think we can do this in a follow-up PR also.

kartheekp-ms (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/VSSettings.cs
```diff
@@ -76,29 +88,90 @@ private bool ResetSolutionSettingsIfNeeded()
...
+                    catch (NuGetConfigurationException ex)
+                    {
+                        settings = NullSettings.Instance;
+
+                        // Show the message box in a different task that does not block this AsyncLazy's GetValueAsync. For more details, see:
+                        // https://github.com/NuGet/NuGet.Client/pull/4939#issuecomment-1351481367
+                        NuGetUIThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+                        {
+                            await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+                            MessageHelper.ShowErrorMessage(Common.ExceptionUtilities.DisplayMessage(ex), Strings.ConfigErrorDialogBoxTitle);
```
dominoFire, line 167: If showing this error message can cause a deadlock, can we remove it and report the error somewhere else, such as Activity Log or VS Status Bar?

dominoFire (dismissed):
Signing off. Consider the case when there are a lot of changes due to scripted nuget.config changes. An 'escape hatch' similar to package icon loading or a way to disable file watching (feature flag) can be useful in this case.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Utility/FileWatchers/SolutionConfigFileWatcher.cs
zivkan: I started with this as the implementation, but once I started writing tests, and thinking about all the cases to achieve (close to) 100% coverage, it added a lot of work for zero benefit, since the code that subscribes to IFileWatcher events wouldn't use the filename, even if it was in the event args. So, borrowing from the YAGNI principal, we just don't need it.

As we advance our diagnosability capabilities, I think it would be worthwhile having an ETW event, so that "report a problem" traces can tell us that config was reloaded. But that can be done within the two watcher implementations, it doesn't have to propagate the filename via event args.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/VSSettings.cs
zivkan: no, "file in use" exceptions are always `IOException`.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/VSSettings.cs
zivkan: Because casting 0xF0070020 to `int` doesn't work (this surprised me, I thought it was possible), and it was easier than calculating what the int value is in signed base-10. Plus, keeping the hex value makes it easier to figure out what the value is, why it has the value it does. Converting to signed base 10 just to remove a cast in my opinion obscures the value and therefore reduces readability.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/VSSettings.cs
zivkan: I didn't feel comfortable changing the error experience for a PR that has a different primary focus. Mitigating the deadlock wasn't much effort, certainly much less effort than changing it to report an error in another way.

I certainly believe we should improve how config errors are handled, but I don't think that scope creep in other issues (even when related) is the best way to achieve it.

dominoFire (approved):
Signing off.

PR #4944, opened 2022-11-22
Forward CancellationToken to methods that take one

martinrrm:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/1871

## Description

Forward the CancellationToken parameter to methods that take one

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/SourceRepositoryDependencyProvider.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/SourceRepositoryDependencyProvider.cs
@@ -346,7 +346,7 @@ namespace NuGet.Commands
 
                 if (_throttle != null)
                 {
-                    await _throttle.WaitAsync();
+                    await _throttle.WaitAsync(cancellationToken);
                 }
 
                 // Read package info, this will download the package if needed.
@@ -440,7 +440,7 @@ namespace NuGet.Commands
 
                 if (_throttle != null)
                 {
-                    await _throttle.WaitAsync();
+                    await _throttle.WaitAsync(cancellationToken);
                 }
 
                 cancellationToken.ThrowIfCancellationRequested();
@@ -609,7 +609,7 @@ namespace NuGet.Commands
             {
                 if (_throttle != null)
                 {
-                    await _throttle.WaitAsync();
+                    await _throttle.WaitAsync(cancellationToken);
                 }
                 if (_findPackagesByIdResource == null)
                 {
--- a/src/NuGet.Core/NuGet.Packaging/PackageExtraction/StreamExtensions.cs
+++ b/src/NuGet.Core/NuGet.Packaging/PackageExtraction/StreamExtensions.cs
@@ -5,12 +5,20 @@ using System;
 using System.Buffers;
 using System.IO;
 using System.IO.MemoryMappedFiles;
+using System.Threading.Tasks;
+using System.Threading;
 using NuGet.Common;
 
 namespace NuGet.Packaging
 {
     public static class StreamExtensions
     {
+        public static readonly int BufferSize = 8192;
+        public static async Task CopyToAsync(this Stream stream, Stream destination, CancellationToken token)
+        {
+            await stream.CopyToAsync(destination, BufferSize, token);
+        }
+
         public static string CopyToFile(this Stream inputStream, string fileFullPath)
         {
             return Testable.Default.CopyToFile(inputStream, fileFullPath);
--- a/src/NuGet.Core/NuGet.Packaging/PackageExtraction/StreamExtensions.cs
+++ b/src/NuGet.Core/NuGet.Packaging/PackageExtraction/StreamExtensions.cs
@@ -5,20 +5,12 @@ using System;
 using System.Buffers;
 using System.IO;
 using System.IO.MemoryMappedFiles;
-using System.Threading.Tasks;
-using System.Threading;
 using NuGet.Common;
 
 namespace NuGet.Packaging
 {
     public static class StreamExtensions
     {
-        public static readonly int BufferSize = 8192;
-        public static async Task CopyToAsync(this Stream stream, Stream destination, CancellationToken token)
-        {
-            await stream.CopyToAsync(destination, BufferSize, token);
-        }
-
         public static string CopyToFile(this Stream inputStream, string fileFullPath)
         {
             return Testable.Default.CopyToFile(inputStream, fileFullPath);
--- a/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpCacheUtility.cs
+++ b/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpCacheUtility.cs
@@ -73,7 +73,11 @@ namespace NuGet.Protocol
                 FileShare.None,
                 BufferSize))
             {
+#if NETCOREAPP2_0_OR_GREATER
+                using (var networkStream = await response.Content.ReadAsStreamAsync(cancellationToken))
+#else
                 using (var networkStream = await response.Content.ReadAsStreamAsync())
+#endif
                 {
                     await networkStream.CopyToAsync(fileStream, BufferSize, cancellationToken);
                 }
--- a/build/NuGet.ruleset
+++ b/build/NuGet.ruleset
@@ -10,6 +10,9 @@
   <Rules AnalyzerId="Microsoft.CodeAnalysis.CSharp.CodeStyle" RuleNamespace="Microsoft.CodeQuality.Analyzers">
     <Rule Id="IDE0055" Action="Warning" />
   </Rules>
+  <Rules AnalyzerId="Microsoft.CodeAnalysis.CSharp.NetAnalyzers" RuleNamespace="Microsoft.CodeAnalysis.CSharp.NetAnalyzers">
+    <Rule Id="CA2016" Action="Warning" />
+  </Rules>
   <Rules AnalyzerId="Microsoft.CodeAnalysis.NetAnalyzers" RuleNamespace="Microsoft.CodeAnalysis.NetAnalyzers">
     <Rule Id="CA3061" Action="Warning" />
     <Rule Id="CA3076" Action="Warning" />
--- a/src/NuGet.Core/NuGet.Common/ConcurrencyUtilities.cs
+++ b/src/NuGet.Core/NuGet.Common/ConcurrencyUtilities.cs
@@ -87,14 +87,14 @@ namespace NuGet.Common
 
                             // This can occur when the file is being deleted
                             // Or when an admin user has locked the file
-                            await Task.Delay(SleepDuration);
+                            await Task.Delay(SleepDuration, token);
                             continue;
                         }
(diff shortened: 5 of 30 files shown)


src/NuGet.Core/NuGet.Packaging/PackageExtraction/StreamExtensions.cs
```diff
@@ -5,12 +5,20 @@
...
+using System.Threading.Tasks;
+using System.Threading;
 using NuGet.Common;
 
 namespace NuGet.Packaging
 {
     public static class StreamExtensions
     {
+        public static readonly int BufferSize = 8192;
+        public static async Task CopyToAsync(this Stream stream, Stream destination, CancellationToken token)
```
nkolev92: Can we compile this conditionally only for the frameworks that need it?
For example, .NET 5 has it, no need for us to declare extensions for that.

src/NuGet.Core/NuGet.Packaging/PackageExtraction/StreamExtensions.cs
```diff
@@ -5,12 +5,20 @@
...
 using System.IO.MemoryMappedFiles;
+using System.Threading.Tasks;
+using System.Threading;
 using NuGet.Common;
 
 namespace NuGet.Packaging
 {
     public static class StreamExtensions
     {
+        public static readonly int BufferSize = 8192;
```
nkolev92: Please refer to https://github.com/NuGet/NuGet.Client/blob/dev/docs/nuget-sdk.md#changes-to-apis. 

We should be mindful about adding public types and methods. 
Is there a benefit to making buffer size public?

src/NuGet.Core/NuGet.Protocol/HttpSource/HttpCacheUtility.cs
```diff
@@ -73,7 +73,11 @@ public static class HttpCacheUtility
                 FileShare.None,
                 BufferSize))
             {
+#if NETCOREAPP2_0_OR_GREATER
```
nkolev92, line 76: In some other places you declared an extension method. 
Should we follow the same approach here?

nkolev92 (commented):
A  couple of questions, primarily revolving around public APIs.

Looks good overall!

src/NuGet.Core/NuGet.Protocol/HttpSource/HttpCacheUtility.cs
martinrrm: I tried doing the same thing, but I wasn't able to succeed. I started a thread in the Teams channel and Kartheek commented that maybe this is the best solution 

> ReadStreamAsync() don't have any overload method that accepts cancellation token in .NET Framework 472. We may have to live with If else suggestion you mentioned above. Even VS repo has If else thing for `ReadStreamAsync` [Utilities.cs - Repos (visualstudio.com)](https://devdiv.visualstudio.com/DevDiv/_git/VS?path=/src/env/shell/PackageFramework/Current/Shell/Utilities.cs&version=GBmain&_a=contents)

nkolev92: I'm not sure I understand why this instance is different from other ones.

kartheekp-ms: [Stream.CopyToAsync()]( https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.copytoasync?view=netframework-4.7.2#system-io-stream-copytoasync(system-io-stream-system-int32-system-threading-cancellationtoken)) has an overloaded method that accepts `CancellationToken` in .NET Framework 472 whereas [HttpContent.ReadStreamAsync()](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpcontent.readasstreamasync?view=netframework-4.7.2) does not have any overloaded methods in .NET Framework 472. Hence, I suggested @martinrrm to add extension method for `Stream.CopyToAsync()` and use conditional compilation symbols for `HttpContent.ReadStreamAsync()`

nkolev92: Unfortunately I don't follow.

Maybe a matrix would help: 

|  | net472 | netstandard2.0 | net5.0 |
|-|---------|------------------|---------|
| Stream.CopyToAsync | | | | 
| HttpContent.ReadStreamAsync | | | | 

You can always add more rows for methods.

src/NuGet.Core/NuGet.Packaging/PackageExtraction/StreamExtensions.cs
kartheekp-ms: `Stream.CopyToAsync()` has an overloaded method that accepts `CancellationToken` both in .NET 472 and .NET 5. The only difference is .NET 5 has [CopyToAsync(Stream, CancellationToken)](https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.copytoasync?view=net-5.0#system-io-stream-copytoasync(system-io-stream-system-threading-cancellationtoken)) but .NET 472 does not have this overload. Similar approach followed in `NuGet.Protocol.StreamExtensions`.
https://github.com/NuGet/NuGet.Client/blob/dcb355d1a989be369e5ac3e3cd6eeea449a4b5eb/src/NuGet.Core/NuGet.Protocol/Utility/StreamExtensions.cs#L16-L19

nkolev92: What's your proposal in this case @kartheekp-ms 

I'm not sure what you said makes my suggestion invalid.

kartheekp-ms: Your suggestion is valid. We can add a conditional compilation so that the extension method is called just for `.NET Framework 472`. I tried to explain the differences in the API between .NET 5 and .NET Framework and the current approach that we follow in `NuGet.Protocol` library.

nkolev92: I see. 

My idea is that by not making it part of the public API, if at some point we migrate away from netstandard2.0 and net472, we can simply not have the redundant extension method anymore.

zivkan: +1 to not making it public. `NuGet.Packaging` is a library for dealing with nupkgs.  Having methods for `Syste.IO.Stream` is out of scope. I'd like to make the other APIs internal, but that's a breaking change.

src/NuGet.Core/NuGet.Commands/RestoreCommand/SourceRepositoryDependencyProvider.cs
```diff
@@ -346,7 +346,7 @@ public class SourceRepositoryDependencyProvider : IRemoteDependencyProvider
 
                 if (_throttle != null)
                 {
-                    await _throttle.WaitAsync();
+                    await _throttle.WaitAsync(cancellationToken);
```
zivkan, line 349: The linked issue makes it look like this PR is supposed to resolve a bunch of `CA2016` warnings (which our build would promote to errors). Shouldn't there also be a removal of `CA2016` from `NoWarn`, or `GlobalSuppressions.cs`?

src/NuGet.Core/NuGet.Packaging/PackageExtraction/StreamExtensions.cs
nkolev92: If it needs to be shared in more than 1 project, I like https://github.com/NuGet/NuGet.Client/tree/dev/build/Shared for those purposes :D
Especially if we add conditional compilation.

src/NuGet.Core/NuGet.Protocol/HttpSource/HttpCacheUtility.cs
kartheekp-ms: @martinrrm - Can you please add the details in the table so that it would be easy for review?

src/NuGet.Core/NuGet.Commands/RestoreCommand/SourceRepositoryDependencyProvider.cs
kartheekp-ms: `CA2016` was never enabled in the repo and that's why IDE showed these warnings as `Info` messages (please see screenshot in the linked issue). Hence you don't see these warnings being suppressed in the `GlobalSuppressions.cs`.

zivkan: Then this PR should enable CA2016, so that after it's merged people don't introduce new violations?

Code changes should not only resolve the linked issue, but try to prevent regressions as well. That's what tests are for, in regards to changes in product behaviour. Therefore in the same spirit, these changes, which might not change product behaviour, need to consider other means to avoid future regressions.

src/NuGet.Core/NuGet.Protocol/HttpSource/HttpCacheUtility.cs
martinrrm: Sorry for the delay in the response. I'm not sure what is expected in the matrix but this is the information I know.

|  | net472 | netstandard2.0 | net5.0 |
|-|---------|------------------|---------|
| Stream.CopyToAsync | No overload with `CancellationToken` only, but there is a overload with `CopyToAsync(destination, bufferSize, CancellationToken.None)` | No overload with `CancellationToken` only, but there is a overload with `CopyToAsync(destination, bufferSize, CancellationToken.None)` | Yes overload with `CancellationToken` | 
| HttpContent.ReadStreamAsync | No overload with `CancellationToken` | No overload with `CancellationToken` | Yes overload with `CancellationToken` | 

After doing this matrix and with the comments from the above discussion I think is better to do a conditional compile and avoid adding new API's.

src/NuGet.Core/NuGet.Commands/RestoreCommand/SourceRepositoryDependencyProvider.cs
martinrrm: I enabled the rule in the ruleset and now is a `Warning`

src/NuGet.Core/NuGet.Protocol/HttpSource/HttpCacheUtility.cs
nkolev92: Looks like you're using conditional compilation everywhere. 

I'm fine with that as it's consistent. 

My feedback was primarily that we either use a conditional compilation at the site of the offending code or we declare an extension method that's conditionally compiled.

nkolev92 (dismissed):

erdembayar (approved):

jeffkl (approved):

PR #4951, opened 2022-11-28
Cache solution dir evaluation

erdembayar:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/1893

## Description
Currently getting solution dir need VS solution still be open, but we have some processing([VsDeleteOnRestartManager.DeleteMarkedPackageDirectoriesAsync](https://github.com/NuGet/NuGet.Client/blob/84d99ce064a5f70628c89af9f30693581e84dc37/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSDeleteOnRestartManager.cs#L168)) works which happen when VS close the solution, due to some race condition [it couldn't access](https://github.com/NuGet/NuGet.Client/blob/84d99ce064a5f70628c89af9f30693581e84dc37/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSDeleteOnRestartManager.cs#L174) the `solution dir` property for solution just closed. It waits for it on main thread, then blocks other thread pool threads. 
We could cache the value and use later even solution is closed for above method, of course we reset to `null` at end of solution closing to prevent cross solution contamination. 
I believe we gain some perf improvement considering this is hot path and [we don't switch to UI thread often](https://github.com/NuGet/NuGet.Client/blob/84d99ce064a5f70628c89af9f30693581e84dc37/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSSolutionManager.cs#L441), only once.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSSolutionManager.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSSolutionManager.cs
@@ -434,15 +434,34 @@ namespace NuGet.PackageManagement.VisualStudio
             });
         }
 
-        public string SolutionDirectory => NuGetUIThreadHelper.JoinableTaskFactory.Run(GetSolutionDirectoryAsync);
+        private string _solutionDir;
+        public string SolutionDirectory
+        {
+            get
+            {
+                if (!string.IsNullOrEmpty(_solutionDir))
+                {
+                    return _solutionDir;
+                }
+
+                _solutionDir = NuGetUIThreadHelper.JoinableTaskFactory.Run(GetSolutionDirectoryAsync);
+                return _solutionDir;
+            }
+        }
 
         public async Task<string> GetSolutionDirectoryAsync()
         {
+            if (!string.IsNullOrEmpty(_solutionDir))
+            {
+                return _solutionDir;
+            }
+
             await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
             var vsSolution = await _asyncVSSolution.GetValueAsync();
             if (IsSolutionOpenFromVSSolution(vsSolution))
             {
-                return (string)GetVSSolutionProperty(vsSolution, (int)__VSPROPID.VSPROPID_SolutionDirectory);
+                _solutionDir = (string)GetVSSolutionProperty(vsSolution, (int)__VSPROPID.VSPROPID_SolutionDirectory);
+                return _solutionDir;
             }
             return null;
         }
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSSolutionManager.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSSolutionManager.cs
@@ -544,6 +544,7 @@ namespace NuGet.PackageManagement.VisualStudio
             SolutionClosed?.Invoke(this, EventArgs.Empty);
 
             _solutionOpenedRaised = false;
+            _solutionDir = null;
         }
 
         private void OnBeforeClosing()
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSSolutionManager.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSSolutionManager.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.Diagnostics;
-using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
@@ -67,6 +66,7 @@ namespace NuGet.PackageManagement.VisualStudio
 
         private bool _initialized;
         private bool _cacheInitialized;
+        private string _solutionDirectory;
 
         //add solutionOpenedRasied to make sure ProjectRename and ProjectAdded event happen after solutionOpened event
         private bool _solutionOpenedRaised;
@@ -434,34 +434,33 @@ namespace NuGet.PackageManagement.VisualStudio
             });
         }
 
-        private string _solutionDir;
         public string SolutionDirectory
         {
             get
             {
-                if (!string.IsNullOrEmpty(_solutionDir))
+                if (!string.IsNullOrEmpty(_solutionDirectory))
                 {
-                    return _solutionDir;
+                    return _solutionDirectory;
                 }
 
-                _solutionDir = NuGetUIThreadHelper.JoinableTaskFactory.Run(GetSolutionDirectoryAsync);
-                return _solutionDir;
+                _solutionDirectory = NuGetUIThreadHelper.JoinableTaskFactory.Run(GetSolutionDirectoryAsync);
+                return _solutionDirectory;
             }
         }
 
         public async Task<string> GetSolutionDirectoryAsync()
         {
-            if (!string.IsNullOrEmpty(_solutionDir))
+            if (!string.IsNullOrEmpty(_solutionDirectory))
             {
-                return _solutionDir;
+                return _solutionDirectory;
             }
 
             await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
             var vsSolution = await _asyncVSSolution.GetValueAsync();
             if (IsSolutionOpenFromVSSolution(vsSolution))
             {
-                _solutionDir = (string)GetVSSolutionProperty(vsSolution, (int)__VSPROPID.VSPROPID_SolutionDirectory);
-                return _solutionDir;
+                _solutionDirectory = (string)GetVSSolutionProperty(vsSolution, (int)__VSPROPID.VSPROPID_SolutionDirectory);
+                return _solutionDirectory;
             }
             return null;
         }
@@ -544,7 +543,7 @@ namespace NuGet.PackageManagement.VisualStudio
             SolutionClosed?.Invoke(this, EventArgs.Empty);
 
             _solutionOpenedRaised = false;
-            _solutionDir = null;
+            _solutionDirectory = null;
         }
 
         private void OnBeforeClosing()


src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSSolutionManager.cs
```diff
@@ -434,15 +434,34 @@ public void EnsureSolutionIsLoaded()
             });
         }
 
-        public string SolutionDirectory => NuGetUIThreadHelper.JoinableTaskFactory.Run(GetSolutionDirectoryAsync);
+        private string _solutionDir;
+        public string SolutionDirectory
```
nkolev92, line 437: The VSSolutionManager is shared across solutions.

nkolev92 (changes requested):
The correct approach is in https://github.com/NuGet/NuGet.Client/pull/4879. 

We need to coordinate solution open + solution directory.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSSolutionManager.cs
nkolev92: There's also some timing concerns https://github.com/NuGet/NuGet.Client/pull/4879 is trying to address that are not being considered here.

erdembayar: What do you mean by VSSolutionManager is shared across solutions?
Could we load multiple solutions same time?

nkolev92: VSSolutionManager is a shared component. Check the Importing declaration.

- open solution A
- Close solution
- Open Solution B


As long as you don't turn close VS and shut down the process, it's the same solution manager.

nkolev92: I would just try to figure out what @zivkan's PR was at as he was trying to solve the same problem.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSSolutionManager.cs
```diff
@@ -434,15 +434,34 @@ public void EnsureSolutionIsLoaded()
...
+        {
+            get
+            {
+                if (!string.IsNullOrEmpty(_solutionDir))
```
nkolev92: I see you're relying on a null, while Andy was doing the is initialized approach. 

Maybe this is better than that. 

The one scenario that Andy's implementation had issues with was the temporary project one. 

You can see which test was failing and see why.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSSolutionManager.cs
erdembayar: > VSSolutionManager is a shared component. Check the Importing declaration.
> 

What is the Importing declaration?

> * open solution A
> * Close solution
> * Open Solution B
> 
> As long as you don't turn close VS and shut down the process, it's the same solution manager.

But I do reset the` _solutionDirectory` to null every time we close the solution. Isn't that enough?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSSolutionManager.cs
erdembayar: My PR is passing all the tests. Is it still concern?

nkolev92: IMO yes, as the approaches are different. 

This particular change has undergone multiple iterations, and we've slightly been getting better. 

Maybe this is the right approach, but I'm not 100% confident I trust the tests in all scenarios.

In case of a null this retries, but Andy's didn't. The difference between 2 approaches is minor but understanding why exactly there's a difference would allow to be confident about the change.

erdembayar: I'll look into Andy's PR, I'm not familiar with all the changes in that PR. 
Could we ask vendor to test this for us with different scenarios?

nkolev92: Please take a look at that PRs. 

The differences between the 2 are easy to count. 

IMO, having high confidence in the changes requires us to understand why the 2 slightly different implementations are behaving differently. 

I don't believe the vendors are the going to help us testing here. 
The problems we've had are timing and more often theoretical than fully deterministic.

erdembayar: Do you know where is this [Get-Package](https://github.com/NuGet/NuGet.Client/blob/84d99ce064a5f70628c89af9f30693581e84dc37/test/EndToEnd/tests/GetPackageTest.ps1#L393) defined for End2nd test? 
Not sure what is this GetPackageDoesNotThrowIfSolutionIsTemporary for now.

erdembayar: Is it NuGet package manager console command?

nkolev92: If you search for GetPackageCommand.cs you get https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.PowerShellCmdlets/Cmdlets/GetPackageCommand.cs.

erdembayar: This is the exception call stack I get with https://github.com/NuGet/NuGet.Client/pull/4879
It's on UI thread and initialized still throw exception, tbh I'm not familiar with VS Com interface. Not what to do next here?

image
image

nkolev92: > It's on UI thread and initialized still throw exception, tbh I'm not familiar with VS Com interface.

It's just a native API that's being called, so the error communication is not the same. 

When `OnAfterOpenSolution` is called, is what's the value of fNewSolution https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.interop.ivssolutionevents.onafteropensolution?view=visualstudiosdk-2022#parameters? 

Are the params any different from usual? 

What Andy was trying to ensure is that we don't have any synchronization problems.
Basically there are solution open/close and that's the ideal time to update the solution directory. 

I don't know if that's a real concern as long as we fallback if the value is null in the getter for Solution Directory which was added later.

erdembayar: > I don't believe the vendors are the going to help us testing here. The problems we've had are timing and more often theoretical than fully deterministic.

https://github.com/NuGet/NuGet.Client/pull/4879 has more changes regarding if solution is open or not besides `solution dir`. Here in this PR we're only interested in `solution dir` value caching, change is much less.  I believe https://github.com/NuGet/Home/issues/11090 is much larger issue than this current issue I'm working on. Could we scope into current issue instead of solving both?

nkolev92: > https://github.com/NuGet/NuGet.Client/pull/4879 has more changes regarding if solution is open or not besides solution dir

I'm not sure I agree. The changes are very, very similar. 

- The other PR does the initial calculation in the InitializeAsync part, https://github.com/NuGet/NuGet.Client/pull/4879/files#diff-bed53a8b2e737e49ff6141c4279e525dfa2d100eabf6510ef6e0a1746de10505R208-R216. This PR is skipping that. Think of it that this PR is retrieving the directory in a lazy way, while the other change is just doing it at the first point where the solution directory is available. 
- IsSolutionOpen check is the extra code, but it's really the same solution. If you look at the implementation it's just merely checking that the solution directory is not null. 
- The other PRs solution directory check is again relying on a cookie being set, which means InitializeAsync is being run, while you check for null value. 
- OnAfterOpenSolution is calculating the solution directory at the first time it should be available. Think of it InitializeAsync, but for new solutions potentially. 

> I believe https://github.com/NuGet/Home/issues/11090 is much larger issue than this current issue I'm working on

The solution directory calculation is equivalent in both approaches. 

I think this is change is going to work, but understanding the difference between the 2 might allow us to potentially have an even better solution. 

GetSolutionDirectory(Async) could just literally return the value. There should never be a need to do a calculation, as long as we get the correct value at initialization time.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSSolutionManager.cs
nkolev92: > What is the Importing declaration?

This indicates it's shared in the process: https://github.com/NuGet/NuGet.Client/blob/c0371f7165bea201aded1a8013ca0b17e2ad0092/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSSolutionManager.cs#L39-L40

nkolev92: Resolving this as you reset the value.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSSolutionManager.cs
```diff
@@ -434,15 +434,33 @@ public void EnsureSolutionIsLoaded()
...
+        {
+            get
+            {
+                if (!string.IsNullOrEmpty(_solutionDirectory))
+                {
+                    return _solutionDirectory;
+                }
+
+                _solutionDirectory = NuGetUIThreadHelper.JoinableTaskFactory.Run(GetSolutionDirectoryAsync);
+                return _solutionDirectory;
```
zivkan, line 447: I think this has a bug if a solution with only legacy csproj is opened, something calls `ISolutionManager.SolutionDirectory`, without `InitializeAsync` ever being run, then solution is closed. In this case, `OnAfterClosing` will never be called, since `InitializeAsync` was never called, and then `VsSolutionManager` will still return a solution directory for a solution that's now closed.

Understanding in what scenarios` .SolutionDirectory` is called, without `InitializeAsync()` being called is difficult. Opening PM UI, PMC, or doing a restore (including a build) will cause `VsSolutionManager.InitializeAsync()` to be called, so we have to think of different scenarios.

One example might be right clicking the solution in solution explorer, as the code that checks if the "restore nuget packages" command should be enabled/disabled calls something on `ISolutionManager`, although it's more likely to call `.IsSolutionOpen`, rather than `.SolutionDirectory`.  The analysis of checking every single reference of `.SolutionDirectory` to ensure that all of them cause `InitializeAsync()` to be called was more effort than just implementing code that should "do the right thing", however, that one failing test prevented me from merging my PR.

erdembayar: > I think this has a bug if a solution with only legacy csproj is opened, something calls `ISolutionManager.SolutionDirectory`, without `InitializeAsync` ever being run, then solution is closed. In this case, `OnAfterClosing` will never be called, since `InitializeAsync` was never called, and then `VsSolutionManager` will still return a solution directory for a solution that's now closed.

I just tested with solution with only PC and PR legacy csproj, both cases `OnAfterClosing` was called and `solution directory` value reset to null. Am I missing something? 
1. I opened [LegacyProjectPC](https://github.com/NuGet/NuGet.Client/files/10119776/LegacyProjectPC.zip) solution in VS then closed the solution in VS, observed if solution directory value reset in `OnAfterClosing`.
2. Next, I opened [LegacyProjectPR](https://github.com/NuGet/NuGet.Client/files/10119782/LegacyProjectPR.zip) solution in VS then closed the solution in VS, observed if solution directory value reset in `OnAfterClosing`.

jeffkl (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSSolutionManager.cs
erdembayar: @zivkan 
Could you be able to test possible breaking scenarios with vsix artifacts from [here](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7024573&view=artifacts&pathAsName=false&type=publishedArtifacts) when you have time, please?
Currently it's passing RPS tests.

nkolev92: I found a scenario that doesn't seem to work. 

1. Load a C++ project (make sure no PMC, or anything like that is loaded). 
2. Open the settings via `Options -> NuGet Pakcage Manager Settings`.
3. Close the settings dialog
4. Close the solution (not VS).
5. Load any other solution.

image


This is basically what @zivkan is referring to by `ISolutionManager.SolutionDirectory, without InitializeAsync ever being run, then solution is closed`. 

In cases that the solution manager does not get initialized, we don't get a call to reset the property.

nkolev92: btw you might want to rebase cause there's been so many project fixes, and it pretty much destroyed all the local caches :D

zivkan: I believe I fixed my other PR: https://github.com/NuGet/NuGet.Client/pull/4879/commits/fd029a7386b50328354b64c2e9d603712aa542ed The test passes locally, and I'm now waiting for CI results.

Therefore, I believe we can close this PR and accept mine: https://github.com/NuGet/NuGet.Client/pull/4879

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSSolutionManager.cs
zivkan: I believe I fixed my other PR: https://github.com/NuGet/NuGet.Client/pull/4879/commits/fd029a7386b50328354b64c2e9d603712aa542ed The test passes locally, and I'm now waiting for CI results.

Therefore, I believe we can close this PR and accept mine: https://github.com/NuGet/NuGet.Client/pull/4879

erdembayar:
@nkolev92 @NuGet/nuget-client 
Based on @zivkan 's [comment](https://github.com/NuGet/NuGet.Client/pull/4951#discussion_r1043001396) I'm closing this PR. End result would be same, bit different approach.

PR #4953, opened 2022-11-29
Effective PrivateAssets of centrally managed transitive dependencies should be an intersection of parent dependencies 

### LINKED ISSUE #12270: PrivateAssets flow incorrectly to transitively pinned centrally managed dependencies
### NuGet Product Used

dotnet.exe

### Product Version

7.0.100

### Worked before?

no

### Impact

### Repro Steps & Context

For project:
``` xml
  <PropertyGroup>
    <TargetFramework>netstandard2.0</TargetFramework>
    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
    <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Newtonsoft.Json.Bson" PrivateAssets="Compile;Runtime" />
    <PackageReference Include="Microsoft.Rest.ClientRuntime" PrivateAssets="Compile" />
  </ItemGroup>
</Project>
```
and `Directory.Packages.props`:
``` xml
<Project>
    <ItemGroup>
        <PackageVersion Include="Newtonsoft.Json" Version="13.0.1" />
        <PackageVersion Include="Newtonsoft.Json.Bson" Version="1.0.2" />
        <PackageVersion Include="Microsoft.Rest.ClientRuntime" Version="2.3.24" />
    </ItemGroup>
</Project>
```

The expected content of `centralTransitiveDependencyGroups` is:
``` json
"centralTransitiveDependencyGroups": {
    ".NETStandard,Version=v2.0": {
      "Newtonsoft.Json": {
        "include": "Runtime, Compile, Native, BuildTransitive",
        "version": "[13.0.1, )"
    }
  }
}
```
, but the SDK 7.0.100 produces empty `centralTransitiveDependencyGroups`.

PS:
I'm going to submit a PR with a fix.

### Verbose Logs

_No response_

marcin-krystianc:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12270

Regression? Last working version: no

## Description

The existing code was using bitwise `OR` for calculating the effective value of PrivateAssets for central managed transitive dependencies. But instead, it should perform bitwise `AND` to get the intersection of PrivateAssets values of relevant parent dependencies.

Also, during Pack, the asset of `BuildTransitive` was becoming `Build | BuildTransitive` so a new overload was added to get only the specified values.

The unit test was also updated to check more combinations of things.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/LockFileBuilder.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/LockFileBuilder.cs
@@ -474,7 +474,7 @@ namespace NuGet.Commands
 
         private void AddCentralTransitiveDependencyGroupsForPackageReference(PackageSpec project, LockFile lockFile, IEnumerable<RestoreTargetGraph> targetGraphs)
         {
-            if (project.RestoreMetadata == null || !project.RestoreMetadata.CentralPackageVersionsEnabled)
+            if (project.RestoreMetadata == null || !project.RestoreMetadata.CentralPackageVersionsEnabled || !project.RestoreMetadata.CentralPackageTransitivePinningEnabled)
             {
                 return;
             }
@@ -491,7 +491,7 @@ namespace NuGet.Commands
                 }
 
                 // The transitive dependencies enforced by the central package version management file are written to the assets to be used by the pack task.
-                List<LibraryDependency> centralEnforcedTransitiveDependencies = GetLibraryDependenciesForCentralTransitiveDependencies(targetGraph, targetFrameworkInformation, project.RestoreMetadata.CentralPackageTransitivePinningEnabled).ToList();
+                List<LibraryDependency> centralEnforcedTransitiveDependencies = GetLibraryDependenciesForCentralTransitiveDependencies(targetGraph, targetFrameworkInformation).ToList();
 
                 if (centralEnforcedTransitiveDependencies.Any())
                 {
@@ -513,7 +513,7 @@ namespace NuGet.Commands
         /// <param name="targetFrameworkInformation">The <see cref="TargetFrameworkInformation" /> for the target framework to get centrally defined transitive dependencies for.</param>
         /// <param name="centralPackageTransitivePinningEnabled">A value indicating whether or not central transitive dependency version pinning is enabled.</param>
         /// <returns>An <see cref="IEnumerable{LibraryDependency}" /> representing the centrally defined transitive dependencies for the specified <see cref="RestoreTargetGraph" />.</returns>
-        private IEnumerable<LibraryDependency> GetLibraryDependenciesForCentralTransitiveDependencies(RestoreTargetGraph targetGraph, TargetFrameworkInformation targetFrameworkInformation, bool centralPackageTransitivePinningEnabled)
+        private IEnumerable<LibraryDependency> GetLibraryDependenciesForCentralTransitiveDependencies(RestoreTargetGraph targetGraph, TargetFrameworkInformation targetFrameworkInformation)
         {
             foreach (GraphNode<RemoteResolveResult> rootNode in targetGraph.Graphs)
             {
@@ -528,39 +528,34 @@ namespace NuGet.Commands
                     CentralPackageVersion centralPackageVersion = targetFrameworkInformation.CentralPackageVersions[node.Item.Key.Name];
                     Dictionary<string, LibraryIncludeFlags> dependenciesIncludeFlags = _includeFlagGraphs[targetGraph];
 
-                    LibraryIncludeFlags suppressParent = LibraryIncludeFlags.None;
+                    LibraryIncludeFlags suppressParent = LibraryIncludeFlags.All;
 
-                    if (centralPackageTransitivePinningEnabled)
+                    // Centrally pinned dependencies are not directly declared but the intersection of all of the PrivateAssets of the parents that pulled it in should apply to it
+                    foreach (GraphNode<RemoteResolveResult> parentNode in EnumerateParentNodes(node))
                     {
-                        // Centrally pinned dependencies are not directly declared but the PrivateAssets from the top-level dependency that pulled it in should apply to it also
-                        foreach (GraphNode<RemoteResolveResult> parentNode in EnumerateParentNodes(node))
-                        {
-                            LibraryDependency parentDependency = rootNode.Item.Data.Dependencies.FirstOrDefault(i => i.Name.Equals(parentNode.Item.Key.Name, StringComparison.OrdinalIgnoreCase));
+                        LibraryDependency parentDependency = rootNode.Item.Data.Dependencies.FirstOrDefault(i => i.Name.Equals(parentNode.Item.Key.Name, StringComparison.OrdinalIgnoreCase));
 
-                            // A transitive dependency that is a few levels deep won't be a top-level dependency so skip it
-                            if (parentDependency == null || parentDependency.ReferenceType != LibraryDependencyReferenceType.Direct)
-                            {
-                                continue;
-                            }
-
-                            suppressParent |= parentDependency.SuppressParent;
-                        }
-
-                        // If all assets are suppressed then the dependency should not be added
-                        if (suppressParent == LibraryIncludeFlags.All)
+                        // A transitive dependency that is a few levels deep won't be a top-level dependency so skip it
+                        if (parentDependency == null || parentDependency.ReferenceType != LibraryDependencyReferenceType.Direct)
                         {
                             continue;
                         }
+
+                        suppressParent &= parentDependency.SuppressParent;
                     }
 
-                    yield return new LibraryDependency()
+                    // If all assets are suppressed then the dependency should not be added
+                    if (suppressParent != LibraryIncludeFlags.All)
                     {
-                        LibraryRange = new LibraryRange(centralPackageVersion.Name, centralPackageVersion.VersionRange, LibraryDependencyTarget.Package),
-                        ReferenceType = LibraryDependencyReferenceType.Transitive,
-                        VersionCentrallyManaged = true,
-                        IncludeType = dependenciesIncludeFlags[centralPackageVersion.Name],
-                        SuppressParent = suppressParent,
-                    };
+                        yield return new LibraryDependency()
+                        {
+                            LibraryRange = new LibraryRange(centralPackageVersion.Name, centralPackageVersion.VersionRange, LibraryDependencyTarget.Package),
+                            ReferenceType = LibraryDependencyReferenceType.Transitive,
+                            VersionCentrallyManaged = true,
+                            IncludeType = dependenciesIncludeFlags[centralPackageVersion.Name],
+                            SuppressParent = suppressParent,
+                        };
+                    }
                 }
             }
         }
--- a/test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests.cs
@@ -2563,9 +2563,19 @@ namespace NuGet.Commands.Test
         /// <summary>
         /// Verifies that when a transitive package version is pinned and is referenced by multiple parents, the PrivateAssets flow from the package that pulled it into the graph to the pinned dependency.
         /// </summary>
-        /// <returns></returns>
-        [Fact]
-        public async Task RestoreCommand_CentralVersion_AssetsFile_PrivateAssetsFlowsToPinnedDependenciesWithMultipleParents()
+        [Theory]
+        [InlineData(
+            LibraryIncludeFlags.All, // PrivateAssets="All"
+            LibraryIncludeFlags.Build | LibraryIncludeFlags.ContentFiles | LibraryIncludeFlags.Analyzers, // Default PrivateAssets
+            LibraryIncludeFlags.Build | LibraryIncludeFlags.ContentFiles | LibraryIncludeFlags.Analyzers)] // Expect only the intersection, in this case the default
+        [InlineData(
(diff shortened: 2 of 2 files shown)


marcin-krystianc:
Next one is https://github.com/NuGet/NuGet.Client/pull/4954

src/NuGet.Core/NuGet.Commands/RestoreCommand/LockFileBuilder.cs
```diff
@@ -528,7 +528,7 @@ private IEnumerable<LibraryDependency> GetLibraryDependenciesForCentralTransitiv
                     CentralPackageVersion centralPackageVersion = targetFrameworkInformation.CentralPackageVersions[node.Item.Key.Name];
                     Dictionary<string, LibraryIncludeFlags> dependenciesIncludeFlags = _includeFlagGraphs[targetGraph];
 
-                    LibraryIncludeFlags suppressParent = LibraryIncludeFlags.None;
+                    LibraryIncludeFlags suppressParent = LibraryIncludeFlags.All;
```
erdembayar, line 531: Should this be conditional depending on if `centralPackageTransitivePinningEnabled` is true or false?
Now it's always `LibraryIncludeFlags.All`.

marcin-krystianc: I think that the check `if (centralPackageTransitivePinningEnabled)` is even not necessary. We know that it must be always true, because the current node is central transitive (`node.Item.IsCentralTransitiv == true` ) and  it is only possible when `centralPackageTransitivePinningEnabled` is true. So, in my opinion, there is no point in making the `LibraryIncludeFlags suppressParent = LibraryIncludeFlags.All;` to be conditional.

marcin-krystianc: I've pushed a change that checks whether the `centralPackageTransitivePinningEnabled` is true before calling the `GetLibraryDependenciesForCentralTransitiveDependencies`.

jeffkl (dismissed):

src/NuGet.Core/NuGet.LibraryModel/LibraryIncludeFlagUtils.cs
```diff
@@ -58,7 +66,7 @@ public static LibraryIncludeFlags GetFlags(IEnumerable<string> flags)
                         result |= LibraryIncludeFlags.Analyzers;
                         break;
                     case "buildtransitive":
-                        result |= LibraryIncludeFlags.BuildTransitive | LibraryIncludeFlags.Build;
+                        result |= includeBuildWithBuildTransitive ? LibraryIncludeFlags.BuildTransitive | LibraryIncludeFlags.Build : LibraryIncludeFlags.BuildTransitive;
```
jeffkl: This is needed because the assets file is written with `BuildTransitive` but when its read for Pack, the value ends up being `Build | BuildTransitive`

jeffkl:
@NuGet/nuget-client this is ready for review

src/NuGet.Core/NuGet.LibraryModel/LibraryIncludeFlagUtils.cs
marcin-krystianc: I think adding an integration test for it would be a good idea. I tried to test locally, but I struggled to find the scenario in which this change has an effect.

src/NuGet.Core/NuGet.LibraryModel/GlobalSuppressions.cs
```diff
@@ -25,3 +25,4 @@
 [assembly: SuppressMessage("Build", "CA2227:Change 'Items' to be read-only by removing the property setter.", Justification = "<Pending>", Scope = "member", Target = "~P:NuGet.LibraryModel.Library.Items")]
 [assembly: SuppressMessage("Build", "CA2227:Change 'NoWarn' to be read-only by removing the property setter.", Justification = "<Pending>", Scope = "member", Target = "~P:NuGet.LibraryModel.LibraryDependency.NoWarn")]
 [assembly: SuppressMessage("Build", "CA1067:Type NuGet.LibraryModel.FrameworkDependency should override Equals because it implements IEquatable<T>", Justification = "<Pending>", Scope = "type", Target = "~T:NuGet.LibraryModel.FrameworkDependency")]
+[assembly: SuppressMessage("Globalization", "CA1308:Normalize strings to uppercase", Justification = "<Pending>", Scope = "member", Target = "~M:NuGet.LibraryModel.LibraryIncludeFlagUtils.GetFlags(System.Collections.Generic.IEnumerable{System.String},System.Boolean)~NuGet.LibraryModel.LibraryIncludeFlags")]
```
zivkan: Putting on my API design review hat on, do we need this new API?

Instead, can the caller add `buildTransitive` to the `IEnumerable<string>` before calling the existing public API?

Not that I'm an expert on `NuGet.LibraryModel`, but my gut reaction is that it's a bit of a leaky abstraction to add the bool parameter, and therefore might not be a good API design.

jeffkl: The problem is that the assets file contains only `BuildTransitive` and when its read by pack, the value comes back as both.  In this code path, we are going to be using the value in the `.nuspec` so it should really only contain what the user specified.  Since its a transitive dependency, `Build` should not flow unless specified right?  

Given that, I couldn't come up with a better API change, I need this method to return the true values.  I am open to suggestions though so please let me know.

zivkan: > The problem is that the assets file contains only BuildTransitive and when its read by pack, the value comes back as both

That sounds like a bug to me.

When I do `<PackageReference Include="Package1" Version="1.0.0" IncludeAssets="compile;runtime;build;buildtransitive" />` and restore, in the assets file I see `"include": "Runtime, Compile, Build, BuildTransitive"`.  So, the assets file does seem capable of specifying `Build` and `BuildTransitive` independently.

When I try packing a project (that does not use CPM) with PackageReferences modifiers like `PrivateAssets="none"`, `PrivateAssets="analyzers"`, or the above IncludeAssets, the generated nuspec contains the include and exclude attributes that I'd expect, including being capable of specifying `Build` and `BuildTransitive` independently.

This has me wondering, is this PR trying to solve the problem for CPM in a different way that it's already implemented for non-CPM projects?

Perhaps I don't understand your comment very well. I certainly don't know the code being modified, so if there are indeed practical constraints that aren't obvious from a theoretical knowledge of the asset selection feature, then I'm unaware. But I don't understand why reading the assets file will parse `BuildTransitive` as both `Build` and `BuildTransitive`, when the assets file is capable of representing both. From a pack point of view, it also doesn't appear to be a limitation of non-CPM projects.

src/NuGet.Core/NuGet.LibraryModel/LibraryIncludeFlagUtils.cs
```diff
@@ -24,6 +24,14 @@ public static class LibraryIncludeFlagUtils
         /// Convert set of flag strings into a LibraryIncludeFlags.
```
jeffkl: Team Review: Update this XML doc comment to say it treats `BuildTranstive` as `BuildTranstive | Build`

src/NuGet.Core/NuGet.Commands/RestoreCommand/LockFileBuilder.cs
```diff
@@ -528,29 +528,26 @@ private IEnumerable<LibraryDependency> GetLibraryDependenciesForCentralTransitiv
...
+                        if (parentDependency == null || parentDependency.ReferenceType != LibraryDependencyReferenceType.Direct)
                         {
                             continue;
                         }
+
+                        suppressParent &= parentDependency.SuppressParent;
+                    }
+
+                    // If all assets are suppressed then the dependency should not be added
+                    if (suppressParent == LibraryIncludeFlags.All)
```
nkolev92: Given that we're doing the intersection, isn't an early exit is incorrect now?

src/NuGet.Core/NuGet.LibraryModel/GlobalSuppressions.cs
nkolev92: There's a related bug: https://github.com/NuGet/Home/issues/9672

Build transitive is implemented in a way that might not be the most intuitive. 
build and buidltransitive aren't exclusive like all the other asset types. 

if a package has build and buildtransitive files, only buildtransitive will be used. 

By extension, if you write `PrivateAssets="buildtransitive"`, pretty sure it tries to suppress both build and buildtransitive

src/NuGet.Core/NuGet.LibraryModel/LibraryIncludeFlagUtils.cs
```diff
@@ -24,6 +24,14 @@ public static class LibraryIncludeFlagUtils
...
         /// </summary>
         public static LibraryIncludeFlags GetFlags(IEnumerable<string> flags)
+        {
+            return GetFlags(flags, includeBuildWithBuildTransitive: true);
+        }
+
+        /// <summary>
+        /// Convert set of flag strings into a LibraryIncludeFlags.
+        /// </summary>
+        public static LibraryIncludeFlags GetFlags(IEnumerable<string> flags, bool includeBuildWithBuildTransitive)
```
nkolev92: There's still an API discussion, so I'm not expecting them added immediately, but given that this is a new public method, and it's super easy to test, we msut add unit tests for this.

src/NuGet.Core/NuGet.LibraryModel/GlobalSuppressions.cs
marcin-krystianc: Yeah, this is super confusing. Accordign to my tests this reference:
```xml
<PackageReference Include="PackageWithBuildTransitive" Version="1.0.0" IncludeAssets="Build" ExcludeAssets="BuildTransitive" />
```
doesn't include `Build` assets although it is explicitly requested.

marcin-krystianc (commented):
Given that there is no clear consensus about "Build" and "BuildTransitive" assets and the fact the problem is not strictly connected to the CPM - maybe it should be addressed in another PR?

src/NuGet.Core/NuGet.LibraryModel/GlobalSuppressions.cs
zivkan: Probably getting off-topic for this PR now, but does this also mean that the use case for https://github.com/NuGet/Home/issues/6938 isn't going to work?  Even with that feature implemented, the package trying to include its dependency's `build` assets isn't going to work as expected?

nkolev92: build/buildtransitive and contentfiles are the "weird' asset types. The rest work. 
So technically will work https://github.com/NuGet/Home/issues/6938, but we'd need to fix the other behaviors.

src/NuGet.Core/NuGet.Commands/RestoreCommand/LockFileBuilder.cs
marcin-krystianc: I think this the code is correct, maybe we could improve its readability by reversing the condition and write it instead like this?:
```
if (suppressParent != LibraryIncludeFlags.All)
{
	yield return new LibraryDependency()
	{
		LibraryRange = new LibraryRange(centralPackageVersion.Name, centralPackageVersion.VersionRange, LibraryDependencyTarget.Package),
		ReferenceType = LibraryDependencyReferenceType.Transitive,
		VersionCentrallyManaged = true,
		IncludeType = dependenciesIncludeFlags[centralPackageVersion.Name],
		SuppressParent = suppressParent,
	}; 
}
```

src/NuGet.Core/NuGet.LibraryModel/GlobalSuppressions.cs
jeffkl: So what's the consensus here?  With my code change, I feel that i get the expected result.  Without changing it, it stays as confusing as ever.  It only applies when someone tries to exclude `BuildTransitive`  independently which I would imagine isn't too common.

src/NuGet.Core/NuGet.Commands/RestoreCommand/LockFileBuilder.cs
jeffkl: If the intersection yields `LibraryIncludeFlags.All` then exiting in this case means that the `LibraryDependency` will be left out of the `centralTransitiveDependencyGroups` section of the `project.assets.json` file and then during `Pack` it will be left out of the `.nuspec`.  This would only happen if all parents of a package set `PrivateAssets="All"` which means they aren't in the `.nuspec` and neither should this transitively pinned package.

nkolev92: My bad, I thought this was in the foreach loop. I misread it. 

I agree with @marcin-krystianc that flipping the condition might be more readable than using a continue.

src/NuGet.Core/NuGet.LibraryModel/GlobalSuppressions.cs
nkolev92: I think it'd be great if the build/buildtransitive chnage is fixed in all scenarios. 
Having it change only for transitive pinning is a probably not enough?

jeffkl (approved):

nkolev92 (approved):

zivkan (approved):

PR #4954, opened 2022-11-29
Improved performance of calculation of PrivateAssets for transitively pinned centrally managed dependencies

### LINKED ISSUE #12269: Performance regression of NuGet restores in the sdk v7.0.100 due to calculation of "CentralTransitiveDependencyGroups"
### NuGet Product Used

dotnet.exe

### Product Version

7.0.100

### Worked before?

6.0.400

### Impact

### Repro Steps & Context

NuGet restore got considerably slower due to changes in how the `CentralTransitiveDependencyGroups` is calculated (https://github.com/NuGet/NuGet.Client/pull/4669.
For some of our internal solutions (which, unfortunately, we cannot share publicly ) the restore time was increased to **tens of minutes**.  It makes it impossible for us to migrate to dotnet 7. We can't share our internal solution but, we have a public repro that demonstrates the problem on a smaller scale where the restore time is doubled.

1: checkout https://github.com/marcin-krystianc/TestSolutions/tree/master/LargeAppWithPrivatePackagesCentralisedNGBVRemoved
2. create `global.json` for older versions e.g. `6.0.400`
3. Go to `LargeAppWithPrivatePackagesCentralisedNGBVRemoved\solution` directory and run `dotnet restore --force -clp:summary /p:RestoreUseStaticGraphEvaluation=true` to measure performance before the problem existed
4. set version 7.0.100 in `global.json`
5. run `dotnet restore --force -clp:summary /p:RestoreUseStaticGraphEvaluation=true` again to see performance regression (45s vs 90s on my machine)

PS:
I'm going to submit a PR with a fix for this problem.

### Verbose Logs

_No response_

marcin-krystianc:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12269

Regression? Last working version: 6.0.400

## Description

Existing code enumerated parent nodes in an inefficient fashion, it could yield the same node many times. Another performance issue was a linear scan to get the dependency node for the given parent node.
It is being fixed by:
- using a HashSet to avoid traversing the same nodes multiple times 
- using a dictionary lookup instead of a linear scan.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


jeffkl:
@marcin-krystianc what's the status of this?  I'd like to get this in the next release if there's a large performance increase.

marcin-krystianc:
> @marcin-krystianc what's the status of this? I'd like to get this in the next release if there's a large performance increase.

I needed previous PRs to be merged. Now I can resolve remaining conflicts and fine tune it. It should be ready to review soon.

marcin-krystianc:
@jeffkl It is ready for a review. I've tested the restore time using the test scenario from the https://github.com/NuGet/Home/issues/12269 and now the restore time is more or less the same as it was for 6.0.400.

src/NuGet.Core/NuGet.Commands/RestoreCommand/LockFileBuilder.cs
```diff
@@ -517,6 +517,8 @@ private IEnumerable<LibraryDependency> GetLibraryDependenciesForCentralTransitiv
         {
             foreach (GraphNode<RemoteResolveResult> rootNode in targetGraph.Graphs)
             {
+                var dependencyDictionary = rootNode.Item.Data.Dependencies.ToDictionary(x => x.Name, x => x, StringComparer.OrdinalIgnoreCase);
```
jeffkl: Are there cases where this dictionary won't end up being used?  On line 525 nodes can be skipped, so perhaps don't initialize this dictionary until its needed on line 538?

src/NuGet.Core/NuGet.Commands/RestoreCommand/LockFileBuilder.cs
```diff
@@ -531,17 +533,10 @@ private IEnumerable<LibraryDependency> GetLibraryDependenciesForCentralTransitiv
...
-                        LibraryDependency parentDependency = rootNode.Item.Data.Dependencies.FirstOrDefault(i => i.Name.Equals(parentNode.Item.Key.Name, StringComparison.OrdinalIgnoreCase));
-
-                        // A transitive dependency that is a few levels deep won't be a top-level dependency so skip it
-                        if (parentDependency == null || parentDependency.ReferenceType != LibraryDependencyReferenceType.Direct)
-                        {
-                            continue;
-                        }
-
-                        suppressParent &= parentDependency.SuppressParent;
+                        var dependency = dependencyDictionary[dependencyNode.Key.Name];
```
jeffkl: Use explicit type instead of `var` please.

src/NuGet.Core/NuGet.Commands/RestoreCommand/LockFileBuilder.cs
```diff
@@ -561,35 +556,44 @@ private IEnumerable<LibraryDependency> GetLibraryDependenciesForCentralTransitiv
...
                 }
-                else if (item.OuterNode != null)
+                else
                 {
-                    // Normal transitive nodes use OuterNode to track their parent
-                    foreach (GraphNode<T> outerNode in EnumerateParentNodes(item.OuterNode))
+                    if (node.OuterNode == rootNode)
                     {
-                        yield return outerNode;
+                        // It's what we are looking for
```
jeffkl: Please clarify this comment that the node is a parent

src/NuGet.Core/NuGet.Commands/RestoreCommand/LockFileBuilder.cs
marcin-krystianc: Yeah, it is possible that the dictionary will not be eventually used but it is not very likely scenario. I don't have strong opinion here..

jeffkl (dismissed):
Thanks for the contribution.  I ran the changes under the debugger and everything seems to still work.  Would it be possible to get some benchmarks just so we know _how_ much faster this is?

src/NuGet.Core/NuGet.Commands/RestoreCommand/LockFileBuilder.cs
```diff
@@ -561,35 +561,45 @@ private IEnumerable<LibraryDependency> GetLibraryDependenciesForCentralTransitiv
...
         /// <typeparam name="T">The type of the node.</typeparam>
+        /// <param name="rootNode">The <see cref="GraphNode{TItem}" /> to know which nodes are first level inner nodes.</param>
         /// <param name="graphNode">The <see cref="GraphNode{TItem}" /> to enumerate the parent nodes of.</param>
-        /// <returns>An <see cref="IEnumerable{T}" /> containing a top down list of parent nodes of the specied node.</returns>
-        private static IEnumerable<GraphNode<T>> EnumerateParentNodes<T>(GraphNode<T> graphNode)
+        /// <returns>An <see cref="IEnumerable{T}" /> containing list of parent nodes of the specified node.</returns>
+        private static IEnumerable<GraphNode<T>> EnumerateNodesForDependencyChecks<T>( GraphNode<T> rootNode, GraphNode<T> graphNode)
         {
-            foreach (GraphNode<T> item in graphNode.ParentNodes)
+            var visitedNodes = new HashSet<GraphNode<T>>();
```
nkolev92: We're running this for each package, in sequence, correct?

Can we reuse the sets/queues to save on allocations?

src/NuGet.Core/NuGet.Commands/RestoreCommand/LockFileBuilder.cs
```diff
@@ -561,35 +561,45 @@ private IEnumerable<LibraryDependency> GetLibraryDependenciesForCentralTransitiv
...
+                else
                 {
-                    // Normal transitive nodes use OuterNode to track their parent
-                    foreach (GraphNode<T> outerNode in EnumerateParentNodes(item.OuterNode))
+                    if (node.OuterNode == rootNode)
                     {
-                        yield return outerNode;
+                        // We were traversing the graph upwards and now found a parent node (an inner node of the root node).
+                        // Now we return it, so it can be used to calculate the effective value of SuppressParent for the initial node.
+                        yield return node;
```
nkolev92, line 601: I think it'd be further optimized if we used a List instead of yield return and an IEnumerable.

nkolev92 (commented):
Thanks for this fix. 

A few suggestions to reduce allocations.

src/NuGet.Core/NuGet.Commands/RestoreCommand/LockFileBuilder.cs
nkolev92: Similarly to the queue, we can probably reuse the list too.

marcin-krystianc:
> Thanks for the contribution. I ran the changes under the debugger and everything seems to still work. Would it be possible to get some benchmarks just so we know _how_ much faster this is?

Sure, just got some measurements for https://github.com/marcin-krystianc/TestSolutions/tree/master/LargeAppWithPrivatePackagesCentralisedNGBVRemoved on my dev laptop (multiple runs, but posting the best result only):
- 6.0.400, Time Elapsed 00:00:26.15 
- 7.0.100, Time Elapsed 00:00:57.18
- 7.0.100 (with this fix), Time Elapsed 00:00:28.28

 I see that there is still some noticeable difference when compared to 6.0.400, but I think it is justified by the extra work needed to correctly calculate effective PrivateAssets value for central transitive dependencies.

jeffkl (approved):

src/NuGet.Core/NuGet.Commands/RestoreCommand/LockFileBuilder.cs
marcin-krystianc: Yes, indeed. Reusing the `Queue` and the `HashSet` saved noticeable allocations (in my test scenario it was ~200MB.
I tried the same trick for returning results, but this time I didn't notice any significant savings. Thus I think it is better to keep IEnumerable pattern because it is more "elegant" in my opinion.

nkolev92: Thanks for profiling that!

jeffkl:
@marcin-krystianc something broke with our PR build pipeline but is now fixed, can you please rebase your change onto latest `dev` branch and force push?

src/NuGet.Core/NuGet.Commands/RestoreCommand/LockFileBuilder.cs
```diff
@@ -525,23 +530,21 @@ private IEnumerable<LibraryDependency> GetLibraryDependenciesForCentralTransitiv
                         continue;
                     }
 
+                    if (dependencyDictionary == null)
+                    {
+                        dependencyDictionary = rootNode.Item.Data.Dependencies.ToDictionary(x => x.Name, x => x, StringComparer.OrdinalIgnoreCase);
```
nkolev92, line 535: The fact that GraphNode doesn't implement hashcode made me a bit nervous, but I think it's ok in this case, as the data itself will never be duplicated.

nkolev92 (approved):

src/NuGet.Core/NuGet.Commands/RestoreCommand/LockFileBuilder.cs
marcin-krystianc: Hmm, not sure what was your concern really. Was it about `HashSet<GraphNode<T>>` for visited nodes? I think it is ok that it doesn't have proper equality comparer. We care about not visiting the same graph node twice, but it is ok when there are multiple nodes representing same project or package.

marcin-krystianc:
> @marcin-krystianc something broke with our PR build pipeline but is now fixed, can you please rebase your change onto latest `dev` branch and force push?

Done

davkean:
@marcin-krystianc Nice work on this. Looking our allocation telemetry, I can see this addressed the majority of the issues on this path.

PR #4956, opened 2022-11-30
Move all projects to C# 11

nkolev92:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2039

## Description

Note that the reason why we're using a specific version is that latest can someone cause build failures due to treat warnings as errors. 

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/build/common.project.props
+++ b/build/common.project.props
@@ -87,7 +87,7 @@
     <Features>strict</Features>
     <!-- Same as SDK default, but without CandidateAssemblyFiles in front, which would search in Content items -->
     <AssemblySearchPaths>{HintPathFromItem};{TargetFrameworkDirectory};{RawFileName}</AssemblySearchPaths>
-    <LangVersion>9</LangVersion>
+    <LangVersion>11</LangVersion>
     <LangVersion Condition="'$(DotNetBuildFromSource)' == 'true'">latest</LangVersion>
   </PropertyGroup>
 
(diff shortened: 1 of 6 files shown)


build/common.project.props
```diff
@@ -87,7 +87,7 @@
     <Features>strict</Features>
     <!-- Same as SDK default, but without CandidateAssemblyFiles in front, which would search in Content items -->
     <AssemblySearchPaths>{HintPathFromItem};{TargetFrameworkDirectory};{RawFileName}</AssemblySearchPaths>
-    <LangVersion>9</LangVersion>
+    <LangVersion>11</LangVersion>
```
dominoFire, line 90: Why not int Directory.Build.props? I think using retail MSBuild would benetit simpliflyfing code.

erdembayar:
There is some error regarding option 11 in [CI pipeline](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7026570&view=logs&j=eed5f894-0bfe-523e-c106-bdc7cd9a7fb6&t=7a483108-1d15-5878-75dd-386380f1e5fb&l=581).

nkolev92:
Apparently we're building linux with 6.0.400 somehow. Not ideal :) 

We use CPM right now which is dependent on using 7.0.100 and later tooling.

build/common.project.props
nkolev92: This file is imported by DBP already, so there's really no different. Keeping it in this file where all the current properties are is preferrable right now. 

At a later point someone can move these props if they wish, but that can be its own change.

nkolev92:
@erdembayar I bumped the SDK we use on linux to 7.x. 

Hopefully it passes now.

jeffkl (dismissed):

erdembayar (approved):

PR #4960, opened 2022-12-01
Remove custom property usage for ProjectReference from all NuGet projects, simplify dependencies where appropriate

nkolev92:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2043

## Description

As part of the effort make all NuGet projects work with the retail tooling out of the box, we should remove the properties in ProjectReference.

cc @jeffkl who suggested we should do this.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/build/common.project.props
+++ b/build/common.project.props
@@ -35,7 +35,6 @@
     <DotnetExePath Condition=" '$(IsXPlat)' == 'true' ">$(RepositoryRootDirectory)cli\dotnet</DotnetExePath>
     <SharedDirectory>$(BuildCommonDirectory)Shared</SharedDirectory>
     <NuGetCoreSrcDirectory>$(RepositoryRootDirectory)src\NuGet.Core\</NuGetCoreSrcDirectory>
-    <NuGetClientsSrcDirectory>$(RepositoryRootDirectory)src\NuGet.Clients\</NuGetClientsSrcDirectory>
     <NupkgOutputDirectory>$(ArtifactsDirectory)nupkgs\</NupkgOutputDirectory>
     <SolutionPackagesFolder>$(RepositoryRootDirectory)packages\</SolutionPackagesFolder>
     <ILMergeExePath>$(SolutionPackagesFolder)ilmerge\3.0.21\tools\net452\ILMerge.exe</ILMergeExePath>
--- a/build/common.project.props
+++ b/build/common.project.props
@@ -34,7 +34,6 @@
     <DotnetExePath>$(RepositoryRootDirectory)cli\dotnet.exe</DotnetExePath>
     <DotnetExePath Condition=" '$(IsXPlat)' == 'true' ">$(RepositoryRootDirectory)cli\dotnet</DotnetExePath>
     <SharedDirectory>$(BuildCommonDirectory)Shared</SharedDirectory>
-    <NuGetCoreSrcDirectory>$(RepositoryRootDirectory)src\NuGet.Core\</NuGetCoreSrcDirectory>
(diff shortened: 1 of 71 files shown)


jeffkl (commented):
🎉

nkolev92:
I had a messed up path. Should be fixed now :) That's why we have a CI :D

heng-liu:
Hi @jeffkl , may I know if there is any guidelines/best practice about writing msbuild related files? I think I was told to do in the opposite way before :)

jeffkl:
> Hi @jeffkl , may I know if there is any guidelines/best practice about writing msbuild related files? I think I was told to do in the opposite way before :)

Do you mean for `<ProjectReference />` items specifically or more top-level guidance?

heng-liu:
> Do you mean for `<ProjectReference />` items specifically or more top-level guidance?

For this PR, it's only `<ProjectReference />` items. 
But if there are more general guidance, that would be great! E.g. I'm not sure if I remember correctly that, you mentioned using Directory.Build.props and Directory.Build.targets instead of customized props and targets.
I tried to check [MSBuild best practices](https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-best-practices?view=vs-2022), but it's not very helpful.

jeffkl:
> For this PR, it's only <ProjectReference /> items.

Okay so for large internal enterprise builds, people used to recommend that you had a property that specified the folder where projects live and each ProjectReference would then use that as a base value for building the path.  The only scenario I'm aware of that this helps is when you move projects around, you could theoretically just update the property to the new path.  But the major downside is that every single time you add a new ProjectReference via Visual Studio or other tooling, it won't know about your custom property.  This means you really need enforcement to ensure that no one adds a new relative path that isn't using the property.  If you don't use these properties and just use the tooling (like Visual Studio), then you just make all of the paths relative either by hand or let something like Visual Studio do the work for you.  If we move projects around, yes updating these relative paths is a bit of work but honestly we just don't move projects around that much, if at all.

My general guidance to anyone using MSBuild is to avoid customizations whereever possible.  If we all went to college and learned the .NET developer stack using just Visual Studio or Visual Studio Code, we'd be creating projects with templates and adding references with dialogs.  This tooling will do its thing in a uniform way and if you customize a codebase too much you end up fighting the tooling.  So the best practice is to do things in a stock way so no one has to understand customizations, things are just as they expect.  

I don't know of a definitive all encompassing MSBuild best practices doc, but the standard documentation does cover a fair amount of what I'd consider to be best practices.

heng-liu:
> I don't know of a definitive all encompassing MSBuild best practices doc, but the standard documentation does cover a fair amount of what I'd consider to be best practices.

Thanks a lot @jeffkl ! That's very helpful!

erdembayar:
> As part of the effort make all NuGet projects work with the retail tooling out of the box

Which retail tool are you referring here? Trying to understand actual intention behind the change.

jeffkl:
> > As part of the effort make all NuGet projects work with the retail tooling out of the box
> 
> Which retail tool are you referring here? Trying to understand actual intention behind the change.

I'm talking about all of the tooling in the .NET developer ecosystem.  If you right click in Visual Studio and add a project reference or use `dotnet add reference` it will not respect these properties.

erdembayar (approved):

PR #4970, opened 2022-12-05
Have SDK resolver always log an error when SDK resolution in unsuccessful

### LINKED ISSUE #12312: During MSBuild project SDK resolution, always log errors in the case of failure
### NuGet Product(s) Affected

MSBuild.exe

### Current Behavior

Currently, the NuGet-based MSBuild project SDK resolver can only indicate if it successfully resolved and SDK or it did not.  There are many reasons it cannot resolve an SDK that aren't necessarily an error.  For example, you may not have specified a version or you may have misspelled the name of `global.json`.  In this case, the SDK resolver currently logs warnings.  As part of an effort in MSBuild to improve the SDK resolver error experience, the MSBuild team would like to only log errors in the case when an SDK resolver returns non-success.

### Desired Behavior

Any time the SDK resolver returns non-success, it should log errors which is what the MSBuild team is pushing for all SDK resolvers to do.

### Additional Context

_No response_

Forgind:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12312

No

## Description

SDK resolvers can only report success or failure.  In the case when the SDK resolver is disabled or couldn't find a version specified, the SDK resolver will report warnings.  As part of an effort to improve the SDK resolver failure scenario, the MSBuild team would like all SDK resolvers to log errors in the non-success case.

This pull request changes the warnings to errors in the case when the SDK resolver is disabled or a version was not specified.
## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/NuGetSdkResolver.cs
+++ b/src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/NuGetSdkResolver.cs
@@ -73,14 +73,14 @@ namespace Microsoft.Build.NuGetSdkResolver
             // Escape hatch to disable this resolver
             if (DisableNuGetSdkResolver.Value)
             {
-                return factory.IndicateFailure(errors: null, warnings: new List<string>() { Strings.Error_DisabledSdkResolver });
+                return factory.IndicateFailure(errors: new List<string>() { Strings.Error_DisabledSdkResolver }, warnings: null);
             }
 
             // This resolver only works if the user specifies a version in a project or a global.json.
             // Ignore invalid versions, there may be another resolver that can handle the version specified
             if (!TryGetNuGetVersionForSdk(sdkReference.Name, sdkReference.Version, resolverContext, out var parsedSdkVersion))
             {
-                return factory.IndicateFailure(errors: null, warnings: new List<string>() { Strings.Error_NoSdkVersion });
+                return factory.IndicateFailure(errors: new List<string>() { Strings.Error_NoSdkVersion }, warnings: null);
             }
 
             NuGet.Common.Migrations.MigrationRunner.Run();
--- a/src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/NuGetSdkResolver.cs
+++ b/src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/NuGetSdkResolver.cs
@@ -73,6 +73,10 @@ namespace Microsoft.Build.NuGetSdkResolver
             // Escape hatch to disable this resolver
             if (DisableNuGetSdkResolver.Value)
             {
+                // Errors returned to MSBuild aren't logged if another SDK resolver succeeds.  Returning errors on non-succcess is
+                // what all SDK resolvers should be doing and if no SDK resolver succeeds then MSBuild logs all of the errors as
+                // one.  In this case, the SDK resolver is disabled and it might be helpful for a user to see that they've disabled
+                // it in case it was a mistake.
                 return factory.IndicateFailure(errors: new List<string>() { Strings.Error_DisabledSdkResolver }, warnings: null);
             }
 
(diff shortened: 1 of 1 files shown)


src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/NuGetSdkResolver.cs
```diff
@@ -73,14 +73,14 @@ public override SdkResult Resolve(SdkReference sdkReference, SdkResolverContext
             // Escape hatch to disable this resolver
             if (DisableNuGetSdkResolver.Value)
             {
-                return factory.IndicateFailure(errors: null, warnings: new List<string>() { Strings.Error_DisabledSdkResolver });
```
erdembayar, line 76: @Forgind 
Are there any tests need to be updated as well?

Forgind: I wouldn't suspect so unless you've added them since my previous PR, but if CI fails, I can look into any tests that need updates.

rainersigwald (dismissed):

jeffkl (dismissed):
I've created and linked an issue.

@NuGet/nuget-client I talked with Nathan offline and I think we might as well make this change to keep SDK resolver consistent.  It can slightly improve the SDK resolution failure case in the scenario where the NuGet-based SDK resolver is the problem because all of the messages you see are errors rather than our message being a warning and other SDK resolver messages being an error.  The functionality is a bit wonky to log an error in this case but it is what we believe is the best option at the moment.

src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/NuGetSdkResolver.cs
```diff
@@ -73,14 +73,14 @@ public override SdkResult Resolve(SdkReference sdkReference, SdkResolverContext
...
+                return factory.IndicateFailure(errors: new List<string>() { Strings.Error_DisabledSdkResolver }, warnings: null);
```
zivkan, line 80: `DisableNuGetSdkResolver` is true when the customer sets an environment variable to opt-in to this behaviour (to opt-out of the feature). By making this scenario an error, it completely breaks the opt-out feature.  Is this intentional?  If so, would it be better to just delete `DisableNuGetSdkResolver`?

jeffkl: In this case it won't actually "break" anything and no errors are physically logged by the SDK resolver.  Only a list of errors will be returned to MSBuild.  The logic in MSBuild will then move on to the next SDK resolver and if none of them work then _all_ of the errors from _all_ of the SDK resolvers are logged, as well as all of the warnings.  

If you opt out of using the NuGet-based MSBuild project SDK resolver and another SDK resolver is able to resolve the SDK, then these errors returned to MSBuild are never logged.  Clear as mud right?  So in the case where the NuGet SDK resolver is "failing" because you disabled it, you'll get an error if no SDK resolver resolves it so you at least know that you might have forgotten that you disabled it.  But you'll also see errors from the other SDK resolvers so maybe you just misspelled the SDK name or something.

https://github.com/dotnet/msbuild/blob/main/src/Build/BackEnd/Components/SdkResolution/SdkResolverService.cs#L227

zivkan: In this case, I think a comment is warranted, because this is totally non-obvious and completely contradictory to how errors vs warnings work in restore, pack, basically all other parts of nuget.

jeffkl: Great idea, I'll add a comment here.

jeffkl (approved):

src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/NuGetSdkResolver.cs
Forgind: Thanks! I agree that how MSBuild is handling these is a bit confusing.

zivkan (approved):

erdembayar (approved):

PR #4976, opened 2022-12-07
Consider PrivateAsset flag independently from Include/Exclude asset flags for pack and ProjectReference  consumption

### LINKED ISSUE #6938: PrivateAssets and ExcludeAssets should be independent
## Details about Problem

NuGet product used (NuGet.exe | VS UI | Package Manager Console | dotnet.exe): I've tried dotnet.exe and nuget.exe  and msbuild.exe as it ships with my VS.

NuGet version (x.x.x.xxx): 4.6.1

dotnet.exe --version (if appropriate): 2.1.200

VS version (if appropriate): 27713.00.d15.8

OS version (i.e. win10 v1607 (14393.321)): Windows 10 RS4

Worked before? If so, with which NuGet version: No

## Detailed repro steps so we can see the same problem

In a .NET SDK style project, reference a package while specifying `PrivateAssets` and `ExcludeAssets` attributes, like this:

```xml
<PackageReference Include="CodeGeneration.Roslyn.BuildTime" PrivateAssets="none" ExcludeAssets="Build" />
```

### Expected result

I expect that *my* project will *not* consume the build elements of the package, but that my own package's respective consumers *will* consume the build elements of the referenced package.

In other words, I don't want my project importing the build folder .targets file from that package, but I *do* want this expressed in my package's .nuspec file:

```xml
<dependency id="CodeGeneration.Roslyn.BuildTime" version="0.4.42" include="all" />
```

### Actual result

```xml
<dependency id="CodeGeneration.Roslyn.BuildTime" version="0.4.42" include="Runtime,Compile,Native,ContentFiles,Analyzers" />
```

### Additional discussion

The fact that I can specify `PrivateAssets` (which is documented to influence how the dependency is expressed to others), _and_ I can specify `ExcludeAssets`/`IncludeAssets` (which influences how *I* get my own dependency), is fantastic, except that when I use the two together, they interact such that I might as well not set `PrivateAssets` because it seems to be ignored.

In another related case to this, I actually wanted these two attributes on my `PackageReference`: `PrivateAssets="none" IncludeAssets="none"` because I wanted to express a dependency downstream that I myself didn't share. But that simply suppresses the dependency in the .nuspec file. 😦 

### Sample Project

[refissue.zip](https://github.com/NuGet/Home/files/2009553/refissue.zip)

erdembayar:
## Bug

Fixes: https://github.com/NuGet/Home/issues/6938

## Description
Currently Include/Exclude flag completely eclipses PrivateAsset flag when generating nupkg package in many cases. So, it doesn't let package dependencies flow to parent projects of consuming project. This PR implements [spec](https://github.com/NuGet/Home/blob/e9a8465cbf5c51d915d16dd3e8e106f3621774bf/proposed/2022/PrivateAssetIndepdentFromExcludeAsset.md).

As per the restore considerations, https://github.com/NuGet/NuGet.Client/blob/dev/docs/feature-guide.md#restore-considerations, we need the project-system change as well https://github.com/dotnet/project-system/issues/8723

**Case 1.**
```
  <ItemGroup>
    <PackageReference Include="Microsoft.Windows.CsWin32" Version="0.2.138-beta" PrivateAssets="none" ExcludeAssets="build" />
  </ItemGroup>
```

**Before change nuspec file**
```
    <dependencies>
      <group targetFramework=".NETStandard2.0">
        <dependency id="Microsoft.Windows.CsWin32" version="0.2.138-beta" exclude="Runtime,Build,Native,Analyzers,BuildTransitive" />
      </group>
    </dependencies>
```

**After change:**
```
    <dependencies>
      <group targetFramework=".NETStandard2.0">
        <dependency id="Microsoft.Windows.CsWin32" version="0.2.138-beta" include="All" />
      </group>
    </dependencies>
```

**Case 2.**
```
  <ItemGroup>
    <PackageReference Include="Microsoft.Windows.CsWin32" Version="0.2.138-beta" PrivateAssets="none" IncludeAssets="none" />    
  </ItemGroup>
```

**Before change nuspec file**

```
    <dependencies>
      <group targetFramework=".NETStandard2.0" />
    </dependencies>
```

**After change:**
```
    <dependencies>
      <group targetFramework=".NETStandard2.0">
        <dependency id="Microsoft.Windows.CsWin32" version="0.2.138-beta" include="All" />
      </group>
    </dependencies>
```

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.SolutionRestoreManager/VSNominationUtilities.cs
+++ b/src/NuGet.Clients/NuGet.SolutionRestoreManager/VSNominationUtilities.cs
@@ -90,7 +90,7 @@ namespace NuGet.SolutionRestoreManager
         }
 
         internal static TargetFrameworkInformation ToTargetFrameworkInformation(
-            IVsTargetFrameworkInfo targetFrameworkInfo, bool cpvmEnabled, string projectFullPath)
+            IVsTargetFrameworkInfo targetFrameworkInfo, bool cpvmEnabled, bool privateAssetIndependentEnabled, string projectFullPath)
         {
             var tfi = new TargetFrameworkInformation
             {
@@ -117,7 +117,7 @@ namespace NuGet.SolutionRestoreManager
                 tfi.Dependencies.AddRange(
                     targetFrameworkInfo.PackageReferences
                         .Cast<IVsReferenceItem>()
-                        .Select(pr => ToPackageLibraryDependency(pr, cpvmEnabled)));
+                        .Select(pr => ToPackageLibraryDependency(pr, cpvmEnabled, privateAssetIndependentEnabled)));
             }
 
             if (targetFrameworkInfo is IVsTargetFrameworkInfo2 targetFrameworkInfo2)
@@ -300,6 +300,15 @@ namespace NuGet.SolutionRestoreManager
             return GetSingleNonEvaluatedPropertyOrNull(tfms, ProjectBuildProperties.CentralPackageTransitivePinningEnabled, MSBuildStringUtility.IsTrue);
         }
 
+        /// <summary>
+        /// Evaluates the msbuild properties and returns the value of the PrivateAssetIndependent property.
+        /// If it is not defined the default value will be disabled.
+        /// </summary>
+        internal static bool IsPrivateAssetIndependentEnabled(IEnumerable tfms)
+        {
+            return GetSingleNonEvaluatedPropertyOrNull(tfms, ProjectBuildProperties.ManagePackageVersionsCentrally, MSBuildStringUtility.IsTrue);
+        }
+
         private static NuGetFramework GetToolFramework(IEnumerable targetFrameworks)
         {
             return GetSingleNonEvaluatedPropertyOrNull(
@@ -371,7 +380,7 @@ namespace NuGet.SolutionRestoreManager
 
         #region IVSReferenceItemAPIs
 
-        private static LibraryDependency ToPackageLibraryDependency(IVsReferenceItem item, bool cpvmEnabled)
+        private static LibraryDependency ToPackageLibraryDependency(IVsReferenceItem item, bool cpvmEnabled, bool privateAssetIndependentEnabled)
         {
             if (!TryGetVersionRange(item, "Version", out VersionRange versionRange))
             {
@@ -391,7 +400,8 @@ namespace NuGet.SolutionRestoreManager
                 AutoReferenced = GetPropertyBoolOrFalse(item, "IsImplicitlyDefined"),
                 GeneratePathProperty = GetPropertyBoolOrFalse(item, "GeneratePathProperty"),
                 Aliases = GetPropertyValueOrNull(item, "Aliases"),
-                VersionOverride = versionOverrideRange
+                VersionOverride = versionOverrideRange,
+                PrivateAssetIndependentEnabled = privateAssetIndependentEnabled
             };
 
             // Add warning suppressions
--- a/src/NuGet.Clients/NuGet.SolutionRestoreManager/VSNominationUtilities.cs
+++ b/src/NuGet.Clients/NuGet.SolutionRestoreManager/VSNominationUtilities.cs
@@ -306,7 +306,7 @@ namespace NuGet.SolutionRestoreManager
         /// </summary>
         internal static bool IsPrivateAssetIndependentEnabled(IEnumerable tfms)
         {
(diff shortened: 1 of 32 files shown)


src/NuGet.Core/NuGet.Commands/CommandRunners/PackCommandRunner.cs
```diff
@@ -568,17 +568,32 @@ private static void CalculateExcludes(IncludeExcludeFiles files, out string full
...
             ISet<PackageDependency> packageDependencies = new HashSet<PackageDependency>();
 
             foreach (LibraryDependency dependency in dependencies)
             {
-                LibraryIncludeFlags effectiveInclude = dependency.IncludeType & ~dependency.SuppressParent;
+                LibraryIncludeFlags effectiveInclude = default;
+
+                if (packPrivateAssetsFlow)
+                {
+                    effectiveInclude = dependency.IncludeType | (~dependency.SuppressParent & LibraryIncludeFlags.All);
```
erdembayar: Reasoning for `& LibraryIncludeFlags.All` to limit `~dependency.SuppressParent` to byte size. Otherwise `~none` would result in `65535` not `127`.

erdembayar: Should this be just? 
`effectiveInclude = ~dependency.SuppressParent & LibraryIncludeFlags.All;`

Technically we only care value of `dependency.SuppressParent` if we opted in right?

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/IncludeFlagUtils.cs
```diff
@@ -155,10 +155,12 @@ internal static class IncludeFlagUtils
...
-                        LibraryIncludeFlags typeIntersection =
-                            (node.DependencyType
-                            & dependency.IncludeType
-                            & (~dependency.SuppressParent));
+                        LibraryIncludeFlags typeIntersection = dependency.PrivateAssetIndependentEnabled ?
+                            node.DependencyType
+                                & (dependency.IncludeType | (~dependency.SuppressParent & LibraryIncludeFlags.All))
+                            : node.DependencyType
+                                & dependency.IncludeType
+                                & (~dependency.SuppressParent);
```
erdembayar, line 163: Should this be just? Technically we only care value of `dependency.SuppressParent` if we opted in right?
```
LibraryIncludeFlags typeIntersection = dependency.PrivateAssetIndependentEnabled ?
                            node.DependencyType
                                & (~dependency.SuppressParent & LibraryIncludeFlags.All)
                            : node.DependencyType
                                & dependency.IncludeType
                                & (~dependency.SuppressParent);
```

src/NuGet.Clients/NuGet.SolutionRestoreManager/VSNominationUtilities.cs
```diff
@@ -300,6 +300,15 @@ internal static bool IsCentralPackageTransitivePinningEnabled(IEnumerable tfms)
             return GetSingleNonEvaluatedPropertyOrNull(tfms, ProjectBuildProperties.CentralPackageTransitivePinningEnabled, MSBuildStringUtility.IsTrue);
         }
 
+        /// <summary>
+        /// Evaluates the msbuild properties and returns the value of the PrivateAssetIndependent property.
```
erdembayar, line 304: Need to create project-system issue for this

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
```diff
@@ -61,6 +61,8 @@ Copyright (c) .NET Foundation. All rights reserved.
     <GetReferenceNearestTargetFrameworkTaskSupportsTargetPlatformParameter>true</GetReferenceNearestTargetFrameworkTaskSupportsTargetPlatformParameter>
     <!-- Flag if the Central package file is enabled -->
     <_CentralPackageVersionsEnabled Condition="'$(ManagePackageVersionsCentrally)' == 'true' AND '$(CentralPackageVersionsFileImported)' == 'true'">true</_CentralPackageVersionsEnabled>
+    <!-- Flag if the Private Asset Independent is enabled -->
+    <_PrivateAssetIndependentEnabled Condition="'$(PrivateAssetIndependent)' == 'true'">true</_PrivateAssetIndependentEnabled>
```
erdembayar, line 65: Do we need this? It looks like in above line `_CentralPackageVersionsEnabled` used for some other type of calculation.

PR #4982, opened 2022-12-12
add missing test coverage for AbandonedMutexException in NuGet migrations logic

kartheekp-ms:
## Bug

Tracking: https://github.com/NuGet/Home/issues/12159

## Description

I noticed that test coverage is missing for a scenario when I addressed the linked (tracking) issue in https://github.com/NuGet/NuGet.Client/commit/41c08c5549ac3c738d5bf369014da059eff53fbd and https://github.com/NuGet/NuGet.Client/commit/3f3f9ecc62ead356dd0ec2f129d1c2806f8b73a4 commits. Basically, the scenario is, whenever a thread is abandoned because of the application has been terminated abruptly ([for example, by using Windows Task Manager](https://learn.microsoft.com/en-us/dotnet/api/system.threading.waithandle.waitone?view=net-7.0#Remarks)) then the next migration run executes successfully. No product changes are introduced in this PR.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [x] N/A 

- **Documentation**
  
  - [x] N/A

### DIFF
--- a/test/NuGet.Core.Tests/NuGet.Common.Test/MigrationRunnerTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Common.Test/MigrationRunnerTests.cs
@@ -88,7 +88,7 @@ namespace NuGet.Common.Test
 
             void AbandonMutex()
             {
-                _orphan.WaitOne();
+                _orphan.WaitOne(TimeSpan.FromMinutes(1), false);
                 // Abandon the mutex by exiting the method without releasing
             }
         }
--- a/test/NuGet.Core.Tests/NuGet.Common.Test/MigrationRunnerTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Common.Test/MigrationRunnerTests.cs
@@ -89,7 +89,7 @@ namespace NuGet.Common.Test
             void AbandonMutex()
             {
                 _orphan.WaitOne();
-                // Abandon the mutexes by exiting without releasing the it by invoking _orphan1.ReleaseMutex() method.
+                // Abandon the mutexes by exiting without releasing it by invoking _orphan1.ReleaseMutex() method.
             }
         }
     }
--- a/test/NuGet.Core.Tests/NuGet.Common.Test/MigrationRunnerTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Common.Test/MigrationRunnerTests.cs
@@ -89,7 +89,7 @@ namespace NuGet.Common.Test
             void AbandonMutex()
             {
                 _orphan.WaitOne();
-                // Abandon the mutexes by exiting without releasing it by invoking _orphan1.ReleaseMutex() method.
+                // Abandon the mutex by exiting the method without releasing
             }
         }
     }
--- a/test/NuGet.Core.Tests/NuGet.Common.Test/MigrationRunnerTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Common.Test/MigrationRunnerTests.cs
@@ -88,7 +88,7 @@ namespace NuGet.Common.Test
 
             void AbandonMutex()
             {
-                _orphan.WaitOne(TimeSpan.FromMinutes(1), false);
+                _ = _orphan.WaitOne(TimeSpan.FromMinutes(1), false);
                 // Abandon the mutex by exiting the method without releasing
             }
         }
--- a/test/NuGet.Core.Tests/NuGet.Common.Test/MigrationRunnerTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Common.Test/MigrationRunnerTests.cs
@@ -73,22 +73,21 @@ namespace NuGet.Common.Test
             Thread t = new Thread(new ThreadStart(AbandonMutex));
             t.Start();
             t.Join();
+            Assert.True(signal, userMessage: "Failed to acquire the mutex.");
+            
+            string directory = MigrationRunner.GetMigrationsDirectory();
+            if (Directory.Exists(directory))
+                Directory.Delete(path: directory, recursive: true);
 
-            if (signal)
-            {
-                string directory = MigrationRunner.GetMigrationsDirectory();
-                if (Directory.Exists(directory))
-                    Directory.Delete(path: directory, recursive: true);
+            // Act
+            MigrationRunner.Run();
 
-                // Act
-                MigrationRunner.Run();
-
-                // Assert
-                Assert.True(Directory.Exists(directory));
-                var files = Directory.GetFiles(directory);
-                Assert.Equal(1, files.Length);
-                Assert.Equal(Path.Combine(directory, "1"), files[0]);
-            }
+            // Assert
+            Assert.True(Directory.Exists(directory));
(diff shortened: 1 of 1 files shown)


test/NuGet.Core.Tests/NuGet.Common.Test/MigrationRunnerTests.cs
```diff
@@ -62,5 +62,35 @@ public void Run_WhenExecutedInParallelThenOnlyOneMigrationFileIsCreated_Success(
...
+            Mutex _orphan = new Mutex(false, "NuGet-Migrations");
+
+            // Arrange
+            Thread t = new Thread(new ThreadStart(AbandonMutex));
+            t.Start();
+            t.Join();
+
+            string directory = MigrationRunner.GetMigrationsDirectory();
+            if (Directory.Exists(directory))
+                Directory.Delete(path: directory, recursive: true);
```
dominoFire, line 80: I prefer using braces:

```suggestion
            if (Directory.Exists(directory))
            {
                Directory.Delete(path: directory, recursive: true);
            }
```

dominoFire (dismissed):
LGTM. One comment. Thanks!

test/NuGet.Core.Tests/NuGet.Common.Test/MigrationRunnerTests.cs
```diff
@@ -62,5 +62,35 @@ public void Run_WhenExecutedInParallelThenOnlyOneMigrationFileIsCreated_Success(
...
+            // Assert
+            Assert.True(Directory.Exists(directory));
+            var files = Directory.GetFiles(directory);
+            Assert.Equal(1, files.Length);
+            Assert.Equal(Path.Combine(directory, "1"), files[0]);
+
+            void AbandonMutex()
+            {
+                _orphan.WaitOne();
+                // Abandon the mutex by exiting the method without releasing
```
erdembayar, line 95: Should we log or write something to some file to make sure if mutex was indeed acquired here?

test/NuGet.Core.Tests/NuGet.Common.Test/MigrationRunnerTests.cs
```diff
@@ -62,5 +62,35 @@ public void Run_WhenExecutedInParallelThenOnlyOneMigrationFileIsCreated_Success(
...
+            // Arrange
+            Thread t = new Thread(new ThreadStart(AbandonMutex));
+            t.Start();
+            t.Join();
+
+            string directory = MigrationRunner.GetMigrationsDirectory();
+            if (Directory.Exists(directory))
+                Directory.Delete(path: directory, recursive: true);
+
+            // Act
```
erdembayar, line 82: Should call `t.Abort();` to simulate actual abortion/termination of thread which holding the mutex?

kartheekp-ms: No need because the local function `AbandonMutex` exits without invoking `ReleaseMutex()` on the mutex object.

test/NuGet.Core.Tests/NuGet.Common.Test/MigrationRunnerTests.cs
kartheekp-ms: I changed this line a bit in the latest commit, but the logic remains the same. `WaitOne();` returns `true` only if the mutex is acquired by the current thread.

erdembayar: Did you push your latest change yet? I still don't see any return with `true` here.

kartheekp-ms: I changed this line from `_orphan.WaitOne()` to `_ = _orphan.WaitOne(TimeSpan.FromMinutes(1), false);` because `WaitOne()` waits indefinitely if the mutex is owned by other thread but the overloaded method waits for max 1 minute. It returns `true` if the thread aquires the mutex before 1 minute or else `false` but don't wait indefinitely like `WaitOne();`. We do this in our product code also. https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Common/Migrations/MigrationRunner.cs#L52

Given that these tests are not run in parallel with other tests `_orphan.WaitOne(TimeSpan.FromMinutes(1), false);` always returns `true`. When this method returns true it means that the thread acquired the mutex. Used discard because we ignore the return value. I don't think that we should do anything here unless I am missing something.

test/NuGet.Core.Tests/NuGet.Common.Test/MigrationRunnerTests.cs
kartheekp-ms: I somehow prefer not to add braces for this case.  https://github.com/NuGet/NuGet.Client/blob/dev/docs/coding-guidelines.md suggests `Braces may be omitted only if the body of every block associated with an if/else if/.../else compound statement is placed on a single line.`

test/NuGet.Core.Tests/NuGet.Common.Test/MigrationRunnerTests.cs
erdembayar: How about something like this?
Now just assert return value is `true`
```
      boolean AbandonMutex()
            {
                return _orphan.WaitOne(TimeSpan.FromMinutes(1), false);
             }
```

kartheekp-ms: ThreadStart delegate handler signature matches methods that returns void. https://learn.microsoft.com/dotnet/api/system.threading.threadstart?view=net-7.0 . Given that `AbandonMutex` is a local function, I captured a bool variable to signal that mutex has been acquired and abandoned because the local function exits without releasing the mutex. If `signal is true` then we proceed with the test. If the `signal is false` then it doesn't make sense to run this new test because the `AbandonMutex` method never acquired the mutex to abandon it.

erdembayar: How about bit change like below? We expect `signal` never would be `false` considering we don't run this test in parallel.
```
              Assert.True(signal);
              Everything else is outside if conditions.
```

kartheekp-ms: Fixed in the latest commit.

dominoFire (approved):

erdembayar (approved):

test/NuGet.Core.Tests/NuGet.Common.Test/MigrationRunnerTests.cs
zivkan: @erdembayar just FYI, `Thread.Abort` has always been problematic in .NET, and since .NET Core 1.0 it's obsolete, and in fact throws if you try to call it: https://learn.microsoft.com/en-us/dotnet/api/system.threading.thread.abort?view=net-7.0#system-threading-thread-abort

PR #4990, opened 2022-12-27
Return default format in SemanticVersion's last chance ToString

### LINKED ISSUE #12330: StackOverflow in SemanticVersion.ToString
### NuGet Product Used

NuGet SDK

### Product Version

Nuget.Versioning 6.4.0

### Worked before?

unknown

### Impact

### Repro Steps & Context

See [the following code w/in SemanticVersion.cs](https://github.com/NuGet/NuGet.Client/blob/08a7a7bd17e504d808329dcc1ffc866d9f59d040/src/NuGet.Core/NuGet.Versioning/SemanticVersionBase.cs#L35):

image

This results in a stack overflow when calling `ToString()` when the format provider fails to format, which will happen if you inherit from `SemanticVersion` for your own purposes.

It's a pretty severe bug as `ToString()` is exactly what's called by JSON Serializers, so this blows up any object model that utilizes it and serializes to JSON.

If it's not intended for people to inherit from Semver, mark it `sealed`.

### Verbose Logs

_No response_

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12330

## Description

`SemanticVersion.ToString()` calls `ToNormalizedString()`, which calls `ToString("N", VersionFormatter.Instance)`, however, `VersionFormatter` does type checking and refuses to work except with `SemanticVersion` and `NuGetVersion`. In which case, `ToString(string, IFormatProvider)` will fall back to `ToString()`, which therefore causes infinite recursion until stack overflow.

This PR changes `VersionFormatter` to work with anything that extends `SemanticVersion`, so default `ToString()` works without having to override anything.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Versioning/VersionFormatter.cs
+++ b/src/NuGet.Core/NuGet.Versioning/VersionFormatter.cs
@@ -54,7 +54,7 @@ namespace NuGet.Versioning
         {
             if (formatType == typeof(ICustomFormatter)
                 || formatType == typeof(NuGetVersion)
-                || formatType == typeof(SemanticVersion))
+                || typeof(SemanticVersion).IsAssignableFrom(formatType))
             {
                 return this;
             }
--- a/test/NuGet.Core.Tests/NuGet.Versioning.Test/SemanticVersionTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Versioning.Test/SemanticVersionTests.cs
@@ -97,5 +97,23 @@ namespace NuGet.Versioning.Test
             Assert.False(result);
             Assert.Null(semanticVersion);
         }
+
+        [Fact]
+        public void ToString_ClassExtendingSemanticVersion_ReturnsDefaultFormat()
+        {
+            ExtendedSemanticVersion target = new(1, 2, 3);
+
+            string result = target.ToString();
+
+            Assert.Equal("1.2.3", result);
+        }
+
+        private class ExtendedSemanticVersion : SemanticVersion
+        {
+            public ExtendedSemanticVersion(int major, int minor, int patch)
+                : base(major, minor, patch)
+            {
+            }
+        }
     }
 }
--- a/src/NuGet.Core/NuGet.Versioning/SemanticVersionBase.cs
+++ b/src/NuGet.Core/NuGet.Versioning/SemanticVersionBase.cs
@@ -42,12 +42,10 @@ namespace NuGet.Versioning
         /// </summary>
         public virtual string ToString(string format, IFormatProvider formatProvider)
         {
-            string formattedString = null;
-
             if (formatProvider == null
-                || !TryFormatter(format, formatProvider, out formattedString))
+                || !TryFormatter(format, formatProvider, out string formattedString))
             {
-                formattedString = ToString();
+                formattedString = Major + "." + Minor + "." + Patch;
             }
 
             return formattedString;
--- a/src/NuGet.Core/NuGet.Versioning/SemanticVersionBase.cs
+++ b/src/NuGet.Core/NuGet.Versioning/SemanticVersionBase.cs
@@ -45,7 +45,7 @@ namespace NuGet.Versioning
             if (formatProvider == null
                 || !TryFormatter(format, formatProvider, out string formattedString))
             {
-                formattedString = Major + "." + Minor + "." + Patch;
+                return ToString();
             }
 
             return formattedString;


test/NuGet.Core.Tests/NuGet.Versioning.Test/SemanticVersionTests.cs
```diff
@@ -97,5 +97,23 @@ public void TryParseStrictReturnsFalseIfVersionIsNotStrictSemVer(string version)
             Assert.False(result);
             Assert.Null(semanticVersion);
         }
+
+        [Fact]
+        public void ToString_ClassExtendingSemanticVersion_ReturnsDefaultFormat()
+        {
+            ExtendedSemanticVersion target = new(1, 2, 3);
+
+            string result = target.ToString();
```
donnie-msft, line 106: use var

test/NuGet.Core.Tests/NuGet.Versioning.Test/SemanticVersionTests.cs
```diff
@@ -97,5 +97,23 @@ public void TryParseStrictReturnsFalseIfVersionIsNotStrictSemVer(string version)
             Assert.False(result);
             Assert.Null(semanticVersion);
         }
+
+        [Fact]
+        public void ToString_ClassExtendingSemanticVersion_ReturnsDefaultFormat()
```
donnie-msft, line 102: I don't understand why these tests only test `ToString()`, whereas the more interesting logic, and most of what your PR touches, is `ToString(string format, IFormatProvider formatProvider)`. Is that intentional?

zivkan: I did it this way because what we care about, from an SDK compatibility point of view (the so called "API contract"), is that customers can extend `SemanticVersion`, and then libraries like Newtonsoft.Json, or System.Text.Json, or WPF converters, can be called with reasonable results (not stack overflow).

If at some point in the future, someone refactors SemanticVersion, I don't care about the implementation details of `ToString(string, IFormatProvider)`, or even if that method gets deleted. What's more important to me (what I want to check for regressions), is that `MyCustomVersion.ToString()` works, which doesn't before this PR.

donnie-msft (dismissed):

martinrrm (dismissed):

erdembayar (dismissed):

erdembayar (approved):

donnie-msft (approved):

erdembayar:
I requeued failed tests, hopefully it passes.

brandonh-msft:
Will the SO still happen if one extends `NuGetVersion` now, though? Since we only added `.IsAssignableFrom` to `SemanticVersion`?

zivkan:
@brandonh-msft can you suggest another unit test that will fail, even with the code change?  I wrote the unit test in this PR before writing the code change, so I'm confident it fixes at least one stack overflow.

brandonh-msft:
yeah it would effectively be just like your current one but w/ `NuGetVersion` instead. Though now that I'm thinking about it, since `NuGetVersion` inherits from `SemanticVersion` your current fix is probably fine (but calls into question why check `NuGetVersion` at all now).

```csharp
        [Fact]
        public void ToString_ClassExtendingSemanticVersion_ReturnsDefaultFormat()
        {
            ExtendedNuGetVersion target = new(1, 2, 3, 4);

            string result = target.ToString();

            Assert.Equal("1.2.3.4", result);
        }

        private class ExtendedNuGetVersion : NuGetVersion
        {
            public ExtendedNuGetVersion(int major, int minor, int patch, int revision)
                : base(major, minor, patch, revision)
            {
            }
        }
```

src/NuGet.Core/NuGet.Versioning/VersionFormatter.cs
```diff
@@ -54,7 +54,7 @@ public object GetFormat(Type formatType)
         {
             if (formatType == typeof(ICustomFormatter)
                 || formatType == typeof(NuGetVersion)
-                || formatType == typeof(SemanticVersion))
+                || typeof(SemanticVersion).IsAssignableFrom(formatType))
```
brandonh-msft, line 57:
```suggestion
                || typeof(SemanticVersion).IsAssignableFrom(formatType))
```

zivkan: I put together a benchmark. Here are the results

|                 Method |       Implementation |      Mean |     Error |    StdDev |    Median |
|----------------------- |--------------------- |----------:|----------:|----------:|----------:|
| CompareSemanticVersion |             Fallback |  4.907 ns | 0.1339 ns | 0.1252 ns |  4.875 ns |
|    CompareNuGetVersion |             Fallback |  5.226 ns | 0.1459 ns | 0.2478 ns |  5.179 ns |
|   CompareCustomVersion |             Fallback | 24.932 ns | 0.5226 ns | 0.9816 ns | 24.726 ns |
|          CompareObject |             Fallback | 26.623 ns | 2.5718 ns | 7.5829 ns | 22.854 ns |
| CompareSemanticVersion | Singl(...)eFrom [22] |  6.285 ns | 0.6062 ns | 1.7682 ns |  6.419 ns |
|    CompareNuGetVersion | Singl(...)eFrom [22] | 13.948 ns | 0.3181 ns | 0.6279 ns | 13.804 ns |
|   CompareCustomVersion | Singl(...)eFrom [22] | 13.733 ns | 0.3106 ns | 0.4147 ns | 13.701 ns |
|          CompareObject | Singl(...)eFrom [22] | 17.396 ns | 0.3803 ns | 0.3735 ns | 17.406 ns |

In absolute terms, they're all very quick, the difference wouldn't be noticeable to humans unless a process does it an absurd number of times. Although technically removing the two `==` comparisons, NuGet's primary use case (visual studio, dotnet cli) would become marginally slower. However it would be faster for customers who do implement their own version by extending NuGet's version classes.  Anyway, our team's primary responsibility is the VS & dotnet CLI experience, so I appreciate the suggestion, but I'm not going to create a new PR to merge it.

<details>
<summary>code</summary>
<div>

```cs
using NuGet.Versioning;

public class Program
{
    static void Main() => BenchmarkRunner.Run<Program>();

    public enum Method
    {
        Fallback,
        SingleIsAssignableFrom
    }

    [ParamsAllValues]
    public Method Implementation { get; set; }

    private Func<Type, bool>? _compare;
    private SemanticVersion _semanticVersion = SemanticVersion.Parse("1.2.3");
    private NuGetVersion _nugetVersion = NuGetVersion.Parse("1.2.3");
    private CustomVersion _customVersion = new CustomVersion(1, 2, 3);

    [GlobalSetup]
    public void GlobalSetup()
    {
        _compare = Implementation switch
        {
            Method.Fallback =>
                type =>
                    type == typeof(SemanticVersion)
                    || type == typeof(NuGetVersion)
                    || typeof(SemanticVersion).IsAssignableFrom(type),
            Method.SingleIsAssignableFrom =>
                type =>
                    typeof(SemanticVersion).IsAssignableFrom(type),
            _ => throw new NotImplementedException()
        };
    }

    [Benchmark]
    public bool CompareSemanticVersion()
    {
        return _compare!(typeof(SemanticVersion));
    }

    [Benchmark]
    public bool CompareNuGetVersion()
    {
        return _compare!(typeof(NuGetVersion));
    }

    [Benchmark]
    public bool CompareCustomVersion()
    {
        return _compare!(typeof(CustomVersion));
    }

    [Benchmark]
    public bool CompareObject()
    {
        return _compare!(typeof(object));
    }

    private class CustomVersion : NuGetVersion
    {
        public CustomVersion(int major, int minor, int patch)
            : base(major, minor, patch)
        {
        }
    }
}
```

</div>
</details>

PR #4698, opened 2022-06-23
Rewrite solution initialization with vs hierarchy

nkolev92:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/1701

Regression? Last working version:

## Description

NuGet currently uses the DTE to get all the projects in the solution. Currently that involves a lot of conversions into IVsHierarchy. Sometimes more than once per check. This PR cleans up some of those unnecessary conversions in the GetAllEnvDTEProjectsAsync method and attempts to use IVsHierarchy as often as possible. 

- Switch from using `DTE.Solution` to `IVsSolution` to enumerate the projects. 
- In VSProjectBuildProperties, the initialization is usually done with a Project as a fallback for reading properties. To avoid actively using the DTE project, we have a different constructor that lazily uses the DTE project instead. 
- VSProjectAdapter had a guid field that wasn't being used. I removed that.
- I did deconstruct some extra async calls as they were unnecessary (only called from the UI thread anyways).
- When checking IsSupportedAsync for the DTE.Project, I reduced the number of DTE.Project to IVsHierarchy conversions to 1 exactly.
- The biggest complexity/risk is with the compatible check, rewriting it from DTE to IVsHierarchy 
  - IsSupportedAsync
     - VsHierarchyUtility.IsProjectCapabilityCompliant - runs on IVsHierarchy
     - envDTEProject.Kind != null && ProjectType.IsSupported(envDTEProject.Kind) - Runs on Project.
       - Replaced DTE.Project.Kind with VsHierarchyUtility.GetProjectTypeGuids 
     - VsHierarchyUtility.HasUnsupportedProjectCapability - runs on IVsHierarchy
     - I also added a utility method in ProjectNames so we don't have to lookup the hierarchy again.
- I have also made this feature flag toggleable. I want to be able to revert things without a code change in case I have an oversight. I have not added this feature flag in the vs repo yet. I'll do that once I confirm these changes have been approved.
 
Final notes:
* Not everything is being addressed here. There's a bunch of async that's unnecessary that can be cleaned up, but I'm trying to keep the changesets manageable. There will be more changes in this area.
* This doesn't mean `Project` won't be used during any project initialization. There are still a few places where we stll use Project. Similarly, future work can remove that.

@davkean , I'd appreciate your review. A few specific points I'd love your thoughts on.
- Is the IVsSolution project enum walk correct for all projects in VS? Any risk of double counting/special project types that I should be checking for in addition to what we have there? 
[...]

PR #4992, opened 2022-12-28
Fix WebSite projects opened from IIS

### LINKED ISSUE #12337: WebSite projects opened from IIS fail to install packages
### NuGet Product Used

Visual Studio Package Management UI, Visual Studio Package Manager Console

### Product Version

VS 17.4

### Worked before?

17.3

### Impact

### Repro Steps & Context

1. In IIS Manager, create a virtual directory as an application.
2. In VS, use "open website", select "local IIS", and then select the directory
3. Try to install a package that has `lib/<tfm>/*.dll` files.

An obscure error message is reported: 

> Failed to add reference to ‘Newtonsoft.Json’.
> 
> Invalid URI: The Authority/Host could not be parsed.

### Verbose Logs

_No response_

zivkan:
IPersistFileFormat.GetCurFile was strangely returning the IIS http(s) URL, not the local folder name.
GetCanonicalName is returning the path, with trailing \, so that Path.GetDirectory will still return the same path

## Bug

Fixes: https://github.com/NuGet/Home/issues/12337

Regression? **yes**
Last working version: 6.3.x

## Description

https://github.com/NuGet/NuGet.Client/pull/4698 was aimed at reducing usage of `DTE`, but it also changed `VsHierarchyUtility.GetProjectPath` from using `IVsHierarchy.GetCanonicalName` to `IPersistFileFormat.GetCurFile`.

However, when a WebSite project is loaded from the UI to select a local IIS directory, rather than loading a solution file that already has the path to the folder, `IPersistFileFormat` will strangely return the `http(s)` URL, not the filesystem path, to the project/website folder.

Given the previous PR was aimed at reducing `DTE` usage, and `IVsHierarchy` is not `DTE`, it's not clear to me why this was changed. It's not described in the PR either, so I don't see any reason to assume there's a problem changing it back.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] **Test exception**: Requires IIS to be installed on the test machine, IIS must have a web application configured, and the website project to be loaded from the IIS configuration, not a solution file or new project.
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.VisualStudio.Common/IDE/VsHierarchyUtility.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Common/IDE/VsHierarchyUtility.cs
@@ -25,8 +25,7 @@ namespace NuGet.VisualStudio
         public static string GetProjectPath(IVsHierarchy project)
         {
             ThreadHelper.ThrowIfNotOnUIThread();
-            var format = (IPersistFileFormat)project;
-            format.GetCurFile(out string projectPath, out uint _);
+            project.GetCanonicalName((uint)VSConstants.VSITEMID.Root, out string projectPath);
             return projectPath;
         }
 


src/NuGet.Clients/NuGet.VisualStudio.Common/IDE/VsHierarchyUtility.cs
```diff
@@ -25,8 +25,7 @@ public static class VsHierarchyUtility
         public static string GetProjectPath(IVsHierarchy project)
         {
             ThreadHelper.ThrowIfNotOnUIThread();
-            var format = (IPersistFileFormat)project;
-            format.GetCurFile(out string projectPath, out uint _);
+            project.GetCanonicalName((uint)VSConstants.VSITEMID.Root, out string projectPath);
```
zivkan, line 28: @nkolev92 any chance you remember why you changed this in your PR?  It's unrelated to DTE, so I don't think it has any of the perf/other issues of using DTE.

donnie-msft (approved):
Isn't it kind of a red flag that no test exists that attempted to install packages to a Website project? 
Edit: Oh right, we have EndToEnd tests for Websites, but not using DTE/VS. Hmm..

zivkan:
> Oh right, we have EndToEnd tests for Websites, but not using DTE/VS. Hmm..

I'm not sure what you mean by "not using DTE/VS".  But as I wrote in the PR checklist's test exception section, it's difficult to test because this bug happens only when the WebSite project is loaded from IIS. It doesn't happen when the WebSite project is loaded from the filesystem. Setting this up programmatically is non-trivial (assuming VS even has an API to programmatically "open project from local IIS URL"). Our test vendors have a manual test for this now, so they'll periodically/regularly check for regressions.

src/NuGet.Clients/NuGet.VisualStudio.Common/IDE/VsHierarchyUtility.cs
nkolev92: I'd reach out to David about this.
I changed it with his guidance so he'll likely know better.
I think it was not returning the correct value for some type of projects. It was null but we needed a non null value.

nkolev92: It was a special project type such like maybe shared project or c++/CLI

zivkan: I should have posted this a few days ago, but when debugging I found that for native C++ and C++/CLI projects, both APIs return the same value (the full path, including the `vcxproj` file. However, for .NET projects, `IVsHierarchy.GetCanonicalName` returns only the directory, not the project file, whereas `IPersistFileFormat.GetCurFile` does return the project file. This gets passed to the our project adapter in a parameter whose name expects the project filename.  I didn't do much manual testing, and I assumed that E2E test failures were flakiness, not regressions.

So, this needs to be investigated, but assuming it's flakiness, it's not clear to me how important it is that we get this working  the same. If we need it to keep returning the project file for C# projects (a concept that doesn't even exist for WebSite projects), then our `VsProjectAdapter` class can no longer be project type agnostic. We'll need it to intrinsically understand different project types and how different ones need to use different APIs. And doing that will be a much bigger change.

zivkan:
I spent half a day on this today, trying to change NuGet's assumption that all `VsProjectAdapter`s can work the same, and try to make it part of the project system specific abstractions.   It's really not easy 😞. This is going to take quite a bit of extra work to get working properly.

Switching to draft mode, although I'm not sure I'll have time to work on it this sprint, before the bot closes it as stale.

PaavoNieminen:
Hope this dos not get buried as the pain caused by this bug is somewhat annoying ...

zivkan:
Sorry, there's been a lot going on, so I haven't made the time to work on this. The PR that introduced the regression is a perf improvement, so at this point I think it's easiest to revert the whole commit. We'll need to find a better solution to get the perf win back again.

PR #4995, opened 2023-01-09
Recalculate PackageSource generated properties on Source change

### LINKED ISSUE #10276: PackageSource:  returns possibly incorrect hash code
`NuGet.Configuration.PackageSource` [computes its hash code in its constructor](https://github.com/NuGet/NuGet.Client/blob/788bc01a1b063a37841cdd6d035feb320e90e475/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSource.cs#L128) based on `name` and `source` values.  However, [the `Source` property is read-write](https://github.com/NuGet/NuGet.Client/blob/788bc01a1b063a37841cdd6d035feb320e90e475/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSource.cs#L26), which means that the hash code may be incorrect after a new source is set.

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/10276

Regression? No

## Description

* Stop using `Lazy<bool>` for `IsHttp`, `IsHttps`, and `IsLocal`. I don't believe the time to calculate these are great, so I don't see the benefit in lazy-evaluating them. Especially when the most important code path, Restore, will certainly cause `IsHttp` and `IsHttps` to be evaluated.
* remove the `readonly` modified from the calculated hash code.
* Do these calculations in `PackageSource.Source`'s setter, so that if the source value changes, the values are recalcuated.
* I took the opportunity to enable nullability checking on this file.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSource.cs
+++ b/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSource.cs
@@ -1,6 +1,8 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+#nullable enable
+
 using System;
 using System.Collections.Generic;
 using System.Globalization;
@@ -55,7 +57,7 @@ namespace NuGet.Configuration
                 {
                     _isHttps = false;
                     _isHttp = false;
-                    Uri uri = TrySourceAsUri;
+                    Uri? uri = TrySourceAsUri;
                     _isLocal = uri != null ? uri.IsFile : false;
                 }
             }
@@ -64,7 +66,7 @@ namespace NuGet.Configuration
         /// <summary>
         /// Returns null if Source is an invalid URI
         /// </summary>
-        public Uri TrySourceAsUri => UriUtility.TryCreateSourceUri(Source, UriKind.Absolute);
+        public Uri? TrySourceAsUri => UriUtility.TryCreateSourceUri(Source, UriKind.Absolute);
 
         /// <summary>
         /// Throws if Source is an invalid URI
@@ -81,15 +83,15 @@ namespace NuGet.Configuration
 
         public bool IsEnabled { get; set; }
 
-        public PackageSourceCredential Credentials { get; set; }
+        public PackageSourceCredential? Credentials { get; set; }
 
-        public string Description { get; set; }
+        public string? Description { get; set; }
 
         public bool IsPersistable { get; private set; }
 
         public int MaxHttpRequestsPerSource { get; set; }
 
-        public IReadOnlyList<X509Certificate> ClientCertificates { get; set; }
+        public IReadOnlyList<X509Certificate>? ClientCertificates { get; set; }
 
         /// <summary>
         /// Gets or sets the protocol version of the source. Defaults to 2.
@@ -134,7 +136,7 @@ namespace NuGet.Configuration
             bool isPersistable = true)
         {
             Name = name ?? throw new ArgumentNullException(nameof(name));
-            Source = source ?? throw new ArgumentNullException(nameof(source));
+            Source = _source = source ?? throw new ArgumentNullException(nameof(source));
             IsEnabled = isEnabled;
             IsOfficial = isOfficial;
             IsPersistable = isPersistable;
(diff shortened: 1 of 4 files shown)


src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSource.cs
```diff
@@ -17,20 +18,54 @@ public class PackageSource : IEquatable<PackageSource>
...
+                {
+                    _isHttps = true;
+                    _isHttp = true;
+                    _isLocal = false;
+                }
+                else if (value.StartsWith("http://", StringComparison.OrdinalIgnoreCase))
+                {
+                    _isHttps = false;
+                    _isHttp = true;
+                    _isLocal = true;
```
erdembayar:
```suggestion
                    _isLocal = false;
```

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSource.cs
```diff
@@ -17,20 +18,54 @@ public class PackageSource : IEquatable<PackageSource>
         /// </summary>
         public const int DefaultProtocolVersion = 2;
 
-        private readonly int _hashCode;
+        private int _hashCode;
+
+        private string _source;
```
erdembayar, line 22: Maybe we no need line 22(blank).

test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceTests.cs
```diff
@@ -49,5 +49,29 @@ public void AsSourceItem_WorksCorrectly()
 
             SettingsTestUtils.DeepEquals(source.AsSourceItem(), expectedItem).Should().BeTrue();
         }
+
+        [Fact]
```
erdembayar, line 54: Could we add more tests for `http://my.test/v3/index.json`, `localhost` to prevent from regression?

erdembayar (commented):
Looks good, just 3 comments.

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSource.cs
zivkan: Good catch!  Thank you!

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSource.cs
```diff
@@ -17,20 +18,53 @@ public class PackageSource : IEquatable<PackageSource>
...
+
+        public string Name { get; }
 
-        private bool? _isHttp;
-        private bool? _isHttps;
-        private bool? _isLocal;
+        public string Source
+        {
+            get => _source;
+            set
```
kartheekp-ms, line 32: Reading the issue description https://github.com/NuGet/Home/issues/10276, I guess, we should be good if we just make `Source` property read-only. Do we really need to go into this path of reevaluating all the properties based on a source change? `PackageSource` has `Credentials` & `ClientCertificates` properties also which is tied to a single source. If we are going in this direction, I think we should also ensure that other related properties are also set with appropriate values.

erdembayar (approved):

erdembayar (approved):

jeffkl (approved):

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSource.cs
zivkan: I started a conversation in Teams about exactly that idea, on the 4th of January, before I wrote tests and spent time making all the code prod-quality. Although the 4th of January was right after the New Year break, everyone was at work (well, except one who is on extended leave), so I think everyone had an opportunity to have a conversation. The only feedback I got at the time was that they thought it was overkill (although the idealist in me agrees with you, I wish this type was immutable). So, unfortunately there wasn't enough agreement in the team that spending a lot more effort to make API breaking changes was worth the benefit.

PR #4997, opened 2023-01-10
Switched to AsyncKeyedLock

MarkCiliaVincenti:
The current implementation of `KeyedLock` has a number of rare race conditions which can lead to problems. This PR switches to a third-party library that handles such race conditions and on top offers superior performance and reduced allocations.


NuGet.Config
```diff
@@ -7,10 +7,11 @@
     <add key="myget-legacy@Local" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/myget-legacy%40Local/nuget/v3/index.json" />
     <add key="nuget-build" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/nuget-build/nuget/v3/index.json" />
     <add key="vside" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/msft_consumption/nuget/v3/index.json" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
```
kartheekp-ms, line 10: As per https://learn.microsoft.com/en-us/nuget/concepts/security-best-practices#nuget-feeds guidance, we can't have nuget.org as package source.

kartheekp-ms (changes requested):
> The current implementation of KeyedLock has a number of rare race conditions which can lead to problems. 

Thank you for taking the time to create this PR. Can you please explain the race conditions you have noticed in the current implementation so that we can first check if we can address them without adding a new dependency? It would be hard for us (I am not saying impossible) to have a dependency on third party software because NuGet code ships in Visual Studio, dotnet SDK, NuGet.exe and NuGet packages.

MarkCiliaVincenti:
> > The current implementation of KeyedLock has a number of rare race conditions which can lead to problems.
> 
> Thank you for taking the time to create this PR. Can you please explain the race conditions you have noticed in the current implementation so that we can first check if we can address them without adding a new dependency? It would be hard for us (I am not saying impossible) to have a dependency on third party software because NuGet code ships in Visual Studio, dotnet SDK, NuGet.exe and NuGet packages.

It needs better analysis but with a quick look it's easy to see a couple of race conditions.

For example, if two (or more) threads concurrently try to get a key from the dictionary but it's not found, they will all try to insert into the dictionary, overwriting each other. Given that they wouldn't be using the same `LockState`, this effectively allows multiple threads running concurrently using the same key, exactly what this class is supposed to be preventing.

Also, if thread A manages to get the value from the dictionary, it will increment through `Interlocked.Increment`, but this might be temporarily locked by another thread B trying to decrement the count from 1 to 0. Thread B would have `count` set to 0 and would then dispose of the semaphore and remove the key from the dictionary, while thread A increments the count back to 1 and returns the `lockState` with the disposed semaphore. Meanwhile another thread C would be unable to find the key in the dictionary and creates it.

NuGet.Config
MarkCiliaVincenti: I understand and was expecting this, but in order to test this it needs this for now, otherwise the NuGet package cannot be found in those other resources.

zivkan:
> For example, if two (or more) threads concurrently try to get a key from the dictionary but it's not found, they will all try to insert into the dictionary, overwriting each other. Given that they wouldn't be using the same LockState, this effectively allows multiple threads running concurrently using the same key, exactly what this class is supposed to be preventing.

You appear to be describing [the method `GetOrCreate`](https://github.com/NuGet/NuGet.Client/blob/ac1c958f82932b52bf71f0ff7b81714327a66d3b/src/NuGet.Core/NuGet.Common/KeyedLock.cs#L82). However, that method is private and called in two places, [`Enter`](https://github.com/NuGet/NuGet.Client/blob/ac1c958f82932b52bf71f0ff7b81714327a66d3b/src/NuGet.Core/NuGet.Common/KeyedLock.cs#L64) and [`EnterAsync`](https://github.com/NuGet/NuGet.Client/blob/ac1c958f82932b52bf71f0ff7b81714327a66d3b/src/NuGet.Core/NuGet.Common/KeyedLock.cs#L34). Both of those are guarded by the class field `SeamphoreSlim _dictionaryLock`. So, only one thread should be capable of running `GetOrCreate()` at a time, unless there's a bug in how `_dictionaryLock` is being used.

> Also, if thread A manages to get the value from the dictionary, it will increment through Interlocked.Increment, but this might be temporarily locked by another thread B trying to decrement the count from 1 to 0. Thread B would have count set to 0 and would then dispose of the semaphore and remove the key from the dictionary, while thread A increments the count back to 1 and returns the lockState with the disposed semaphore. Meanwhile another thread C would be unable to find the key in the dictionary and creates it.

[`Exit`](https://github.com/NuGet/NuGet.Client/blob/ac1c958f82932b52bf71f0ff7b81714327a66d3b/src/NuGet.Core/NuGet.Common/KeyedLock.cs#L124-L132) and [`ExitAsync`](https://github.com/NuGet/NuGet.Client/blob/ac1c958f82932b52bf71f0ff7b81714327a66d3b/src/NuGet.Core/NuGet.Common/KeyedLock.cs#L106-L114) are also guarded by the `_dictionaryLock`, so it should be impossible for `Cleanup()` to be called by more than one thread concurrently, or for it to be called at the same time that `GetOrCreate()` is being called.

The reason that `CleanUp` and `GetOrCreate` exist is to minimise duplicated code since `Enter` and `EnterAsync`, and `Exit` and `ExitAsync` need to use different APIs on `SemaphoreSlim` to avoid doing sync over async, or calling async APIs in a sync method.

> it's easy to see a couple of race conditions.

I suggest avoiding the use of such language ("easy to see") because if it was an obvious bug, it could be perceived as being judgemental against the PR author (disclaimer, it was me) and the reviewers, all of which are responsible for catching and preventing easy to see bugs when this code was being introduced. Saying something like "I think that there are bugs, for example" allows you to send the same message (about the bug) without risk of implying anything about the work other people did.

Meanwhile, I notice that your appear the be the author of the library that this PR is switching to, so in the context of this PR it's not actually a 3rd party library (you being one of the parties involved). While I admire your attempt in getting your library used in more products, I don't think that claiming there are bugs without evidence, and suggesting that the library is 3rd party, is the best way to go about it.

If there are bugs in my implementation, I absolutely want to know about it, so I can learn and improve. However, I'm concerned that the motive behind this pull request might not be entirely about fixing a bug in NuGet.

MarkCiliaVincenti:
Apologies @zivkan, I totally missed the guards by Enter/EnterAsync. Since performance is not critical for this path, I am closing this PR because I understand the cons outweigh the pros within this context.

MarkCiliaVincenti:
Just one thought though, @zivkan, what is the point of using Interlocked.Increment in a part of code that can only have one concurrent thread modifying the value? Unless I'm missing something, it feels superfluous.

PR #4999, opened 2023-01-10
Add 3 new test cases for Dotnet SDK style project to Install/Update/Un…

v-luzh:
## Bug
Fixes: https://github.com/NuGet/Client.Engineering/issues/2116

## Description
**Purpose:**
Add test cases to test installing/updating/uninstalling package into a Dotnet SDK style project from PM UI with “PackageReference” format by default.

**Add three test cases:**

InstallPackageToNetCoreProjectFromUI: Install particular package to a C# Net Standard Class Library Project from PM UI.
UpdatePackageToNetCoreProjectFromUI: Update package for a C# Net Standard Class Library Project in PM UI.
UninstallPackageFromNetCoreProjectFromUI: Uninstall particular package to a C# Net Standard Class Library Project from PM UI.

**Notes:**
The test cases were added into the end of the ...\NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NetCoreProjectTestCase.cs file.

## PR Checklist

- [X] PR has a linked issue.

- **Tests**
  - [X] Automated tests added

### DIFF
--- a/test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NetCoreProjectTestCase.cs
+++ b/test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NetCoreProjectTestCase.cs
@@ -258,5 +258,100 @@ namespace NuGet.Tests.Apex
         {
             yield return new object[] { ProjectTemplate.NetStandardClassLib };
         }
+
+        [NuGetWpfTheory]
+        [MemberData(nameof(GetNetCoreProjects))]
+        public void InstallPackageToNetCoreProjectFromUI(ProjectTemplate projectTemplate, string projectName, string packageName, string packageVersion)
+        {
+            // Arrange
+            EnsureVisualStudioHost();
+            var dte = VisualStudio.Dte;
+            var solutionService = VisualStudio.Get<SolutionService>();
+            solutionService.CreateEmptySolution();
+
+            var project = solutionService.AddProject(ProjectLanguage.CSharp, projectTemplate, projectName);
+            VisualStudio.ClearOutputWindow();
+            solutionService.SaveAll();
+
+            // Act
+            CommonUtility.OpenNuGetPackageManagerWithDte(VisualStudio, XunitLogger);
+            var nugetTestService = GetNuGetTestService();
+            var uiwindow = nugetTestService.GetUIWindowfromProject(project);
+            uiwindow.InstallPackageFromUI(packageName, packageVersion);
+
+            // Assert
+            VisualStudio.AssertNuGetOutputDoesNotHaveErrors();
+            CommonUtility.AssertPackageReferenceExists(VisualStudio, project, packageName, packageVersion, XunitLogger);
+        }
+
+        public static IEnumerable<object[]> GetNetCoreProjects()
+        {
+            yield return new object[] { ProjectTemplate.NetStandardClassLib, "NetStandardClassLib", "NetStandard.Library", "2.0.1" };
+        }
+
+
+        [NuGetWpfTheory]
+        [MemberData(nameof(UpdateNetCoreProjects))]
+        public void UpdatePackageToNetCoreProjectFromUI(ProjectTemplate projectTemplate, string projectName, string packageName, string installVersion, string updateVersion)
+        {
+            // Arrange
+            EnsureVisualStudioHost();
+            var dte = VisualStudio.Dte;
+            var solutionService = VisualStudio.Get<SolutionService>();
+            solutionService.CreateEmptySolution();
+
+            var project = solutionService.AddProject(ProjectLanguage.CSharp, projectTemplate, projectName);
+            VisualStudio.ClearOutputWindow();
+            solutionService.SaveAll();
+
+            // Act
+            CommonUtility.OpenNuGetPackageManagerWithDte(VisualStudio, XunitLogger);
+            var nugetTestService = GetNuGetTestService();
+            var uiwindow = nugetTestService.GetUIWindowfromProject(project);
+            uiwindow.InstallPackageFromUI(packageName, installVersion);
+            uiwindow.SwitchTabToUpdate();
+            uiwindow.UpdatePackageFromUI(packageName, updateVersion);
+
+            // Assert
+            VisualStudio.AssertNuGetOutputDoesNotHaveErrors();
+            CommonUtility.AssertPackageReferenceExists(VisualStudio, project, packageName, updateVersion, XunitLogger);
+        }
+
+        public static IEnumerable<object[]> UpdateNetCoreProjects()
+        {
+            yield return new object[] { ProjectTemplate.NetStandardClassLib, "NetStandardClassLib", "NetStandard.Library", "2.0.2", "2.0.3" };
+        }
+
+        [NuGetWpfTheory]
+        [MemberData(nameof(UninstallNetCoreProjects))]
+        public void UninstallPackageFromNetCoreProjectFromUI(ProjectTemplate projectTemplate, string projectName, string packageName, string packageVersion)
+        {
+            // Arrange
+            EnsureVisualStudioHost();
+            var dte = VisualStudio.Dte;
+            var solutionService = VisualStudio.Get<SolutionService>();
+            solutionService.CreateEmptySolution();
+
+            var project = solutionService.AddProject(ProjectLanguage.CSharp, projectTemplate, projectName);
+            VisualStudio.ClearOutputWindow();
+            solutionService.SaveAll();
+
+            // Act
+            CommonUtility.OpenNuGetPackageManagerWithDte(VisualStudio, XunitLogger);
+            var nugetTestService = GetNuGetTestService();
+            var uiwindow = nugetTestService.GetUIWindowfromProject(project);
+            uiwindow.InstallPackageFromUI(packageName, packageVersion);
+            uiwindow.SwitchTabToInstalled();
+            uiwindow.UninstallPackageFromUI(packageName);
+
+            // Assert
+            VisualStudio.AssertNuGetOutputDoesNotHaveErrors();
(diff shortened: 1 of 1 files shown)


test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NetCoreProjectTestCase.cs
```diff
@@ -258,5 +258,100 @@ public static IEnumerable<object[]> GetNetCoreTemplates()
         {
             yield return new object[] { ProjectTemplate.NetStandardClassLib };
         }
+
+        [NuGetWpfTheory]
+        [MemberData(nameof(GetNetCoreProjects))]
+        public void InstallPackageToNetCoreProjectFromUI(ProjectTemplate projectTemplate, string projectName, string packageName, string packageVersion)
```
zivkan, line 264: Since there's only one test case defined in `GetNetCoreProjects`, how about making it a `NuGetWpfFact`, and using the string literals in the test code below?  That way when reading the test code, I don't need to jump around to different parts of the file to see what values are actually being passed in.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NetCoreProjectTestCase.cs
```diff
@@ -258,5 +258,100 @@ public static IEnumerable<object[]> GetNetCoreTemplates()
...
+            uiwindow.InstallPackageFromUI(packageName, packageVersion);
+
+            // Assert
+            VisualStudio.AssertNuGetOutputDoesNotHaveErrors();
+            CommonUtility.AssertPackageReferenceExists(VisualStudio, project, packageName, packageVersion, XunitLogger);
+        }
+
+        public static IEnumerable<object[]> GetNetCoreProjects()
+        {
+            yield return new object[] { ProjectTemplate.NetStandardClassLib, "NetStandardClassLib", "NetStandard.Library", "2.0.1" };
```
zivkan, line 289: I thought the .NET SDK automatically adds `NetStandard.Library` to `netstandard` TFMs. If that's right, then using `NetStandard.Library` as a package to install in the project is not a good choice. `Newtonsoft.Json` version `13.0.1` is probably a better choice, but ideally it's a package that's already installed by other tests, to minimize the number of network calls tests will make (in the past we had flaky tests when the CI agent's network stopped working temporarily, although we might have had an HTTP retry bug as well).

My preference would actually be to install a generated package from a local filesystem source, because I wish none of our tests made any network requests, but I'm not sure we have any examples in Apex tests already that do this. I think it would be great to have a test fixture that create the test package(s) one time, and multiple tests share them, to avoid having tests re-create similar test packages over and over in a single test execution.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NetCoreProjectTestCase.cs
v-luzh: Hi @zivkan, actually we have 5 test cases defined in GetNetCoreProjects as below (changed the last package to "Newtonsoft.Json" just now) when we were writing the series of test cases. However, we just kept one because we worry about the time problem if adding them all to everyday build. So, should we try the five cases with [NuGetWpfTheory] as the original idea or change them to [NuGetWpfFact]?

```
        public static IEnumerable<object[]> GetNetCoreProjects()
        {
            yield return new object[] { ProjectTemplate.NetCoreConsoleApp, "NetCoreConsole", "Automapper", "11.0.0" };
            yield return new object[] { ProjectTemplate.NetCoreClassLib, "NetCoreClassLib", "Microsoft.NETCore.Platforms", "6.0.6" };
            yield return new object[] { ProjectTemplate.NetCoreXUnitTest, "NetCoreXUnit", "NUnit", "3.13.2" };
            yield return new object[] { ProjectTemplate.ASPNETCoreWebApplication, "NetCoreWeb", "Microsoft.NETCore.Targets", "3.1.0" };
            yield return new object[] { ProjectTemplate.NetStandardClassLib, "NetStandardClassLib", "Newtonsoft.Json", "13.0.1" };
        }
```

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NetCoreProjectTestCase.cs
v-luzh: Thanks for your suggestion, we will change the package `NetStandard.Library` to `Newtonsoft.Json`.

The reason why we chose the package with network request, because the test point of the case is to test installing/updating/uninstalling packages from network source. If we change it to installing/updating/uninstalling packages from a local source, that would be another case talking about. We will use the package that's already installed by other tests to minimize the number of network calls.

zivkan: Network requests are made via `System.Net.Http.HttpClient`, which is part of the .NET Base Class Library, so in my opinion, it's the .NET team's responsibility to test. NuGet has unit tests to make sure that NuGet.Protocol uses HttpClient correctly (both via `HttpClient` with a mock `HttpMessageHandler`, and also via HTTP server's listening on `localhost`, so it doesn't actually use the network). I don't see it as something that provides us with much safety in a Visual Studio integration test.

It's been a long time since I've seen a flaky network cause a test failure to nuget.org, but I'd still like to minimize risk of unreliable tests preventing pull requests from being merged.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NetCoreProjectTestCase.cs
zivkan: You must be looking at something different to me, because in this PR I see:

* [`GetNetCoreProjects`](https://github.com/NuGet/NuGet.Client/pull/4999/files#diff-15febf9b9c2640538162ca786717f618186c3c0b448c4d41dde8537354b23e23R287-R290)
* [`UpdateNetCoreProjects`](https://github.com/NuGet/NuGet.Client/pull/4999/files#diff-15febf9b9c2640538162ca786717f618186c3c0b448c4d41dde8537354b23e23R320-R323)
* [`UninstallNetCoreProjects`](https://github.com/NuGet/NuGet.Client/pull/4999/files#diff-15febf9b9c2640538162ca786717f618186c3c0b448c4d41dde8537354b23e23R352-R355)

Each of which only have 1 case.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NetCoreProjectTestCase.cs
zivkan: Also, maybe this is similar to the discussion we started with the previous Apex test you added, `NuGet.Tests.Apex` runs on every pull request. If you'd like to write tests that have a lot more duplicated, exhaustive tests, even though we believe it's testing the same implementation as what we already have tests for, then we can investigate [adding a new test project that runs once a day/week](https://github.com/NuGet/Client.Engineering/issues/2020).

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NetCoreProjectTestCase.cs
v-luzh: Sorry, I didn't describe clearly. At the beginning, we wanted to define 5 test cases as above. But we thought it would waste time for our Dev build if merge all the 5 ones. So, we deleted 4 from it and reserved the fifth project "NetStandardClassLib" as you see in the PR.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NetCoreProjectTestCase.cs
v-luzh: > Network requests are made via `System.Net.Http.HttpClient`, which is part of the .NET Base Class Library, so in my opinion, it's the .NET team's responsibility to test. NuGet has unit tests to make sure that NuGet.Protocol uses HttpClient correctly (both via `HttpClient` with a mock `HttpMessageHandler`, and also via HTTP server's listening on `localhost`, so it doesn't actually use the network). I don't see it as something that provides us with much safety in a Visual Studio integration test.
> 
> It's been a long time since I've seen a flaky network cause a test failure to nuget.org, but I'd still like to minimize risk of unreliable tests preventing pull requests from being merged.

I see. As you mentioned above, should we use a package (e.g. Newtonsoft.Json) that's already installed by other tests in `NuGet.Tests.Apex` instead?

v-luzh: > Also, maybe this is similar to the discussion we started with the previous Apex test you added, `NuGet.Tests.Apex` runs on every pull request. If you'd like to write tests that have a lot more duplicated, exhaustive tests, even though we believe it's testing the same implementation as what we already have tests for, then we can investigate [adding a new test project that runs once a day/week](https://github.com/NuGet/Client.Engineering/issues/2020).

Yes, we'd like to write tests which can replace at least part of manual testing we do every day. We plan to add more and more projects gradually to our test cases after the test pipeline created. Before [the task](https://github.com/NuGet/Client.Engineering/issues/2020) is done, we just want to pick one project to write test case for one scenario and merge it to NuGet build, wishing this will bring minimum impact to your work. Does the plan make sense?

v-luzh:
I will close this PR since we got an error said I have unstaged changes when we trying to rebase but we don't remember having any changes after the previous pushing.

PR #5002, opened 2023-01-10
CLI: Add -protocolVersion option to nuget sources add/update commands

### LINKED ISSUE #9170: Add protocolVersion argument to nuget source add
Please read the following information before posting the issue.

## Details about Problem

NuGet product used (NuGet.exe | VS UI | Package Manager Console | dotnet.exe):

NuGet version (x.x.x.xxx): any

Worked before? If so, with which NuGet version: No

## Feature request

Ability to specify protocolVersion property when adding new package source with CLI

dotnokato:
## Bug

Fixes: https://github.com/NuGet/Home/issues/9170

## Description

Adds ability to specify `-protocolVersion` property when adding/updating package source with CLI

Accepted protocol versions are 2, 3.
When source provided is a local source, protocol version is always set to 2, hence not being written.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added

- **Documentation**
  
  <https://github.com/NuGet/docs.microsoft.com-nuget/pull/2981>


src/NuGet.Clients/NuGet.CommandLine/NuGetCommand.resx
```diff
@@ -921,4 +921,7 @@ nuget client-certs List</value>
   <data name="Option_Help" xml:space="preserve">
     <value>Show command help and usage information.</value>
   </data>
+  <data name="SourcesCommandProtocolVersionDescription" xml:space="preserve">
+    <value>The NuGet server protocol version to be used. The current version is "3". Supported in NuGet 3.0+.</value>
```
erdembayar:
```suggestion
    <value>The NuGet server protocol version to be used. The default version is "3". Supported in NuGet 3.0+.</value>
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -812,4 +812,10 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
     <value>The package reference {0} specifies a VersionOverride but the ability to override a centrally defined version is currently disabled.</value>
     <comment>0 - packagereference name</comment>
   </data>
-</root>
+  <data name="SourcesCommandProtocolVersionDescription" xml:space="preserve">
+    <value>The NuGet server protocol version to be used. The current version is "3". Supported in NuGet 3.0+.</value>
```
erdembayar:
```suggestion
    <value>The NuGet server protocol version to be used. The default version is "3". Supported in NuGet 3.0+.</value>
```

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
```diff
@@ -398,5 +408,15 @@ public static void ValidateCredentials(string username, string password, string
                 throw new CommandException(Strings.SourcesCommandCredentialsRequiredWithAuthTypes);
             }
         }
+
+        public static int ParseProtocolVersion(string protocolVersion)
+        {
+            if (!int.TryParse(protocolVersion, out var protocolVersionValue))
```
erdembayar: With current logic customer can request invalid protocol version like `5,6` etc.
We need to validate it like below.

https://github.com/NuGet/NuGet.Client/blob/d76a117c590f8a91b844013bba7ea9b60e469aa1/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/ListPackageCommand.cs#L195-L204

src/NuGet.Core/NuGet.Configuration/Settings/Items/SourceItem.cs
```diff
@@ -77,8 +77,7 @@ public override bool Equals(object other)
                 return true;
             }
 
-            return string.Equals(Key, source.Key, StringComparison.Ordinal) &&
-                string.Equals(ProtocolVersion, source.ProtocolVersion, StringComparison.Ordinal);
+            return string.Equals(Key, source.Key, StringComparison.Ordinal);
```
erdembayar: Why are we making change here? I believe it should stay as it's now.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Verbs.cs
```diff
@@ -43,6 +43,10 @@ internal partial class AddVerbParser
                         "--valid-authentication-types",
                         Strings.SourcesCommandValidAuthenticationTypesDescription,
                         CommandOptionType.SingleValue);
+                    CommandOption protocolVersion = SourceCmd.Option(
+                        "-v|--protocol-version",
```
erdembayar: `-v` abbreviation usually reserved verbosity, so in order to avoid confusion let's not use `-v` here.
Technically `-p` is closer to `--protocol-version`, but we don't have to implement shorter abbreviations for option. Just look at `--valid-authentication-types` 3-4 lines above, it doesn't have one.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Verbs.cs
```diff
@@ -377,6 +382,10 @@ internal partial class UpdateVerbParser
...
```
erdembayar: Same as above.

src/NuGet.Core/NuGet.Commands/Strings.resx
```diff
@@ -1041,4 +1041,7 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
     <value>Warning As Error: {0}</value>
     <comment>{0} is a warning message</comment>
   </data>
+  <data name="SourcesCommandValidProtocolVersion" xml:space="preserve">
+    <value>The protocol version specified is invalid. Provide a valid protocol version.</value>
```
erdembayar: Should we just list what are valid versions here?

erdembayar (commented):
Thank you for your contribution. 👏 
LGTM. Few comments.

src/NuGet.Core/NuGet.Configuration/Settings/Items/SourceItem.cs
dotnokato: This change is needed as otherwise using `nuget sources update` with protocol version specified results in adding new item to sources list with the same name as already existing instead of updating.

This additionally fixes a bug that is there right now where if you use `nuget sources update` on a source that has protocol version specified, e.g. nuget v3 feed, this operation is resulting in nuget config with two sources with the same name.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Verbs.cs
dotnokato: Changed

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Verbs.cs
dotnokato: Changed

dotnokato:
Thanks for all the comments. I've fixed some of them already. Will fix the rest later this week.

src/NuGet.Core/NuGet.Configuration/Settings/Items/SourceItem.cs
erdembayar: Thank you for explanation, sounds reasonable for me. What others think?

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
dotnokato: See above.

src/NuGet.Clients/NuGet.CommandLine/NuGetCommand.resx
dotnokato: I've further changed the wording of this description to better match the behaviour of code
- "2" is the default version which is not being written in xml
- "3" is called 'current' as per [nuget.config File Reference](https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file#packagesources)

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
dotnokato: I've added validation and defined 2, 3, 4 as valid versions  (as per <https://learn.microsoft.com/en-us/nuget/api/nuget-protocols#nuget-protocol-version-410>) 
Everything else is now treated as invalid.
I've added tests to verify that.

src/NuGet.Core/NuGet.Commands/Strings.resx
dotnokato: I've updated the error message to list versions which are valid.

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
```diff
@@ -398,5 +408,17 @@ public static void ValidateCredentials(string username, string password, string
                 throw new CommandException(Strings.SourcesCommandCredentialsRequiredWithAuthTypes);
             }
         }
+
+        public static int ParseProtocolVersion(string protocolVersion)
+        {
+            var currentlySupportedProtocolVersions = new List<int> { 2, 3, 4 };
```
erdembayar: I'm not aware of protocol version 4. Is it documented somewhere?

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/NuGetSourcesCommandTest.cs
```diff
@@ -416,6 +416,173 @@ public void SourcesCommandTest_AddWithUserNamePassword_UserDefinedConfigFile()
...
+                    string.Join(" ", args),
+                    true);
+
+                // Assert
+                Assert.Equal(0, result.ExitCode);
+                ISettings loadedSettings = Configuration.Settings.LoadDefaultSettings(configFileDirectory, configFileName, null);
+                SettingSection packageSourcesSection = loadedSettings.GetSection("packageSources");
+                Assert.Single(packageSourcesSection.Items);
+                SourceItem sourceItem = packageSourcesSection?.GetFirstItemWithAttribute<SourceItem>("key", "test_source");
+                Assert.Equal("3", sourceItem.ProtocolVersion);
```
erdembayar, line 583: Please assert logged error message too.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
```diff
@@ -493,6 +493,100 @@ private static void CreateXmlFile(string configFilePath, string nugetConfigStrin
...
+                Assert.True(result.Success, result.Output + " " + result.Errors);
+
+                ISettings loadedSettings = Settings.LoadDefaultSettings(root: configFileDirectory, configFileName: null, machineWideSettings: null);
+
+                SettingSection packageSourcesSection = loadedSettings.GetSection("packageSources");
+                SourceItem sourceItem = packageSourcesSection?.GetFirstItemWithAttribute<SourceItem>("key", "test_source");
+                Assert.Equal("3", sourceItem.ProtocolVersion);
+            }
+        }
+
```
erdembayar, line 714: Please add unhappy path tests for dotnet too.

src/NuGet.Clients/NuGet.CommandLine/NuGetCommand.resx
```diff
@@ -921,4 +921,7 @@ nuget client-certs List</value>
   <data name="Option_Help" xml:space="preserve">
     <value>Show command help and usage information.</value>
   </data>
+  <data name="SourcesCommandProtocolVersionDescription" xml:space="preserve">
+    <value>The NuGet server protocol version to be used. The current version is "3". Supported in NuGet 3.0+. Defaults to "2" if not specified.</value>
```
erdembayar: I don't understand it says currently version is `3`, same time it defaults to `2`. 
Do you mean `3` current effective version used by server team? Please clarify further.

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
dotnokato: I've found is referenced here <https://learn.microsoft.com/en-us/nuget/api/nuget-protocols#nuget-protocol-version-410> trying to find some list of valid versions and was really confused. Should I leave it or remove it?
I know that other places in code reference 3 as the max version.

erdembayar: @joelverhagen
Which versions are available? If I understood correctly only v2, v3 are available.

src/NuGet.Clients/NuGet.CommandLine/NuGetCommand.resx
dotnokato: I've tried to use wording similar to what is said in the table here <https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file#package-source-sections>.

Any suggestions to make it clearer are welcome.

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
joelverhagen: The naming here is pretty confusing. V2 (OData), V3 (Service index, JSON) are the protocol versions most people talk about. The `X-NuGet-Protocol-Version` header referenced here is a header that allows feeds (mainly NuGet.org) to reject older client implementations specifically on the push endpoint. The value could have been anything. `4.1.0` was picked because the original header name was "client version" but then we changed that a bit later on. So `4.1.0` is actually the first NuGet _client_ version that is compatible with the latest breaking change added to package push on NuGet.org.

I don't want to write a book here, so I'll just say that `2, 3` is correct. `2, 3, 4` is not correct.

See https://learn.microsoft.com/en-us/nuget/api/overview#versioning for more (general, not NuGet.org push specific) protocol versions info.

dotnokato: Thanks for clarification.
I've pushed update with only `2, 3` as correct versions.

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
dotnokato: Changed valid versions to `2,3` as per discussion in other thread.

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/NuGetSourcesCommandTest.cs
dotnokato: Added tests validating unhappy path and error message.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
dotnokato: Added tests validating unhappy path and error message.

dotnokato:
Thank you for the review and all the comments. They were very helpful. I have pushed changes for all issues raised. Please let me know if there is anything else missing or needs amending.

src/NuGet.Clients/NuGet.CommandLine/NuGetCommand.resx
dotnokato: I've pushed an update to the description which should be clearer.

src/NuGet.Core/NuGet.Configuration/Settings/Items/SourceItem.cs
jeffkl: Team Review: We think it would best if the code path that needs a custom comparison implements it that way rather than changing how all comparisons work.  This could be a breaking change at the moment to alter how these objects are compared.

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -812,4 +812,10 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
     <value>The package reference {0} specifies a VersionOverride but the ability to override a centrally defined version is currently disabled.</value>
     <comment>0 - packagereference name</comment>
   </data>
-</root>
+  <data name="SourcesCommandProtocolVersionDescription" xml:space="preserve">
+    <value>The NuGet server protocol version to be used. Currently supported versions are 2 and 3. See https://learn.microsoft.com/en-us/nuget/api/overview for information about the version 3 protocol. Defaults to "2" if not specified.</value>
+  </data>
```
erdembayar, line 817: @JonDouglas 
Could you please quick review newly added string resources in this PR?

erdembayar:
> Thank you for the review and all the comments. They were very helpful. I have pushed changes for all issues raised. Please let me know if there is anything else missing or needs amending.

I just queued the PR for CI build, let's make sure it passes all the tests. Please check 'Team triage' comment about hashcode change.

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
zivkan: The URL should not contain `/en-us`. Remove it, so that `learn.microsoft.com` can detect the customer's language from their browser's language settings: https://learn.microsoft.com/nuget/api/overview

src/NuGet.Core/NuGet.Configuration/Settings/Items/SourceItem.cs
erdembayar: Several tests are failing, please check why they're failing: [#1](https://github.com/NuGet/NuGet.Client/blob/9836b4469a575b6b1aa7c9072c28b2f4ef6ec91d/test/NuGet.Core.Tests/NuGet.Configuration.Test/SettingsFileParsingTests/SourceItemTests.cs#L128), [#2](https://github.com/NuGet/NuGet.Client/blob/9836b4469a575b6b1aa7c9072c28b2f4ef6ec91d/test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs#L1029), [#3](https://github.com/NuGet/NuGet.Client/blob/9836b4469a575b6b1aa7c9072c28b2f4ef6ec91d/test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs#L1624).

erdembayar:
> > Thank you for the review and all the comments. They were very helpful. I have pushed changes for all issues raised. Please let me know if there is anything else missing or needs amending.
> 
> I just queued the PR for CI build, let's make sure it passes all the tests. Please check 'Team triage' comment about hashcode change.

Few tests are failing, check my [comment](https://github.com/NuGet/NuGet.Client/pull/5002#discussion_r1073771881). We're almost ready to merge after addressing current PR comments.

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
dotnokato: `/en-us` removed from the URL

src/NuGet.Core/NuGet.Configuration/Settings/Items/SourceItem.cs
dotnokato: Thank you for comments and running tests.

@erdembayar, the tests are failing due to aforementioned change in the comparison. My bad for missing that those tests failed. I'll revert the change.

@jeffkl 
What should be the expected behaviour in the below scenario?


Source item in `nuget.config` before change: `<add key="Source-Name" value="Legacy-Source" />`
Command: `nuget sources update -Name "Source-Name" -Source "New-Source" -ProtocolVersion 3`

Looking at [#2](https://github.com/NuGet/NuGet.Client/blob/9836b4469a575b6b1aa7c9072c28b2f4ef6ec91d/test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs#L1029) seems that the above command should result in `nuget.config` ending up with two lines for `Source-Name`:
- `<add key="Source-Name" value="Legacy-Source" />`
- `<add key="Source-Name" value="New-Source" protocolVersion="3" />`

Please confirm if my understanding is correct or different behaviour is expected.
I will then push the revert for comparison and updated tests.

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
erdembayar: Thank you for addressing this comment.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
```diff
@@ -493,6 +493,208 @@ private static void CreateXmlFile(string configFilePath, string nugetConfigStrin
...
+        [PlatformTheory(Platform.Windows)]
```
erdembayar: By the way why we're only testing on Windows? Please use `Theory` instead.

src/NuGet.Core/NuGet.Configuration/Settings/Items/SourceItem.cs
erdembayar: If I understand correctly Team triage [comment](https://github.com/NuGet/NuGet.Client/pull/5002#discussion_r1072625616) asks you revert this comparison logic change.  I understand your intention, but it might break other scenarios we're not aware of. Once this's reverted then I believe failing tests would be passing.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
dotnokato: I have used the same attribute as other tests in this file for consistency. The only platform that those are marked for is Windows.
Shall I change whole class or only newly added tests?

src/NuGet.Core/NuGet.Configuration/Settings/Items/SourceItem.cs
dotnokato: I have pushed the revert for comparison logic. Previously failing tests [#1](https://github.com/NuGet/NuGet.Client/blob/9836b4469a575b6b1aa7c9072c28b2f4ef6ec91d/test/NuGet.Core.Tests/NuGet.Configuration.Test/SettingsFileParsingTests/SourceItemTests.cs#L128), [#2](https://github.com/NuGet/NuGet.Client/blob/9836b4469a575b6b1aa7c9072c28b2f4ef6ec91d/test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs#L1029), [#3](https://github.com/NuGet/NuGet.Client/blob/9836b4469a575b6b1aa7c9072c28b2f4ef6ec91d/test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs#L1624) succeeded when running locally.

That however causes change in behaviour of `dotnet sources update` command when `-protocolVersion` is specified and is different than the value in nuget.config. That, in turn, causes tests added for new option to fail.

While taking closer look at failed tests mentioned by you, I am not sure if the behaviour of `sources update` command I implemented for `-ProtocolVersion` is the correct one since it is inconsistent with other commands (e.g. in test [#2](https://github.com/NuGet/NuGet.Client/blob/9836b4469a575b6b1aa7c9072c28b2f4ef6ec91d/test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs#L1029)).
That's the reason for [my previous comment](https://github.com/NuGet/NuGet.Client/pull/5002#discussion_r1080616770).

Without the comparison logic change, the code behaves as in scenario described. `nuget.config` has two lines for the source with the same name and different values and protocol versions.
If that is the expected behaviour, I need to fix tests for sources command with Protocol Version specified. They are failing right now, expecting only one line for source in `nuget.config` after sources update command.

Alternatively, I need to amend the behaviour of the `PackageSourceProvider.UpdatePackageSource` for this case. That does seem quite cumbersome without the comparison logic change, so any suggestions and pointers are welcome. 

With that said, I am happy to go either way with the implementation.

I hope that better explains my intent 😃

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
erdembayar: In that case, please keep it as it's now.

src/NuGet.Core/NuGet.Configuration/Settings/Items/SourceItem.cs
dotnokato: I've pushed a change that retains earlier behaviour from this PR but does not require comparison logic change.

Please let me know if that is directionally correct.

erdembayar: Thank you for pushing new change. I just requeued the CI build, let's see how it goes.

erdembayar: Some tests failed, could be flaky test, I just requeued the failed tests.

dotnokato: Hello @erdembayar

I can see that some of the tests failed in CI.
Would you be able to provide information which tests need my attention?

What else do you need me to do so that this can be merged?

erdembayar: It looks those tests are not connected to your change, there're some flaky Apex tests.
Please rebase to dev and push your changes, we need to requeue it.

dotnokato: Thanks for the reply. I've rebased and squashed my change and pushed it to the branch.

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
```diff
@@ -561,13 +561,19 @@ public void UpdatePackageSource(PackageSource source, bool updateCredentials, bo
...
+                if (!string.Equals(newSource.Source, existingSource.Source, StringComparison.OrdinalIgnoreCase) && newSource.IsPersistable)
                 {
                     Settings.AddOrUpdate(ConfigurationConstants.PackageSources, newSource.AsSourceItem());
                     isDirty = true;
                 }
 
+                if ((newSource.ProtocolVersion != existingSource.ProtocolVersion) && newSource.IsPersistable)
+                {
+                    Settings.Remove(ConfigurationConstants.PackageSources, existingSource.AsSourceItem());
+                    Settings.AddOrUpdate(ConfigurationConstants.PackageSources, newSource.AsSourceItem());
```
erdembayar: @dotnokato 
PR looks great. Just 1 tiny improvement. 
Here we're writing to config file 2nd time if protocol version is different after writing same info in line 566 if new and old sources urls are different and protocols are different too.
Could you improve to make it just 1 write to disc?

jeffkl (dismissed):

nkolev92 (commented):
Can we handle the protocol version scenario with local feeds? They protocol version doesn't apply in that case, so we shouldn't write it.

dotnokato:
> Can we handle the protocol version scenario with local feeds? They protocol version doesn't apply in that case, so we shouldn't write it.

I've pushed change that handles this scenario and added tests for the same.

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
dotnokato: Good catch!
I've pushed updated code for the above.

erdembayar: Thank you. LGTM, I requeued the CI build and waiting for result.

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
```diff
@@ -561,8 +561,13 @@ public void UpdatePackageSource(PackageSource source, bool updateCredentials, bo
         {
             if (string.Equals(newSource.Name, existingSource.Name, StringComparison.OrdinalIgnoreCase))
             {
-                if ((!string.Equals(newSource.Source, existingSource.Source, StringComparison.OrdinalIgnoreCase) ||
-                    newSource.ProtocolVersion != existingSource.ProtocolVersion) && newSource.IsPersistable)
+                if ((newSource.ProtocolVersion != existingSource.ProtocolVersion) && newSource.IsPersistable)
+                {
+                    Settings.Remove(ConfigurationConstants.PackageSources, existingSource.AsSourceItem());
```
erdembayar: @dotnokato 
Why we're doing `remove` here? But not in other case, what is the diff?

dotnokato: @erdembayar 
I used `remove` here as a way to work around issue with equality operator including ProtocolVersion in comparison. As a result `AddOrUpdate` was adding new line in config instead of updating the current one if protocol version value was being edited.
When changing the code, I've left the original logic (using only `AddOrUpdate`) for the path that was working correctly. I only used the `Remove`+`AddOrUpdate` in the one scenario that was necessary.

Come to think of it, I think those two cases can be handled together by only having the `Remove` branch.
Shall I update the code?

dotnokato: I've combined those two code paths and pushed rebased code.

zivkan: >I used remove here as a way to work around issue with equality operator including ProtocolVersion in comparison.

There are some people who are very particular about their config files, so I'm very confident that we'll get bug reports that the command has changed the order of sources, as we can see in the tests in this PR.

It's not my intention to add scope creep to this PR, but given the config file has an attribute `key`, it's not obvious to me why anything of the fields other than `key` (which I think is `PackageSource.Name`) should matter to `Settings.AddOrUpdate`.  Perhaps that should be changed, rather than this remove-add workaround?

dotnokato: I've pushed change that retains the original order of sources in config file. It is using `Remove`+`AddOrUpdate` hack, though.

@zivkan, original code submitted in for this PR contained change dropping `protocolVersion` from equality comparer in `SourceItem.cs`. It was reverted after [Team Review comment](https://github.com/NuGet/NuGet.Client/pull/5002#discussion_r1072625616).

If you believe that this is better course of action and other reviewers agree, I am happy to rework the code.

zivkan: Thanks for linking to the other comment. Unfortunately, I'm in a radically different time zone than the rest of the team, so I'm unable to participate in most of their meetings, like their PR review meeting, so I don't know what they discussed. I have some questions/objections to comments in that thread, so I'll reply there.

src/NuGet.Core/NuGet.Configuration/Settings/Items/SourceItem.cs
zivkan: > Team Review: We think it would best if the code path that needs a custom comparison implements it that way rather than changing how all comparisons work. This could be a breaking change at the moment to alter how these objects are compared.

While I usually agree with being conservative with potential breaking changes, or introducing regressions, what we've seen in this PR with this one command needing a hack to work around a genuine problem means that other commands are likely to have the same bug. So, it's an engineering tradeoff.  Although, creating a new comparison class and in other PRs migrating those other commands to the new comparison might be an option. But if the current comparison ends up with nothing using it, that means it's buggy, so if we fix it, then we don't need to add new public APIs.

I haven't read this PR carefully enough, or know NuGet.Configuration's code well enough, to suggest my ideal solution. My opinion will certainly change as I learn more information. But given my limited knowledge right now, I think the original change to the comparison feels the best to me, unless someone can convince me otherwise.

> Several tests are failing, please check why they're failing: [#1](https://github.com/NuGet/NuGet.Client/blob/9836b4469a575b6b1aa7c9072c28b2f4ef6ec91d/test/NuGet.Core.Tests/NuGet.Configuration.Test/SettingsFileParsingTests/SourceItemTests.cs#L128), [#2](https://github.com/NuGet/NuGet.Client/blob/9836b4469a575b6b1aa7c9072c28b2f4ef6ec91d/test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs#L1029), [#3](https://github.com/NuGet/NuGet.Client/blob/9836b4469a575b6b1aa7c9072c28b2f4ef6ec91d/test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs#L1624).

I strongly believe all 3 tests are buggy, which means the product is buggy, and tests are asserting the bugs continue, rather than helping the bugs get fixed.

The first test is for the `SourceItem.Equals` method, which validates that a different key or a different protocol version are treated as not equal.  But why is `protocolVersion` being compared, whereas `value` (url/path) is not?  It makes so sense. Either `SourceItem.Equals` should validate all all properties (key, value, protocolVersion), or only `key`. The current implementation doesn't make sense.

The other two tests show the impact of this bug:

The second test assets that after "saving" two package sources, one of which has the same key as the test setup ("arrange"), that [`children[0]`](https://github.com/NuGet/NuGet.Client/blob/9836b4469a575b6b1aa7c9072c28b2f4ef6ec91d/test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs#L1059) and [`children[2]`](https://github.com/NuGet/NuGet.Client/blob/9836b4469a575b6b1aa7c9072c28b2f4ef6ec91d/test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs#L1063) have the same key. To me this really feels like a bug. When I hear `key` in a programming context, I think dictionary key, which needs to be unique. The test assertion shows that it's not unique. Why is this test asserting this?

Similarly, I find the third test very strange. In the test setup, it [pre-configures two sources with the same key](https://github.com/NuGet/NuGet.Client/blob/9836b4469a575b6b1aa7c9072c28b2f4ef6ec91d/test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs#L1633-L1634). Why? Like I commented on the previous test, I expect keys to be unique. The test also asserts that there are still two source items with the same key.

As described by our docs on NuGet configuration, NuGet "accumulates" settings from multiple files. If I have `c:\nuget.config` with `<add key="a" value="https://url1.test/" />`, and `c:\src\nuget.config` with `<add key="a" value="https://url2.test/" />`, then I'd expect the "accumulated" settings to have only one source with `key="a"`, and the rules that "closest wins", meaning "url2". The bug that this PR exposes is that if one of the two configs sets `protocolVersion="3"`, then the package source won't be updated, it'll be duplicated.

And within a single config file, there's no reason why two sources with the same key should be allowed to exist. In fact, for package source credentials, the source key is used as the XML element name, preventing the possibility of duplicates (but also forcing encoding of XML invalid characters).  If a nuget.config has two different package sources with the same key (but different protocolVersion, and different URL), and there's `packageSourceCredentials` for that key, which URL(s) should NuGet use the credentials for?

A different example, something I believe this PR is adding, is if I have a config file with 1 source that does not specify protocolVersion, and I run `dotnet nuget update source MySource --protocolVersion 3`, it would be very unexpected for a new source to be added with the same key/URL as the already existing source, right? What I'd expect as a customer is that `protocolVersion="3"` be added to the existing source in my config file.

I've edited this comment a few times as I've been writing it, so I can't remember if I've kept this my next comment in what I wrote above, but I believe this PR has exposed bugs in both the product and the tests. Normally it's not the place of of a PR author fixing adding a feature to fix a different bug, especially when the PR author is an external contributor. But in this case, the bug is directly related to the feature being added.

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
```diff
@@ -83,6 +83,11 @@ public static void Run(AddSourceArgs args, Func<ILogger> getLogger)
                 newPackageSource.Credentials = credentials;
             }
 
+            if (!newPackageSource.IsLocal && !string.IsNullOrEmpty(args.ProtocolVersion))
```
erdembayar, line 86: Sorry for late review, I don't understand why this code path is updated. I believe it's separate problem.
Should we just address the `sources add/update` part this time?

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
erdembayar: Sorry, for looking back at code changes tiny improvement added 30 lines of code. Considering this code path is not hot code path previous implementation is good enough for me, I prefer clarity over 30 lines of code.

erdembayar (commented):
Sorry, I changed my mind on based on your latest commit. Please check my comment.

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
```diff
@@ -561,8 +561,36 @@ public void UpdatePackageSource(PackageSource source, bool updateCredentials, bo
...
+                            Settings.Remove(ConfigurationConstants.PackageSources, source);
+                            Settings.AddOrUpdate(ConfigurationConstants.PackageSources, source);
+                        }
+                        else if (source is SourceItem && (source as SourceItem).Key.Equals(newSource.Name, StringComparison.OrdinalIgnoreCase))
+                        {
+                            // Remove existing source and add new source it its place
+                            Settings.Remove(ConfigurationConstants.PackageSources, source);
+                            Settings.AddOrUpdate(ConfigurationConstants.PackageSources, newSource.AsSourceItem());
+                            newSourceAlreadyAdded = true;
+                            isDirty = true;
```
erdembayar: I believe this should be outside for loop.

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
erdembayar: I don't feel comfortable adding another 20+ lines of code just for the sake of keeping order in the config. If the user is really concerned about ordering, they can fix it manually. For me, clarity of code is more important than ordering or optimizations in this case. Considering this is not a hot path, I'm okay with reverting back to two writes to disk. With the latest code, we're making more than two read/writes, and actually, the performance is not better than before.

It seems that if the source is changed instead of the protocol, we are still removing and adding, which means that the ordering is not preserved. Why should we make an exception in this case with significant code changes?
Sorry, I edited my comments few times.

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
erdembayar: Sorry, I see another commend regarding this, I still believe we need to address one problem at the time.

jeffkl:
Team Review: We discussed this today and Erick is going to see if a custom comparer would be a good solution

zivkan:
@jeffkl (cc @erdembayar) in case I wasn't clear in the meeting yesterday, my vote was that the PR author's original implementation, where they modified the existing `SourceItem.Equals` implementation that caused the 3 tests to fail, should be re-instated, and that the tests should be fixed because [I strongly believe that the tests have bugs](https://github.com/NuGet/NuGet.Client/pull/5002#discussion_r1112783280). I don't think a new comparer is the best solution here.

This stack overflow question shows that `dotnet nuget add source -name` already has code somewhere to prevent having two sources configured with the same name, so to me it really makes sense that a single nuget.config file shouldn't allow it either: https://stackoverflow.com/questions/75661647/global-nuget-config-out-of-sync

dotnokato:
Hello,
Has the conclusion been reached on how this feature should be implemented and what is the scope?

erdembayar:
> Hello, Has the conclusion been reached on how this feature should be implemented and what is the scope?

Sorry, I was super busy with other priorities. I'll attend this PR soon, I need sometime.

dotnokato:
No worries. I understand as I was unable to work on this for the last few weeks myself.
Just wanted to make sure that the PR did not fall off the radar and won't get automatically closed :)

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
```diff
@@ -398,5 +408,17 @@ public static void ValidateCredentials(string username, string password, string
                 throw new CommandException(Strings.SourcesCommandCredentialsRequiredWithAuthTypes);
             }
         }
+
+        public static int ParseProtocolVersion(string protocolVersion)
```
kartheekp-ms: Please change the access modifier to `private` instead of `public` because this method is used in the same class.

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
```diff
@@ -398,5 +408,17 @@ public static void ValidateCredentials(string username, string password, string
...
+        {
+            var currentlySupportedProtocolVersions = new List<int> { 2, 3 };
```
kartheekp-ms: instead of hard coding protocol versions here, I would use PackageSource.DefaultProtocolVersion instead of 2. May be add a new variable named `MaxProtocolVersion` and then update the condition to accept a protocol version that falls within the the range of `DefaultProtocolVersion` to `MaxProtocolVersion`.

https://github.com/NuGet/NuGet.Client/blob/80c3c74ca5c6d67912a19833e8cc2b10f875dc95/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSource.cs#L19

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
```diff
@@ -561,8 +561,36 @@ public void UpdatePackageSource(PackageSource source, bool updateCredentials, bo
...
+                        }
+                    }
+                }
+                else if (!string.Equals(newSource.Source, existingSource.Source, StringComparison.OrdinalIgnoreCase) && newSource.IsPersistable)
```
kartheekp-ms:
``` suggestion
                else if (newSource.IsPersistable && !string.Equals(newSource.Source, existingSource.Source, StringComparison.OrdinalIgnoreCase))
```

kartheekp-ms (commented):
I am not yet done with reviewing the PR fully yet. Left minor comments to start with.

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
dotnokato: This method cannot be made private as it is in class [`internal static class RunnerHelper`](https://github.com/NuGet/NuGet.Client/blob/34f2638d7ebe5d9f4a33f74e9b900c0abad19437/src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs#LL325C39-L325C39) and is used in `AddSourceRunner` and `UpdateSourceRunner` classes.
It was marked `public static` similarly to other methods in this class.

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
dotnokato: I've pushed code addressing this.

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
dotnokato: Shall I revert this change then?

erdembayar:
Team triage: There was shift in priority for team.
Someone from team will drive this.

pf-uvonceumer:
I stumbled upon this issue while trying to integrate a Artifactory remote registry for an enterprise GitHub nuget registry. Turns out it is required to change the protocol version to version 3 to use the registry in dotnet.

So I would really appreciate that this feature is added as we need to enable our users to add the registry without having to manually edit %appdata%\NuGet\NuGet.Config.

ghost:
This PR has been automatically marked as stale because it has no activity for **7 days**. It will be closed if no further activity occurs **within another 90 days** of this comment. If it is closed, you may reopen it anytime when you're ready again, as long as you don't delete the branch.

src/NuGet.Core/NuGet.Configuration/Settings/Items/SourceItem.cs
zivkan: coincidentally, a Microsoft-internal tester found a user-scenario where this bug (different protocolVersion considered not equal) causes a user-experience bug. I have a PR to fix that, and once it's merged, I can update this PR and get this one merged: https://github.com/NuGet/NuGet.Client/pull/5334

dotnokato: Hello @zivkan,
Looking forward to #5334 being merged. After that I should be able to restart working on this PR.

dotnokato:
I have pushed update code to the branch with the following changes:
- merged dev to branch to get equality comparer changes #5334
- fixed merge conflicts
- fixed compilation errors and failing tests that were the result of codebase changes from when the PR was first submitted
- updated the implementation, effectively reverting back close to the initially proposed solution.

I've also run tests locally and all passed. That is after changing my OS language and regional format to `English (US)`.

Please let me know if there is anything else needing to be changed.

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
dotnokato: I've pushed new changes (see my comment [here](https://github.com/NuGet/NuGet.Client/pull/5002#issuecomment-1679707317)) reverting this bit as it is no longer needed.

src/NuGet.Core/NuGet.Configuration/Settings/Items/SourceItem.cs
dotnokato: Updated code pushed.

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/NuGetSourcesCommandTest.cs
```diff
@@ -411,6 +411,191 @@ public void SourcesCommandTest_AddWithUserNamePassword_UserDefinedConfigFile()
             }
         }
 
+        [Theory]
+        [InlineData(@"https://source.test", true)]
+        [InlineData(@"\\myserver\packages", false)]
+        public void SourcesCommandTest_AddWithProtocolVersion(string source, bool shouldWriteProtocolVersion)
```
zivkan: This test failed on Mac (we don't run Mono tests on Linux):

> Assert.Equal() Failure
> Expected: (null)
> Actual:   3

The test has me wondering if the behaviour is good though. On one hand `protocolVersion` isn't used for local feeds, so I understand not using it for local feeds. On the other hand, if a customer passes `-ProtocolVersion 3` on the command line, I don't like that it gets silently ignored. To me, an error message saying "local feeds don't support protocolVersion" feels better, as it's more explicit about what's going on.

dotnokato: Behaviour with local feeds was implemented as per [this comment](https://github.com/NuGet/NuGet.Client/pull/5002#pullrequestreview-1287837289).

dotnokato: I believe that the reason why test failed on Mac was due to `\\myserver\packages` not being treated as local source. I've split the test into two: one for remote, one for local. Test for local source uses different source path depending on the platform it runs on.

dtivel:
FYI @dotnokato:  https://github.com/NuGet/Home/issues/12820

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
```diff
@@ -475,6 +475,211 @@ private static void CreateXmlFile(string configFilePath, string nugetConfigStrin
...
+        [InlineData("1", false)]
+        [InlineData("2", true)]
+        [InlineData("3", true)]
+        [InlineData("4", false)]
+        [InlineData("5", false)]
+        public void Sources_WhenAddingSourceWithProtocolVersion_ProtocolVersionIsValidated(string protocolVersion, bool shouldSucceed)
```
zivkan: Why is this test marked as Windows only? I don't see anything obvious about this test that should prevent it from working on Linux or Mac.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
```diff
@@ -475,6 +475,211 @@ private static void CreateXmlFile(string configFilePath, string nugetConfigStrin
             }
         }
 
+        [PlatformTheory(Platform.Windows)]
+        [InlineData(@"https://source.test", true)]
+        [InlineData(@"\\myserver\packages", false)]
+        public void Sources_WhenAddingSourceWithProtocolVersion_GotAdded(string source, bool shouldWriteProtocolVersion)
```
zivkan: Since the test uses `SimpleTestPathContext`, it can use a sub-directory as the source path for the second case, and therefore avoid hardcoding a Windows UNC path, and therefore no longer be limited to running only on Windows.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
```diff
@@ -475,6 +475,211 @@ private static void CreateXmlFile(string configFilePath, string nugetConfigStrin
...
+                    string expectedErrorMessage = "The protocol version specified is invalid.";
+                    Assert.True(result.Output.Contains(expectedErrorMessage), "Expected error is " + expectedErrorMessage + ". Actual error is " + result.Output);
+                }
+            }
+        }
+
+        [PlatformTheory(Platform.Windows)]
+        [InlineData(@"https://source.test", true)]
+        [InlineData(@"\\myserver\packages", false)]
+        public void Sources_WhenUpdatingSourceWithProtocolVersion_GotUpdated(string source, bool shouldWriteProtocolVersion)
```
zivkan: Like the "add" tests above, using a subdirectory of the `TestDirectory` as a local package source, this can unblock the test from being restricted to Windows only.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
```diff
@@ -475,6 +475,211 @@ private static void CreateXmlFile(string configFilePath, string nugetConfigStrin
...
+            }
+        }
+
+        [PlatformTheory(Platform.Windows)]
+        [InlineData("1", false)]
+        [InlineData("2", true)]
+        [InlineData("3", true)]
+        [InlineData("4", false)]
+        [InlineData("5", false)]
+        public void Sources_WhenUpdatingSourceWithProtocolVersion_ProtocolVersionIsValidated(string protocolVersion, bool shouldSucceed)
```
zivkan: Like the other comment, I don't think this test needs to be limited to running on Windows only.

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
zivkan: @kartheekp-ms this conversation is not marked as resolved. Can you please check if you have any other comments?

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
zivkan: @kartheekp-ms this conversation is not marked as resolved. Can you please check if you have any other comments?

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/NuGetSourcesCommandTest.cs
dotnokato: My bad! I updated only tests for `update` and forgot about `add`. I've just pushed updated tests for `add`.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
dotnokato: This was set as Windows only as per [this comment](https://github.com/NuGet/NuGet.Client/pull/5002#discussion_r1083228447).

I've changed it to `[Theory]`.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
dotnokato: I've decided to split those tests for clarity, similarly to what I did in NuGetSourcesCommandTest.cs.

When I tried using relative path for source, `IsLocal()` was returning `false`

zivkan (approved):

PR #5010, opened 2023-01-18
Properly handle alternate console stream encodings in StaticGraphRestoreArguments

### LINKED ISSUE #12373: Static graph-based restore crashes on systems with alternate console encodings
### NuGet Product Used

dotnet.exe, MSBuild.exe

### Product Version

Visual Studio 17.5

### Worked before?

Yes, Visual Studio 17.4

### Impact

### Repro Steps & Context

In Windows Region settings, enable "Use Unicode UTF-8 for worldwide language support"

image

This seems to make Console.StandardInput have a 3-byte preamble, causing the logic that read the static graph restore arguments to not work correctly.

Reported by @vlada-shubina

Caused by https://github.com/NuGet/NuGet.Client/pull/4772

This is fixed in 17.6

### Workaround
Disable the Windows Region setting and reboot.

### Verbose Logs

_No response_

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12373

Regression? Yes Last working version: 17.4

## Description

When you enable alternate region encodings, Windows appears to write a preamble to `Console.StandardInput` causing the logic in `StaticGraphRestoreArguments` to fail to read the stream correctly.

My fix is to write take into account the encoding of the stream and skip the preamble if one is found.   This is done by looking at the encoding's preamble, reading the first 4 bytes of the stream, skipping the preamble, and reading the next set of bytes.  This results in the buffer containing the first integer in the stream so the rest of the code can read the content normally.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Build.Tasks/StaticGraphRestoreArguments.cs
+++ b/src/NuGet.Core/NuGet.Build.Tasks/StaticGraphRestoreArguments.cs
@@ -35,7 +35,7 @@ namespace NuGet.Build.Tasks
         {
             using var reader = new BinaryReader(stream, encoding, leaveOpen: true);
 
-            int count = SkipPreamble();
+            int count = SkipPreamble(reader, encoding);
 
             return new StaticGraphRestoreArguments
             {
@@ -45,10 +45,7 @@ namespace NuGet.Build.Tasks
 
             Dictionary<string, string> ReadDictionary(int count = -1)
             {
-                if (count == -1)
-                {
-                    count = reader.ReadInt32();
-                }
+                count = count == -1 ? reader.ReadInt32() : count;
 
                 var dictionary = new Dictionary<string, string>(capacity: count, StringComparer.OrdinalIgnoreCase);
 
@@ -60,51 +57,6 @@ namespace NuGet.Build.Tasks
                 }
 
                 return dictionary;
-            }
-
-            int SkipPreamble()
-            {
-                // Get the preamble from the current encoding which should only be a maximum of 4 bytes
-                byte[] preamble = encoding.GetPreamble();
-
-                if (preamble.Length == 0)
-                {
-                    // Return -1 to the caller if there is no preamble for the encoding meaning that the stream is at the beginning of the expected content
-                    return -1;
-                }
-
-                // Create a buffer for the preamble which should be a maximum of 4 bytes
-                byte[] buffer = new byte[4];
-
-                // Read 
-                int readBytes = reader.Read(buffer, 0, buffer.Length);
-
-                int matchingPreambleLength = 0;
-
-                for (int i = 0; i < preamble.Length; i++)
-                {
-                    if (buffer[i] != preamble[i])
-                    {
-                        break;
-                    }
-
-                    matchingPreambleLength++;
-                }
-
-                if (matchingPreambleLength == preamble.Length)
-                {
-                    int index = matchingPreambleLength;
-
-                    for (int i = 0; i < buffer.Length - matchingPreambleLength; i++)
-                    {
-                        buffer[i] = buffer[index];
-                    }
-
-                    reader.Read(buffer, buffer.Length - matchingPreambleLength, matchingPreambleLength);
-
-                }
-
-                return BitConverter.ToInt32(buffer, 0);
             }
         }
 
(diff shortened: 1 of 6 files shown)


Forgind (dismissed):
I'm wondering if a specific random byte might be better in case someone else does something similar, but I think this is good 🙂

jeffkl:
> I'm wondering if a specific random byte might be better in case someone else does something similar, but I think this is good 🙂

Good point, it appears that both `byte.MaxValue` (`255`/`0xFF`) and `byte.MinValue` (`0`/`0x00`) can be used in the byte order mark.  In my testing I was seeing `0xFF 0xBB 0xBF` (UTF8 BOM) which wouldn't collide, but I need something more future proof.  More research required...

donnie-msft (dismissed):

jeffkl:
I've realized now that I can emulate a stream with this functionality and write unit tests for it, will update soon...

src/NuGet.Core/NuGet.Build.Tasks/StaticGraphRestoreArguments.cs
```diff
@@ -63,7 +68,10 @@ public static StaticGraphRestoreArguments Read(Stream stream)
         /// <param name="stream">A <see cref="Stream" /> to write the arguments to.</param>
         public void Write(Stream stream)
         {
-            using BinaryWriter writer = new BinaryWriter(stream, Encoding.Default, leaveOpen: true);
+            using BinaryWriter writer = new BinaryWriter(stream, Encoding.UTF8, leaveOpen: true);
+
+            // Write a byte indicating the beginning of the content
+            writer.Write(byte.MaxValue);
```
erdembayar: Is just `byte.MaxValue` future proof enough?
Probably write several predefined sequential bytes something like `!NuGetStartSignal!` as bytes instead of just 1 byte.

jeffkl: Yes sorry completely disregard my first iteration of this pull request, I've updated it to actually look at the encoding preamble which is variable length and added unit tests.

jeffkl:
I've completely revamped my implementation, please re-review!

src/NuGet.Core/NuGet.Build.Tasks/StaticGraphRestoreArguments.cs
```diff
@@ -55,15 +61,60 @@ public static StaticGraphRestoreArguments Read(Stream stream)
...
 
         /// <summary>
         /// Writes the current arguments to the specified stream.
         /// </summary>
         /// <param name="stream">A <see cref="Stream" /> to write the arguments to.</param>
-        public void Write(Stream stream)
+        public void Write(StreamWriter streamWriter)
         {
-            using BinaryWriter writer = new BinaryWriter(stream, Encoding.Default, leaveOpen: true);
+            using BinaryWriter writer = new BinaryWriter(streamWriter.BaseStream, streamWriter.Encoding, leaveOpen: true);
```
zivkan, line 69: I don't know if this is a good or bad idea, but another option is to create our own instance of [`UTF8Encoding`, using the overload where we can disable the preamble](https://learn.microsoft.com/en-us/dotnet/api/system.text.utf8encoding.-ctor?view=net-7.0#system-text-utf8encoding-ctor(system-boolean)) for both the reader and writer. Then we don't need a `SkipPreamble()` method.

But honestly, I don't know enough about localization and internationalization to know if could create problems. Is it possible for customers to use a non-UTF8 encoding in their terminal? 🤷 I believe that for some languages, like chineese, japanese, koren, that UTF8 encoding results in more bytes than UTF16 (maybe UTF32 as well?). But for our use-case, I don't think that matters. Are there other risks? 🤷

src/NuGet.Core/NuGet.Build.Tasks/StaticGraphRestoreArguments.cs
```diff
@@ -55,15 +61,60 @@ public static StaticGraphRestoreArguments Read(Stream stream)
...
+                }
+
+                if (matchingPreambleLength == preamble.Length)
+                {
+                    int index = matchingPreambleLength;
+
+                    for (int i = 0; i < buffer.Length - matchingPreambleLength; i++)
+                    {
+                        buffer[i] = buffer[index];
+                    }
```
zivkan: [`Array.Copy`](https://learn.microsoft.com/en-us/dotnet/api/system.array.copy?view=net-7.0#system-array-copy(system-array-system-int32-system-array-system-int32-system-int32)). Although for max of 4 bytes, it really won't matter.

zivkan (dismissed):

src/NuGet.Core/NuGet.Build.Tasks/StaticGraphRestoreArguments.cs
jeffkl: So I spent a lot of time trying to figure out why the preamble is there.  If we created the stream, then you're absolutely right that using an instance of `UTF8Encoding` with the preamble off would work.

However, what I found under the debugger is that the operating system creates the Standard Input stream and something else is writing the preamble _before_ our code writes anything to the stream.  I tried to find a way to alter the encoding of the stream when its created but I couldn't find a way to.  Ideally during process creation I could specify the encoding and of course tell it to not write a preamble.  

The `ProcessStartInfo.StandardOutputEncoding` has always existed for all versions of .NET, but the `ProcessStartInfo.StandardInputEncoding` property was added until .NET Standard 2.1/.NET Core 2.1: https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.standardinputencoding?view=net-7.0#applies-to

What's better is that the `ProcessStartInfo.StandardInputEncoding` property's remarks say:

> If the value of the StandardInputEncoding property is null, the process uses the default standard input encoding for the standard input. The StandardInputEncoding property must be set before the process is started. Setting this property does not guarantee that the process will use the specified encoding. The application should be tested to determine which encodings the process supports.

This leads me to believe that when you're dealing with Standard Input, you'll have to compensate for the preamble no matter what.

At least my latest iteration of the code is reading the encoding set by the stream and detecting it.  I've also written tests to verify my code can handle these variable length preambles.  I've also manually tested that the code properly detects the encoding of the stream on a VM with and without this Region setting enabled.  

I do fear that there's some much simpler solution that I missed though, so please double check all my assumptions!

zivkan (dismissed):

zivkan (approved):

PR #5011, opened 2023-01-18
Disable floating versions in project level for CPM 

### LINKED ISSUE #12229: [Bug]: Disable the Update/Install button for floating versions in CPM in PMUI
CPM doesn't allow the usage of Floating Versions, disable this in the PM UI

martinrrm:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12229

## Description

Floating versions is not enabled for CPM, this change disables this functionality of the combobox.

### Project Level
image

### Solution Level
image

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A  

- **Documentation**
  
  - **OR**
  - [x] N/A


martinrrm:
After creating the PR, I realized that I should add tests for `IsCentralPackageManagementEnabled` property.

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageDetailControlModel.cs
```diff
@@ -339,6 +339,14 @@ public bool IsProjectPackageReference
             }
         }
 
+        public bool IsEditableComboboxEnabled
+        {
+            get
+            {
+                return IsProjectPackageReference && !IsCentralPackageManagementEnabled;
```
erdembayar, line 346: If I understand correctly, we enable editable combobox if CPM enabled.
Here it looks like opposite is happening. Please correct me if I'm wrong.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
```diff
@@ -567,6 +572,19 @@ private async Task PackageSourcesChangedAsync(IReadOnlyCollection<PackageSourceC
...
 
+        private async Task IsCentralPackageManagementEnabledAsync(CancellationToken cancellationToken)
+        {
+            await NuGetUIThreadHelper.JoinableTaskFactory.RunAsync(async delegate
+            {
+                // Go off the UI thread to perform non-UI operations
+                await TaskScheduler.Default;
+                foreach (IProjectContextInfo project in Model.Context.Projects)
+                {
+                    _detailModel.IsCentralPackageManagementEnabled = await project.IsCentralPackageManagementEnabledAsync(Model.Context.ServiceBroker, cancellationToken);
```
erdembayar: If there are more than 1 projects, then `_detailModel.IsCentralPackageManagementEnable` going to have value from last project evaluation. 
Is it correct logic? Should it be aggregate of all projects or just last one?

kartheekp-ms: Maybe we should terminate the loop when `IsCentralPackageManagementEnabledAsync` method returns for any project.

kartheekp-ms:
> disables the ability of installing packages in project level since CPM is a Solution level functionality.

I think in `dotnet add package` command we add package reference to the project file and package version to the `Directory.Packages.Props`. I think we should have the same experience across CLI and VS. What do others think? cc @jeffkl

donnie-msft:
> > disables the ability of installing packages in project level since CPM is a Solution level functionality.
> 
> I think in `dotnet add package` command we add package reference to the project file and package version to the `Directory.Packages.Props`. I think we should have the same experience across CLI and VS. What do others think? 

@kartheekp-ms  I'm not seeing in docs nor in manual testing that `dotnet add package` supports floating versions. Is it supposed to be supported?

`dotnet add package NuGet.Core --version 2.*`
> **error: Value cannot be null. (Parameter 'version')**

..but an explicit version does work...

`dotnet add package NuGet.Core --version 2.6.0`
> info : PackageReference for package 'NuGet.Core' version '2.6.0' added to file '[...].csproj'.

kartheekp-ms:
> I'm not seeing in docs nor in manual testing that `dotnet add package` supports floating versions. Is it supposed to be supported?

Thank you for taking the time to check if `dotnet add package` supports floating versions right currently. Good that know that using a floating version throws an exception. My comment was related to `disabling the ability to install packages in project level since CPM is a Solution level functionality` statement. IMHO, CLI and VS should have same user experience for non-floating package versions at least.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
martinrrm: You're right @erdembayar, I have some logic before calling this function and this should only be run in Project level, but that makes this function non reusable, I'll modify it to get the aggregate even if it is just one project

martinrrm:
Thanks @donnie-msft for checking the behavior, you're right floating versions are not supported in CPM, this PR disables floating versions in the CPM and talking with Jeff we thought about disabling the button from the Project Level too to avoid confusion in the users. But as @kartheekp-ms says, adding a package using dotnet inside a project works and the behavior should be the same.

I think we should not disable adding/updating a package in the Project Level

donnie-msft:
We can disable floating versions controls in PM UI when CPM is enabled, and that's the point of the linked issue and this PR, right?
If we want to support CPM in the PM UI in the future, that to me is an entirely different topic and would warrant a new issue/PR.

This PR is a bug fix, and I'm not seeing how a new feature request is blocking it. Please schedule time with me if I'm still not on the same page.

martinrrm:
Updated the PR, this no longer disabled the Update/Install button in project level.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
```diff
@@ -567,6 +569,23 @@ private async Task PackageSourcesChangedAsync(IReadOnlyCollection<PackageSourceC
...
         }
 
+        private async Task IsCentralPackageManagementEnabledAsync(CancellationToken cancellationToken)
+        {
+            await NuGetUIThreadHelper.JoinableTaskFactory.RunAsync(async delegate
+            {
+                // Go off the UI thread to perform non-UI operations
+                await TaskScheduler.Default;
+
+                if (!Model.IsSolution)
```
donnie-msft: An optimization would be to perform this check before invoking this method at all, or at the top of the method. Don't want Solution PMUI switching threads unnecessarily.

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageDetailControlModel.cs
```diff
@@ -339,6 +339,14 @@ public bool IsProjectPackageReference
             }
         }
 
+        public bool IsEditableComboboxEnabled
```
donnie-msft: naming suggestion..

```suggestion
        public bool IsFloatingVersionSupported
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
```diff
@@ -567,6 +569,23 @@ private async Task PackageSourcesChangedAsync(IReadOnlyCollection<PackageSourceC
...
+                {
+                    foreach (IProjectContextInfo project in Model.Context.Projects)
+                    {
```
donnie-msft: I don't understand this loop... so the last project wins? It looks like it's overriding a shared member variable.

martinrrm: There is only 1 project because is not a Solution, I prefer using a foreach here to be clear that Model.Context.Projects is an array but since is not a solution it can only have 1 element (or maybe empty?).

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
donnie-msft: I prefer this approach. What do you think?
https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs#L210

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
donnie-msft: [In other places](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs#LL293C27-L293C58), we're not afraid to make this assertion: `Model.Context.Projects.First();`

I'd say it's more clear than a loop that should always be 1 item. If it ever were more than 1 item, I'd rather an exception occur rather than "last project wins".

donnie-msft: In fact, not sure why we aren't using `Model.Context.Projects.Single();` in both places 🤔

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
```diff
@@ -567,6 +569,22 @@ private async Task PackageSourcesChangedAsync(IReadOnlyCollection<PackageSourceC
             }
         }
 
+        private async Task IsCentralPackageManagementEnabledAsync(CancellationToken cancellationToken)
```
erdembayar, line 572: If I understand correctly then CPM is enabled for all projects or not, so just checking for 1st project would be suffice.

```Suggestion
        private async Task IsCentralPackageManagementEnabledAsync(CancellationToken cancellationToken)
        {
            if (!Model.IsSolution)
            {
                await NuGetUIThreadHelper.JoinableTaskFactory.RunAsync(async delegate
                {
                    // Go off the UI thread to perform non-UI operations
                    await TaskScheduler.Default;
                    if (Model.Context.Projects.Any())
                    {
                        _detailModel.IsCentralPackageManagementEnabled = await Model.Context.Projects.First().IsCentralPackageManagementEnabledAsync(Model.Context.ServiceBroker, cancellationToken);
                    }
                });
            }
        }
```

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetProjectManagerService.cs
```diff
@@ -274,6 +274,30 @@ public async ValueTask<IProjectContextInfo> GetProjectAsync(string projectId, Ca
             return targetFrameworks;
         }
 
+        public async ValueTask<bool> IsCentralPackageManagementEnabledAsync(string projectId, CancellationToken cancellationToken)
```
erdembayar, line 277: This one looks very similar to one in IProjectContextInfoExtensions.cs. Could they considerate into one?

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
donnie-msft: You can skip the `Any()` and just use the `First()` as I suggested here: https://github.com/NuGet/NuGet.Client/pull/5011#discussion_r1103360097

erdembayar: But that case we're making assumption, there would be at least 1 project, `Any` is for safeguarding any unexpected null exception error.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetProjectManagerService.cs
martinrrm: I looked at the code but don't see a clear way to make it simpler, do you see something here?

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
martinrrm: I'll go with `First()` because as donnie said, we use it on other places and we're not afraid to make that assertion. I would normally go with `Any()` to avoid errors but the PM UI cannot be opened without a project

erdembayar: You still didn't implement `First()`. For you no need to loop. You can use above code without `Any()` check too.

martinrrm: Sorry I forgot to push new changes.

zivkan: > If I understand correctly then CPM is enabled for all projects or not

That's not correct. CPM (and PackageReference) is implemented via MSBuild, and MSBuild is strictly per-project.  Trying to enforce anything to be "all of solution or none" in MSBuild is effectively impossible.

martinrrm:
@donnie-msft @erdembayar Can a get a review again? Thanks!

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
```diff
@@ -818,6 +818,8 @@ public OptionsViewModel Options
 
         public string PackagePath => _searchResultPackage?.PackagePath;
 
+        public bool IsCentralPackageManagementEnabled { get; set; }
```
donnie-msft, line 821: Can we get rid of this property? Unless I missed a XAML binding, it's only used to store the result of IsCentralPackageManagementEnabledAsync. If we can avoid maintaining state, that's usually best. If we do want state, then I'm not sure where we're utilizing it?

martinrrm: I'm using it to know if we should have floating versions enabled in the combobox https://github.com/NuGet/NuGet.Client/pull/5011/files#diff-0b6aa73227964074dc06409994723b4f7d158fe4edc669b5c8229c1c3d235bd4R342

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
erdembayar: @martinrrm 
Have you looked into above scenario? Do we consider it CPM enabled if some projects are CPM enabled but not all? Maybe check with @jeffkl

martinrrm: This code will only run in the project PM UI, so if the user has disabled CPM for that project, it can install any version. I understand that disable CPM in just a project is not a desired behavior for this feature but since its msbuild it difficult to restrain users from doing it. @jeffkl can you confirm this?

donnie-msft (approved):

PR #5016, opened 2023-01-19
Embed NuGet files in the binary log after restore completes

### LINKED ISSUE #12374: NuGet should include the files emitted during restore in the binary log
### NuGet Product(s) Involved

MSBuild.exe, dotnet.exe

### The Elevator Pitch

During a build, it is possible to embed arbitrary files in the binary log.  The .NET SDK by default includes the `project.assets.json` file during build: https://github.com/dotnet/sdk/pull/16840

However, during restore the file does not exist yet so a binlog for just restore can be missing files.  NuGet should include these files after restore is complete.

_No response_

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12374

## Description

This updates the Restore task to embed NuGet files in the binlog after Restore is complete.  Even if restore fails, any files that exist will be embedded:

image

In this case the package name is wrong and the `.dgspec` is still in the binlog.

For static graph-based restore, the output paths are logged back to the RestoreTaskEx so it can embed the files in the binlog at the end of the actual build.  I could not find a way to embed them in the `nuget.binlog` that contains evaluation...

Related to https://github.com/dotnet/sdk/pull/16840

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception - It would be hard to verify the contents of the binlog at the moment.
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
+++ b/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
@@ -140,6 +140,9 @@ Copyright (c) .NET Foundation. All rights reserved.
       Interactive="$(NuGetInteractive)"
       RestoreForceEvaluate="$(RestoreForceEvaluate)"
       RestorePackagesConfig="$(RestorePackagesConfig)"/>
+
+    <!-- When Restore fails, still try to embed the results of restore in the binary log -->
+    <OnError ExecuteTargets="EmbedNuGetFilesInBinLog" />
   </Target>
 
   <!--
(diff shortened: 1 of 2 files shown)


jeffkl:
FYI @KirillOsenkov!

donnie-msft (approved):

erdembayar (approved):
Cool. 👏

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
```diff
@@ -1457,4 +1460,20 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+    EmbedNuGetFilesInBinLog
+    Adds files to be embedded in the binary log.  These files are results of a restore
+    so they need to be added after restore is complete.
+    ============================================================
+  -->
+  <Target Name="EmbedNuGetFilesInBinLog"
+        AfterTargets="Restore"
+        Condition="'$(EmbedNuGetFilesInBinLog)' != 'false'">
+    <ItemGroup>
+      <EmbedInBinlog Include="$(ProjectAssetsFile)" Condition="Exists('$(ProjectAssetsFile)') AND $(EmbedProjectAssetsFile) != false" />
```
KirillOsenkov, line 1474: I would swap the AND branches and check for EmbedProjectAssetsFile first, then check for Exists after AND. I can't remember why, but we broke someone by still hitting the disk for Exists when it was turned off.

See here:
https://github.com/dotnet/sdk/pull/20624/files

KirillOsenkov: Here was the issue where we were reading the file unnecessarily and this broke deterministic/distributed build:
https://github.com/dotnet/sdk/issues/20556

jeffkl:
I'm closing this for now because I'm not 100% happy with it.  Currently, only the `project.assets.json` file of the entry project is captured since it runs after the `Restore` target.  So if you restore 50 projects, you only get one, which I don't think is very useful.  I'll spend some time coming up with a better solution...

PR #5026, opened 2023-01-21
Fix project evaluation information in binlog for static graph-based restore

### LINKED ISSUE #11485: [Bug]: Static graph restore binlog doesn't have evaluation data
### NuGet Product Used

MSBuild.exe

### Product Version

msbuild 17

### Worked before?

_No response_

### Impact

### Repro Steps & Context

```
SET RESTORE_TASK_BINLOG_PARAMETERS=LogFile=%cd%\nuget.binlog
msbuild /t:restore /bl /p:RestoreUseStaticGraphEvaluation=true
```

Inspect `nuget.binlog` and notice the Evaluation node is empty/childless:
image

It should look more like this:
image

### Verbose Logs

_No response_

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Home/issues/11485

## Description

Something changed in MSBuild in the last year or so where the evaluation information for projects weren't showing up in the binlogs for static graph-based restore.  This change alters how projects are loaded and built.

Before:
* ProjectGraph is loaded which transitively loads all projects
* During project load, but before its returned, build the necessary targets for restore
* Return the project

After
* ProjectGraph is loaded which transitively loads all projects
* Return all the projects
* Loop through each project and build the necessary targets for restore

This seems to fix the problem and evaluation information is in the binlog:
image

I was worried about loss of parallelism since the `ProjectGraph` is loaded via multiple threads and is very efficient.  But I finally
realized that builds are run serially.  We only use the one in-proc build node in MSBuild because spawning out-of-proc build
nodes for targets that run so fast adds too much overhead.  Running the builds out-of-proc actually takes longer.  So from
my testing, the loading of projects and building the targets takes the same amount of time.  On a repository with 800 projects, it used to take 9 seconds and it still takes 9 seconds.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added - Existing tests exist for this scenario
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
+++ b/src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
@@ -866,51 +866,63 @@ namespace NuGet.Build.Tasks.Console
         /// <returns>An <see cref="ICollection{ProjectWithInnerNodes}" /> object containing projects and their inner nodes if they are targeting multiple frameworks.</returns>
         private ICollection<ProjectWithInnerNodes> LoadProjects(IEnumerable<ProjectGraphEntryPoint> entryProjects)
         {
-            var loggers = new List<Microsoft.Build.Framework.ILogger>
-            {
-                LoggingQueue
-            };
-
-            // Get user specified parameters for a binary logger
-            string binlogParameters = Environment.GetEnvironmentVariable("RESTORE_TASK_BINLOG_PARAMETERS");
-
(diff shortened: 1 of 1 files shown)


jeffkl:
FYI @zivkan I did this to make you happy 😄

zivkan (approved):
😁

ghost:

erdembayar (approved):

PR #5028, opened 2023-01-24
Add Package Source Mapping Dialog uses native Window theming & will Autofocus on first control

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2034

## Description

1.  I noticed that opening this dialog window didn't automatically focus on the Package ID textbox. Fixed that in this PR.
1.  I discovered that some properties we had configured were causing this dialog to lose default, native border properties from Windows. Specifically, Win11 theming which can include rounded corners.
    - We now get Title bar text for free.
    - The X in the upper right now looks like any Windows close button (and we get it for free).
    - Note: some repos roll their own Window, and instead set properties directly. I decided to keep using the VS `DialogWindow`, which gives us some free. 
 
    - Windows 10: 
image
image

    - Windows 11 (opted-in to rounded corners):
image
image

## PR Checklist

- [ ] PR has a linked issue.
- [ ] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception - manually tested, see screenshots
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Options/AddMappingDialog.xaml
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Options/AddMappingDialog.xaml
@@ -14,7 +14,8 @@
              ResizeMode="NoResize"
              ShowInTaskbar="False"
              Title="{x:Static nuget:Resources.VSOptions_Label_AddPackageNamespace}"
-             WindowStartupLocation="CenterOwner">
+             WindowStartupLocation="CenterOwner"
+             Loaded="AddMappingDialogWindow_Loaded">
   <vsui:DialogWindow.Resources>
     <ResourceDictionary>
       <ResourceDictionary.MergedDictionaries>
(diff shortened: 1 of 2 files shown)


src/NuGet.Clients/NuGet.PackageManagement.UI/Options/AddMappingDialog.xaml
```diff
@@ -11,14 +11,11 @@
              x:Uid="AddMappingDialogWindow"
              Height="348"
              Width="480"
+             ResizeMode="NoResize"
```
zivkan, line 14: Is there a good reason to block resizing? In your screenshot, there's a package source with a very long name. If a customer also has a source with a long name, or there is a large number of sources, I think it would be useful to be able to resize the window.

Or is this "NoResize" needed to fix something unintuitive?

donnie-msft: This one is intuitive. I did consider allowing resize. Most dialogs do not, but some do. I feel that the scrollbar is sufficient, for now, and didn't want to enable new behavior while trying to fix this issue. One concern is that there may be an expectation that window size is remembered, and I believe that would require some extra work to make that happen.

donnie-msft: Thinking about it... maybe now is the time to enable it. If customers want it to remember width, we can address that when it comes up.
Here's the commit to enable it. Maybe I should do a separate PR since it's scope creep? https://github.com/NuGet/NuGet.Client/commit/d8615a5ea6640dcf3ac2a4012244466488d60092
image

donnie-msft: Draft PR waiting on this PR: https://github.com/NuGet/NuGet.Client/pull/5029

kartheekp-ms (approved):

PR #5042, opened 2023-02-07
Embed NuGet files in binlog after restore

### LINKED ISSUE #12374: NuGet should include the files emitted during restore in the binary log

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12374

## Description
This is a follow-up of https://github.com/NuGet/NuGet.Client/pull/5016 but now it includes the files from all projects that took part in restore. 

This updates the Restore task to embed NuGet files in the binlog after Restore is complete.  Even if restore fails, any files that exist will be embedded:

* project.assets.json
* ProjectFile.csproj.g.nuget.props
* ProjectFile.csproj.g.nuget.targets
* ProjectFile.csproj.dgspec.json

image

In this case the package name is wrong and the `.dgspec` is still in the binlog.

For static graph-based restore, the output paths are logged back to the RestoreTaskEx so it can embed the files in the binlog at the end of the actual build.  I could not find a way to embed them in the `nuget.binlog` that contains evaluation...

Related to https://github.com/dotnet/sdk/pull/16840

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception - It would be hard to verify the contents of the binlog at the moment.
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Core/NuGet.Build.Tasks/NuGet.RestoreEx.targets
```diff
@@ -53,5 +60,5 @@ Copyright (c) .NET Foundation. All rights reserved.
         MSBuildStartupDirectory="$(MSBuildStartupDirectory)"
         />
   </Target>
-
+
```
donnie-msft: undo?

src/NuGet.Core/NuGet.Build.Tasks/StaticGraphRestoreTaskBase.cs
```diff
@@ -136,6 +144,18 @@ public override bool Execute()
...
                             }
                         }
+
+                        if (restoredOutputPaths.Any())
+                        {
+                            RestoredProjectOutputPaths = new ITaskItem[restoredOutputPaths.Count];
+
+                            int i = 0;
+
+                            foreach (var item in restoredOutputPaths)
```
donnie-msft: use explicit type

test/NuGet.Core.Tests/NuGet.Build.Tasks.Test/TaskLoggingQueueTests.cs
```diff
@@ -27,7 +29,7 @@ public void TaskLoggingQueue_Process_HandleBadMessage()
 
             var buildEngine = new TestBuildEngine();
 
-            using (var loggingQueue = new TaskLoggingQueue(new TaskLoggingHelper(buildEngine, nameof(TaskLoggingQueueTests))))
+            using (var loggingQueue = new TaskLoggingQueue(new TaskLoggingHelper(buildEngine, nameof(TaskLoggingQueueTests)), new List<string>()))
```
donnie-msft: use named parameter

src/NuGet.Core/NuGet.Build.Tasks/TaskLoggingQueue.cs
```diff
@@ -19,13 +20,18 @@ internal class TaskLoggingQueue : LoggingQueue<string>
...
         private readonly TaskLoggingHelper _log;
 
+        private readonly ICollection<string> _restoredProjectOutputPaths;
+
         /// <summary>
         /// Initializes a new instance of the TaskLoggingHelperQueue class.
         /// </summary>
         /// <param name="taskLoggingHelper">The <see cref="Task" /> to create a logging queue for.</param>
-        public TaskLoggingQueue(TaskLoggingHelper taskLoggingHelper)
+        /// <param name="restoredProjectOutputPaths">Receives the list of output paths for the restored projects.</param>
```
donnie-msft: This is often (always) an empty list? The dequeuing of a task can modify the list, right? Currently sounds like it's supposed to be a complete list of paths.

If it's just always initializing the list, maybe it shouldn't be a constructor parameter?

```suggestion
        /// <param name="restoredProjectOutputPaths">Receives the list of initial output paths (can be empty) for the restored projects. </param>
```

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
```diff
@@ -127,7 +127,25 @@ public async Task<bool> RestoreAsync(string entryProjectFilePath, IDictionary<st
...
+                if (string.Equals(Environment.GetEnvironmentVariable("MSBUILDBINARYLOGGERENABLED"), bool.TrueString, StringComparison.OrdinalIgnoreCase))
+                {
+                    IReadOnlyList<PackageSpec> projects = dependencyGraphSpec.Projects;
+
+                    for (int i = 0; i < projects.Count; i++)
+                    {
+                        LoggingQueue.Enqueue(new ConsoleOutLogMessage
+                        {
+                            Message = projects[i].RestoreMetadata.OutputPath,
+                            MessageType = ConsoleOutLogMessageType.RestoredProjectOutputPath
```
nkolev92: I might've sighed when I saw this :D

src/NuGet.Core/NuGet.Build.Tasks/NuGet.RestoreEx.targets
```diff
@@ -31,8 +31,15 @@ Copyright (c) .NET Foundation. All rights reserved.
...
-        MSBuildStartupDirectory="$(MSBuildStartupDirectory)"
-        />
+        MSBuildStartupDirectory="$(MSBuildStartupDirectory)">
+      <Output TaskParameter="RestoredProjectOutputPaths" ItemName="_RestoredProjectOutputPaths" />
+    </RestoreTaskEx>
+
+    <ItemGroup Condition="@(_RestoredProjectOutputPaths->Count()) > 0">
+      <EmbedInBinlog Include="%(_RestoredProjectOutputPaths.Identity)project.assets.json;
+                              %(_RestoredProjectOutputPaths.Identity)*.nuget.dgspec.json;
+                              %(_RestoredProjectOutputPaths.Identity)*.nuget.g.*" />
```
nkolev92: Aren't these files already included as they are a part of the msbuild evaluation?

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
```diff
@@ -127,7 +127,25 @@ public async Task<bool> RestoreAsync(string entryProjectFilePath, IDictionary<st
...
                     cleanupAssetsForUnsupportedProjects: IsOptionTrue(nameof(RestoreTaskEx.CleanupAssetsForUnsupportedProjects), options),
                     log: MSBuildLogger,
-                    cancellationToken: CancellationToken.None);
+                cancellationToken: CancellationToken.None);
+
+                // If the MSBuildBinaryLoggerEnabled environment variable is set, return the paths to all of the projects used during restore so that the assets file and other NuGet files
+                // can be embedded in the binlog
+                if (string.Equals(Environment.GetEnvironmentVariable("MSBUILDBINARYLOGGERENABLED"), bool.TrueString, StringComparison.OrdinalIgnoreCase))
+                {
+                    IReadOnlyList<PackageSpec> projects = dependencyGraphSpec.Projects;
```
nkolev92: This covers all projects, which may included projects that don't contain nuget targets + packages.config projects etc. 

We can probably handle that better.

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
```diff
@@ -139,7 +139,15 @@ Copyright (c) .NET Foundation. All rights reserved.
       HideWarningsAndErrors="$(HideWarningsAndErrors)"
       Interactive="$(NuGetInteractive)"
       RestoreForceEvaluate="$(RestoreForceEvaluate)"
-      RestorePackagesConfig="$(RestorePackagesConfig)"/>
+      RestorePackagesConfig="$(RestorePackagesConfig)">
+      <Output TaskParameter="RestoredProjectOutputPaths" ItemName="_RestoredProjectOutputPaths" />
```
nkolev92: Can we return the exact files here instead of paths to do globbing for? We know the exact paths from within the task + we can account for packages.config projects.

nkolev92 (commented):
Looks pretty good!

One small suggestion about handling packages.config projects too.

If we're doing it now, might as well get it done fully :)

src/NuGet.Core/NuGet.Build.Tasks/NuGet.RestoreEx.targets
donnie-msft: I think yes, if they existed when msbuild was invoked. 
I believe this covers the scenario when they're created during restore, so this step injects those newly created files.

nkolev92: Ah, great point!
They might be different and we'd benefit from having both.

zivkan: I think it's just the assets file that's included by the .NET SDK. But I think that when the .NET SDK adds the assets file to the binlog, it attaches the assets file before restore runs, so when the restore is not already up to date, it attaches the wrong data. The assets file is still very useful in a `--no-restore` build, so we probably want the .NET SDK to keep including the assets file on build (unless we add some targets on build to do it, so we can remove the logic from the .NET SDK itself).

nkolev92: > I think it's just the assets file that's included by the .NET SDK

image

It's not the SDK, it's just the simple concept of msbuidl files that are part of the evaluation are included.

aortiz-msft (commented):
Could you please add the files to be included for each case as part of the PR description?

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
```diff
@@ -1065,5 +1070,35 @@ private void LogErrorFromException(Exception exception)
...
+                if (project.RestoreMetadata.ProjectStyle == ProjectStyle.PackageReference)
+                {
+                    LoggingQueue.Enqueue(new ConsoleOutLogEmbedInBinlog(Path.Combine(project.RestoreMetadata.OutputPath, LockFileFormat.AssetsFileName)));
+                    LoggingQueue.Enqueue(new ConsoleOutLogEmbedInBinlog(Path.Combine(project.RestoreMetadata.OutputPath, DependencyGraphSpec.GetDGSpecFileName(Path.GetFileName(project.RestoreMetadata.ProjectPath)))));
+                    LoggingQueue.Enqueue(new ConsoleOutLogEmbedInBinlog(BuildAssetsUtils.GetMSBuildFilePathForPackageReferenceStyleProject(project, BuildAssetsUtils.PropsExtension)));
+                    LoggingQueue.Enqueue(new ConsoleOutLogEmbedInBinlog(BuildAssetsUtils.GetMSBuildFilePathForPackageReferenceStyleProject(project, BuildAssetsUtils.TargetsExtension)));
+                }
+                else if (project.RestoreMetadata.ProjectStyle == ProjectStyle.PackagesConfig)
+                {
+                    LoggingQueue.Enqueue(new ConsoleOutLogEmbedInBinlog(Path.Combine(Path.GetDirectoryName(project.RestoreMetadata.ProjectPath), NuGetConstants.PackageReferenceFile)));
```
zivkan: `NuGetConstants.PackageReferenceFile` is a constant with value `packages.config`. However, projects can also use `packages.<project name>.config`: https://github.com/NuGet/NuGet.Client/blob/a897a1478161a2b5cbf730879977cab3a0b1cef9/src/NuGet.Core/NuGet.Build.Tasks/BuildTasksUtility.cs#L428-L440

src/NuGet.Core/NuGet.Build.Tasks/RestoreTask.cs
```diff
@@ -167,5 +180,39 @@ protected virtual void Dispose(bool disposing)
...
+                if (project.RestoreMetadata.ProjectStyle == ProjectStyle.PackageReference)
+                {
+                    restoredProjectOutputPaths.Add(new TaskItem(Path.Combine(project.RestoreMetadata.OutputPath, LockFileFormat.AssetsFileName)));
+                    restoredProjectOutputPaths.Add(new TaskItem(Path.Combine(project.RestoreMetadata.OutputPath, DependencyGraphSpec.GetDGSpecFileName(Path.GetFileName(project.RestoreMetadata.ProjectPath)))));
+                    restoredProjectOutputPaths.Add(new TaskItem(BuildAssetsUtils.GetMSBuildFilePathForPackageReferenceStyleProject(project, BuildAssetsUtils.PropsExtension)));
+                    restoredProjectOutputPaths.Add(new TaskItem(BuildAssetsUtils.GetMSBuildFilePathForPackageReferenceStyleProject(project, BuildAssetsUtils.TargetsExtension)));
+                }
+                else if (project.RestoreMetadata.ProjectStyle == ProjectStyle.PackagesConfig)
+                {
+                    restoredProjectOutputPaths.Add(new TaskItem(Path.Combine(Path.GetDirectoryName(project.RestoreMetadata.ProjectPath), NuGetConstants.PackageReferenceFile)));
```
zivkan: Like my other comment, this needs to account for `packages.<project name>.config`.

zivkan (commented):
❤️ this feature!

nkolev92 (dismissed):
LGTM. @zivkan had some feedback, but looks great other than that.

nkolev92 (dismissed):

zivkan (dismissed):

test/NuGet.Core.Tests/NuGet.Build.Tasks.Console.Test/BuildTasksUtilityTests.cs
```diff
@@ -12,6 +12,82 @@ namespace NuGet.Build.Tasks.Console.Test
...
+        [Theory]
+        [InlineData("packages.config")]
+        [InlineData("packages.Test.config")]
+        public void GetPackagesConfigFilePath_WithNonExistentPackagesConfig_ReturnsNull(string packagesConfigFilename)
+        {
+            using (var testDirectory = TestDirectory.Create())
+            {
+                string projectFullPath = Path.Combine(testDirectory, "Test.csproj");
+
+                string packagesConfigFilePath = Path.Combine(testDirectory, packagesConfigFilename);
```
donnie-msft: Remove these unused lines

src/NuGet.Core/NuGet.Build.Tasks/BuildTasksUtility.cs
```diff
@@ -478,7 +464,7 @@ private static async Task<RestoreSummary> PerformNuGetV2RestoreAsync(Common.ILog
 
                 settings = settings ?? Settings.LoadSettingsGivenConfigPaths(pcRestoreMetadata.ConfigFilePaths);
 
-                var packagesConfigPath = Path.Combine(Path.GetDirectoryName(pcRestoreMetadata.ProjectPath), NuGetConstants.PackageReferenceFile);
+                var packagesConfigPath = GetPackagesConfigFilePath(pcRestoreMetadata.ProjectPath);
```
donnie-msft: use explicit type

donnie-msft (approved):

zivkan (approved):

PR #5044, opened 2023-02-10
Use int.d17.5 for VSTargetChannelForTests

erdembayar:
Related issue: https://github.com/NuGet/Client.Engineering/issues/2115
Currently we're using `int.d17.4` instead of current `int.d17.6` for our End2End and Apex tests.
This temporary step until real problem is fixed, at least testing `int.d17.5` is better than testing with `int.d17.4`. Currently `int.d17.6` is still failing.


nkolev92:
Does this imply you tried with 17.6 and it's not working?

erdembayar:
> Does this imply you tried with 17.6 and it's not working?

Yes, I already tried it, it's [not working](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7310239&view=results). This one is passing the test, by mistake I requeued the tests 1 more time for this change.

nkolev92 (approved):

PR #5048, opened 2023-02-10
Telemetry for Floating Versions

martinrrm:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/1700

## Description

Adding Telemetry to know if the user is using Floating Versions in the PM UI and understand the rate of usage of Floating Versions in projects.

### Install
image

### Update
image

### Uninstall
image

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A  I wasn't able to Mock the behavior of installing a package through the PM UI

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
@@ -18,6 +18,7 @@ using NuGet.PackageManagement.VisualStudio;
 using NuGet.Packaging.Core;
 using NuGet.ProjectManagement;
 using NuGet.Protocol.Core.Types;
+using NuGet.Versioning;
 using NuGet.VisualStudio;
 using NuGet.VisualStudio.Internal.Contracts;
 using Task = System.Threading.Tasks.Task;
@@ -281,10 +282,10 @@ namespace NuGet.PackageManagement.UI
             }
         }
 
-        private static Tuple<string, string> CreatePackageTuple(IPackageReferenceContextInfo pkg)
+        private static Tuple<string, string, string> CreatePackageTuple(IPackageReferenceContextInfo pkg)
         {
             PackageIdentity package = pkg.Identity;
-            return Tuple.Create(package.Id, package.Version == null ? string.Empty : package.Version.ToNormalizedString());
+            return Tuple.Create(package.Id, package.Version == null ? string.Empty : package.Version.ToNormalizedString(), pkg.AllowedVersions.OriginalString);
         }
 
         private async Task PerformActionImplAsync(
@@ -304,7 +305,7 @@ namespace NuGet.PackageManagement.UI
             var acceptedLicense = true;
 
             List<string> removedPackages = null;
-            var existingPackages = new HashSet<Tuple<string, string>>();
+            var existingPackages = new HashSet<Tuple<string, string, string>>();
             List<Tuple<string, string>> addedPackages = null;
             List<Tuple<string, string>> updatedPackagesOld = null;
             List<Tuple<string, string>> updatedPackagesNew = null;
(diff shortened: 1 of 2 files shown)


src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
```diff
@@ -281,10 +282,10 @@ private static string GenerateUpgradeReport(string projectName, string backupPat
             }
         }
 
-        private static Tuple<string, string> CreatePackageTuple(IPackageReferenceContextInfo pkg)
+        private static Tuple<string, string, string> CreatePackageTuple(IPackageReferenceContextInfo pkg)
         {
             PackageIdentity package = pkg.Identity;
-            return Tuple.Create(package.Id, package.Version == null ? string.Empty : package.Version.ToNormalizedString());
+            return Tuple.Create(package.Id, package.Version == null ? string.Empty : package.Version.ToNormalizedString(), pkg.AllowedVersions.OriginalString);
```
nkolev92: Can AlllowedVersions ever be null? Even when we're not using special versions but the dropdown ones? 

What about original string? Is it ever nulL? I think we fixed it so that it's not, but thought I'dm double check :)

nkolev92: Are we sending this data for packages.config projects too?

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
```diff
@@ -665,7 +679,21 @@ static List<TelemetryEvent> ToTelemetryPackageList(List<Tuple<string, string>> p
             // two collections for updated packages: pre and post upgrade
             if (updatedPackagesNew?.Count > 0)
             {
-                actionTelemetryEvent.ComplexData["UpdatedPackagesNew"] = ToTelemetryPackageList(updatedPackagesNew);
+                var packages = new List<TelemetryEvent>();
```
nkolev92, line 682: Any chance we can add some tests for these?

These are static methods so they should be testable.

nkolev92 (commented):
Looks good.
A few questions about testing. 
I think lots of these methods are static so can be tested.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
martinrrm: Added tests for this functions.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
martinrrm: AllowedVersions and OriginalString can be nulls, I added a null check for that

martinrrm: Yes we send the same information for packages.config and for that case versionRange is going to be the same as version, its like installing a non floating version. Should we avoid sending that telemetry when the project is not PR?

nkolev92: As long as we use/interpret the data in a correct way, then it probably doesn't matter. 
Ideally we have the project type in the same event.

martinrrm: I was wrong, this function will be used for existingPackages and for packages.config projects we dont have AllowedVersions property, so versionRange will not be send in the Telemetry. Also, we dont have the project type in the same event

image

nkolev92: I think the project type would be helpful in either case. 


> for packages.config projects we dont have AllowedVersions property, so versionRange will not be send in the Telemetry

The original concept actually comes form packages.config

Test out:
```
<packages>
  <package id="Newtonsoft.Json" version="9.0.1" targetFramework="net45" allowedVersions="[6.0.0, 10.0.0)" />
</packages>
```

martinrrm: You're right, it will be added to telemetry as `{"version":"8.0.2","versionRange":"[6.0.0, 10.0.0)","id":"newtonsoft.json"}`

Should I try to add the projectType in this PR or follow up in a new one?

nkolev92: Whatever you prefer. I don't mind having different PRs.

nkolev92: Just let us know what you picked so that we know when we should review again :)

martinrrm: Created bug to follow up https://github.com/NuGet/Client.Engineering/issues/2166

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
```diff
@@ -561,6 +562,25 @@ private static string GenerateUpgradeReport(string projectName, string backupPat
                 }
             }, cancellationToken);
         }
+        internal static TelemetryEvent ToTelemetryPackage(string packageId, string packageVersion, string packageVersionRange)
```
nkolev92, line 566: format: You're missing an extra line. Should have been flagged by the analyzer.

nkolev92 (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
martinrrm: Fixed!

nkolev92 (approved):

jeffkl (approved):

PR #5051, opened 2023-02-14
Fix an NRE that occurs when versions selection is changed

### LINKED ISSUE #12184: [Bug]: Versions_SelectionChanged throws NullReferenceException when changing selected package
### NuGet Product Used

### Product Version

17.5 Preview 1.0 33023.41.main

### Worked before?

_No response_

### Impact

Other

### Repro Steps & Context

I don't see any direct customer impact from this, but when changing selected version, the debugger shows exceptions being thrown from `Versions_SelectionChanged` due to `null` value in the AddedItem[0] element (see https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs#L279)

1. Attach a debugger.
2. Launch Project PM UI.
3. Change selected packages.
4. Observe the NRE.

image

### Verbose Logs

_No response_

jebriede:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12184

## Description
An NRE occurs when the package version selection is changed, and the added item array contains a null item as its first element. A previous attempt to fix this in https://github.com/NuGet/NuGet.Client/pull/4935 still attempted to dereference the element to compare it to null. This fix uses a null safe comparison and short circuits if the item is null, setting the PreviousSelectedVersion to an empty string.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
Code is in the UI code behind. Manually tested by attaching a debugger, turning on exceptions, setting a breakpoint, and validating that when AddedItems[0] is null, an NRE is not thrown.
  - [x] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
@@ -276,7 +276,7 @@ namespace NuGet.PackageManagement.UI
 
         private void Versions_SelectionChanged(object sender, SelectionChangedEventArgs e)
         {
-            PackageDetailControlModel.PreviousSelectedVersion = e.AddedItems.Count > 0 && !e.AddedItems[0].Equals(null) ? e.AddedItems[0].ToString() : string.Empty;
+            PackageDetailControlModel.PreviousSelectedVersion = e.AddedItems.Count > 0 && e.AddedItems[0] != null ? e.AddedItems[0].ToString() : string.Empty;
 
             return;
         }


erdembayar (approved):
👏

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
```diff
@@ -276,7 +276,7 @@ private void InstallButton_Clicked(object sender, RoutedEventArgs e)
 
         private void Versions_SelectionChanged(object sender, SelectionChangedEventArgs e)
         {
-            PackageDetailControlModel.PreviousSelectedVersion = e.AddedItems.Count > 0 && !e.AddedItems[0].Equals(null) ? e.AddedItems[0].ToString() : string.Empty;
+            PackageDetailControlModel.PreviousSelectedVersion = e.AddedItems.Count > 0 && e.AddedItems[0] != null ? e.AddedItems[0].ToString() : string.Empty;
```
erdembayar, line 279: Wow, quite subtle bug.

zivkan, line 279: > Code is in the UI code behind. Manually tested by attaching a debugger, turning on exceptions, setting a breakpoint, and validating that when AddedItems[0] is null, an NRE is not thrown.

Can the method's access modifier be changed to `internal`, and then a unit test written to minimize risk of future regressions?

zivkan (approved):

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
jebriede: > > Code is in the UI code behind. Manually tested by attaching a debugger, turning on exceptions, setting a breakpoint, and validating that when AddedItems[0] is null, an NRE is not thrown.
> 
> Can the method's access modifier be changed to `internal`, and then a unit test written to minimize risk of future regressions?

We chatted offline and are satisfied with the test exception. This scenario was not regressed but rather the first attempted fix continued to cause an NRE which I noticed while debugging. A unit test could have caught the issue but there are several challenges, mainly that WPF components are not unit testable so this logic should live in a separate layer which is out of scope for this change.

PR #5054, opened 2023-02-15
Improvements to the patching of dependencies for the dotnet.exe tests, don't overwrite higher versions of assemblies

nkolev92:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2157

## Description

When the new version of the .NET SDK shipped yesterday, it broke some of our tests as our infrastructure started downloading 7.0.200 instead of 7.0.1xx SDK it used.

The correct behavior is to test against newer versions, so we need fix our tests to work with that. 
This is a multi-fix PR;

- Stop calling dotnet.exe with `--output` on a solution file. This was disallowed. See https://github.com/dotnet/sdk/issues/30624, 
There was a fix in https://github.com/dotnet/sdk/pull/29065 for https://github.com/dotnet/sdk/issues/15607 which errors/warns for specifying output path for solution.

- In the past for our dotnet.exe integration tests, prior to https://github.com/NuGet/NuGet.Client/commit/84d99ce064a5f70628c89af9f30693581e84dc37, we used to patch only the NuGet assemblies. If we needed an extra assembly we'd hardcode it. In the linked change, we started copying the new dependencies automatically. Unfortunately this can cause problems sometimes because the version we'd copy would be older than the version in the .NET SDK and cause problems. 
  - You might wonder why only now?  https://github.com/dotnet/runtime/pull/80158/files  and https://github.com/dotnet/runtime/issues/78875 have the answer. 
  Instead of having a platform specific implementation, for the protecteddata dll, in newer runtimes, ashared implementation of the protecteddata.dll will be called. Unfortunately our patching would overwrite that shared implementation with an older version, which is NS2.0, and not appropriate. 
 
## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/test/TestUtilities/Test.Utility/TestDotnetCLiUtility.cs
+++ b/test/TestUtilities/Test.Utility/TestDotnetCLiUtility.cs
@@ -213,7 +213,7 @@ SDKs found: {string.Join(", ", Directory.EnumerateDirectories(SdkDirSource).Sele
                     {
                         var dependencyTargetPath = Path.Combine(pathToSdkInCli, file.Name);
 
-                        if (file.Name.StartsWith("NuGet"))
+                        if (file.Name.Contains("NuGet"))
                         {
                             file.CopyTo(dependencyTargetPath, overwrite: true);
                         }
--- a/test/TestUtilities/Test.Utility/TestDotnetCLiUtility.cs
+++ b/test/TestUtilities/Test.Utility/TestDotnetCLiUtility.cs
@@ -210,7 +210,10 @@ SDKs found: {string.Join(", ", Directory.EnumerateDirectories(SdkDirSource).Sele
                 {
                     foreach (FileInfo file in frameworkArtifactsFolder.EnumerateFiles($"*{fileExtension}"))
                     {
-                        file.CopyTo(Path.Combine(pathToSdkInCli, file.Name), overwrite: true);
+                        if (file.Name.Contains("NuGet"))
+                        {
+                            file.CopyTo(Path.Combine(pathToSdkInCli, file.Name), overwrite: true);
+                        }
                     }
                 }
 
--- a/test/TestUtilities/Test.Utility/TestDotnetCLiUtility.cs
+++ b/test/TestUtilities/Test.Utility/TestDotnetCLiUtility.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using Newtonsoft.Json;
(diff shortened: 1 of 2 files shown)


test/TestUtilities/Test.Utility/TestDotnetCLiUtility.cs
```diff
@@ -210,7 +211,28 @@ private static void CopyRestoreArtifacts(string artifactsDirectory, string pathT
...
+                        else
+                        {
+                            if (File.Exists(dependencyTargetPath)) // If a dependency exists in the SDK, only copy it if our version is higher than the SDK version.
+                            {
+                                var targetFileVersion = new Version(FileVersionInfo.GetVersionInfo(dependencyTargetPath).FileVersion);
+                                var fileToPatchVersion = new Version(FileVersionInfo.GetVersionInfo(file.FullName).FileVersion);
+                                if (fileToPatchVersion > targetFileVersion)
+                                {
+                                    file.CopyTo(dependencyTargetPath, overwrite: true);
+                                }
```
donnie-msft, line 229: So the `else` isn't needed because, this means we can use whatever is in the SDK without needed to copy to the target path (right?)

donnie-msft (dismissed):

erdembayar (dismissed):

test/TestUtilities/Test.Utility/TestDotnetCLiUtility.cs
```diff
@@ -210,7 +211,28 @@ private static void CopyRestoreArtifacts(string artifactsDirectory, string pathT
                 {
                     foreach (FileInfo file in frameworkArtifactsFolder.EnumerateFiles($"*{fileExtension}"))
                     {
-                        file.CopyTo(Path.Combine(pathToSdkInCli, file.Name), overwrite: true);
+                        var dependencyTargetPath = Path.Combine(pathToSdkInCli, file.Name);
+
+                        if (file.Name.StartsWith("NuGet"))
```
heng-liu: Shall we add "Microsoft.Build.NuGetSdkResolver" at here?

nkolev92: Great point. I had contains originally and then I removed it :D

test/TestUtilities/Test.Utility/TestDotnetCLiUtility.cs
nkolev92: Yes, because the version in the SDK is higher than what we're referencing for compilation.

zivkan (approved):

heng-liu (approved):

PR #5066, opened 2023-02-24
Validate and show info for missing readme files

heng-liu:
## Bug

Fixes:https://github.com/NuGet/Home/issues/12415

## Description
Validate and show information(not a warning or an error) during pack about missing readme file.
1.No matter if packing is successful or not, the message will be shown if there is no readme: 
`The package PackageA.1.0.0 is missing a readme. Go to https://aka.ms/nuget/authoring-best-practices/readme to learn why package readmes are important.`
2.This is a message so it won't break people if they have `TreatWarningsAsErrors` set to true. But it also means people could not suppress the message with putting NUxxx in NoWarn.

With this change, running dotnet pack on a solution of multiple projects, the output is like:
```
C:\PatchedSDK0306\dotnet.exe pack C:\Users\Source\Repos\ClassLibrary1\ClassLibrary1.sln
MSBuild version 17.5.0-preview-23061-01+040e2a90e for .NET
  Restored C:\Users\Source\Repos\ClassLibrary1\ClassLibrary4\ClassLibrary4.csproj (in 717 ms).
  Restored C:\Users\Source\Repos\ClassLibrary1\ClassLibrary3\ClassLibrary3.csproj (in 717 ms).
  Restored C:\Users\Source\Repos\ClassLibrary1\ClassLibrary2\ClassLibrary2.csproj (in 717 ms).
  Restored C:\Users\Source\Repos\ClassLibrary1\ClassLibrary1\ClassLibrary1.csproj (in 717 ms).
  ClassLibrary2 -> C:\Users\Source\Repos\ClassLibrary1\ClassLibrary2\bin\Debug\net7.0\ClassLibrary2.dll
  ClassLibrary3 -> C:\Users\Source\Repos\ClassLibrary1\ClassLibrary3\bin\Debug\net7.0\ClassLibrary3.dll
  ClassLibrary4 -> C:\Users\Source\Repos\ClassLibrary1\ClassLibrary4\bin\Debug\net7.0\ClassLibrary4.dll
  Successfully created package 'C:\Users\Source\Repos\ClassLibrary1\ClassLibrary2\bin\Debug\ClassLibrary2.1.0.0.nupkg'.
  The package ClassLibrary3.1.0.0 is missing a readme. Go to https://aka.ms/nuget/authoring-best-practices/readme to learn why package readmes are important.
  Successfully created package 'C:\Users\Source\Repos\ClassLibrary1\ClassLibrary3\bin\Debug\ClassLibrary3.1.0.0.nupkg'.
  Successfully created package 'C:\Users\Source\Repos\ClassLibrary1\ClassLibrary4\bin\Debug\ClassLibrary4.1.0.0.nupkg'.
  ClassLibrary1 -> C:\Users\Source\Repos\ClassLibrary1\ClassLibrary1\bin\Debug\net7.0\ClassLibrary1.dll
  The package ClassLibrary1.1.0.0 is missing a readme. Go to https://aka.ms/nuget/authoring-best-practices/readme to learn why package readmes are important.
  Successfully created package 'C:\Users\Source\Repos\ClassLibrary1\ClassLibrary1\bin\Debug\ClassLibrary1.1.0.0.nupkg'.
```

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Packaging/PublicAPI/net472/PublicAPI.Unshipped.txt
+++ b/src/NuGet.Core/NuGet.Packaging/PublicAPI/net472/PublicAPI.Unshipped.txt
@@ -1,6 +1,2 @@
 #nullable enable
-NuGet.Packaging.Rules.MissingReadmeRule
-~NuGet.Packaging.Rules.MissingReadmeRule.MessageFormat.get -> string
-~NuGet.Packaging.Rules.MissingReadmeRule.MissingReadmeRule(string messageFormat) -> void
-~NuGet.Packaging.Rules.MissingReadmeRule.Validate(NuGet.Packaging.PackageArchiveReader builder) -> System.Collections.Generic.IEnumerable<NuGet.Common.PackagingLogMessage>
 ~static NuGet.Packaging.Rules.AnalysisResources.MissingReadmeInformation.get -> string
--- a/src/NuGet.Core/NuGet.Packaging/PublicAPI/net472/PublicAPI.Unshipped.txt
+++ b/src/NuGet.Core/NuGet.Packaging/PublicAPI/net472/PublicAPI.Unshipped.txt
@@ -1,3 +1,2 @@
 #nullable enable
 ~static NuGet.Packaging.Rules.AnalysisResources.MissingReadmeInformation.get -> string
-~static NuGet.Packaging.Rules.RuleSet.PackageCreationBestPracticeRuleSet.get -> System.Collections.Generic.IEnumerable<NuGet.Packaging.Rules.IPackageRule>
--- a/src/NuGet.Core/NuGet.Packaging/PublicAPI/net472/PublicAPI.Unshipped.txt
+++ b/src/NuGet.Core/NuGet.Packaging/PublicAPI/net472/PublicAPI.Unshipped.txt
@@ -1,1 +1,6 @@
 #nullable enable
+NuGet.Packaging.Rules.MissingReadmeRule
+~NuGet.Packaging.Rules.MissingReadmeRule.MessageFormat.get -> string
+~NuGet.Packaging.Rules.MissingReadmeRule.MissingReadmeRule(string messageFormat) -> void
+~NuGet.Packaging.Rules.MissingReadmeRule.Validate(NuGet.Packaging.PackageArchiveReader builder) -> System.Collections.Generic.IEnumerable<NuGet.Common.PackagingLogMessage>
+~static NuGet.Packaging.Rules.AnalysisResources.MissingReadmeInformation.get -> string
--- a/src/NuGet.Core/NuGet.Packaging/PublicAPI/net472/PublicAPI.Unshipped.txt
+++ b/src/NuGet.Core/NuGet.Packaging/PublicAPI/net472/PublicAPI.Unshipped.txt
@@ -1,2 +1,3 @@
 #nullable enable
 ~static NuGet.Packaging.Rules.AnalysisResources.MissingReadmeInformation.get -> string
+~static NuGet.Packaging.Rules.RuleSet.PackageCreationBestPracticeRuleSet.get -> System.Collections.Generic.IEnumerable<NuGet.Packaging.Rules.IPackageRule>
--- a/src/NuGet.Core/NuGet.Packaging/PublicAPI/netcoreapp5.0/PublicAPI.Unshipped.txt
+++ b/src/NuGet.Core/NuGet.Packaging/PublicAPI/netcoreapp5.0/PublicAPI.Unshipped.txt
@@ -1,6 +1,2 @@
 #nullable enable
-NuGet.Packaging.Rules.MissingReadmeRule
-~NuGet.Packaging.Rules.MissingReadmeRule.MessageFormat.get -> string
-~NuGet.Packaging.Rules.MissingReadmeRule.MissingReadmeRule(string messageFormat) -> void
-~NuGet.Packaging.Rules.MissingReadmeRule.Validate(NuGet.Packaging.PackageArchiveReader builder) -> System.Collections.Generic.IEnumerable<NuGet.Common.PackagingLogMessage>
 ~static NuGet.Packaging.Rules.AnalysisResources.MissingReadmeInformation.get -> string
--- a/src/NuGet.Core/NuGet.Packaging/PublicAPI/netcoreapp5.0/PublicAPI.Unshipped.txt
+++ b/src/NuGet.Core/NuGet.Packaging/PublicAPI/netcoreapp5.0/PublicAPI.Unshipped.txt
@@ -1,3 +1,2 @@
 #nullable enable
 ~static NuGet.Packaging.Rules.AnalysisResources.MissingReadmeInformation.get -> string
-~static NuGet.Packaging.Rules.RuleSet.PackageCreationBestPracticeRuleSet.get -> System.Collections.Generic.IEnumerable<NuGet.Packaging.Rules.IPackageRule>
--- a/src/NuGet.Core/NuGet.Packaging/PublicAPI/netcoreapp5.0/PublicAPI.Unshipped.txt
+++ b/src/NuGet.Core/NuGet.Packaging/PublicAPI/netcoreapp5.0/PublicAPI.Unshipped.txt
@@ -1,1 +1,6 @@
 #nullable enable
+NuGet.Packaging.Rules.MissingReadmeRule
+~NuGet.Packaging.Rules.MissingReadmeRule.MessageFormat.get -> string
+~NuGet.Packaging.Rules.MissingReadmeRule.MissingReadmeRule(string messageFormat) -> void
+~NuGet.Packaging.Rules.MissingReadmeRule.Validate(NuGet.Packaging.PackageArchiveReader builder) -> System.Collections.Generic.IEnumerable<NuGet.Common.PackagingLogMessage>
+~static NuGet.Packaging.Rules.AnalysisResources.MissingReadmeInformation.get -> string
--- a/src/NuGet.Core/NuGet.Packaging/PublicAPI/netcoreapp5.0/PublicAPI.Unshipped.txt
+++ b/src/NuGet.Core/NuGet.Packaging/PublicAPI/netcoreapp5.0/PublicAPI.Unshipped.txt
@@ -1,2 +1,3 @@
 #nullable enable
 ~static NuGet.Packaging.Rules.AnalysisResources.MissingReadmeInformation.get -> string
+~static NuGet.Packaging.Rules.RuleSet.PackageCreationBestPracticeRuleSet.get -> System.Collections.Generic.IEnumerable<NuGet.Packaging.Rules.IPackageRule>
--- a/src/NuGet.Core/NuGet.Packaging/Rules/AnalysisResources.resx
+++ b/src/NuGet.Core/NuGet.Packaging/Rules/AnalysisResources.resx
@@ -214,7 +214,7 @@
     <value>The transform file '{0}' is outside the 'content' folder and hence will not be transformed during installation of this package. Move it into the 'content' folder.</value>
   </data>
   <data name="MissingReadmeInformation" xml:space="preserve">
-    <value>The created package is missing a readme. Go to aka.ms/nuget/authoring-best-practices#readme to learn why package readmes are important.</value>
+    <value>The created package is missing a readme. Go to https://aka.ms/nuget/authoring-best-practices#readme to learn why package readmes are important.</value>
   </data>
   <data name="NoRefOrLibFolderInPackage" xml:space="preserve">
     <value>This package does not contain a lib/ or ref/ folder, and will therefore be treated as compatible for all frameworks. Since framework specific files were found under the build/ directory for {0}, consider creating the following empty files to correctly narrow the compatibility of the package:
--- a/src/NuGet.Core/NuGet.Packaging/Rules/AnalysisResources.resx
+++ b/src/NuGet.Core/NuGet.Packaging/Rules/AnalysisResources.resx
@@ -213,6 +213,9 @@
   <data name="MisplacedTransformFileWarning" xml:space="preserve">
     <value>The transform file '{0}' is outside the 'content' folder and hence will not be transformed during installation of this package. Move it into the 'content' folder.</value>
   </data>
+  <data name="MissingReadmeInformation" xml:space="preserve">
+    <value>The created package is missing a readme. Go to https://learn.microsoft.com/en-us/nuget/create-packages/package-authoring-best-practices#readme to learn why package readmes are important.</value>
+  </data>
   <data name="NoRefOrLibFolderInPackage" xml:space="preserve">
     <value>This package does not contain a lib/ or ref/ folder, and will therefore be treated as compatible for all frameworks. Since framework specific files were found under the build/ directory for {0}, consider creating the following empty files to correctly narrow the compatibility of the package:
 {1}</value>
--- a/src/NuGet.Core/NuGet.Packaging/Rules/AnalysisResources.resx
+++ b/src/NuGet.Core/NuGet.Packaging/Rules/AnalysisResources.resx
@@ -214,7 +214,8 @@
     <value>The transform file '{0}' is outside the 'content' folder and hence will not be transformed during installation of this package. Move it into the 'content' folder.</value>
   </data>
   <data name="MissingReadmeInformation" xml:space="preserve">
-    <value>The created package is missing a readme. Go to https://aka.ms/nuget/authoring-best-practices/readme to learn why package readmes are important.</value>
+    <value>The package {0} is missing a readme. Go to https://aka.ms/nuget/authoring-best-practices/readme to learn why package readmes are important.</value>
+    <comment>0 - package identity</comment>
   </data>
   <data name="NoRefOrLibFolderInPackage" xml:space="preserve">
     <value>This package does not contain a lib/ or ref/ folder, and will therefore be treated as compatible for all frameworks. Since framework specific files were found under the build/ directory for {0}, consider creating the following empty files to correctly narrow the compatibility of the package:
--- a/src/NuGet.Core/NuGet.Packaging/Rules/AnalysisResources.resx
+++ b/src/NuGet.Core/NuGet.Packaging/Rules/AnalysisResources.resx
@@ -214,7 +214,7 @@
     <value>The transform file '{0}' is outside the 'content' folder and hence will not be transformed during installation of this package. Move it into the 'content' folder.</value>
   </data>
   <data name="MissingReadmeInformation" xml:space="preserve">
-    <value>The created package is missing a readme. Go to https://learn.microsoft.com/en-us/nuget/create-packages/package-authoring-best-practices#readme to learn why package readmes are important.</value>
+    <value>The created package is missing a readme. Go to aka.ms/nuget/authoring-best-practices#readme to learn why package readmes are important.</value>
   </data>
   <data name="NoRefOrLibFolderInPackage" xml:space="preserve">
     <value>This package does not contain a lib/ or ref/ folder, and will therefore be treated as compatible for all frameworks. Since framework specific files were found under the build/ directory for {0}, consider creating the following empty files to correctly narrow the compatibility of the package:
--- a/src/NuGet.Core/NuGet.Packaging/Rules/AnalysisResources.resx
+++ b/src/NuGet.Core/NuGet.Packaging/Rules/AnalysisResources.resx
@@ -214,7 +214,7 @@
     <value>The transform file '{0}' is outside the 'content' folder and hence will not be transformed during installation of this package. Move it into the 'content' folder.</value>
   </data>
   <data name="MissingReadmeInformation" xml:space="preserve">
-    <value>The created package is missing a readme. Go to https://aka.ms/nuget/authoring-best-practices#readme to learn why package readmes are important.</value>
+    <value>The created package is missing a readme. Go to https://aka.ms/nuget/authoring-best-practices/readme to learn why package readmes are important.</value>
   </data>
   <data name="NoRefOrLibFolderInPackage" xml:space="preserve">
     <value>This package does not contain a lib/ or ref/ folder, and will therefore be treated as compatible for all frameworks. Since framework specific files were found under the build/ directory for {0}, consider creating the following empty files to correctly narrow the compatibility of the package:
--- a/src/NuGet.Core/NuGet.Packaging/Rules/BestPractice/MissingReadmeRule.cs
+++ b/src/NuGet.Core/NuGet.Packaging/Rules/BestPractice/MissingReadmeRule.cs
@@ -29,7 +29,7 @@ namespace NuGet.Packaging.Rules
             if (readmeMetadata == null)
             {
                 yield return PackagingLogMessage.CreateMessage(
-                    string.Format(CultureInfo.CurrentCulture, Environment.NewLine + MessageFormat),
+                    string.Format(CultureInfo.CurrentCulture, MessageFormat),
                     LogLevel.Minimal);
             }
         }
--- a/src/NuGet.Core/NuGet.Packaging/Rules/BestPractice/MissingReadmeRule.cs
+++ b/src/NuGet.Core/NuGet.Packaging/Rules/BestPractice/MissingReadmeRule.cs
@@ -26,10 +26,12 @@ namespace NuGet.Packaging.Rules
             var nuspecReader = builder?.NuspecReader;
             var readmeMetadata = nuspecReader?.GetReadme();
 
+            var packageIdentity = nuspecReader?.GetIdentity().ToString();
+
             if (readmeMetadata == null)
             {
                 yield return PackagingLogMessage.CreateMessage(
-                    string.Format(CultureInfo.CurrentCulture, MessageFormat),
+                    string.Format(CultureInfo.CurrentCulture, MessageFormat, packageIdentity),
                     LogLevel.Minimal);
             }
         }
--- a/src/NuGet.Core/NuGet.Packaging/Rules/BestPractice/MissingReadmeRule.cs
+++ b/src/NuGet.Core/NuGet.Packaging/Rules/BestPractice/MissingReadmeRule.cs
@@ -12,7 +12,7 @@ using NuGet.Common;
 
 namespace NuGet.Packaging.Rules
 {
-    public class MissingReadmeRule : IPackageRule
+    internal class MissingReadmeRule : IPackageRule
     {
         public string MessageFormat { get; }
 
--- a/src/NuGet.Core/NuGet.Packaging/Rules/MissingReadmeRule.cs
+++ b/src/NuGet.Core/NuGet.Packaging/Rules/MissingReadmeRule.cs
@@ -1,0 +1,37 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using NuGet.Common;
+
+
+namespace NuGet.Packaging.Rules
+{
+    public class MissingReadmeRule : IPackageRule
+    {
+        public string MessageFormat { get; }
+
+        public MissingReadmeRule(string messageFormat)
+        {
+            MessageFormat = messageFormat ?? throw new ArgumentNullException(nameof(messageFormat));
+        }
+
+        public IEnumerable<PackagingLogMessage> Validate(PackageArchiveReader builder)
+        {
+            var nuspecReader = builder?.NuspecReader;
+            var readmeMetadata = nuspecReader?.GetReadme();
+
+            if (readmeMetadata == null)
+            {
+                yield return PackagingLogMessage.CreateMessage(
+                    string.Format(CultureInfo.CurrentCulture, MessageFormat),
+                    LogLevel.Minimal);
+            }
+        }
+    }
+}
--- a/src/NuGet.Core/NuGet.Packaging/Rules/MissingReadmeRule.cs
+++ b/src/NuGet.Core/NuGet.Packaging/Rules/MissingReadmeRule.cs
@@ -29,7 +29,7 @@ namespace NuGet.Packaging.Rules
             if (readmeMetadata == null)
             {
                 yield return PackagingLogMessage.CreateMessage(
-                    string.Format(CultureInfo.CurrentCulture, MessageFormat),
+                    string.Format(CultureInfo.CurrentCulture, Environment.NewLine + MessageFormat),
                     LogLevel.Minimal);
             }
         }
--- a/test/NuGet.Clients.Tests/NuGet.CommandLine.Test/NuGetPackCommandTest.cs
+++ b/test/NuGet.Clients.Tests/NuGet.CommandLine.Test/NuGetPackCommandTest.cs
@@ -6555,41 +6555,6 @@ $@"<package xmlns='http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd'>
         }
 
         [Fact]
-        public void PackCommand_PackReadme_MissingReadme_Succeeds()
-        {
-            var nuspec = NuspecBuilder.Create();
-            var testDirBuilder = TestDirectoryBuilder.Create();
-            var s = Path.DirectorySeparatorChar;
-
-            nuspec
-                .WithPackageId("test")
-                .WithFile("dummy.txt");
-
-            testDirBuilder
-                .WithFile("dummy.txt", 6)
-                .WithNuspec(nuspec);
-
-            using (testDirBuilder.Build())
-            {
-                var nupkgPath = Path.Combine(testDirBuilder.BaseDir, $"{testDirBuilder.NuspecBuilder.PackageIdEntry}.{testDirBuilder.NuspecBuilder.PackageVersionEntry}.nupkg");
-
-                // Act
-                var result = CommandRunner.Run(
-                    Util.GetNuGetExePath(),
-                    testDirBuilder.BaseDir,
-                    $"pack {testDirBuilder.NuspecPath}",
-                    waitForExit: true);
-
-                // Assert
-                Assert.Equal(0, result.ExitCode);
-                Assert.True(File.Exists(nupkgPath));
-
-                // MissingReadmeInformation in output.
-                Assert.Contains(AnalysisResources.MissingReadmeInformation, result.Output);
-            }
-        }
-
-        [Fact]
         public void PackCommand_PackReadme_EmptyReadmeEntry_Fails()
         {
             var nuspecBuilder = NuspecBuilder.Create();
--- a/test/NuGet.Clients.Tests/NuGet.CommandLine.Test/NuGetPackCommandTest.cs
+++ b/test/NuGet.Clients.Tests/NuGet.CommandLine.Test/NuGetPackCommandTest.cs
@@ -6555,6 +6555,41 @@ $@"<package xmlns='http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd'>
         }
 
         [Fact]
+        public void PackCommand_PackReadme_MissingReadme_Succeeds()
+        {
+            var nuspec = NuspecBuilder.Create();
+            var testDirBuilder = TestDirectoryBuilder.Create();
+            var s = Path.DirectorySeparatorChar;
+
+            nuspec
+                .WithPackageId("test")
+                .WithFile("dummy.txt");
+
+            testDirBuilder
+                .WithFile("dummy.txt", 6)
+                .WithNuspec(nuspec);
+
+            using (testDirBuilder.Build())
+            {
+                var nupkgPath = Path.Combine(testDirBuilder.BaseDir, $"{testDirBuilder.NuspecBuilder.PackageIdEntry}.{testDirBuilder.NuspecBuilder.PackageVersionEntry}.nupkg");
+
+                // Act
+                var result = CommandRunner.Run(
+                    Util.GetNuGetExePath(),
+                    testDirBuilder.BaseDir,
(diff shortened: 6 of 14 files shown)


src/NuGet.Core/NuGet.Packaging/Rules/AnalysisResources.resx
```diff
@@ -213,6 +213,9 @@
   <data name="MisplacedTransformFileWarning" xml:space="preserve">
     <value>The transform file '{0}' is outside the 'content' folder and hence will not be transformed during installation of this package. Move it into the 'content' folder.</value>
   </data>
+  <data name="MissingReadmeInformation" xml:space="preserve">
+    <value>The created package is missing a readme. Go to https://learn.microsoft.com/en-us/nuget/create-packages/package-authoring-best-practices#readme to learn why package readmes are important.</value>
```
zivkan:
```suggestion
    <value>The created package is missing a readme. Go to https://learn.microsoft.com/nuget/create-packages/package-authoring-best-practices#readme to learn why package readmes are important.</value>
```

Should we use an aka.ms link instead, to reduce risk of docs URLs changing in the future? (this includes the header that is used as the anchor used in the deep link)

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/NuGetPackCommandTest.cs
```diff
@@ -6554,6 +6554,41 @@ public void PackCommand_PackReadme_FolderNested_Succeeds(char readmeSeparator)
             TestPackReadmeSuccess(testDirBuilder, $"docs{s}nested{s}readme.md");
         }
 
+        [Fact]
+        public void PackCommand_PackReadme_MissingReadme_Succeeds()
+        {
+            var nuspec = NuspecBuilder.Create();
+            var testDirBuilder = TestDirectoryBuilder.Create();
+            var s = Path.DirectorySeparatorChar;
```
zivkan: I'm not sure if this variable is used, but if it is, it should have a more meaningful name than `s`.

test/NuGet.Core.Tests/NuGet.Packaging.Test/BestPracticeRuleTests.cs
```diff
@@ -0,0 +1,94 @@
...
+using NuGet.Common;
+using NuGet.Packaging.Rules;
+using NuGet.Test.Utility;
+using Xunit;
+
+namespace NuGet.Packaging.Test
+{
+    //Package authoring best practices are an ever evolving set of rules and guidelines.
+    //Those soft recommendations are non-breaking validation message. 
+    public class BestPracticeRuleTests
```
zivkan: I kind of see what you're going for, having a `BestPracticeRuleTests`, but my personal opinion is that I'd rather keep the convention of tests having the same name as the production code they're testing. To me, best practises are a documentation issue, and are "likely" to change over time. We shouldn't need to change tests when best practice documentation changes.

src/NuGet.Core/NuGet.Packaging/Rules/MissingReadmeRule.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        public IEnumerable<PackagingLogMessage> Validate(PackageArchiveReader builder)
+        {
+            var nuspecReader = builder?.NuspecReader;
+            var readmeMetadata = nuspecReader?.GetReadme();
+
+            if (readmeMetadata == null)
+            {
+                yield return PackagingLogMessage.CreateMessage(
+                    string.Format(CultureInfo.CurrentCulture, Environment.NewLine + MessageFormat),
```
zivkan: I don't think warning messages should start with new lines

src/NuGet.Core/NuGet.Packaging/PublicAPI/net472/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,6 @@
 #nullable enable
+NuGet.Packaging.Rules.MissingReadmeRule
+~NuGet.Packaging.Rules.MissingReadmeRule.MessageFormat.get -> string
+~NuGet.Packaging.Rules.MissingReadmeRule.MissingReadmeRule(string messageFormat) -> void
+~NuGet.Packaging.Rules.MissingReadmeRule.Validate(NuGet.Packaging.PackageArchiveReader builder) -> System.Collections.Generic.IEnumerable<NuGet.Common.PackagingLogMessage>
+~static NuGet.Packaging.Rules.AnalysisResources.MissingReadmeInformation.get -> string
```
zivkan: Are all rules public? From an API design point of view, I guess it might be useful if someone using the Client SDK wants to run a specific rule on a package, without running all other rules. But it also feels a bit overkill.

However, if we do want to keep it public, I don't think the constructor makes sense. I guess if someone calling the API from their own app wants a translation of the message that we don't provide ourselves it might be useful, but the XML doc for the constructor doesn't say how many parameters (`{x}`) the message takes, or what each of those substitutions mean. If the message is passed into the constructor, I guess maybe it makes sense to provide a property getting to get the value back again. But if I was writing an app using the API, I think only the `Validate(..)` method is useful. So, I think we can simplify the public API if we remove the getter.

src/NuGet.Core/NuGet.Packaging/PublicAPI/netcoreapp5.0/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,6 @@
 #nullable enable
+NuGet.Packaging.Rules.MissingReadmeRule
+~NuGet.Packaging.Rules.MissingReadmeRule.MessageFormat.get -> string
+~NuGet.Packaging.Rules.MissingReadmeRule.MissingReadmeRule(string messageFormat) -> void
+~NuGet.Packaging.Rules.MissingReadmeRule.Validate(NuGet.Packaging.PackageArchiveReader builder) -> System.Collections.Generic.IEnumerable<NuGet.Common.PackagingLogMessage>
+~static NuGet.Packaging.Rules.AnalysisResources.MissingReadmeInformation.get -> string
```
zivkan: I think it would be great if we start adding nullable annotations to our new public APIs, and we'll get the added benefit of having the compiler check nulls in our code as well. It's not always easy to enable nullable checks on old code, but I think it would be great, and reduce bugs, if we start enabling nullable on new files.

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/NuGetPackCommandTest.cs
```diff
@@ -6554,6 +6554,41 @@ public void PackCommand_PackReadme_FolderNested_Succeeds(char readmeSeparator)
...
+            testDirBuilder
+                .WithFile("dummy.txt", 6)
+                .WithNuspec(nuspec);
+
+            using (testDirBuilder.Build())
+            {
+                var nupkgPath = Path.Combine(testDirBuilder.BaseDir, $"{testDirBuilder.NuspecBuilder.PackageIdEntry}.{testDirBuilder.NuspecBuilder.PackageVersionEntry}.nupkg");
+
+                // Act
+                var result = CommandRunner.Run(
```
zivkan: Considering there a "unit" test in `NuGet.Packagaing.Test`, does this test prevent any other regression? Tests like this that run a process, wait for the process the finish, and then assert the outcome as significantly slower than tests that run code in the test host process. Plus, duplicated tests make CI run slower without providing benefit.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/PackCommandTests.cs
```diff
@@ -5513,6 +5520,30 @@ public void PackCommand_PackageReadmeFile_BasicFunc_WithSymbol_Succeeds()
...
+
+            projectBuilder
+                .WithProjectName("test")
+                .WithProperty("IncludeSymbols", "true")
+                .WithProperty("SymbolPackageFormat", "snupkg");
+
+            using (var srcDir = msbuildFixture.Build(testDirBuilder))
+            {
+                projectBuilder.Build(msbuildFixture, srcDir.Path);
+                var result = msbuildFixture.PackProject(projectBuilder.ProjectFolder, projectBuilder.ProjectName, string.Empty, validateSuccess: false);
```
zivkan: Considering there a "unit" test in `NuGet.Packagaing.Test`, does this test prevent any other regression? Tests like this that run a process, wait for the process the finish, and then assert the outcome as significantly slower than tests that run code in the test host process. Plus, duplicated tests make CI run slower without providing benefit.

src/NuGet.Core/NuGet.Packaging/Rules/AnalysisResources.resx
heng-liu: Thanks for reviewing!
I'm not sure what would be the best url here, when people misses adding a readme file into the package.
Hi @JonDouglas @albarry4 , what do you think?

src/NuGet.Core/NuGet.Packaging/PublicAPI/net472/PublicAPI.Unshipped.txt
nkolev92: Another thing is that this newly added rule, contrary to all the other rules, does not really raise a warning. 

Do we need different sets for both rules?

src/NuGet.Core/NuGet.Packaging/Rules/AnalysisResources.resx
nkolev92: I agree with aka.ms idea. 

> aka.ms/nugetauthoringbestpractices

src/NuGet.Core/NuGet.Packaging/Rules/MissingReadmeRule.cs
nkolev92: Originalyl I was wondering if this was raising a warning, but now I'm not clear. 

@heng-liu can we get a more detailed explanation of what we're implementing here? The issues says the first part, but it doesn't go in detail.

test/NuGet.Core.Tests/NuGet.Packaging.Test/BestPracticeRuleTests.cs
nkolev92: For unit tests, I feel like following the same name is super important because of discoverability.

nkolev92 (commented):
I don't really have many concerns with the implementation, but I have questions about the UX and the expectation for the change.

src/NuGet.Core/NuGet.Packaging/Rules/MissingReadmeRule.cs
heng-liu: Thanks for reviewing @zivkan !
I tried without newline at first, the output is as below, and the message seems very difficult to discover.
Without newline:
```
C:\PatchedSDK1\dotnet.exe pack C:\Users\username\Repos\Classlib1\Classlib1.csproj
MSBuild version 17.5.0-preview-23061-01+040e2a90e for .NET
  Restored C:\Users\username\Repos\Classlib1\Classlib1.csproj (in 458 ms).
  Classlib1 -> C:\Users\username\Repos\Classlib1\bin\Debug\net7.0\Classlib1.dll
  The created package is missing a readme. Go to https://learn.microsoft.com/en-us/nuget/create-packages/package-author
  ing-best-practices#readme to learn why package readmes are important.
  Successfully created package 'C:\Users\username\Repos\Classlib1\bin\Debug\Classlib1.1.0.0.nupkg'.
```
With newline:
```
C:\PatchedSDK2\dotnet.exe pack C:\Users\username\Repos\Classlib1\Classlib1.csproj
MSBuild version 17.5.0-preview-23061-01+040e2a90e for .NET
  Restored C:\Users\username\Repos\Classlib1\Classlib1.csproj (in 549 ms).
  Classlib1 -> C:\Users\username\Repos\Classlib1\bin\Debug\net7.0\Classlib1.dll

  The created package is missing a readme. Go to https://learn.microsoft.com/en-us/nuget/create-packages/package-author
  ing-best-practices#readme to learn why package readmes are important.
  Successfully created package 'C:\Users\username\Repos\Classlib1\bin\Debug\Classlib1.1.0.0.nupkg'.
```
Hi @JonDouglas @albarry4 , may I know what do you think? Thanks!

zivkan: @heng-liu that's basically just how MSBuild works. Do we want to "work against" the normal for this one rule, or stay consistent with everything else?

Also try a single project with multiple pack warnings, not just one.  Also try multiple projects in the solution (each of which have multiple warnings) and pack the solution. Basically, when trying to optimise the output, look at as many different scenarios as possible, not just one.

src/NuGet.Core/NuGet.Packaging/Rules/MissingReadmeRule.cs
```diff
@@ -0,0 +1,37 @@
...
+                    LogLevel.Minimal);
```
zivkan: This message doesn't have an NU code. I believe all other pack warnings have NU codes. I know you're putting a URL in the message itself. Was it intentional to be different than the other rules and not have an NU code?

I'm thinking, what happens when the customer opts into "generate package on build", so this warning comes up as part of build. Will it appear in Visual Studio's Error List window?  If so, an NU code is very useful, because it's clickable, taking the customer (hopefuly) directly to the docs page.

src/NuGet.Core/NuGet.Packaging/Rules/AnalysisResources.resx
heng-liu: Thanks! Changed the URL into aka.ms/nuget/authoring-best-practices#readme

src/NuGet.Core/NuGet.Packaging/PublicAPI/net472/PublicAPI.Unshipped.txt
heng-liu: Thanks for the suggestion! @zivkan 
There are many internal rules, and several public rules.
So I changed it into internal instead.

Thanks for reviewing @nkolev92 !
I created a subfolder BestPractice under Rules folder. All rules need to implement the IPackageRule interface, this might be an easy way to differentiate the two sets of rules.

src/NuGet.Core/NuGet.Packaging/Rules/AnalysisResources.resx
```diff
@@ -213,6 +213,9 @@
   <data name="MisplacedTransformFileWarning" xml:space="preserve">
     <value>The transform file '{0}' is outside the 'content' folder and hence will not be transformed during installation of this package. Move it into the 'content' folder.</value>
   </data>
+  <data name="MissingReadmeInformation" xml:space="preserve">
+    <value>The created package is missing a readme. Go to aka.ms/nuget/authoring-best-practices#readme to learn why package readmes are important.</value>
```
zivkan:
```suggestion
    <value>The created package is missing a readme. Go to https://aka.ms/nuget/authoring-best-practices#readme to learn why package readmes are important.</value>
```

Some terminals auto-detect http(s) links and make them clickable. Let's make it as easy as possible for customers to be able to follow the link.

src/NuGet.Core/NuGet.Packaging/Rules/BestPractice/MissingReadmeRule.cs
```diff
@@ -0,0 +1,37 @@
...
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using NuGet.Common;
+
+
+namespace NuGet.Packaging.Rules
```
zivkan, line 13: By default Visual Studio will put new files in a namespace that includes project sub-directories as part of the namespace, now that this file was moved, it's no longer following that convention.  NuGet does this often, but personally I really don't like it. Both directories and namespaces are used for organization, so if it's worthwhile organizing a file, then I think it's also worthwhile organizing the namespace.

Was this done intentionally, or just a mistake since the file was moved, rather than created from the new subdirectory?

src/NuGet.Core/NuGet.Packaging/Rules/AnalysisResources.resx
zivkan: Out of curiosity, did you try putting the `#readme` in the aka.ms url's redirection, rather than in the URL we output to customers?

I'm thinking just in case the markdown document for the docs changes `## README` to something else, the `#readme` is the URL will no longer be valid. If we can make it part of the redirect URL (I have no idea if that works), then we'll be able to maintain it by changing the redirection URL.

test/NuGet.Core.Tests/NuGet.Packaging.Test/BestPracticeRuleTests.cs
zivkan: From https://github.com/NuGet/NuGet.Client/pull/5066#discussion_r1120997940

> I created a subfolder BestPractice under Rules folder.

As I commented above in this conversation, best practises evolve over time as we learn from experience. Do we planning on modifying client every time best practises change? Are we planning on creating pack validations for every best practise that can be programantically checked?

In this specific case, having a readme seems like a good idea for packages on nuget.org, and also on private feeds, so I don't see a problem in "forcing" this rule on everyone (but I feel strongly that it should be possible to NoWarn to hide the message. It would actually be much better if we implemented roslyn analyzer style editorconfig settings, so that customers can change defaults about error, warning, info, none on a code by code basis)

src/NuGet.Core/NuGet.Packaging/Rules/MissingReadmeRule.cs
zivkan: Also, can you please show sample output when running `dotnet pack my.sln`, where there are two projects.  I'm used to MSBuild messages saying which project file (or in the case of C#, which code file) the message is related to.

If my solution has multiple projects that get packed, is the message clear about which project it's referring to?

test/NuGet.Core.Tests/NuGet.Packaging.Test/BestPracticeRuleTests.cs
heng-liu: Thanks! Changed the test class name to `MissingReadmeRuleTests`.

heng-liu: According to our previous discussion, this is going to be a quick first step experiment. 
There are concerns that it takes some time for users to adopt the client tooling to get the latest best practices.
We will evaluate if readme adoption increases significantly after this  implementation, then we can decide what would be the next steps.

src/NuGet.Core/NuGet.Packaging/Rules/MissingReadmeRule.cs
heng-liu: > @heng-liu that's basically just how MSBuild works. Do we want to "work against" the normal for this one rule, or stay consistent with everything else?

Thanks for the explanation! Removed the newline in the message.

heng-liu: > Originalyl I was wondering if this was raising a warning, but now I'm not clear.
> 
> @heng-liu can we get a more detailed explanation of what we're implementing here? The issues says the first part, but it doesn't go in detail.

The logger in pack task is [MSBuildLogger](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Build.Tasks.Pack/PackTask.cs#L80), and it's implementation of Log method is https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Build.Tasks/Common/MSBuildLogger.cs#L53.

The [CreateMessage](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Common/Errors/PackagingLogMessage.cs#L70) doesn't set `logLevel `to `LogLevel.Warning` nor `LogLevel.Error` so it's not a warning nor an error. I set it to `LogLevel.Minimal`, which is the same with the messages [Log_PackageCommandSuccess](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Commands/CommandRunners/PackCommandRunner.cs#L202-L204)
@zivkan also mentioned it needs minimal verbosity as `dotnet pack` defaults to minimal in [this comment](https://github.com/NuGet/Home/issues/12070#issuecomment-1242520006) .
BTW, the loglevel mapping could be found [here](https://github.com/NuGet/Home/blob/dev/implemented/2021/Verify-Verbosity.md#log-level-mapping---the-following-details-are-copied-from-here-thanks-to-joel-verhagen-for-the-detailed-information)

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/NuGetPackCommandTest.cs
heng-liu: Thanks for the comments!
Yes, I feel hesitate when writing the functional tests (NuGet commandline, and dotnet.intergration.tests), for the same reason.
It might be good to let people know how to trade off.
I removed the two functional tests.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/PackCommandTests.cs
heng-liu: Thanks! I removed the two functional tests.

src/NuGet.Core/NuGet.Packaging/Rules/MissingReadmeRule.cs
heng-liu: Thanks! This is not a warning or an error, so there is no NU code.
It's a message similar with [Log_PackageCommandSuccess](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Commands/CommandRunners/PackCommandRunner.cs#L202-L204)

src/NuGet.Core/NuGet.Packaging/Rules/AnalysisResources.resx
heng-liu: Thanks! Fixed.

src/NuGet.Core/NuGet.Packaging/Rules/AnalysisResources.resx
heng-liu: I created a Aka link [nuget/authoring-best-practices/readme](https://aka.ms/nuget/authoring-best-practices/readme) for https://learn.microsoft.com/en-us/nuget/create-packages/package-authoring-best-practices#readme. Changed `#` into `/`, as `#` is not allowed in the Aka link.
Looks like there is a benefit to track the [traffic data](https://microsoft.sharepoint.com/teams/RedirectionServiceInfo/SitePages/Aka-Link-Manager-Tool-FAQ.aspx#traffic-data) for this link. Thanks for the suggestion!

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/NuGetPackCommandTest.cs
heng-liu: This functional test is removed.

src/NuGet.Core/NuGet.Packaging/Rules/MissingReadmeRule.cs
heng-liu: > Also try a single project with multiple pack warnings, not just one. Also try multiple projects in the solution (each of which have multiple warnings) and pack the solution. Basically, when trying to optimise the output, look at as many different scenarios as possible, not just one.

Tried a solution with 4 projects: ClassLibrary1(missing readme), ClassLibrary2, ClassLibrary3(missing readme), ClassLibrary4. 
```
C:\PatchedSDK0301\dotnet.exe pack C:\Users\Source\Repos\ClassLibrary1\ClassLibrary1.sln
MSBuild version 17.5.0-preview-23061-01+040e2a90e for .NET
  Restored C:\Users\Source\Repos\ClassLibrary1\ClassLibrary3\ClassLibrary3.csproj (in 1.7 sec).
  Restored C:\Users\Source\Repos\ClassLibrary1\ClassLibrary1\ClassLibrary1.csproj (in 1.7 sec).
  Restored C:\Users\Source\Repos\ClassLibrary1\ClassLibrary2\ClassLibrary2.csproj (in 1.7 sec).
  Restored C:\Users\Source\Repos\ClassLibrary1\ClassLibrary4\ClassLibrary4.csproj (in 1.7 sec).
  ClassLibrary2 -> C:\Users\Source\Repos\ClassLibrary1\ClassLibrary2\bin\Debug\net7.0\ClassLibrary2.dll
  ClassLibrary1 -> C:\Users\Source\Repos\ClassLibrary1\ClassLibrary1\bin\Debug\net7.0\ClassLibrary1.dll
  ClassLibrary3 -> C:\Users\Source\Repos\ClassLibrary1\ClassLibrary3\bin\Debug\net7.0\ClassLibrary3.dll
  ClassLibrary4 -> C:\Users\Source\Repos\ClassLibrary1\ClassLibrary4\bin\Debug\net7.0\ClassLibrary4.dll
  Successfully created package 'C:\Users\Source\Repos\ClassLibrary1\ClassLibrary2\bin\Debug\ClassLibrary2.1.0.0.nupkg'.
  The created package is missing a readme. Go to https://aka.ms/nuget/authoring-best-practices/readme to learn why package readmes are important.
  Successfully created package 'C:\Users\Source\Repos\ClassLibrary1\ClassLibrary4\bin\Debug\ClassLibrary4.1.0.0.nupkg'.
  Successfully created package 'C:\Users\Source\Repos\ClassLibrary1\ClassLibrary3\bin\Debug\ClassLibrary3.1.0.0.nupkg'.
  The created package is missing a readme. Go to https://aka.ms/nuget/authoring-best-practices/readme to learn why package readmes are important.
  Successfully created package 'C:\Users\Source\Repos\ClassLibrary1\ClassLibrary1\bin\Debug\ClassLibrary1.1.0.0.nupkg'.
```

zivkan: I don't know if other pack warnings/errors have the same problem, but I think this sample output shows a problem, because the output makes it look like either ClassLibrary2 and ClassLibrary3 have issues, or maybe ClassLibrary4 and ClassLibrary1 (it's not clear), but your description says that the messages are related to ClassLibrary1 and ClassLibrary3, but there's no way for us to know that looking at the output.

Please check a different warning type (I don't know, iconURL is deprecated?). If it has the same problem, then please create a new issue and we can ignore it here. Otherwise, I think we need to improve this message's output.

src/NuGet.Core/NuGet.Packaging/Rules/MissingReadmeRule.cs
zivkan: I disagree. The Log_PackageCommandSuccess reports the outcome of the action, whereas this PR's message is an analyzer message displayed when a condition is met. Just like all the other warnings and errors.

Let's say there's a customer who really doesn't want a readme in their package.  How can they suppress this message? If it had an NU code, they could put that NU code in a `NoWarn`.

It's out of scope for this PR, but since pack warnings are really a lot like Roslyn Analyzers, I wish we could have `.editorconfig` settings to allow customers to change [severity levels](https://learn.microsoft.com/en-us/visualstudio/code-quality/use-roslyn-analyzers?view=vs-2022#configure-severity-levels). From a customer experience point of view, it's really really good, whereas MSBuild NoWarn, WarningAsError, WarningNotAsError is much more limited, and the language (MSBuild) to configure it is much more clunky.

heng-liu: As the description in https://github.com/NuGet/Home/issues/12070, `TreatWarningsAsError` isn't an uncommon thing, so warnings in these cases can be disruptive.
So if we add this info as warning, then yes, customers could suppress this message, but some customers might encounter build break for this best practice validation, if they use `TreatWarningsAsError`.

heng-liu:
Thanks for all the reviews! Here are the remaining open questions:
1. Shall we treat the best practice validation results as warnings, or just info?
If we treat them as warnings, then customers could suppress this message. But some customer might encounter build break for this best practice validation, if they use `TreatWarningsAsError`

2. Shall we show best practice validation results only if the package is created successfully? Or show them no matter if the package is created or not?
The message might need to be different:
`The package to be created is missing a readme. Go to https://aka.ms/nuget/authoring-best-practices/readme to learn why package readmes are important.`
` The created package is missing a readme. Go to https://aka.ms/nuget/authoring-best-practices/readme to learn why package readmes are important.`

3. When we run `dotnet pack A.sln` to pack multiple projects, the message will be shown multiple times and it's not clear which package it's about. Shall we add the package path or project path in the message?
Other validation results are warnings so they have 
`C:\PatchedSDK0301\sdk\7.0.200\Sdks\NuGet.Build.Tasks.Pack\build\NuGet.Build.Tasks.Pack.targets(221,5): warning NU5048:
The 'PackageIconUrl'/'iconUrl' element is deprecated. Consider using the 'PackageIcon'/'icon' element instead. Learn mo
re at https://aka.ms/deprecateIconUrl [C:\Users\henli\Source\Repos\ClassLibrary1\ClassLibrary1\ClassLibrary1.csproj]`

nkolev92:
For 2, I think you can tweak the message to make it fit both scenarios> 

> The package, A 1.0.0 is missing a readme. Go to https://aka.ms/nuget/authoring-best-practices/readme to learn why package readmes are important.

Do you see any concerns with that?


For 3, if we're keeping it as a simple message, then I think we need to include the package name.

Another option is if we have more validation is that we can combine them together.

src/NuGet.Core/NuGet.Packaging/Rules/MissingReadmeRule.cs
nkolev92: I'm a fan of severity levels too, but I think the idea behind this was to do a low effort thing to see if it's even worth it.

heng-liu:
> For 2, I think you can tweak the message to make it fit both scenarios>
> 
> > The package, A 1.0.0 is missing a readme. Go to https://aka.ms/nuget/authoring-best-practices/readme to learn why package readmes are important.
> 
> Do you see any concerns with that?

Thanks @nkolev92 !
If someone use `TreatWarningsAsError`, and there is a warning when validating other rules(e.g. Icon url), it will be escalated into an error so the package A 1.0.0 will be disposed afterwards.
If we still show this best practice message, it indicates the package A 1.0.0 is created. So it might be confusing as the package is not created successfully.

So I'm wondering if we show best practice validation results only if the package is created successfully, or show them no matter if the package is created or not?

nkolev92:
> If we still show this best practice message, it indicates the package A 1.0.0 is created. So it might be confusing as the package is not created successfully.

In this case, there'll certainly be another failure that they have to address.

IMO, it's OK if this is extra details are also shown. 

If I'm fixing metadata, I might choose to address more thna just the actual error.

src/NuGet.Core/NuGet.Packaging/Rules/MissingReadmeRule.cs
nkolev92: Thinking about it again, I think the cost of implementing severity levels in pack is pretty low, we'd just need a forward looking design.

heng-liu:
Thanks for your reviews @nkolev92 and @zivkan !
I addressed the open questions and also added into the description of the PR. 
Please let me know if you have different opinions. Thanks!
(I may schedule a short review meeting if that's faster for discussions on different opinions)

src/NuGet.Core/NuGet.Packaging/PublicAPI/net472/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,3 @@
 #nullable enable
+~static NuGet.Packaging.Rules.AnalysisResources.MissingReadmeInformation.get -> string
```
nkolev92, line 2: Can we keep this private for now? 

We're trying to make this experimental change, but by making it part of the API, we're committing too much imo.

test/NuGet.Core.Tests/NuGet.Packaging.Test/MissingReadmeRuleTests.cs
```diff
@@ -0,0 +1,92 @@
...
+readmeMetadata +
+"    <dependencies>" +
+"      <dependency id=\"System.Collections.Immutable\" version=\"4.3.0\" />" +
+"    </dependencies>" +
+"    </metadata>" +
+"</package>";
+
+            var readmeContent = "Test readme file.";
+
+            using (var testDirectory = TestDirectory.Create())
```
nkolev92: You can probably just mock PackageArchiveReader and avoid having to do integration tests.

nkolev92 (commented):
Looks reasonable to me. 
A few small comments.

src/NuGet.Core/NuGet.Packaging/PublicAPI/net472/PublicAPI.Unshipped.txt
heng-liu: I checked our [coding-guideline](https://github.com/NuGet/NuGet.Client/blob/dev/docs/coding-guidelines.md#when-to-use-internals-vs-public-and-when-to-use-internalsvisibleto), but it's not recommended to use `InternalsVisibleTo` in this case.
Shall I just add PackageCreationBestPracticeRules into PackageCreationRules in RuleSet then?

nkolev92: Oh, I didn't notice this was a different assembly from the one where it's being used. 

I'm ambivalent about this. 

Curious what others think.

nkolev92: cc @NuGet/nuget-client

heng-liu: So there are 3 options:
1. Keep the PackageCreationBestPracticeRuleSet.get a public API.
2. Change PackageCreationBestPracticeRuleSet into private, add PackageCreationBestPracticeRuleSet into PackageCreationBestPracticeRules.
3. Change the PackageCreationBestPracticeRuleSet.get into internal, and use ` InternalsVisibleTo`, which violates our [coding-guideline](https://github.com/NuGet/NuGet.Client/blob/dev/docs/coding-guidelines.md#when-to-use-internals-vs-public-and-when-to-use-internalsvisibleto)
`InternalsVisibleTo is used only to allow a unit test to test internal types and members of its runtime assembly. We do not use InternalsVisibleTo between two runtime assemblies.`

If 2 is better, then when people who calls into PackageCreationBestPracticeRuleSet.get  API, they will get an extra item from the collection. Is that acceptable?

nkolev92: I can't imagine too many people are running the validations. 

You're right, 3 is not really an option. 
I don't see a problem with 2.

heng-liu: Thanks! Applied option 2.

test/NuGet.Core.Tests/NuGet.Packaging.Test/MissingReadmeRuleTests.cs
heng-liu: Thanks for the info! Fixed~

heng-liu:
Team review from 3/7: Thanks for @jeffkl and @zivkan 's input! 
We agreed to keep it as a message, as a warning is a breaking change and it should only be considered for major version change.
But we will still need to check if there is a way to suppress the message like this. And it could be in a separate PR.

test/NuGet.Core.Tests/NuGet.Packaging.Test/MissingReadmeRuleTests.cs
```diff
@@ -0,0 +1,81 @@
...
+                    zip.AddEntry("package.nuspec", nuspecContent, Encoding.UTF8);
+                    if (readmeMetadataExists)
+                    {
+                        zip.AddEntry("README.md", readmeContent, Encoding.UTF8);
+                    }
+                }
+
+                // Act
+                var ruleSet = RuleSet.PackageCreationRuleSet;
+                using (var reader = new PackageArchiveReader(stream))
```
nkolev92, line 59: I think my suggestion was to mock the archive reader. Right now, you're still creating a stream, which isn't as bad as writing a file :)

src/NuGet.Core/NuGet.Packaging/PublicAPI/net472/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,2 @@
...
```
nkolev92, line 2: I feel like we need to redesign this. It's pretty poor that we need to make the analysis resource public.

nkolev92 (approved):

jeffkl (approved):

test/NuGet.Core.Tests/NuGet.Packaging.Test/MissingReadmeRuleTests.cs
heng-liu: I see. Thanks! I will address this in quality week in a separate PR (since the CI is flaky now).

oback:
I upgraded my NuGet client from 6.5.0.154 to 6.6.1.2 and noticed this new message. I'm working with some older nuspec files where it's not feasible (right now at least) to change the nuspec file itself. The files are missing a `<PackageReadmeFile>`, but they also use `<PackageIconUrl>` instead of `<PackageIcon>` (NU5048). I've suppressed that warning with `NoWarn`, but have no easy way of ignoring this new message.

It's not a big issue since it's just some clutter in my log, but the inconsistency does confuse me slightly.

To look at it from the other side: If I have set up a project with `TreatWarningsAsErrors` enabled and I don't pin the version of the NuGet binary (or MSBuild or any other part of the toolchain that can generate warnings), I would expect my build to break occasionally (and in this instance, inform me that my nuspec or project is not configured "optimally").

Apologies if this should have been an Issue instead, but there was no discussion on the parent issue.

PR #5070, opened 2023-02-28
Remove bootstrap.proj

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2186

## Description

This removes `bootstrap.proj` and replaces it with more retail MSBuild logic and allows a user to clone our repository and build it as long as they have Visual Studio and .NET SDK installed.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


build.ps1
```diff
@@ -157,10 +157,15 @@ Invoke-BuildStep $VSMessage {
 -ev +BuildErrors
 
 Invoke-BuildStep 'Creating the EndToEnd test package' {
-        param($Configuration)
-        $EndToEndScript = Join-Path $PSScriptRoot scripts\cibuild\CreateEndToEndTestPackage.ps1 -Resolve
-        $OutDir = Join-Path $Artifacts VS15
-        & $EndToEndScript -c $Configuration -out $OutDir
+        $msbuildArgs = "test\TestUtilities\CreateEndToEndTestPackage\CreateEndToEndTestPackage.proj", "/p:Configuration=$Configuration", "/restore:false", "/property:BuildProjectReferences=false"
```
zivkan, line 160: I think that MSBuild (at least outside of MSFT employees) is significantly less well known, and much harder to find help online than powershell.  PowerShell also has PowerShell ISE that allows step by step debugging.

So, I think that converting powershell scripts into MSBuild is a step backwards.

jeffkl: In some cases I agree.  If this Powershell script ran locally and was used by us, Powershell could be the right answer.  I am not advocating that we take every Powershell script we have and replace it with MSBuild logic.

However, in this case the script is doing the following:

1. Assumes `bootstrap.proj` has been restored via `MSBuild /t:restore` or `dotnet restore`
2. Parses `bootstrap.proj` as XML so it knows the version of a package and forms a path to the NuGet package that was restored by `bootstrap.proj`.
3. Gets the output after _building_ `API.Test.csproj` and copies them to a specific location
4. Gets the output after _building_ `GenerateTestPackages.csproj` and copies them to a specific location
5. Copies files from the repository to a specific location

Most of that is relying on `MSBuild` to have restored packages and built projects.  Replacing the logic with an MSBuild project itself with standard `<ProjectReference />` items, a set of items, and a call to a `<Robocopy />` task doesn't seem like that big of a shift.

Also, this script runs on a remote machine most of the time and debugging what Powershell is doing remotely isn't great in my opinion.  The MSBuild binary logger is great for capturing what took place on a different machine.

build/BuildValidator.proj
```diff
@@ -0,0 +1,63 @@
+<Project Sdk="Microsoft.Build.NoTargets">
```
nkolev92, line 1: Can we name this `LocalizationValidator` or something like that/

I know the package is not named super well, but we can do better with the project.

build/loc.proj
```diff
@@ -99,7 +106,4 @@
     <Copy SourceFiles="@(_EnglishBinaries)" DestinationFiles="@(_EnglishBinaries->'%(DestinationPath)')" />
     <Copy SourceFiles="@(_LocalizeFiles)" DestinationFiles="@(_LocalizeFiles->'%(DestinationPath)')" />
   </Target>
-
-  <Target Name="AfterBuild" DependsOnTargets="LocalizeNonProjectFiles;BatchLocalize"/>
-  <Import Project="$(MicroBuildDirectory)Microsoft.VisualStudioEng.MicroBuild.Core.targets" />
```
nkolev92, line 104: How are these targets included now?

setup/After.Microsoft.Common.targets
```diff
@@ -0,0 +1,6 @@
+<Project>
+  <PropertyGroup Condition="'$(MSBuildProjectExtension)' == '.vsmanproj'">
+    <OutputType />
```
nkolev92, line 3: I'm imaging it's getting overwritten if we put it in the project file itself?

nkolev92 (commented):
Love the overall intention. 

Would it be easier to have 1 PR to minimize the risk of the changes? 
All the changes are easy to follow individually, but there's a lot :D 

It would be helpful to summarize the changes in the PR description as well.

Have you verified that the outputs generated by this changes are equivalent before/after?

ghost:

PR #5072, opened 2023-03-01
Account for multiple lines of output when getting values from config.props via MSBuild

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2170
Fixes: https://github.com/NuGet/Client.Engineering/issues/2169

## Description

A change in MSBuild 17.5 causes an extra newline to be logged to the output which is causing out build logic that captures the output with PowerShell to break.  This is because PowerShell by default will store the output in an array if it contains more than one line.  This change tells PowerShell to store the command output as a single string instead so that the rest of the logic works.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/eng/pipelines/templates/Initialize_Build.yml
+++ b/eng/pipelines/templates/Initialize_Build.yml
@@ -27,16 +27,16 @@ steps:
         if ([System.String]::IsNullOrEmpty($env:VsTargetChannelOverride) -eq $false) {
           $targetChannel = $env:VsTargetChannelOverride
         } else {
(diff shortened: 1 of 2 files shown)


nkolev92:
There's already a bug for this: https://github.com/NuGet/Client.Engineering/issues/2169. 

Maybe you can close that too?

nkolev92 (approved):
At some point I compared the versions of MSBuild across a working and a failing build, so there's probably something else that changed too.

martinrrm (approved):

erdembayar (approved):

PR #5084, opened 2023-03-10
Add test case-verify cache file inside obj folder.

CiciLi1:
## Bug 
Fixes: https://github.com/NuGet/Client.Engineering/issues/2193
## Description

1. Add test case-verify cache file inside obj folder after building/rebuilding/cleaning/restoring the test project to "C:\NuGet.Client\test\NuGet.Tests.Apex\NuGet.Tests.Apex\NuGetEndToEndTests\NuGetConsoleTestCase.cs".
2. Move WaitForFileExists() and WaitForFileNotExists() methods from [NuGetUITestCase.cs](https://github.com/NuGet/NuGet.Client/blob/3ab7d1462029e0b211a88040c952c9b8a679708d/test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetUITestCase.cs#L385) to [CommonUtility.cs](https://github.com/NuGet/NuGet.Client/blob/3ab7d1462029e0b211a88040c952c9b8a679708d/test/NuGet.Tests.Apex/NuGet.Tests.Apex/Utility/CommonUtility.cs#L28) for reduced code duplication, as this new case added also uses these two methods.

## PR Checklist

- [X] PR has a linked issue.
- [X] Described changes

- **Tests**
  - [X] Automated tests added

### DIFF
--- a/test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetConsoleTestCase.cs
+++ b/test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetConsoleTestCase.cs
@@ -683,37 +683,29 @@ namespace NuGet.Tests.Apex
         }
 
         [NuGetWpfTheory]
-        [MemberData(nameof(GetPackagesConfigTemplates))]
-        public async Task VerifyCatchFileInsideObjFolder(ProjectTemplate projectTemplate)
+        [MemberData(nameof(GetNetCoreTemplates))]
+        public async Task VerifyCacheFileInsideObjFolder(ProjectTemplate projectTemplate)
         {
             using (var testContext = new ApexTestContext(VisualStudio, projectTemplate, XunitLogger))
             {
                 //Arrange
-                var packageName = "TestPackage";
+                var packageName = "VerifyCacheFilePackage";
                 var packageVersion = "1.0.0";
-                await CommonUtility.CreatePackageInSourceAsync(testContext.PackageSource, packageName, packageVersion);
-                var project = testContext.SolutionService.AddProject(ProjectLanguage.CSharp, ProjectTemplate.NetCoreConsoleApp, "NetCore");
-                var solutionService = VisualStudio.Get<SolutionService>();
+                await CommonUtility.CreatePackageInSourceAsync(testContext.PackageSource, packageName, packageVersion);                
                 var nugetConsole = GetConsole(testContext.Project);
                 nugetConsole.InstallPackageFromPMC(packageName, packageVersion);
-                FileInfo CatchFilePath = CommonUtility.GetCatchFilePath(project.FullPath);
+                FileInfo CacheFilePath = CommonUtility.GetCacheFilePath(testContext.Project.FullPath);
 
                 //Act
                 testContext.SolutionService.Build();
                 testContext.NuGetApexTestService.WaitForAutoRestore();
-                CommonUtility.WaitForFileExists(CatchFilePath);
+                CommonUtility.WaitForFileExists(CacheFilePath);
 
-                project.Rebuild();
-                CommonUtility.WaitForFileExists(CatchFilePath);
+                testContext.Project.Rebuild();
+                CommonUtility.WaitForFileExists(CacheFilePath);
 
-                project.Clean();
-                CommonUtility.WaitForFileNotExists(CatchFilePath);
-
-                nugetConsole.Execute("dotnet restore");
-                CommonUtility.WaitForFileExists(CatchFilePath);
-
-                nugetConsole.Clear();
-                solutionService.Save();
+                testContext.Project.Clean();
+                CommonUtility.WaitForFileNotExists(CacheFilePath);
             }
         }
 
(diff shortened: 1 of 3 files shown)


test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetConsoleTestCase.cs
```diff
@@ -682,6 +682,41 @@ public async Task UpdatePackageForPR_PackageNamespace_WithMultipleFeedsWithIdent
...
+                testContext.NuGetApexTestService.WaitForAutoRestore();
+                CommonUtility.WaitForFileExists(CatchFilePath);
+
+                project.Rebuild();
+                CommonUtility.WaitForFileExists(CatchFilePath);
+
+                project.Clean();
+                CommonUtility.WaitForFileNotExists(CatchFilePath);
+
+                nugetConsole.Execute("dotnet restore");
```
nkolev92: Can we skip this step? `dotnet restore` tests are separate from this. 

Plus, there's no guarantee that dotnet restore and VS restore contain the same code.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetConsoleTestCase.cs
```diff
@@ -682,6 +682,41 @@ public async Task UpdatePackageForPR_PackageNamespace_WithMultipleFeedsWithIdent
...
+                CommonUtility.WaitForFileExists(CatchFilePath);
+
+                nugetConsole.Clear();
```
nkolev92: The last 2 lines are probably unnecessary as Save is happening through Dispose of the ApexTestContext..

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetConsoleTestCase.cs
v-luzh: Hi @nkolev92, we are not expect to test project restoring in this test case. The test purpose of executing `dotnet restore` (as below step5) is to see if the cache file can get back after cleaning solution in previous step. Corresponding manual test case is here:

**Steps:** 
1.Create above project and install corresponding PackageReference package.
2.Build solution, observe there should generate a restore cache file inside obj/ folder.
3.Rebuild solution, observe there should generate a restore cache file inside obj/ folder.
4.Clean solution, observe there should clean the restore cache file inside obj/ folder.
5.Restore NuGet Package, observe there should also generate or persist this restore cache file inside obj/ folder. 
**Expect Result:** 
There should generate a restore cache file inside obj/ folder in step2 & step3 & step5, and no cache file inside obj/ folder in step4. 

Do you mean the `dotnet restore` would not work for above scenario? If it doesn't work, should we use VS restore? (We used `dotnet restore` instead of VS restore since that's easier.) 

If all the restore should be skipped in the test case, we can do that. Have any existed automation test case covered above step5 (verifying cache file generated after doing restore)?

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetConsoleTestCase.cs
v-luzh: OK, we can delete the last 2 lines. Thanks.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetConsoleTestCase.cs
nkolev92: > Hi @nkolev92, we are not expect to test project restoring in this test case. The test purpose of executing dotnet restore (as below step5) is to see if the cache file can get back after cleaning solution in previous step. Corresponding manual test case is here:

Restore is what generates this file. So we're testing restore. By calling dotnet restore, you're testing dotnet restore since it's what's creating the file. 

> Do you mean the dotnet restore would not work for above scenario? If it doesn't work, should we use VS restore? (We used dotnet restore instead of VS restore since that's easier.)

I think it's more appropriate to call Visual Studio restore. 

> If all the restore should be skipped in the test case, we can do that. Have any existed automation test case covered above step5 (verifying cache file generated after doing restore)?

We actually have tests that verify the behavior. https://github.com/NuGet/NuGet.Client/blob/ac7977c16f6d4c7d48ad3c624b6d5caa46c4447f/test/EndToEnd/tests/NetCoreProjectTest.ps1#L141. So maybe this isn't as valuable. 
https://github.com/NuGet/NuGet.Client/blob/ac7977c16f6d4c7d48ad3c624b6d5caa46c4447f/test/EndToEnd/tests/NetCoreProjectTest.ps1#L106-L120 has a test that clean deletes the cache file. 

I think the test here should call VS restore and check that or just remove it as duplicate of the above 2.

v-luzh: OK, we will remove the `dotnet restore` step from this test case. Thanks.

nkolev92 (dismissed):

nkolev92 (approved):

jeffkl (approved):

PR #5085, opened 2023-03-10
Always log an error when static graph-based restore fails

### LINKED ISSUE #12372: Static graph-based restore should always log an error on failure
### NuGet Product(s) Affected

MSBuild.exe, dotnet.exe

### Current Behavior

When static graph-based restore runs, the restore is run in a different process.  The logic in this process has a few spots where something can go wrong and the EXE will not log an error.  In rare cases, no error is logged and restore appears to succeed.

### Desired Behavior

All places in the EXE's code where something can go wrong should log an error and the RestoreTaskEx should consider looking at the exit code in case its non-zero instead of only determining success/failure based on if any errors were logged.

### Additional Context

_No response_

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12372

## Description

This ensures that an error is always logged when:
* Arguments could not be parsed
* An unhandled exception occurs
* A non-zero exit code is returned but an error was not logged

I also brought back the refactoring of having `RestoreTaskEx` and `GenerateRestoreGraphFileTask` share the same base class since I had to revert it with other changes.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


nkolev92 (dismissed):

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
```diff
@@ -651,10 +651,10 @@ private DependencyGraphSpec GetDependencyGraphSpec(string entryProjectPath, IDic
                 // Load the projects via MSBuild and create an array of them since Parallel.ForEach is optimized for arrays
                 var projects = LoadProjects(entryProjects)?.ToArray();
 
-                // If no projects were loaded, return null indicating that the projects could not be loaded.
+                // If no projects were loaded, return an empty DependencyGraphSpec
                 if (projects == null || projects.Length == 0)
                 {
-                    return null;
+                    return new DependencyGraphSpec();
```
nkolev92, line 657: Why do we need tihs change?

jeffkl: This method returns `null` if there's an exception after it logs an error but in the case where you have a solution file with no projects to restore, a message is logged and this returns `null` but that ends up making the EXE return an exit code of `1` and you get an error message that static graph-based restore returned `1` but didn't log an error.  I checked regular restore and it just reports success.  So returning an empty `DependencyGraphSpec` makes more sense because now the caller can tell if it wasn't possible to get one (because of an error) or if there are just no projects to restore.  Now the task logs no error but reports success and there's a message indicating why:

Before:
```
Restore:
  The solution contains '1' project(s) 'a' that are not known to MSBuild. Ensure that all projects are known to be MSBuild before running restore on the solution.
  Evaluated 0 project(s) in 326ms (0 builds, 0 failures).
D:\nuget3\artifacts\NuGet.Build.Tasks.Console\bin\Debug\net472\NuGet.RestoreEx.targets(19,5): error : Static graph-based restore failed with exit code '1' but did not log an error. Please file an issue at https://github
.com/NuGet/Home. [D:\nuget3\.test\work\e3ff513d\04b28e2f\solution\solution.sln]

Build FAILED.

"D:\nuget3\.test\work\e3ff513d\04b28e2f\solution\solution.sln" (Restore target) (1) ->
(Restore target) ->
  D:\nuget3\artifacts\NuGet.Build.Tasks.Console\bin\Debug\net472\NuGet.RestoreEx.targets(19,5): error : Static graph-based restore failed with exit code '1' but did not log an error. Please file an issue at https://gith
ub.com/NuGet/Home. [D:\nuget3\.test\work\e3ff513d\04b28e2f\solution\solution.sln]

    0 Warning(s)
    1 Error(s)

Time Elapsed 00:00:58.55
```

After:
```
Restore:
  The solution contains '1' project(s) 'a' that are not known to MSBuild. Ensure that all projects are known to be MSBuild before running restore on the solution.
  Evaluated 0 project(s) in 196ms (0 builds, 0 failures).
  The solution did not have any projects to restore, ensure that all projects are known to be MSBuild and that the projects exist.

Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:01.66
```

The new behavior matches regular restore now!

nkolev92 (approved):

PR #3206, opened 2020-01-24
Implementation of `dotnet nuget add|remove|list|etc... source`

rrelyea:
## Bug

Fixes: https://github.com/nuget/home/issues/4126
Regression: No

Spec link: https://github.com/NuGet/Home/blob/dev/designs/DotNet-Sources-Support.md

## Fix

Breaking Changes
Move and rename type : NuGet.CommandLine!NuGet.CommandLine.CommandLineException --> NuGet.Commands!NuGet.Commands.CommandException
	(discussed this change with Nikolche…NuGet.CommandLine is not a library. Ok to move types from there…)

NuGet.CommandLine\Commands\SourcesCommand.cs
Use new SourceArgs type, in common with NuGet.CommandLine.Xplat usage too.

NuGetCommandLine.csproj
Changed ILMergeNuGetExe Target to:
	Merge in each languages NuGet.Commands.resources.dll into NuGet.exe so that localization still works.
	It used to be that NuGet.exe strings were all localized in the .resx file directly. (each string had ~13 different languages directly in the string table. NuGet localization process has not been able to deal with that since at least Feb 2016.
	Since we are moving some code and strings from NuGet.CommandLine to NuGet.Commands, we opted to go through normal localization process (build .resources.dll) and then add them to the ILMerged nuget.exe
	
NuGet.CommandLine\NuGetResources.resx
	Delete a bunch of strings… the english ones are moving to NuGet.Commands.dll
	The other language ones are no longer necessary. See ILMerge note above.
	
Commands.xml
	A data format to define verbs/commands args/options.
	(a verb has a noun after it. A command doesn't.)
	This code change only uses verbs. 

CommandParsers.tt/.cs
	Codegen engine to parse commands.xml and create a Parser for each command.
	
Verbs.cs/.tt
	CodeGen engine to create the code for help/command parsing for all the Verbs. Based on data in commands.cs
	
CommandOutputLogger.cs
	On dotnet.exe codepaths, we want to use the logger, but don't like the prefixes outputed in standard scenarios.So, we've made this class support both modes. 
	Using console.writeline was another option considered and discussed with nikolche.
	
NuGet.CommandLine.Xplat\Program.cs
Verbosity wasn't functional before - tracked by dotnet-nuget doesn't support verbosity [#6374](https://github.com/nuget/home/issues/6374)
RegisterCommands() is now done as part of InitializeApp()

NuGet.Commands\CommandArgs\VerbArgs.cs/.tt - via CodeGen
[...]

PR #5088, opened 2023-03-13
Dotnet NuGet Config CLI Config Paths Command

dainizzz:
## Bug

Fixes: https://github.com/NuGet/Home/issues/8420

## Description

Adds dotnet nuget config paths command, which lists all the paths of NuGet configuration files that will be applied, when invoking NuGet command from the current working directory path.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Core/NuGet.Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -0,0 +1,32 @@
...
+                foreach (var filePath in filePaths)
+                {
+                    getLogger().LogMinimal(filePath);
+                }
+            }
+            else
+            {
+                throw new CommandException(string.Format(CultureInfo.CurrentCulture, Strings.Error_PathNotFound, args.WorkingDirectory));
+            }
+
```
donnie-msft: nit: space

src/NuGet.Core/NuGet.Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -0,0 +1,32 @@
...
+        }
+
```
donnie-msft: nit: space

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -0,0 +1,114 @@
...
+
+                // Act
+                var exitCode = Program.MainInternal(args.ToArray(), log);
+
+                // Assert
+                Assert.Equal(string.Empty, log.ShowErrors());
+                Assert.Equal(0, exitCode);
+                Assert.Contains(Path.Combine(testInfo.WorkingPath.Path, "NuGet.Config"), log.Messages);
+            }
+
```
donnie-msft, line 53: nit: space

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -0,0 +1,114 @@
...
+                // Act
+                var exitCode = Program.MainInternal(args.ToArray(), log);
+                var expectedError = @"The path 'C:\Test\NonExistingRepos' doesn't exist.";
+
+                // Assert
+                Assert.Contains(expectedError, log.ShowErrors());
+                Assert.Equal(1, exitCode);
+            }
+
```
donnie-msft, line 127: nit: space

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+
```
donnie-msft: nit: extra space

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -0,0 +1,114 @@
...
+    <packageSources>
+        <add key=""Foo"" value=""https://contoso.com/v3/index.json"" />
+    </packageSources>
+</configuration>
+");
+            }
+
+            public TestDirectory WorkingPath { get; }
+            public string ConfigFile { get; }
+            //public string InvalidWorkingPath { get; }
```
donnie-msft: Remove comment

src/NuGet.Core/NuGet.CommandLine.XPlat/NuGet.CommandLine.XPlat.csproj
```diff
@@ -22,7 +22,7 @@
 
 <!-- Microsoft.Build.Locator is only used when debugging, and the compiler will skip copying this dependency from Release assemblies we insert because we only refer to it conditionally with the DEBUG configuration.
   Uncomment the following when debugging. Also uncomment the MSBuildLocator code from Program.cs -->
-  <!-- <ItemGroup>
+   <!-- <ItemGroup>
```
donnie-msft: Can just revert this file

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Verbs.cs
```diff
@@ -475,4 +475,39 @@ internal partial class UpdateVerbParser
...
+                ConfigCmd.Description = Strings.Config_Description;
+                ConfigCmd.OnExecute(() =>
+                {
+                    app.ShowHelp("config");
+                    return 0;
+                });
+            });
+        }
+    }
+
```
donnie-msft: nit: extra space

donnie-msft (commented):
Looks pretty good! 🎉 Looking forward to a test drive once the build is green!
Sending a few nit-picks in the meantime..

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -0,0 +1,109 @@
...
+using System;
+using System.IO;
+using System.Linq;
+using NuGet.CommandLine.XPlat;
+using NuGet.Test.Utility;
+using Xunit;
+
+namespace NuGet.XPlat.FuncTest
+{
+    [Collection("NuGet XPlat Config Test Collection")]
```
donnie-msft: I'm not seeing a test fixture that corresponds to this collection. Does this do anything, or are you planning to use it in a future commit?

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -0,0 +1,109 @@
...
+                var exitCode = Program.MainInternal(args.ToArray(), log);
+                var expectedError = @"The path 'C:\Test\NonExistingRepos' doesn't exist.";
+
+                // Assert
+                Assert.Contains(expectedError, log.ShowErrors());
+                Assert.Equal(1, exitCode);
+            }
+        }
+
+        internal class TestInfo : IDisposable
```
donnie-msft, line 132: I think this design works in the case where you'll eventually want more test methods which don't need a `TestInfo`, so you can choose whether to create one, and maybe add parameters to it later. 
However, some ideas I had that you could consider:
1. If you're going to have a mix of tests which do and do not need `TestInfo`, maybe best to keep what you have.
1. If you want the exact `TestInfo` in each test (as currently is the case), you could use a constructor/dispose on the test class directly (without a separate class). See https://xunit.net/docs/shared-context.
1. Consider if it makes sense to create 2 (or more) test classes with different contexts (like another variation of your `TestInfo`) for the same source code. It doesn't have to be 1-to-1 for source-to-test class even though a lot of our test projects cram everything into one.
1. FYI, our test architecture varies widely across projects, so don't be hesitant to challenge any patterns we already have.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -0,0 +1,109 @@
...
+            public TestInfo(string configPath)
+            {
+                WorkingPath = TestDirectory.Create();
+                ConfigFile = configPath;
+                CreateFile(WorkingPath.Path,
+                           Path.GetFileName(ConfigFile),
+                           $@"
+<configuration>
+    <packageSources>
+        <add key=""Foo"" value=""https://contoso.com/v3/index.json"" />
```
zivkan:
```suggestion
        <add key=""Foo"" value=""https://contoso.test/v3/index.json"" />
```

`.test` is a reserved top level domain, so we can be confident that this domain won't ever exist. Using `.test` domains for fake examples makes sure that if a mistake is made, we're not sending HTTP requests to a potentially real website.

src/NuGet.Core/NuGet.Commands/PublicAPI/net472/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,7 @@
 #nullable enable
+NuGet.Commands.ConfigPathsArgs
```
nkolev92: Given that this is only going to be implement in `dotnet nuget`, should we move this to the NuGet.CommandLine.Xplat assembly?

cc @NuGet/nuget-client thoughts?

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -0,0 +1,109 @@
...
+{
+    [Collection("NuGet XPlat Config Test Collection")]
+    public class XPlatConfigTests
+    {
+
+        [Fact]
+        public void ConfigPathsCommand_ListConfigPathsWithArgs_Success()
+        {
+            // Arrange
+            using (var testInfo = new TestInfo("NuGet.Config"))
```
nkolev92: style: It may look neater if you use a simple using statement: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0063.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -0,0 +1,109 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
```
nkolev92, line 4: These tests are great, but given how minimal your implementation is, you can probably do most of your testing on the ConfigRunners directly.

nkolev92: Those tests are more efficient.

nkolev92 (commented):
Looks great. 

1 question to the team, and a few stylistic suggestions.

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -812,4 +812,13 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
     <value>The package reference {0} specifies a VersionOverride but the ability to override a centrally defined version is currently disabled.</value>
     <comment>0 - packagereference name</comment>
   </data>
-</root>
+  <data name="ConfigPathsCommandDescription" xml:space="preserve">
+    <value>Lists all the paths of NuGet configuration files that will be applied when invoking NuGet command in a specific directory</value>
```
jeffkl:
```suggestion
    <value>Lists the paths to all NuGet configuration files that will be applied when invoking NuGet command in a specific directory</value>
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -812,4 +812,13 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
...
+  </data>
+  <data name="ConfigPathsWorkingDirectoryDescription" xml:space="preserve">
+    <value>Run this command as if working directory is set to the specified directory. If it's not specified, the current working directory will be used.</value>
```
jeffkl:
```suggestion
    <value>Specifies the directory to start from when listing configuration files. If not specified, the current directory is used.</value>
```

I kind of feel like the term "working directory" is very technical here and I lean towards "starting directory" or something less technical.  Is the term working directory already heavily used in our tooling? If so, I'd hate to be different.

src/NuGet.Core/NuGet.Commands/PublicAPI/net472/PublicAPI.Unshipped.txt
jeffkl: I agree

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
```diff
@@ -336,6 +336,27 @@ public static ISettings GetSettings(string configfile)
             }
         }
 
+        public static ISettings GetSettingsFromDirectory(string directory)
+        {
+            if (Directory.Exists(directory))
```
jeffkl: Wow I thought for sure that `Directory.Exists()` would throw if you pass `null` or `string.Empty` but apparently it does not:

https://dotnetfiddle.net/4VjQxC

I still think this method should check `string.IsNullOrEmpty()` first.

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
```diff
@@ -336,6 +336,27 @@ public static ISettings GetSettings(string configfile)
...
+            {
+                return NuGet.Configuration.Settings.LoadDefaultSettings(directory,
+                   configFileName: null,
+                   machineWideSettings: new XPlatMachineWideSetting());
+            }
+            else if (string.IsNullOrEmpty(directory))
```
jeffkl: since the `if` statement returns something, there should be no need for an `else if` and `else`.

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
```diff
@@ -336,6 +336,27 @@ public static ISettings GetSettings(string configfile)
...
+            {
+                var currentDirectory = Directory.GetCurrentDirectory();
+                return NuGet.Configuration.Settings.LoadDefaultSettings(currentDirectory,
+                   configFileName: null,
+                   machineWideSettings: new XPlatMachineWideSetting());
+            }
+            else
+            {
+                return null;
```
jeffkl: I'm not sure exactly if this method should return `null` or just throw the exception that the caller throws if the specified directory does not exist.  What does everyone think is cleaner?

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
```diff
@@ -336,6 +336,27 @@ public static ISettings GetSettings(string configfile)
             }
         }
 
+        public static ISettings GetSettingsFromDirectory(string directory)
+        {
+            if (Directory.Exists(directory))
+            {
+                return NuGet.Configuration.Settings.LoadDefaultSettings(directory,
+                   configFileName: null,
+                   machineWideSettings: new XPlatMachineWideSetting());
```
jeffkl:
```suggestion
                return NuGet.Configuration.Settings.LoadDefaultSettings(
                   directory,
                   configFileName: null,
                   machineWideSettings: new XPlatMachineWideSetting());
```

Nit: If one arg is on a new line, all args should be.

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
```diff
@@ -336,6 +336,27 @@ public static ISettings GetSettings(string configfile)
...
+            }
+            else if (string.IsNullOrEmpty(directory))
+            {
+                var currentDirectory = Directory.GetCurrentDirectory();
+                return NuGet.Configuration.Settings.LoadDefaultSettings(currentDirectory,
+                   configFileName: null,
+                   machineWideSettings: new XPlatMachineWideSetting());
```
jeffkl:
```suggestion
                return NuGet.Configuration.Settings.LoadDefaultSettings(
                   currentDirectory,
                   configFileName: null,
                   machineWideSettings: new XPlatMachineWideSetting());
```

src/NuGet.Core/NuGet.Commands/Strings.resx
```diff
@@ -1041,4 +1041,8 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
     <value>Warning As Error: {0}</value>
     <comment>{0} is a warning message</comment>
   </data>
+  <data name="Error_PathNotFound" xml:space="preserve">
+    <value>The path '{0}' doesn't exist.</value>
```
jeffkl:
```suggestion
    <value>The specified path '{0}' does not exist.</value>
```
Nit: wording

We could also maybe just throw a [DirectoryNotFoundException](https://learn.microsoft.com/en-us/dotnet/api/system.io.directorynotfoundexception?view=net-8.0) which has wording for this.

src/NuGet.Core/NuGet.Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            var settings = RunnerHelper.GetSettingsFromDirectory(args.WorkingDirectory);
+
+            if (settings != null)
+            {
+                var filePaths = settings.GetConfigFilePaths();
+                foreach (var filePath in filePaths)
+                {
+                    getLogger().LogMinimal(filePath);
+                }
```
jeffkl:
```suggestion
                ILogger logger = getLogger();
                foreach (var filePath in filePaths)
                {
                    logger.LogMinimal(filePath);
                }
```

```suggestion
                    getLogger().LogMinimal(filePath);
```
I'm not sure the precedent in our code, but it seems like it would be more efficient and cleaner to call `getLogger` once and then call `LogMinimal()` multiple times instead of calling `getLogger().LogMinimal()` over and over.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Verbs.cs
```diff
@@ -475,4 +475,38 @@ internal partial class UpdateVerbParser
...
+    {
+        internal static void Register(CommandLineApplication app,
+                                      Func<ILogger> getLogger)
+        {
+            app.Command("config", ConfigCmd =>
+            {
+                ConfigCmd.Command("paths", PathsCmd =>
+                {
+                    CommandArgument workingdirectory = PathsCmd.Argument(
+                        "working-directory", Strings.ConfigPathsWorkingDirectoryDescription);
```
jeffkl:
```suggestion
                    CommandArgument workingdirectory = PathsCmd.Argument(
                        "working-directory",
                        Strings.ConfigPathsWorkingDirectoryDescription);
```

jeffkl (commented):
Looks great overall, mostly stylistic and nit comments!

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
heng-liu: From the [spec](https://github.com/NuGet/Home/blob/dev/proposed/2022/dotnet-nuget-config-spec.md#arguments),
`If the specified WORKING_DIRECTORY doesn't exist, an error is displayed indicating the WORKING_DIRECTORY doesn't exist.`
The [example ](https://github.com/NuGet/Home/blob/dev/proposed/2022/dotnet-nuget-config-spec.md#examples)in the spec:
```
List all the NuGet configuration file that will be applied, but passing a non-existing argument WORKING_DIRECTORY.
dotnet nuget config paths C:\Test\NonExistingRepos

Error: The path "C:\Test\NonExistingRepos" doesn't exist.
```

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
dainizzz: Hi Donnie! I'm going to add more tests to this file but I'm not 100% sure what they'll look like yet, so for now I think it'll be good to keep these tests in this format, and then try implementing the ideas you suggested for future tests.

src/NuGet.Core/NuGet.Commands/PublicAPI/net472/PublicAPI.Unshipped.txt
martinrrm: When we we're looking at other `dotnet nuget` commands, we saw that we were using autogenerated files using `dotnet-t4 global tool` (used in dotnet nuget ... source https://github.com/NuGet/NuGet.Client/pull/3206), so since these are files are autogenerated I suggest not moving them if we plan to continue using the tool. 

These files are a lot of boilerplates and I think if we move this code to `NuGet.CommandLine.Xplat` it would work the same, but I think it's more a discussion of if we should continue using the tool or manually create the files. cc: @NuGet/nuget-client

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
donnie-msft: Suggestion seems in line with the what we have in docs https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-new#options

> name of the current directory is used.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
dainizzz: I removed this for now since I'm not 100% sure if I will use it in the future!

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Verbs.cs
dainizzz: Hi Jeff! At the top of this file, it says:
> Do not manually edit this autogenerated file: instead modify the neighboring .tt file (text template) and/or NuGet.CommandLine.Xplat\Commands\Commands.xml (data file), then re-execute the text template via "run custom tool" on VS context menu for .tt file, or via dotnet-t4 global tool.
Given this, is it OK for me to directly make these changes?

src/NuGet.Core/NuGet.Commands/Strings.resx
dainizzz: The [current design](https://github.com/NuGet/Home/blob/dev/proposed/2022/dotnet-nuget-config-spec.md#examples) says:
```
dotnet nuget config paths C:\Test\NonExistingRepos

Error: The path "C:\Test\NonExistingRepos" doesn't exist.
```
Using a DirectoryNotFoundException would change the design; is that OK since the language is similar?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Verbs.cs
jeffkl: Dang, I guess in a future commit we can update the `.tt` file to use better formatting 😞

src/NuGet.Core/NuGet.Commands/Strings.resx
jeffkl: The `DirectoryNotFoundException` has language like "could not find part of the path ...` which is probably good enough here, I'd just hate for us to localize our own string when there's something already in the CLR.  But still just my suggestion, I'm fine with whatever.

src/NuGet.Core/NuGet.Commands/PublicAPI/net472/PublicAPI.Unshipped.txt
nkolev92: Do we absolutely have to move all the files?

martinrrm: The only one is not autogenerated is `NuGet.Commands.ConfigPathsRunner`, so we can move that one

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
dainizzz: Martin and I discussed this, and throwing the exception instead of returning null made more sense to us.

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
```diff
@@ -336,6 +336,27 @@ public static ISettings GetSettings(string configfile)
...
+            }
+            if (Directory.Exists(directory))
+            {
+                return NuGet.Configuration.Settings.LoadDefaultSettings(
+                   directory,
+                   configFileName: null,
+                   machineWideSettings: new XPlatMachineWideSetting());
+            }
+
+            throw new CommandException(string.Format(CultureInfo.CurrentCulture, Strings.Error_PathNotFound, directory));
```
jeffkl:
```suggestion
            if (string.IsNullOrEmpty(directory))
            {
                directory = Directory.GetCurrentDirectory();
            }
            
            if (!Directory.Exists(directory))
            {
                throw new CommandException(string.Format(CultureInfo.CurrentCulture, Strings.Error_PathNotFound, directory));
            }

            return NuGet.Configuration.Settings.LoadDefaultSettings(
                       directory,
                       configFileName: null,
                       machineWideSettings: new XPlatMachineWideSetting());
```

Sorry now that this method was refactored, I see more improvements.

jeffkl (approved):

src/NuGet.Core/NuGet.Commands/PublicAPI/net472/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,7 @@
...
+NuGet.Commands.ConfigPathsArgs.ConfigPathsArgs() -> void
+NuGet.Commands.ConfigPathsRunner
+~NuGet.Commands.ConfigPathsArgs.WorkingDirectory.get -> string
+~NuGet.Commands.ConfigPathsArgs.WorkingDirectory.set -> void
+~static NuGet.Commands.ConfigPathsRunner.Run(NuGet.Commands.ConfigPathsArgs args, System.Func<NuGet.Common.ILogger> getLogger) -> void
```
zivkan: It would be really great to add `#nullable enable` in the `.cs` file that defines `ConfigPathArgs`, and then annotate nullable/non-nullable types accordingly. It will really help us minimize null reference exception bugs.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -0,0 +1,103 @@
...
+        [Fact]
+        public void ConfigPathsCommand_ListConfigPathsNonExistingDirectory_Fail()
+        {
+            // Arrange & Act
+            using var testInfo = new TestInfo("NuGet.Config");
+            {
+                var result = CommandRunner.Run(
+                    DotnetCli,
+                    Directory.GetCurrentDirectory(),
+                    $"{XplatDll} config paths {@"C:\Test\NonExistingRepos"}",
```
zivkan: You should pass a non-existant path under the directory that `testInfo` created, because if I happen to have `c:\test\NonExistingRepos` on my machine, this test will fail. Plus, `c:\` is not valid on Mac or Linux.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -0,0 +1,103 @@
...
+                    waitForExit: true
+                    );
+                var expectedError = @"The specified path 'C:\Test\NonExistingRepos' does not exist.";
```
jeffkl: Let's have this string come from the resources rather than having it hard-coded in the test.  We recently had a change to a resource and then the unit test was broken because the string was duplicated.  I don't think there's a reason to have these strings duplicated.

src/NuGet.Core/NuGet.Commands/PublicAPI/net472/PublicAPI.Unshipped.txt
nkolev92: Where we going to move `ConfigPathsRunner`?

dainizzz: We ended up moving three files to the Commands folder in `NuGet.CommandLine.XPlat` and are no longer using the tool to autogenerate code.

jeffkl (approved):

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigArgs.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+#nullable enable
+
+namespace NuGet.CommandLine.XPlat
+{
+    internal partial class ConfigPathsArgs
```
nkolev92: why partial?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -0,0 +1,47 @@
...
+
+            var filePaths = settings.GetConfigFilePaths();
+            foreach (var filePath in filePaths)
+            {
+                logger.LogMinimal(filePath);
+            }
+        }
+    }
+
+    internal static class RunnerHelper
```
nkolev92, line 43: you don't need a separate class for 1 static method when you're already in a static class :)

I'd just merge these 2 in 1.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -0,0 +1,47 @@
...
+    {
+        public static ISettings GetSettingsFromDirectory(string directory)
+        {
+            if (string.IsNullOrEmpty(directory))
```
nkolev92: I think a more natural implementation might be to do this check outside of the utility method for getting the settings. 

The name is `GetSettingsFromDirectory`, but it accepts null, which is not obvious from the method signature.

src/NuGet.Core/NuGet.Commands/NuGet.Commands.csproj
```diff
@@ -65,4 +65,8 @@
   <ItemGroup>
     <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
   </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="ConfigCommands\" />
```
nkolev92: This should be reverted.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -0,0 +1,100 @@
...
+
+namespace NuGet.XPlat.FuncTest
+{
+    public class XPlatConfigTests
+    {
+        private static readonly string XplatDll = DotnetCliUtil.GetXplatDll();
+        private static readonly string DotnetCli = TestFileSystemUtility.GetDotnetCli();
+
+        [Fact]
+        public void ConfigPathsCommand_ListConfigPathsWithArgs_Success()
```
nkolev92, line 19: We're missing for dotnet.exe with no parameters.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -0,0 +1,100 @@
...
+using System;
+using System.Globalization;
+using System.IO;
+using NuGet.CommandLine.XPlat;
+using NuGet.Test.Utility;
+using Xunit;
+
+namespace NuGet.XPlat.FuncTest
+{
+    public class XPlatConfigTests
```
nkolev92, line 13: We're missing some negative tests, and tests for things like help command.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.IO;
+using NuGet.Commands;
+using NuGet.Common;
+using NuGet.Configuration;
+
+namespace NuGet.CommandLine.XPlat
+{
+    internal static class ConfigPathsRunner
+    {
+        public static void Run(ConfigPathsArgs args, Func<ILogger> getLogger)
```
nkolev92, line 15: I'd add some null checks for both parameters here.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigCommand.cs
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.CommandLineUtils;
+using NuGet.Common;
+
+namespace NuGet.CommandLine.XPlat
+{
+    internal partial class ConfigCommand
```
kartheekp-ms: Similar question like https://github.com/NuGet/NuGet.Client/pull/5088/files#r1146874447. Why `partial`? 
[dotnet add package](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/AddPackageReferenceCommand.cs) command and others don't have `partial` keyword.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigCommand.cs
```diff
@@ -0,0 +1,45 @@
...
+                        return 0;
+                    });
+                });
+                ConfigCmd.HelpOption("-h|--help");
+                ConfigCmd.Description = Strings.Config_Description;
+                ConfigCmd.OnExecute(() =>
+                {
+                    app.ShowHelp("config");
+                    return 0;
+                });
```
kartheekp-ms, line 41: I think these lines of code are applicable to `config` command block not `paths` command block. What do you think? Please let me know incase if I missed something.

```cs
app.Command("config", ConfigCmd => 
 {
 }
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigCommand.cs
```diff
@@ -0,0 +1,45 @@
...
+        internal static void Register(CommandLineApplication app,
+                                      Func<ILogger> getLogger)
+        {
+            app.Command("config", ConfigCmd =>
+            {
+                ConfigCmd.Command("paths", PathsCmd =>
+                {
+                    CommandArgument workingdirectory = PathsCmd.Argument(
+                        "working-directory",
+                        Strings.ConfigPathsWorkingDirectoryDescription);
```
kartheekp-ms, line 21: If the `working-directory` argument is not passed, then can the command consider the current directory from where the command was invoked as working directory?

Looking at the code below, it looks like current directory is used when working directory is not explicitly passed to the command. In that case, I think working directory should be an `option` instead of `argument`.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -0,0 +1,47 @@
...
+using System;
+using System.Globalization;
+using System.IO;
+using NuGet.Commands;
+using NuGet.Common;
+using NuGet.Configuration;
+
+namespace NuGet.CommandLine.XPlat
+{
+    internal static class ConfigPathsRunner
```
kartheekp-ms, line 13: I think we can add unit tests for the `ConfigPathsRunner` class so that we can only add a smaller number of dotnet integration tests. The advantage is that unit tests doesn't have to be invoked in a separate process where as every dotnet integration tests is executed in a new dotnet.exe process.

kartheekp-ms, line 13: Can this runner be moved to `NuGet.Commands` project? https://github.com/NuGet/NuGet.Client/tree/dev/src/NuGet.Core/NuGet.Commands

nkolev92: I had provided the exact opposite feedback earlier. https://github.com/NuGet/NuGet.Client/pull/5088#discussion_r1137520319
This implementation will only be a part of the .NET CLI, so there's no point in tying it to a public API unnecessarily.

kartheekp-ms: Make sense Nikolche. I am resolving this thread.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigArgs.cs
dainizzz: This code was originally autogenerated using a template, but now that we moved it to `NuGet.CommandLine.XPlat` we're no longer using the template tool to generate it; I'll remove the partial modifier.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigCommand.cs
dainizzz: We were using a template to generate the code before so that's why it had a partial keyword; I'll remove it since we're no longer using the template tool.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigCommand.cs
martinrrm: @kartheekp-ms The approved design in the section [Considerations](https://github.com/NuGet/Home/blob/dev/proposed/2022/dotnet-nuget-config-spec.md#considerations) in point 6 there is an explanation on why `working-directory` was changed from option to argument. 

>WORKING_DIRECTORY is changed from an option to an argument. So we have `dotnet nuget config paths <WORKING_DIRECTORY>` and `dotnet nuget config get <ALL|CONFIG_KEY> <WORKING_DIRECTORY>`. Is it okey if we have two arguments in the second command? And if it's okey, only the second argument could be optional, right?
Confirmed with .NET sdk folks. There is no problem of having two arguments in the command and the second argument could be optional.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
dainizzz: We're trying to follow what other commands look like, so our plan is to add in additional classes for the other `dotnet nuget config` commands to this file. Having `RunnerHelper` be a separate class will be more meaningful once we implement more commands.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigCommand.cs
dainizzz: Lines 35-40 are for just the config command; we're adding tests that will demonstrate this more clearly.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
nkolev92: I guess, it's just hard to see that without more changes :) 

I'll let that be your call (whether you change it now), and I can provide feedback on it when I see the rest of the code. Feel free to resolve the thread once you've decided what you want to do.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigCommand.cs
kartheekp-ms: Thanks for the feedback, Martin. Can we add a `dotnet integration test` to ensure that `dotnet nuget config paths` command returns all the appropriate NuGet.Config paths using the current directory as working directory?

dainizzz: Yes, we added that test! :)

jeffkl (approved):

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -0,0 +1,194 @@
...
+            var args = new ConfigPathsArgs()
+            {
+            };
+
+            // Act & Assert
+            Assert.Throws<ArgumentNullException>(() => ConfigPathsRunner.Run(args, null));
+        }
+
+        [Fact]
+        public void ConfigPathsCommand_TypoInCommand_Fail()
```
jeffkl: I think this test isn't necessary

nkolev92 (approved):

jeffkl (approved):

nkolev92:
This should *probably* wait to be merged until next release. 

I'll share more info offline.

nkolev92 (approved):

martinrrm:
@nkolev92 This is going to merge into a feature branch, do we still need to wait for next release?

nkolev92:
oh nope, my bad, I didn't see that :)

PR #5092, opened 2023-03-14
Process GlobalPackageReference items outside of a target

### LINKED ISSUE #12368: GlobalPackageReference is not working for legacy-style csproj projects
### NuGet Product Used

### Product Version

VS 17.4.4

### Worked before?

_No response_

### Impact

### Repro Steps & Context

1. Create a solution with both a SDK-style csproj file and a legacy-style csproj file (e.g. .NET Framework project).
2. Add a NuGet package to both.
3. Enable Central Package Management per [documentation](https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management#).
4. Note that Central Package Management works fine for both project types.
5. Add a Global Package Reference (per [documentation](https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management#global-package-references)) to the `Directory.Packages.props` file, e.g. [PolySharp](https://www.nuget.org/packages/PolySharp/).
6. Note that the global package reference works in the SDK-style project but is **not** working in the legacy-style project.

`<PackageReference>` and the base feature of Central Package Management both work with legacy-style csproj files, so why doesn't `<GlobalPackageReference>` work too?

### Verbose Logs

_No response_

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12368

## Description

This is a re-do of https://github.com/NuGet/NuGet.Client/pull/5064 which didn't move the logic of creating `<PackageVersion />` items with versions.

## PR Checklist

- [x] PR has a linked issue.
- [ x Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception - We don't have a good way of testing a restore when the targets like `CollectPackageReferences` haven't run like in this scenario.
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
+++ b/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
@@ -92,25 +92,6 @@ Copyright (c) .NET Foundation. All rights reserved.
     </_GenerateRestoreGraphProjectEntryInputProperties>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(ManagePackageVersionsCentrally)' == 'true' And '$(RestoreEnableGlobalPackageReference)' != 'false'">
-    <!--
-        Add GlobalPackageReference items to the PackageReference item group with no version.  The PackageVersion items are added
-        in the CollectCentralPackageVersions target.
-
-        Global package references only include the same assets as a development dependency (runtime; build; native; contentfiles; analyzers)
-        because those kind of packages are the best candidate for a global package reference.  They are generally packages that
-        extend the build.
-
-        Global package references have all assets private because central package references are generally packages that provide
-        versioning, signing, etc and should not flow to downstream dependencies.  Also, central package references are already
-        referenced by every project in the tree so they don't need to be transitive.
-      -->
-    <PackageReference Include="@(GlobalPackageReference)"
-                      Version=""
-                      IncludeAssets="Runtime;Build;Native;contentFiles;Analyzers"
-                      PrivateAssets="All" />
-  </ItemGroup>
-
   <!-- Tasks -->
   <UsingTask TaskName="NuGet.Build.Tasks.RestoreTask" AssemblyFile="$(RestoreTaskAssemblyFile)" />
   <UsingTask TaskName="NuGet.Build.Tasks.WriteRestoreGraphTask" AssemblyFile="$(RestoreTaskAssemblyFile)" />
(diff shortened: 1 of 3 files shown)


nkolev92 (dismissed):

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
```diff
@@ -109,6 +108,12 @@ Copyright (c) .NET Foundation. All rights reserved.
                       Version=""
                       IncludeAssets="Runtime;Build;Native;contentFiles;Analyzers"
                       PrivateAssets="All" />
+
+      <!--
+        Add GlobalPackageReference items to the PackageVersion item group with the version.
+      -->
+    <PackageVersion Include="@(GlobalPackageReference)"
```
nkolev92, line 115: We can probably use a test for this.

jeffkl: I will need to pull `MSBuildLocator` into `NuGet.Build.Tasks.Test` so I can do an MSBuild evaluation without calling targets to verify that the transformation is taking place.

The existing unit tests verify that it works end-to-end but doesn't test in the legacy CSPROJ scenario (ie CollectPackageReferences hasn't run).  What do you think?

nkolev92: hmmm...I'm wondering if a VS test would be more appropriate? An Apex test for example?

jeffkl: What if I make a functional test that just logs the PackageReferences after importing NuGet.props/NuGet.targets?  It would just need to verify that the items and their metadata is set and I could accomplish that with a relatively hacky test.

nkolev92: Use evaluated items instead of calling a target?
That definitely tests the problem, but probably not the scenario. 

I don't expect issues like this often though, so I'm fine with that test.

jeffkl: I have added a unit test that just verifies that the `<GlobalPackageReference />` items are processed without having to call a target.

nkolev92 (approved):

zivkan (approved):

PR #5093, opened 2023-03-15
Apex method `CreateAndInitProject` Saves All files and Logs Exceptions

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2201

## Description

- `CreateAndInitProject` fails regularly and makes our CI flaky. A guess I have is that there's a file unsaved (maybe out of scope of our tests) which is affecting how the API saves the solution file. 

- Logging any exception with an easily searchable string `[NuGet Fault in Apex]:` for convenience and clarity when looking at logs
  - I think `ToString` here should show inner exceptions. If not, we'll need to modify that as the interesting error probably is there.
- I don't know if this PR solves our Apex flakiness problem, but I also think it doesn't hurt. The successful builds for my branch could be a coincidence:
image

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added - modified a test
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/build/config.props
+++ b/build/config.props
@@ -27,7 +27,7 @@
     <VsTargetMajorVersion>$([MSBuild]::Add(11, $(MajorNuGetVersion)))</VsTargetMajorVersion>
     <VsTargetBranch>main</VsTargetBranch>
     <VsTargetChannel>int.$(VsTargetBranch)</VsTargetChannel>
-    <VsTargetChannelForTests>int.d17.5</VsTargetChannelForTests>
+    <VsTargetChannelForTests>$(VsTargetChannel)</VsTargetChannelForTests>
 
     <!-- NuGet SDK VS package Semantic Version -->
     <NuGetSdkVsSemanticVersion>$(VsTargetMajorVersion).$(MinorNuGetVersion).$(PatchNuGetVersion)</NuGetSdkVsSemanticVersion>
--- a/test/NuGet.Tests.Apex/NuGet.Tests.Apex/Utility/CommonUtility.cs
+++ b/test/NuGet.Tests.Apex/NuGet.Tests.Apex/Utility/CommonUtility.cs
@@ -456,7 +456,7 @@ namespace NuGet.Tests.Apex
             }
             catch (Exception ex)
             {
-                logger.LogInformation("DGTEST " + ex.ToString());
+                logger.LogInformation("[NuGet Fault in Apex]: " + ex.ToString());
                 throw ex;
             }
         }
--- a/test/NuGet.Tests.Apex/NuGet.Tests.Apex/Utility/CommonUtility.cs
+++ b/test/NuGet.Tests.Apex/NuGet.Tests.Apex/Utility/CommonUtility.cs
@@ -443,6 +443,9 @@ namespace NuGet.Tests.Apex
                 logger.LogInformation("Adding project");
                 var project = solutionService.AddProject(ProjectLanguage.CSharp, projectTemplate, ProjectTargetFramework.V46, "TestProject");
 
+                logger.LogInformation("Saving all open and dirty files");
+                solutionService.SaveAll();
+
                 logger.LogInformation("Saving solution");
                 solutionService.Save();
 
@@ -453,7 +456,7 @@ namespace NuGet.Tests.Apex
             }
             catch (Exception ex)
             {
-                logger.LogError(ex.ToString());
+                logger.LogInformation("DGTEST " + ex.ToString());
                 throw ex;
             }
         }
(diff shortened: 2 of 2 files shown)


build/config.props
```diff
@@ -27,7 +27,7 @@
     <VsTargetMajorVersion>$([MSBuild]::Add(11, $(MajorNuGetVersion)))</VsTargetMajorVersion>
     <VsTargetBranch>main</VsTargetBranch>
     <VsTargetChannel>int.$(VsTargetBranch)</VsTargetChannel>
-    <VsTargetChannelForTests>int.d17.5</VsTargetChannelForTests>
+    <VsTargetChannelForTests>$(VsTargetChannel)</VsTargetChannelForTests>
```
nkolev92, line 30: I had a different branch for this, but glad we're reverting this :)

test/NuGet.Tests.Apex/NuGet.Tests.Apex/Utility/CommonUtility.cs
```diff
@@ -435,19 +435,30 @@ public static void UIInvoke(Action action)
...
-            project.Build();
+                logger.LogInformation("Building solution");
+                project.Build();
 
-            return project;
+                return project;
+            }
+            catch (Exception ex)
+            {
+                logger.LogInformation("[NuGet Fault in Apex]: " + ex.ToString());
```
nkolev92, line 459: I think you need to `unwrap` the exception. 
See example in https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Common/ExceptionUtilities.cs#L30.

nkolev92 (commented):
I think we might need to unwrap the exception to get even more information.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/Utility/CommonUtility.cs
donnie-msft: The reason I didn't invoke the unwrap utils or write additional code for inner exceptions is that it appears to be the default for `ToString()`. Am I missing something?

> The default implementation of [ToString](https://learn.microsoft.com/en-us/dotnet/api/system.exception.tostring?view=netframework-4.8) obtains the name of the class that threw the current exception, the message, the result of calling [ToString](https://learn.microsoft.com/en-us/dotnet/api/system.exception.tostring?view=netframework-4.8) on the inner exception, and the result of calling [Environment.StackTrace](https://learn.microsoft.com/en-us/dotnet/api/system.environment.stacktrace?view=netframework-4.8).

https://learn.microsoft.com/en-us/dotnet/api/system.exception.tostring?view=netframework-4.8#remarks

donnie-msft: Checking my own work:

 ```
        static void Main(string[] args)
        {
            try
            {
                DoWork();
            }
            catch (Exception ex) 
            { 
                Console.WriteLine(ex);
            }

            Console.ReadLine();
        }

        private static void DoWork()
        {
            throw new Exception("Outer exception", new Exception("Inner exception"));
        }
```

image

build/config.props
donnie-msft: Feel free to PR that if you prefer the separation.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/Utility/CommonUtility.cs
nkolev92: AggregateException breaks that since there can be more than 1 exception, https://learn.microsoft.com/en-us/dotnet/api/system.aggregateexception.-ctor?view=netframework-4.7.2#overloads.

Not sure if AggregateException is what gets thrown anywhere in the stack though, so I'm comfortable with whatever you decide. 
We can always update it easily.

build/config.props
nkolev92: nope, I'm good. It's a trivial change.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/Utility/CommonUtility.cs
donnie-msft: I see. No, I expect it to be `SolutionException`.

donnie-msft:
Moved back to draft as the failure still occurred, and the logging shows the same IPC error without new information.

PR #5097, opened 2023-03-16
Add nullable annotations to NuGet.Versioning and tests

zivkan:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2204

Regression? No

## Description

* Enable `<nullable>` on NuGet.Versioning and NuGet.Versioning.Tests
* Copied `NullableAttributes.cs` from the runtime, since these attributes were only added to .NET Standard 2.1, and .NET Core App 3.1 and higher. Therefore, to use these attributes (for example, `[NotNullWhen(..)]`) in .NET Standard 2.0, and .NET Framework TFMs, we need to copy these attributes locally.
* Since the BCL didn't add annotate (many) APIs until .NET 5, there's a bunch of places I had to suppress the analyzer, or use `!` to tell the compiler I know strings are not null following `!string.IsNullOrEmpty`
* Changed a bunch of `String.` to `string.` to make analyzer warnings go away, when close to code I was changing for nullable checks.

***NOTE*** This is not an ABI breaking change, but it can be an API breaking change, if the project referencing/using our package also has nullable checks enabled. I think the break changes are limited to methods that return a nullable value, or out parameters that are nullable, which should be somewhat limited. Similar to how missing annotations on the BCL cause the compiler to think that in parameters are always non-null, before this PR, the compiler probably thinks that all APIs are non-null, which is more restrictive than some APIs really are, which is not an API breaking change when we add nullable annotations.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception: no logic was added or changed.
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Core/NuGet.Versioning/FloatRange.cs
```diff
@@ -162,25 +163,25 @@ public bool Satisfies(NuGetVersion version)
...
         /// Parse a floating version into a FloatRange
         /// </summary>
-        public static FloatRange Parse(string versionString)
+        public static FloatRange? Parse(string versionString)
         {
-            _ = TryParse(versionString, out FloatRange range);
+            _ = TryParse(versionString, out FloatRange? range);
 
             return range;
         }
```
zivkan: This looks a lot like a bug. Since there's already a `TryParse` method, the `Parse` probably shouldn't return null and should throw instead. However, that would be a breaking change, so I'm not sure if we should fix the bug, or annotate the method as being capable of returning null.

src/NuGet.Core/NuGet.Versioning/FloatRange.cs
```diff
@@ -336,6 +335,7 @@ private static int CalculateVersionParts(string line)
         public override string ToString()
         {
             var result = string.Empty;
+#pragma warning disable CS8602 // Dereference of a possibly null reference. - possible bugs?
```
zivkan: I don't know if the constructor validates that when `_floatBehavior` has certain values, that `MinVersion`, `MaxVersion`, etc can not be null.

If it does not, then this method is full of NullReferenceException bugs.

src/NuGet.Core/NuGet.Versioning/SemanticVersionBase.cs
```diff
@@ -133,15 +137,15 @@ public virtual int CompareTo(SemanticVersion other, VersionComparison versionCom
...
+        public static bool operator ==(SemanticVersion? version1, SemanticVersion? version2)
         {
             return Equals(version1, version2);
         }
 
         /// <summary>
         /// Not equal
         /// </summary>
-        public static bool operator !=(SemanticVersion version1, SemanticVersion version2)
+        public static bool operator !=(SemanticVersion? version1, SemanticVersion? version2)
```
zivkan, line 148: I added nullable annotations to operators `==` and `!=`, but I wasn't sure if we should for `>`, `<`, `<=`, `>=`. The code handles it, but without looking at the implementation, it's not intuitive to know if `null` is before or after non-null values. Hence, I thought maybe we should not annotate those operators. However, `IComparable<T>` in .NET 5+ is annotated as allowing nulls in `Compare(T? x, T? y)`, so maybe the operators should be annotated as allowing nulls?

I'm leaning towards no, but I welcome any other input.

src/NuGet.Core/NuGet.Versioning/VersionComparer.cs
```diff
@@ -68,16 +68,19 @@ public bool Equals(SemanticVersion x, SemanticVersion y)
...
+            // The BCL is missing nullable annotations on IComparable<T> before net5.0
             return comparer.Compare(version1, version2);
+#pragma warning restore CS8604 // Possible null reference argument.
         }
 
         /// <summary>
         /// Gives a hash code based on the normalized version string.
         /// </summary>
-        public int GetHashCode(SemanticVersion version)
+        public int GetHashCode(SemanticVersion? version)
```
zivkan: `IEqualityComparer<T>`, even in .NET 5+, does not make `GetHashCode(T)` nullable, so I wasn't sure if I should annotate as nullable here or not. I did because the implementation checks for null and handles it gracefully (doesn't throw an exception), but I'm open to keeping the same annotations as `IEqualityComparer<T>`. In fact, if we re-target this project to .NET 5 or higher, I'm pretty sure the compiler is going to warn us that we're changing the nullable annotations for a method inherited from an interface.

src/NuGet.Core/NuGet.Versioning/VersionExtensions.cs
```diff
@@ -47,9 +47,12 @@ public static class VersionExtensions
         /// <summary>
         /// Find the version that best matches the VersionRange and the floating behavior.
         /// </summary>
-        public static INuGetVersionable FindBestMatch(this IEnumerable<INuGetVersionable> items, VersionRange ideal)
+        public static INuGetVersionable? FindBestMatch(this IEnumerable<INuGetVersionable> items, VersionRange ideal)
         {
+#pragma warning disable CS8602 // Dereference of a possibly null reference.
+            // Looks like a NRE bug
             return FindBestMatch<INuGetVersionable>(items, ideal, (e => e.Version));
+#pragma warning restore CS8602 // Dereference of a possibly null reference.
```
zivkan: This 2 arg overload calls the 3 arg overload above. The implementation of the 3 arg overload is that it iterates the `IEnumerable`, and in the first loop, the value of `e` will be null, hence if I'm reading the code correctly, this code will always throw a `NullReferenceException`.

However, if we change the 3 arg overload, so that `Func<T?, NuGetVersion> selector` uses non-nullable `T` instead of `T?` (and we fix the implementation), then this method won't have an "always crash" bug.

Since I was focusing on getting the rest of the project annotated for nullable flags, I didn't think very long about this, so I'm open to suggestions to avoid having a useless method in our public API.

src/NuGet.Core/NuGet.Versioning/VersionFormatter.cs
```diff
@@ -20,16 +20,24 @@ public class VersionFormatter : IFormatProvider, ICustomFormatter
...
                 throw new ArgumentNullException(nameof(arg));
             }
 
             if (string.IsNullOrEmpty(format) || arg is not SemanticVersion version)
             {
+#pragma warning disable CS8603 // Possible null reference return.
+                // Looks like a violation of the ICustomFormatter.Format interface
                 return null;
+#pragma warning restore CS8603 // Possible null reference return.
             }
```
zivkan: `ICustomFormatter.Format` in .NET5+ is not annotated to allow null return values. Testing the implementation of `string.Format`, it throws an exception if `format` is null, even though `ICustomFormatter.Format` claims to allow `string? format`.

I opted to supress the compiler warning and keep existing behaviour, but I'm wondering if anyone things we should break behaviour and do something different.

src/NuGet.Core/NuGet.Versioning/VersionRange.cs
```diff
@@ -253,7 +254,10 @@ public bool IsBetter(NuGetVersion current, NuGetVersion considering)
                 else
                 {
                     // neither are in range
+#pragma warning disable CS8604 // Possible null reference argument.
+                    // possible bug?
                     var curToLower = current < _floatRange.MinVersion;
+#pragma warning restore CS8604 // Possible null reference argument.
```
zivkan: As mentioned in an earlier comment, I wasn't sure if `operator <` should allow nulls or not. Since I opted not to, and this method doesn't otherwise validate `_floatRange.MinVersion != null`, the compiler is warning us that we're possibly doing something wrong.

Is this an example of why we should allow `operator <` to accept nulls?

src/NuGet.Core/NuGet.Versioning/VersionRange.cs
```diff
@@ -297,9 +301,12 @@ public VersionRange ToNonSnapshotRange()
             {
                 var minVersion = MinVersion;
 
-                if (Float.FloatBehavior == NuGetVersionFloatBehavior.Prerelease)
+                if (Float!.FloatBehavior == NuGetVersionFloatBehavior.Prerelease)
                 {
+#pragma warning disable CS8604 // Possible null reference argument.
+                    // Possible bug?
                     minVersion = GetNonSnapshotVersion(minVersion);
+#pragma warning restore CS8604 // Possible null reference argument.
```
zivkan: `GetNonSnapshotVersion` will throw `NullReferenceException` if `minVersion == null`. I don't know if `FloatBehaviour == Prerelease` means that `minVersion` will definately not be null. If not, this is a potential bug.

src/NuGet.Core/NuGet.Versioning/VersionRangeComparer.cs
```diff
@@ -90,8 +93,11 @@ public int GetHashCode(VersionRangeBase obj)
 
             combiner.AddObject(obj.IsMinInclusive);
             combiner.AddObject(obj.IsMaxInclusive);
+#pragma warning disable CS8604 // Possible null reference argument.
+            // Do we have a null bug here?
             combiner.AddObject(_versionComparer.GetHashCode(obj.MinVersion));
             combiner.AddObject(_versionComparer.GetHashCode(obj.MaxVersion));
+#pragma warning restore CS8604 // Possible null reference argument.
```
zivkan: `IEqualityComparer<T>.GetHashCode(T)` doesn't allow nulls, even in TFMs where the BCL is annotated. However, our implementation "gracefully" handles nulls. So, not sure what to do here. Should I wrap them?
```suggestion
            if (obj.MinVersion != null) { combiner.AddObject(_versionComparer.GetHashCode(obj.MinVersion)); }
            if (obj.MaxVersion != null) { combiner.AddObject(_versionComparer.GetHashCode(obj.MaxVersion)); }
```

src/NuGet.Core/NuGet.Versioning/VersionRangeFormatter.cs
```diff
@@ -65,18 +69,22 @@ public string Format(string format, object arg, IFormatProvider formatProvider)
...
+                        return formatted;
                     }
                 }
             }
 
-            return formatted;
+#pragma warning disable CS8603 // Possible null reference return.
+            // bug? ICustomFormatter.Format doesn't appear to allow null return values
+            return null;
+#pragma warning restore CS8603 // Possible null reference return.
```
zivkan: As the comment says, `ICustomFormatter.Format` doesn't allow null return values. Not sure if we should break current behaviour to comply with nullable annotations, or just leave this suppressed.

zivkan (commented):
What would probably be most useful to me, is if reviewers look carefully at PublicAPI.Shipped.txt, and provide feedback about any API that accepts null (`?`) that shouldn't, or if there's anywhere that doesn't allow null (`!`) that should. Changes to public APIs are breaking and can make it hard to upgrade package versions, so it would be really great to get it "right" the first time.

jeffkl (dismissed):
🤞 hopefully this doesn't break anyone...

src/NuGet.Core/NuGet.Versioning/FloatRange.cs
nkolev92: This is definitely a bug, lol :D 
I think we should throw rather than annotate, but I'm open to argument against that. 

I wouldn't expected too many people to be using this API directly.

src/NuGet.Core/NuGet.Versioning/FloatRange.cs
nkolev92: For there to be a floating version, MinVersion has to exist. Everything else is unparsable.

nkolev92: Should we maybe refactor out ToString to a different method and then pass MinVersion as a null ignore. 
pragra that we will never fix make readability worse.

src/NuGet.Core/NuGet.Versioning/SemanticVersionBase.cs
```diff
@@ -178,9 +182,12 @@ public virtual int CompareTo(SemanticVersion other, VersionComparison versionCom
             return Compare(version1, version2) >= 0;
         }
 
-        private static int Compare(SemanticVersion version1, SemanticVersion version2)
+        private static int Compare(SemanticVersion? version1, SemanticVersion? version2)
         {
+#pragma warning disable CS8604 // Possible null reference argument.
```
nkolev92, line 187: Should permanent suppressions be in GlobalSuppressions instead?
Hide them, pretend it doesn't exist :D

src/NuGet.Core/NuGet.Versioning/VersionComparer.cs
nkolev92: Do we know why GetHashCode has no annotations? 
I'd lean towards not annotating to match the behavior if the reasoning is good.

src/NuGet.Core/NuGet.Versioning/VersionExtensions.cs
nkolev92: Given how simple it is, IMO we should fix the bug. 

Bug fixes > incorrect annotation.

src/NuGet.Core/NuGet.Versioning/VersionFormatter.cs
nkolev92: That sounds a bit odd. Do they have a tracking issue for that? Has anyone complained?

Earlier in a code I said, I'd prefer to match the compiler where the reasoning is good, but I'm not sure it helps us here.

src/NuGet.Core/NuGet.Versioning/VersionRange.cs
nkolev92: min version is always non-null for float.

src/NuGet.Core/NuGet.Versioning/VersionRangeComparer.cs
nkolev92: I'm ok with wrapping them.
We'd be doing the same thing in the rest of our code.

src/NuGet.Core/NuGet.Versioning/VersionRangeFormatter.cs
nkolev92: Format will never really be null correct? Like we created a version object, so it will have a string representation?

src/NuGet.Core/NuGet.Versioning/VersionComparer.cs
zivkan: [The docs for `IEqualityComparer<T>.GetHashCode` say](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.iequalitycomparer-1.gethashcode?view=net-7.0)

> Exceptions
> [ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception?view=net-7.0)
> The type of obj is a reference type and obj is null.

My guess is that it's like asking what colour a non-existent ball is. It doesn't fundamentally make sense to ask what the hash code for a null value is. Our implementation returns `0` for this scenario, but `0` is a perfectly valid hash code for a non-null value to have, so now there's a potential hash collision between null and non-null values (although there's also potentially hash collisions for other non-null values, which is why it's critical to check equality and don't assume hash codes are unique).

I guess this is also the reason why `Dictionary<T1, T2>` doesn't allow null keys (even though there are times where that would be useful).

I'll remove the nullable annotation, but I won't change the implementation (to throw an exception, rather than return 0) for now.

src/NuGet.Core/NuGet.Versioning/VersionFormatter.cs
zivkan: Thinking about this more, and glancing over [the docs for ICustomFormatter](https://learn.microsoft.com/en-us/dotnet/api/system.icustomformatter?view=net-7.0) (but still not reading it very carefully), I was confusing `format == null` and the return value.

Firstly about the parameter `format == null`, consider a line `string.Format(new MyCustomFormatter(), "{0} {0:A}")`. I believe that `MyCustomFormatter.Format(...)` will be called twice, the first time with `format: null`, the second time with `format: "A"`.  Therefore, when `format == null`, it means "use the default format". This makes it intuitive that this is a supported, and designed for, scenario.

But about the return value being null, using C# Interactive, `string a = null; string.Format("{0}", a)` return the empty string `""`. Therefore, I believe the "correct" thing to do is return the empty string, not null.  I'm kind worried about regressions in this change of behaviour though.  I'll make changes, see if tests fail, and make a decision. Having code work more accurately according to interface docs is a good thing, but might be a little frustrating if it breaks long standing behaviour.

src/NuGet.Core/NuGet.Versioning/VersionRangeFormatter.cs
zivkan: Similar to another comment, `format` can be null with this example: `string.Format(VersionRangeFormatter.Instance, "{0}", versionRange);`. If `{0}` is replaced with something like `{0:N}`, then `format == "N"`, but when the format string is `{0}`, then `format == null`.

However, this code has other codepaths where it will reach this null return value, like when the type passed into `arg` is neither castable to `IFormattable` or `VersionRange`.

My preference is to throw an `ArgumentException`. I don't suspect that any of NuGet's own code will ever hit the exception, but since this class is public, who knows if someone somewhere is doing something like that.

src/NuGet.Core/NuGet.Versioning/VersionRange.cs
zivkan: In that case, this constructor shouldn't exist with the implementation that it does: https://github.com/NuGet/NuGet.Client/blob/fc41cff354547a3c7f84d88d10d4e0bdf79d9c4b/src/NuGet.Core/NuGet.Versioning/FloatRange.cs#L23-L26

It's passing `null` as the value for `minVersion`.

Also, this constructor shouldn't be handling the `minVersion == null` case: https://github.com/NuGet/NuGet.Client/blob/fc41cff354547a3c7f84d88d10d4e0bdf79d9c4b/src/NuGet.Core/NuGet.Versioning/FloatRange.cs#L44-L57

What I find interesting is that `VersionRange.Parse("*").MinVersion == "[0.0]"`. So, I'm tempted to change the implementation so that the ctor without a `NuGetVersion minVersion` will pass `NuGetVersion.Parse("0.0")` as the value, and change the other constructor(s) to throw when `minVersion == null`, but this is another example of where it feels scary to make breaking changes to the API.  NuGet as a product may never do it that way, but it's a public API. On the other hand, I can already get null reference exceptions doing `new FloatRange(NuGetVersionFloatBehaviour.Minor).ToString()`. So, the constructor probably should have validated, but didn't.

Another option is to add validation, and allow `minVersion == null` when FloatBehaviour == Major, AbsoluteLatest, and maybe even PreReleaseMajor. But then we'll have to use `!` to override the compiler's nullability check when we believe it should never be null, and any mistakes will still result in `NullReferenceException` bugs. 🤷

src/NuGet.Core/NuGet.Versioning/FloatRange.cs
zivkan: > For there to be a floating version, MinVersion has to exist. Everything else is unparsable.

This might be the intention, and it may be true when NuGet is used as a tool, but as mentioned in [this comment](https://github.com/NuGet/NuGet.Client/pull/5097#discussion_r1143367788), the API design doesn't enforce it, and there are several scenarios where it doesn't appear to be true.

I was spending too much time on trying to understand `FloatRange` and `VersionRange`, so I didn't end up making any changes to these two places where the compiler is warning us about potential issues. The pragmas don't help readability, but at the same time they warn us about potential problems with the software design of these classes. It's certainly possible to use the APIs to cause null reference exceptions, but it might be true that our CLI and VS experiences will never hit any of the NRE scenarios.

I propose to leave it as is for now, and if someone in the future is modifying (or perhaps just reading) this code, they might have more time to deeply understand and refactor to allow the compiler to give us better null checking.

src/NuGet.Core/NuGet.Versioning/VersionRange.cs
zivkan: From Nikolche's other comments, the intention is that floating versions should always have a `MinVersion`. However, the API doesn't enforce this, so it's possible for someone using the NuGet.Versioning package in their own app to have `_floatRange.MinVersion == null` at this line.

src/NuGet.Core/NuGet.Versioning/SemanticVersionBase.cs
zivkan: For these one-line methods, GlobalSuppressions could be used. Although, if anyone refactors the method in the future, then they won't have null checking protection from the compiler. A problem with GlobalSuppressions.cs is that it applies to entire methods, so I definitely wouldn't want to use it for multi-line method implementations, but that's the not case for this specific example. Personally, I prefer the inline suppressions. I don't like the GlobalSuppressions file, even when I want a suppression to apply to the entire method body.

src/NuGet.Core/NuGet.Versioning/FloatRange.cs
nkolev92: Sure, please remove the code comments in that case.

src/NuGet.Core/NuGet.Versioning/VersionRange.cs
nkolev92: VersionRangeFactory uses the HasMinVersion thing as a means to detect whether a float range is valid...https://github.com/NuGet/NuGet.Client/blob/fc41cff354547a3c7f84d88d10d4e0bdf79d9c4b/src/NuGet.Core/NuGet.Versioning/VersionRangeFactory.cs#L204-L214...which is yeah, not great :D 

I think it'd be great to not annotate incorrectly, so I'm ok with changing the constructor to validate if you are. 

I don't expect people to be playing with invalid versions/ranges.

zivkan: Improved 👍

src/NuGet.Core/NuGet.Versioning/SemanticVersionBase.cs
nkolev92: I'll resolve this. There's so many other changes not worth focusing on this. 
Alternatively we can suppress on top of the method too.

src/NuGet.Core/NuGet.Versioning/FloatRange.cs
```diff
@@ -73,7 +74,7 @@ public FloatRange(NuGetVersionFloatBehavior floatBehavior, NuGetVersion minVersi
         /// <summary>
         /// The original release label. Invalid labels are allowed here.
         /// </summary>
-        public string OriginalReleasePrefix => _releasePrefix;
+        public string? OriginalReleasePrefix => _releasePrefix;
```
nkolev92, line 82: This can probably have a MemberNotNullWhen IsPrerelease correct?

src/NuGet.Core/NuGet.Versioning/FloatRange.cs
```diff
@@ -105,12 +106,13 @@ public bool Satisfies(NuGetVersion version)
...
                        && _minVersion.Minor == version.Minor
                        && _minVersion.Patch == version.Patch
-                       && ((version.IsPrerelease && version.Release.StartsWith(_releasePrefix, StringComparison.OrdinalIgnoreCase))
+                       && ((version.IsPrerelease && version.Release.StartsWith(_releasePrefix!, StringComparison.OrdinalIgnoreCase))
                            || !version.IsPrerelease);
                 }
                 else if (_floatBehavior == NuGetVersionFloatBehavior.PrereleasePatch)
                 {
                     // allow the stable version to match
+#pragma warning disable CS8604 // Possible null reference argument. - usage of _releasePrefix don't appear to be null-safe
```
nkolev92: Wonder if the NotNullWhen helps.

src/NuGet.Core/NuGet.Versioning/VersionRange.cs
```diff
@@ -297,9 +298,12 @@ public VersionRange ToNonSnapshotRange()
             {
                 var minVersion = MinVersion;
 
-                if (Float.FloatBehavior == NuGetVersionFloatBehavior.Prerelease)
+                if (Float!.FloatBehavior == NuGetVersionFloatBehavior.Prerelease)
                 {
+#pragma warning disable CS8604 // Possible null reference argument.
+                    // Possible bug?
                     minVersion = GetNonSnapshotVersion(minVersion);
```
nkolev92: Would MemberNotNullWhen help here as well?

nkolev92 (commented):
A couple of ideas about using MemberNotNullWhen.

src/NuGet.Core/NuGet.Versioning/FloatRange.cs
zivkan: I had to add a new public API, `IncludesPrerelease`, but yes, it was possible.

src/NuGet.Core/NuGet.Versioning/VersionRange.cs
zivkan: This was quite problematic. See this commit: https://github.com/NuGet/NuGet.Client/pull/5097/commits/92917270a9526c115f9914de784678c6aad7ae83

Firstly, `MemberNotNullWhenAttribute` appears to only allow you to reference members that are in the same class. `MinVersion` is defined in `VersionRangeBase`, whereas all float stuff is in `VersionRange`. Therefore, I had to add `public new NuGetVersion? MinVersion => base.MinVersion;`, but then all the annotations from `VersionRangeBase` that says `MinVersion` is not null when `IncludeLowerBound`, `HasMinVersion`, or `IncludeLowerAndUpperBound` is true. So, then I had to copy those properties, but since `IncludeLowerAndUpperBound` has `MemberNotNullWhen(true, nameof(MaxVersion))`, then I had to copy all the `MaxVersion` properties too.

Secondly, I added a check in the constructor so when `floatRange != null && floatRange.FloatBehaviour != null && minVersion == null` it throws, so that we ensure that the `MemberNotNullWhen(true, nameof(MinVersion))` is actually validated. However, the two obsolete properties `VersionRange.AllFloating` and `VersionRange.AllStableFloating` violated this assertion and threw exceptions. Changing the constructors to pass in `NuGetVersion.Parse("0.0.0")` as the min version caused the tests that assert the `ToNormalizedString` output changed.  Maybe I could add an internal contructor for the two obsolete static instances in `VersionRangeFactory` to avoid throwing, but then they violate the not-null annotation, and those code paths might throw exceptions.

If we care most about code-correctness, I think we should keep the commit. But it really feels bad.

The new public API changes in the commit could be removed if we delete the pointless `VersionRangeBase` abstraction, but unfortunately that's public, so would be a public API breaking change. That wouldn't solve the behaviour change in the obsolete `AllFloating` and `AllStableFloating` `ToNormalizedString` output changes.  If we're going to break public APIs, maybe we just delete those APIs, since they were marked obsolete in NuGet 5.5 in 2019. It would be nice to follow SemVer and wait until the next major NuGet version increment, but our last attempt at making breaking changes at a major version change wasn't successful.

jeffkl (approved):

nkolev92 (approved):

PR #5103, opened 2023-03-21
Fix project restore when CentralPackageVersionOverrideEnabled=false

### LINKED ISSUE #12500: Set CentralPackageVersionOverrideEnabled=false in project with CPM broke project restore
### NuGet Product Used

dotnet.exe

### Product Version

8.0.0-preview.2.23128.3

### Worked before?

_No response_

### Impact

### Repro Steps & Context

1. Create project with CPM enabled. 
2. Add `<CentralPackageVersionOverrideEnabled>false</CentralPackageVersionOverrideEnabled>` to project file.
3. Run `dotnet restore`

The restore fails but does not log an error.  This is because of bad logic here: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs#L478

### Verbose Logs

_No response_

kant2002:
This is fix restoring of the project under CPM. If adding `<CentralPackageVersionOverrideEnabled>false</CentralPackageVersionOverrideEnabled>` to clean project then `dotnet restore` fails to work.

## Bug

Fixes: https://github.com/nuget/home/issues/12500

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests.cs
@@ -1524,67 +1524,6 @@ namespace NuGet.Commands.Test
             }
         }
 
-        [Fact]
-        public async Task RestoreCommand_CentralVersion_NoErrorForCleanProjectWhenVersionOverrideDisabled()
-        {
-            // Arrange
-            using (var pathContext = new SimpleTestPathContext())
-            {
-                var projectName = "TestProject";
-                var projectPath = Path.Combine(pathContext.SolutionRoot, projectName);
-                var outputPath = Path.Combine(projectPath, "obj");
-                var dependencyBar = new LibraryDependency(new LibraryRange("bar", LibraryDependencyTarget.All),
-                        LibraryIncludeFlags.All,
-                        LibraryIncludeFlags.All,
-                        new List<NuGetLogCode>(),
-                        autoReferenced: false,
-                        generatePathProperty: true,
-                        versionCentrallyManaged: false,
-                        LibraryDependencyReferenceType.Direct,
-                        aliases: null,
-                        versionOverride: null);
-
-                await SimpleTestPackageUtility.CreateFolderFeedV3Async(
-                    pathContext.PackageSource,
-                    PackageSaveMode.Defaultv3,
-                    new SimpleTestPackageContext("foo", "1.0.0"),
-                    new SimpleTestPackageContext("bar", "2.0.0")
-                    );
-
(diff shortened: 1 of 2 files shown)


test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests.cs
jeffkl, line 1: The fix looks good but I think I'd rather have you extend the existing test, [RestoreCommand_CentralVersion_ErrorWhenVersionOverrideUsedButIsDisabled](https://github.com/NuGet/NuGet.Client/blob/dev/test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests.cs#L2701), instead of adding a new one.  I'm surprised the existing test didn't catch this so I think it would be better to improve it.

Have that unit test take two booleans:
```c#
[Theory]
[InlineData(true, true)]
[InlineData(true, false)]
[InlineData(false, true)]
[InlineData(false, false)]
public async Task RestoreCommand_CentralVersion_ErrorWhenVersionOverrideUsedButIsDisabled(bool isCentralPackageVersionOverrideDisabled, bool isVersionOverrideUsed)
```
Then we can test all of the combinations of the feature is enabled/disabled and the feature is used/not used.

Thanks for contributing!  I can elaborate more or push just a test change to your branch if that doesn't make sense.

kant2002: Isn't that lead to have `Assert` block of that test to have 4 combiation. Right now is 2 variants here https://github.com/NuGet/NuGet.Client/blob/fc41cff354547a3c7f84d88d10d4e0bdf79d9c4b/test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests.cs#L2756-L2773

Maybe there better way for that?

jeffkl: Yes you'll need to update this section to verify all 4 combinations (which this test should have been doing in the first place)

kant2002: Add some noise while fix typo in variable. Otherwise changes looks not so bad.

jeffkl (approved):
Your contribution is greatly appreciated!

nkolev92 (approved):

jeffkl:
@nkolev92 can I get a merge?  The only failing test is flaky.

PR #5107, opened 2023-03-24
Add missing RegistrationsBaseUrls to prevent exceptions when loading valid Service Indexes.

### LINKED ISSUE #12455: SourceRepository.GetResource throws if type is not an exact match
### NuGet Product Used

Other/NA

### Product Version

Chocolatey v2.0.0-alpha-20230124

### Worked before?

_No response_

### Impact

### Repro Steps & Context

[As per the NuGet V3 API specification](https://learn.microsoft.com/en-us/nuget/api/registration-base-url-resource), `RegistrationsBaseUrl`, `RegistrationsBaseUrl/3.0.0-beta` and `RegistrationsBaseUrl/3.0.0-rc` are all aliases for the same `@type`. However, if the server provides only one of them (e.g., `RegistrationsBaseUrl`), the client throws an ArgumentNullException.

### Verbose Logs

```shell
System.ArgumentNullException: Value cannot be null.
Parameter name: baseUrl
   at NuGet.Protocol.RegistrationResourceV3..ctor(IHttpSource client, Uri baseUrl)
   at NuGet.Protocol.RegistrationResourceV3Provider.<TryCreate>d__1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at NuGet.Protocol.Core.Types.SourceRepository.<GetResourceAsync>d__16`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NuGet.Protocol.Core.Types.SourceRepository.<GetResourceAsync>d__15`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NuGet.Protocol.PackageMetadataResourceV3Provider.<TryCreate>d__1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at NuGet.Protocol.Core.Types.SourceRepository.<GetResourceAsync>d__16`1.MoveNext()
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at NuGet.Protocol.Core.Types.SourceRepository.GetResource[T](CancellationToken token)
   at NuGet.Protocol.Core.Types.SourceRepository.GetResource[T]()
   at chocolatey.infrastructure.app.nuget.NugetCommon.<GetRepositoryResource>d__7`1.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at chocolatey.infrastructure.app.services.NugetService.install_run(ChocolateyConfiguration config, Action`2 continueAction)
   at chocolatey.infrastructure.app.services.ChocolateyPackageService.perform_source_runner_function[T](ChocolateyConfiguration config, Func`2 function)
   at chocolatey.infrastructure.app.services.ChocolateyPackageService.install_run(ChocolateyConfiguration config)
   at chocolatey.infrastructure.app.runners.GenericRunner.run(ChocolateyConfiguration config, Container container, Boolean isConsole, Action`1 parseArgs)
   at chocolatey.infrastructure.app.runners.ConsoleApplication.run(String[] args, ChocolateyConfiguration config, Container container)
   at chocolatey.console.Program.Main(String[] args)
```

atamagaii:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12455

Not a regression.
I'm not sure this ever worked fully as corner cases have always been able to cause problems

## Description

Moved the RegistrationsBaseUrl types into their own static class for the sake of clarity.

Added the missing versions based on the api documentation [here](https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/api/registration-base-url-resource.md)
Extracted the ClientVersions into their own class for clarity.
Tests for all the version literals
Added check so that the TryCreate method returns false instead of throwing an exception since that's what we expect "try" methods to do.
## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added

- **Documentation**
  
https://github.com/NuGet/Home/issues/12455


src/NuGet.Core/NuGet.Protocol/PublicAPI/net472/PublicAPI.Shipped.txt
```diff
@@ -2033,15 +2033,6 @@ static readonly NuGet.Protocol.ProxyAuthenticationHandler.MaxAuthRetries -> int
...
 ~static readonly NuGet.Protocol.ServiceTypes.SymbolPackagePublish -> string[]
-~static readonly NuGet.Protocol.ServiceTypes.Version200 -> string
-~static readonly NuGet.Protocol.ServiceTypes.Version300 -> string
-~static readonly NuGet.Protocol.ServiceTypes.Version300beta -> string
-~static readonly NuGet.Protocol.ServiceTypes.Version340 -> string
-~static readonly NuGet.Protocol.ServiceTypes.Version470 -> string
-~static readonly NuGet.Protocol.ServiceTypes.Version490 -> string
-~static readonly NuGet.Protocol.ServiceTypes.Version500 -> string
-~static readonly NuGet.Protocol.ServiceTypes.Version510 -> string
-~static readonly NuGet.Protocol.ServiceTypes.Versioned -> string
```
zivkan: While I believe it's very unlikely that anyone using the NuGet.Protocol package in their own app is using any of these APIs, it's a breaking change to move the strings to a different type. In my opinion, the "benefit" doesn't warrant the risk of breaking APIs.

src/NuGet.Core/NuGet.Protocol/PublicAPI/net472/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,30 @@
...
+~static readonly NuGet.Protocol.ClientVersions.Versioned -> string
+*REMOVED*~static readonly NuGet.Protocol.ServiceTypes.Version200 -> string
+*REMOVED*~static readonly NuGet.Protocol.ServiceTypes.Version300 -> string
+*REMOVED*~static readonly NuGet.Protocol.ServiceTypes.Version300beta -> string
+*REMOVED*~static readonly NuGet.Protocol.ServiceTypes.Version340 -> string
+*REMOVED*~static readonly NuGet.Protocol.ServiceTypes.Version470 -> string
+*REMOVED*~static readonly NuGet.Protocol.ServiceTypes.Version490 -> string
+*REMOVED*~static readonly NuGet.Protocol.ServiceTypes.Version500 -> string
+*REMOVED*~static readonly NuGet.Protocol.ServiceTypes.Version510 -> string
+*REMOVED*~static readonly NuGet.Protocol.ServiceTypes.Versioned -> string
```
zivkan: Git already tracks that APIs were removed in PublicAPI.Shipped.txt, so I don't think this really adds anything.  Additionally, does the PublicApiAnalyzer even work with these lines in this file?  Even if it does, it adds extra process when we ship the new version of NuGet, as we'd need to remember to manually delete these lines when everything else is otherwise copied into `PublicAPI.Shipped.txt`

src/NuGet.Core/NuGet.Protocol/PublicAPI/net472/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,30 @@
...
+~static readonly NuGet.Protocol.RegistrationsBaseUrlTypes.RegistrationsBaseUrl -> string
+~static readonly NuGet.Protocol.RegistrationsBaseUrlTypes.RegistrationsBaseUrlVersion300beta -> string
+~static readonly NuGet.Protocol.RegistrationsBaseUrlTypes.RegistrationsBaseUrlVersion300rc -> string
+~static readonly NuGet.Protocol.RegistrationsBaseUrlTypes.RegistrationsBaseUrlVersion340 -> string
+~static readonly NuGet.Protocol.RegistrationsBaseUrlTypes.RegistrationsBaseUrlVersion360 -> string
+~static readonly NuGet.Protocol.RegistrationsBaseUrlTypes.RegistrationsBaseUrlVersioned -> string
+~static readonly NuGet.Protocol.RegistrationsBaseUrlTypes.RegistrationsBaseUrls -> string[]
```
zivkan: I wrote some docs on the NuGet SDK and API changes a while ago: https://github.com/NuGet/NuGet.Client/blob/dev/docs/nuget-sdk.md#changes-to-apis

Basically, are these types useful to customers who use NuGet.Protocol in their own apps? I don't think they are, as I'd expect developers to use `SourceRepository.GetResourceAsync<T>`, and all these constants are just an internal implementation detail.

Unfortunately people on the NuGet team years ago weren't mindful about what they made public vs internal, but this has bitten us, making it harder to make perf improvements, and sometimes bug fixes, which is why we're much more deliberate about public API changes now. As much as I'd love to hide a bunch of public APIs that have already shipped, as mentioned, we've had problems in the past with people using them, even though I believe there were better ways of doing what they wanted to achieve, but in any case we try not to break people. Therefore, we don't want to add new APIs that are not useful and therefore add technical debt.

src/NuGet.Core/NuGet.Protocol/RegistrationsBaseUrlTypes.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// <summary>
+        /// These registrations are compressed using Content-Encoding: gzip.
+        /// SemVer 2.0.0 packages are included in this hive.
+        /// </summary>
+        public static readonly string RegistrationsBaseUrlVersion360 = $"{TypeName}{ClientVersions.Version360}";
+
+        /// <summary>
+        /// An alias for RegistrationsBaseUrlVersion360
+        /// </summary>
+        public static readonly string RegistrationsBaseUrlVersioned = $"{TypeName}{ClientVersions.Versioned}";
```
nkolev92: These can all just be declared in the static string. They're only used in tests.

src/NuGet.Core/NuGet.Protocol/RegistrationsBaseUrlTypes.cs
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.Protocol
+{
+    /// <summary>
+    /// This represents three distinct registration hives available for various client versions.
+    /// </summary>
+    public static class RegistrationsBaseUrlTypes
```
nkolev92: This can stay internal.

src/NuGet.Core/NuGet.Protocol/ServiceTypes.cs
```diff
@@ -1,29 +1,19 @@
 // Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+// Licensed under the Apache License, ClientVersions.Version 2.0. See License.txt in the project root for license information.
 
 namespace NuGet.Protocol
 {
     public static class ServiceTypes
     {
-        public static readonly string Version200 = "/2.0.0";
```
nkolev92, line 8: We try not to break public APIs.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/RegistrationsBaseUrlTypesTests.cs
```diff
@@ -0,0 +1,62 @@
...
+        [Fact]
+        public void RegistrationsBaseUrlTypes_RegistrationsBaseUrl_Has_Correct_Coresponding_String_Literal()
+        {
+            RegistrationsBaseUrlTypes.RegistrationsBaseUrl.Should().Be("RegistrationsBaseUrl");
+        }
+
+        [Fact]
+        public void RegistrationsBaseUrlTypes_RegistrationsBaseUrl300beta_Has_Correct_Coresponding_String_Literal()
+        {
+            RegistrationsBaseUrlTypes.RegistrationsBaseUrlVersion300beta.Should().Be("RegistrationsBaseUrl/3.0.0-beta");
```
nkolev92: This is testing the value of the string, but not testing the code that's actually used directly which is the static array. 

I think all tests should rely on that instead of the static string that's not really used anywhere outside of the RegistrationsBaseUrlTypes class.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/Resources/RegistrationResourceV3ProviderTests.cs
```diff
@@ -0,0 +1,44 @@
...
+using Newtonsoft.Json.Linq;
+using NuGet.Protocol.Core.Types;
+using Xunit;
+
+namespace NuGet.Protocol.Tests.Resources
+{
+    public class RegistrationResourceV3ProviderTests
+    {
+        [Fact]
+        public async Task TryCreate_When_RegistrationsBaseUrl_Version_Is_Unuspported_Returns_False()
```
nkolev92, line 18: I really appreciate this test :)
Great job.

nkolev92 (commented):
Thank you for your contribution.

Andy linked you to our policy on public APIs, so please address that.

src/NuGet.Core/NuGet.Protocol/PublicAPI/net472/PublicAPI.Unshipped.txt
atamagaii: Okay, I have pushed some updates and paired down the code. In this go around I added the missing versions alongside the existing ones in the ServiceTypes file. This does create new entries in the public api, but follows the pattern of exposing the version strings that the class was already doing.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/RegistrationsBaseUrlTypesTests.cs
atamagaii: In an effort to reduce some noise I ended up ditching the extracted class and reverted back to initializing the array in ServiceTypes. I did update the test to check the expected version strings directly from ServiceTypes.RegstrationsBaseUrl.

src/NuGet.Core/NuGet.Protocol/PublicAPI/net472/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,2 @@
-#nullable enable
```
kartheekp-ms, line 1: The current dev branch has `#nullable enable`. I think we should add it back for all the 3 public api files modified in this PR.
https://github.com/NuGet/NuGet.Client/blob/80c3c74ca5c6d67912a19833e8cc2b10f875dc95/src/NuGet.Core/NuGet.Protocol/PublicAPI/netstandard2.0/PublicAPI.Unshipped.txt#L1

src/NuGet.Core/NuGet.Protocol/ServiceTypes.cs
```diff
@@ -7,16 +7,18 @@ public static class ServiceTypes
     {
         public static readonly string Version200 = "/2.0.0";
         public static readonly string Version300beta = "/3.0.0-beta";
+        public static readonly string Version300rc = "/3.0.0-rc";
```
nkolev92, line 10: If this is internal, does anything break? I'm guessing no, so we can probably do that too. No need to increase the public API despite the precedent. 
Thoughts @NuGet/nuget-client

nkolev92 (approved):

src/NuGet.Core/NuGet.Protocol/ServiceTypes.cs
kartheekp-ms: I agree with above comment. There is no need to add new public API surface if it is not required.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/ServiceTypesTests.cs
```diff
@@ -20,5 +21,20 @@ public void RepositorySignatures_AllResourcesAreVersioned()
...
+        public void RegistrationsBaseUrls_Returns_All_Versions_In_Desc_Order()
+        {
+            string[] expected =
+            {
+                "RegistrationsBaseUrl/Versioned",
+                "RegistrationsBaseUrl/3.6.0",
+                "RegistrationsBaseUrl/3.4.0",
+                "RegistrationsBaseUrl/3.0.0-rc",
+                "RegistrationsBaseUrl/3.0.0-beta",
+                "RegistrationsBaseUrl"
```
erdembayar, line 35: Is it correct ordering assumption? 
Does non-versioned index need to be the last one to try?

jeffkl: Team Review: Yes

erdembayar (approved):

src/NuGet.Core/NuGet.Protocol/ServiceTypes.cs
jeffkl: Team Review: In this case it doesn't matter since they are readonly strings and we'll never change them anyway

jeffkl (approved):

PR #5109, opened 2023-03-28
Implement creating PackageSourceMapping through PackageManagement & Restore

### LINKED ISSUE #11797: Package Source Mapping UI - PM UI Actions support source pinning & create mappings in nuget.config
Refer to https://github.com/NuGet/Home/issues/11361 and https://github.com/NuGet/Home/blob/dev/implemented/2021/PackageSourceMapping.md

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Home/issues/11797

## Description

Plumbing changes and tests to support creating a new Package Source Mapping via PackageManagement actions. These changes will serve as an entry point for PM UI changes in a separate PR.

-  Passing `ISettings` into `DependencyGraphSpecRequestProvider` so that a manipulated object can be passed in.
  - Since the settings object has a long life within `NuGetPackageManager`, I am storing the original mappings prior to beginning preview restore, and then resetting the `ISettings` object to its original state after the preview restore finishes. 

-  Looks at existing mappings and only adds mappings if no existing mappings have patterns that already satisfy the new package pattern.

    - eg,  If `Microsoft.*` is already mapped, and an install introduced `Microsoft.Win32.Registry`, we will not write a mapping explicitly for `Microsoft.Win32.Registry` as it's already satisfied.

- With this PR, there's no analysis of the GPF to identify where a package came from.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Core/NuGet.Commands/PublicAPI/net472/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,5 @@
 #nullable enable
+~NuGet.Commands.RestoreArgs.NewMappingID.get -> string
+~NuGet.Commands.RestoreArgs.NewMappingID.set -> void
+~NuGet.Commands.RestoreArgs.NewMappingSource.get -> string
+~NuGet.Commands.RestoreArgs.NewMappingSource.set -> void
```
zivkan: This doesn't feel right to me. Merging two different package source mappings doesn't feel like something that Restore should know about.

Take PM UI preview install as a similar example (in fact, this PR modifies the PreviewInstall API signature). it doesn't have separate "actually installed packages" and "new packages to install" inputs. It has just one installed packages list (the data type is PackageSpec), and PM UI modifies the project's actual/current installed packages list with the changes from the PM UI, and then runs a restore, and from RestoreArg's point of view, it looks like every other restore.

Looking at RestoreArgs, it doesn't have the data types needed to override default values, like which packages a project references, or the settings to use. This makes sense, since `RestoreArgs` is a representations of the command line arguments of a restore. `RestoreRunner` has another overload that takes in `RestoreRequest`, which does have the right data types. So, putting aside the layers of abstraction, fundamentally I think that it should 1. Clone settings, and merge the new package source mapping, so it looks like the changes have been written to disk, but it's really just in-memory. 2. Create the appropriate RestoreRequests for the solution, using the custom settings, rather than the "real" settings. I expect that the PreviewInstall already does this to enable package install preview. 3. Call RestoreRunner.RunAsync with the custom RestoreRequests

In fact, since `RestoreArgs` is a representation of the command line args for a restore, it's another reason why this doesn't feel right. This PR/feature isn't adding a way to use different package source mapping config on the command line, so this doesn't seem like the right data structure to modify.

donnie-msft: > Merging two different package source mappings doesn't feel like something that Restore should know about.

Hm, I see your point. Especially if CLI won't need a feature like "enable automatically mapping newly installed packages to default sources" or something, then CLI would never need to use these args to inject a new mapping.
It's probably trivial, but I'll have to think about refactoring that after vacation.

Appreciate all input in the meantime!

nkolev92: I looked over the changes, and I like @zivkan's idea. Beyond that, it seems pretty reasonable what we're trying to do. 

- Adding `packageid` ensures that the new id can come from the selected source.
- Adding `*` in the preview restore allows new packages to come from the selected source only. 

That in addition to the post-restore handling in case of failure/user rejecting, I think this should be pretty good.

donnie-msft:
Think I have an idea for changing the provider to manipulate settings and avoid the `DependencyGraphSpecRequestProvider` change.
Back to draft...

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
```diff
@@ -459,6 +471,37 @@ private static string GenerateUpgradeReport(string? projectName, string backupPa
...
+                                .ToDictionary(pair => pair.Key, pair => (IReadOnlyList<string>)(pair.Patterns.Select(p => p.Pattern).ToList()));
+
+                            PackageSourceMapping packageSourceMapping = new(patternsReadOnly);
+
+                            // Expand all patterns/globs so we can later check if this package ID was already mapped.
+                            var addedPackageIds = addedPackages.Select(action => action.Item1);
+                            var addedPackageIdsWithoutExistingMappings = new List<string>();
+                            foreach (string? addedPackageId in addedPackageIds)
+                            {
+                                var configuredSource = packageSourceMapping.GetConfiguredPackageSources(addedPackageId);
```
donnie-msft: This is where the heavy lifting of determining if an existing mapping (pattern or glob) covers the package being added. Only the delta of packages being brought in that were not previously mapped will be mapped.

Possibly this logic could eventually move into the `PackageSourceMappingProvider`. I think it's safe to put here for preview 1 and I can move it in a later iteration if that's a valuable change.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
```diff
@@ -1596,7 +1596,9 @@ await PreviewBuildIntegratedProjectActionsAsync(buildIntegratedProject, actions,
             INuGetProjectContext nuGetProjectContext,
             IReadOnlyCollection<SourceRepository> activeSources,
             VersionRange versionRange,
-            CancellationToken token)
+            CancellationToken token,
+            string newMappingID = null,
+            string newMappingSource = null)
```
zivkan: I don't know why we don't have [CA1068](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1068) enabled, but normally cancellation tokens should always be the last parameter of a method.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
```diff
@@ -38,6 +38,11 @@ public sealed class UIActionEngine
         private readonly NuGetPackageManager _packageManager;
         private readonly INuGetLockService _lockService;
 
+        /// <summary>
+        /// Property is referenced by the Preview Window.
+        /// </summary>
+        public bool IsPackageSourceMappingEnabled { get; private set; }
```
zivkan: This doesn't appear to be used.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
zivkan: Is it feasible to move everything inside the `if (userAction?.SourceMappingSourceName != null && addedPackages != null)` to a method? Maybe even move the if into the method, so this part of the diff is more like:

```diff
                    if (!cancellationToken.IsCancellationRequested)
                    {
+                       ConfigureNewPackageSourceMapping(...);
                        await projectManagerService.ExecuteActionsAsync(
                            actions,
                            cancellationToken);
```

It's easier to remember guidelines and best practises when reviewing other people's code, rather than when writing it yourself, so I don't always follow my own advice. But when possible, I like methods to either do business logic, or do orchestration. I think when a method mixes both it becomes more difficult to understand. Plus this method, `PerformActionImplAsync` is over 300 lines long, so I hope it'll be slightly easier to understand if split into smaller methods. But I know this isn't always easy depending on which variables need mutation, and other reasons.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
```diff
@@ -889,6 +934,43 @@ private PackageLicenseInfo GeneratePackageLicenseInfo(IPackageSearchMetadata met
             return results;
         }
 
+        private static void CreateAndSavePackageSourceMappings(string sourceName, string[] newPackageIdsToSourceMap, PackageSourceMappingProvider mappingProvider)
```
zivkan: Could this method be moved into a different file? I know it's very opinionated, but I don't like that NuGet has so many long (1000+ line) files. I love that this method is static, as that usually means it's easier to unit test. Even if this method isn't reused elsewhere, I think having shorter classes, with methods whose name apply more directly to the class name, helps understanding.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
```diff
@@ -889,6 +934,43 @@ private PackageLicenseInfo GeneratePackageLicenseInfo(IPackageSearchMetadata met
...
+                        .Where(mappingItem => mappingItem.Key == sourceName)
+                        .FirstOrDefault();
+
+            // Source is being mapped for the first time.
+            if (existingPackageSourceMappingItemForSource is null)
+            {
+                existingPackageSourceMappingItemForSource = new(sourceName, newPackagePatternItems);
+                newAndExistingPackageSourceMappingItems.Add(existingPackageSourceMappingItemForSource);
+            }
+            else // Source already had an existing mapping.
```
zivkan: A few lines above where this method is called, a variable named `packageIdsNeedingNewSourceMappings` is defined, with a comment saying that it's already calculated packages that don't already have source mappings.

I haven't read this code carefully enough to figure out what the answer is, but are "package IDs that don't already have mappings" being calculated twice? Or is one of the two comments no longer correct?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetProjectManagerService.cs
```diff
@@ -441,7 +441,9 @@ public async ValueTask ExecuteActionsAsync(IReadOnlyList<ProjectAction> actions,
             DependencyBehavior dependencyBehavior,
             IReadOnlyList<string> packageSourceNames,
             VersionRange? versionRange,
-            CancellationToken cancellationToken)
+            CancellationToken cancellationToken,
+            string? newMappingID = null,
+            string? newMappingSource = null)
```
zivkan: I think I commented elsewhere already, but CA1068 usually gets us to put CancellationTokens as the last method parameter. I don't know why this analyzer isn't running for us.

src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/INuGetProjectManagerService.cs
```diff
@@ -71,8 +71,10 @@ public interface INuGetProjectManagerService : IDisposable
             bool includePrelease,
             DependencyBehavior dependencyBehavior,
             IReadOnlyList<string> packageSourceNames,
-            VersionRange versionRange,
-            CancellationToken cancellationToken);
+            VersionRange? versionRange,
+            CancellationToken cancellationToken,
+            string? newMappingID = null,
+            string? newMappingSource = null);
```
zivkan: CA1068 usually asks for CancellationToken to be the last parameter of a method. I'm not sure why it's disabled in our repo.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -123,8 +123,8 @@ public async Task<RestoreResult> ExecuteAsync(CancellationToken token)
             {
                 telemetry.TelemetryEvent.AddPiiData(ProjectFilePath, _request.Project.FilePath);
 
-                bool isPackageSourceMappingEnabled = _request?.PackageSourceMapping.IsEnabled ?? false;
-                telemetry.TelemetryEvent[PackageSourceMappingIsMappingEnabled] = isPackageSourceMappingEnabled;
+                bool isPackageSourceMappingAlreadyEnabled = _request.PackageSourceMapping?.IsEnabled ?? false;
+                telemetry.TelemetryEvent[PackageSourceMappingIsMappingEnabled] = isPackageSourceMappingAlreadyEnabled;
```
zivkan: This variable name change doesn't make sense to me. Restore doesn't enable package source mapping half way through restore, so I expect it to be enabled or disabled throughout the entire operation. I don't understand what the "already" in the new variable name is trying to communicate.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
```diff
@@ -2844,7 +2848,9 @@ var projectUniqueNamesForBuildIntToUpdate
             IReadOnlyCollection<SourceRepository> primarySources,
             INuGetProjectContext nuGetProjectContext,
             VersionRange versionRange,
-            CancellationToken token)
+            CancellationToken token,
+            string newMappingID = null,
+            string newMappingSource = null)
```
zivkan: As mentioned a few times, CA1068 recommends cancellation tokens be the last parameter, which is a convention used throughout most (if not all?) libraries.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
zivkan: I very strongly discourage public APIs from using parameters with default values. Method overloads are, in my opinion, better in every way.

zivkan: Firstly, calling this API with the extra parameters doesn't appear to be in this PR, so I assume this PR is setting up the public API for other work which will call it.

However, I have two "bigger picture" thoughts:

1. If the customer is installing a package from a private feed, which has dependencies that come from a public feed, is this a scenario that's supposed to work?
2. Do we think in the future it will be possible to install more than 1 package in a single action?
    * Currently it's possible in PM UI's upgrade tab to update more than one package. But unless the new version changes dependencies, an update generally shouldn't need a different package source mapping.
    * But I'm wondering if a future improvement to PM UI might be to allow installing multiple packages "at once", to reduce waiting time (install package 1, wait for it to finish, install package 2, wait for it to finish).

I'm wondering if "needing more than 1 package source mapping" is an "obvious" future need, and we can avoid future public API changes if we do it now, even if we only call it with 1 package id/source per call initially.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
donnie-msft: It is used in this branch here (https://github.com/NuGet/NuGet.Client/blob/dev-donnie-msft-psmRestoreAndTests/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs#L546), although it's main purpose will be to serve as a XAML binding in the next PR (referred to in this PR's description).
Since I moved where the PackageSourceMappings are loading from settings, and since the telemetry event needs to use this flag, I wanted to start using this property now. I agree it is less necessary to use a public property in this PR so technically we could have a middle-ground commit that does the refactoring above without a new property, and then my next PR moves it to a new property. I like the idea of doing the change now since it's tangentially relevant.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
donnie-msft: Created a shiny brand new utilities class. See what you think.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
donnie-msft: Yes, I think I was retrieving existing mappings again in the save method, and technically those shouldn't be modified during the preview restore, so now I am passing them down into the save.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
donnie-msft: Discussed a bit offline. The Preview restore accepts 1 package ID. I'm not sure I see how it supports batching multiple actions or if this PR should be the thing that implements such batching.
Seeking more clarification...

jeffkl (dismissed):

src/NuGet.Core/NuGet.PackageManagement/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,3 @@
 #nullable enable
+~NuGet.PackageManagement.NuGetPackageManager.PreviewProjectsInstallPackageAsync(System.Collections.Generic.IReadOnlyCollection<NuGet.ProjectManagement.NuGetProject> nuGetProjects, NuGet.Packaging.Core.PackageIdentity packageIdentity, NuGet.PackageManagement.ResolutionContext resolutionContext, NuGet.ProjectManagement.INuGetProjectContext nuGetProjectContext, System.Collections.Generic.IReadOnlyCollection<NuGet.Protocol.Core.Types.SourceRepository> activeSources, NuGet.Versioning.VersionRange versionRange, System.Threading.CancellationToken token) -> System.Threading.Tasks.Task<System.Collections.Generic.IEnumerable<NuGet.PackageManagement.ResolvedAction>>
```
zivkan: This is the same API as was removed from the shipping txt file, so should be undeleted from there, rather than added here.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
zivkan: Github not showing me all the method parameters confused me. I thought this was one of the methods that took `List<NuGetAction>`. I know the method that has all the logic for install & uninstall is like that, so I made a bad assumption, sorry.

The first point, about a dependency that lives on a different package source stands. In fact, I'm now wondering if this feature will work for packages that have dependencies. If I try to install PackageA, which depende on PackageB, then this method will be called with PackageA, Source1 as parameters, but as far as I can tell, nothing will set up the source mapping for PackageB. Of course, NuGet won't know about PackageB until it resolves the graph, which happens as part of the preview restore.  Anyway, all I've managed to do is convince myself this is a difficult feature to automate in a GUI.

zivkan (dismissed):

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
donnie-msft: Yes, all of the dependencies will be source mapped. The UIActionEngine is taking the addedPackages, which is everything Restore deemed as necessary to add as part of this install. I then diff that with existing mappings and only write mappings that aren't currently satisfied by another mapping.

Added packages are analyzed here: https://github.com/NuGet/NuGet.Client/pull/5109/files/e93884166bf77bc495041788db16dc9efd9b687e#diff-ac2848ccdf455d1a83de39b70803b630e3b17f94534b18d15aa8fcf6c8191921R484

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
```diff
@@ -1596,6 +1617,8 @@ await PreviewBuildIntegratedProjectActionsAsync(buildIntegratedProject, actions,
             INuGetProjectContext nuGetProjectContext,
             IReadOnlyCollection<SourceRepository> activeSources,
             VersionRange versionRange,
+            string newMappingID,
+            string newMappingSource,
```
jebriede, line 1621: The newMappingID and newMappingSource pair are related. I propose the introduction of a new type that enforces the relationship between these two and encapsulates the logic of what happens when one is null and not the other?

Currently, I am not seeing a test that checks what will happen if only one of these strings is null and the other is not. But rather than just adding that test, extracting this into a type that either allows or disallows that and putting that type under test would isolate those rules in one place and allow it to be tested there.

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -0,0 +1,95 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using NuGet.Configuration;
+
+namespace NuGet.PackageManagement.UI
+{
+    internal static class PackageSourceMappingUtility
+    {
+        internal static void ConfigureNewPackageSourceMapping(
```
jebriede, line 15: Do we need this new static utility class? The method here mutates the state of the PackageSourceMappingProvider, so it would seem more appropriate for this functionality to exist on that class directly rather than having it in a separate class.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
jebriede: The UIActionEngine should not keep state of whether package source mapping is enabled. It also looks like this value is only set after a particular operation is performed, such as `PerformInstallOrUninstallAsync` which means its value is actually unknown until that operation and only updates after the next operation.

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMappingProvider.cs
```diff
@@ -16,6 +16,22 @@ public PackageSourceMappingProvider(ISettings settings)
...
+        public PackageSourceMappingProvider(ISettings settings, bool shouldSkipSave)
+            : this(settings)
+        {
+            ShouldSkipSave = shouldSkipSave;
+        }
+
+        /// <summary>
+        /// Avoid saving to disk and only modify the <see cref="ISettings"/> in memory.
+        /// </summary>
+        public bool ShouldSkipSave { get; private set; }
```
jebriede: The word "save" means different things in this class. In the method SavePackageSourceMappings, it seems to add items to a list, but not persist to disk.

This variable `ShouldSkipSave` reflects whether the settings should be persisted to disk. This inconsistency is confusing, especially when `SavePackageSourceMappings` is called in places where there are comments that say it is specifically not writing to disk. Consider updating ShouldSkipSave to use a different prefix such as `SkipPersisting` or `SkipWritingToDisk`

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMappingProvider.cs
```diff
@@ -44,7 +60,10 @@ internal void Remove(IReadOnlyList<PackageSourceMappingSourceItem> packageSource
                 catch { }
             }
 
-            _settings.SaveToDisk();
+            if (!ShouldSkipSave)
+            {
+                _settings.SaveToDisk();
+            }
```
jebriede, line 66: It looks like this `PackageSourceMappingProvider` takes in a ISettings and manipulates it for package source mapping, and decides when to write it to disk, and in some cases, skips the part about persisting to disk. A better design would be to have an implementation of ISettings that is the in-memory only representation of the settings and whose SaveToDisk method implementation no-ops.

This would also make it so we don't have to restore the settings' previous state after we've made modifications to it. We would make a copy that is an in-memory only representation but still implements ISettings, and pass that around, then throw it away when we're done with it and the settings we were passed remain untouched. It also would make it so we don't need the ShouldSkipSave on the PackageSourceMappingProvider and this class can function exactly the same way no matter what settings it was passed, without having to have deeper knowledge of whether it was supposed to persist the specific settings or not.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
```diff
@@ -2878,6 +2908,16 @@ var projectUniqueNamesForBuildIntToUpdate
             var result = new List<ResolvedAction>();
 
             var lockFileLookup = new Dictionary<string, LockFile>(PathUtility.GetStringComparerBasedOnOS());
+            PackageSourceMappingProvider packageSourceMappingProvider = null;
+            IReadOnlyList<PackageSourceMappingSourceItem> originalPackageSourceMappings = null;
+
+            if (newMappingID != null && newMappingSource != null)
```
jebriede, line 2914: As mentioned in another comment, this check could be handled by a type that relates these two parameters rather than checking them individually for null.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
```diff
@@ -2878,6 +2908,16 @@ var projectUniqueNamesForBuildIntToUpdate
...
+            {
+                packageSourceMappingProvider = new PackageSourceMappingProvider(Settings, shouldSkipSave: true);
```
jebriede, line 2916: If I'm understanding this correctly, it seems the responsibility of `PackageSourceMappingProvider` is to mutate the settings to add or remove details about Package Source Mappings based on supplied parameters about package source mappings and decide whether to persist the settings or not. If that is the case, the name PackageSourceMappingProvider doesn't seem to fit its responsibility. It also appears that the details about package source mapping items is actually provided by an outside source for most or all of this class' methods. I'd like to understand the overall intention of this class better before I could suggest an alternate name. Happy to chat offline or in the next PR review meeting.

jebriede (commented):
A lot of great work here! This is a big PR and I had to limit the time reviewing it. There are opportunities to encapsulate the functionality of the package source mapping code and separate responsibilities more cleanly. I left some comments about the design and am happy to go into more detail if you'd like to discuss it further offline in a chat or in our next PR review meeting.

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
donnie-msft: The reason I didn't put these UI related concepts into the `PackageSourceMappingProvider` is to keep a separation of concerns and to minimize the public API changes. I wouldn't want that provider knowing about things like `UserAction`.

It may be that the provider can have additional logic later, but I don't want to create public APIs until we know exactly what that provider's surface should look like.

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMappingProvider.cs
donnie-msft: Yes, this is how our providers work. They orchestrate an `ISettings`. The `ShouldSkipSave` pattern and terminology is also seen in the package source provider: https://github.com/NuGet/NuGet.Client/blob/63edda646848553467639abe6bbade5ebc5dec61/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs#L464

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
donnie-msft: Yes, our providers typically handle loading and manipulation of objects. For example, package sources have a provider here: https://github.com/NuGet/NuGet.Client/blob/63edda646848553467639abe6bbade5ebc5dec61/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs#L464

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
donnie-msft: > The newMappingID and newMappingSource pair are related. I propose the introduction of a new type that enforces the relationship between these two and encapsulates the logic of what happens when one is null and not the other?

I agree it could be more convenient to have a type, especially in the future if we want to support more configurable properties from the client. However, I think it's safe to iterate on this since the API isn't one that we ship, and as far as I know nothing else in VS depends on this API. In other words, we don't have to minimize our changes to this API right now, and when we have more interesting capabilities we want to support that warrant it, then we can create and swap this over to a new type.

> Currently, I am not seeing a test that checks what will happen if only one of these strings is null and the other is not.

The source code only uses them if both are not null. I'm not sure a separate test for one being null and not the other would be a useful test.

There's a test for this code path with both being `null` (which exercises the same code as only 1 being `null`): https://github.com/NuGet/NuGet.Client/blob/5e03c332991fcf02574f9e8b7a3eb7918aea91a5/test/NuGet.Core.Tests/NuGet.PackageManagement.Test/NuGetPackageManagerTests.cs#L7111-L7111

and there's a test for both being not null, ensuring it affects the restore summary:
https://github.com/NuGet/NuGet.Client/blob/5e03c332991fcf02574f9e8b7a3eb7918aea91a5/test/NuGet.Core.Tests/NuGet.PackageManagement.Test/NuGetPackageManagerTests.cs#L7182-L7182

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
donnie-msft: Addressed in https://github.com/NuGet/NuGet.Client/pull/5109#discussion_r1174063461

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMappingProvider.cs
donnie-msft: > This inconsistency is confusing, especially when SavePackageSourceMappings is called in places where there are comments that say it is specifically not writing to disk. 

I could only find one comment in this helper method, which is technically accurate, but I removed it based on your feedback as well as the fact that I'm not enforcing the helper to receive a provider which has `ShouldSkipSave==false`.
https://github.com/NuGet/NuGet.Client/blob/5e03c332991fcf02574f9e8b7a3eb7918aea91a5/src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs#L3190

> Consider updating ShouldSkipSave to use a different prefix such as SkipPersisting or SkipWritingToDisk

The nomenclature "save" does have multiple meanings in our repo. We save changes to an `ISettings` in memory, and we also `SaveToDisk`. At least we're explicit about saving to disk, since we want that to be obvious. I don't think there's a common understanding that "save" must always mean "to disk". If so, we should probably update the terms across the entire repo at some point. 
The naming I've used here was chosen to be consistent with other providers, such as https://github.com/NuGet/NuGet.Client/blob/63edda646848553467639abe6bbade5ebc5dec61/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs#L464

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -0,0 +1,95 @@
...
+            UserAction? userAction, List<Tuple<string, string>>? addedPackages,
```
dtivel:
```suggestion
            UserAction? userAction, IReadOnlyList<Tuple<string, string>>? addedPackages,
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -0,0 +1,95 @@
...
+            PackageSourceMappingProvider sourceMappingProvider,
+            IReadOnlyList<PackageSourceMappingSourceItem> existingPackageSourceMappingSourceItems)
+        {
+            if (userAction?.SourceMappingSourceName == null || addedPackages == null)
```
dtivel: This [`is null` pattern](https://learn.microsoft.com/en-US/dotnet/csharp/language-reference/operators/is) is great.  Consider using it instead of `== null`.  (There's also `is not null` instead of `!= null`.)

In most cases we write `== null` when we really mean `is null`.  See [this nice writeup](https://www.thomasclaudiushuber.com/2020/03/19/c-why-you-should-prefer-the-is-keyword-over-the-operator/).

You already do this elsewhere (e.g.:  [here](https://github.com/NuGet/NuGet.Client/pull/5109/files#diff-969dec071837d9b4d625b8920426d453e4ab75f07efec5c856b76392c8cbf39eR61) and [here](https://github.com/NuGet/NuGet.Client/pull/5109/files#diff-969dec071837d9b4d625b8920426d453e4ab75f07efec5c856b76392c8cbf39eR76)).  Search your PR for `== null`.

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
dtivel: Ideally, `Tuple<string, string>` would not leak from the [call site](https://github.com/NuGet/NuGet.Client/pull/5109/files#diff-ac2848ccdf455d1a83de39b70803b630e3b17f94534b18d15aa8fcf6c8191921R468) into this method.  I can't look at this parameter and understand it fully; what is `Item1` and what is `Item2`?  It looks like this method doesn't even use `Item2`.

I suggest reworking the parameter to be:

`IReadOnlyList<string> addedPackageIds`

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -0,0 +1,95 @@
...
+
+            Dictionary<string, IReadOnlyList<string>> patternsReadOnly = existingPackageSourceMappingSourceItems
+                .ToDictionary(pair => pair.Key, pair => (IReadOnlyList<string>)(pair.Patterns.Select(p => p.Pattern).ToList()));
+
+            PackageSourceMapping packageSourceMapping = new(patternsReadOnly);
+
+            // Expand all patterns/globs so we can later check if this package ID was already mapped.
+            IEnumerable<string> addedPackageIds = addedPackages.Select(action => action.Item1);
+            List<string> addedPackageIdsWithoutExistingMappings = new();
+            foreach (string? addedPackageId in addedPackageIds)
```
dtivel: I don't think `string?` is right.  It has been non-nullable up to now.  We'd get an [`ArgumentException` later](https://github.com/NuGet/NuGet.Client/blob/e3016693e01536996921d613152648d66eb0ce99/src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/SearchTree.cs#L85-L88) if it were nullable.

If you are sure it is nullable, then other annotations up until now are incorrect.  Like [this](https://github.com/NuGet/NuGet.Client/pull/5109/files#diff-969dec071837d9b4d625b8920426d453e4ab75f07efec5c856b76392c8cbf39eR31) and the `addedPackages` parameter, for instance.

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -0,0 +1,95 @@
...
+            {
+                IReadOnlyList<string> configuredSource = packageSourceMapping.GetConfiguredPackageSources(addedPackageId);
```
dtivel: `IReadOnlyList<string>?`

I'm guessing that from this and my previous comment that nullable warnings are not enabled.  Argument validation for the [`CreateAndSavePackageSourceMappings(...)`](https://github.com/NuGet/NuGet.Client/pull/5109/files#diff-969dec071837d9b4d625b8920426d453e4ab75f07efec5c856b76392c8cbf39eR55) method below is another example.

Meticulously ensure that your annotations are correct.

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -0,0 +1,95 @@
...
+                if (configuredSource == null || configuredSource.Count == 0)
+                {
+                    addedPackageIdsWithoutExistingMappings.Add(addedPackageId);
+                }
+            }
+
+            // Get all newly added package IDs that were not previously Source Mapped.
+            // Always include the Package ID being installed since it takes precedence over any globbing.
+            string[] packageIdsNeedingNewSourceMappings = addedPackageIdsWithoutExistingMappings
+                .Union(new string[] { userAction.PackageId })
```
dtivel:
```suggestion
                .Append(userAction.PackageId)
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -0,0 +1,95 @@
...
+                .ToArray();
+
+            CreateAndSavePackageSourceMappings(
+                sourceName: userAction.SourceMappingSourceName,
+                newPackageIdsToSourceMap: packageIdsNeedingNewSourceMappings,
+                sourceMappingProvider,
```
dtivel, line 51: The named parameters seem unnecessary.

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -0,0 +1,95 @@
...
+            }
+
+            IEnumerable<PackagePatternItem> newPackagePatternItems = newPackageIdsToSourceMap.Select(packageId => new PackagePatternItem(packageId));
+
+            List<PackageSourceMappingSourceItem> newAndExistingPackageSourceMappingItems = new(existingPackageSourceMappingSourceItems);
+
+            PackageSourceMappingSourceItem packageSourceMappingItemForSource =
+                        existingPackageSourceMappingSourceItems
+                        .Where(mappingItem => mappingItem.Key == sourceName)
+                        .FirstOrDefault();
```
dtivel: Fix indentation.

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -0,0 +1,95 @@
...
+                return;
+            }
+
+            IEnumerable<PackagePatternItem> newPackagePatternItems = newPackageIdsToSourceMap.Select(packageId => new PackagePatternItem(packageId));
+
+            List<PackageSourceMappingSourceItem> newAndExistingPackageSourceMappingItems = new(existingPackageSourceMappingSourceItems);
+
+            PackageSourceMappingSourceItem packageSourceMappingItemForSource =
+                        existingPackageSourceMappingSourceItems
+                        .Where(mappingItem => mappingItem.Key == sourceName)
```
dtivel: Is it possible for there to be differences in casing?

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
```diff
@@ -841,14 +849,16 @@ private PackageLicenseInfo GeneratePackageLicenseInfo(IPackageSearchMetadata met
             UserAction userAction,
             bool removeDependencies,
             bool forceRemove,
-            CancellationToken token)
+            CancellationToken token,
+            string? newMappingID = null,
+            string? newMappingSource = null)
```
dtivel: A `CancellationToken` parameter, if present, is typically last (like [here](https://github.com/NuGet/NuGet.Client/pull/5109/files#diff-77c5ac57c969eaa5ab522f95d16bbf96ab31f78a256489bd165a5b72c3500aeaR444-R446).

https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1068

src/NuGet.Clients/NuGet.PackageManagement.UI/UserAction.cs
```diff
@@ -24,20 +26,27 @@ private UserAction(NuGetProjectActionType action, string packageId, NuGetVersion
...
+        }
+
         public NuGetProjectActionType Action { get; private set; }
         public bool IsSolutionLevel { get; private set; }
         public ContractsItemFilter ActiveTab { get; private set; }
         public string PackageId { get; }
-        public NuGetVersion Version { get; }
-        public VersionRange VersionRange { get; }
+        public NuGetVersion? Version { get; }
+        public VersionRange? VersionRange { get; }
```
dtivel, line 46: Can you explain how/why each of these two can be `null` now?

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMappingProvider.cs
```diff
@@ -16,6 +16,22 @@ public PackageSourceMappingProvider(ISettings settings)
             _settings = settings ?? throw new ArgumentNullException(nameof(settings));
         }
 
+        /// <summary>
+        /// Supports the disabling of saving to disk for any <see cref="ISettings"/> changes.
+        /// </summary>
+        /// <param name="shouldSkipSave">True to avoid saving any changes to disk and only modify the <see cref="ISettings"/> in memory.
```
dtivel, line 22: Missing a `<param />` for `settings`.

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMappingProvider.cs
dtivel: This can be read-only?
```suggestion
        public bool ShouldSkipSave { get; }
```

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMappingProvider.cs
```diff
@@ -16,6 +16,22 @@ public PackageSourceMappingProvider(ISettings settings)
...
+        /// Default is false.</param>
+        public PackageSourceMappingProvider(ISettings settings, bool shouldSkipSave)
```
dtivel, line 24: Would `ShouldSave` (or `ShouldPersist`) with a default of `true` be better?  Turning off a feature (like saving) with `false` seems more intuitive than `true`.  `ShouldSkipSave` is basically "should not save", and the double negative default is yucky.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
```diff
@@ -3137,6 +3183,26 @@ var projectUniqueNamesForBuildIntToUpdate
             return result;
         }
 
+        /// <summary>
+        /// Reads existing Package Source Mappings from settings and appends a new mapping.
+        /// Additionally, a glob "*" pattern is appended for the the new mapping source.
+        /// The intention is that Preview Restore can run and expect all newly installed packages to be source mapped to the new source.
+        /// </summary>
+        /// <param name="settings">Reads existing Package Source Mappings, but does not write them.</param>
+        /// <returns>If a new mapping was provided, returns all persisted mappings appended with the new mapping. Otherwise, null.</returns>
```
dtivel: This parameter `settings` does not exist.  The method does not return anything.

Double check all documentation for correctness.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
```diff
@@ -3137,6 +3183,26 @@ var projectUniqueNamesForBuildIntToUpdate
...
+        private void AddNewPackageSourceMappingToSettings(string newMappingID, string newMappingSource, PackageSourceMappingProvider mappingProvider)
+        {
+            List<PackagePatternItem> newPatternItems = new()
+            {
+                new PackagePatternItem(newMappingID),
+                new PackagePatternItem("*")
+            };
+
+            var newAndExistingPackageSourceMappingItems = mappingProvider.GetPackageSourceMappingItems().ToList();
```
dtivel: Use explicit type instead of `var`.

test/NuGet.Clients.Tests/NuGet.PackageManagement.UI.Test/Actions/UIActionEngineTests.cs
```diff
@@ -140,6 +140,81 @@ public async Task GetPreviewResultsAsync_WithMultipleActions_SortsPackageIdentit
             Assert.Equal(packageIdentityC.Id, addedResults[2].Id);
         }
 
+        [Theory]
+        [MemberData(nameof(GetInstallActionTestData))]
+        public async Task CreateInstallAction_OnInstallingProject_WithNewSourceMapping_DoesNotLogTelemetry(ContractsItemFilter activeTab, bool isSolutionLevel, string packageIdToInstall, bool? expectedPkgWasTransitive)
```
dtivel:
```suggestion
        public async Task CreateInstallAction_OnInstallingProject_WithNewSourceMapping_DoesNotLogTelemetry(ContractsItemFilter activeTab, bool isSolutionLevel, string packageIdToInstall, bool? expectedPackageWasTransitive)
```

test/NuGet.Clients.Tests/NuGet.PackageManagement.UI.Test/Actions/UIActionEngineTests.cs
```diff
@@ -140,6 +140,81 @@ public async Task GetPreviewResultsAsync_WithMultipleActions_SortsPackageIdentit
...
+            TelemetryEvent lastTelemetryEvent = null;
+            telemetrySession
+                .Setup(x => x.PostEvent(It.IsAny<TelemetryEvent>()))
+                .Callback<TelemetryEvent>(x => lastTelemetryEvent = x);
+            var telemetryService = new NuGetVSTelemetryService(telemetrySession.Object);
+            TelemetryActivity.NuGetTelemetryService = telemetryService;
+
+            var sourceProvider = new Mock<ISourceRepositoryProvider>();
+            var settings = new Mock<ISettings>();
+            var nugetPM = new NuGetPackageManager(sourceProvider.Object, settings.Object, @"\packagesFolder");
```
dtivel, line 158: @"\packagesFolder" seems like a bad idea.  A single leading '\' usually means "from the root of the drive" on Windows.  A temporary directory or a local directory (@".\packgesFolder") seems better.

test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommandTests.cs
```diff
@@ -2207,6 +2209,172 @@ public async Task RestoreCommand_RestoreFloatingVersionWithIgnoreFailingHttpSour
...
+        {
+            // Arrange
+            using var pathContext = new SimpleTestPathContext();
+
+            var packageA = new SimpleTestPackageContext("a", "1.0.0");
+            var logger = new TestLogger();
+
+            PackageSpec project1Spec = ProjectTestHelpers.GetPackageSpec(projectName: "Project1",
+                                                                 rootPath: pathContext.SolutionRoot,
+                                                                 framework: "net5.0",
```
dtivel: Fix indentation.

dtivel (dismissed):
I didn't scrutinize tests... out of time.

jeffkl (dismissed):

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMappingProvider.cs
donnie-msft: I removed it since our policy is "if you don't add comment content for a param, then don't have a param element at all". Didn't have anything interesting to write for it.

src/NuGet.Clients/NuGet.PackageManagement.UI/UserAction.cs
donnie-msft: I enabled nullable for this class in this PR. 
It's possible to remove the nullable with some broader refactoring which I believe is outside the scope of my change. 

`VersionRange` is null (not set) by this constructor:
https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.UI/UserAction.cs#L12

`Version` is provided a null when uninstalling:
https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.UI/UserAction.cs#L62

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
donnie-msft: This is a great question. My assumption was that it should be explicitly the casing of the source requested.
I double-checked and fascinatingly, you **can** provide different casing of the same source name when hand-editing a nuget.config. I got no errors in VS from loading a config with nuget.org and NuGet.OrG both existing in the same config but point to different servers.

Using VS Options does show an error if you attempt to add a different casing of an existing source through the Options UI.

Therefore, since we don't have guarantees that there are unique names, I think this utility should enforce case checking as written..

test/NuGet.Clients.Tests/NuGet.PackageManagement.UI.Test/Actions/UIActionEngineTests.cs
donnie-msft: I tried the refactoring and got this error.
` System.ArgumentException : 'rootDirectory' must contain an absolute path '.\packagesFolder'.`

Since this test pattern was copied from an existing test in this class, I've filed this issue to find a way to refactor both of them in a separate PR: https://github.com/NuGet/Home/issues/12556

jeffkl (dismissed):

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMappingProvider.cs
donnie-msft: I got other feedback about the naming being confusing. We use this naming already in other providers, such as https://github.com/NuGet/NuGet.Client/blob/63edda646848553467639abe6bbade5ebc5dec61/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs#L464

I think it's worth a separate PR to address refactoring both providers if it's necessary. I don't know if the suggestion seems more intuitive to me, personally.

zivkan (approved):

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -0,0 +1,95 @@
...
+                .ToDictionary(pair => pair.Key, pair => (IReadOnlyList<string>)(pair.Patterns.Select(p => p.Pattern).ToList()));
+
+            PackageSourceMapping packageSourceMapping = new(patternsReadOnly);
+
+            // Expand all patterns/globs so we can later check if this package ID was already mapped.
+            List<string> addedPackageIdsWithoutExistingMappings = new();
+            foreach (string addedPackageId in addedPackageIds)
+            {
+                IReadOnlyList<string>? configuredSource = packageSourceMapping.GetConfiguredPackageSources(addedPackageId);
+                if (configuredSource is null || configuredSource.Count == 0)
```
nkolev92, line 36: The only thing I'd consider whether we should do differently is whether anything that gets brought in via a `*` should *actually* be explicitly mapped if direct. 

I don't think there's a good answer here without CD ofc :)

PR #5115, opened 2023-03-29
Add VulnerabilityInfo to NuGet.Protocol public API

### LINKED ISSUE #12518: Add VulnerabilityInfo APIs into NuGet.Protocol
### NuGet Product(s) Involved

NuGet SDK

### The Elevator Pitch

[The Vulnerabilities In Restore spec](https://github.com/NuGet/Home/blob/dev/proposed/2022/vulnerabilities-in-restore.md) introduces a new "resource" in the V3 protocol service index document.

There needs to be APIs added in NuGet.Protocol to download them, so that not only can the rest of NuGet download and use the data, but also other apps using the NuGet.Protocol package in their own apps can use it.

See the PR for the actual implementation, but the proposed API can be something along the lines of:

```cs
var vulnResource = await sourceRepository.GetResourceAsync<VulnerabilityInfoResource>();
IReadOnlyList<IDictionary<string, IReadOnlyList<VulnerabilityInfo>>> knownVulnerabilities =
    await vulnResource.GetVulnerabilityInfoAsync(...);
```

It's not the prettiest of APIs, specifically returning a list of dictionaries instead of a a single dictionary the the resource merges itself. However, it's going to be used in a perf-critical hot-path, and merging dictionaries will take more time than iterating multiple dictionaries, so here we are.

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12518

## Description

Implements VulnerabilityInfo resource as per the [vulnerabilities in restore spec](https://github.com/NuGet/Home/blob/dev/proposed/2022/vulnerabilities-in-restore.md).

* Added some model classes for JSON to be deserialized into
* Created a class that implements `IReadOnlyDictionary`, only to have a `JsonConverter` that allows us to create an instance of `Dictionary` with `StringComparer.InvariantIgnoreCase`.
* Created interface that all sources can (theoretically) implement for vulnerability data
* Created V3 HTTP resource and provider for the vulnerability data interface.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [x] Documentation PR or issue filled: https://github.com/NuGet/docs.microsoft.com-nuget/issues/3041
  - **OR**
  - [ ] N/A


zivkan:
## Public API review suggestions

Since we use Roslyn's Public API Analyzers, we can see all changes to public APIs in `PublicAPI.Shipped.txt` and `PublicAPI.Unshipped.txt`. For more info, see https://github.com/NuGet/NuGet.Client/blob/dev/docs/nuget-sdk.md.

In this PR, I added `#nullable enable` to some files that did not previously have them, so those are the changes we see to `PublicAPI.Shipped.txt` files. By all means validate, but none of these APIs should have changed, it's just adding nullable annotations. The new APIs are all in `PublicAPI.Unshipped.txt`

There are some new public APIs due to necessary NuGet.Protocol design, but the most important API in this PR is `IVulnerabilityInfoResource`. Basically everything else is just to support that API.

### `IVulnerabilityInfoResource`

It only has one method, `GetVulnerabilityInfoAsync`, which returns a `GetVulnerabilityInfoResult`. More on that after the sample.

The goal is for usage of the API to look something like this:

```csharp
SourceRepository source = GetSource();
IVulnerabilityInfoResource? vulnResource = await source.GetResourceAsync(IVulnerabilityInfoResource();
if (vulnResource == null)
{
  // this source does not have vulnerability info
  return null;
}

GetVulnerabilityInfoResult vulnData = await vulnResource.GetVulnerabilityInfoAsync(...);

IReadOnlyList<IDictionary<string, IReadOnlyList<VulnerabilityInfo>>>? knownVulnerabilities
    = vulnData.KnownVulnerabilities;
List<VulnerabilityInfo> packageVulnerabilities = new();
if (knownVulnerabilities != null)
{
    for (int file = 0; file < knownVulnerabilities.Count; file++)
    {
        if (knownVulnerabilities[file].TryGet(packageId, out IReadOnlyList<VulnerabilityInfo>? knownPackageVulnerabilities))
        {
            foreach (VulnerabilityInfo knownPackageVulnerability in knownPackageVulerabilities)
            {
                if (packageVersion.Satisfies(knownPackageVulnerability.Versions))
                {
                    packageVulnerabilities.Add(knownPackageVulnerability);
                }
            }
        }
    }
}

// Present package vulnerabilities however appropriate for the current tool

if (vulnData.Errors?.Count > 0)
{
    // Getting vulnerability data from source was not completely successful.
    // Present errors as appropriate
}
```

The big nested loop in the middle, that checks all files and checks each known vulnerability if the package version being checked satisfies the vulnerability data range, might become a public helper method in another assembly, but it doesn't feel like it belongs in NuGet.Protocol. To be honest, I don't know which assembly such a public API should belong in. So maybe by the time the feature is complete it will be in NuGet.Protocol. But filtering vulerabilities based on package ID and version is not related to "communication with NuGet feeds", hence it doesn't feel like the right place currently.

#### Why an interface

(Almost?) All of NuGet.Protocol's other resources are abstract classes. Why not follow that design, and put common code that all implementations (HTTP and in the future local folder source) will need?

Firstly, I'm a believer in [composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance). While well intentioned, I don't believe that inheritance is a good solution to code sharing. Helper functions are better, in my opinion, and when necessary, use delegate parameters. C# 3.5 added `Func` and `Action` which makes it much easier to use. Although in perf-critical code it's best to avoid using (anonymous) lamba functions, since that causes allocations. Where possible, it's best to pass in a method name with a matching signature.

More importantly, I'm also a string believer in writing testable code. While NuGet.Protocol.Tests and NuGet.Protocol.FuncTests does test a lot of NuGet.Protocol's code, if you read the code and read the test infrastructure, I think it's clear that it's not easy. For example, look at, and debug while stepping into every method call, anything that uses `StaticHttpSource`. The intention is to use a dictionary where the key is the URL and the value is the response payload. However, it runs a whole lot of production code, including needing workarounds to avoid the http cache possibly returning a payload from a different test. Not to mention that `NuGet.Protocol.FuncTests` takes multiple minutes to execute, despite a small number of tests. By making the base resource, which everything outside of NuGet.Protocol uses, an interface, it allows tests to mock the return code, avoiding a whole lot of test infrastructure, and making the whole thing faster to execute.

#### Why not report failures with `ILogger`

For performance reasons, it is intended to have a shared context object get the vulerability data from all sources just once. My assumption might turn out to be wrong by the time I get to using these APIs in `RestoreCommand`, but I believe that reporting errors and warnings with `ILogger` will cause the message to be reported to the first project that calls the API. However, if a package source's vulnerability data can't be parsed, this means the vulnerability checking on all projects is incomplete. Therefore, any warnings or errors related to incomplete vulnerability checking applies to all projects. By using a list of `Exception`s, it allows the per-project restore code to add errors to that project, meaning next restore will fail no-op checks, and will therefore try to recheck the vulnerability data, which will hopefully succeed in a following attempt.

### `GetVulnerabilityInfoResult`

This type was created to avoid using a `Tuple` or `ValueTuple`. While `ValueTuple` can use the syntax `(int Thing, string Whatever)`, so the tuple's items have "real" names, rather than `Item1`, and `Item2`, since the two items types have long definitions, it wouldn't be easy to use without `var`. While we could [add a `Deconstruct` method](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/deconstruct#user-defined-types), I'd rather wait and see when it's useful, rather than preemptively adding another public API that might not be useful.

Another thing to note (and it looks like I need to fill in the XMLDoc) is that either of the two properties can be null. If there's some kind of failure that prevents any of the vulnerability files from being parsed, then the `KnownVulerabilities` will be null. If there are no errors, then `Exceptions` will be null. If there are some exceptions, but some files were successfully parsed, then neither will be null, and the caller is expected to still understand that this is an error situation, but one that allows degraded functionality.

#### Why not have `IVulnerabilityInfoResource.GetVulnerabilityInfoAsync` throw when any error happens?

Since the design spec of the feature explicitly recommends that package sources split their vulnerability data into large, periodically updated "base" files, and a second frequently updated file, we encourage sources to provide multiple files. In case a server has error, for example the vulnerability index points to a missing file, or a file that is in the process of being updated and therefore the client gets a JSON file that can't be parsed correctly, it allows the client to run in a degraded state, where it can still check vulnerabilities for data is successfully downloaded, before providing appropriate error handling.

Think of it as a kind of "try-get", but since async methods can't use `out` parameters, a new data type is being used instead of a tuple.

#### Why not merge multiple files into a single dictionary

`IVulnerabilityInfoResult.KnownVulnerabilities` is a list of dictionaries. I'll agree that this feels like a leaky abstraction. However, I believe this may be a case where it's warranted because this code is on a known hot path. It's not on the "hottest path" (no-op restore), but any time no-op fails, then this code will be run, even if no packages need to be downloaded. Since speed or performance is often one of the biggest complaint areas from customer when we run surveys, we know it's important to customers that scenarios like this are fast. While my own testing/benchmarking can't possibly cover 100% of all scenarios, and I am making assumptions about how this not-yet-released API is going to be used in the future, benchmarking makes it appear that the time to merge the multiple dictionaries into a single dictionary, or reading the vulnerabilities files serially into a single dictionary, rather than in parallel into multiple dictionaries, is slower than having the vulnerability checking code iterate over multiple dictionaries.

### `VulerabilityInfoResourceV3Provider`

This type is public because `SourceRepository` is designed to be extensible, so someone using NuGet.Protocol can either filter out, or inject their own providers when they create their own `SourceRepository`. Whether they use `GetCoreV3().Where(p => !(p.Value is VulnerabilityInfoResourceV3Provider`, or they create their own provider and pass `nameof(VulnerabilityInfoResourceV3Provider)` to the `beforeProviders` argument, both cases need the type to be public. Well, technically the second doesn't, but if it wasn't public they'd have to provide a string without allowing the compiler to validate the correct name for them.

### `VulnerabilityInfo`

This is a 1:1 mapping of the deepest object within the vulnerability data file that package sources need to provide to the client. While I was originally concerned about this being a leaky abstraction, in the end I thought it was the ideal data structure. Clients need the URL and severity to show to customers. And the versions list could be useful as well. Imaging something like VS's Package Manager UI, where you get the package details of a package, and it not only tells you that it has a known vulnerability, but it also tells you the version range the vulerability applies to, so you can know in advance what the lowest version with a fix is, so you don't need to check every version individually to see when the known vulnerability information stops.

kartheekp-ms:
> GetVulnerabilityInfoAsync, which returns a GetVulnerabilityInfoResult

I am curious to understand why not return `VulnerabilityInfoResult` instead of `GetVulnerabilityInfoResult`

kartheekp-ms:
> The big nested loop in the middle, that checks all files and checks each known vulnerability if the package version being checked satisfies the vulnerability data range, might become a public helper method in another assembly, but it doesn't feel like it belongs in NuGet.Protocol. 

May be the decision is already made, have we considered an option to download package vulnerabilities just like package versions by appending package name to the registration base URL for example https://pkgs.dev.azure.com/dnceng/9ee6d478-d288-47f7-aacc-f6e6d082ae6d/_packaging/45bacae2-5efb-47c8-91e5-8ec20c22b4f8/nuget/v3/flat2/nuget.common/index.json. I don't have the performance metrics yet, but I think if the vulnerabilities database size increases in NuGet.org then performance may degrade.

loic-sharma:
> Since the design spec of the feature explicitly...

Is this design spec public? Would you mind sharing it?

EDIT: Here it is: https://github.com/NuGet/Home/blob/72a2e173e1c7d82880f21b8574a81874eeb15ead/proposed/2022/vulnerabilities-in-restore.md

loic-sharma:
This looks great! Thanks for the excellent write up to make this easy to review, it's very appreciated. 

Did you consider making `VulnerabilityInfoResourceV3.GetVulnerabilityFilesAsync` and `VulnerabilityInfoResourceV3.GetVulnerabilityDataAsync` public and part of `IVulnerabilityInfoResourceV3`? Scenarios where this might be useful:

1. I want a tool that notifies me whenever nuget.org updates its vulnerability data
2. I know of nuget.org's files and only want to ingest the `update` file without ever reading the `base` file.

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
```diff
@@ -0,0 +1,168 @@
...
+                {
+                    knownVulnerabilities = new();
+                    return new GetVulnerabilityInfoResult(knownVulnerabilities, exceptions);
+                }
+
+                var tasks = new Task<IReadOnlyDictionary<string, IReadOnlyList<VulnerabilityInfo>>>[indexEntries.Count];
+                for (int i = 0; i < tasks.Length; i++)
+                {
+                    V3VulnerabilityIndexEntry indexEntry = indexEntries[i];
+                    tasks[i] = GetVulnerabilityDataAsync(indexEntry, cacheContext, logger, cancellationToken);
```
loic-sharma, line 160: I would clamp this to an upper bound to avoid flooding the thread pool with tasks if nuget.org ever decides to have thousands of files.

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
```diff
@@ -0,0 +1,168 @@
...
+            Uri uri,
+            string name,
+            SourceCacheContext cacheContext,
+            ILogger logger,
+            CancellationToken cancellationToken)
+        {
+            HttpSourceResource httpSourceResource = await _sourceRepository.GetResourceAsync<HttpSourceResource>(cancellationToken);
+
+            HttpSourceCacheContext httpSourceCacheContext = HttpSourceCacheContext.Create(cacheContext, isFirstAttempt: true);
+            var request = new HttpSourceCachedRequest(uri.OriginalString, "vuln_data_" + name, httpSourceCacheContext);
```
loic-sharma: I'm a bit concerned using `@name` for the cache key. Per the spec:

> @name, a user friendly name for the page, must be unique.

Does NuGet's caching normalize the key? Should we add limits to the `@name` to ensure it cannot break NuGet's cache key logic?

kartheekp-ms:
> I know of nuget.org's files and only want to ingest the update file without ever reading the base file.

I am not sure if this is possible because as per https://github.com/NuGet/Home/blob/dev/proposed/2022/vulnerabilities-in-restore.md#dedicated-vulnerability-information-resource, it is possible to update the base file also.
```
The vulnerability resource contains 2 pages.
One page, base, represents the data up to a certain point in time.
The second page, update, represents the data from the last update of base, to present.
Periodically, for example once a month, the data from update should be migrated to base.
If an entry needs to be removed from base, base should be updated.
```

erdembayar:
> The big nested loop in the middle, that checks all files and checks each known vulnerability if the package version being checked satisfies the vulnerability data range,

If performance is a concern here then have you considered other data structure types? Perhaps a `Trie` data structure could be an alternative instead of a `big nested loop`. We have already implemented one for [Package Source mapping lookup](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/SearchNode.cs) by pattern.  Actually, it could be further improved by using string compression. I haven't had time to implement this, and so far there have been no reported performance concerns regarding that.
The only high cost would be the initial `Trie` tree building, and then the lookup would be cheap. Still tbh, without actual perf test it's hard to say what would be perf difference, it depends on parameters like load size and number of packages etc.

kartheekp-ms:
> If performance is a concern here then have you considered other data structure types? Perhaps a `Trie` data structure could be an alternative instead of a `big nested loop`.

In my opinion, the Trie data structure may not be appropriate for this scenario as our objective is not to match prefixes, as in the case of a package source mapping feature.

src/NuGet.Core/NuGet.Protocol/Model/GetVulnerabilityInfoResult.cs
```diff
@@ -0,0 +1,34 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace NuGet.Protocol.Model
+{
+    /// <summary>Represents the result of <see cref="IVulnerabilityInfoResource.GetVulnerabilityInfoAsync(Core.Types.SourceCacheContext, Common.ILogger, System.Threading.CancellationToken)"/></summary>
+    /// <remarks>Vulnerability data from a package source typically is spread across multiple files. Therefore, it's
+    /// possible for some data to be retrieved successfully, while other data failed. This type allows partial data
+    /// to be returned along with error information.</remarks>
+    public class GetVulnerabilityInfoResult
```
zivkan: From https://github.com/NuGet/NuGet.Client/pull/5115#issuecomment-1491164057, since GitHub mysteriously doesn't supported threaded/resolvable conversations unless it's linked to a specific line in a file

> I am curious to understand why not return VulnerabilityInfoResult instead of GetVulnerabilityInfoResult

@kartheekp-ms my thought process was that the data type for the payload data (list of dictionary) is very specific to one API and might not be interesting to other APIs. So, rather than using an overly generic name for a specific data type, give a specific name.  Although there's a good chance I was over-thinking it.

I also experimented with:

``` cs
public class ApiResult<T>
{
    public T? Result { get; }
    public IReadOnlyList<Exception>? Exceptions { get; }
}
```

However, I had some problem trying to get it to work. I no longer remember exactly what problems, possibly something to do with Newtonsoft.Json deserialization.

src/NuGet.Core/NuGet.Protocol/Model/GetVulnerabilityInfoResult.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        /// <summary></summary>
+        public IReadOnlyList<IReadOnlyDictionary<string, IReadOnlyList<VulnerabilityInfo>>>? KnownVulnerabilities { get; }
+        public IReadOnlyList<Exception>? Exceptions { get; }
```
zivkan: It occurs to me now that perhaps this should be `AggregateException` instead.  It contains a list internally.  Thoughts?

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
```diff
@@ -0,0 +1,168 @@
...
+                    vulnerabilityIndexUrl.OriginalString);
+                throw new FatalProtocolException(message);
+            }
+
+            return vulnFiles;
+
+            async Task<Uri> GetIndexUrlAsync(CancellationToken cancellationToken)
+            {
+                ServiceIndexResourceV3 serviceIndex = await _sourceRepository.GetResourceAsync<ServiceIndexResourceV3>(cancellationToken);
+                return serviceIndex.GetServiceEntryUri(ServiceTypes.VulnerabilityInfo);
```
zivkan, line 79: From https://github.com/NuGet/NuGet.Client/pull/5115#issuecomment-1491185896 since GitHub doesn't provide threaded/resolvable conversations unless linked to a specific line in a file:

> May be the decision is already made, have we considered an option to download package vulnerabilities just like package versions by appending package name to the registration base URL for example https://pkgs.dev.azure.com/dnceng/9ee6d478-d288-47f7-aacc-f6e6d082ae6d/_packaging/45bacae2-5efb-47c8-91e5-8ec20c22b4f8/nuget/v3/flat2/nuget.common/index.json. I don't have the performance metrics yet, but I think if the vulnerabilities database size increases in NuGet.org then performance may degrade.

@kartheekp-ms the design spec considered different options, and also for different considerations/use-cases, like performance: https://github.com/NuGet/Home/blob/dev/proposed/2022/vulnerabilities-in-restore.md

Basically, even if the file(s) keep getting larger and larger over time, the latency penalty of making many requests far outweighs any reduced size benefit. Consider the performance of `dotnet list package --vulnerable --include-transitive`. Adding that much delay into restore is unacceptable. It needs to be much, much faster.  But also it wouldn't meet some other design goals outlined in the spec.

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
```diff
@@ -0,0 +1,168 @@
...
+            }
+        }
+
+        internal async Task<IReadOnlyDictionary<string, IReadOnlyList<VulnerabilityInfo>>> GetVulnerabilityDataAsync(
```
zivkan: From https://github.com/NuGet/NuGet.Client/pull/5115#issuecomment-1492470821 since GitHub doesn't support threaded/resolvable conversations unless linked to a specific line of code

> Did you consider making `VulnerabilityInfoResourceV3.GetVulnerabilityFilesAsync` and `VulnerabilityInfoResourceV3.GetVulnerabilityDataAsync` public and part of `IVulnerabilityInfoResourceV3`? Scenarios where this might be useful:
> 1. I want a tool that notifies me whenever nuget.org updates its vulnerability data
> 2. I know of nuget.org's files and only want to ingest the update file without ever reading the base file.

@loic-sharma excellent scenarios!  I was losing time overthinking how to design the public APIs, so I tried to just get stuff done by focusing on the absolute minimum APIs required. Now I need to open this can of worms again 😆. As we discussed years ago, we both wish that NuGet.Protocol had "low level" APIs that make the "raw" protocol available to advanced SDK users, and have separate "high level" APIs that make it easier to use.

I don't want to make the APIs public on the "high level" API. We also haven't yet designed how vulnerability data will work for local sources, or some other kind of "download vulnerability data from a server, and use your local copy during a restore". So, I don't know if "there could be multiple files" is going to be a NuGet v3 HTTP specific implementation detail, or if it'll be common for all APIs that deal with these vulnerability data files.

You probably remember, but NuGet.Client has been impacted before by having too many public APIs exposing what should be internal implementation details, and when we try to fix bugs (perf or behaviour), and broke those APIs. Later we get feedback that it affected someone because they were using those APIs. Sometimes our insertions into the .NET SDK broke because the .NET SDK itself was using the APIs. So, now I'm being overly cautious to not cause future problems when someone needs to refactor what is hopefully internal details. But APIs that mimic the HTTP NuGet protocol aren't really internal, so I'll have a think about how to make it public without making too much public.

src/NuGet.Core/NuGet.Protocol/Model/GetVulnerabilityInfoResult.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        public GetVulnerabilityInfoResult(
```
zivkan, line 32: Since GitHub doesn't provide threaded/resolvable conversations unless assigned to a specific line of code:

From @erdembayar in https://github.com/NuGet/NuGet.Client/pull/5115#issuecomment-1492755715

> If performance is a concern here then have you considered other data structure types? Perhaps a Trie data structure could be an alternative instead of a big nested loop. We have already implemented one for [Package Source mapping lookup](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/SearchNode.cs) by pattern. Actually, it could be further improved by using string compression. I haven't had time to implement this, and so far there have been no reported performance concerns regarding that.
The only high cost would be the initial Trie tree building, and then the lookup would be cheap. Still tbh, without actual perf test it's hard to say what would be perf difference, it depends on parameters like load size and number of packages etc.

From @kartheekp-ms in https://github.com/NuGet/NuGet.Client/pull/5115#issuecomment-1492769962

> In my opinion, the Trie data structure may not be appropriate for this scenario as our objective is not to match prefixes, as in the case of a package source mapping feature.

As I wrote in [the PR review suggestions](https://github.com/NuGet/NuGet.Client/pull/5115#issuecomment-1488678765), under a header titled "Why not merge multiple files into a single dictionary"

> benchmarking makes it appear that the time to merge the multiple dictionaries into a single dictionary, or reading the vulnerabilities files serially into a single dictionary, rather than in parallel into multiple dictionaries, is slower than having the vulnerability checking code iterate over multiple dictionaries.

Putting it another way, the perf-bottleneck is merge the data into a single data structure, not making many lookups for a single package ID.

erdembayar: > Putting it another way, the perf-bottleneck is merge the data into a single data structure, not making many lookups for a single package ID.

I agree with that. Actually, building the initial tree for trie is expensive, but lookups would be quite cheap. The reason I propose using trie is that, if I'm not mistaken, for a large set of package IDs it would save space since the size does not grow linearly with the number of items in it. Therefore, it solves the problem of vulnerability data growing over time.

If the internal implementation is abstracted away and changing it wouldn't cause a public API change later, then I'm okay with the current implementation.

zivkan: It is something that can change later, but I feel like it's missing the point. Synthetic benchmarks so far suggest that loading the file takes more time than lookups. So, if we make lookups even faster, by making loading even slower, then it's going to make the total duration worse, not better. We can validate once the real implementation is complete to validate that the synthetic benchmarks were representative of real performance.

src/NuGet.Core/NuGet.Protocol/Model/IgnoreCaseStringKeyDictionary.cs
```diff
@@ -0,0 +1,45 @@
...
+        public int Count => _inner.Count;
+
+        public bool ContainsKey(string key) => _inner.ContainsKey(key);
+
+        public IEnumerator<KeyValuePair<string, IReadOnlyList<VulnerabilityInfo>>> GetEnumerator() => _inner.GetEnumerator();
+
+        public bool TryGetValue(string key, out IReadOnlyList<VulnerabilityInfo> value) => _inner.TryGetValue(key, out value);
+
+        IEnumerator IEnumerable.GetEnumerator() => _inner.GetEnumerator();
+    }
```
jeffkl:
```suggestion
    internal class VulnerabilityFileData : Dictionary<string, IReadOnlyList<VulnerabilityInfo>>, IReadOnlyDictionary<string, IReadOnlyList<VulnerabilityInfo>>
    {
        public VulnerabilityFileData()
          : base(StringComparer.OrdinalIgnoreCase)
        {
        }
    }
```

Not 100% sure this will work but you can try having this class inherit from `Dictionary<string, IReadOnlyList<VulnerabilityInfo>>` then call the base constructor with `StringComparer.OrdinalIgnoreCase` making this class always be a case-insensitive dictionary and not needing to have an inner dictionary.  That could also mean that you would not need a custom JSON converter?

src/NuGet.Core/NuGet.Protocol/Model/GetVulnerabilityInfoResult.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// <summary>The known package vulnerabilities.</summary>
+        /// <remarks>The outer IReadOnlyList represents the number of files the package source split the vulnerability data into.
+        /// The IReadOnlyDictionary's key is the package ID.
+        /// The inner IReadOnlyList is all of the vulnerabilities known for the package ID.
+        /// <para>May be null if no vulnerability data was parsed successfully.</para></remarks>
+        public IReadOnlyList<IReadOnlyDictionary<string, IReadOnlyList<VulnerabilityInfo>>>? KnownVulnerabilities { get; }
+
+        /// <summary>Any exceptions while reading vulnerability data.</summary>
+        /// <remarks>May be null if no exceptions occurred.</remarks>
+        public AggregateException? Exceptions { get; }
```
jeffkl, line 26: Does it make sense for this to return the base `Exception` class instead of `AggregateException`?  It could still be an `AggregateException` for the caller but it seems odd to return a concrete type?  What do you think?

src/NuGet.Core/NuGet.Protocol/Converters/IgnoreCaseStringKeyDictionaryConverter.cs
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+#nullable enable
```
jeffkl: +100 points for doing `#nullable enable` for all this 😄

src/NuGet.Core/NuGet.Protocol/Converters/IgnoreCaseStringKeyDictionaryConverter.cs
```diff
@@ -0,0 +1,81 @@
...
+            var data = (VulnerabilityFileData?)value;
+
+            if (data != null)
+            {
+                writer.WriteStartObject();
+
+                foreach (var kvp in data.OrderBy(kvp => kvp.Key, StringComparer.InvariantCultureIgnoreCase))
+                {
+#pragma warning disable CA1308 // Normalize strings to uppercase - NuGet has been using lower ids for years
+                    string lowerKey = kvp.Key.ToLowerInvariant();
```
jeffkl: If the API endpoint only returned lowercase strings, this would not be needed.  Is that something worth considering?  I guess there'd be no guarantee that the server implementation abides by that...

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
```diff
@@ -0,0 +1,205 @@
...
+            IReadOnlyList<V3VulnerabilityIndexEntry>? vulnFiles;
+            try
+            {
+                vulnFiles = await httpSourceResource.HttpSource.GetAsync(request,
+                    result =>
+                {
+                    using (var textReader = new StreamReader(result.Stream))
+                    using (var jsonReader = new JsonTextReader(textReader))
+                    {
+                        var parsed = JsonExtensions.JsonObjectSerializer.Deserialize<IReadOnlyList<V3VulnerabilityIndexEntry>>(jsonReader);
```
jeffkl: Do you know if this call will have any effect on Visual Studio?  Will we end up scheduling this on a background thread and not have to worry about it?  Just curious the implications on VS.

src/NuGet.Core/NuGet.Protocol/Model/GetVulnerabilityInfoResult.cs
zivkan: `AggregateException` contains a list of exceptions. Previously I had the type as `IReadOnlyList<Exception>`, but thought maybe `AggregateException` might be "better" since it's a built-in type for basically the same purpose.

I think that returning the "real" exception when there's only 1, and returning AggregateException when there are multiple is more confusing, because then it ends up like JavaScript/PowerShell "is there one or more than one" code everywhere since you never really know what you're going to get.  An array of 1 allows code to always work the same way, and maybe sometimes it'll only iterate a single item.

Having written all that, since you're asking, clearly it's not obvious.  Should I change it back to `IReadOnlyList<Exception>`?  Is all my justification unconvincing and you think I should still use `Exception` anyway?

src/NuGet.Core/NuGet.Protocol/Converters/IgnoreCaseStringKeyDictionaryConverter.cs
zivkan: This is in the serialization code, not the deserialization code.  NuGet.Client will never use it, but other people might find it useful (maybe the server team in the future? it might be useful to other server implementors who want to generate vulnerability data for their implementation).

What I'm thinking is they use NuGet.Packaging to parse a bunch of nupkg files, whose nuspec files will use "package preferred" casing, not lowercase. They then populate this data structure and then use `string json = JsonConverter.Serialize(data)`. In order to comply with NuGet's protocol spec, it then needs to lowercase the package ids (and version strings, though that's only relevant for versions with pre-release strings, and it's also a violation of the SemVer spec, but that ship has long sailed).

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
zivkan: Are you referring to the fact that it's not async, or something else?

Ideally NuGet.Protocol should never run on the UI thread. But NuGet.Protocol is a "generic" library that runs in, and ships with, CLI msbuild and the .NET SDK. Therefore, we can't use the vs-threading library and `ThreadHelper.ThrowWhenOnUIThread()`. If you have an idea of some other way we can easily detect it (and future regressions), that'll be great (I'm thinking something around ETW events and a post-test analyzer, similar to VS's DDRITs/RPS tests, but this involves a lot of infrastructure NuGet doesn't currently have). But ASP.NET (non-core) has thread-affinity, so we probably don't want to throw any time the task's synchronization context is non-null, because that'll break anyone who might be using NuGet.Protocol in an ASP.NET app (well, it'll force them to be more explicit about switching between the thread pool, and the appropriate synchronization context).

Anyway, Newtonsoft.Json doesn't have an async deserialize method. The only async code I'm aware of is JsonTextReader has a `NextAsync`, but that means we'd have to convert everything to custom converters (and not classes implementing `JsonConverter`). It'd be easier to switch to System.Text.Json instead, although since that only supports UTF-8, there's theoretically a risk of breaking any server that serves any other text encoding. All I can do now is throw my hands in the air 🤷‍♂️

If you're not referring to async, then you need to explain what specific concerns about VS you have.  But it's not fundamentally deserializing JSON differently to how the rest of NuGet.Protocol is already deserializing JSON. Well, it is, most of NuGet.Protocl is much more awful, first converting to JObject, and then deserializing, but that's a pointless performance waste.

src/NuGet.Core/NuGet.Protocol/Model/IgnoreCaseStringKeyDictionary.cs
zivkan: This worked well, thanks!

I also made a slight change, making it a generic type `CaseInsenstiveDictionary<TValue>`, so theoretically it could be re-used if it ends up being useful, but we no longer need our own JsonConverter.

src/NuGet.Core/NuGet.Protocol/Model/GetVulnerabilityInfoResult.cs
jeffkl: I don't think it would make sense for this to return a list of `Exception` objects.  I'm only saying it could return `Exception` and the caller could check if its an `AggregateException`.  Otherwise our code has to always wrap exceptions in an `AggregateException`.  I don't see anywhere else in our code where we return an `AggregateException`, only cases where we return `Exception`.
```
src\NuGet.Core\NuGet.PackageManagement\IDE\PackageRestoreFailedEventArgs.cs:        public Exception Exception { get; private set; }
src\NuGet.Core\NuGet.Protocol\Plugins\FaultedPluginEventArgs.cs:        public Exception Exception { get; }
src\NuGet.Core\NuGet.Protocol\Plugins\PluginCreationResult.cs:        public Exception Exception { get; }
src\NuGet.Core\NuGet.Protocol\Plugins\ProtocolErrorEventArgs.cs:        public Exception Exception { get; }
```

In my opinion, it would be best to return a more abstract type instead of a concrete type and is a more standard pattern.  The XML documentation can indicate to callers that it could be an `AggregateException`.

This would mean that `GetVulnerabilityInfoAsync` would need to still keep track of the exceptions but if only one was thrown return that, otherwise return an `AggregateException`.  We could also just keep its implementation and this property is just always an `Exception` but current implementation is really always an `AggregateException`.  Just my vote.

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
```diff
@@ -0,0 +1,299 @@
...
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                AddException(ref exceptions, ex);
+            }
+
+            GetVulnerabilityInfoResult result = new(knownVulnerabilities, ToAggregateException(exceptions));
+            return result;
```
jeffkl, line 193: Okay now I have another thought, what if this method just threw the exception at the end and the `GetVulnerabilityInfoResult` doesn't have a property indicating the exceptions.  Wouldn't that be a bit more aligned with what exceptions are for?  The caller would use `try...catch` instead of getting back an object and checking a property.  What do you think?

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
```diff
@@ -0,0 +1,299 @@
...
+                const int maxNameLength = 32;
+                if (name.Length > maxNameLength)
+                {
+                    string message = string.Format("Vulnerability page index {0} name is longer than {1} characters (Source: '{2}')",
+                        i, maxNameLength, _sourceRepository.PackageSource.Name);
+                    AddException(new FatalProtocolException(message), ref exceptions);
+                    continue;
+                }
+
+                bool hasOnlyValidChars = true;
```
jeffkl, line 235: nit: this could be a local or static function to be a little cleaner

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
zivkan: I added some limits to prevent malicious (or just careless) servers from causing clients too many problems. Here's a PR to update the spec: https://github.com/NuGet/Home/pull/12529#discussion_r1159438144

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
zivkan: APIs are public now.

src/NuGet.Core/NuGet.Protocol/Model/GetVulnerabilityInfoResult.cs
zivkan: > I'm only saying it could return Exception and the caller could check if its an AggregateException.

And this is what I'm referring to that frustrates developers with PowerShell (and maybe also Javascript?)  when a method returns more than 1 items, the return type is an array, but when it returns exactly one item, it's not an array. For example the version numbering problem we had a few weeks ago when MSBuild started adding empty lines to its output, and `output.Trim()` in PowerShell silently trimmed each line individually (since it was treated as an array), rather than trimming the new line characters out of we expected to be a single string.

It sounds easy to "check if it's an AggregateException", but so is "check if the PowerShell method returned an array or a single item". Not everyone will read the docs, and when it behaves differently to the simple tests people tried while developing, bugs will happen.  Good APIs have few surprises, and that was my intention here. It's designed to potentially have more than 1 error, so the data type represents that.

[The BCL's Task.Exception](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.exception?view=net-7.0#system-threading-tasks-task-exception) is a kind of precedent. It has type `AggregateException`, even though a task usually represents a single code path execution, not a `Task.WhenAll` or similar.

> Otherwise our code has to always wrap exceptions in an AggregateException

I don't follow. Why would other code, that have different design goals, need to wrap all exceptions in an AggregateException?

>  I don't see anywhere else in our code where we return an AggregateException, only cases where we return Exception.

And I'm trying to do something different to basically anything else already in the NuGet repo:  graceful degradation when there's a problem. The closest precedent I can think of is when a package version can't be found or downloaded during the first stage of graph resolution. It doesn't instantly crash the whole restore process, it still allows other packages to be restored, but the project will error out since part of the graph couldn't be determined. However, a failed package doesn't have the same "partial data" concept that vulnerability data does (multiple pages). Other examples of graceful error handling are the second stage of graph resolution, or pack validations, but in both of those cases, they're types of validation on existing data, its not simultaneously creating new data types to return, at the same time as detecting errors gracefully.

If there's a problem with one vulnerability page in a server, I still want to use the successful pages to check for known vulnerabilities. If one server's vulnerability data is completely broken, I still want to use different server's known vulnerabilities.

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
zivkan: As explained in [another comment I just posted](https://github.com/NuGet/NuGet.Client/pull/5115#discussion_r1159487887), I want this to degrade gracefully when errors occur.  If a server reports 5 pages of vulnerability data, but only one of them has an error, I think it's still useful (and important) to use the 4 good pages of data to check if restored packages have known vulnerabilities.

I will still fail the restore, but if one of the good pages has a known vulnerability on a package the project references, I don't think we should hide that information from the customer just because some different file on the server has a JSON error.

src/NuGet.Core/NuGet.Protocol/Model/GetVulnerabilityInfoResult.cs
jeffkl: Okay I suppose this is fine, I don't feel that strongly about it.  It seems weird but it does fit the implementation where an exception isn't thrown and potentially incomplete results are returned.

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
jeffkl: I was just thinking that most other calls aren't potentially huge and just wanted to be sure that this moderately intense operation won't cause UI blocking in VS.  I'm not familiar enough with JTF to know, just wanted to get your thoughts.  Sounds like we're good.

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
jeffkl: Okay I obviously missed the design of returning potentially incomplete results.

jeffkl (dismissed):
Great work!

src/NuGet.Core/NuGet.Protocol/Model/GetVulnerabilityInfoResult.cs
nkolev92, line 26: Another way of "replicating" the exception would've been to internally put the call in a lazy task that gets awaited. 

https://github.com/NuGet/NuGet.Client/blob/91e9a3db594160975170dc226646af7b6c4a50bb/src/NuGet.Core/NuGet.Commands/RestoreCommand/SourceRepositoryDependencyProvider.cs#L223

That way you're not putting the errors in a result.

src/NuGet.Core/NuGet.Protocol/Model/VulnerabilityInfo.cs
```diff
@@ -0,0 +1,75 @@
...
+
+using System;
+using Newtonsoft.Json;
+using NuGet.Shared;
+using NuGet.Versioning;
+
+namespace NuGet.Protocol.Model
+{
+    /// <summary>A class that represents one known vulnerability of a package.</summary>
+    public sealed class VulnerabilityInfo : IEquatable<VulnerabilityInfo>
```
nkolev92: Should this be a more specific name, such as PackageVulnerabilityInfo, as vulnerability info is pretty generic.

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
```diff
@@ -0,0 +1,299 @@
...
+            _sourceRepository = sourceRepository ?? throw new ArgumentNullException(nameof(sourceRepository));
+        }
+
+        /// <summary>Get the vulnerability pages the server contains.</summary>
+        /// <param name="cacheContext">The cache settings to use when making HTTP requests.</param>
+        /// <param name="log">The logger for any messages.</param>
+        /// <param name="cancellationToken">The cancellation token to cancel operation.</param>
+        /// <returns>The list of vulnerability data pages the server claims to have.</returns>
+        /// <exception cref="FatalProtocolException">When various HTTP or deserialization exceptions occur.</exception>
+        public async Task<IReadOnlyList<V3VulnerabilityIndexEntry>> GetVulnerabilityFilesAsync(SourceCacheContext cacheContext, ILogger log, CancellationToken cancellationToken)
```
nkolev92, line 38: I'm guessing this being public is a consequence of some of the earlier discussions?

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
nkolev92: I think any `WhenAll` http communication should have an upper bound. This one is a sneaky/unobvious one.

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
```diff
@@ -0,0 +1,299 @@
...
+            {
+                IReadOnlyList<V3VulnerabilityIndexEntry> indexEntries = await GetVulnerabilityFilesAsync(cacheContext, logger, cancellationToken);
+
+                if (indexEntries.Count == 0)
+                {
+                    knownVulnerabilities = new();
+                    return new GetVulnerabilityInfoResult(knownVulnerabilities, ToAggregateException(exceptions));
+                }
+
+                var tasks = new Task<IReadOnlyDictionary<string, IReadOnlyList<VulnerabilityInfo>>>[indexEntries.Count];
```
nkolev92: How do we account for an data file being already on disk an up to date?
I'm not sure I see that handled anywhere. Aren't we re-downloading the files constantly now?

src/NuGet.Core/NuGet.Protocol/ServiceTypes.cs
```diff
@@ -27,5 +28,6 @@ public static class ServiceTypes
         public static readonly string[] PackageBaseAddress = { "PackageBaseAddress" + Versioned, "PackageBaseAddress" + Version300 };
         public static readonly string[] RepositorySignatures = { "RepositorySignatures" + Version500, "RepositorySignatures" + Version490, "RepositorySignatures" + Version470 };
         public static readonly string[] SymbolPackagePublish = { "SymbolPackagePublish" + Version490 };
+        internal static readonly string[] VulnerabilityInfo = { "VulnerabilityInfo" + Version660 };
```
nkolev92: Your spec has 6.7.0.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/Resources/VulnerabilityInfoResourceV3Tests.cs
```diff
@@ -0,0 +1,139 @@
...
+using NuGet.Protocol.Model;
+using NuGet.Protocol.Resources;
+using NuGet.Protocol.Tests.Providers;
+using NuGet.Test.Utility;
+using Test.Utility;
+using Xunit;
+
+namespace NuGet.Protocol.Tests.Resources
+{
+    public class VulnerabilityInfoResourceV3Tests
```
nkolev92, line 22: We could use a few more tests (can be in a future PR). 

- GetVulnerabilityDataAsync 
  - Fetches data from all index entries (testing more than 1).
  - Multiple calls use http cache.
  - Multiple calls when 1 file gets updated only fetches the updated file.

nkolev92 (commented):
Looks great! Some great tests. 

I don't see any handling of individual file updates. Is that intentional?

src/NuGet.Core/NuGet.Protocol/Model/IgnoreCaseDictionary.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace NuGet.Protocol.Model
+{
+    internal class IgnoreCaseDictionary<TValue> : Dictionary<string, TValue>
```
jebriede: A class name that would avoid using the verb "Ignore" in its name would be:

```suggestion
    internal class CaseInsensitiveDictionary<TValue> : Dictionary<string, TValue>
```

test/NuGet.Core.Tests/NuGet.Protocol.Tests/Converters/VulnerabilityFileDataConverterTests.cs
```diff
@@ -0,0 +1,33 @@
...
+            string json = @"{
+  ""packageid"": []
+}";
+
+            // Act
+            var result = JsonConvert.DeserializeObject<IgnoreCaseDictionary<IReadOnlyList<VulnerabilityInfo>>>(json);
+
+            // Assert
+            result.Should().NotBeNull();
+            result!.ContainsKey("packageid").Should().BeTrue();
```
jebriede, line 29: I notice the "!" [null-forgiving operator](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving) here but not in the other assertions using `result`. Is it needed here?

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
```diff
@@ -0,0 +1,299 @@
...
+                vulnFiles = await httpSourceResource.HttpSource.GetAsync(request,
+                    result =>
+                {
+                    using (var textReader = new StreamReader(result.Stream))
+                    using (var jsonReader = new JsonTextReader(textReader))
+                    {
+                        var parsed = JsonExtensions.JsonObjectSerializer.Deserialize<IReadOnlyList<V3VulnerabilityIndexEntry>>(jsonReader);
+                        return Task.FromResult(parsed);
+                    }
+                },
```
jebriede: Needs further indentation since this code block is related to the result parameter and not vulnFiles.
```suggestion
                    {
                        using (var textReader = new StreamReader(result.Stream))
                        using (var jsonReader = new JsonTextReader(textReader))
                        {
                            var parsed = JsonExtensions.JsonObjectSerializer.Deserialize<IReadOnlyList<V3VulnerabilityIndexEntry>>(jsonReader);
                            return Task.FromResult(parsed);
                        }
                    },
```

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
```diff
@@ -0,0 +1,299 @@
...
+                V3VulnerabilityIndexEntry entry = indexEntries[i];
+
+                string name = entry.Name;
+                if (string.IsNullOrWhiteSpace(name))
+                {
+                    string message = string.Format("Vulnerability page index {0} does not contain a name (Source: '{1}')",
+                        i, _sourceRepository.PackageSource.Name);
+                    AddException(new FatalProtocolException(message), ref exceptions);
+                    continue;
+                }
```
jebriede: This method has a series of validations that follow the same pattern:
Test a predicate. If true, build a message, throw an exception, and continue.

We can reduce code duplication and make it easier to read by making a method or object for the repeated validation pattern.

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
```diff
@@ -0,0 +1,299 @@
...
+            return result;
+
+            static void AddException(ref List<Exception>? exceptions, Exception ex)
+            {
+                if (exceptions == null)
+                {
+                    exceptions = new();
+                }
+                exceptions.Add(ex);
+            }
```
jebriede: There are two definitions of AddException with different order of parameters. Consider making this just one static method definition on the class.

jebriede (dismissed):
Approving with suggestions! 👍

src/NuGet.Core/NuGet.Protocol/Model/GetVulnerabilityInfoResult.cs
zivkan: I'm sorry, I don't really follow.

Perhaps it's a similar misunderstand to what Jeff had.  I want this code to handle graceful degradation when some vulnerability pages are successful, while others failed. I'm intentionally avoiding the "this method either succeeds and returns a valid result, or it throws an exception". It's what I tried to explain in this class's XMLdoc.

If that's not what you mean by your comment, I just don't understand.

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
zivkan: yes: https://github.com/NuGet/NuGet.Client/pull/5115#discussion_r1155728138

Basically, people who maintain their own servers, or server implementations, can use these APIs to mirror another source's vulnerability data in their own instance.

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
zivkan: It will use the cache settings on the passed-in `SourceCacheContext`. This is a perf improvement, but not a feature correctness issue, so will be implemented later when there aren't more pressing things to work on: https://github.com/NuGet/Client.Engineering/issues/2244

test/NuGet.Core.Tests/NuGet.Protocol.Tests/Converters/VulnerabilityFileDataConverterTests.cs
zivkan: The compiler knows that once a variable is not null, it cannot be null without having a new value assigned. So, the `!` operator isn't so much forgiving one statement to say it's allowed to be null, it's basically saying "Treat this value as being definitely not-null".

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
zivkan: I put in a basic limit of 16 for now. A better throttle so that the max limit isn't all done concurrently can be a perf optimization later.

src/NuGet.Core/NuGet.Protocol/Model/GetVulnerabilityInfoResult.cs
nkolev92: I'm ok with your implementation, I don't love AggregateException as a prop in a result file, but I get your intention.

src/NuGet.Core/NuGet.Protocol/Model/V3VulnerabilityIndexEntry.cs
```diff
@@ -0,0 +1,40 @@
...
+#nullable enable
+
+using System;
+using Newtonsoft.Json;
+
+namespace NuGet.Protocol.Model
+{
+    /// <summary>Represents one page of vulnerability data.</summary>
+    /// <remarks>More details at <a href="https://learn.microsoft.com/nuget/api/vulnerability-info-resource">https://learn.microsoft.com/nuget/api/vulnerability-info-resource</a></remarks>
+    public sealed class V3VulnerabilityIndexEntry
```
nkolev92, line 13: nit: V3 in the name is probably unnecessary. 
Or if it is, I think it should be at the end instead of the beginning.

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
nkolev92: My comment was about the parallelization degree instead of the total count of pages actually.

nkolev92 (approved):

jeffkl (approved):

PR #5117, opened 2023-03-29
Fix transitive dependencies cache problems

### LINKED ISSUE #12333: The feature “Transitive Package” doesn’t work in the “Installed” tab until reopening the .NET SDK style project
### NuGet Product Used

### Product Version

NuGet Client Dev\6.5.0.135

### Worked before?

_No response_

### Impact

##### Repro Steps:  

1. Set the NUGET_TRANSITIVE_DEPENDENCIES_IN_PMUI environment variable to 1. 

2. Create a .NET Core Console App project via VS.  
3. Right-click the project in Solution Explorer and select ‘Manage NuGet Packages…’ menu item to open PM UI.  
 4. Click “Browse” tab, search a package (e.g. “Microsoft.EntityFrameworkCore”) that has dependencies, select a package version “6.0.4” in the "Version" drop-down list in the right panel and click the “Install” button on the right.  
 5. Click “OK” button in the “Preview Changes” dialog and click “I Accept” button in the “License Acceptance” dialog.  
 6. Click “Installed” tab. 

##### Expected:  

The feature “Transitive Package” works in the “Installed” tab. “Transitive packages” label is expanded by default and all the transitive packages showed in the “Transitive packages” list (under the “Top-level packages” list). 

##### Actual:  

The feature “Transitive Package” doesn’t work in the “Installed” tab until reopening the .NET SDK style project as below. 
MicrosoftTeams-image

##### Note: 

1. The feature “Transitive Package” will work in the “Installed” tab after reopening the .NET SDK style project and PM UI. 
2. Started to repro from Dev\6.5.0.117.

### Verbose Logs

_No response_

### LINKED ISSUE #12360: The count of the transitive packages never changes in the “Installed” tab after installing transitive packages
### NuGet Product Used

### Product Version

NuGet Client Dev\6.5.0.140

### Worked before?

_No response_

### Impact

#### Repro Steps:
1. Create a .NET Core Console App project via VS.  
2. Right-click the project in Solution Explorer and select ‘Manage NuGet Packages…’ menu item to open PM UI.  
3. Click “Browse” tab, search a package (e.g. “Microsoft.EntityFrameworkCore”) that has dependencies, select a package version (e.g. 6.0.4) in the "Version" drop-down list in the right panel and click the “Install” button on the right.  
4. Click “OK” button in the “Preview Changes” dialog and click “I Accept” button in the “License Acceptance” dialog.
5. Switch to the “Installed” tab, and Install a transitive package “Microsoft.Extensions.Logging” in the package list.

#### Expected:
The count of the transitive packages should show correctly (it should be 12 for this example).

#### Actual:
The count of the transitive packages never changes in the “Installed” tab (it showed 13 for this example) as below screenshot.
image
#### Notes:
1. The count never changes no matter how many transitive packages are installed.
2. It started from Dev\6.5.0.129.

### Verbose Logs

_No response_

### LINKED ISSUE #12428: [Bug Bash] The transitive packages will not disappear in the “Installed” tab after uninstalling the top-level package
### NuGet Product Used

### Product Version

Dev\6.6.0.19

### Worked before?

_No response_

### Impact

#### Repro Steps:
1. Create a C# Console App (.NET Core 7.0) project in VS.
2. Right-Click the project in the Solution Explorer window and select ‘Manage NuGet Packages…’.
3. Click “Browse” tab, search a package (e.g. “Microsoft.EntityFrameworkCore”) that has dependencies and click the “Install” button (↓) on the right.
4. Click “OK” button in the “Preview Changes” dialog and click “I Accept” button in the “License Acceptance” dialog.
5. Click “Installed” tab and observe the package list.
6. Uninstall the top-level package (e.g. “Microsoft.EntityFrameworkCore”) and observe the package list in “Installed” tab.

#### Expected:
The transitive packages would disappear in the “Installed” tab.

#### Actual:
The transitive packages will not disappear in the “Installed” tab as below.
image

#### Notes:
It doesn't repro on Dev\6.5.0.107, repro on Dev\6.5.0.127.

martinrrm:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12333
Fixes: https://github.com/NuGet/Home/issues/12360
Fixes: https://github.com/NuGet/Home/issues/12428

Regression? Last working version: Regression introduced by https://github.com/NuGet/NuGet.Client/commit/e7ea3a13d3fc4738c375f2a48306d0cc46207e36

## Description

This PR fixed multiple bugs, those can be resumed in that transitive dependencies where not displaying correctly in the PM UI when you installed or uninstalled packages or transitive packages, this happened because the cache was not cleared correctly.  

The main problem is that the function `GetInstalledAndTransitivePackagesAsync` handles a lot of states using booleans, this function can return only Installed packages and not Transitives even though the name says otherwise. This causes that we cache wrong information when using `IsInstalledAndTransitiveComputationNeeded`.  GIATPA is called a lot of times when using the PM UI, sometimes there are multiple threads, so that complicates even more the handling of the booleans and its states.

GIATPA = GetInstalledAndTransitivePackagesAsync
IsComputationNeeded = IsInstalledAndTransitiveComputationNeeded

`IsComputationNeeded ` can tell us if there was a change in the assets file, and before this PR if we don't request for transitive data in GIATPA `IsComputationNeeded` will be set to false and transitive data won't be calculated, if we later ask GIATPA to return transitive and `IsComputationNeeded` will be false and returning wrong cached data. 

In this PR when `IsComputationNeeded` is `true`, we will clear all cached data from transitive since we don't know if there was a change that affected transitive data, this will force the calculation of transitive packages and origins when explicitly requested in GIATPA and the cache is clear. 

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
@@ -134,14 +134,13 @@ namespace NuGet.PackageManagement.VisualStudio
             IList<LockFileTarget> targetsList = null;
             T installedPackages;
             T transitivePackages = default;
+            Dictionary<string, TransitiveEntry> transitiveOrigins = default;
             if (IsInstalledAndTransitiveComputationNeeded)
             {
                 // clear the transitive packages cache, since we don't know when a dependency has been removed
                 installedPackages = new T();
-                if (includeTransitivePackages)
-                {
-                    transitivePackages = new T();
-                }
+                transitivePackages = new T();
+                transitiveOrigins = new Dictionary<string, TransitiveEntry>();
                 targetsList = await GetTargetsListAsync(assetsFilePath, token);
             }
             else
@@ -188,6 +187,11 @@ namespace NuGet.PackageManagement.VisualStudio
             IEnumerable<PackageReference> calculatedTransitivePackages = Enumerable.Empty<PackageReference>();
             if (includeTransitivePackages)
             {
+                if (targetsList == null)
+                {
+                    targetsList = await GetTargetsListAsync(assetsFilePath, token);
+                }
+
                 calculatedTransitivePackages = packageSpec
                     .TargetFrameworks
                     .SelectMany(f => ResolvedTransitivePackagesList(f.FrameworkName, targetsList, installedPackages, transitivePackages))
@@ -202,7 +206,6 @@ namespace NuGet.PackageManagement.VisualStudio
                 if (includeTransitiveOrigins && await ExperimentUtility.IsTransitiveOriginExpEnabled.GetValueAsync(token))
                 {
                     // Compute Transitive Origins
-                    Dictionary<string, TransitiveEntry> transitiveOrigins;
                     if (IsInstalledAndTransitiveComputationNeeded // Cache invalidation
                         || TransitiveOriginsCache == null // If any data race left the cache as null
                         || (!TransitiveOriginsCache.Any() && calculatedTransitivePackages.Any())) // We have transitive packages, but no transitive origins and the call is requesting transitive origins
@@ -254,12 +257,13 @@ namespace NuGet.PackageManagement.VisualStudio
             {
                 InstalledPackages = installedPackages;
             }
-            if (includeTransitivePackages)
+            lock (_transitiveOriginsLock)
             {
-                lock (_installedAndTransitivePackagesLock)
-                {
-                    TransitivePackages = transitivePackages;
-                }
+                TransitiveOriginsCache = transitiveOrigins;
+            }
+            lock (_installedAndTransitivePackagesLock)
+            {
+                TransitivePackages = transitivePackages;
             }
 
             IsInstalledAndTransitiveComputationNeeded = false;
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
@@ -257,13 +257,18 @@ namespace NuGet.PackageManagement.VisualStudio
             {
                 InstalledPackages = installedPackages;
             }
-            lock (_transitiveOriginsLock)
+            // if includeTransitivePackages, update the cache with the new transitive packages information
+            // or if IsInstalledAndTransitiveComputationNeeded, clear the transitive packages cache, since we don't know when a dependency has been removed
+            if (includeTransitivePackages || IsInstalledAndTransitiveComputationNeeded)
             {
-                TransitiveOriginsCache = transitiveOrigins;
-            }
-            lock (_installedAndTransitivePackagesLock)
-            {
-                TransitivePackages = transitivePackages;
+                lock (_transitiveOriginsLock)
+                {
+                    TransitiveOriginsCache = transitiveOrigins;
+                }
+                lock (_installedAndTransitivePackagesLock)
+                {
+                    TransitivePackages = transitivePackages;
+                }
             }
 
             IsInstalledAndTransitiveComputationNeeded = false;
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
@@ -163,6 +163,7 @@ namespace NuGet.PackageManagement.VisualStudio
                     if (TransitivePackages == null)
                     {
                         transitivePackages = new T();
+                        transitiveOrigins = new Dictionary<string, TransitiveEntry>();
                     }
                     else
                     {
--- a/test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
+++ b/test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
@@ -4198,8 +4198,6 @@ namespace NuGet.PackageManagement.VisualStudio.Test
 
             // do a restore
             project = PrepareCpsRestoredProject(initialProjectSpec, cache);
-
-            var installedPackages = await project.GetInstalledPackagesAsync(CancellationToken.None); // This will change IsInstalledAndTransitiveComputationNeeded to false
             result = await project.GetInstalledAndTransitivePackagesAsync(false, false, CancellationToken.None);
 
             // Assert I: Cache for transitive was cleared
@@ -4207,7 +4205,6 @@ namespace NuGet.PackageManagement.VisualStudio.Test
             Assert.Equal(0, result.TransitivePackages.Count);
             Assert.All(result.TransitivePackages, pkg => Assert.Equal(0, pkg.TransitiveOrigins.Count()));
 
-            project = PrepareCpsRestoredProject(initialProjectSpec, cache);
             result = await project.GetInstalledAndTransitivePackagesAsync(true, true, CancellationToken.None);
 
             // Assert II: Transitive packages calculated correctly
--- a/test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
+++ b/test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
@@ -4127,11 +4127,11 @@ namespace NuGet.PackageManagement.VisualStudio.Test
         }
 
         [Theory]
-        [InlineData(false, false)]
-        [InlineData(false, true)]
-        [InlineData(true, false)]
-        [InlineData(true, true)]
-        public async Task GetInstalledAndTransitivePackagesAsync_IsInstalledAndTransitiveComputationNeeded_False_Succeed(bool includeTransitivePackages, bool includeTransitiveOrigins)
+        [InlineData(false, false, 0, 0)]
+        [InlineData(false, true, 0, 0)]
+        [InlineData(true, false, 1, 0)]
+        [InlineData(true, true, 1, 1)]
+        public async Task GetInstalledAndTransitivePackagesAsync_InstalledAndTransitiveComputationNotNeeded_ReturnsExpectedTransitiveData(bool includeTransitivePackages, bool includeTransitiveOrigins, int transitivePackagesCount, int transitiveOriginsCount)
         {
             using var rootDir = new SimpleTestPathContext();
             await CreatePackagesAsync(rootDir, packageAVersion: "1.0.0");
@@ -4142,27 +4142,24 @@ namespace NuGet.PackageManagement.VisualStudio.Test
             await RestorePackageSpecsAsync(rootDir, output: null, initialProjectSpec);
 
             CpsPackageReferenceProject project = PrepareCpsRestoredProject(initialProjectSpec, cache);
+            var installedPackagesBefore = await project.GetInstalledPackagesAsync(CancellationToken.None); // This will change IsInstalledAndTransitiveComputationNeeded to false
 
             // Act
-            var installedPackages = await project.GetInstalledPackagesAsync(CancellationToken.None); // This will change IsInstalledAndTransitiveComputationNeeded to false
-            ProjectPackages result = await project.GetInstalledAndTransitivePackagesAsync(includeTransitivePackages, includeTransitiveOrigins, CancellationToken.None);
+            ProjectPackages installedAndTransitivePackagesAfter = await project.GetInstalledAndTransitivePackagesAsync(includeTransitivePackages, includeTransitiveOrigins, CancellationToken.None);
 
             // Assert
-            var transitivePackagesCount = includeTransitivePackages ? 1 : 0;
-            var transitiveOriginsCount = (includeTransitivePackages && includeTransitiveOrigins) ? 1 : 0;
-
-            Assert.Equal(result.InstalledPackages.Count, installedPackages.Count());
-            Assert.Equal(1, result.InstalledPackages.Count);
-            Assert.Equal(transitivePackagesCount, result.TransitivePackages.Count);
-            Assert.All(result.TransitivePackages, pkg => Assert.Equal(transitiveOriginsCount, pkg.TransitiveOrigins.Count()));
+            Assert.Equal(installedAndTransitivePackagesAfter.InstalledPackages.Count, installedPackagesBefore.Count());
+            Assert.Equal(1, installedAndTransitivePackagesAfter.InstalledPackages.Count);
+            Assert.Equal(transitivePackagesCount, installedAndTransitivePackagesAfter.TransitivePackages.Count);
+            Assert.All(installedAndTransitivePackagesAfter.TransitivePackages, pkg => Assert.Equal(transitiveOriginsCount, pkg.TransitiveOrigins.Count()));
         }
 
         [Theory]
-        [InlineData(false, false)]
-        [InlineData(false, true)]
-        [InlineData(true, false)]
-        [InlineData(true, true)]
-        public async Task GetInstalledAndTransitivePackagesAsync_IsInstalledAndTransitiveComputationNeeded_True_Succeed(bool includeTransitivePackages, bool includeTransitiveOrigins)
+        [InlineData(false, false, 0, 0)]
+        [InlineData(false, true, 0, 0)]
+        [InlineData(true, false, 1, 0)]
+        [InlineData(true, true, 1, 1)]
+        public async Task GetInstalledAndTransitivePackagesAsync_IsInstalledAndTransitiveComputationNeeded_ReturnsExpectedTransitiveData(bool includeTransitivePackages, bool includeTransitiveOrigins, int transitivePackagesCount, int transitiveOriginsCount)
         {
             using var rootDir = new SimpleTestPathContext();
             await CreatePackagesAsync(rootDir, packageAVersion: "1.0.0");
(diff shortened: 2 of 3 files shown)


src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
```diff
@@ -188,6 +187,11 @@ internal async Task<ProjectPackages> GetInstalledAndTransitivePackagesAsync(bool
             IEnumerable<PackageReference> calculatedTransitivePackages = Enumerable.Empty<PackageReference>();
             if (includeTransitivePackages)
             {
+                if (targetsList == null)
+                {
+                    targetsList = await GetTargetsListAsync(assetsFilePath, token);
```
martinrrm, line 193: `targetsList` was only calculated when `IsComputationNeeded` was true and we need to update the list when getting transitives, for example when opening the Browse tab, there is no computation needed but transitive are required for the recommender

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
```diff
@@ -3900,7 +3900,6 @@ public async Task GetTransitivePackagesAsync_WithTransitivePackageReferences_Ret
                 await result.CommitAsync(logger, CancellationToken.None);
                 ProjectPackages packages = await project.GetInstalledAndTransitivePackagesAsync(includeTransitiveOrigins: false, CancellationToken.None);
                 DateTime lastWriteTime = File.GetLastWriteTimeUtc(lockFilePath);
-                File.WriteAllText(lockFilePath, "** replaced file content to test cache **");
```
martinrrm, line 3903: This was causing that the assets file couldn't be read because it is an invalid JSON and there is a new call to load the JSON in https://github.com/NuGet/NuGet.Client/pull/5117/files#diff-cf69ebf676e61580349108d38e5dab1bbff91468194c6a8d342149d264139b15R192

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
```diff
@@ -4127,6 +4126,107 @@ public async Task GetInstalledAndTransitivePackagesAsync_TransitiveOriginsParamS
             Assert.All(packages.TransitivePackages, pkg => Assert.NotEmpty(pkg.TransitiveOrigins));
         }
 
+        [Theory]
+        [InlineData(false, false)]
+        [InlineData(false, true)]
+        [InlineData(true, false)]
+        [InlineData(true, true)]
+        public async Task GetInstalledAndTransitivePackagesAsync_IsInstalledAndTransitiveComputationNeeded_False_Succeed(bool includeTransitivePackages, bool includeTransitiveOrigins)
```
jebriede: The suffix of this test method is "Succeed" but that doesn't tell me what is supposed to happen. Can you follow the pattern for naming tests here: https://github.com/NuGet/NuGet.Client/blob/dev/docs/coding-guidelines.md#unit-test-method-naming

The last part should be "with what expectations"

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
```diff
@@ -4127,6 +4126,107 @@ public async Task GetInstalledAndTransitivePackagesAsync_TransitiveOriginsParamS
...
+            IProjectSystemCache cache = new ProjectSystemCache();
+
+            // Arrange
+            PackageSpec initialProjectSpec = ProjectTestHelpers.GetPackageSpec("MyProject", rootDir.SolutionRoot, framework: "net472", dependencyName: "PackageA");
+            await RestorePackageSpecsAsync(rootDir, output: null, initialProjectSpec);
+
+            CpsPackageReferenceProject project = PrepareCpsRestoredProject(initialProjectSpec, cache);
+
+            // Act
+            var installedPackages = await project.GetInstalledPackagesAsync(CancellationToken.None); // This will change IsInstalledAndTransitiveComputationNeeded to false
```
jebriede: Nit: this looks like part of the Arrange step if it's setting up the test.

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
```diff
@@ -4127,6 +4126,107 @@ public async Task GetInstalledAndTransitivePackagesAsync_TransitiveOriginsParamS
...
+            ProjectPackages result = await project.GetInstalledAndTransitivePackagesAsync(includeTransitivePackages, includeTransitiveOrigins, CancellationToken.None);
+
+            // Assert
+            var transitivePackagesCount = includeTransitivePackages ? 1 : 0;
+            var transitiveOriginsCount = (includeTransitivePackages && includeTransitiveOrigins) ? 1 : 0;
```
jebriede: Conditionals in a test makes it harder to read the test and reason what is expected, but most importantly, add complexity to tests that make the test code more vulnerable to bugs so conditionals should be avoided in tests.

The expected conditions for the tests should be included as parameters in the InlineData above so they are constant and not based on logical operations.

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
```diff
@@ -4127,6 +4126,107 @@ public async Task GetInstalledAndTransitivePackagesAsync_TransitiveOriginsParamS
...
+
+            Assert.Equal(result.InstalledPackages.Count, installedPackages.Count());
```
jebriede: Can we have a variable with something descriptive for these installed package counts? For example, something like "installedPackagesBefore" and "installedPackagesAfter". It's possible I'm not even understanding the comparison being made here so feel free to choose a different name, but one that describes what the first and second collections are for would make this assertion clearer.

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
```diff
@@ -4127,6 +4126,107 @@ public async Task GetInstalledAndTransitivePackagesAsync_TransitiveOriginsParamS
...
+            Assert.Equal(1, result.InstalledPackages.Count);
+            Assert.Equal(transitivePackagesCount, result.TransitivePackages.Count);
```
jebriede: Same as with installedPackages above.

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
```diff
@@ -4127,6 +4126,107 @@ public async Task GetInstalledAndTransitivePackagesAsync_TransitiveOriginsParamS
...
+            Assert.All(result.TransitivePackages, pkg => Assert.Equal(transitiveOriginsCount, pkg.TransitiveOrigins.Count()));
+        }
+
+        [Theory]
+        [InlineData(false, false)]
+        [InlineData(false, true)]
+        [InlineData(true, false)]
+        [InlineData(true, true)]
+        public async Task GetInstalledAndTransitivePackagesAsync_IsInstalledAndTransitiveComputationNeeded_True_Succeed(bool includeTransitivePackages, bool includeTransitiveOrigins)
```
jebriede: Same feedback as above test applies to all tests in the PR.

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
```diff
@@ -4127,6 +4126,107 @@ public async Task GetInstalledAndTransitivePackagesAsync_TransitiveOriginsParamS
...
+            var transitivePackagesCount = includeTransitivePackages ? 1 : 0;
+            var transitiveOriginsCount = (includeTransitivePackages && includeTransitiveOrigins) ? 1 : 0;
+
+            Assert.Equal(1, result.InstalledPackages.Count);
+            Assert.Equal(transitivePackagesCount, result.TransitivePackages.Count);
+            Assert.All(result.TransitivePackages, pkg => Assert.Equal(transitiveOriginsCount, pkg.TransitiveOrigins.Count()));
+        }
+
+        [Fact]
+        public async Task GetInstalledAndTransitivePackagesAsync_HasExpectedCache_Succeed()
```
jebriede: What does it mean to have expected cache and succeed? Please make sure the test method describes that more clearly.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
```diff
@@ -134,14 +134,13 @@ internal async Task<ProjectPackages> GetInstalledAndTransitivePackagesAsync(bool
...
+            Dictionary<string, TransitiveEntry> transitiveOrigins = default;
             if (IsInstalledAndTransitiveComputationNeeded)
             {
                 // clear the transitive packages cache, since we don't know when a dependency has been removed
                 installedPackages = new T();
-                if (includeTransitivePackages)
-                {
-                    transitivePackages = new T();
-                }
+                transitivePackages = new T();
```
jebriede, line 142: On [line 165 before your fix](https://github.com/NuGet/NuGet.Client/blob/87fc99d5049cfa8fd23795e96d30c78aace32cb6/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs#L165): there's another time where the transitivePackages cache is reset. Should the transitiveOrigins be reset there too? If not, what is different about that case?

jebriede (commented):
Code looks good but some comments that should be addressed for the tests. One question about invaliding transitive origins cache that might have been missed.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
martinrrm: Good catch!

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
```diff
@@ -4127,6 +4126,107 @@ public async Task GetInstalledAndTransitivePackagesAsync_TransitiveOriginsParamS
...
+
+            CpsPackageReferenceProject project = PrepareCpsRestoredProject(initialProjectSpec, cache);
+
+            // Act
+            ProjectPackages result = await project.GetInstalledAndTransitivePackagesAsync(true, true, CancellationToken.None);
+
+            // Assert
+            Assert.Equal(1, result.InstalledPackages.Count);
+            Assert.Equal(1, result.TransitivePackages.Count);
+            Assert.All(result.TransitivePackages, pkg => Assert.Equal(1, pkg.TransitiveOrigins.Count()));
```
jebriede: I believe these assertions under the same conditions are already covered in other tests above. Please correct me if I'm wrong. If it's true, then we shouldn't be testing them again in this test. We can assume they are correct because another test covers this. We can simply set up our test here.

In general, whenever we see Arrange I / Act I / Assert I and Arrange II / Act II / Assert II, ..., etc. it suggests that we're testing more than one thing in a test and that should probably be broken up into multiple tests or we are asserting things we already covered in another test already.

martinrrm: Removing these 3 asserts, I was trying to be more clear that the cache was "something" and changes to other state, but we can assume its true since we have tests that cover these asserts.

nkolev92:
Is https://github.com/NuGet/NuGet.Client/pull/5099 still happening?

martinrrm:
@nkolev92 No, I just closed it, this is the correct fix

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
jebriede: You can put a comment in the test that explains the state the cache should be after each step of the arrange steps which should help explain that. That way, we can still be clear about the state of the cache but not have multiple tests failing for the same reason if this code were to break.

dominoFire: I think those tests are trying to test how the cache evolves over time. e.g. Step 1: Read on Installed tab, Step 2: go to browse tab and so on.

dominoFire (dismissed):
Looks good so far. Great work!

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
martinrrm: @jebriede Yes, this test is trying to test how the cache changes when calling GetInstalledPackages or GIATPA. I didn't specify in the test which user action triggers the GetInstalledPackages or GIATPA, because they can be called from multiple actions (doing a refresh in a tab, changing tabs, installing package) this is just testing that even if the cache was cleared, we can calculate transitive data correctly.

There is a comment here https://github.com/NuGet/NuGet.Client/pull/5117/files#diff-f3327a2a5bc6e2fc4bc816d16cea0eee2f993d91e354fee24676d717efd1a74eR4205, where we are expecting the cache for transitive's to be clear

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
```diff
@@ -4127,6 +4126,96 @@ public async Task GetInstalledAndTransitivePackagesAsync_TransitiveOriginsParamS
...
+
+            CpsPackageReferenceProject project = PrepareCpsRestoredProject(initialProjectSpec, cache);
+
+            // Act
+            ProjectPackages result = await project.GetInstalledAndTransitivePackagesAsync(true, true, CancellationToken.None);
+
+            // do a restore
+            project = PrepareCpsRestoredProject(initialProjectSpec, cache);
+
+            var installedPackages = await project.GetInstalledPackagesAsync(CancellationToken.None); // This will change IsInstalledAndTransitiveComputationNeeded to false
```
nkolev92: this is a bit of the test smell as it assumes the implementation of the API.

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
```diff
@@ -4127,6 +4126,96 @@ public async Task GetInstalledAndTransitivePackagesAsync_TransitiveOriginsParamS
...
+            result = await project.GetInstalledAndTransitivePackagesAsync(false, false, CancellationToken.None);
+
+            // Assert I: Cache for transitive was cleared
+            Assert.Equal(1, result.InstalledPackages.Count);
+            Assert.Equal(0, result.TransitivePackages.Count);
+            Assert.All(result.TransitivePackages, pkg => Assert.Equal(0, pkg.TransitiveOrigins.Count()));
+
+            project = PrepareCpsRestoredProject(initialProjectSpec, cache);
```
nkolev92: Is this needed?
I thought the idea is/was to just call multiple times in a row.

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
martinrrm: Actually I can delete this line, GetInstalledPackagesAsync does a  GetInstalledAndTransitivePackagesAsync(false, false, CancellationToken.None)

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
martinrrm: I can delete this too

nkolev92 (approved):

jebriede (approved):

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
```diff
@@ -188,6 +188,11 @@ internal async Task<ProjectPackages> GetInstalledAndTransitivePackagesAsync(bool
             IEnumerable<PackageReference> calculatedTransitivePackages = Enumerable.Empty<PackageReference>();
             if (includeTransitivePackages)
             {
+                if (targetsList == null)
+                {
+                    targetsList = await GetTargetsListAsync(assetsFilePath, token);
+                }
```
kartheekp-ms, line 194:
```suggestion
                targetsList ??= await GetTargetsListAsync(assetsFilePath, token);
```

kartheekp-ms: not a blocking comment

PR #5120, opened 2023-03-31
added a comment indicating that 'ItemFilter.All' represents "Browse" tab in PM UI

kartheekp-ms:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2222

## Description

During [recent data analysis](https://github.com/NuGet/Client.Engineering/issues/2097#issuecomment-1490823705), I discovered that the PM UI `Browse` tab was named `All` in the code, causing confusion. To address this inconsistency, I added a comment in the source code. I initially proposed to rename the enum but as per team feedback ended up adding a comment.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] N/A 

- **Documentation**
  
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/ItemFilter.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/ItemFilter.cs
@@ -6,7 +6,7 @@ namespace NuGet.PackageManagement.UI
     public enum ItemFilter
     {
         /// <summary>
-        /// Represents Browse tab in PM UI
+        /// The value All represents the Browse tab in PM UI
         /// </summary>
         All,
         Installed,
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/ItemFilter.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/ItemFilter.cs
@@ -5,6 +5,9 @@ namespace NuGet.PackageManagement.UI
 {
     public enum ItemFilter
     {
+        /// <summary>
+        /// Represents Browse tab in PM UI
+        /// </summary>
         All,
         Installed,
         UpdatesAvailable,
--- a/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/ItemFilter.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/ItemFilter.cs
@@ -6,7 +6,7 @@ namespace NuGet.VisualStudio.Internal.Contracts
     public enum ItemFilter
     {
         /// <summary>
-        /// Represents Browse tab in PM UI
+        /// The value All represents the Browse tab in PM UI
         /// </summary>
         All,
         Installed,
--- a/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/ItemFilter.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/ItemFilter.cs
@@ -5,6 +5,9 @@ namespace NuGet.VisualStudio.Internal.Contracts
 {
     public enum ItemFilter
     {
+        /// <summary>
+        /// Represents Browse tab in PM UI
+        /// </summary>
         All,
         Installed,
         UpdatesAvailable,


nkolev92 (dismissed):

erdembayar:
Thank you for creating this PR.

Did you complete the telemetry assessment before creating this PR?
In my opinion, the ship has already sailed. The older VS will always emit telemetry with All, and I don't understand why it was named "All" from the beginning. Renaming it now would cause the dashboard to track two different sets of telemetry for the same thing. I'm not sure if it's worth doing, because the current dashboard would stop working. We would need to do more engineering work to make this change functional, probably involving duplicating any investigation work for the previous All and the renamed Browse.
The ROI seems very low, in my opinion.

erdembayar (changes requested):
Thank you for creating this PR.

Did you complete the telemetry assessment before creating this PR?
In my opinion, the ship has already sailed. The older VS will always emit telemetry with All, and I also don't understand why it was named "All" from the beginning. Renaming it now would cause the dashboard to track two different sets of telemetry for the same thing. I'm not sure if it's worth doing, because the current dashboard would stop working. We would need to do more engineering work to make this change functional, probably involving duplicating any investigation work for the previous `All` and the renamed `Browse`.
In addition, I see public api change.
The ROI seems very low, in my opinion. Maybe adding clarifying comment in code would be better, that way we don't create more engineering work.

jeffkl (dismissed):

kartheekp-ms:
> Thank you for creating this PR.
> 
> Did you complete the telemetry assessment before creating this PR? In my opinion, the ship has already sailed. The older VS will always emit telemetry with All, and I also don't understand why it was named "All" from the beginning. Renaming it now would cause the dashboard to track two different sets of telemetry for the same thing. I'm not sure if it's worth doing, because the current dashboard would stop working. We would need to do more engineering work to make this change functional, probably involving duplicating any investigation work for the previous `All` and the renamed `Browse`. In addition, I see public api change. The ROI seems very low, in my opinion. Maybe adding clarifying comment in code would be better, that way we don't create more engineering work.

@erdembayar I have responded to your feedback in https://github.com/NuGet/Client.Engineering/issues/2222#issuecomment-1495374663. If you have any further questions, please do not hesitate to leave a comment.

kartheekp-ms:
@erdembayar @jebriede - As per our conversation during the team review on this PR, I updated this PR by adding a comment that "ItemFilter.All" represents Browse tab in PM UI. Please review the latest changes and let me know your feedback.

src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/ItemFilter.cs
```diff
@@ -5,6 +5,9 @@ namespace NuGet.VisualStudio.Internal.Contracts
 {
     public enum ItemFilter
     {
+        /// <summary>
+        /// Represents Browse tab in PM UI
+        /// </summary>
```
kartheekp-ms: As per Andy's comment in https://github.com/NuGet/Client.Engineering/issues/2222#issuecomment-1497100858, this entire project can be removed but that is out of scope for this PR. I added a comment here to avoid confusion.

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/ItemFilter.cs
```diff
@@ -5,6 +5,9 @@ namespace NuGet.PackageManagement.UI
 {
     public enum ItemFilter
     {
+        /// <summary>
+        /// Represents Browse tab in PM UI
```
jebriede: This might make it clearer that the comment is specifically for the All value.

```suggestion
        /// The value All represents the Browse tab in PM UI
```

jebriede (dismissed):
Approved with suggestion on modifying the wording of the comment. Thanks!

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/ItemFilter.cs
kartheekp-ms: Thank you for the feedback. Fixed in the latest commit.

jeffkl (approved):

erdembayar (approved):

PR #5126, opened 2023-04-11
Dotnet NuGet Config CLI Config Get Command

dainizzz:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12475

## Description

Adds the `dotnet nuget config get` command, which gets the NuGet configuration settings that will be applied. When provided the `ALL` argument, this command will get all merged NuGet configuration settings from multiple NuGet configuration files that will be applied, when invoking the command from the working directory path. When provided a `CONFIG_KEY` argument, this will get the effective value of the specified configuration settings of the config section.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -128,12 +128,12 @@ public void ConfigGetCommand_ConfigKeyArgNoDirectoryArg_Success()
...
         {
             // Arrange & Act
-            using var testInfo = new TestInfo("NuGet.Config", "/subfolder");
+            using var testInfo = new TestInfo("NuGet.Config", "\\subfolder");
             {
                 var result = CommandRunner.Run(
                     DotnetCli,
                     Directory.GetCurrentDirectory(),
-                    $"{XplatDll} config get all {testInfo.WorkingPath + "/subfolder"}",
+                    $"{XplatDll} config get all {testInfo.WorkingPath + "\\subfolder"}",
```
zivkan:
```suggestion
                    $"{XplatDll} config get all {Path.Combine(testInfo.WorkingPath, "subfolder")}",
```

Always use `Path.Combine` to combine paths, as it will use the correct path separator character for the operating system. Tehnically, `/` should work on all, but using `\` will not do what you expect on Linux, and probably Mac.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -54,5 +96,108 @@ public static ISettings GetSettingsFromDirectory(string directory)
...
+            }
+            return item.Value;
+        }
+
+        /// <summary>
+        /// Logs each section of the configuration settings that will be applied, grouped by file path.
+        /// </summary>
+        public static void LogSectionsWithPaths(Dictionary<string, VirtualSettingSection> sections, ILogger logger)
+        {
+            foreach (var section in sections)
```
donnie-msft: This will result in a null reference exception (NRE) when the `NullSettings` object is used.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -54,5 +96,108 @@ public static ISettings GetSettingsFromDirectory(string directory)
...
+            {
+                logger.LogMinimal(section.Key + ":");
+                var items = section.Value.Items;
+
+                var groupByConfigPathsQuery =
```
donnie-msft: Should use an explicit type

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -825,4 +825,16 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
...
     <comment>0 - path</comment>
   </data>
+  <data name="ConfigCommandKeyNotFound" xml:space="preserve">
+    <value>Key '{0}' not found.</value>
+  </data>
+  <data name="ConfigGetAllOrConfigKeyDescription" xml:space="preserve">
+    <value>ALL: Get all merged NuGet configuration settings from multiple NuGet configuration files that will be applied, when invoking NuGet command from the working directory path.  CONFIG_KEY: Get the effective value of the specified configuration settings of the config section.</value>
+  </data>
+  <data name="ConfigGetShowPathDescription" xml:space="preserve">
+    <value>Indicate that the NuGet configuration file path will be shown besides the configuration settings.</value>
```
donnie-msft:
```suggestion
    <value>Indicate that the NuGet configuration file path will be shown beside the configuration settings.</value>
```

src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs
```diff
@@ -43,6 +43,11 @@ public class Settings : ISettings
 
         private readonly Dictionary<string, VirtualSettingSection> _computedSections;
 
+        public Dictionary<string, VirtualSettingSection> GetComputedSections()
```
donnie-msft: You can instead make your `_computedSections` field a Property that has a (public) `get` and a `private set`.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -69,6 +69,171 @@ public void ConfigPathsCommand_HelpMessage_Success()
             DotnetCliUtil.VerifyResultSuccess(result, helpMessage);
         }
 
+        [Fact]
+        public void ConfigGetCommand_ConfigKeyArg_Success()
+        {
+            // Arrange & Act
+            using var testInfo = new TestInfo("NuGet.Config");
+            {
```
donnie-msft: Can drop the brackets on `using` statements to avoid indenting the rest of the method.

src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs
dainizzz: I was getting an error when I tried to make `_computedSections` public, so I created a property called `ComputedSections` instead that gets the value of `_computedSections`. 

The only thing I'm unsure about now is that I had to cast the ISettings object I was working with in `ConfigRunner.cs` to a Settings object to be able to use the new property. I don't know if this might cause any problems. The method I'm using to get the settings is in the same file as `_computedSections`, and so far I haven't encountered any issues, so it looks like it's fine?

martinrrm: +1 to `ComputedSections`. I think it's good too, you're getting the settings from a function that is only defined in the `Settings` class https://github.com/NuGet/NuGet.Client/pull/5126/files#diff-3783860f3f9dc2e36126a8a162f70f9ab66956adab808551c91ebbf8a224b63aR48.

src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs
```diff
@@ -43,6 +43,8 @@ public class Settings : ISettings
 
         private readonly Dictionary<string, VirtualSettingSection> _computedSections;
 
+        public Dictionary<string, VirtualSettingSection> ComputedSections { get { return _computedSections; } }
```
zivkan:
```suggestion
        public Dictionary<string, VirtualSettingSection> ComputedSections { get; }
```

Since the getter does nothing more than return a private field, we can let the compiler do this automatically for us with an "auto prop". You'll need to change all references of `_compuedSections` to `ComputedSections`, but otherwise it works the same. (well, almost the same, if you want to be pedantic?  Technically the IL will have an additional method call, the getter function, but the JIT might inline/optimize it away anyway, so real runtime perf probably isn't affected)

This is what @donnie-msft  suggested in https://github.com/NuGet/NuGet.Client/pull/5126#discussion_r1167154519, but I think you misunderstood his suggestion.

src/NuGet.Core/NuGet.Configuration/Settings/Items/AddItem.cs
```diff
@@ -76,6 +76,16 @@ public virtual string GetValueAsPath()
             return Value;
         }
 
+        public string GetConfigFilePath()
```
zivkan: From an API design point of view, this feels like something more appropriate as a property, rather than a method.

src/NuGet.Core/NuGet.Configuration/Settings/Items/AddItem.cs
```diff
@@ -76,6 +76,16 @@ public virtual string GetValueAsPath()
...
+        {
+            if (Origin != null)
+            {
+                return Origin.ConfigFilePath;
+            }
+
+            return string.Empty;
```
zivkan: From an API design point of view, why return an empty string?  If a setting was created programatically in-memory, so doesn't have a backing file, `null` feels like a reasonable return value. I think if I'm writing code, trying to get config file paths and I get an empty string that's more confusing.

But also, the method should have XMLDoc that explains this (it's really a shame we suppress the warnings telling us to always add xmldoc on public APIs)

src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs
donnie-msft: > I was getting an error when I tried to make _computedSections public,

Right, you don't want to do this (that's a field, not a property)

> so I created a property called ComputedSections instead that gets the value of _computedSections

👍.  Also, reiterating Andy's comment, make sure you use an autoprop.

> The only thing I'm unsure about now is that I had to cast the ISettings object I was working with in ConfigRunner.cs to a Settings object to be able to use the new property

That's because the new property is only defined in that `Settings` class, and not the `ISettings`. That's a design decision for you to make. The cost of the current implementation is that it requires the `Settings` type, and all our other implementations (eg, `ImmutableSettings`, `NullSettings`) will not be aware of `ComputedSections`. If you're good with that, then the API change you've proposed makes sense to me.
I'd be careful for the case when there's only invalid nuget.configs: That looks like it would result in your code encountering `NullSettings`, which your cast will fail. See: https://github.com/NuGet/NuGet.Client/blob/d0bb0a180105de35a29ee89f52282a0eabd20588/src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs#L511

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -54,5 +100,108 @@ public static ISettings GetSettingsFromDirectory(string directory)
...
+        /// <exception cref="ArgumentNullException"></exception>
+        public static string GetValueForConfigKey(ISettings settings, string key, bool showPath)
+        {
+            if (settings == null)
+            {
+                throw new ArgumentNullException(nameof(settings));
+            }
+
+            var sectionElement = settings.GetSection(ConfigurationConstants.Config);
+            var item = sectionElement?.GetFirstItemWithAttribute<AddItem>(ConfigurationConstants.KeyAttribute, key);
```
donnie-msft: Use explicit types

martinrrm (approved):
LGTM, let's wait for previous reviewers to see if they have more comments on your recent changes.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigCommand.cs
```diff
@@ -17,7 +17,7 @@ internal class ConfigCommand
                 ConfigCmd.Command("paths", PathsCmd =>
                 {
                     CommandArgument workingdirectory = PathsCmd.Argument(
-                        "working-directory",
+                        "WORKING_DIRECTORY",
```
jebriede: Can we make this a constant to ensure consistency throughout? I see it's used in multiple places in this file.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigArgs.cs
```diff
@@ -8,4 +8,11 @@ internal class ConfigPathsArgs
     {
         public string? WorkingDirectory { get; set; }
     }
+
+    internal partial class ConfigGetArgs
```
jebriede: I am not seeing why this needs to be a partial class. Is there another part of this class definition elsewhere that is not part of this PR?

src/NuGet.Core/NuGet.Configuration/Settings/SettingItem.cs
```diff
@@ -101,5 +103,11 @@ internal virtual void Update(SettingItem setting)
                 AddOrUpdateAttribute(attribute.Key, attribute.Value);
             }
         }
+
+        public IEnumerable<XAttribute> GetXElementAttributes()
+        {
+            var xElement = Node as XElement;
+            return xElement.Attributes();
```
jebriede: Because we use the implicit cast above (Node as XElement), we are saying that it is possible for Node to not be an XElement, in which case xElement will be null here. If this indeed a valid scenario, we need to guard against null here, otherwise, when we dereference xElement, it will throw an exception.

If we know for certain that Node will always be an XElement, we can instead explicitly cast it and then we don't need to check for xElement being null.

src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs
```diff
@@ -41,11 +41,11 @@ public class Settings : ISettings
             new[] { "*.config" } :
             new[] { "*.Config", "*.config" };
 
-        private readonly Dictionary<string, VirtualSettingSection> _computedSections;
+        public Dictionary<string, VirtualSettingSection> ComputedSections { get; }
```
jebriede: It looks like this class is designed to hide _computedSections from external callers and GetSection is supposed to be used to obtain data about the setting section. By exposing this ComputedSection publicly, we are changing the encapsulation of this class and we should be cautious about doing.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -40,10 +34,66 @@ public static void Run(ConfigPathsArgs args, Func<ILogger> getLogger)
...
+            var settings = RunnerHelper.GetSettingsFromDirectory(args.WorkingDirectory) as Settings;
+            if (settings == null)
+            {
+                return;
+            }
+            ILogger logger = getLogger();
+
+            if (args.AllOrConfigKey.Equals("all", StringComparison.OrdinalIgnoreCase))
+            {
+                var sections = settings.ComputedSections;
```
jebriede: Can this call GetSection instead of directly accessing the internal dictionary? This would also prevent us from having to expose that publicly.

jebriede (dismissed):
Great work! Left a few comments with requested changes and questions.

src/NuGet.Core/NuGet.Configuration/Settings/SettingItem.cs
```diff
@@ -101,5 +103,11 @@ internal virtual void Update(SettingItem setting)
                 AddOrUpdateAttribute(attribute.Key, attribute.Value);
             }
         }
+
+        public IEnumerable<XAttribute> GetXElementAttributes()
```
zivkan: Similar to one of JP's other comments, this class is designed to encapsulate the underlying XML file, and provide a higher level abstraction. So, to me, adding this as a public API feels like [a leaky abstraction](https://en.wikipedia.org/wiki/Leaky_abstraction).

On one hand, the way the settings classes are currently designed, it's supposed to hide the underlying file format, so exposing the internals violates the design.

On the other hand, if `dotnet nuget config` can't use the settings classes, that feels like there might be a design flaw too. Although in this case, you're using a "low level" API within NuGet.Configuration (SettingsItem), rather than the section specific abstractions that basically everything else is intended to use (PackageSource, PackageSourceMapping, etc).

Perhaps adding this as public API is the least bad option. It's not immediately clear to me what to recommend, so I'm not sure if this comment is actually helpful in any way. At least it's further evidence that API design is hard.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigArgs.cs
dainizzz: Oops, this was from when I was using a template to autogenerate code; I'll remove the partial keyword!

src/NuGet.Core/NuGet.Configuration/Settings/SettingItem.cs
donnie-msft: Looks like other places check for null, so probably can continue that. (Although, I wonder what other types a Node can be)

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
dainizzz: Initially, that was my plan, but since I don't know which sections exist and I want all the sections in the settings, using `ComputedSections` seemed like a better option.

src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs
dainizzz: Initially, I was using a method that returned the value of `_computedSections` to access this value, but [Andy](https://github.com/NuGet/NuGet.Client/pull/5126#discussion_r1167389921) and [Donnie](https://github.com/NuGet/NuGet.Client/pull/5126#discussion_r1167154519) suggested using an auto property to get the value instead. I wasn't able to figure out a way to use `GetSection`, since I don't know which section names might exist in the settings I'm using.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -54,5 +104,108 @@ public static ISettings GetSettingsFromDirectory(string directory)
...
+        /// Logs each section of the configuration settings that will be applied.
+        /// </summary>
+        public static void LogSectionsNoPaths(Dictionary<string, VirtualSettingSection> sections, ILogger logger)
+        {
+            foreach (var section in sections)
+            {
+                logger.LogMinimal(section.Key + ":");
+                var items = section.Value.Items;
+                LogSectionItems(items, logger);
+                logger.LogMinimal("\n");
```
jeffkl:
```suggestion
                logger.LogMinimal(Environment.NewLine);
```

src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs
jeffkl: I would vote to have a new method on `Settings` that returns an `IEnumerable<string>` which is just `_computedSections.Keys`

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigCommand.cs
```diff
@@ -9,6 +9,7 @@ namespace NuGet.CommandLine.XPlat
 {
     internal class ConfigCommand
     {
+        public const string WORKING_DIRECTORY = "WORKING_DIRECTORY";
```
jebriede: I believe only this class references this constant, in which case this should be private.

```suggestion
        private const string WORKING_DIRECTORY = "WORKING_DIRECTORY";
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -136,12 +136,12 @@ public static string GetValueForConfigKey(ISettings settings, string key, bool s
...
         /// </summary>
-        public static void LogSectionsWithPaths(Dictionary<string, VirtualSettingSection> sections, ILogger logger)
+        public static void LogSectionsWithPaths(IEnumerable<string> sections, Settings settings, ILogger logger)
         {
             foreach (var section in sections)
             {
-                logger.LogMinimal(section.Key + ":");
-                var items = section.Value.Items;
+                logger.LogMinimal(section + ":");
+                var items = settings.GetSection(section).Items;
```
jebriede: We must guard against the `settings.GetSection()` being null because that code explicitly returns null as a possibility.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -152,22 +152,22 @@ public static void LogSectionsWithPaths(Dictionary<string, VirtualSettingSection
...
         /// </summary>
-        public static void LogSectionsNoPaths(Dictionary<string, VirtualSettingSection> sections, ILogger logger)
+        public static void LogSectionsNoPaths(IEnumerable<string> sections, Settings settings, ILogger logger)
         {
             foreach (var section in sections)
             {
-                logger.LogMinimal(section.Key + ":");
-                var items = section.Value.Items;
+                logger.LogMinimal(section + ":");
+                var items = settings.GetSection(section).Items;
```
jebriede: Guard against null here too.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -136,12 +136,12 @@ public static string GetValueForConfigKey(ISettings settings, string key, bool s
         /// <summary>
         /// Logs each section of the configuration settings that will be applied, grouped by file path.
         /// </summary>
-        public static void LogSectionsWithPaths(Dictionary<string, VirtualSettingSection> sections, ILogger logger)
+        public static void LogSectionsWithPaths(IEnumerable<string> sections, Settings settings, ILogger logger)
```
jebriede: `LogSectionsWithPaths` and `LogSectionsNoPaths` have a lot of the same code. Can we deduplicate that code into a helper method they both use or the same method with a usePaths bool parameter?

src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs
```diff
@@ -41,18 +41,20 @@ public class Settings : ISettings
...
-            if (ComputedSections.TryGetValue(sectionName, out var section))
+            if (_computedSections.TryGetValue(sectionName, out var section))
             {
                 return section.Clone() as SettingSection;
             }
 
             return null;
         }
 
+        public IEnumerable<string> GetAllSettingSections() { return _computedSections.Keys; }
```
jebriede, line 56: `_computedSections` can be null here so we have to guard against that. We can do something like what `GetSection` does, with a lazy initialization of this dictionary if it hasn't already been initialized.

jebriede (commented):
Looks good! There are three `NullReferenceException`s that we need to guard against and I left a few other comments.

src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs
donnie-msft: How can it be null? Am I missing a constructor that fails to initialize it?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -54,5 +104,111 @@ public static ISettings GetSettingsFromDirectory(string directory)
...
+
+                logger.LogMinimal(setting);
+            }
+        }
+
+        /// <summary>
+        /// Throws an exception if any of the config runner arguments are null.
+        /// </summary>
+        /// <exception cref="ArgumentNullException"></exception>
+        public static void ValidateArguments<TArgs>(TArgs args, Func<ILogger> logger)
```
donnie-msft: I don't know what's best, but I'd suggest indicating in the name what's being validated. When I see "Validate args", I think of business logic.

```suggestion
        public static void ValidateArgumentsNotNull<TArgs>(TArgs args, Func<ILogger> logger)
```

```suggestion
        public static void EnsureArgumentsNotNull<TArgs>(TArgs args, Func<ILogger> logger)
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -54,5 +104,111 @@ public static ISettings GetSettingsFromDirectory(string directory)
...
+        /// Combines the attributes from each item in a collection of SettingItems into a string, then logs the string.
+        /// </summary>
+        /// <param name="items"></param>
+        /// <param name="logger"></param>
+        public static void LogSectionItems(IEnumerable<SettingItem> items, ILogger logger)
+        {
+            foreach (SettingItem item in items)
+            {
+                var setting = $"\t{item.ElementName}";
+                var attributes = item.GetAttributes();
```
donnie-msft: I know it's currently being debated, but I'd prefer to see at least something here with an explicit type. Because as it is, I can't easily scan this and know the types you're working with.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
donnie-msft: I think it's fine? There aren't currently any codepaths that I'm noticing which will pass a null.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -40,10 +34,66 @@ public static void Run(ConfigPathsArgs args, Func<ILogger> getLogger)
...
+            {
+                throw new CommandException(string.Format(CultureInfo.CurrentCulture, Strings.ConfigCommandKeyNotFound, args.AllOrConfigKey));
+            }
+
+            var settings = RunnerHelper.GetSettingsFromDirectory(args.WorkingDirectory) as Settings;
+            if (settings == null)
+            {
+                return;
+            }
+            ILogger logger = getLogger();
```
donnie-msft, line 53: Just FYI something I checked: The logger returned by `getLogger` in theory could still be null. But from a quick check, I don't think it's possible since our app initialization always logic creates one.

src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs
dainizzz: I was talking with Martin and we also couldn't figure out when `_computedSections` would be null. It looks like the constructor for `_computedSections` initializes an empty dictionary, and from what I understand, if `_computedSections` was empty, the `Keys` property would also be empty, and not null.

donnie-msft (approved):

PR #5134, opened 2023-04-18
Improve installation of .NET for building NuGet.Client repository

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2250

## Description

This makes it so that our hosted build always uses the .NET SDKs and runtimes we need for our build to work.
* Moved the specification of which versions to install from MSBuild logic to a text file.  This allows any machine to bootstrap .NET even if .NET isn't already installed.
* Introduced `configure.sh` so non-Windows machines can initialize the .NET SDKs like Windows
* Scripts set `DOTNET_ROOT`, `DOTNET_MULTILEVEL_LOOKUP=0`, and add `dotnet` to the `PATH` for that command window
* You can set an environment variable, `CLIBRANCHFORTESTING` to use different versions for testing something else or having a different configuration for different pipelines

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


build/CliVersions.txt
dtivel, line 1: Consider renaming the file to `DotNetSdkVersionsForTesting.txt`.  Similarly, consider renaming environment variable `CLIBRANCHFORTESTING` (`DOTNETSDK_VERSION_FOR_TESTING`?).  Point is:  the thing that we install and test is the .NET SDK not just its CLI.

dtivel (dismissed):

build/CliVersions.txt
jeffkl: The installed .NET SDKs are used for build and test.  I just re-used existing naming but am open to changing it.  Maybe the file should be named `DotNetSdkVersions.txt` and the env var `DOTNET_SDK_VERSIONS`?

build/DotNetSdkVersions.txt
```diff
@@ -0,0 +1,5 @@
+# Each line represents arguments for the .NET SDK installer script (https://learn.microsoft.com/dotnet/core/tools/dotnet-install-script)
+-Channel 7.0
+-Channel 6.0 -Runtime dotnet
```
zivkan, line 3: What's the purpose of the `-runtime dotnet` on the line, but not others?

We've had contributors using some IBM mainframe contribute to NuGet, but it uses the mono runtime (not Mono Project, that implements .NET Framework, but the mono runtime of the .NET Core App runtime). By specifying `-runtime dotnet`, is it going to cause problems for customers on platforms that the dotnet runtime doesn't have binaries for?

configure.ps1
```diff
@@ -26,7 +26,8 @@ Param (
     [switch]$CleanCache,
```
zivkan, line 26: Any chance while you're at it, you could move the strong name key disable code out of build.ps1 and into configure.ps1?  It's something that needs to be configured once per machine, not every build.

build/DotNetSdkVersions.txt
jeffkl: I found that some of our tests need the .NET 6 runtime but don't need the full SDK.  I've experimented a little and hope to trim this down as soon as all of our tests target .NET 7 or .NET 8.

configure.ps1
jeffkl: Yeah I wasn't sure if it needed to be configured unless you want to debug our VSIX?  Or is it safe to disable it on everyone's machine that runs this script?

build/CliVersions.txt
jeffkl: I've renamed it to `DotNetSdkVersions.txt`

configure.ps1
zivkan: I think it needs to be disabled not only to run the VSIX, but also to run the .NET Framework restore task, the non-ilmerged nuget.exe (maybe even the ilmerged nuget.exe). I'm not sure about running tests. Maybe it's needed for that too.  Basically, I think strong name validation for our 2 keys should be disabled for everyone running the script on Windows.

zivkan (approved):

build/DotNetSdkVersions.txt
```diff
@@ -0,0 +1,5 @@
...
+-Channel 5.0
```
nkolev92, line 4: Why do we need the 5.0 and 3.1 SDKs?
Aren't the runtimes sufficient for this? 

We'll always build with the latest installed SDK anyways.

PR #5136, opened 2023-04-19
Check PackageReference projects for packages with known vulnerabilities

### LINKED ISSUE #12289: Warn when vulnerabilities are detected during PackageReference restore

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12289

Regression? no

## Description

* Spec for feature: https://github.com/NuGet/Home/blob/dev/proposed/2022/vulnerabilities-in-restore.md
  * If you see anything in the implementation that does not follow the spec, please let me know!
* Added VulnerabilityInformationProvider, which caches vulnerability data per-source, so when multiple projects need to load the known vulnerability database in a restore, it's only parsed once
* Added AuditUtility class, which has all the business logic in matching known vulnerabilities to packages used in the project, and generates all the warnings
* Added relevant code to get `NuGetAudit` and `NuGetAuditLevel` MSBuild properties from PackageReference projects on CLI and VS.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception: Due to time constraints to meet the next preview ship date, tests were postponed, but will be followed up on: https://github.com/NuGet/Client.Engineering/issues/2253
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [x] Documentation PR or issue filled: https://github.com/NuGet/docs.microsoft.com-nuget/issues/3049
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
@@ -406,13 +406,10 @@ namespace NuGet.PackageManagement.VisualStudio
                 }
             }
 
-            string nugetAuditValue = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAudit);
-            bool auditEnabled = string.Equals("enable", nugetAuditValue, StringComparison.InvariantCultureIgnoreCase);
-            string auditLevel = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditLevel);
             RestoreAuditProperties auditProperties = new RestoreAuditProperties()
             {
-                EnableAudit = auditEnabled,
-                AuditLevel = auditLevel
+                EnableAudit = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAudit),
+                AuditLevel = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditLevel)
             };
 
             var msbuildProjectExtensionsPath = await GetMSBuildProjectExtensionsPathAsync();
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
@@ -406,6 +406,15 @@ namespace NuGet.PackageManagement.VisualStudio
                 }
             }
 
+            string nugetAuditValue = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAudit);
+            bool auditEnabled = string.Equals("enable", nugetAuditValue, StringComparison.InvariantCultureIgnoreCase);
+            string auditLevel = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditLevel);
+            RestoreAuditProperties auditProperties = new RestoreAuditProperties()
+            {
+                EnableAudit = auditEnabled,
+                AuditLevel = auditLevel
+            };
+
             var msbuildProjectExtensionsPath = await GetMSBuildProjectExtensionsPathAsync();
             return new PackageSpec(tfis)
             {
@@ -448,6 +457,7 @@ namespace NuGet.PackageManagement.VisualStudio
                     CentralPackageVersionsEnabled = isCpvmEnabled,
                     CentralPackageVersionOverrideDisabled = GetPropertySafe(_vsProjectAdapter.BuildProperties, ProjectBuildProperties.CentralPackageVersionOverrideEnabled).EqualsFalse(),
                     CentralPackageTransitivePinningEnabled = MSBuildStringUtility.IsTrue(GetPropertySafe(_vsProjectAdapter.BuildProperties, ProjectBuildProperties.CentralPackageTransitivePinningEnabled)),
+                    RestoreAuditProperties = auditProperties,
                 }
             };
         }
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
@@ -406,11 +406,15 @@ namespace NuGet.PackageManagement.VisualStudio
                 }
             }
 
-            RestoreAuditProperties auditProperties = new RestoreAuditProperties()
-            {
-                EnableAudit = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAudit),
-                AuditLevel = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditLevel)
-            };
+            string enableAudit = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAudit);
+            string auditLevel = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditLevel);
+            RestoreAuditProperties auditProperties = !string.IsNullOrEmpty(enableAudit) || !string.IsNullOrEmpty(auditLevel)
+                ? new RestoreAuditProperties()
+                {
+                    EnableAudit = enableAudit,
+                    AuditLevel = auditLevel
+                }
+                : null;
 
             var msbuildProjectExtensionsPath = await GetMSBuildProjectExtensionsPathAsync();
             return new PackageSpec(tfis)
--- a/src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
+++ b/src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
@@ -799,6 +799,8 @@ namespace NuGet.Build.Tasks.Console
 
             (bool isCentralPackageManagementEnabled, bool isCentralPackageVersionOverrideDisabled, bool isCentralPackageTransitivePinningEnabled) = GetCentralPackageManagementSettings(project, projectStyleOrNull);
 
+            RestoreAuditProperties auditProperties = GetRestoreAuditProperties(project);
+
             List<TargetFrameworkInformation> targetFrameworkInfos = GetTargetFrameworkInfos(projectsByTargetFramework, isCentralPackageManagementEnabled);
 
             (ProjectStyle ProjectStyle, bool IsPackageReferenceCompatibleProjectStyle, string PackagesConfigFilePath) projectStyleResult = BuildTasksUtility.GetProjectRestoreStyle(
@@ -843,6 +845,7 @@ namespace NuGet.Build.Tasks.Console
                     CentralPackageVersionsEnabled = isCentralPackageManagementEnabled && projectStyle == ProjectStyle.PackageReference,
                     CentralPackageVersionOverrideDisabled = isCentralPackageVersionOverrideDisabled,
                     CentralPackageTransitivePinningEnabled = isCentralPackageTransitivePinningEnabled,
+                    RestoreAuditProperties = auditProperties
                 };
             }
 
@@ -865,6 +868,20 @@ namespace NuGet.Build.Tasks.Console
             restoreMetadata.TargetFrameworks = GetProjectRestoreMetadataFrameworkInfos(targetFrameworkInfos, projectsByTargetFramework);
 
             return (restoreMetadata, targetFrameworkInfos);
+        }
+
+        private RestoreAuditProperties GetRestoreAuditProperties(IMSBuildProject project)
+        {
+            string enableAudit = project.GetProperty("NuGetAudit");
+            string auditLevel = project.GetProperty("AuditLevel");
+
+            return !string.IsNullOrEmpty(enableAudit) || !string.IsNullOrEmpty(auditLevel)
+                ? new RestoreAuditProperties()
+                {
+                    EnableAudit = enableAudit,
+                    AuditLevel = auditLevel
+                }
+                : null;
         }
 
         /// <summary>
--- a/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
+++ b/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
@@ -849,6 +849,8 @@ Copyright (c) .NET Foundation. All rights reserved.
         <_CentralPackageVersionsEnabled>$(_CentralPackageVersionsEnabled)</_CentralPackageVersionsEnabled>
         <CentralPackageVersionOverrideEnabled>$(CentralPackageVersionOverrideEnabled)</CentralPackageVersionOverrideEnabled>
         <CentralPackageTransitivePinningEnabled>$(CentralPackageTransitivePinningEnabled)</CentralPackageTransitivePinningEnabled>
+        <NuGetAudit>$(NuGetAudit)</NuGetAudit>
+        <NuGetAuditLevel>$(NuGetAuditLevel)</NuGetAuditLevel>
       </_RestoreGraphEntry>
     </ItemGroup>
 
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
@@ -16,7 +16,9 @@ using NuGet.Frameworks;
 using NuGet.LibraryModel;
 using NuGet.Packaging.Core;
 using NuGet.ProjectModel;
+using NuGet.Protocol;
 using NuGet.Protocol.Core.Types;
+using NuGet.Protocol.Model;
 using NuGet.Repositories;
 using NuGet.RuntimeModel;
 using NuGet.Versioning;
@@ -81,6 +83,7 @@ namespace NuGet.Commands
         private const string ValidateRestoreGraphsDuration = nameof(ValidateRestoreGraphsDuration);
         private const string CreateRestoreResultDuration = nameof(CreateRestoreResultDuration);
         private const string IsCentralPackageTransitivePinningEnabled = nameof(IsCentralPackageTransitivePinningEnabled);
+        private const string VulnerablePackageCheck = nameof(VulnerablePackageCheck);
 
         // PackageSourceMapping names
         private const string PackageSourceMappingIsMappingEnabled = "PackageSourceMapping.IsMappingEnabled";
@@ -274,6 +277,10 @@ namespace NuGet.Commands
                         return RestoreTargetGraph.Create(_request.Project.RuntimeGraph, Enumerable.Empty<GraphNode<RemoteResolveResult>>(), contextForProject, _logger, e.Framework, e.RuntimeIdentifier);
                     });
                 }
+
+                telemetry.StartIntervalMeasure();
+                await CheckPackageVulnerabilitiesAsync(graphs, _logger, token);
+                telemetry.EndIntervalMeasure(VulnerablePackageCheck);
 
                 telemetry.StartIntervalMeasure();
                 // Create assets file
@@ -1300,6 +1307,152 @@ namespace NuGet.Commands
                 project,
                 msbuildProjectPath: null,
                 projectReferences: Enumerable.Empty<string>());
+        }
+
+#nullable enable
+
+        private async Task CheckPackageVulnerabilitiesAsync(IEnumerable<RestoreTargetGraph> graphs, RestoreCollectorLogger logger, CancellationToken cancellationToken)
+        {
+            GetVulnerabilityInfoResult? allVulnerabilityData = await GetAllVulnerabilityDataAsync(logger, cancellationToken);
+            if (allVulnerabilityData == null) return;
+
+            foreach (var graph in graphs)
+            {
+                foreach (LibraryIdentity package in graph.Flattened.Select(i => i.Key).Where(p => p.Type == "package"))
+                {
+                    var knownVulnerabilities = GetKnownVulnerabilities(package.Name, package.Version, allVulnerabilityData.KnownVulnerabilities);
+
+                    if (knownVulnerabilities?.Count() > 0)
+                    {
+                        var sb = StringBuilderPool.Shared.Rent(1);
+                        try
+                        {
+                            sb.Clear();
+                            sb.Append("Package ");
+                            sb.Append(package.Name);
+                            sb.Append("/");
+                            sb.Append(package.Version);
+                            sb.AppendLine(" has known vulnerabilities:");
+
+                            foreach (var vuln in knownVulnerabilities)
+                            {
+                                sb.Append("  ");
+                                sb.AppendLine(vuln.Url.OriginalString);
+                                sb.Append("    Severity: ");
+                                sb.AppendLine(vuln.Severity.ToString());
+                                sb.Append("    Affected versions: ");
+                                sb.AppendLine(vuln.Versions.ToShortString());
+                            }
+
+                            var message = RestoreLogMessage.CreateWarning(NuGetLogCode.NU1000, sb.ToString(),
+                                package.Name,
+                                graph.TargetGraphName);
+                            message.ProjectPath = _request.Project.FilePath;
+                            logger.Log(message);
+                        }
+                        finally
+                        {
+                            StringBuilderPool.Shared.Return(sb);
+                        }
+                    }
+                }
+            }
+
+            List<PackageVulnerabilityInfo>? GetKnownVulnerabilities(
+                string name,
+                NuGetVersion version,
+                IReadOnlyList<IReadOnlyDictionary<string, IReadOnlyList<PackageVulnerabilityInfo>>>? knownVulnerabilities)
+            {
+                HashSet<PackageVulnerabilityInfo>? vulnerabilities = null;
+
+                if (knownVulnerabilities == null) return null;
+
+                foreach (var file in knownVulnerabilities)
+                {
+                    if (file.TryGetValue(name, out var packageVulnerabilities))
+                    {
+                        foreach (var vulnInfo in packageVulnerabilities)
+                        {
+                            if (vulnInfo.Versions.Satisfies(version))
+                            {
+                                if (vulnerabilities == null)
+                                {
+                                    vulnerabilities = new();
+                                }
+                                vulnerabilities.Add(vulnInfo);
+                            }
+                        }
+                    }
+                }
+
+                return vulnerabilities != null ? vulnerabilities.ToList() : null;
+            }
+        }
+
+        private async Task<GetVulnerabilityInfoResult?> GetAllVulnerabilityDataAsync(ILogger logger, CancellationToken cancellationToken)
+        {
+            var sources = new List<SourceRepository>(_request.DependencyProviders.RemoteProviders.Count);
+            sources.AddRange(_request.DependencyProviders.RemoteProviders.Select(p => p.SourceRepository).Where(s => s is not null));
+
+            var results = new Task<GetVulnerabilityInfoResult?>[sources.Count];
+            for (int i = 0; i < sources.Count; i++)
+            {
+                SourceRepository source = sources[i];
+                results[i] = GetVulnerabilityInfoAsync(source, logger, cancellationToken);
+            }
+
+            await Task.WhenAll(results);
+            if (cancellationToken.IsCancellationRequested)
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+            }
+
+            List<Exception>? errors = null;
+            List<IReadOnlyDictionary<string, IReadOnlyList<PackageVulnerabilityInfo>>>? knownVulnerabilities = null;
+            foreach (var resultTask in results)
+            {
+                var result = await resultTask;
+                if (result is null) continue;
+
+                if (result.KnownVulnerabilities != null)
+                {
+                    if (knownVulnerabilities == null)
+                    {
+                        knownVulnerabilities = new();
+                    }
+
+                    knownVulnerabilities.AddRange(result.KnownVulnerabilities);
+                }
+
+                if (result.Exceptions != null)
+                {
+                    if (errors == null)
+                    {
+                        errors = new();
+                    }
+
+                    errors.AddRange(result.Exceptions.InnerExceptions);
+                }
+            }
+
+            GetVulnerabilityInfoResult? final =
+                knownVulnerabilities != null || errors != null
+                ? new(knownVulnerabilities, errors != null ? new AggregateException(errors) : null)
+                : null;
+            return final;
+
+            static async Task<GetVulnerabilityInfoResult?> GetVulnerabilityInfoAsync(SourceRepository source, ILogger logger, CancellationToken cancellationToken)
+            {
+                IVulnerabilityInfoResource? vulnerabilityInfoResource = await source.GetResourceAsync<IVulnerabilityInfoResource>(cancellationToken);
+                if (vulnerabilityInfoResource == null)
+                {
+                    return null;
+                }
+
+                using SourceCacheContext cacheContext = new();
+                GetVulnerabilityInfoResult result = await vulnerabilityInfoResource.GetVulnerabilityInfoAsync(cacheContext, logger, cancellationToken);
+                return result;
+            }
         }
     }
 }
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
@@ -1316,35 +1316,84 @@ namespace NuGet.Commands
             GetVulnerabilityInfoResult? allVulnerabilityData = await GetAllVulnerabilityDataAsync(logger, cancellationToken);
             if (allVulnerabilityData == null) return;
 
-            foreach (var graph in graphs)
-            {
-                foreach (LibraryIdentity package in graph.Flattened.Select(i => i.Key).Where(p => p.Type == "package"))
-                {
-                    var knownVulnerabilities = GetKnownVulnerabilities(package.Name, package.Version, allVulnerabilityData.KnownVulnerabilities);
+            // multi-targeting projects often use the same package across multiple TFMs, so group to reduce output spam.
+            Dictionary<PackageIdentity, Dictionary<PackageVulnerabilityInfo, List<string>>>? packagesWithKnownVulnerabilities =
+                allVulnerabilityData.KnownVulnerabilities != null
+                ? FindPackagesWithKnownVulnerabilities(graphs, allVulnerabilityData.KnownVulnerabilities)
+                : null;
 
-                    if (knownVulnerabilities?.Count() > 0)
+            if (packagesWithKnownVulnerabilities != null)
+            {
+                // .NET Framework and .NET Standard don't have Deconstructor methods for KeyValuePair :(
+                foreach (var kvp1 in packagesWithKnownVulnerabilities.OrderBy(p => p.Key.Id))
+                {
+                    PackageIdentity package = kvp1.Key;
+                    Dictionary<PackageVulnerabilityInfo, List<string>> vulnerabilities = kvp1.Value;
+                    foreach (var kvp2 in vulnerabilities.OrderBy(v => v.Key.Url.OriginalString))
                     {
-                        foreach (var knownVulnerability in knownVulnerabilities)
-                        {
-                            (string severityLabel, NuGetLogCode logCode) = GetSeverityLabelAndCode(knownVulnerability.Severity);
-                            string message = string.Format(Strings.Warning_PackageWithKnownVulnerability,
-                                package.Name,
-                                package.Version.ToNormalizedString(),
-                                severityLabel,
-                                knownVulnerability.Url);
-                            RestoreLogMessage restoreLogMessage =
-                                RestoreLogMessage.CreateWarning(logCode,
-                                message,
-                                package.Name,
-                                graph.TargetGraphName);
-                            restoreLogMessage.ProjectPath = _request.Project.FilePath;
-                            logger.Log(restoreLogMessage);
-                        }
+                        PackageVulnerabilityInfo vulnerability = kvp2.Key;
+                        List<string> affectedGraphs = kvp2.Value;
+                        (string severityLabel, NuGetLogCode logCode) = GetSeverityLabelAndCode(vulnerability.Severity);
+                        string message = string.Format(Strings.Warning_PackageWithKnownVulnerability,
+                            package.Id,
+                            package.Version.ToNormalizedString(),
+                            severityLabel,
+                            vulnerability.Url);
+                        RestoreLogMessage restoreLogMessage =
+                            RestoreLogMessage.CreateWarning(logCode,
+                            message,
+                            package.Id,
+                            affectedGraphs.ToArray());
+                        restoreLogMessage.ProjectPath = _request.Project.FilePath;
+                        logger.Log(restoreLogMessage);
                     }
                 }
             }
 
-            List<PackageVulnerabilityInfo>? GetKnownVulnerabilities(
+            static Dictionary<PackageIdentity, Dictionary<PackageVulnerabilityInfo, List<string>>>?
+                FindPackagesWithKnownVulnerabilities(
+                    IEnumerable<RestoreTargetGraph> graphs,
+                    IReadOnlyList<IReadOnlyDictionary<string, IReadOnlyList<PackageVulnerabilityInfo>>> knownVulnerabilities)
+            {
+                Dictionary<PackageIdentity, Dictionary<PackageVulnerabilityInfo, List<string>>>? result = null;
+
+                foreach (var graph in graphs)
+                {
+                    foreach (LibraryIdentity package in graph.Flattened.Select(i => i.Key).Where(p => p.Type == "package"))
+                    {
+                        var fromFile = GetKnownVulnerabilities(package.Name, package.Version, knownVulnerabilities);
+
+                        if (fromFile?.Count() > 0)
+                        {
+                            PackageIdentity packageIdentity = new(package.Name, package.Version);
+
+                            if (result == null)
+                            {
+                                result = new();
+                            }
+
+                            if (!result.TryGetValue(packageIdentity, out Dictionary<PackageVulnerabilityInfo, List<string>>? knownPackageVulnerabilities))
+                            {
+                                knownPackageVulnerabilities = new();
+                                result.Add(packageIdentity, knownPackageVulnerabilities);
+                            }
+
+                            foreach (PackageVulnerabilityInfo knownVulnerability in fromFile)
+                            {
+                                if (!knownPackageVulnerabilities.TryGetValue(knownVulnerability, out List<string>? affectedGraphs))
+                                {
+                                    affectedGraphs = new();
+                                    knownPackageVulnerabilities.Add(knownVulnerability, affectedGraphs);
+                                }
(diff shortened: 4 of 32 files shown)


zivkan:
## Perf Results

|Repo|dev branch|NuGetAudit=disable|NuGetAudit=enable|Duration increase|
|--|--:|--:|--:|--:|
|NuGet.Client|8.0 s|7.8 s|10.7 s|37%|
|orleans|6.9 s|6.9 s|11.4 s|65%|
|OrchardCore|24.4 s|22.5 s|25.3 s|12%|

* "dev branch" was the last commit on the dev branch before my branch's first commit, to ensure perf is as comparable as possible
* OrchardCore restores much, much faster on the dotnet CLI vs msbuild, but only msbuild was tested here.

<details>
<summary>Source code</summary>

<div>

```csharp
using System.Diagnostics;

[SimpleJob(launchCount: 1, warmupCount: 1, iterationCount: 10, invocationCount: 1)]
public class Program
{
    static void Main()
    {
    }

    public enum TestScenario { mainBranch, disable, enable }

    [ParamsAllValues]
    public TestScenario Feature;

    private string processArguments;

    private string beforeTargets;
    private string afterTargets;
    private string NuGetClientRepoPath;
    private string OrleansRepoPath;
    private string OrchardCoreRepoPath;

    [GlobalSetup]
    public void GlobalSetup()
    {
        beforeTargets = Environment.GetEnvironmentVariable("BeforeTargets") ?? throw new Exception("Set environment variable 'BeforeTargets' to NuGet.targets without the feature");
        if (!File.Exists(beforeTargets)) { throw new FileNotFoundException("Before targets file does not exist", beforeTargets); }
        afterTargets = Environment.GetEnvironmentVariable("BeforeTargets") ?? throw new Exception("Set environment variable 'AfterTargets' to NuGet.targets with the feature");
        if (!File.Exists(afterTargets)) { throw new FileNotFoundException("After targets file does not exist", afterTargets); }
        NuGetClientRepoPath = Environment.GetEnvironmentVariable("NuGetClientRepoPath") ?? throw new Exception("Set environment variable 'NuGetClientRepoPath' to where https://github.com/NuGet/NuGet.Client has been cloned");
        if (!Directory.Exists(NuGetClientRepoPath)) { throw new DirectoryNotFoundException($"Directory '{NuGetClientRepoPath}' not found"); }
        OrleansRepoPath = Environment.GetEnvironmentVariable("OrleansRepoPath") ?? throw new Exception("Set environment variable 'OrleansRepoPath' to where https://github.com/dotnet/orleans has been cloned");
        if (!Directory.Exists(OrleansRepoPath)) { throw new DirectoryNotFoundException($"Directory '{OrleansRepoPath}' not found"); }
        OrchardCoreRepoPath = Environment.GetEnvironmentVariable("OrchardCoreRepoPath") ?? throw new Exception("Set environment variable 'OrchardCoreRepoPath' to where https://github.com/OrchardCMS/OrchardCore has been cloned");
        if (!Directory.Exists(OrchardCoreRepoPath)) { throw new DirectoryNotFoundException($"Directory '{OrchardCoreRepoPath}' not found"); }

        List<string> args = new List<string>()
        {
            "-t:restore",
            "-v:q",
            "-p:RestoreAdditionalProjectSources=https://devapi.nugettest.org/v3/index.json",
            "-p:RestoreForce=true"
        };
        if (Feature == TestScenario.mainBranch)
        {
            args.Add(@"-p:NuGetRestoreTargets=" + beforeTargets);
        }
        else
        {
            args.Add(@"-p:NuGetRestoreTargets=" + afterTargets);
            if (Feature == TestScenario.enable)
            {
                args.Add("-p:NuGetAudit=enable");
            }
        }

        processArguments = string.Join(" ", args);
    }

    [Benchmark]
    public void NuGetClient()
    {
        Restore(NuGetClientRepoPath);
    }

    [Benchmark]
    public void Orleans()
    {
        Restore(OrleansRepoPath);
    }

    [Benchmark]
    public void OrchardCore()
    {
        Restore(OrchardCoreRepoPath);
    }

    private void Restore(string repo)
    {
        var processInfo = new ProcessStartInfo("msbuild", processArguments);
        processInfo.WorkingDirectory = repo;
        processInfo.RedirectStandardOutput = true;
        processInfo.RedirectStandardError = true;

        using Process process = Process.Start(processInfo)!;
        DrainStream(process.StandardOutput);
        DrainStream(process.StandardError);
        process.WaitForExit();
    }

    private void DrainStream(StreamReader stream)
    {
        Task.Run(() =>
        {
            try
            {
                var buffer = new char[4096];
                while (true)
                {
                    if (stream.Read(buffer, 0, buffer.Length) <= 0)
                    {
                        return;
                    }
                }
            }
            catch
            {
            }
        });
    }
}
```

</div>
</summary>

src/NuGet.Core/NuGet.DependencyResolver.Core/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,3 @@
 #nullable enable
+~NuGet.DependencyResolver.IRemoteDependencyProvider.SourceRepository.get -> NuGet.Protocol.Core.Types.SourceRepository
+~NuGet.DependencyResolver.LocalDependencyProvider.SourceRepository.get -> NuGet.Protocol.Core.Types.SourceRepository
```
zivkan, line 3: These Public API isn't needed if we break RestoreCommandProvider's constructor. But considering that `PackageSource` is already public on `IRemoteDependencyProvider`, adding `SourceRepository` doesn't seem unreasonable to me.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -275,14 +277,27 @@ public async Task<RestoreResult> ExecuteAsync(CancellationToken token)
                     });
                 }
 
+                if (string.Equals(_request.Project.RestoreMetadata?.RestoreAuditProperties?.EnableAudit, "enable", StringComparison.InvariantCultureIgnoreCase))
```
zivkan: @JonDouglas Nuget already has a bunch of MSBuild properties which use `true` and `false` for boolean values.  I kept `enable` here, since that's what the spec called for, but I just wanted to call it out. While implementing this, I'm wondering if it has implications on project systems that want to make a properties page to change the value. They probably already have controls for true/false, so we might be causing them extra work if we want to keep "enable", in addition to any possible customer confusion about why this is the only "boolean like" property that doesn't use "true".

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -0,0 +1,286 @@
...
+            {
+                CheckPackageVulnerabilities(allVulnerabilityData.KnownVulnerabilities);
+            }
+        }
+
+        private void ReplayErrors(AggregateException exceptions)
+        {
+            foreach (Exception exception in exceptions.InnerExceptions)
+            {
+                string messageText = "Error occurred while getting package vulnerability data: " + exception.Message;
```
zivkan: 🤦 This sting needs to be localized.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -0,0 +1,286 @@
...
+            switch (severity)
+            {
+                case 1:
+                    return (Strings.Vulnerability_Severity_1, NuGetLogCode.NU1901);
+                case 2:
+                    return (Strings.Vulnerability_Severity_2, NuGetLogCode.NU1902);
+                case 3:
+                    return (Strings.Vulnerability_Severity_3, NuGetLogCode.NU1903);
+                case 4:
+                    return (Strings.Vulnerability_Severity_1, NuGetLogCode.NU1901);
```
zivkan: copy paste error 😞

src/NuGet.Core/NuGet.ProjectModel/ProjectRestoreMetadata.cs
```diff
@@ -12,6 +12,10 @@ namespace NuGet.ProjectModel
 {
     public class ProjectRestoreMetadata : IEquatable<ProjectRestoreMetadata>
     {
+        public ProjectRestoreMetadata()
```
zivkan: This was useful to find where the class instances were created, but can be deleted now.

src/NuGet.Core/NuGet.ProjectModel/ProjectRestoreMetadata.cs
```diff
@@ -223,7 +230,8 @@ public bool Equals(ProjectRestoreMetadata other)
                    EqualityUtility.EqualsWithNullCheck(RestoreLockProperties, other.RestoreLockProperties) &&
                    EqualityUtility.EqualsWithNullCheck(CentralPackageVersionsEnabled, other.CentralPackageVersionsEnabled) &&
                    EqualityUtility.EqualsWithNullCheck(CentralPackageVersionOverrideDisabled, other.CentralPackageVersionOverrideDisabled) &&
-                   EqualityUtility.EqualsWithNullCheck(CentralPackageTransitivePinningEnabled, other.CentralPackageTransitivePinningEnabled);
+                   EqualityUtility.EqualsWithNullCheck(CentralPackageTransitivePinningEnabled, other.CentralPackageTransitivePinningEnabled) &&
+                   RestoreAuditProperties == other.RestoreAuditProperties;
```
zivkan, line 230: I guess this should use the same helper function as everything else?  The helper function seems a bit pointless though. It's longer than a direct comparison 😕

zivkan: Actually, looking further up the file, there are plenty of other properties that compare directly with `==`, so no need to change this.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
jeffkl: I would vote for `true/false` or at least check for `true` or `enable`.

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
```diff
@@ -849,6 +849,8 @@ Copyright (c) .NET Foundation. All rights reserved.
         <_CentralPackageVersionsEnabled>$(_CentralPackageVersionsEnabled)</_CentralPackageVersionsEnabled>
         <CentralPackageVersionOverrideEnabled>$(CentralPackageVersionOverrideEnabled)</CentralPackageVersionOverrideEnabled>
         <CentralPackageTransitivePinningEnabled>$(CentralPackageTransitivePinningEnabled)</CentralPackageTransitivePinningEnabled>
+        <NuGetAudit>$(NuGetAudit)</NuGetAudit>
+        <NuGetAuditLevel>$(NuGetAuditLevel)</NuGetAuditLevel>
```
jeffkl, line 853: Do you need to update static graph-based restore as well?

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -0,0 +1,287 @@
...
+        private readonly IReadOnlyList<IVulnerabilityInformationProvider> _vulnerabilityInfoProviders;
+        private readonly ILogger _logger;
+
+        public AuditUtility(
+            ProjectModel.RestoreAuditProperties restoreAuditProperties,
+            string projectFullPath,
+            IEnumerable<RestoreTargetGraph> graphs,
+            IReadOnlyList<IVulnerabilityInformationProvider> vulnerabilityInformationProviders,
+            ILogger logger)
+        {
```
jeffkl, line 33: Should any of these inputs be checked for `null`?  It looks like only our code will new it up but those inputs might come from a 3rd party caller.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -0,0 +1,287 @@
...
+                            package.Id,
+                            affectedGraphs.OrderBy(s => s).ToArray());
+                        restoreLogMessage.ProjectPath = _projectFullPath;
+                        _logger.Log(restoreLogMessage);
+                    }
+                }
+            }
+        }
+
+        private static List<PackageVulnerabilityInfo>? GetKnownVulnerabilities(
```
jeffkl, line 101: Is it worth having the methods in this class return read-only versions of the collections?
```suggestion
        private static ReadOnlyCollection<PackageVulnerabilityInfo>? GetKnownVulnerabilities(
```

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -0,0 +1,287 @@
...
+            GetVulnerabilityInfoResult? final =
+                knownVulnerabilities != null || errors != null
+                ? new(knownVulnerabilities, errors != null ? new AggregateException(errors) : null)
+                : null;
+            return final;
+        }
+
+        private int ParseAuditLevel()
+        {
+            string? auditLevel = _restoreAuditProperties.AuditLevel;
```
jeffkl: Do we need to `.Trim()` the value?

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
```diff
@@ -891,6 +895,23 @@ private static RestoreLockProperties GetRestoreLockProperites(IMSBuildItem specI
                 IsPropertyTrue(specItem, "RestoreLockedMode"));
         }
 
+        private static RestoreAuditProperties GetRestoreAuditProperties(IMSBuildItem specItem)
+        {
+            string enableAudit = specItem.GetProperty("NuGetAudit");
+            string auditLevel = specItem.GetProperty("NuGetAuditLevel");
```
jeffkl, line 901: Can we use the constants in this context?

```suggestion
            string enableAudit = specItem.GetProperty(ProjectBuildProperties.NuGetAudit);
            string auditLevel = specItem.GetProperty(ProjectBuildProperties.NuGetAuditLevel);
```

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommandProviders.cs
```diff
@@ -118,12 +147,20 @@ public class RestoreCommandProviders
                 remoteProviders.Add(provider);
             }
 
+            var vulnerabilityInfoProviders = new List<IVulnerabilityInformationProvider>(sources.Count());
```
donnie-msft: Just an observation: I suppose the work to filter sources upfront to only ones which provide vulnerability info is worse than just creating providers and letting a `null` come back when vulnerabilities are requested.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommandProviders.cs
```diff
@@ -118,12 +147,20 @@ public class RestoreCommandProviders
...
+            foreach (var source in sources)
```
donnie-msft: use explicit type

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommandProvidersCache.cs
```diff
@@ -115,7 +117,14 @@ public class RestoreCommandProvidersCache
                 remoteProviders.Add(remoteProvider);
             }
 
-            return new RestoreCommandProviders(globalCache, fallbackFolders, localProviders, remoteProviders, _fileCache);
+            var vulnerabilityInfoProviders = new List<IVulnerabilityInformationProvider>(sources.Count);
+            foreach (var source in sources)
```
donnie-msft: use explicit type

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommandProvidersCache.cs
```diff
@@ -115,7 +117,14 @@ public class RestoreCommandProvidersCache
                 remoteProviders.Add(remoteProvider);
             }
 
-            return new RestoreCommandProviders(globalCache, fallbackFolders, localProviders, remoteProviders, _fileCache);
+            var vulnerabilityInfoProviders = new List<IVulnerabilityInformationProvider>(sources.Count);
```
donnie-msft: nit...
```suggestion
            var vulnerabilityInformationProviders = new List<IVulnerabilityInformationProvider>(sources.Count);
```

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -275,14 +277,27 @@ public async Task<RestoreResult> ExecuteAsync(CancellationToken token)
...
+                {
+                    telemetry.StartIntervalMeasure();
+                    AuditUtility audit = new AuditUtility(
```
donnie-msft, line 285: use implicit type

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -0,0 +1,287 @@
...
+            _vulnerabilityInfoProviders = vulnerabilityInformationProviders;
+            _logger = logger;
+        }
+
+        public async Task CheckPackageVulnerabilitiesAsync(CancellationToken cancellationToken)
+        {
+            GetVulnerabilityInfoResult? allVulnerabilityData = await GetAllVulnerabilityDataAsync(cancellationToken);
+            if (allVulnerabilityData == null) return;
+
+
```
donnie-msft: extra line

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -0,0 +1,287 @@
...
+            {
+                CheckPackageVulnerabilities(allVulnerabilityData.KnownVulnerabilities);
+            }
+        }
+
+        private void ReplayErrors(AggregateException exceptions)
+        {
+            foreach (Exception exception in exceptions.InnerExceptions)
+            {
+                string messageText = string.Format(Strings.Error_VulnerabilityDataFetch, exception.Message);
```
donnie-msft, line 61: use implicit type

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -0,0 +1,287 @@
...
+                _logger.Log(logMessage);
+            }
+        }
+
+        private void CheckPackageVulnerabilities(IReadOnlyList<IReadOnlyDictionary<string, IReadOnlyList<PackageVulnerabilityInfo>>> knownVulnerabilities)
+        {
+            Dictionary<PackageIdentity, Dictionary<PackageVulnerabilityInfo, List<string>>>? packagesWithKnownVulnerabilities =
+                FindPackagesWithKnownVulnerabilities(knownVulnerabilities);
+
+            if (packagesWithKnownVulnerabilities != null)
```
donnie-msft: nit/style consideration to avoid indentation: just return if null.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -0,0 +1,287 @@
...
+            IReadOnlyList<IReadOnlyDictionary<string, IReadOnlyList<PackageVulnerabilityInfo>>> knownVulnerabilities)
+        {
+            // multi-targeting projects often use the same package across multiple TFMs, so group to reduce output spam.
+            Dictionary<PackageIdentity, Dictionary<PackageVulnerabilityInfo, List<string>>>? result = null;
+
+            int minSeverity = ParseAuditLevel();
+
+            foreach (var graph in _targetGraphs)
+            {
+                foreach (LibraryIdentity package in graph.Flattened.Select(i => i.Key).Where(p => p.Type == "package"))
```
donnie-msft:
```suggestion
                foreach (LibraryIdentity package in graph.Flattened.Select(i => i.Key).Where(p => p.Type == LibraryType.Project ))
```

Maybe?

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -0,0 +1,287 @@
...
+                                result.Add(packageIdentity, knownPackageVulnerabilities);
+                            }
+
+                            if (!knownPackageVulnerabilities.TryGetValue(knownVulnerability, out List<string>? affectedGraphs))
+                            {
+                                affectedGraphs = new();
+                                knownPackageVulnerabilities.Add(knownVulnerability, affectedGraphs);
+                            }
+
+                            // Multiple package sources might list the same known vulnerability, so de-duple those too.
```
donnie-msft:
```suggestion
                            // Multiple package sources might list the same known vulnerability, so de-dupe those too.
```

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -0,0 +1,287 @@
...
+            if (cancellationToken.IsCancellationRequested)
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+            }
+
+            List<Exception>? errors = null;
+            List<IReadOnlyDictionary<string, IReadOnlyList<PackageVulnerabilityInfo>>>? knownVulnerabilities = null;
+            foreach (var resultTask in results)
+            {
+                var result = await resultTask;
```
donnie-msft: use explicit type

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
donnie-msft: I'd almost rather that happen in the `setter` for `AuditLevel` if so, that way all the `.Equals` logic/etc gets the benefit.

src/NuGet.Core/NuGet.Commands/Strings.resx
```diff
@@ -1041,4 +1041,41 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
...
+  </data>
+  <data name="Vulnerability_Severity_unknown" xml:space="preserve">
+    <value>unknown</value>
+    <comment>As in "This package has a unknown severity vulnerability"</comment>
+  </data>
+  <data name="Warning_PackageWithKnownVulnerability" xml:space="preserve">
+    <value>Package '{0}' {1} has a known {2} severity vulnerability, {3}.</value>
+    <comment>{0} - package id
+{1} - package version
+{2} - severity, for example low, moder, high, or critical
```
donnie-msft:
```suggestion
{2} - severity, for example low, moderate, high, or critical
```
also should the severities be translated here? Or not, as Error_InvalidNuGetAuditLevelValue is not?

src/NuGet.Core/NuGet.Commands/Strings.resx
```diff
@@ -1041,4 +1041,41 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
     <value>Warning As Error: {0}</value>
     <comment>{0} is a warning message</comment>
   </data>
+  <data name="Vulnerability_Severity_1" xml:space="preserve">
+    <value>low</value>
+    <comment>As in "This package has a low severity vulnerability"</comment>
```
donnie-msft, line 1046: Should the severities be translated here? Or not, as Error_InvalidNuGetAuditLevelValue is not?

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.cs
```diff
@@ -1039,6 +1040,30 @@ private static void ReadMSBuildMetadata(JsonTextReader jsonReader, PackageSpec p
...
+                            }
+                        });
+                        auditProperties = new RestoreAuditProperties()
+                        {
+                            EnableAudit = enableAudit,
+                            AuditLevel = auditLevel
+                        };
+                        break;
+
+
```
donnie-msft: extra line

src/NuGet.Core/NuGet.ProjectModel/PackageSpecWriter.cs
```diff
@@ -163,6 +161,9 @@ private static void SetMSBuildMetadata(IObjectWriter writer, PackageSpec package
             // write NuGet lock file msbuild properties
             WriteNuGetLockFileProperties(writer, msbuildMetadata);
 
+            // write NuGetAudit properties
```
donnie-msft: Can omit comment

src/NuGet.Core/NuGet.ProjectModel/PackageSpecWriter.cs
```diff
@@ -200,6 +201,18 @@ private static void WriteNuGetLockFileProperties(IObjectWriter writer, ProjectRe
             }
         }
 
+        private static void WriteNuGetAuditProperties(IObjectWriter writer, RestoreAuditProperties auditProperties)
+        {
+            if (auditProperties == null) { return; }
```
donnie-msft: No braces, or new lines for each brace.

src/NuGet.Core/NuGet.ProjectModel/RestoreAuditProperties.cs
```diff
@@ -0,0 +1,70 @@
...
+{
+    public class RestoreAuditProperties : IEquatable<RestoreAuditProperties>
+    {
+        /// <summary>
+        /// Gets or sets a value indicating whether NuGet audit (check packages for known vulnerabilities) is enabled.
+        /// </summary>
+        public string? EnableAudit { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating the NuGet audit level threshold which which vulnerabilities are reported.
```
donnie-msft:
```suggestion
        /// Gets or sets a value indicating the NuGet audit level threshold with which vulnerabilities are reported.
```

src/NuGet.Core/NuGet.ProjectModel/RestoreAuditProperties.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// </summary>
+        /// <value>low, moderate, high, critical</value>
+        public string? AuditLevel { get; set; }
+
+        public bool Equals(RestoreAuditProperties? other)
+        {
+            if (other is null) { return false; }
```
donnie-msft: new lines for braces

src/NuGet.Core/NuGet.ProjectModel/RestoreAuditProperties.cs
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        internal RestoreAuditProperties Clone()
+        {
+            var clone = new RestoreAuditProperties()
+            {
+                EnableAudit = EnableAudit,
+                AuditLevel = AuditLevel
+            };
+            Debug.Assert(clone == this && clone.GetHashCode() == GetHashCode());
```
donnie-msft: I'm not sure I follow how the hash codes could be different if Equals is true. Not important to discuss now if you're sure about it.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommandProviders.cs
kartheekp-ms: Would it be possible to replace the usage of `sources.Count()` with `remoteProviders.Count` since the latter has a constant time complexity?

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
zivkan: The struct is `internal`, so it's not (yet?) a public API. If it was a public API, then yes, definitely. However, I think that `RestoreTargetGraph` is so specific to restore that there isn't a benefit to making it public for other developers to use in their own apps. So ultimately, I chose to be conservative, because it's easy to add APIs in the future, but it's hard to remove APIs when we make bug fixes or add new features.

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
zivkan: Yes, yes I did. I added it in a new commit.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
zivkan: `List` overrides `GetEnumerator` to return a struct enumerator, so `foreach` loops do not allocate. If we switch to `ReadOnlyCollection`, then it will return `IEnumerable`, which means it will allocate.  Since restore is our main perf hot path, I  was trying to minimize allocations. If the method was public, then I'd probably agree, but it's private, so I think it's reasonable to optimize for perf.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
zivkan: RestoreAuditProperties comes from the PackageSpec, which is the input parameter for the entire restore process. In other words, if we validate and throw exceptions when settings `RestoreAuditProperties.AuditLevel`, then no packages will be restored, no vulnerabilities will be checked. If there is anything else that would output warnings or errors, they won't be output until the customer fixes their project file first and then runs again.  I don't think that's a good experience for a dev tool (the compiler doesn't stop at the first syntax error), hence the PackageSpec data structure is a faithful representation of project inputs, and validation happens here.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
zivkan: Unfortunately no. `ProjectBuildProperties` is defined in NuGet.PackageManagement, which has a ProjectReference to NuGet.Commands. Therefore, we cannot add a circular reference from NuGet.Commands back to NuGet.PackageManagement in order to use the API.

Why the const strings are defined in NuGet.PackageManagement instead of NuGet.Commands, I don't know. But I'm not a "reduce duplication at all costs!!!!!" purist. Once we ship a version of NuGet that uses "NuGetAudit" and "NuGetAuditLevel" MSBuild properties, it becomes part of our public API, and therefore effectively set in stone permanently. The benefit to reducing duplication is when a value changes, you don't have to change multiple places. But this value will never change, so there's effectively little benefit.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommandProviders.cs
zivkan: Yes. In case of a no-op restore, we'll never check the servers if they have vulnerability data.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
zivkan: Why?  While I don't usually agree with "don't use var", at least the theoretical benefit is to improve understanding of what datatype the reader might not know. But I don't understand the benefit of using implicit type, especially when the type has a short name.

src/NuGet.Core/NuGet.Commands/Strings.resx
zivkan: Either, this is the translation for the low severity (see the `<value>` element). You're commenting on the `<comment>`, which in theory is for translators to get additional context, although I suspect that everything is machine translated so the comments are probably useless.  Otherwise, I don't understand your question.

src/NuGet.Core/NuGet.Commands/Strings.resx
zivkan: I'm not sure I understand the question. The severity here is in the resource comment. The severity is translated in the resources above (Vulnerability_Severity_*), and is substitued for `{2}` as per this resource's comment.

src/NuGet.Core/NuGet.ProjectModel/RestoreAuditProperties.cs
zivkan: It was to help catch bugs in `==` and `GetHashCode`, as much as bugs in Clone itself. But I'll have a follow up PR with unit tests, so we won't need this in the product code.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
donnie-msft: > if we validate and throw exceptions 

`Trim()` is sanitation, not validation. How would `Trim()` cause an exception to be thrown?

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
donnie-msft: Is it different than this example in # 10 of https://github.com/NuGet/NuGet.Client/blob/dev/docs/coding-guidelines.md#c-coding-style?

> The following are incorrect: `List<Fruit> fruits = new List<Fruit>();`
> the following are correct: `var fruits = new List<Fruit>();`

src/NuGet.Core/NuGet.Commands/Strings.resx
donnie-msft: Edit: Nevermind, yes I see this is just a parameter

src/NuGet.Core/NuGet.Commands/Strings.resx
donnie-msft: Edit: Nevermind, I see it's the severity itself.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
zivkan: If the only reason to change the code is because of the coding guidelines say to do it this way, without justifying why it benefits the repo, then the coding guidelines document probably should be changed.

Some of the guidelines help reduce churn by being opinionated about whitespace, so people don't keep changing whitespace between PRs. Some guidelines help reduce risk of bugs. Some help improve readability. But I don't feel like the "don't use var" guideline does any of those things

donnie-msft: If you feel our coding guidelines don't benefit our repo, then certainly propose your changes and rationale for that. However, since they're in place now, I think we should follow them, just as we ask our community members to do.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
donnie-msft: It looks like you're going to error right now for "low " with a message using NuGetLogCode.NU1014. It also sounds like you're confident we should let this happen instead of sanitizing, so I'll resolve my part of the comment.

donnie-msft (approved):
Think most of my comments are resolved. Consider either following the guideline # 10 or creating an issue to discuss with the team your proposal to change it.

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
```diff
@@ -867,6 +870,20 @@ private PackageSpec GetPackageSpec(IMSBuildProject project, IReadOnlyDictionary<
             return (restoreMetadata, targetFrameworkInfos);
         }
 
+        private RestoreAuditProperties GetRestoreAuditProperties(IMSBuildProject project)
```
jeffkl, line 873: static graph-based restore has access to MSBuildRestoreUtility if you want to make `GetRestoreAuditProperties` internal to re-use that code

jeffkl (approved):

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
JonDouglas: This is likely oversight on my part. When I wrote that, I was looking at a minimal sdk-style project and the only new features being enabled(in templates) had this `enable/disable` usage. Upon looking at https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props I believe 99% of properties use true/enable.

Let's do what is most consistent.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
```diff
@@ -406,6 +406,16 @@ private async Task<PackageSpec> GetPackageSpecAsync(ISettings settings)
...
 
+            string enableAudit = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAudit);
+            string auditLevel = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditLevel);
+            RestoreAuditProperties auditProperties = !string.IsNullOrEmpty(enableAudit) || !string.IsNullOrEmpty(auditLevel)
+                ? new RestoreAuditProperties()
+                {
+                    EnableAudit = enableAudit,
+                    AuditLevel = auditLevel
+                }
+                : null;
```
kartheekp-ms, line 417: I am curious if there is any specific reason to use `null` in this case. How about using `false` instead?

```cs
RestoreAuditProperties auditProperties = new RestoreAuditProperties()
{
    EnableAudit = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAudit) ?? false,
    AuditLevel = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditLevel) ?? null
};
```

src/NuGet.Core/NuGet.Commands/RestoreCommand/VulnerabilityInformationProvider.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        public VulnerabilityInformationProvider(SourceRepository source, ILogger logger)
+        {
+            _source = source;
+            _logger = logger;
+
+            _vulnerabilityInfo = new AsyncLazy<GetVulnerabilityInfoResult?>(GetVulnerabilityInfoAsync);
+        }
+
+        public async Task<GetVulnerabilityInfoResult?> GetVulnerabilityInformationAsync(CancellationToken cancellationToken)
```
kartheekp-ms, line 30: If I understand correctly, `cancellationToken` paramter is unused here.

src/NuGet.Core/NuGet.Commands/RestoreCommand/VulnerabilityInformationProvider.cs
```diff
@@ -0,0 +1,50 @@
...
+using NuGet.Protocol;
+using NuGet.Protocol.Core.Types;
+using NuGet.Protocol.Model;
+
+namespace NuGet.Commands
+{
+    internal sealed class VulnerabilityInformationProvider : IVulnerabilityInformationProvider
+    {
+        private SourceRepository _source;
+        private ILogger _logger;
```
kartheekp-ms, line 18: I think these 2 variables can be marked as readonly

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
```diff
@@ -406,6 +406,16 @@ private async Task<PackageSpec> GetPackageSpecAsync(ISettings settings)
                 }
             }
 
+            string enableAudit = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAudit);
+            string auditLevel = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditLevel);
+            RestoreAuditProperties auditProperties = !string.IsNullOrEmpty(enableAudit) || !string.IsNullOrEmpty(auditLevel)
```
kartheekp-ms, line 411: I am curious what happens when customers set `auditLevel` and don't set `enableAudit` by mistake. As per the current logic, an instance of `RestoreAuditProperties` type will be created if one of them is set.

kartheekp-ms (commented):
I'm currently in the process of reviewing these changes and have made a few initial comments to begin with. Thank you for your patience.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
zivkan: I did it this way for a two main reasons.

1. The method `GetPackageSpecAsync` returns a `PackageSpec` object, whose purpose is to have an abstracted representation of project inputs. If the project doesn't explicitly disable the feature, then setting the value to false isn't exactly an accurate representation of the project inputs
2. If we want to be able to do A/B experiments, or a checkbox in VS options to enable this, it becomes important to know the difference between "project opts-out" and "project is using defaults". Having the PackageSpec pass through the (lack of) actual value allows us to implement that in a single place. Otherwise, we'd have to implement it in every project system we generate a PackageSpec.

src/NuGet.Core/NuGet.Commands/RestoreCommand/VulnerabilityInformationProvider.cs
zivkan: Currently unused due to the implementation details. But it means that cancellation isn't "instant", and that fetching the vulnerability database from a remote HTTP server can't be cancelled. If we want to fix this in the future, this API will need a cancellation token. When I first wrote this, I thought this class (well, the interface it implements) was going to have to be public, and I wanted to make sure we didn't need to change the public API. However, in the end I found a way to keep it all internal, so it's less justifyable.  I do have an idea how we can implement a `CancellableAsyncLazy`, but it's a non-trivial implementation and it's a perf-improvement that can be done in the future. But from an API design point of view, methods that do synchronous work shouldn't be async, and async methods should be cancellable, so a cancellation token seems like the right API choice. 🤷

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
zivkan: As stated previously, a `PackageSpec` is an abstracted representation of project inputs. If that's how the project is authored, then it's an accurate representation. It also ties to your other comment, because if at some point we flip the default so that audit is enabled by default, unless explicitly opt-out, then you'd want the `RestoreAuditProperties` to contain the AuditLevel, even hwne `EnableAudit` is null/false.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
zivkan: > Trim() is sanitation, not validation. How would Trim() cause an exception to be thrown?

Oops, I misunderstood. My thoughts were elsewhere.  I'll have a look how other parts of our "create a PackageSpec instance" code handles trimming.

zivkan:
I'm merging this to ensure it can be inserted into .NET and VS ASAP, but I'll address more of the comments in a follow-up PR, as I feel like the remaining comments aren't likely to lead to a significant change to public APIs or general software design.

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
zivkan: static restore doesn't have access to access to NuGet.Commands internals, so the API has to be public 😞

src/NuGet.Core/NuGet.Commands/RestoreCommand/VulnerabilityInformationProvider.cs
Nigusu-Allehu: https://github.com/NuGet/Home/issues/13644

PR #5139, opened 2023-04-20
Be more permissive with explicit type in coding guidelines

zivkan:
## Bug

Fixes: N/A

Regression? no

## Description

Personally, I don't agree with "don't use var when it's not obvious", but I can understand that some people want more certainly about the data type when reading code than I do.

However, I can't see any benefit to preventing explicit types when it's obvious. For example, `string something = "value"`. Since I don't see any downside, I propose that we officially allow it.

## PR Checklist

- [ ] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/docs/coding-guidelines.md
+++ b/docs/coding-guidelines.md
@@ -48,7 +48,7 @@ The general rule we follow is "use Visual Studio defaults".
 
 1. If a file happens to differ in style from these guidelines (e.g. private members are named `m_member` rather than `_member`), the existing style in that file takes precedence. Changes/refactorings are possible, but depending on the complexity, change frequency of the file, might need to be considered on their own merits in a separate pull request.
 
-1. Don't use `var` when it's not obvious what the variable type is.
+1. Explicitly state the data type, and avoid using `var`, in variable declarations where the type is not obvious.
 For example the following are correct:
 
     ```cs
@@ -66,7 +66,7 @@ For example the following are correct:
     var flavor = fruit.GetFlavor();
     ```
 
-    The following are optional and can either use explicit type or var
+    Using an explicit type or `var` in the following examples are both permissible:
 
     ```cs
     string fruit = "Apple";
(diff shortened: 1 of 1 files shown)


docs/coding-guidelines.md
```diff
@@ -48,7 +48,7 @@ The general rule we follow is "use Visual Studio defaults".
 
 1. If a file happens to differ in style from these guidelines (e.g. private members are named `m_member` rather than `_member`), the existing style in that file takes precedence. Changes/refactorings are possible, but depending on the complexity, change frequency of the file, might need to be considered on their own merits in a separate pull request.
 
-1. We only use `var` when it's obvious what the variable type is.
+1. Don't use `var` when it's not obvious what the variable type is.
```
jebriede: Rephrasing to avoid the double use of "not" and phrasing as more what we should do.

```suggestion
1. Explicitly state the data type, and avoid using `var`, in variable declarations where the type is not obvious.
```

docs/coding-guidelines.md
```diff
@@ -60,10 +60,15 @@ For example the following are correct:
...
     ```
 
-    The following are incorrect:
+    The following is incorrect:
 
     ```cs
     var flavor = fruit.GetFlavor();
+    ```
+
+    The following are optional and can either use explicit type or var
```
jebriede:
```suggestion
    Using an explicit type or `var` in the following examples are both permissible:
```

jebriede (dismissed):
Thanks for updating this! Suggesting wording changes for clarity.

jeffkl (approved):

jebriede (approved):

martinrrm (approved):

heng-liu (approved):

donnie-msft (commented):
I don't understand the meaning of this PR, as it seems to just change words around that mean the same thing as before. I'd like at least one example to be added of what the delta is between what we have been doing and what you suggest we start doing.

zivkan:
@donnie-msft previously the coding guidelines explicitly said that `string fruit = "Apple";` and `List<Fruit> fruits = new List<Fruit>();` were incorrect. Now it's saying that these examples are permissible.

donnie-msft:
What is the purpose of documenting that using either an implicit or explicit type is up to your own discretion? Should we just remove the guideline?

PR #5141, opened 2023-04-20
Reduce allocations in VirtualFileInfo.Name

### LINKED ISSUE #12550: Reduce allocations in VirtualFileInfo.Name
`VisualFileInfo` is an implementation of `IFileInfo` used with `Microsoft.Extensions.FileSystemGlobbing`. It's used when matching files in the package to files listed in the `contentFiles` section of package `nuspec` files.

The property is getting the filename from a path, but it does so with `PhysicalPath.Split('/').LastOrDefault();`. Both `Split` and `LastOrDefault` cause allocations, plus the result is not cached and debugging shows the same instance can have the `Name` getter called multiple times.

This code is used when generating the assets file when restoring a PackageReference project that uses a package that has `contentFiles/*` assets.

zivkan:
## Bug

Fixes: internal bug 1806171
Fixes: https://github.com/NuGet/Home/issues/12550

Regression? no

## Description

The previous code was doing a one-liner `PhysicalPath.Split('/').LastOrDefault();`. However, this causes the following allocations:

* An array for the `Split` results
* One string per array index (hopefully zero allocation when the array length is 1)
* An enumerator for `LastOrDefault`

Additionally, it wasn't caching the result, and running in the debugger shows that many instances have `Name` called multiple times during execution. Prism was showing this at a top allocation impact in VS.

So, the fix is to find the start index of the required string, which is zero allocation, and then get the substring as the only allocation in the method body. When the `PhysicalPath` doesn't contain any `/` characters, the code is now zero-allocation.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception: perf improvement, no functional changes
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/ContentFiles/VirtualFileInfo.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/ContentFiles/VirtualFileInfo.cs
@@ -59,11 +59,24 @@ namespace NuGet.Commands
             }
         }
 
+        private string _name;
         public string Name
         {
             get
             {
-                return PhysicalPath.Split('/').LastOrDefault();
+                if (_name == null)
+                {
+                    int lastSlashIndex = PhysicalPath.LastIndexOf('/');
+                    if (lastSlashIndex >= 0)
+                    {
+                        _name = PhysicalPath.Substring(lastSlashIndex + 1);
+                    }
+                    else
+                    {
+                        _name = PhysicalPath;
+                    }
+                }
+                return _name;
             }
         }
 


martinrrm (approved):

jeffkl (approved):

src/NuGet.Core/NuGet.Commands/RestoreCommand/ContentFiles/VirtualFileInfo.cs
```diff
@@ -59,11 +59,24 @@ public long Length
...
+                {
+                    int lastSlashIndex = PhysicalPath.LastIndexOf('/');
+                    if (lastSlashIndex >= 0)
+                    {
+                        _name = PhysicalPath.Substring(lastSlashIndex + 1);
+                    }
+                    else
+                    {
+                        _name = PhysicalPath;
+                    }
```
kartheekp-ms, line 77: Perhaps we can use the [Path.GetFileName](https://learn.microsoft.com/en-us/dotnet/api/system.io.path.getfilename?view=net-8.0#system-io-path-getfilename(system-string)) method, which retrieves the file name and extension from a given path - this is exactly what we need to do. Please let me know if my understanding is incorrect.

zivkan: `Path.GetFileName` will also use `\` as a separator, at least on Windows. Whereas currently only `/` is used. Therefore it would be a change in behaviour, and it might cause NuGet on Windows to work differently to Linux/Mac (when a package was "incorrectly" packed with Windows `\` separators rather than `/` as pack normalizes to).

It's an edge case, but these 9 lines of code aren't so complex that I feel it's worthwhile investigating this more deeply to understand whether it will change any behaviour.

kartheekp-ms (approved):

PR #5144, opened 2023-04-21
Build and test with net8.0 by default on CI only

zivkan:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2171

Regression? No

## Description

https://github.com/NuGet/NuGet.Client/pull/5137 added `net8.0` to the build. However, there is no version of Visual Studio at this time which installs the .NET 8 SDK. This makes it harder for external contributors to contribute, as well as NuGet team members who don't want to manually install and update a .NET SDK version that isn't managed by VS's installer.

Since NuGet's dev branch isn't yet targeting .NET 8, it's still being inserted primarily into .NET 7, and then flowing into .NET 8 previews because we don't have a .NET 8 specific branch, I believe this is low risk or impact.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/build/common.project.props
+++ b/build/common.project.props
@@ -15,8 +15,8 @@
     <NETFXTargetFramework>net472</NETFXTargetFramework>
     <NETCoreTargetFramework>netcoreapp3.1</NETCoreTargetFramework>
     <NETCoreTargetFramework Condition="'$(DotNetBuildFromSource)' == 'true'">net7.0</NETCoreTargetFramework>
-    <NETCoreTestTargetFrameworks Condition=" '$(NUGET_CLIENT_VS_INSTALLED_TOOLS_ONLY)' != 'true' ">net7.0;net8.0</NETCoreTestTargetFrameworks>
-    <NETCoreTestTargetFrameworks Condition=" '$(NUGET_CLIENT_VS_INSTALLED_TOOLS_ONLY)' == 'true' ">net7.0</NETCoreTestTargetFrameworks>
+    <NETCoreTestTargetFrameworks>net7.0</NETCoreTestTargetFrameworks>
+    <NETCoreTestTargetFrameworks Condition=" ('$(CI)' == 'true' AND '$(BUILD_NET8)' != 'false') OR '$(BUILD_NET8)' == 'true' ">net7.0;net8.0</NETCoreTestTargetFrameworks>
     <NetStandardVersion>netstandard2.0</NetStandardVersion>
     <TargetFrameworksExe>$(NETFXTargetFramework);$(NETCoreTargetFramework)</TargetFrameworksExe>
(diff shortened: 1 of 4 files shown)


build/common.project.props
```diff
@@ -15,7 +15,8 @@
     <NETFXTargetFramework>net472</NETFXTargetFramework>
     <NETCoreTargetFramework>netcoreapp3.1</NETCoreTargetFramework>
     <NETCoreTargetFramework Condition="'$(DotNetBuildFromSource)' == 'true'">net7.0</NETCoreTargetFramework>
-    <NETCoreTestTargetFrameworks>net7.0;net8.0</NETCoreTestTargetFrameworks>
+    <NETCoreTestTargetFrameworks Condition=" '$(NUGET_CLIENT_VS_INSTALLED_TOOLS_ONLY)' != 'true' ">net7.0;net8.0</NETCoreTestTargetFrameworks>
+    <NETCoreTestTargetFrameworks Condition=" '$(NUGET_CLIENT_VS_INSTALLED_TOOLS_ONLY)' == 'true' ">net7.0</NETCoreTestTargetFrameworks>
```
zivkan: I'm open to suggestions for better environment variable names.  `NUGET_CLIENT_VS_INSTALLED_TOOLS_ONLY` isn't great. It's not even good. But I thought it should contain the name of our repo to reduce risk we use a name that conflicts with other software, and then the rest of my proposed name is descriptive 🤷

martinrrm (dismissed):

build/common.project.props
jeffkl: Hopefully its just temporary so the name doesn't matter.

jeffkl (dismissed):

jeffkl (dismissed):

jeffkl (dismissed):

eng/pipelines/templates/pipeline.yml
```diff
@@ -84,6 +84,7 @@ resources:
 
 variables:
   DOTNET_NOLOGO: 1
+  CI: 'true'
```
jeffkl: Was this not set before somewhere?  We have the CI-only facts and theories which is looking for this env var.  Now some tests are crashing on this PR build consistently.  I'm wondering if this change is making more tests run than before?

zivkan: It's set up as either a pipeline variable, or in our "variable group". I wanted to add it to the yaml just to be more explicit about which variables our CI depends on, but I really wasn't expecting any behaviour differences 😕

jeffkl (approved):

PR #5146, opened 2023-04-24
Show diagnostic beneath unresolved package/project reference in Solution Explorer

### LINKED ISSUE #8947: Unresolved packages/projects should show errors in dependency tree
The following project references a NuGet package that doesn't exist:

```xml

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="NonExistantPackage" Version="1.2.3" />
  </ItemGroup>

</Project>
```

`project.assets.json` has the following entry:

```json
  "logs": [
    {
      "code": "NU1101",
      "level": "Error",
      "message": "Unable to find package NonExistantPackage. No packages exist with this id in source(s): C:\\Program Files\\dotnet\\library-packs, Microsoft Visual Studio Offline Packages, NuGet local, nuget.org",
      "libraryId": "NonExistantPackage",
      "targetGraphs": [
        "net7.0"
      ]
    }
  ]
```

This message should be displayed in the tree, yet isn't.

image

The package node should have a child node containing the error message. Otherwise it is difficult for the user to know why the restore failed, or what the yellow triangle means.

### LINKED ISSUE #8998: When changing a package version, update rather than replace the tree node
The dependencies tree currently looks for the node to update based on its caption.

When we change a package `MyPackage` from version `1.0` to `2.0`, the caption changes from `MyPackage (1.0)` to `MyPackage (2.0)`.

The current code thinks these are different packages and therefore removes the old tree and attaches a new one. This has the effect of causing the package's node to collapse if it were expanded, which is annoying.

Note any change here will need a corresponding change on the NuGet.Client side, as that code doesn't currently update attached child nodes if the package's version changes.

drewnoakes:
## Bug

Fixes https://github.com/dotnet/project-system/issues/8947
Fixes https://github.com/dotnet/project-system/issues/8998

## Description

Previously, if a project referenced a non-existent package or project, the dependency would appear with a yellow triangle without any further explanation.

A log message would be present in the `project.assets.json` file, however we would not attach it to the tree.

This is because unresolved packages/projects such as these do not have entries in the `libraries` section of the lock file. This meant that we would not create an `IRelatableItem` for the library, and would then not be able to attach child nodes (such as diagnostics).

This change ensures that libaries mentioned in log messages have at least a dummy entry in the `LibraryByName` collection, so that tree construction completes successfully.

Another required change here was the normalization of unresolved project paths. In some instances these were identified with the relative path used in the project file (such as `..\OtherProject\OtherProject.csproj`), yet in the log messages the identity would be the full path. To fix this issue and allow diagnostics to be attached to unresolved project references, absolute library IDs are made relative to the project file during snapshot construction. It's possible to author a project file for which this would not work correctly, but that seems unlikely to occur very often in practice.

## Before

image

## After

image

When one of the diagnostic nodes is focussed, the _Properties_ window shows more information.

image

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/Models/AssetsFileLibraryType.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/Models/AssetsFileLibraryType.cs
@@ -11,6 +11,22 @@ namespace NuGet.VisualStudio.SolutionExplorer.Models
     internal enum AssetsFileLibraryType : byte
     {
         Package,
-        Project
+
+        Project,
+
+        /// <summary>
+        /// When it is unknown whether the library represents a package or a project.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// This occurs, for example, when a referenced item is not found. It will be present in the lock file's
+        /// messages, but not elsewhere in the lock file.
+        /// </para>
+        /// <para>
+        /// This enum member exists only to support attaching diagnostic items to their package or project reference
+        /// node in the tree.
+        /// </para>
+        /// </remarks>
+        Unknown
     }
 }
--- a/src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/AssetsFileTopLevelDependenciesCollectionSourceProvider.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/AssetsFileTopLevelDependenciesCollectionSourceProvider.cs
@@ -26,7 +26,7 @@ namespace NuGet.VisualStudio.SolutionExplorer
     /// <remarks>
     /// Templates out common code with a bunch of protected methods to override for specific item types.
     /// </remarks>
-    internal abstract class AssetsFileTopLevelDependenciesCollectionSourceProvider<TIdentity, TItem> : DependenciesAttachedCollectionSourceProviderBase
+    internal abstract class AssetsFileTopLevelDependenciesCollectionSourceProvider<TItem> : DependenciesAttachedCollectionSourceProviderBase
         where TItem : class, IRelatableItem
     {
         protected AssetsFileTopLevelDependenciesCollectionSourceProvider(ProjectTreeFlags flags)
@@ -34,9 +34,9 @@ namespace NuGet.VisualStudio.SolutionExplorer
         {
         }
 
-        protected abstract bool TryGetIdentity(Properties properties, out TIdentity identity);
+        protected abstract bool TryGetLibraryName(Properties properties, [NotNullWhen(returnValue: true)] out string? libraryName);
 
-        protected abstract bool TryGetLibrary(AssetsFileTarget target, TIdentity identity, [NotNullWhen(returnValue: true)] out AssetsFileTargetLibrary? library);
+        protected abstract bool TryGetLibrary(AssetsFileTarget target, string libraryName, [NotNullWhen(returnValue: true)] out AssetsFileTargetLibrary? library);
 
         protected abstract TItem CreateItem(AssetsFileTarget targetData, AssetsFileTargetLibrary library);
 
@@ -51,7 +51,7 @@ namespace NuGet.VisualStudio.SolutionExplorer
         {
             ThreadHelper.ThrowIfNotOnUIThread();
 
-            TIdentity identity;
+            string? libraryName;
 
             if (!ErrorHandler.Succeeded(hierarchyItem.HierarchyIdentity.Hierarchy.GetProperty(
                 hierarchyItem.HierarchyIdentity.ItemID, (int)__VSHPROPID.VSHPROPID_ExtObject, out object projectItemObject)))
@@ -63,7 +63,7 @@ namespace NuGet.VisualStudio.SolutionExplorer
             {
                 Properties? properties = (projectItemObject as ProjectItem)?.Properties;
 
-                if (properties == null || !TryGetIdentity(properties, out identity))
+                if (properties == null || !TryGetLibraryName(properties, out libraryName))
                 {
                     containsCollectionSource = null;
                     return false;
@@ -99,7 +99,7 @@ namespace NuGet.VisualStudio.SolutionExplorer
                     AssetsFileDependenciesSnapshot snapshot = versionedValue.Value;
                     if (snapshot.TryGetTarget(target, out AssetsFileTarget? targetData))
                     {
-                        if (TryGetLibrary(targetData, identity, out AssetsFileTargetLibrary? library))
+                        if (TryGetLibrary(targetData, libraryName, out AssetsFileTargetLibrary? library))
                         {
                             if (item == null)
                             {
--- a/src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/Items/PackageAssemblyItem.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/Items/PackageAssemblyItem.cs
@@ -67,7 +67,9 @@ namespace NuGet.VisualStudio.SolutionExplorer
             {
                 get
                 {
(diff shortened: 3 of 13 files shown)


drewnoakes:
@kartheekp-ms any feedback on this PR? I'd like to get it in early for 17.7p2 if possible.

I'll probably have one or two more PRs for tree nodes coming for that release too.

drewnoakes:
> I'll probably have one or two more PRs for tree nodes coming for that release too.

Actually the first of these will conflict with this one, so I'll extend this PR.

drewnoakes:
> Actually the first of these will conflict with this one, so I'll extend this PR.

Done in fe0f6053df5d204649a630b801cfb97d910f41ed:

> Don't include version in package identity
> 
> Changes to the Dependencies tree in 17.7 will mean that when changing a package's version, the tree node is no longer removed and re-added, but rather updated in place.
> 
> Before this change, the package's version was part of its identity. This meant new data for an updated version would not be applied to the node again. Previously this wasn't an issue, as we recreated the node. But now we re-use it, we want the data from a different version to be applied to the tree.
> 
> This change enables that.

src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/Models/AssetsFileLibraryType.cs
```diff
@@ -11,6 +11,22 @@ namespace NuGet.VisualStudio.SolutionExplorer.Models
...
+        /// <para>
+        /// This occurs, for example, when a referenced item is not found. It will be present in the lock file's
+        /// messages, but not elsewhere in the lock file.
+        /// </para>
+        /// <para>
+        /// This enum member exists only to support attaching diagnostic items to their package or project reference
+        /// node in the tree.
+        /// </para>
+        /// </remarks>
+        Unknown
```
kartheekp-ms, line 30: I think using `Unknown` can be confusing because the non-existant type is either `Package` or `Project`. To make the code easier to understand, we could consider adding a readonly property called `LibraryType` of type `AssetsFileLibraryType` to `AssetsFileLogMessage`. This would allow us to use `AssetsFileLogMessage.LibraryType` instead of `Unknown` when creating a placeholder or dummy `AssetFileTargetLibrary` object. By implementing this change, we could eliminate the need to check for `Unknown` in `TryGetPackage` or `TryGetProject` methods. What do you think? Please let me know if my understanding is incorrect.

kartheekp-ms (commented):
I tested the behavior introduced in this PR by launching the extension from your branch. To ensure that the code is reliable and robust, would it be possible to add some tests to this PR?

Also, I think you should rebase your branch on top of upstream dev branch to fix some CI failures.

src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/Models/AssetsFileLibraryType.cs
drewnoakes: Unfortunately we still wouldn't know what value to put in a `AssetsFileLogMessage.LibraryType` property. The `project.assets.json` file doesn't contain enough information to model that. Hence the `Unknown` value.

For example, here's are the log messages for a package and project, respectively:

```json
"logs": [
  {
    "code": "NU1101",
    "level": "Error",
    "message": "Unable to find package NonExistantPackage. No packages exist with this id in source(s): C:\\Program Files\\dotnet\\library-packs, Microsoft Visual Studio Offline Packages, NuGet local, nuget.org",
    "libraryId": "NonExistantPackage",
    "targetGraphs": [
      "net7.0"
    ]
  },
  {
    "code": "NU1104",
    "level": "Error",
    "message": "Unable to find project 'D:\\repos\\issues\\Issue8947UnresolvedPackageWarningMissing\\NonExistantProject\\NonExistantProject.csproj'. Check that the project reference is valid and that the project file exists.",
    "libraryId": "D:\\repos\\issues\\Issue8947UnresolvedPackageWarningMissing\\NonExistantProject\\NonExistantProject.csproj",
    "targetGraphs": [
      "net7.0"
    ]
  }
]
```

drewnoakes:
> would it be possible to add some tests to this PR?

There's not much actual logic here to test as the attached collections stuff is quite declarative. The most valuable testing for these code paths would be integration tests, which my team is investigating at the moment.

I've added some tests for the new parsing behaviour.

> Also, I think you should rebase your branch on top of upstream dev branch to fix some CI failures.

Done. I've pushed, but CI doesn't run automatically for my PRs unfortunately so I cannot tell whether it'll pass or not.

src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/Models/AssetsFileLibraryType.cs
kartheekp-ms: Sorry for the confusion. I should have been clearer. Given that `Path.IsRooted("libraryId")` will be `true` for projects but not for packages, I think we can identify whether the library is a package or project easily.

```cs
    /// <summary>
    /// Models a diagnostic found in the assets file. Immutable.
    /// </summary>
    internal readonly struct AssetsFileLogMessage
    {
      // remaining code
        public AssetFileTargetLibrary LibraryType { get; } 
        public AssetsFileLogMessage(string projectFilePath, IAssetsLogMessage logMessage)
        {
            Code = logMessage.Code;
            Level = logMessage.Level;
            WarningLevel = logMessage.WarningLevel;
            Message = logMessage.Message;
            LibraryName = NormalizeLibraryName(logMessage.LibraryId, projectFilePath);
            LibraryType = Path.IsRooted(logMessage.LibraryId) ? AssetFileTargetLibrary.Project : AssetFileTargetLibrary.Package
        }
    //remaining code
    }
```

With these changes, when we are creating the dummy/placeholder object, we can pass logmessage.LibraryType instead of using Unknown as the `LibraryType` property for `AssetFileTargetLibrary`. This allows us to eliminate the need for checking if the library type is unknown.

``` cs
        public static AssetsFileTargetLibrary CreatePlaceholder(string name, AssetFileTargetLibrary libraryType)
        {
            return new AssetsFileTargetLibrary(name, libraryType);
        }

        private AssetsFileTargetLibrary(string name, AssetFileTargetLibrary libraryType)
        {
            Name = name;
            Version = null;
            Type = libraryType;
            Dependencies = ImmutableArray<string>.Empty;
            FrameworkAssemblies = ImmutableArray<string>.Empty;
            CompileTimeAssemblies = ImmutableArray<string>.Empty;
            ContentFiles = ImmutableArray<AssetsFileTargetLibraryContentFile>.Empty;
            BuildFiles = ImmutableArray<string>.Empty;
            BuildMultiTargetingFiles = ImmutableArray<string>.Empty;
            DocumentationFiles = ImmutableArray<string>.Empty;
        }
```

drewnoakes: Oh I see. I guess it comes down to whether you think and absolute path is a good indicator that is a project or package. I don't know it to be, so prefer to capture that uncertainty in the data. It also means the unknown flag helps explain why so many other fields in the object have no value.

From a design point of view I prefer modelling that uncertainty explicitly, even if it means an extra check in two places.

A better fix overall would be for the lock file to include the library type in the log message JSON. That must be known by the writer of the file, but then that information is not serialised and is therefore lost to consumers.

kartheekp-ms:
> The most valuable testing for these code paths would be integration tests, which my team is investigating at the moment.

Thank you. Is there any tracking issue?

kartheekp-ms (approved):

jeffkl (approved):

drewnoakes:
> Is there any tracking issue?

Discussed offline.

PR #5150, opened 2023-04-26
 Dotnet NuGet Config CLI Config Set Command

dainizzz:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12476

## Description

Adds the `dotnet nuget config set` command, which sets the value of a NuGet configuration setting in the config section. By default, the user-level NuGet configuration file is used, and users can specify which configuration file they want to update using the option `--configfile`.
## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -54,5 +74,57 @@ public static ISettings GetSettingsFromDirectory(string directory)
...
+                && !configKey.Equals(ConfigurationConstants.RepositoryPath, StringComparison.OrdinalIgnoreCase)
+                && !configKey.Equals(ConfigurationConstants.DefaultPushSource, StringComparison.OrdinalIgnoreCase)
+                && !configKey.Equals(ConfigurationConstants.HostKey, StringComparison.OrdinalIgnoreCase)
+                && !configKey.Equals(ConfigurationConstants.UserKey, StringComparison.OrdinalIgnoreCase)
+                && !configKey.Equals(ConfigurationConstants.PasswordKey, StringComparison.OrdinalIgnoreCase)
+                && !configKey.Equals(ConfigurationConstants.NoProxy, StringComparison.OrdinalIgnoreCase)
+                && !configKey.Equals(ConfigurationConstants.MaxHttpRequestsPerSource, StringComparison.OrdinalIgnoreCase)
+                && !configKey.Equals(ConfigurationConstants.SignatureValidationMode, StringComparison.OrdinalIgnoreCase))
+            {
+                throw new CommandException(string.Format(CultureInfo.CurrentCulture, Strings.Error_ConfigSetInvalidKey, configKey));
```
martinrrm: Maybe this if statements its too large. I was looking at enums and maybe you can use TryParse from https://learn.microsoft.com/en-us/dotnet/api/system.enum.tryparse?view=net-7.0.

You can pass a boolean as parameter to do the operation  case-insensitive

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -40,10 +32,43 @@ public static void Run(ConfigPathsArgs args, Func<ILogger> getLogger)
...
+            RunnerHelper.ValidateConfigKey(args.ConfigKey);
+            ISettings settings = null;
+            if (string.IsNullOrEmpty(args.ConfigFile))
+            {
+                settings = RunnerHelper.GetSettingsFromDirectory(null);
+            }
+            else
+            {
+                settings = RunnerHelper.GetSettingsFromFile(args.ConfigFile);
+            }
```
jeffkl: I prefer this syntax:
```suggestion
            ISettings settings = string.IsNullOrEmpty(args.ConfigFile)
                ? RunnerHelper.GetSettingsFromDirectory(null)
                : RunnerHelper.GetSettingsFromFile(args.ConfigFile);
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -40,10 +32,43 @@ public static void Run(ConfigPathsArgs args, Func<ILogger> getLogger)
...
+            var encrypt = false;
+            if (args.ConfigKey.Equals(ConfigurationConstants.PasswordKey, StringComparison.OrdinalIgnoreCase))
+            {
+                encrypt = true;
+            }
```
jeffkl:
```suggestion
            bool encrypt = args.ConfigKey.Equals(ConfigurationConstants.PasswordKey, StringComparison.OrdinalIgnoreCase);
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -40,10 +32,43 @@ public static void Run(ConfigPathsArgs args, Func<ILogger> getLogger)
...
+        /// Creates a settings object using the directory argument,
+        /// or using the current directory if no argument is passed.
+        /// </summary>
+        /// <exception cref="CommandException"></exception>
         public static ISettings GetSettingsFromDirectory(string directory)
         {
+            if (string.IsNullOrEmpty(directory))
+            {
+                directory = Directory.GetCurrentDirectory();
+            }
```
jeffkl: nit: Add a line break
```suggestion
            }
            
```

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -20,35 +22,33 @@ public void ConfigPathsCommand_ListConfigPathsWithArgs_Success()
...
-                // Assert
-                DotnetCliUtil.VerifyResultSuccess(result, Path.Combine(testInfo.WorkingPath.Path, "NuGet.Config"));
-            }
+        
+            var result = CommandRunner.Run(
+                DotnetCli,
+                Directory.GetCurrentDirectory(),
+                $"{XplatDll} config paths {testInfo.WorkingPath}",
+                waitForExit: true
+                );
```
jeffkl: Can you please fix this whole source file to not have the trailing parenthesis and semi-colon on a new line?  I see a lot of instances of it.
```suggestion
                waitForExit: true);
```

jeffkl (approved):

jeffkl (approved):

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -192,5 +220,26 @@ public static void EnsureArgumentsNotNull<TArgs>(TArgs args, Func<ILogger> logge
...
             _ = logger ?? throw new ArgumentNullException(nameof(logger));
         }
+
+        /// <summary>
+        /// Throws an exception if the value passed in is not a valid config key.
+        /// </summary>
+        /// <exception cref="CommandException"></exception>
+        public static void ValidateConfigKey(string configKey)
+        {
+            if (!configKey.Equals(ConfigurationConstants.DependencyVersion, StringComparison.OrdinalIgnoreCase)
```
jebriede: Check configKey is not null before dereferencing it.

jebriede: Alternatively, you can write this as:

`
!ConfigurationConstants.DependencyVersion.Equals(configKey, StringComparison.OrdinalIgnoreCase)
`
etc.

Since the constants cannot be null.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -217,6 +219,119 @@ public void ConfigGetConfigKeyCommand_HelpMessage_Success()
...
+                $"{XplatDll} config set {key} {value} --configfile {filePath}",
+                waitForExit: true);
+
+            var settings = Configuration.Settings.LoadDefaultSettings(
+                testInfo.WorkingPath,
+                configFileName: filePath,
+                machineWideSettings: new XPlatMachineWideSetting());
+
+            var configSection = settings.GetSection("config");
+            var values = configSection?.Items.Select(c => c as AddItem).Where(c => c != null).ToList();
```
jebriede: Tests should be deterministic and always produce the same result so configSection should always have a result or should always be null, which means we should not allow configSection to be null in some cases and handle it quietly in the test.

If configSection can be null in some cases, and we don't have a test that covers that, or we want to add a test for how this config set command will handle that scenario, we should add a separate test for that.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -217,6 +219,119 @@ public void ConfigGetConfigKeyCommand_HelpMessage_Success()
...
+            var configItems = values.Where(i => i.Key == key);
```
jebriede, line 244: My comment above suggests that we should be explicit if configSection is expected to be null or not in this test. Note that in the code's current state, this line could result in a NullReferenceException because `values` would be null if configSection is null.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -217,6 +219,119 @@ public void ConfigGetConfigKeyCommand_HelpMessage_Success()
...
+                $"{XplatDll} config set {key} {value} --configfile {filePath}",
+                waitForExit: true);
+
+            var settings = Configuration.Settings.LoadDefaultSettings(
+                testInfo.WorkingPath,
+                configFileName: filePath,
+                machineWideSettings: new XPlatMachineWideSetting());
+
+            var configSection = settings.GetSection("config");
+            var values = configSection?.Items.Select(c => c as AddItem).Where(c => c != null).ToList();
```
jebriede: Same feedback about null configSection applies here as in the test above.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -217,6 +219,119 @@ public void ConfigGetConfigKeyCommand_HelpMessage_Success()
...
```
jebriede: Same feedback about null configSection applies here as in the test above.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -349,6 +462,48 @@ public void ConfigGetCommand_NullGetLogger_Fail()
...
+                $"{XplatDll} config set {key} {value} --configfile {filePath}",
+                waitForExit: true
+                );
+            var expectedError = string.Format(CultureInfo.CurrentCulture, Strings.Error_ConfigSetInvalidKey, key);
+            // Assert
+            DotnetCliUtil.VerifyResultFailure(result, expectedError);
+        }
+
+        [Fact]
+        public void ConfigSetCommand_NullArgs_Fail()
```
jebriede: Can you rename the test methods to follow the pattern described in the coding guidelines:
https://github.com/NuGet/NuGet.Client/blob/dev/docs/coding-guidelines.md#unit-test-method-naming

Specifically:
`Format: <what is being tested>_<under what conditions>_<with what expectations>`

With that pattern, the suffixes of _Fail and _Success should be more descriptive of what the expectation is of how they succeed or fail.

jebriede (changes requested):
Great work! Left a few comments of requested changes. Let me know if you have any questions!

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -217,6 +219,119 @@ public void ConfigGetConfigKeyCommand_HelpMessage_Success()
...
+
+            var settings = Configuration.Settings.LoadDefaultSettings(
+                testInfo.WorkingPath,
+                configFileName: filePath,
+                machineWideSettings: new XPlatMachineWideSetting());
+
+            var configSection = settings.GetSection("config");
+            var values = configSection?.Items.Select(c => c as AddItem).Where(c => c != null).ToList();
+            var configItems = values.Where(i => i.Key == key);
+            var configFilePath = configItems.FirstOrDefault().ConfigPath;
```
heng-liu: You may use [Enumerable.Single()](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.single?view=net-7.0), like [this line](https://github.com/NuGet/NuGet.Client/blob/dev/test/NuGet.Core.Tests/NuGet.Commands.Test/RuntimeTargetsTests.cs#L84)
So you don't have to assert the count later.

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -828,6 +828,22 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
...
+    <value>The NuGet configuration file (nuget.config) whose value will be set.</value>
+  </data>
+  <data name="ConfigSetConfigKeyDescription" xml:space="preserve">
+    <value>The key of the settings that are to be set.</value>
+  </data>
+  <data name="ConfigSetConfigValueDescription" xml:space="preserve">
+    <value>The value of the settings that are to be set.</value>
+  </data>
+  <data name="Error_ConfigSetInvalidKey" xml:space="preserve">
+    <value>'{0}' is not a valid config key.</value>
```
heng-liu: How about `<value>'{0}' is not a valid key in config section.</value>`?
Users might misunderstand that they could set keys in other sections in settings.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -100,6 +115,19 @@ public static ISettings GetSettingsFromDirectory(string directory)
...
 
         /// <summary>
+        /// Creates a settings object utilizing a NuGet configuration file path.
+        /// </summary>
+        /// <param name="filePath"></param>
+        /// <returns></returns>
+        public static ISettings GetSettingsFromFile(string filePath)
+        {
+            var directory = Path.GetDirectoryName(filePath);
+            return NuGet.Configuration.Settings.LoadDefaultSettings(
```
heng-liu: This will only load the settings from the specified config file, right?
When user specifies `--configfile`  in `set` command, what would happen if the config file path doesn't exist? (e.g. user enters a wrong path)
Will we show error message saying the config file doesn't exist? Thanks!

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
zivkan: Another alternative is to use `#nullable enable` in this file, and let the compiler do null checking for us. `#nullable enable` will also need to be set in any file that calls a public method in this class, or be very careful when calling the method in other files. But letting the compiler do null checking for us will reduce the amount of code we have to write checking nulls in different places.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
zivkan: Also, `.FirstOrDefault()` can return null, which means that `ConfigPath` might throw a NullReferenceException, which is why it's good to use `.First()` or `.Single()` when you know it's not going to be null (but when it really might be null, you need to handle it and give a better error message).

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
zivkan: I'm not sure that `LoadDefaultSettings` is correct when a specific file is requested. Isn't the first argument a directory name, not a file name?

Also, the XMLDoc for the method says it loads config files in parent directories, and probably also loads the user-profile config: https://github.com/NuGet/NuGet.Client/blob/25d848af193c5dc17530525be8d0952d0335e8f1/src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs#L262-L295

I don't know which API you're using to save settings, and if it's the same API that `dotnet nuget add source` and `nuget.exe config set` commands use, but if it is, it has this unintuitive behaviour: https://github.com/NuGet/Home/blob/dev/accepted/2023/dotnet-nuget-setapikey.md#which-nugetconfig-file

If a customer uses `dotnet nuget config set --config-file path\to\nuget.config key value`, then I think most customers would expect the value to be set in the specified file. Hopefully your tests already validate that it works like that, I haven't checked the tests. But a method that sounds like it's reading settings from a speicifc file, but then using NuGet's API for multi-level config loading, feels like it might not be right.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -192,5 +220,26 @@ public static void EnsureArgumentsNotNull<TArgs>(TArgs args, Func<ILogger> logge
...
+            if (!configKey.Equals(ConfigurationConstants.DependencyVersion, StringComparison.OrdinalIgnoreCase)
+                && !configKey.Equals(ConfigurationConstants.GlobalPackagesFolder, StringComparison.OrdinalIgnoreCase)
+                && !configKey.Equals(ConfigurationConstants.RepositoryPath, StringComparison.OrdinalIgnoreCase)
+                && !configKey.Equals(ConfigurationConstants.DefaultPushSource, StringComparison.OrdinalIgnoreCase)
+                && !configKey.Equals(ConfigurationConstants.HostKey, StringComparison.OrdinalIgnoreCase)
+                && !configKey.Equals(ConfigurationConstants.UserKey, StringComparison.OrdinalIgnoreCase)
+                && !configKey.Equals(ConfigurationConstants.PasswordKey, StringComparison.OrdinalIgnoreCase)
+                && !configKey.Equals(ConfigurationConstants.NoProxy, StringComparison.OrdinalIgnoreCase)
+                && !configKey.Equals(ConfigurationConstants.MaxHttpRequestsPerSource, StringComparison.OrdinalIgnoreCase)
+                && !configKey.Equals(ConfigurationConstants.SignatureValidationMode, StringComparison.OrdinalIgnoreCase))
```
zivkan: On one hand, this helps catch typos.

On the other hand, this feels high risk of causing bugs when in the future we add more config settings. Anyone adding a new config for some feature is unlikely to know/remember that this code here enumerates the full list of allowed settings.

A new public API in `ConfigurationConstants` which contains an `IReadOnlyList<string>` of all the config options, has a better chance of being maintained correctly when new config keys are added.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
dainizzz: Yes, you get this message: `error: File 'filepath' does not exist.`

dainizzz: That makes sense! I misunderstood how `LoadDefaultSettings` works. I'm editing this to use `LoadSpecificSettings` instead, which appears to do what I actually wanted:
https://github.com/NuGet/NuGet.Client/blob/25d848af193c5dc17530525be8d0952d0335e8f1/src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs#LL346C1-L363C10

I'll take a closer look at the tests I have, because they passed when I was using `LoadDefaultSettings`, and I'm now not sure that they should have passed.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
dainizzz: `configSection` should not be null in any of these tests, since I'm only accessing it after adding a value to it. Is it sufficient to remove the `?` suffix, or do I need add an assertion that `configSection` is not null as well?

jebriede: Since you're setting up the mock settings to have the section "config" and when calling settings.GetSection("config") we know it will return a section that is not null, you should be able to simply remove the `?`.

martinrrm:
@NuGet/nuget-client Hi team! Can we get some reviews again on this PR? Thanks!

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -218,6 +219,117 @@ public void ConfigGetConfigKeyCommand_HelpMessage_Success()
...
+            // Arrange & Act
+            using var testInfo = new TestInfo("NuGet.Config");
+            var key = "http_proxy";
+            var value = "http://company-octopus:8765@contoso.test";
+            var filePath = Path.Combine(testInfo.WorkingPath, "NuGet.Config");
+
+            var result = CommandRunner.Run(
+                DotnetCli,
+                Directory.GetCurrentDirectory(),
+                $"{XplatDll} config set {key} {value} --configfile {filePath}",
```
donnie-msft, line 272: I'm not following how this test is different than `ConfigSetCommand_WithConfigFileArg_AddsSetting`.
Was the intention to have an existing `key` entry, then update that entry with `value`? I'm not sure this test is doing that, currently. If it is, please help me understand how it works.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -218,6 +219,117 @@ public void ConfigGetConfigKeyCommand_HelpMessage_Success()
...
+            var value = "accept";
+            var filePath = Path.Combine(testInfo.WorkingPath, "NuGet.Config");
+
+            var result = CommandRunner.Run(
+                DotnetCli,
+                Directory.GetCurrentDirectory(),
+                $"{XplatDll} config set {key} {value} --configfile {filePath}",
+                waitForExit: true);
+
+            var settings = Configuration.Settings.LoadDefaultSettings(
```
donnie-msft: I believe the goal is to not have a `<config>` section, then run the command, then ensure there is a `<config>` section?
How do I know the section didn't exist in the Arrange? From what I see, the command runs, creates the section, then the test loads settings (from DefaultSettings) and ensures the section is there.

Consider this pattern:
1. LoadDefaultSettings 
2.  Read and store the existence of `<config>` section (variable x)
3. Run the command
4.  Read and store the existence of `<config>` section (variable y)
5. Assert x was null/empty
6. Assert y exists

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
dainizzz: Yes, the `TestInfo` used by this test has an existing `http_proxy` key, and the value is then updated during the test. Would it be better to get the current value of `http_proxy`, change it and get the new value, and then compare them?

donnie-msft: > Would it be better to get the current value of http_proxy, change it and get the new value, and then compare them?

I like that. Seems more clear. Thank you

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -407,6 +519,46 @@ public void ConfigGetCommand_NullGetLogger_Fail()
...
+
+        [Fact]
+        public void ConfigSetCommand_WithNullGetLogger_ThrowsArgumentNullException()
+        {
+            // Arrange
+            var args = new ConfigSetArgs();
+
+            // Act & Assert
+            Assert.Throws<ArgumentNullException>(() => ConfigSetRunner.Run(args, getLogger: null));
+        }
```
jebriede: Are these two test names mixed up? It looks like the first test tests for a null logger and the second test tests for null args but the names of the tests suggest otherwise.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -407,6 +519,46 @@ public void ConfigGetCommand_NullGetLogger_Fail()
...
+                $"{XplatDll} config set {key} {value} --configfile {filePath}",
+                waitForExit: true);
+            var expectedError = string.Format(CultureInfo.CurrentCulture, Strings.Error_ConfigSetInvalidKey, key);
+
+            // Assert
+            DotnetCliUtil.VerifyResultFailure(result, expectedError);
+        }
+
+        [Fact]
+        public void ConfigSetCommand_WithNullArgs_ThrowsArgumentNullException()
```
jebriede: This test and the next test could be combined into a single Theory test with two different inputs: In one input, the args would be null and in the other, the logger would be null. The expected outcome is the same, that a null ref exception would be thrown.

jebriede: By its nature, this is a unit test since it's just testing the unit (the Run method) in isolation and does not require any other resources that are needed in a functional test. This test should be moved into the Unit Test assembly.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
dainizzz: In the new Theory test, I tried to make it more evident which arguments are null

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigCommand.cs
```diff
@@ -55,11 +55,36 @@ internal class ConfigCommand
...
+                {
+                    CommandArgument configKey = SetCmd.Argument(
+                        "CONFIG_KEY",
+                        Strings.ConfigSetConfigKeyDescription);
+                    CommandArgument configValue = SetCmd.Argument(
+                        "CONFIG_VALUE",
+                        Strings.ConfigSetConfigValueDescription);
+                    var configFile = SetCmd.Option(
+                        "--configfile",
+                        Strings.ConfigSetConfigFileDescription,
```
kartheekp-ms: how about using the existing string `Strings.Option_ConfigFile` like other commands? https://github.com/search?q=repo%3ANuGet%2FNuGet.Client+Option_ConfigFile&type=code

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -74,6 +74,21 @@ public static void Run(ConfigGetArgs args, Func<ILogger> getLogger)
...
 
+    internal static class ConfigSetRunner
+    {
+        public static void Run(ConfigSetArgs args, Func<ILogger> getLogger)
+        {
+            RunnerHelper.EnsureArgumentsNotNull(args, getLogger);
+            RunnerHelper.ValidateConfigKey(args.ConfigKey);
+            ISettings settings = string.IsNullOrEmpty(args.ConfigFile)
+                ? RunnerHelper.GetSettingsFromDirectory(null)
+                : Settings.LoadSpecificSettings(Path.GetDirectoryName(args.ConfigFile), args.ConfigFile);
```
kartheekp-ms: How about using existing utility method `XPlatUtility.ProcessConfigFikle()` to fetch the settings.
https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/XPlatUtility.cs#L72

For example, dotnet nuget verify command uses this utility method https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Signing/VerifyCommand.cs#LL65C50-L65C67

test/NuGet.Core.Tests/NuGet.CommandLine.Xplat.Tests/XPlatConfigUnitTests.cs
```diff
@@ -0,0 +1,104 @@
...
+using Xunit;
+
+namespace NuGet.CommandLine.Xplat.Tests
+{
+    public class XPlatConfigUnitTests
+    {
+        [Theory]
+        [InlineData(true, false)]
+        [InlineData(false, true)]
+        public void ConfigPathsCommand_WithNullArguments_ThrowsArgumentNullException(bool argsIsNull, bool loggerIsNull )
```
jebriede: nit: remove extra spaces

```suggestion
        public void ConfigPathsCommand_WithNullArguments_ThrowsArgumentNullException(bool argsIsNull, bool loggerIsNull)
```

test/NuGet.Core.Tests/NuGet.CommandLine.Xplat.Tests/XPlatConfigUnitTests.cs
```diff
@@ -0,0 +1,104 @@
...
+        {
+            // Arrange
+            ConfigPathsArgs args = null;
+            CommandOutputLogger getLogger = null;
+
+            if ( !argsIsNull )
```
jebriede: nit: remove extra spaces. If you use Ctrl + K + D in Visual Studio, it should format your code for you and remove these spaces. There are other occurrences of spacing around parentheses, but I'll just leave this comment for all of them. If you use the format feature, it should clean them all up.

```suggestion
            if (!argsIsNull)
```

jebriede (approved):
Approving with suggestion of formatting code to remove extra spaces.

martinrrm (approved):

donnie-msft (approved):

PR #5155, opened 2023-05-02
Add nullable annotations to NuGet.Frameworks

### LINKED ISSUE #12570: Add nullable annotations to NuGet.Frameworks
### NuGet Product(s) Affected

NuGet SDK

### Current Behavior

Currently NuGet.Frameworks doesn't have annotations to allow the compiler to work with the compiler's "nullable reference types" feature.

### Desired Behavior

Add null annotations, so the compiler knows when null might be returned, and when it's definitely not.

### Additional Context

Anyone using NuGet.Frameworks in their project might find that upgrading to the newer version of NuGet.Frameworks might result in compiler warnings or errors, depending on if your project opts into nullable checks.

Additionally, it may be that while annotating the project that null-related bugs are found and fixed. Any app that depends on the behaviour considered buggy might find their app behaves differently since NuGet.Frameworks might throw exceptions when it previously didn't, or have a different result for the same inputs.

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12570

Regression? No

## Description

* Add nullable annotations to NuGet.Frameworks and its test project
* Found two places where `#if NET40_CLIENT` was used. Delete it, since we stopped building for .NET 4.0 a while ago
* Some minor code cleanup in some places (for example, use auto-props)
* Added null checks and throw `ArgumentNullException` on some public APIs where nulls are not expected/allowed, but wasn't previously being enforced (would have thrown `NullReferenceException`, so not has a better error experience).
* I think only 1 bug was found and fixed. See comments below.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception: PR goal is to annotate nullability. Product behaviour, and therefore tests, shouldn't be affected.
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Frameworks/CompatibilityProvider.cs
+++ b/src/NuGet.Core/NuGet.Frameworks/CompatibilityProvider.cs
@@ -122,7 +122,7 @@ namespace NuGet.Frameworks
             if (target.IsPCL)
             {
                 // do not include optional frameworks here since we might be unable to tell what is optional on the other framework
-                if (!_mappings.TryGetPortableFrameworks(target.Profile, false, out targetFrameworks))
+                if (!_mappings.TryGetPortableFrameworks(target.Profile, includeOptional: false, out targetFrameworks))
                 {
                     targetFrameworks = Array.Empty<NuGetFramework>();
                 }
@@ -135,7 +135,7 @@ namespace NuGet.Frameworks
             if (candidate.IsPCL)
             {
                 // include optional frameworks here, the larger the list the more compatible it is
-                if (!_mappings.TryGetPortableFrameworks(candidate.Profile, true, out candidateFrameworks))
+                if (!_mappings.TryGetPortableFrameworks(candidate.Profile, includeOptional: true, out candidateFrameworks))
                 {
                     candidateFrameworks = Array.Empty<NuGetFramework>();
                 }
--- a/src/NuGet.Core/NuGet.Frameworks/CompatibilityProvider.cs
+++ b/src/NuGet.Core/NuGet.Frameworks/CompatibilityProvider.cs
@@ -17,7 +17,7 @@ namespace NuGet.Frameworks
 
         public CompatibilityProvider(IFrameworkNameProvider mappings)
         {
-            _mappings = mappings;
+            _mappings = mappings ?? throw new ArgumentNullException(nameof(mappings));
             _expander = new FrameworkExpander(mappings);
             _cache = new ConcurrentDictionary<CompatibilityCacheKey, bool>();
         }
@@ -30,15 +30,8 @@ namespace NuGet.Frameworks
         /// <returns>True if framework supports other</returns>
         public bool IsCompatible(NuGetFramework target, NuGetFramework candidate)
         {
-            if (target == null)
-            {
-                throw new ArgumentNullException(nameof(target));
-            }
-
-            if (candidate == null)
-            {
-                throw new ArgumentNullException(nameof(candidate));
-            }
+            if (target == null) throw new ArgumentNullException(nameof(target));
+            if (candidate == null) throw new ArgumentNullException(nameof(candidate));
 
             // check the cache for a solution
             var cacheKey = new CompatibilityCacheKey(target, candidate);
@@ -123,13 +116,16 @@ namespace NuGet.Frameworks
                 return IsCompatibleWithTarget(target, candidate);
             }
 
-            IEnumerable<NuGetFramework> targetFrameworks;
-            IEnumerable<NuGetFramework> candidateFrameworks;
+            IEnumerable<NuGetFramework>? targetFrameworks;
+            IEnumerable<NuGetFramework>? candidateFrameworks;
 
             if (target.IsPCL)
             {
                 // do not include optional frameworks here since we might be unable to tell what is optional on the other framework
-                _mappings.TryGetPortableFrameworks(target.Profile, false, out targetFrameworks);
+                if (!_mappings.TryGetPortableFrameworks(target.Profile, false, out targetFrameworks))
+                {
+                    targetFrameworks = Array.Empty<NuGetFramework>();
+                }
             }
             else
             {
@@ -139,7 +135,10 @@ namespace NuGet.Frameworks
             if (candidate.IsPCL)
             {
                 // include optional frameworks here, the larger the list the more compatible it is
-                _mappings.TryGetPortableFrameworks(candidate.Profile, true, out candidateFrameworks);
+                if (!_mappings.TryGetPortableFrameworks(candidate.Profile, true, out candidateFrameworks))
+                {
+                    candidateFrameworks = Array.Empty<NuGetFramework>();
+                }
             }
             else
             {
@@ -241,9 +240,7 @@ namespace NuGet.Frameworks
             {
                 var frameworkToExpand = toExpand.Pop();
 
-                IEnumerable<NuGetFramework> compatibleFrameworks = null;
-
-                if (_mappings.TryGetEquivalentFrameworks(frameworkToExpand, out compatibleFrameworks))
+                if (_mappings.TryGetEquivalentFrameworks(frameworkToExpand, out IEnumerable<NuGetFramework>? compatibleFrameworks))
                 {
                     foreach (var curFramework in compatibleFrameworks)
                     {
--- a/src/NuGet.Core/NuGet.Frameworks/FrameworkNameHelpers.cs
+++ b/src/NuGet.Core/NuGet.Frameworks/FrameworkNameHelpers.cs
@@ -14,14 +14,14 @@ namespace NuGet.Frameworks
             return String.Format(CultureInfo.InvariantCulture, "Profile{0}", profileNumber);
         }
 
-        public static string GetFolderName(string identifierShortName, string versionString, string profileShortName)
+        public static string GetFolderName(string identifierShortName, string versionString, string? profileShortName)
         {
             return String.Format(CultureInfo.InvariantCulture, "{0}{1}{2}{3}", identifierShortName, versionString, String.IsNullOrEmpty(profileShortName) ? string.Empty : "-", profileShortName);
         }
 
         public static string GetVersionString(Version version)
         {
-            string versionString = null;
+            string? versionString = null;
 
             if (version != null)
             {
@@ -38,12 +38,12 @@ namespace NuGet.Frameworks
                 }
             }
 
-            return versionString;
+            return versionString!;
         }
 
-        public static Version GetVersion(string versionString)
+        public static Version GetVersion(string? versionString)
         {
-            Version version = null;
+            Version version;
 
             if (string.IsNullOrEmpty(versionString))
             {
@@ -51,7 +51,7 @@ namespace NuGet.Frameworks
             }
             else
             {
-                if (versionString.IndexOf('.') > -1)
+                if (versionString!.IndexOf('.') > -1)
                 {
                     // parse the version as a normal dot delimited version
                     _ = Version.TryParse(versionString, out version);
--- a/src/NuGet.Core/NuGet.Frameworks/FrameworkNameHelpers.cs
+++ b/src/NuGet.Core/NuGet.Frameworks/FrameworkNameHelpers.cs
@@ -54,7 +54,7 @@ namespace NuGet.Frameworks
                 if (versionString!.IndexOf('.') > -1)
                 {
                     // parse the version as a normal dot delimited version
-                    _ = Version.TryParse(versionString, out version);
+                    version = Version.Parse(versionString);
                 }
                 else
                 {
@@ -67,7 +67,7 @@ namespace NuGet.Frameworks
                     // take only the first 4 digits and add dots
                     // 451 -> 4.5.1
                     // 81233 -> 8123
-                    _ = Version.TryParse(string.Join(".", versionString.ToCharArray().Take(4)), out version);
+                    version = Version.Parse(string.Join(".", versionString.ToCharArray().Take(4)));
                 }
             }
 
--- a/src/NuGet.Core/NuGet.Frameworks/FrameworkNameHelpers.cs
+++ b/src/NuGet.Core/NuGet.Frameworks/FrameworkNameHelpers.cs
@@ -43,18 +43,16 @@ namespace NuGet.Frameworks
 
         public static Version GetVersion(string? versionString)
         {
-            Version version;
-
             if (string.IsNullOrEmpty(versionString))
             {
-                version = FrameworkConstants.EmptyVersion;
+                return FrameworkConstants.EmptyVersion;
             }
             else
             {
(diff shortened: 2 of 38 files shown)


src/NuGet.Core/NuGet.Frameworks/FrameworkNameProvider.cs
```diff
@@ -993,8 +968,12 @@ public bool TryGetSubSetFrameworks(string frameworkIdentifier, out IEnumerable<s
             return false;
         }
 
-        public int CompareFrameworks(NuGetFramework x, NuGetFramework y)
+        public int CompareFrameworks(NuGetFramework? x, NuGetFramework? y)
         {
+            if (x is null && y is null) return 0;
+            if (x is null) return -1;
+            if (y is null) return 1;
```
zivkan, line 975: This method, and `CompareUsingPrecedence` below, would have previously thrown `NullReferenceExceptions` if either of the inputs were null. However, in versions of the BCL where `IComparable` is annotated, `Compare` methods allow nulls. So, in the spirit of "least surprise", I thought we should also support nulls. But, it is a behaviour change.

src/NuGet.Core/NuGet.Frameworks/FrameworkReducer.cs
```diff
@@ -389,18 +396,21 @@ private IEnumerable<NuGetFramework> GetNearestPCLtoPCL(NuGetFramework framework,
                 // from all possible frameworks, find the best match
                 var nearestForSub = GetNearest(sub, allPclFrameworks);
 
-                // +1 each framework containing the best match
-                foreach (var pair in pclToFrameworks)
+                if (nearestForSub != null)
```
zivkan, line 399: I don't think it's possible for `nearestForSub` to be null here, but that's because in the call to `GetNearest` the arguments will always have a result, whereas with different inputs `GetNearest()` could return null, hence it's annotated in that way. Anyway, in order to keep the compiler happy, something had to be done, but I think that `paris.Value.Contains` would have thrown an `ArgumentException`.

It only occurs to me now, that I could have used `!` to suppress the analyzer and keep old behaviour. Does anyone think I should do that instead?  Or throw an exception with a more clear error message?  If noone replies to this comment, I'll assume people are fine with this change.

src/NuGet.Core/NuGet.Frameworks/FrameworkNameHelpers.cs
```diff
@@ -38,23 +38,23 @@ public static string GetVersionString(Version version)
...
                 version = FrameworkConstants.EmptyVersion;
             }
             else
             {
-                if (versionString.IndexOf('.') > -1)
+                if (versionString!.IndexOf('.') > -1)
                 {
                     // parse the version as a normal dot delimited version
-                    _ = Version.TryParse(versionString, out version);
+                    version = Version.Parse(versionString);
```
zivkan: Previously, `TryParse` would end up setting `version` to a default value (null), but it doesn't make sense to me that our method named `GetVersion` should return null. Especially when `versionString` is allowed to be null, in which time we return a default, non-null value. There, this is a change in behaviour since now it can throw.

Same thing for the `else` branch's `TryParse` to `Parse` change, 13 lines below.

src/NuGet.Core/NuGet.Frameworks/CompatibilityProvider.cs
```diff
@@ -123,13 +116,16 @@ private bool IsPCLCompatible(NuGetFramework target, NuGetFramework candidate)
...
-            IEnumerable<NuGetFramework> targetFrameworks;
-            IEnumerable<NuGetFramework> candidateFrameworks;
+            IEnumerable<NuGetFramework>? targetFrameworks;
+            IEnumerable<NuGetFramework>? candidateFrameworks;
 
             if (target.IsPCL)
             {
                 // do not include optional frameworks here since we might be unable to tell what is optional on the other framework
-                _mappings.TryGetPortableFrameworks(target.Profile, false, out targetFrameworks);
+                if (!_mappings.TryGetPortableFrameworks(target.Profile, false, out targetFrameworks))
```
jeffkl: Since you're here, do you want to add parameter names to constants?
```suggestion
                if (!_mappings.TryGetPortableFrameworks(target.Profile, includeOptional: false, out targetFrameworks))
```

src/NuGet.Core/NuGet.Frameworks/CompatibilityProvider.cs
```diff
@@ -139,7 +135,10 @@ private bool IsPCLCompatible(NuGetFramework target, NuGetFramework candidate)
             if (candidate.IsPCL)
             {
                 // include optional frameworks here, the larger the list the more compatible it is
-                _mappings.TryGetPortableFrameworks(candidate.Profile, true, out candidateFrameworks);
+                if (!_mappings.TryGetPortableFrameworks(candidate.Profile, true, out candidateFrameworks))
```
jeffkl:
```suggestion
                if (!_mappings.TryGetPortableFrameworks(candidate.Profile, includeOptional: true, out candidateFrameworks))
```

src/NuGet.Core/NuGet.Frameworks/FrameworkNameHelpers.cs
```diff
@@ -38,23 +38,23 @@ public static string GetVersionString(Version version)
...
 
-            return versionString;
+            return versionString!;
         }
 
-        public static Version GetVersion(string versionString)
+        public static Version GetVersion(string? versionString)
         {
-            Version version = null;
+            Version version;
```
jeffkl: You could also simplify this method to just return the values and reduce nesting while you're at it.  Line 50, 57, and 70 could just `return something`.

src/NuGet.Core/NuGet.Frameworks/FrameworkNameHelpers.cs
jeffkl: This does seem a little dangerous to start throwing now but the method is not a `Try` method...  My only worry is that if this one breaking change causes problems, we'd need to revert this whole thing.  But I'm not sure if its worth keeping the existing behavior, then sending another PR to adjust the behavior.  It all comes down to if anyone is broken by it and how many people are affected...

src/NuGet.Core/NuGet.Frameworks/FrameworkReducer.cs
jeffkl: This is probably better, I don't think the rest of the code would have thrown if `nearestForSub` is `null` since it only passes it to `Contains` which just returns `false`.  At least now the lookup is skipped.

src/NuGet.Core/NuGet.Frameworks/NuGetFramework.cs
```diff
@@ -41,12 +41,12 @@ public NuGetFramework(string framework, Version version)
         /// <summary>
         /// Creates a new NuGetFramework instance, with an optional profile (only available for netframework)
         /// </summary>
-        public NuGetFramework(string frameworkIdentifier, Version frameworkVersion, string frameworkProfile)
+        public NuGetFramework(string frameworkIdentifier, Version frameworkVersion, string? frameworkProfile)
             : this(frameworkIdentifier, frameworkVersion, profile: ProcessProfile(frameworkProfile), platform: string.Empty, platformVersion: FrameworkConstants.EmptyVersion)
```
jeffkl: I would vote to get rid of `ProcessProfile()`
```suggestion
            : this(frameworkIdentifier, frameworkVersion, profile: frameworkProfile ?? string.Empty, platform: string.Empty, platformVersion: FrameworkConstants.EmptyVersion)
```

src/NuGet.Core/NuGet.Frameworks/NuGetFramework.cs
```diff
@@ -270,9 +238,8 @@ public virtual string GetShortFolderName(IFrameworkNameProvider mappings)
                 {
                     sb.Append("-");
 
-                    IEnumerable<NuGetFramework> frameworks = null;
                     if (framework.HasProfile
-                        && mappings.TryGetPortableFrameworks(framework.Profile, false, out frameworks)
+                        && mappings.TryGetPortableFrameworks(framework.Profile, false, out IEnumerable<NuGetFramework>? frameworks)
```
jeffkl:
```suggestion
                        && mappings.TryGetPortableFrameworks(framework.Profile, includeOptional: false, out IEnumerable<NuGetFramework>? frameworks)
```

src/NuGet.Core/NuGet.Frameworks/NuGetFrameworkFactory.cs
```diff
@@ -168,26 +146,15 @@ private static string[] GetParts(string targetPlatformMoniker)
...
-            {
-                throw new ArgumentNullException(nameof(mappings));
-            }
-
-            var parts = frameworkName.Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries).Select(s => s.Trim()).ToArray();
+            if (frameworkName == null) throw new ArgumentNullException(nameof(frameworkName));
+            if (mappings == null) throw new ArgumentNullException(nameof(mappings));
 
-            NuGetFramework result = null;
+            string[] parts = frameworkName.Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries).Select(s => s.Trim()).ToArray();
```
jeffkl: Do we have a `static readonly char[]` with a single comma somewhere we could re-use?  I'd imagine this is a hot path and seeing us allocate a char array for every call makes me sad.

src/NuGet.Core/NuGet.Frameworks/NuGetFramework.cs
```diff
@@ -457,17 +424,23 @@ public override string ToString()
                 : DotNetFrameworkName;
         }
 
-        public bool Equals(NuGetFramework other)
+        public bool Equals(NuGetFramework? other)
         {
+#pragma warning disable CS8604 // Possible null reference argument.
```
jeffkl, line 424: Are these suppressions needed since you updated `NuGetFrameworkFullComparer`?

src/NuGet.Core/NuGet.Frameworks/FrameworkReducer.cs
zivkan: Oops, you're right. `pairs.Value` is an `IEnumerable`, not a `Dictionary`. `Dictionary.ContainsKey` throws on null, but that's not happening here.

src/NuGet.Core/NuGet.Frameworks/NuGetFrameworkFactory.cs
zivkan: There are 20 examples in our code of `new char[] { ',' }` 😱.  I'll fix this one here, but I don't want this PR to touch files outside of NuGet.Frameworks.

src/NuGet.Core/NuGet.Frameworks/NuGetFramework.cs
zivkan: We'd have to change `Comparer` to be of type `NuGetFrameworkFullComparer` instead of `IEqualityComparer<NuGetFramework>`. However, that's a public API, so it would be a breaking change.

As the comment on the next line says/implies, .NET Framework and .NET Standard (and .NETCoreApp before version 5) don't have nullable annotations, and the compiler ("incorrectly") thinks this means nulls are not allowed. But the compiler just assumes that methods without nullable annotations don't allow nulls, which isn't correct in this situation.

src/NuGet.Core/NuGet.Frameworks/FrameworkNameHelpers.cs
zivkan: > My only worry is that if this one breaking change causes problems, we'd need to revert this whole thing.

When we fix a 3 year old bug, we don't revert the commit that introduced the bug, we just fix the bug. Commits don't even need to be that old. Specific example, when fixing a CPM bug, we don't revert CPM. So, I don't see why we'd have to revert this PR, we could just change this one method. no?

src/NuGet.Core/NuGet.Frameworks/NuGetFramework.cs
jeffkl: Ah, makes sense!

src/NuGet.Core/NuGet.Frameworks/FrameworkNameHelpers.cs
```diff
@@ -38,23 +38,21 @@ public static string GetVersionString(Version version)
                 }
             }
 
-            return versionString;
+            return versionString!;
         }
 
-        public static Version GetVersion(string versionString)
+        public static Version GetVersion(string? versionString)
```
jeffkl, line 44: nit: The nesting of the `else`s can be removed now as well which makes this method faster.

src/NuGet.Core/NuGet.Frameworks/FrameworkNameHelpers.cs
jeffkl: > Specific example, when fixing a CPM bug, we don't revert CPM. So, I don't see why we'd have to revert this PR, we could just change this one method. no?

I just recently tried to fix a bug in static graph-based restore and ended up having to revert the fix, bringing back the older, less impactful bug.  I also had to revert some refactorings that I had to add back later and the whole thing was a mess. I'm wondering if there's a bug late in a ship cycle and its too risky to fix forward, we'd be forced to revert.  We're early in the ship cycle so I suppose that reduces risk.  This pull request is just adding nullability and reverting it would not be the same as completely backing out functionality like CPM.

My point is that if the overwhelming majority of this change is non-risky and non-breaky, we can consider a different pull request to follow-up on this one that makes a breaking change in the event that later on we'd need to make a revert due to circumstances out of our control.  If it doesn't break anyone then the whole thing is moot, if it breaks and we can fix it, then its also moot.  If it breaks and we have to revert, I'd hate to lose the rest of this change.  I'll leave it up to you in this case, just trying to help everyone learn from my recent mistake.

jeffkl (approved):

src/NuGet.Core/NuGet.Frameworks/FrameworkNameHelpers.cs
zivkan: I forgot to mention that Intellisense says this method has zero references within our own repo. So, this method is only part of the NuGet Client SDK, so we're unlikely to get any feedback until after we ship the new version and someone using the API upgrades to our new package version.  But considering this method returns `System.Version`, not a Version type from `NuGet.Versioning`, it's a really pointless public API to have. Customers should just use `System.Version.Parse()` directly. This really is an excellent candidate for a public API we should delete at some point.

But for this reason, I'm really not concerned about the breaking change. Such low risk with such an easy mitigation.

jeffkl: :shipit:

PR #5157, opened 2023-05-03
Use PreAuthenticate=true in NuGet.Protocol

### LINKED ISSUE #12514: NuGet should use HttpClientHandler.PreAuthentication to reduce HTTP 401's
### NuGet Product(s) Affected

NuGet.exe, Visual Studio Package Management UI, Visual Studio Package Manager Console, MSBuild.exe, dotnet.exe, NuGet SDK

### Current Behavior

Related to https://github.com/NuGet/Home/issues/11600, NuGet is currently making an unauthenticated HTTP request to every URL. For private feeds, the server then responds with HTTP 401, and then NuGet (technically HttpClientHandler does it automatically, so it's completely hidden from NuGet) resends with credentials.

This means that at least 50% of all HTTP requests NuGet sends to a server results in HTTP 401 responses.

Note that PreAuthenticate has a limitation/design choice where a "base" URL needs to be authenticated first, and then any URL that matches the rough equivilent of `url.StartsWith(baseUrl.SubString(0, baseUrl.IndexOfLast("/"))` will re-use those creds.

Therefore, enabling PreAuthenticate without any other changes means that it will only work when:
1. NuGet does not use a cached service index
2. The server makes all resources (listed in the service index) start with the same URL that the service index itself is in.

In particular, this means that Azure Artifacts' current implementation where customers write a URL with "friendly"/alias names in their nuget.config, but then the service index changes parts of the URL to canonical GUIDs, will not meet HttpClientHandler's PreAuthenticate design.

However, other server implementations already meet the second requirement above, so PreAuthenticate, without any other changes to NuGet, will immediately benefit those servers (when NuGet doesn't use a cached service index, which has a different tracking issue).

### Desired Behavior

Dramatically reduce the number of HTTP 401's the server has to send back to the client.

### Additional Context

_No response_

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12514

Regression? no

## Description

Set `HttpClientHandler.PreAuthenticate` to true. As long as a private NuGet feed puts all resources under sub-directories where the service index is (docs issue created to create server implementation guidelines), then as long as NuGet makes an HTTP request for the service index, then all other URLs will be pre-authenticated.

There are more edge cases (like expiring tokens), but see the [epic for reducing HTTP 401's](https://github.com/NuGet/Home/issues/12517) for other follow up issues. This PR will still provide significant benefit for many scenarios, even if there are some gaps.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [x] Documentation PR or issue filled: https://github.com/NuGet/docs.microsoft.com-nuget/issues/3059
  - **OR**
  - [ ] N/A

### DIFF
--- a/test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
+++ b/test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
@@ -5,8 +5,10 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Net;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Moq;
@@ -86,11 +88,11 @@ namespace NuGet.Protocol.FuncTest
                 string? value = request.Headers["Authorization"];
                 return string.IsNullOrEmpty(value);
             }
-        }
 
-        private static Task<string> ProcessResponse(HttpSourceResult result)
-        {
-            return Task.FromResult(string.Empty);
+            static Task<string> ProcessResponse(HttpSourceResult result)
+            {
+                return Task.FromResult(string.Empty);
+            }
         }
 
         private class RequestCollectingServer
@@ -141,8 +143,51 @@ namespace NuGet.Protocol.FuncTest
                 _httpListener = listener;
                 _baseUrl = baseUrl;
 
-                _serverThread = new Thread(ProcessRequests);
-                _serverThread.Start();
+                listener.BeginGetContext(EndGetContext, listener);
+            }
+
+            public void EndGetContext(IAsyncResult result)
+            {
+                HttpListener httpListener = (HttpListener)result.AsyncState!;
+                if (httpListener.IsListening)
+                {
+                    HttpListenerContext? context = null;
+                    try
+                    {
+                        context = httpListener.EndGetContext(result);
+
+                        _requests.Add(context.Request);
+
+                        string? authorization = context.Request.Headers["Authorization"];
+
+                        if (authorization == null)
+                        {
+                            context.Response.StatusCode = 401;
+                            context.Response.AddHeader("WWW-Authenticate", "Basic");
+                        }
+                        else
+                        {
+                            context.Response.StatusCode = 200;
+                        }
+
+                        _output.WriteLine($"Got request for {context.Request.Url}. Auth: {authorization}");
+                    }
+                    catch (Exception ex)
+                    {
+                        if (context != null)
+                        {
+                            context.Response.StatusCode = 500;
+                            using (var textStream = new StreamWriter(context.Response.OutputStream, Encoding.UTF8))
+                            {
+                                textStream.Write(ex.ToString());
+                            }
+                        }
+                    }
+
+                    context?.Response.Close();
+
+                    httpListener.BeginGetContext(EndGetContext, httpListener);
+                }
             }
 
             public void Stop()
@@ -156,48 +201,11 @@ namespace NuGet.Protocol.FuncTest
                 _httpListener = null;
                 _baseUrl = null;
 
+                listener.Stop();
                 listener.Close();
 
                 _serverThread?.Join();
                 _serverThread = null;
-            }
-
-            private void ProcessRequests()
-            {
-                if (_httpListener == null)
-                {
-                    throw new InvalidOperationException("_httpListener must be created");
-                }
-
-                try
-                {
-                    while (true)
-                    {
-                        HttpListenerContext context = _httpListener.GetContext();
-
-                        _requests.Add(context.Request);
-
-                        string? authorization = context.Request.Headers["Authorization"];
-
-                        if (authorization == null)
-                        {
-                            context.Response.StatusCode = 401;
-                            context.Response.Headers.Add("WWW-Authenticate", "Basic");
-                        }
-                        else
-                        {
-                            context.Response.StatusCode = 200;
-                        }
-
-                        _output.WriteLine($"Got request for {context.Request.Url}. Auth: {authorization}");
-
-                        context.Response.Close();
-                    }
-                }
-                catch (Exception ex)
-                {
-                    string msg = ex.Message;
-                }
             }
         }
     }
--- a/test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
+++ b/test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
@@ -15,11 +15,19 @@ using NuGet.Configuration;
 using NuGet.Protocol.Core.Types;
 using NuGet.Test.Server;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace NuGet.Protocol.FuncTest
 {
     public class AuthenticationHandlerTests
     {
+        private readonly ITestOutputHelper _output;
+
+        public AuthenticationHandlerTests(ITestOutputHelper output)
+        {
+            _output = output;
+        }
+
         [Fact]
         public async Task OnlyOneRequestWithoutAuthorizationHeader()
         {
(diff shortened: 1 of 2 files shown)


jeffkl (dismissed):
🎉

test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
```diff
@@ -0,0 +1,212 @@
...
+                return Task.FromResult(string.Empty);
+            }
+        }
+
+        private class RequestCollectingServer
+        {
+            private string? _baseUrl;
+            private HttpListener? _httpListener;
+            private List<HttpListenerRequest> _requests = new();
+            private Thread? _serverThread;
```
kartheekp-ms: It looks like this `_serverThread` is not declared and only used in the `Server.Stop` method. Can we remove this unused variable?

test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
```diff
@@ -0,0 +1,212 @@
...
+            public void Stop()
+            {
+                if (_httpListener == null)
+                {
+                    throw new InvalidOperationException("Not started");
+                }
+
+                var listener = _httpListener;
+                _httpListener = null;
+                _baseUrl = null;
```
kartheekp-ms, line 211: I see a test added in the PR failed `Cannot access a disposed object.` Is it because `_baseURL` is set to null before invoking `Stop` and `Close` on the listener?

zivkan: no, it's just differences between how HttpListener works on .NET(Core) vs .NET Framework 😞

test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
zivkan: Oops, yes. Forgot to clean it up after migrating the HttpListener's request processing to the async APIs.

jeffkl (dismissed):

test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
```diff
@@ -0,0 +1,215 @@
...
+                listener.Prefixes.Add(baseUrl);
+                listener.AuthenticationSchemes = AuthenticationSchemes.Basic | AuthenticationSchemes.Anonymous;
+                listener.Start();
+                _httpListener = listener;
+                _baseUrl = baseUrl;
+
+                listener.BeginGetContext(EndGetContext, listener);
+            }
+
+            public void EndGetContext(IAsyncResult result)
```
kartheekp-ms: nit: Using a name like `ProcessRequest` for this method would be a better option, as it can help readers to easily understand that this method is responsible for processing the http request. Not a blocking comment.

test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
```diff
@@ -0,0 +1,215 @@
...
+                    server.Stop();
+                }
+
+                // Assert
+                Assert.Equal(
+                    1,
+                    server.Requests.Count(RequestWithoutAuthorizationHeader));
+                Assert.Equal(
+                    2,
+                    server.Requests.Select(r => r.RawUrl).Distinct().Count());
```
kartheekp-ms:
``` suggestion
                Assert.Equal(
                    3,
                    server.Requests.Select(r => r.RawUrl).Count());
```
In my opinion, the initial request returned a 401 response which prompted the client to provide basic credentials, effectively resulting in two requests from the server's perspective. To ensure that the `v3/flatcontainer/packageid/1.2.3/index.json` URL can be accessed without any additional round trips, it would be beneficial to verify a total of 3 requests. What do you think about this approach?

donnie-msft: I think I agree with Kartheek. 
In fact, may be better to just Assert on the URL string rather than on a count. Consider making a variable for the 2-3 URLs in this test and then assert on that variable explicitly.
Quick psuedocode:
```suggestion
                Assert.Contains(
                    v3IndexJson,
                    server.Requests.Select(r => r.RawUrl));

                Assert.Contains(
                    v3FlatcontainerPackageIdIndexjson,
                    server.Requests.Select(r => r.RawUrl));                    
```

test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
```diff
@@ -0,0 +1,215 @@
...
+                return string.IsNullOrEmpty(value);
+            }
+
+            static Task<string> ProcessResponse(HttpSourceResult result)
+            {
+                return Task.FromResult(string.Empty);
+            }
+        }
+
+        private class RequestCollectingServer
```
donnie-msft, line 101: The class appears generic enough to put alongside similar ones in our Test.Utility/TestServer folder.

test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
donnie-msft: I agree because HttpListener uses this name for a function that returns a context, whereas this is just a class method to execute.

test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
zivkan: GitHub was having an outage yesterday, so my attempts at replying failed 😞 

I disagree with "check if total number of requests is 3", because if the test is refactored, by someone not fully understanding the intent, to make 2 requests to the same URL, this would still pass the assertion. But it wouldn't validate that querying a package after querying the service index is also pre-authenticated. The important thing for us is that package URLs, which are typically subdirectories of the service index, get pre-authenticated.

Asserting the 2 URLs individually, but just feels to me like a different way of asserting more or less than same thing 🤷

test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
zivkan: `MockServer` (which interestingly isn't in the `TestServer` subdirectory) is useful generally, because it mimics a NuGet feed, allowing (SLOOOOWWWWWW) tests to run against a server that implements the NuGet protocol. Frankly I think that MockServer is a test smell (test code smell? test design smell?), but I need to avoid getting too side tracked. MockServer doesn't allow us to validate authenticated feeds (accept both Basic and Anonymous auth schemes, and "manually" force 401 responses on anonymous requests). I know this isn't what you suggested, but there's no benefit in extending `MockServer` to do this, as it's something that only needs to be validated in NuGet.Protocol, and would only add complexity, and maybe slow down, TestServer.

Addressing your feedback, for the same reason, this test server not implementing the NuGet protocol, makes it not useful outside of NuGet.Protocol's func tests. I can't think of any other component in NuGet where we'd want it to be able to make arbitrary HTTP requests. The server handling both authenticated and anonymous requests shouldn't be useful to anything outside of validating both anonymous and authenticated requests, which is only relevant to tests validating `HttpMessageHandler.PreAuthenticate = true`. So, I'm not sure it would even be useful to other tests in NuGet.Protocol.FuncTests, unless the `AuthenticationSchemes` and request handler are extensible, at which time it becomes a light wrapper around `HttpListener`.

kartheekp-ms (approved):

donnie-msft (approved):

PR #5159, opened 2023-05-04
Dotnet NuGet Config CLI Config Unset Command

dainizzz:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12477

## Description

Adds the `dotnet nuget config unset command`, which removes a specified NuGet configuration setting in the config section. By default, the user-level NuGet configuration file is used, and users can designate which configuration file they want to update using the option `--configfile`.
## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


martinrrm:
@dainizzz Looking at the conflicts, I see that most of them are because you have code for the get command, I think that if you delete that code most of the conflicts will be fixed

jeffkl (approved):

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -828,6 +828,23 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
...
+    <value>Removes the key-value pair from a specified NuGet configuration setting.</value>
+  </data>
+  <data name="ConfigUnsetConfigFileDescription" xml:space="preserve">
+    <value>The NuGet configuration file from which settings will be removed.</value>
+  </data>
+  <data name="ConfigUnsetConfigKeyDescription" xml:space="preserve">
+    <value>The key of the settings that are to be removed.</value>
+  </data>
+  <data name="Error_ConfigUnsetNonExistingKey" xml:space="preserve">
+    <value>The config key "{0}" could not be unset.</value>
```
heng-liu: Users might want to know why it could not be unset. If the use keys. How about `The key "{0}" could not be found in config section, so it could not be unset.`?

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -828,6 +828,23 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
   <data name="ConfigCommandKeyNotFound" xml:space="preserve">
     <value>Key '{0}' not found.</value>
   </data>
+  <data name="Error_ConfigSetInvalidKey" xml:space="preserve">
+    <value>'{0}' is not a valid config key.</value>
```
heng-liu: How about <value>'{0}' is not a valid key in config section.</value>?
Users might misunderstand that they could set keys in other sections in settings.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -74,6 +74,23 @@ public static void Run(ConfigGetArgs args, Func<ILogger> getLogger)
...
 
+    internal static class ConfigUnsetRunner
+    {
+        public static void Run(ConfigUnsetArgs args, Func<ILogger> getLogger)
+        {
+            RunnerHelper.EnsureArgumentsNotNull(args, getLogger);
+            RunnerHelper.ValidateConfigKey(args.ConfigKey);
+            ISettings settings = string.IsNullOrEmpty(args.ConfigFile)
+                ? RunnerHelper.GetSettingsFromDirectory(null)
+                : RunnerHelper.GetSettingsFromFile(args.ConfigFile);
```
heng-liu: This will only load the settings from the specified config file, right?
When user specifies `--configfile` in `unset` command, what would happen if the config file path doesn't exist? (e.g. user enters a wrong path)
Will we show error message saying the config file doesn't exist? Thanks!

heng-liu (commented):
Great job! I left a few comments. Please let me know if you have any questions.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -192,5 +223,26 @@ public static void EnsureArgumentsNotNull<TArgs>(TArgs args, Func<ILogger> logge
...
+                && !configKey.Equals(ConfigurationConstants.DefaultPushSource, StringComparison.OrdinalIgnoreCase)
+                && !configKey.Equals(ConfigurationConstants.HostKey, StringComparison.OrdinalIgnoreCase)
+                && !configKey.Equals(ConfigurationConstants.UserKey, StringComparison.OrdinalIgnoreCase)
+                && !configKey.Equals(ConfigurationConstants.PasswordKey, StringComparison.OrdinalIgnoreCase)
+                && !configKey.Equals(ConfigurationConstants.NoProxy, StringComparison.OrdinalIgnoreCase)
+                && !configKey.Equals(ConfigurationConstants.MaxHttpRequestsPerSource, StringComparison.OrdinalIgnoreCase)
+                && !configKey.Equals(ConfigurationConstants.SignatureValidationMode, StringComparison.OrdinalIgnoreCase))
+            {
+                throw new CommandException(string.Format(CultureInfo.CurrentCulture, Strings.Error_ConfigSetInvalidKey, configKey));
+            }
```
kartheekp-ms: A potential issue with this approach is that as new entries are added to the `NuGet.Config` file, it requires update to be made to this logic in order for the command to remain functional.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
```diff
@@ -74,6 +74,23 @@ public static void Run(ConfigGetArgs args, Func<ILogger> getLogger)
...
+        {
+            RunnerHelper.EnsureArgumentsNotNull(args, getLogger);
+            RunnerHelper.ValidateConfigKey(args.ConfigKey);
+            ISettings settings = string.IsNullOrEmpty(args.ConfigFile)
+                ? RunnerHelper.GetSettingsFromDirectory(null)
+                : RunnerHelper.GetSettingsFromFile(args.ConfigFile);
+
+            if (!SettingsUtility.DeleteConfigValue(settings, args.ConfigKey))
+            {
+                throw new CommandException(string.Format(CultureInfo.CurrentCulture, Strings.Error_ConfigUnsetNonExistingKey, args.ConfigKey));
```
zivkan: I think the "key does not exist" scenario might warrant some customer development. If a key is not already set, then the nuget.config files are already in the customer's desired state. If a customer is running this is a CI script, do we want to fail their build because their machine state is already in the desired state?

Also, consider `git config --unset user.invalid`. It doesn't output any text, but it does set a non-zero exit code. So, perhaps this is precedent that we return non-zero exit codes.

.NET APIs like `List<T>.Remove`, `Dictionary<TKey, TValue>Remove`, they don't throw when the item isn't found, though they do have a `bool` return value saying whether anything was actually removed.

We could have a comment line option to allow customers to opt-in/out of error scenarios. If we default to "zero exit code", then I'd suggest `--strict` as the option to error when the key doesn't exist. If we default to strict mode, I don't have a good suggestion for an `--arg` that customers could use in scripts to avoid causing the script to error.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
dainizzz: In [the PR for the Set command](https://github.com/NuGet/NuGet.Client/pull/5150#discussion_r1192428485), Andy suggested:

> A new public API in ConfigurationConstants which contains an IReadOnlyList<string> of all the config options, has a better chance of being maintained correctly when new config keys are added.

I'll bring these changes into this branch as well.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -59,7 +60,7 @@ public void ConfigPathsCommand_HelpMessage_Success()
                 Directory.GetCurrentDirectory(),
                 $"{XplatDll} config paths --help",
                 waitForExit: true);
-
+
```
donnie-msft: undo whitespace

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
donnie-msft: Andy's point makes sense to me. The consequences are lower for unsetting a config value which doesn't exist. This reminds me of how CLI environment variables work: they return nothing if the variable doesn't exist.  I'm not sure whether we need a switch or verbosity check, like `--strict` or `--verbosity:quiet`/`--quiet`. Would be worth exploring other dotnet commands to see what's the best practice here.

In contrast, `git` always errors when attempting to delete a non-existing branch. Being wrong about a branch name existing is highly likely user error, and should probably fail any script by default, therefore you get: `error: myBranch 'branch-THISDOESNOTEXIST' not found.`.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -349,6 +393,65 @@ public void ConfigGetCommand_NullGetLogger_Fail()
...
+        }
+
+        [Fact]
+        public void ConfigUnsetCommand_NullArgs_Fail()
+        {
+            // Arrange
+            var log = new TestCommandOutputLogger();
+
+            // Act & Assert
+            Assert.Throws<ArgumentNullException>(() => ConfigUnsetRunner.Run(null, () => log));
```
donnie-msft:
```suggestion
            Assert.Throws<ArgumentNullException>(() => ConfigUnsetRunner.Run(args: null, () => log));
```

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -349,6 +393,65 @@ public void ConfigGetCommand_NullGetLogger_Fail()
...
+        [Fact]
+        public void ConfigUnsetCommand_NullGetLogger_Fail()
+        {
+            // Arrange
+            var args = new ConfigUnsetArgs()
+            {
+            };
+
+            // Act & Assert
+            Assert.Throws<ArgumentNullException>(() => ConfigUnsetRunner.Run(args, null));
```
donnie-msft:
```suggestion
            Assert.Throws<ArgumentNullException>(() => ConfigUnsetRunner.Run(args, getLogger: null));
```

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatConfigTests.cs
```diff
@@ -349,6 +393,65 @@ public void ConfigGetCommand_NullGetLogger_Fail()
...
+            Assert.Throws<ArgumentNullException>(() => ConfigUnsetRunner.Run(null, () => log));
+        }
+
+        [Fact]
+        public void ConfigUnsetCommand_NullGetLogger_Fail()
+        {
+            // Arrange
+            var args = new ConfigUnsetArgs()
+            {
+            };
```
donnie-msft:
```suggestion
            var args = new ConfigUnsetArgs();
```

donnie-msft (commented):
Thanks for another easily digestible PR! 
One big picture question, I know that the existing command logic uses [Microsoft.Extensions.CommandLineUtils](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.commandlineutils?view=dotnet-plat-ext-1.1) and thius PR will add to that. We have a goal of removing all of our implementation for this extension here: https://github.com/NuGet/Client.Engineering/issues/1772. 
Have you acknowledged this and decided it's out of scope for this work?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
dainizzz: Yes, they will see an error message!

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/ConfigCommands/ConfigRunners.cs
dainizzz: Got it! The spec mentions that a message should be displayed to the user if the key could not be unset, so I'll change this to just log the message instead of throwing an exception.

donnie-msft (approved):

dainizzz:
@donnie-msft I actually didn't know that we were planning to remove `Microsoft.Extensions.CommandLineUtils`; thanks for mentioning it! Edit: I spoke with Heng and she also hadn't heard about this, so I'm going to bring it up in the client channel and see if anyone has strong opinions regarding whether this needs to use `System.CommandLine` or not.

martinrrm (approved):

PR #5160, opened 2023-05-05
Enable NuGetAudit by default on .NET 8 and higher

### LINKED ISSUE #12568: NuGetAudit should be on by default with the .NET 8 SDK
### NuGet Product(s) Affected

dotnet.exe

### Current Behavior

NuGetAudit was recently merged into NuGet. In order to use the feature, it's necessary for customers to opt in with an MSBuild property `<NuGetAudit>true</NuGetAudit>`.

### Desired Behavior

The default should be the same (opt-in) for customers using the .NET 7 SDK and earlier. But for early .NET 8 preview SDKs, we'd like to try making NuGetAudit opt-out, so on by default. Depending on feedback, we can revert it to opt-in before .NET 8 gets to RC1, but we think opt-out will give us great feedback during .NET 8's previews.

### Additional Context

Feature spec: https://github.com/NuGet/Home/blob/dev/proposed/2022/vulnerabilities-in-restore.md

NuGetAudit will check for packages with known vulnerabilities during restore.

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12568

Regression? No

## Description

Enable NuGetAudit by default when customers are using the .NET 8 SDK or higher.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/test/NuGet.Core.FuncTests/Dotnet.Integration.Test/ProjectPropertyTests.cs
+++ b/test/NuGet.Core.FuncTests/Dotnet.Integration.Test/ProjectPropertyTests.cs
@@ -27,8 +27,14 @@ namespace Dotnet.Integration.Test
             var projectFileContents = @"<Project Sdk=""Microsoft.NET.Sdk"">
     <Target Name=""ValidateNuGetAuditValue"">
         <Message Importance=""high"" Text=""NuGetAudit: '$(NuGetAudit)'"" />
-        <Error Text=""NuGetAudit property should be set to true by default"" Condition=""'$(NuGetAuditExpected)' == 'true' AND '$(NuGetAudit)' != 'true'"" />
-        <Error Text=""NuGetAudit property should not be set to true by default"" Condition=""'$(NuGetAuditExpected)' != 'true' AND '$(NuGetAudit)' == 'true'"" />
+        <Error Text=""ExpectEnabled was not set"" Condition="" '$(ExpectEnabled)' != 'true' AND '$(ExpectEnabled)' != 'false' "" />
+        <Message Importance=""high"" Text=""SdkVersion: $(NETCoreSdkVersion) "" />
+    </Target>
+    <Target Name=""ValidateEnabled"" Condition="" '$(ExpectEnabled)' == 'true' "" AfterTargets=""ValidateNuGetAuditValue"">
+        <Error Text=""NuGetAudit property should be set to true by default"" Condition="" '$(NuGetAudit)' != 'true' "" />
+    </Target>
+    <Target Name=""ValidateNotSet"" Condition="" '$(ExpectEnabled)' != 'true' "" AfterTargets=""ValidateNuGetAuditValue"">
+        <Error Text=""NuGetAudit property should not be set"" Condition="" '$(NuGetAudit)' != '' "" />
     </Target>
 </Project>";
             File.WriteAllText(projectFilePath, projectFileContents);
@@ -44,7 +50,7 @@ namespace Dotnet.Integration.Test
 #endif
 
             // Act
-            var result = _msbuildFixture.RunDotnet(testDirectory.Path, $"msbuild -t:ValidateNuGetAuditValue -p:NuGetAuditExpected={expected}");
+            var result = _msbuildFixture.RunDotnet(testDirectory.Path, $"msbuild -t:ValidateNuGetAuditValue -p:ExpectEnabled={expected}");
 
             // Assert
             result.Success.Should().BeTrue(result.AllOutput);
--- a/test/NuGet.Core.FuncTests/Dotnet.Integration.Test/ProjectPropertyTests.cs
+++ b/test/NuGet.Core.FuncTests/Dotnet.Integration.Test/ProjectPropertyTests.cs
@@ -1,0 +1,53 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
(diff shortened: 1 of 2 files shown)


test/NuGet.Core.FuncTests/Dotnet.Integration.Test/ProjectPropertyTests.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        private MsbuildIntegrationTestFixture _msbuildFixture;
+
+        public ProjectPropertyTests(MsbuildIntegrationTestFixture fixture)
+        {
+            _msbuildFixture = fixture;
+        }
+
+        [Fact]
+        public void NuGetAudit_EnabledByDefaultOnNet8AndHigher()
```
jeffkl, line 22: I think this should be a unit test, these integration/functional tests are slow and are good for testing something end-to-end like restore, build, and pack as one operation.  What do you think about just making it a unit test?

zivkan: Sure, I'd prefer a unit test, but I'm not really sure how to design it. When I started this PR, I was thinking I really want to have the patched SDK evaluate a project file.  But I guess if we load the NuGet.targets file via MSBuild APIs, and set NETCoreSdkVersion as a global variable, in order to "mock" the SDK, then that's good enough?

Otherwise my only other idea is to use an MSBuild Task with an output, but that's really overkill for what without unit testing is achieved with a single property with a condition.

However, looking at the CI failures, particularly the mono tests failing because the MSBuild intrinsic function to compare versions doesn't exist, is going to make adding this rather difficult.

jeffkl: Yeah it looks like you'll want to use `$([MSBuild]::ValueOrDefault())`:

```
$([MSBuild]::VersionGreaterThanOrEquals($([MSBuild]::ValueOrDefault('$(NETCoreSdkVersion)', '1.0')), '8.0'))
```

Your unit tests could check what happens if `NETCoreSdkVersion` isn't set, its set to lower than 8.0, and higher than 8.0, as well as if the value for `NuGetAudit` is left alone if set by the user.  But I don't see any unit tests that use MSBuild APIs to evaluate a project?

jeffkl (dismissed):

zivkan:
Doing a bit of testing of the NuGetAudit feature, it might not yet be ready to turn on by default.  Will re-open a similar PR when ready.

donnie-msft:
There's an email thread with a few questions about this feature. Are there any potential modifications to make to this PR based on that, or should we review now?

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/ProjectPropertyTests.cs
```diff
@@ -0,0 +1,59 @@
...
+
+using System.IO;
+using FluentAssertions;
+using NuGet.Test.Utility;
+using Xunit;
+
+namespace Dotnet.Integration.Test
+{
+    [Collection(DotnetIntegrationCollection.Name)]
+    public class ProjectPropertyTests
```
donnie-msft, line 12: Maybe just use this fixture since I don't think you need what's in this Collection?

```suggestion
    public class ProjectPropertyTests : ICollectionFixture<MsbuildIntegrationTestFixture>,
```

Edit:  Looks like the cleanup may be "expensive-ish" so perhaps better to keep using the existing Collection?

zivkan:
> There's an email thread with a few questions about this feature. Are there any potential modifications to make to this PR based on that, or should we review now?

Things are changing rapidly/frequently, but it seems to me like we will be enabling this for .NET 8, so unless another new thing pops up soon, it's ready for review.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/ProjectPropertyTests.cs
zivkan: I had another go at trying to write a test using MSBuild APIs to load the targets file (in NuGet.Build.Tasks.Test). Trying to load the project fails at runtime due to Microsoft.Build.dll not being found. So, I tried to use Microsoft.Build.Locator, but now xUnit won't load the test assembly because the dependant (MSBuild) assemblies can't be found.

So, for now I'm giving up on trying to use MSBuild APIs to load the targets file directly. The dotnet integration test is slow, but it works.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/ProjectPropertyTests.cs
zivkan: Is it possible to use `ICollectionFixture` without `CollectionAttribute`?  Or did you typo, and mean `IClassFixture`?

But yes, `MsbuildIntegrationTestFixture` is very expensive to setup and cleanup (mostly setup), which is why all tests in the project use the collection, so it only has to be done once. Unfortunately, this also means that tests don't run in parallel, so maybe we should investigate alternatives (a static class), but for now this is how we copy the .NET SDK and patch it with the current build of NuGet.

jeffkl (approved):

nkolev92 (approved):

PR #5167, opened 2023-05-12
Migrate all NuGet.Client projects to SDK style

zivkan:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/273

Regression? No

## Description

Migrate all non-SDK style projects in the solution to SDK style. VS still providers designers for WPF and WinForms, so I don't believe there's any downside to this.  I tested that code changes and F5 debug into VS experimental instance still works. The solution can now be restored and evaluated with the `dotnet` CLI, but the VSSDK.BuildTools tasks only run on msbuild, so we can't build with the dotnet CLI.

The changes are mostly "simple" conversions, but there's a few non-trivial things that needed to be done:

* SDK style WinForms resources always use project default namespace, forcing all the PM UI options pages to change namespace
* AfterBuild targets needed to be changed due to targets files import order changing.
* Projects using VSSDK.BuildTools have a minor issue. See comment below.
* NuGet.VisualStudio.Client.csproj needs the `<Name>` metadata on all the `<ProjectReference` items that are defined in the vsixmanifest file, because the VSSDK BuildTools needs it. We also need all the ngen metadata. We can convert the syntax to XML attributes rather than child elements, but I don't find either syntax better or worse than the other.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Clients/NuGet.SolutionRestoreManager/Directory.Build.targets
```diff
@@ -0,0 +1,4 @@
+<Project>
+    <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.targets', '$(MSBuildThisFileDirectory)../'))" />
+    <Import Project="$(PkgMicrosoft_VSSDK_BuildTools)\tools\VSSDK\Microsoft.VsSDK.targets" Condition="'$(PkgMicrosoft_VSSDK_BuildTools)' != ''" />
```
zivkan, line 3: The `Microsoft.VsSDK.targets` file needs to be imported AFTER all the .NET/msbuild common targets are imported. Since the project file uses `<Project Sdk="Microsoft.NET.Sdk>`, the project file no longer explicitly imports the targets in the file, allowing us to add extra MSBuild script after the targets import.

However, [MSBuild allows you to explicitly import the SDK's props and targets files in the project file](https://learn.microsoft.com/en-us/visualstudio/msbuild/how-to-use-project-sdk?view=vs-2022). If we change to this syntax, we shouldn't need to add these `Directory.Build.targets` files.  What do people think?

For what it's worth, I also tried moving this import to `build/common.targets`, but for reasons I don't understand, it caused two resources files in the `NuGet.Tools` project to have the same output path, which failed the build.

setup/Microsoft.VisualStudio.NuGet.BuildTools.Swix/Microsoft.VisualStudio.NuGet.BuildTools.Swix.swixproj
```diff
@@ -14,7 +14,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <ReferenceOutputPath>$(ArtifactsDirectory)NuGet.VisualStudio.Client\bin\$(Configuration)\</ReferenceOutputPath>
+    <ReferenceOutputPath>$(ArtifactsDirectory)NuGet.VisualStudio.Client\bin\$(Configuration)\net472\</ReferenceOutputPath>
```
jeffkl, line 17: An alternative would be to set `<AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>` in `NuGet.VisualStudio.Client.csproj` since it won't ever multi-target.

src/NuGet.Clients/NuGet.Console/NuGet.Console.csproj
```diff
@@ -1,157 +1,30 @@
...
-    <Compile Include="Resources.Designer.cs">
-      <AutoGen>True</AutoGen>
-      <DesignTime>True</DesignTime>
-      <DependentUpon>Resources.resx</DependentUpon>
-    </Compile>
-    <Compile Include="GlobalSuppressions.cs" />
-    <Compile Include="Properties\AssemblyInfo.cs" />
-  </ItemGroup>
-  <ItemGroup>
-    <EmbeddedResource Include="Resources.resx">
```
jeffkl, line 103: SDK-style projects usually need an `<EmeddedResource Update="" />` to make things look right in VS.  

https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Versioning/NuGet.Versioning.csproj#L20-L33

src/NuGet.Clients/NuGet.Console/NuGet.Console.csproj
```diff
@@ -1,157 +1,30 @@
...
-      <Project>{eea49a74-6efc-410e-9745-bad367ac151d}</Project>
-      <Name>NuGet.VisualStudio.Common</Name>
-    </ProjectReference>
-    <ProjectReference Include="..\NuGet.VisualStudio.Internal.Contracts\NuGet.VisualStudio.Internal.Contracts.csproj">
-      <Project>{deebf1a1-1c21-4e9e-8ce3-1eb21381ff1e}</Project>
-      <Name>NuGet.VisualStudio.Internal.Contracts</Name>
-    </ProjectReference>
-  </ItemGroup>
-  <ItemGroup>
-    <Page Include="Xamls\ConsoleContainer.xaml">
```
jeffkl, line 148: Does this `<Page />` item get added still?

src/NuGet.Clients/NuGet.SolutionRestoreManager/Directory.Build.targets
jeffkl: I think this more elegant, I'm hoping they fix it in the future too so its even easier to have and SDK-style project for VSIX.

Here's the alternative in action for others to compare: https://github.com/microsoft/slngen/blob/main/src/Microsoft.VisualStudio.SlnGen.Extension/Microsoft.VisualStudio.SlnGen.Extension.csproj#L49-L50

src/NuGet.Clients/NuGet.VisualStudio.Client/NuGet.VisualStudio.Client.csproj
```diff
@@ -97,62 +84,50 @@
       <Link>PackageIconMonikers.imagemanifest</Link>
       <IncludeInVSIX>true</IncludeInVSIX>
     </Content>
+    <Content Include="extension.vsixlangpack">
```
jeffkl, line 87: `<Content />` items are added automatically by the SDK if `EnableDefaultContentItems` is not disabled.  These should probably be `<Content Update="" />` so we don't have duplicates.

src/NuGet.Clients/NuGet.VisualStudio.Client/NuGet.VisualStudio.Client.csproj
```diff
@@ -448,11 +404,28 @@
...
   </Target>
+
+  <!--
+    The VSSDK BuildTools detects assemblies from NuGet packages via an MSBuild item metadata that only exists in
+    non-SDK style projects. Add in the NuGetSourceType=Package metadata to be compatible.
+  -->
+  <Target Name="VSSDKNuGetPackagesCompat" BeforeTargets="GetVsixSourceItems">
+    <ItemGroup>
+      <ReferenceCopyLocalPaths
+        Include="@(ReferenceCopyLocalPaths)"
```
jeffkl: Should this be an `Update`?

src/NuGet.Clients/NuGet.Console/NuGet.Console.csproj
zivkan: Yes, it's done by WPF's targets in the .NET SDK: https://github.com/dotnet/wpf/blob/e69ef93c6275be7afc066388ce0986b401233f2c/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props#L23

src/NuGet.Clients/NuGet.VisualStudio.Client/NuGet.VisualStudio.Client.csproj
zivkan: The SDK is adding this item (and almost all others still in the csproj) as `None`, not `Content`.

jeffkl (approved):
💯

PR #5168, opened 2023-05-17
NuGetAudit should check direct references by default

### LINKED ISSUE #12590: NuGetAudit should check direct PackageReferences by default
### NuGet Product(s) Affected

### Current Behavior

`NuGetAudit`, that first became available in .NET 8 preview 5, checks all (direct and transitive) packages for known vulnerabilities.

### Desired Behavior

Add a new MSBuild property `<NuGetAuditMode>` with allowed values `direct` and `transitive`.

* Direct

Check packages that are directly referenced by the project. This includes not only `<PackageReference` items in the project file, but any `<PackageReference` that might be in a *Directory.Build.props*, or even any "implicit" packages that SDKs and project systems add. Finally, the way that Central Package Management's transitive pinning works is by elevating any transitive package to a direct reference when a pinned version is avaiable, and as a consequence these packages will also be checked for known vulnerabilities.

* Transitive

All NuGet packages in the project's restore graph will be checked for known vulnerabilities, whether those packages are directly referenced by the project, or are transitive packages (dependencies of another package).

Example:

```xml
  <PropertyGroup>
    <NuGetAudit>true</NuGetAudit>
    <NuGetAuditMode>transitive</NuGetAuditMode>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="NuGet.Protocol" Version="5.11.3" />
  </ItemGroup>
</Project>
```

### Additional Context

_No response_

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12590

Regression? No

## Description

Adds `<NuGetAuditMode>` to allow customers to choose if only direct package references should be checked for known vulnerabilities, or all the packages in the restore graph. Sets the default to "direct", when not specified.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception: https://github.com/NuGet/Client.Engineering/issues/2253
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [x] Documentation PR or issue filled: https://github.com/NuGet/docs.microsoft.com-nuget/issues/3049
  - **OR**
  - [ ] N/A


src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -281,5 +290,45 @@ private int ParseAuditLevel()
...
+            else if (string.Equals("direct", auditMode, StringComparison.OrdinalIgnoreCase))
+            {
+                return NuGetAuditMode.Direct;
+            }
+            else if (string.Equals("transitive", auditMode, StringComparison.OrdinalIgnoreCase))
+            {
+                return NuGetAuditMode.Transitive;
+            }
+
+            string messageText = string.Format(Strings.Error_InvalidNuGetAuditModeValue, auditMode, "direct, transitive");
```
kartheekp-ms: The message may not be well formed if the audit mode is set to `string.Empty`. May be consider empty string as `NuGetAuditMode.Unknown`.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -281,5 +290,45 @@ private int ParseAuditLevel()
...
+            {
+                return NuGetAuditMode.Unknown;
+            }
+            else if (string.Equals("direct", auditMode, StringComparison.OrdinalIgnoreCase))
+            {
+                return NuGetAuditMode.Direct;
+            }
+            else if (string.Equals("transitive", auditMode, StringComparison.OrdinalIgnoreCase))
+            {
+                return NuGetAuditMode.Transitive;
```
kartheekp-ms: How about using Enum.TryParse() ? if it returns false then the logic can log an error.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
```diff
@@ -408,11 +408,13 @@ private async Task<PackageSpec> GetPackageSpecAsync(ISettings settings)
 
             string enableAudit = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAudit);
             string auditLevel = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditLevel);
+            string auditMode = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditLevel);
```
kartheekp-ms:
```suggestion
            string auditMode = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditMode);
```

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -152,11 +153,19 @@ private static (string severityLabel, NuGetLogCode code) GetSeverityLabelAndCode
...
+                    NuGetAuditMode.Direct => GetDirectDependencies,
+                    NuGetAuditMode.Transitive => GetAllDependencies,
+                    _ => GetDirectDependencies
+                };
 
-            foreach (var graph in _targetGraphs)
+            foreach (RestoreTargetGraph graph in _targetGraphs)
             {
-                foreach (LibraryIdentity package in graph.Flattened.Select(i => i.Key).Where(p => p.Type == LibraryType.Package))
+                foreach (ResolvedDependencyKey resolvedDependency in getPackagesToCheck(graph))
```
kartheekp-ms: Could you please confirm if we're parsing audit mode for every target framework graph? Assuming that's the case, and since audit mode is set at the project level, I believe there's no need to fetch it in the nested foreach loop.

src/NuGet.Core/NuGet.Commands/Strings.resx
```diff
@@ -1078,4 +1078,10 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
     <value>Error occurred while getting package vulnerability data: {0}</value>
     <comment>{0} is an error message</comment>
   </data>
+  <data name="Error_InvalidNuGetAuditModeValue" xml:space="preserve">
+    <value>Invalid NuGetAuditMode value '{0}'. Expected values: {1}</value>
+    <comment>Don't translate 'NuGetAuditLevel'
```
kartheekp-ms:
```suggestion
    <comment>Don't translate 'NuGetAuditMode'
```

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
zivkan: Enum.Parse is significantly slower, on .NET (CoreApp):

|       Method |                Job |            Runtime |      Mean |     Error |    StdDev |
|------------- |------------------- |------------------- |----------:|----------:|----------:|
|    EnumParse |           .NET 7.0 |           .NET 7.0 | 2.6609 ns | 0.0294 ns | 0.0246 ns |
| StringEquals |           .NET 7.0 |           .NET 7.0 | 0.1974 ns | 0.0028 ns | 0.0024 ns |
|    EnumParse | .NET Framework 4.8 | .NET Framework 4.8 | 7.2990 ns | 0.0162 ns | 0.0143 ns |
| StringEquals | .NET Framework 4.8 | .NET Framework 4.8 | 5.5150 ns | 0.0194 ns | 0.0172 ns |

<details>
<summary>code</summary>

<p>

```cs
[MemoryDiagnoser]
[SimpleJob(RuntimeMoniker.Net48)]
[SimpleJob(RuntimeMoniker.Net70)]
public class Program
{
    private static void Main() => BenchmarkRunner.Run<Program>();

    [Benchmark]
    public bool EnumParse()
    {
        return bool.Parse("true");
    }

    [Benchmark]
    public bool StringEquals()
    {
        return string.Equals("true", bool.TrueString, StringComparison.OrdinalIgnoreCase);
    }
```

</p>
</details>

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
zivkan: Good catch!  Thanks.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
zivkan: Audit mode is parsed in the method `ParseAuditMode`, which you can see if called outside any `foreach` loop. So, it's only parsed once per project.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
donnie-msft: We do similar for when an MSBuild property is blank, then you see messages like "the version '' is invalid". 
Do we like that approach of showing them that they have a blank value? Or would being more descriptive as Kartheek suggests be better?

zivkan: @donnie-msft I pushed an update just over 2 hours before your comment was posted which prevents the code getting here if the value is somehow an empty string.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
kartheekp-ms: Thanks Andy, the syntax for hooking the delegate handler was new to me. Hence wanted to confirm the behavior.

kartheekp-ms (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
donnie-msft: This was marked resolved but it is still showing `NuGetAuditLevel` .... ?

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -281,5 +290,45 @@ private int ParseAuditLevel()
...
+            string messageText = string.Format(Strings.Error_InvalidNuGetAuditModeValue, auditMode, "direct, all");
+            RestoreLogMessage message = RestoreLogMessage.CreateError(NuGetLogCode.NU1014, messageText);
+            message.ProjectPath = _projectFullPath;
+            _logger.Log(message);
+            return NuGetAuditMode.Unknown;
+        }
+
+        IEnumerable<ResolvedDependencyKey> GetDirectDependencies(RestoreTargetGraph graph)
+        {
+            Debug.Assert(graph.Graphs.Count() == 1);
```
donnie-msft: Can't there be 0 here? I see some calls (eg, in `RestoreCommand`) to create `RestoreTargetGraph` which passes an empty enumerable for `graphs`.

src/NuGet.Core/NuGet.Commands/Strings.resx
```diff
@@ -1078,4 +1078,10 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
...
+{0} - is the value from the customer's project file
+{1} is the list of valid values "direct, transitive"</comment>
```
donnie-msft:
```suggestion
{1} is the list of valid values "direct, all"</comment>
```

src/NuGet.Core/NuGet.ProjectModel/RestoreAuditProperties.cs
```diff
@@ -22,12 +21,19 @@ public class RestoreAuditProperties : IEquatable<RestoreAuditProperties>
         /// <value>low, moderate, high, critical</value>
         public string? AuditLevel { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating which audit mode to use.
+        /// </summary>
+        /// <value>direct, transitive</value>
```
donnie-msft: Either delete this line or:

```suggestion
        /// <value>direct, all</value>
```

src/NuGet.Core/NuGet.Commands/Strings.resx
donnie-msft: This was marked resolved but it is still showing NuGetAuditLevel .... ?

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
zivkan: Good find!  Looks like graphs can be empty when there are certain error conditions, which I didn't test (other error conditions still generate a graph). Assuming I correctly commit and pushed this time, this code should be resilient to those error conditions now.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
zivkan: Thank you for verifying, @donnie-msft !  I don't know what happened. I clearly remember making the changes locally last week, but my devbox didn't have staged or unstages changes, or commits that were not pushed. Apparently I undid the changes, but I have no memory of it. Unless I've been doing work in my dreams? 😕

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
```diff
@@ -851,6 +851,7 @@ Copyright (c) .NET Foundation. All rights reserved.
         <CentralPackageTransitivePinningEnabled>$(CentralPackageTransitivePinningEnabled)</CentralPackageTransitivePinningEnabled>
         <NuGetAudit>$(NuGetAudit)</NuGetAudit>
         <NuGetAuditLevel>$(NuGetAuditLevel)</NuGetAuditLevel>
+        <NuGetAuditMode>$(NuGetAuditMode)</NuGetAuditMode>
```
jeffkl, line 854: Looks like static graph-based restore wasn't updated.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -152,11 +153,19 @@ private static (string severityLabel, NuGetLogCode code) GetSeverityLabelAndCode
             Dictionary<PackageIdentity, Dictionary<PackageVulnerabilityInfo, List<string>>>? result = null;
 
             int minSeverity = ParseAuditLevel();
+            Func<RestoreTargetGraph, IEnumerable<ResolvedDependencyKey>> getPackagesToCheck =
```
jeffkl: This is kind of an unusual pattern.  You could just have a single method that returns the packages to check and it could have the logic to parse the audit mode and return the correct packages.  I think that would look more similar to the rest of our code base (unless we use this pattern elsewhere)

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
zivkan: In my previous PR you asked me to make static graph based restore to re-use NuGet.Commands' MSBuildRestoreUtility. So, the one change in that file changed both CLI restore types.

https://github.com/NuGet/NuGet.Client/blob/20618ba68b3a6cd9928d52b4c5f7aa6872747ba4/src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs#L802

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
zivkan: In my opinion, it's just like a method taking a delegate (predicate) as a parameter, just like most LINQ methods. In this case, the `foreach` loop under the highlighted line would be the method the predicate is passed to, except in this case it's inlined, not a separate method. If I refactored the foreach loop into its own method (making this method just 4 or 5 statements long), would that be better?  Or refactored `getPackagesToCheck = ChoosePredicateMethod(auditProperties.Mode)` is that better?

Something that's different, is that the predicate we need to use is different depending on the project settings, but by having multiple predicates, rather than one which has an if/switch statement, we can avoid duplicating that CPU effort every time in the outer loop. Moving business logic to be calculated once outside a loop is a pattern we use in the source already.

VS search finds `Func<` 727 times in our repo, 552 times under `src/`. Most of them are method parameters, or class fields.

I don't like it when people write fancy, but difficult to understand, code. If that's what this method is, then it's unintentional and needs to change. But I wrote it this way because 1. you don't have to scroll up/down to figure out what the logic for determining the predicate and 2. the predicate is used 2 lines underneath where it's defined, so it's easy to see where it's used.

I thought this is the most clear way to express the intention for this method's business logic 😞 I'm not sure what, specifically, is unusual or difficult to understand about it.

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
jeffkl: Okay great, sorry I missed that.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
jeffkl: To me this is hard to read, its using a `switch` to determine a delegate to then call in the foreach.  You could have a single method that returns the correct dependencies to audit:

```c#
IEnumerable<ResolvedDependencyKey> GetDependenciesToAudit(RestoreTargetGraph graph)
{
    NuGetAuditMode auditMode = ParseAuditMode();

    if (auditMode == NuGetAuditMode.All)
    {
        return graph.ResolvedDependencies.Where(dep => dep.Child.Type == LibraryType.Package);
    }

    Debug.Assert(graph.Graphs.Count() <= 1);
    LibraryIdentity? thisProject = graph.Graphs.FirstOrDefault()?.Item.Key;

    if (thisProject == null)
    {
        return Enumerable.Empty<ResolvedDependencyKey>();
    }

    Debug.Assert(thisProject.Type == LibraryType.Project);
    return graph.ResolvedDependencies.Where(dep => dep.Parent == thisProject && dep.Child.Type == LibraryType.Package);
}
```

And then the `foreach` could just call the method that returns the correct packages to audit based on the current configuration:

```c#
foreach (ResolvedDependencyKey resolvedDependency in GetDependenciesToAudit(graph))
```

In my opinion, that looks more like the rest of the code in this class and the rest of the code base.

zivkan: I kept `ParseAuditMode()` outside the foreach loop to avoid wasting CPU cycles, but now I'm no longer using a delegate, so the loop is now executing an additional `if` statement every loop. The default audit mode is also no longer explicitly defined.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -281,5 +284,50 @@ private int ParseAuditLevel()
...
+
+        private NuGetAuditMode ParseAuditMode()
+        {
+            string? auditMode = _restoreAuditProperties.AuditMode?.Trim();
+
+            if (auditMode == null)
+            {
+                return NuGetAuditMode.Unknown;
+            }
+            else if (string.Equals("direct", auditMode, StringComparison.OrdinalIgnoreCase))
```
jeffkl, line 298: These `else if`s are not needed since they each call `return`

jeffkl (approved):
Looks great!

donnie-msft (approved):

PR #5190, opened 2023-05-25
Enable source-build pre-built detection

### LINKED ISSUE #12598: Enable source-build pre-built detection
Contributes to https://github.com/dotnet/source-build/issues/3017

Enable source-build pre-build detection in `nuget.client` repo, with a custom model of tracking prebuilts. Unlike all other repost that contribute to .NET product, `nuget.client` does not use Arcade build infra. It is not currently possible to flow .NET dependencies, i.e. SBRP intermediate package. Without package flow, prebuilts can not be eliminated at the point of detection.

Specifics of this work:
- Enable prebuilt detection, ensuring that we only detect prebuilts in inner source-build
- Use baseline exclusion list (https://github.com/NuGet/NuGet.Client/blob/dev/eng/SourceBuildPrebuiltBaseline.xml) to capture all current prebuilts. All exclusions for .NET packages should use `*` version. All external packages should use current, static version
- Enable `nuget.client` contributors to communicate any new prebuilts back to source-build team, by documenting the process and/or by leaving a hint in build output of prebuilt detection step
- [Optional] Update source-build infra to enable publishing of source-build intermediate package

Process to be followed by repo contributors introducing new prebuilts:
- Follow the upcoming instructions for communicating new prebuilts to source-build team
- Add any new prebuilts to the aforementioned baseline XML, as part of the PR, ensuring source-build leg is clean

NikolaMilosavljevic:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12598

Regression? No

## Description

Enables prebuilt detection, that causes build failure for any new prebuilts not in baseline.

Since it's not possible to flow dependencies, i.e. SBRP intermediate package, all prebuilts have to be listed in prebuilt baseline: https://github.com/NuGet/NuGet.Client/blob/dev/eng/SourceBuildPrebuiltBaseline.xml

All .NET packages are excluded with `*` version, to avoid unnecessary changes in the future - i.e.:
`<UsagePattern IdentityGlob="System.Buffers/*" />`

All instances of other packages are excluded with specific version, to ensure that version update triggers proper prebuilt update process, i.e.:
`<UsagePattern IdentityGlob="Newtonsoft.Json/13.0.1" />`

### DIFF
--- a/eng/SourceBuildPrebuiltBaseline.xml
+++ b/eng/SourceBuildPrebuiltBaseline.xml
@@ -1,5 +1,99 @@
 <UsageData>
   <IgnorePatterns>
-    <UsagePattern IdentityGlob="Microsoft.SourceBuild.Intermediate.*/*" />
+    <!--
+      To ensure that source-building .NET is successful, prebuilt package detection is enabled
+      in all contributing repos.
+
+      This repo doesn't use Arcade infra and does not support dependency flow.
+      As a result, prebuilt detection and resolution is handled differently.
+
+      Repo owners or contributors:
+      If you get an error in source-build leg pointing at new prebuilt package being detected, 
+      resolve the prebuilt issue before merging your PR.
+      If that is not possible, do the following:
+        - contact source-build team
+        - add required UsagePattern entry as provided by source-build team
+        - create an issue in this repo, to address new prebuilt package.
+    -->
+
+    <!--
+      Packages not owned by .NET - exclusion of specifc package version.
+    -->
+    <UsagePattern IdentityGlob="Microsoft.VisualStudioEng.MicroBuild.Core/1.0.0" />
+    <UsagePattern IdentityGlob="Newtonsoft.Json/13.0.1" />
+    <UsagePattern IdentityGlob="System.Windows.Extensions/6.0.0" />
+
+    <!--
+      Packages owned by .NET - exclusion of all versions.
+    -->
+    <UsagePattern IdentityGlob="Microsoft.Build/*" />
+    <UsagePattern IdentityGlob="Microsoft.Build.Framework/*" />
+    <UsagePattern IdentityGlob="Microsoft.Build.Tasks.Core/*" />
+    <UsagePattern IdentityGlob="Microsoft.Build.Tasks.Git/*" />
+    <UsagePattern IdentityGlob="Microsoft.Build.Utilities.Core/*" />
+    <UsagePattern IdentityGlob="Microsoft.CSharp/*" />
+    <UsagePattern IdentityGlob="Microsoft.DotNet.Arcade.Sdk/*" />
+    <UsagePattern IdentityGlob="Microsoft.DotNet.SourceBuild.Tasks/*" />
+    <UsagePattern IdentityGlob="Microsoft.Extensions.CommandLineUtils.Sources/*" />
+    <UsagePattern IdentityGlob="Microsoft.Extensions.FileProviders.Abstractions/*" />
+    <UsagePattern IdentityGlob="Microsoft.Extensions.FileSystemGlobbing/*" />
+    <UsagePattern IdentityGlob="Microsoft.Extensions.Primitives/*" />
+    <UsagePattern IdentityGlob="Microsoft.NET.StringTools/*" />
+    <UsagePattern IdentityGlob="Microsoft.NETCore.Platforms/*" />
+    <UsagePattern IdentityGlob="Microsoft.NETCore.Targets/*" />
+    <UsagePattern IdentityGlob="Microsoft.SourceLink.AzureRepos.Git/*2" />
+    <UsagePattern IdentityGlob="Microsoft.SourceLink.Common/*" />
+    <UsagePattern IdentityGlob="Microsoft.SourceLink.GitHub/*" />
+    <UsagePattern IdentityGlob="Microsoft.Web.Xdt/*" />
+    <UsagePattern IdentityGlob="Microsoft.Win32.Registry/*" />
+    <UsagePattern IdentityGlob="Microsoft.Win32.SystemEvents/*" />
+    <UsagePattern IdentityGlob="NETStandard.Library/*" />
+    <UsagePattern IdentityGlob="System.Buffers/*" />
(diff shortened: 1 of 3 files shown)


NikolaMilosavljevic:
@mmitche @MichaelSimons @mthalman

eng/SourceBuildPrebuiltBaseline.xml
```diff
@@ -1,5 +1,98 @@
...
+      in all contributing repos.
+
+      This repo doesn't use Arcade infra and does not support dependency flow.
+      As a result, prebuilt detection and resolution is handled differently.
+
+      Repo owners or contributors:
+      If you get an error in source-build leg pointing at new prebuilt package being detected,
+      resolve the prebuilt issue before merging your PR.
+      If that is not possible, do the following:
+        - contact source-build team
```
MichaelSimons: I would include the contact info - use the dotnet/source-build-internal github team.

eng/SourceBuildPrebuiltBaseline.xml
```diff
@@ -1,5 +1,98 @@
...
+        - add required UsagePattern entry as provided by source-build team
```
MichaelSimons: I tend to think we should exclude anything besides resolving the prebuilt and contacting the source-build team.  I fear that suggesting creating UsagePatterns is going to lead to the problem we are trying to prevent.

eng/source-build/source-build.proj
```diff
@@ -28,7 +28,7 @@
     <Message Text="Finished restoring Arcade" Importance="High" />
 
     <MSbuild Projects="$(ArcadeDir)/SourceBuild/AfterSourceBuild.proj"
-             Properties="_NETCORE_ENGINEERING_TELEMETRY=AfterSourceBuild;ArcadeBuildFromSource=true;CurrentRepoSourceBuildArtifactsPackagesDir=$(ProjectRoot)artifacts/nupkgs/"
+             Properties="_NETCORE_ENGINEERING_TELEMETRY=AfterSourceBuild;ArcadeBuildFromSource=true;CurrentRepoSourceBuildArtifactsPackagesDir=$(ProjectRoot)artifacts/nupkgs/;FailOnPrebuiltBaselineError=true"
```
MichaelSimons: Does this negatively affect the product source-build?  When doing a full source-build, we don't want this repo to stop the build if it has prebuilts.

eng/source-build/build.sh
```diff
@@ -83,4 +83,5 @@ fi
 
 ReadGlobalVersion Microsoft.DotNet.Arcade.Sdk
 export ARCADE_VERSION=$_ReadGlobalVersion
+export NUGET_PACKAGES=$scriptroot/../../artifacts/source-build/self/package-cache/
```
MichaelSimons: Is this used/exist in the full repo source-build?

NikolaMilosavljevic: It is set in full source-build: https://github.com/dotnet/dotnet/blob/8528c489bcbaeff232b906a3c784016f2de74fe2/build.sh#L248

I think I need to condition this for repo build only.

eng/source-build/source-build.proj
NikolaMilosavljevic: You're right - it will affect full source-build. I think we should be able to make this change specific to repo source-build. Will update - thanks.

eng/SourceBuildPrebuiltBaseline.xml
NikolaMilosavljevic: Makes sense - we want source-build team to provide guidance on a case-by-case basis. Will update - thanks.

eng/SourceBuildPrebuiltBaseline.xml
```diff
@@ -1,5 +1,98 @@
...
+        - create an issue in this repo, to address new prebuilt package.
+    -->
+
+    <!--
+      Packages not owned by .NET - exclusion of specifc package version.
```
mthalman:
```suggestion
      Packages not owned by .NET - exclusion of specific package version.
```

eng/SourceBuildPrebuiltBaseline.xml
NikolaMilosavljevic: Resolved in https://github.com/NuGet/NuGet.Client/pull/5190/commits/eaaaebcd571a201b0bda7ae3e49767824975471a

eng/SourceBuildPrebuiltBaseline.xml
NikolaMilosavljevic: Resolved in https://github.com/NuGet/NuGet.Client/pull/5190/commits/eaaaebcd571a201b0bda7ae3e49767824975471a

eng/source-build/source-build.proj
NikolaMilosavljevic: Resolved in https://github.com/NuGet/NuGet.Client/pull/5190/commits/eaaaebcd571a201b0bda7ae3e49767824975471a

eng/source-build/build.sh
NikolaMilosavljevic: Resolved in https://github.com/NuGet/NuGet.Client/pull/5190/commits/eaaaebcd571a201b0bda7ae3e49767824975471a

MichaelSimons (approved):

NikolaMilosavljevic:
@kartheekp-ms @zivkan can this be approved for CI? This change is specific to .NET source-build.

jeffkl (approved):

jeffkl:
> @kartheekp-ms @zivkan can this be approved for CI? This change is specific to .NET source-build.

Done!

NikolaMilosavljevic:
> > @kartheekp-ms @zivkan can this be approved for CI? This change is specific to .NET source-build.
> 
> Done!

There are some test failures - but since those are on Windows and this PR is only for Linux and source-build, they are unrelated. I see the same failures in other recent PRs.

jeffkl:
> There are some test failures - but since those are on Windows and this PR is only for Linux and source-build, they are unrelated. I see the same failures in other recent PRs.

Yes those test failures are currently known and we'll ignore them.  I'll just give the team one more day to review and I'll merge tomorrow.  Thanks for the contribution, glad this will prevent us from breaking source build going forward.

PR #5198, opened 2023-06-03
Allow FloatRange ctor for AbsoluteLatest without prerelease label

### LINKED ISSUE #12618: FloatRange throws Argument null exception on previously valid constructor
### NuGet Product Used

NuGet SDK

### Product Version

6.7.0-preview

### Worked before?

6.6

### Impact

### Repro Steps & Context

We are using the `FloatRange` in our code [as seen here](https://github.com/dotnet/templating/blob/9ad6c6fc78fbfe50f4327a84a79b0f0641f37bbe/src/Microsoft.TemplateEngine.Edge/Installers/NuGet/NugetApiPackageManager.cs#L212). Before the update we had no issues with this constructor, and later used to check for correct package versions. After the update, this constructor throws a `ArgumentNullException`.

We have a use case on the `dotnet new` CLI command line and some templates are currently (after preview 4) generating incorrectly.

### Verbose Logs

_No response_

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12618

Regression? yes, but in preview version

## Description

A few weeks ago I merged a PR adding nullable annotations (and enabling nullable checks) to NuGet.Versioning](https://github.com/NuGet/NuGet.Client/pull/5097). Due to the class' API design, the compiler can't know that the `_releaseLabel` field is not null under certain conditions, so after getting feedback in the PR, I added an additional validation, so I could use `MemberNotNullWhenAttribute`, in order to simplify API usage and keep the compiler's nullable checks happy.

Unfortunately there was an edge case that did not have an existing unit test, but was used by the .NET SDK, which regressed. Once I found the "same" scenario being tested in a unit test with `FloatRange.Parse`, I extended it to also test calling the `FloatRange` constructor directly, so the two should behave the same now.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Versioning/FloatRange.cs
+++ b/src/NuGet.Core/NuGet.Versioning/FloatRange.cs
@@ -22,7 +22,9 @@ namespace NuGet.Versioning
         /// </summary>
         /// <param name="floatBehavior">Section to float.</param>
         public FloatRange(NuGetVersionFloatBehavior floatBehavior)
-            : this(floatBehavior, new NuGetVersion(0, 0, 0), null)
+            : this(floatBehavior,
+                  minVersion: floatBehavior == NuGetVersionFloatBehavior.None ? new NuGetVersion(0, 0, 0) : new NuGetVersion(0, 0, 0, releaseLabel: "0"),
+                  releasePrefix: null)
         {
         }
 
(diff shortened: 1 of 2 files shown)


jeffkl (approved):
🎉

martinrrm (approved):

PR #5200, opened 2023-06-05
Reduce allocations in ContentItemCollection

### LINKED ISSUE #12645: ContentItemCollection.PopulateItemGroups unnecessarily causing resizes of List<T>, ContentItemCollection.PopulateItemGroups boxing List<T>.Enumerator
Internal issues for more detail: 

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1824305
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1824313

drewnoakes:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12645
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1824305
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1824313
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1837276

## Description

Changes:

1. Allow `ContentItemGroup` to receive collections as constructor arguments. This allows code that initialises these instances to take full ownership of the construction of those collections, rather than having to call `Add` on its properties in a loop during construction.
1. Clone the property dictionary in a way that allows the new object to be constructed more efficiently, avoiding collection resizes (and the temporary allocations that come with them).
1. Swap heap-allocated `Tuple<>` for stack-allocated `ValueTuple<>`, reducing temporary allocations.
1. Stop allocating an unused `List<ContentItem>` in local variable `groupItems`.
1. Typo and minor formatting.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Packaging/ContentModel/ContentItemCollection.cs
+++ b/src/NuGet.Core/NuGet.Packaging/ContentModel/ContentItemCollection.cs
@@ -6,7 +6,6 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using NuGet.Packaging;
 
 namespace NuGet.ContentModel
 {
@@ -63,7 +62,8 @@ namespace NuGet.ContentModel
             {
                 var groupPatterns = definition.GroupExpressions;
 
-                List<Tuple<ContentItem, Asset>> groupAssets = null;
+                List<(ContentItem Item, Asset Asset)> groupAssets = null;
+
                 foreach (var asset in _assets)
                 {
                     foreach (var groupPattern in groupPatterns)
@@ -71,33 +71,19 @@ namespace NuGet.ContentModel
                         var item = groupPattern.Match(asset.Path, definition.PropertyDefinitions);
                         if (item != null)
                         {
-                            if (groupAssets == null)
-                            {
-                                groupAssets = new List<Tuple<ContentItem, Asset>>(1);
-                            }
-
-                            groupAssets.Add(Tuple.Create(item, asset));
+                            groupAssets ??= new List<(ContentItem Item, Asset Asset)>(capacity: 1);
+                            groupAssets.Add((item, asset));
                         }
                     }
                 }
 
                 if (groupAssets?.Count > 0)
                 {
-                    foreach (var grouping in groupAssets.GroupBy(key => key.Item1, GroupComparer.DefaultComparer))
+                    foreach (var grouping in groupAssets.GroupBy(key => key.Item, GroupComparer.DefaultComparer))
                     {
-                        var group = new ContentItemGroup();
-
-                        foreach (var property in grouping.Key.Properties)
-                        {
-                            group.Properties.Add(property.Key, property.Value);
-                        }
-
-                        foreach (var item in FindItemsImplementation(definition, grouping.Select(match => match.Item2)))
-                        {
-                            group.Items.Add(item);
-                        }
-
-                        yield return group;
+                        yield return new ContentItemGroup(
+                            properties: new Dictionary<string, object>(grouping.Key.Properties),
+                            items: FindItemsImplementation(definition, grouping.Select(match => match.Asset)));
                     }
                 }
             }
@@ -114,7 +100,8 @@ namespace NuGet.ContentModel
             {
                 var groupPatterns = definition.GroupExpressions;
 
-                List<Tuple<ContentItem, Asset>> groupAssets = null;
+                List<(ContentItem Item, Asset Asset)> groupAssets = null;
+
                 foreach (var asset in _assets)
                 {
                     foreach (var groupPattern in groupPatterns)
@@ -122,31 +109,19 @@ namespace NuGet.ContentModel
                         var item = groupPattern.Match(asset.Path, definition.PropertyDefinitions);
                         if (item != null)
                         {
-                            if (groupAssets == null)
-                            {
-                                groupAssets = new List<Tuple<ContentItem, Asset>>(1);
-                            }
-
-                            groupAssets.Add(Tuple.Create(item, asset));
+                            groupAssets ??= new List<(ContentItem Item, Asset Asset)>(capacity: 1);
+                            groupAssets.Add((item, asset));
                         }
                     }
                 }
 
-                IList<ContentItem> groupItems = new List<ContentItem>();
                 if (groupAssets?.Count > 0)
                 {
-                    foreach (var grouping in groupAssets.GroupBy(key => key.Item1, GroupComparer.DefaultComparer))
+                    foreach (var grouping in groupAssets.GroupBy(key => key.Item, GroupComparer.DefaultComparer))
                     {
-                        var group = new ContentItemGroup();
-
-                        foreach (var property in grouping.Key.Properties)
-                        {
-                            group.Properties.Add(property.Key, property.Value);
-                        }
-
-                        groupItems = FindItemsImplementation(definition, grouping.Select(match => match.Item2));
-                        group.Items.AddRange(groupItems);
-                        contentItemGroupList.Add(group);
+                        contentItemGroupList.Add(new ContentItemGroup(
+                            properties: new Dictionary<string, object>(grouping.Key.Properties),
+                            items: FindItemsImplementation(definition, grouping.Select(match => match.Asset))));
                     }
                 }
             }
@@ -260,10 +235,11 @@ namespace NuGet.ContentModel
             return null;
         }
 
-        private IList<ContentItem> FindItemsImplementation(PatternSet definition, IEnumerable<Asset> assets)
+        private List<ContentItem> FindItemsImplementation(PatternSet definition, IEnumerable<Asset> assets)
         {
             var pathPatterns = definition.PathExpressions;
-            IList<ContentItem> itemsList = new List<ContentItem>();
+
+            List<ContentItem> items = new();
 
             foreach (var asset in assets)
             {
@@ -274,7 +250,7 @@ namespace NuGet.ContentModel
                     var contentItem = pathPattern.Match(path, definition.PropertyDefinitions);
                     if (contentItem != null)
                     {
-                        //If the item is assembly, populate the "related files extentions property".
+                        //If the item is assembly, populate the "related files extensions property".
                         if (contentItem.Properties.ContainsKey("assembly"))
                         {
                             string relatedFileExtensionsProperty = GetRelatedFileExtensionProperty(contentItem.Path, assets);
@@ -283,13 +259,13 @@ namespace NuGet.ContentModel
                                 contentItem.Properties.Add("related", relatedFileExtensionsProperty);
                             }
                         }
-                        itemsList.Add(contentItem);
+                        items.Add(contentItem);
                         break;
                     }
                 }
             }
 
-            return itemsList;
+            return items;
         }
 
         internal string GetRelatedFileExtensionProperty(string assemblyPath, IEnumerable<Asset> assets)
--- a/src/NuGet.Core/NuGet.Packaging/ContentModel/ContentItemGroup.cs
+++ b/src/NuGet.Core/NuGet.Packaging/ContentModel/ContentItemGroup.cs
@@ -15,6 +15,12 @@ namespace NuGet.ContentModel
             Items = new List<ContentItem>();
         }
 
+        internal ContentItemGroup(IDictionary<string, object> properties, IList<ContentItem> items)
+        {
+            Properties = properties;
+            Items = items;
+        }
+
         public IDictionary<string, object> Properties { get; }
 
         public IList<ContentItem> Items { get; }


zivkan:
@drewnoakes we've had feedback from customers that they enjoy learning about perf improvements in our release notes, which needs a publicly available linked issue in NuGet/Home for our automated release notes tool to find.

drewnoakes:
@zivkan given there'll be a bunch of these small fixes throughout the rest of the month, it probably makes sense to create a single issue that folks can chime in on. A lot of these fixes will be very small and specific and not very relatable to customers beyond "made stuff faster". Did you have something specific in mind?

zivkan:
I was given feedback in the past that my tiny perf improvements should be represented in release notes, so I'm just trying to ensure other peoples contributions are held to the same standard as my own.

We have a bunch of customers who use nuget's SDK packages, in addition to customers using VS, the dotnet CLI and nuget.exe.   The SDK customers in particular will understand technical explanations better than other types of customers.

Here's an example issue I created for the release notes to link to, in case that provides inspration: https://github.com/NuGet/Home/issues/12550

In this case it's a little more difficult, because someone else provided you with a list of perf improvements with zero context of what user scenario it's used in. Normally when either the NuGet team, or an external contributor, makes a perf improvement, they know exactly what scenario it's from, so it's easy for them to write "This benefits restore when a package contains ...." or similar.

src/NuGet.Core/NuGet.Packaging/ContentModel/ContentItemCollection.cs
```diff
@@ -260,10 +235,11 @@ public ContentItemGroup FindBestItemGroup(SelectionCriteria criteria, params Pat
             return null;
         }
 
-        private IList<ContentItem> FindItemsImplementation(PatternSet definition, IEnumerable<Asset> assets)
+        private List<ContentItem> FindItemsImplementation(PatternSet definition, IEnumerable<Asset> assets)
```
nkolev92, line 238: I'm curious how does this improve things? 

The ContentItemGroup constructor takes IList anyways.

nkolev92 (approved):
Looks good, but I have 1 question about a change that seems stylistic

src/NuGet.Core/NuGet.Packaging/ContentModel/ContentItemCollection.cs
zivkan: `List<T>.GetEnumerator()` returns a struct. Anything (like `IList<T>`) that uses `IEnumerator<T>` will require a memory allocation.

Unfortunately it's also true for `IReadOnly[List|Dictionary]`, so then there's a tradeoff between cost of enumeration vs compiler enforced correctness. Or stop using `foreach` and LINQ (or anything else that uses `IEnumerable<T>`), and just use an old-school `for` loop instead.

For people in Microsoft: https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/24181/Garbage-Collection?anchor=%E2%9D%8C-**avoid**-boxing-enumerators-of-built-in-types

nkolev92: I get that, but ContentItemGroup takes an IList, so we're not saving on an allocation.

drewnoakes: You're right. It's left over from an earlier version of the PR where the code was actually using the value in a `foreach`, where it did make a difference. Seems I then went and changed it to store the returned collection directly rather than copying items into a new collection, and never reverted this back. Personally I prefer using concrete types on private APIs so that internal consumers have maximum type information, and using interfaces externally so we have the ability to change implementation behind that abstraction without breaking consumers. But we can revert this line if you prefer.

drewnoakes:
@zivkan could we create a single issue for all perfathon fixes? Most of the fixes will be one-line and fairly abstract. The numbered list in the PR description here covers the main points, but I wonder how helpful any of that would be to any customer. As you mention, I don't have enough context on this repo to understand what user workflows stand to benefit from this change. All I know is that the tooling flagged this on enough user machines in the real world that a fix makes sense.

src/NuGet.Core/NuGet.Packaging/ContentModel/ContentItemCollection.cs
```diff
@@ -114,39 +100,28 @@ public void PopulateItemGroups(PatternSet definition, IList<ContentItemGroup> co
...
                         var item = groupPattern.Match(asset.Path, definition.PropertyDefinitions);
                         if (item != null)
                         {
-                            if (groupAssets == null)
-                            {
-                                groupAssets = new List<Tuple<ContentItem, Asset>>(1);
-                            }
-
-                            groupAssets.Add(Tuple.Create(item, asset));
+                            groupAssets ??= new List<(ContentItem Item, Asset Asset)>(capacity: 1);
```
kartheekp-ms, line 112: I see that the capacity of the list is set to `1` by default and there is no `break` statement in this loop. Is there any way we can estimate the default capacity to avoid the resizing in the next iteration?

drewnoakes: The size could be anywhere between zero and `_assets.Count * groupPatterns.Count`. I can't think of a good way to estimate without walking the loops twice, which would be a net-negative I'm sure.

src/NuGet.Core/NuGet.Packaging/ContentModel/ContentItemCollection.cs
nkolev92: I don't mind it since it's a private method anyways. 

I mostly curiuous whether I'm missing out on a new trick :D

nkolev92 (approved):

src/NuGet.Core/NuGet.Packaging/ContentModel/ContentItemCollection.cs
nkolev92: The capacity also used to be 1, so I'm assuming there might've been some intent behind that.

nkolev92:
The issue discussion is more a matter of policy for us. We just try to keep things consistent. 

I created a quick issue for this one. 

I have not enabled auto-merge like the other ones because there's a few discussions ongoing.

davkean:
> @drewnoakes we've had feedback from customers that they enjoy learning about perf improvements in our release notes, which needs a publicly available linked issue in NuGet/Home for our automated release notes tool to find.

I've ran into this too - has your team considered pulling the PR title if there's no bug listed?

nkolev92:
Issues tend to focus on the why, while PRs tend to focus on the what and how as well, so historically, the issue titles were more useful for automated release notes.

davkean:
Sure, but if knowing this going forward, you could always change PR title to something appropriate. At the moment you are needing to double track internal bugs and file bugs for every change which seems a bit wasteful.

src/NuGet.Core/NuGet.Packaging/ContentModel/ContentItemCollection.cs
```diff
@@ -63,41 +62,28 @@ public IEnumerable<ContentItemGroup> FindItemGroups(PatternSet definition)
...
-                        }
-
-                        foreach (var item in FindItemsImplementation(definition, grouping.Select(match => match.Item2)))
-                        {
-                            group.Items.Add(item);
-                        }
-
-                        yield return group;
+                        yield return new ContentItemGroup(
+                            properties: new Dictionary<string, object>(grouping.Key.Properties),
```
kartheekp-ms, line 85:
```suggestion
                            properties: grouping.Key.Properties,
```
How about the suggested change? If we can avoid creating a new dictionary that would be great.

src/NuGet.Core/NuGet.Packaging/ContentModel/ContentItemCollection.cs
```diff
@@ -114,39 +100,28 @@ public void PopulateItemGroups(PatternSet definition, IList<ContentItemGroup> co
...
-                        foreach (var property in grouping.Key.Properties)
-                        {
-                            group.Properties.Add(property.Key, property.Value);
-                        }
-
-                        groupItems = FindItemsImplementation(definition, grouping.Select(match => match.Item2));
-                        group.Items.AddRange(groupItems);
-                        contentItemGroupList.Add(group);
+                        contentItemGroupList.Add(new ContentItemGroup(
+                            properties: new Dictionary<string, object>(grouping.Key.Properties),
```
kartheekp-ms, line 123: https://github.com/NuGet/NuGet.Client/pull/5200#discussion_r1227141863 comment applies to here also.

src/NuGet.Core/NuGet.Packaging/ContentModel/ContentItemCollection.cs
drewnoakes: The previous code made a copy of this collection. Are you sure it's safe to share objects here? So many of the types here are mutable that it's difficult for me to reason about the safety of removing a defensive copy like this. I agree that if it's safe, however, it would help.

nkolev92: I'd just keep the risk as minimal as possible, especially since we're in the last "easy to get in code" preview.

kartheekp-ms: Looking at the references to `properties`, it looks like the dictionary is only used for lookups. 

image

I quickly made the change and ran tests locally and found that no tests have failed. I think it is a safe change.

```
 Passed!  - Failed:     0, Passed:  1929, Skipped:    37, Total:  1966, Duration: 1 m 6 s - NuGet.Packaging.Test.dll (net472)
  Passed!  - Failed:     0, Passed:  1253, Skipped:     3, Total:  1256, Duration: 18 s - NuGet.ProjectModel.Test.dll (net472)
  Passed!  - Failed:     0, Passed:  1185, Skipped:    18, Total:  1203, Duration: 2 m 9 s - NuGet.Commands.Test.dll (net472)
  Passed!  - Failed:     0, Passed:  2190, Skipped:     7, Total:  2197, Duration: 1 m 52 s - NuGet.Protocol.Tests.dll (net472)
  Passed!  - Failed:     0, Passed:   938, Skipped:     0, Total:   938, Duration: 12 s - NuGet.Versioning.Test.dll (net472)
  Passed!  - Failed:     0, Passed:    73, Skipped:     0, Total:    73, Duration: 1 s - NuGet.Shared.Tests.dll (net472)
  Passed!  - Failed:     0, Passed:   164, Skipped:     0, Total:   164, Duration: 5 s - NuGet.DependencyResolver.Core.Tests.dll (net472)
  Passed!  - Failed:     0, Passed:    80, Skipped:     0, Total:    80, Duration: 1 s - NuGet.LibraryModel.Tests.dll (net472)
  Passed!  - Failed:     0, Passed:    70, Skipped:     0, Total:    70, Duration: 9 s - NuGet.Credentials.Test.dll (net472)
  Passed!  - Failed:     0, Passed:  1726, Skipped:     0, Total:  1726, Duration: 24 s - NuGet.Frameworks.Test.dll (net472)
  Passed!  - Failed:     0, Passed:   597, Skipped:     6, Total:   603, Duration: 26 s - NuGet.Configuration.Test.dll (net472)
  Passed!  - Failed:     0, Passed:   368, Skipped:    23, Total:   391, Duration: 46 s - NuGet.Common.Test.dll (net472)
  Passed!  - Failed:     0, Passed:    73, Skipped:     0, Total:    73, Duration: 3 s - NuGet.Resolver.Test.dll (net472)
  Passed!  - Failed:     0, Passed:   136, Skipped:     0, Total:   136, Duration: 8 s - NuGet.Build.Tasks.Test.dll (net472)
  Passed!  - Failed:     0, Passed:    88, Skipped:     0, Total:    88, Duration: 10 s - NuGet.Build.Tasks.Console.Test.dll (net472)
  Passed!  - Failed:     0, Passed:    64, Skipped:     0, Total:    64, Duration: 32 s - NuGet.CommandLine.Xplat.Tests.dll (net472)
  Passed!  - Failed:     0, Passed:    33, Skipped:     2, Total:    35, Duration: 12 s - NuGet.Build.Tasks.Pack.Test.dll (net472)
  Passed!  - Failed:     0, Passed:    60, Skipped:     0, Total:    60, Duration: 16 s - Microsoft.Build.NuGetSdkResolver.Test.dll (net472)
```

May be the correct thing to do in this case is to change the type to `IReadOnlyDictionary` which can be done in follow-up PR.

drewnoakes: That looks encouraging. Sounds like a follow up for 17.8.

Let me know if there's anything else you'd like to see for this PR. Seems ready to me.

kartheekp-ms: Yup. Created https://github.com/NuGet/Home/issues/12657 to address the code review feedback.

src/NuGet.Core/NuGet.Packaging/ContentModel/ContentItemCollection.cs
```diff
@@ -283,13 +259,13 @@ private IList<ContentItem> FindItemsImplementation(PatternSet definition, IEnume
                                 contentItem.Properties.Add("related", relatedFileExtensionsProperty);
```
kartheekp-ms, line 259: I'm not entirely sure how this PR addresses the issue raised in the work item https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1837276. The bug's call stack points to a resize operation within the Dictionary, we haven't made any changes to that specific portion. My understanding is that the line of code `contentItem.Properties.Add("related", relatedFileExtensionsProperty);` is triggering the dictionary resizing.

nkolev92:
@kartheekp-ms I didn't see your comment before I merged. 

I'm sure you and @drewnoakes can work out if there are any new changes that need to be made and those can be done in a separate PR.

src/NuGet.Core/NuGet.Packaging/ContentModel/ContentItemCollection.cs
davkean: Was that the right bug? This appears to have implemented exactly what is called in the bug you listed.

drewnoakes: I think the confusion here is because the discussion in that bug talks about `List<>` but the callstack in that bug talks about `Dictionary<>`.

The list allocations have been addressed here. The dictionary one has not, as far as I can tell.

I'll pull down the `dev` branch now that this has merged to see whether there's more we can do here.

@davkean I'll leave it to you to decide whether this closes that bug or not.

davkean: Okay, I understand. The call stack is an example call stack from that failure - there are many in this case. In this case I looked at the trace and determined that there was easy fixable path and opened a bug from the failure. The tooling randomly picked a stack from all the stacks and put that in the bug which didn't match what I wanted to get fixed.

drewnoakes: > I'll pull down the dev branch now that this has merged to see whether there's more we can do here.

I can't see an obvious way forward here. It doesn't seem possible to predict the collection size in advance.

davkean: Yeah I think I made a similar conclusion as I didn't file a bug here despite the resize being obvious in the stack.

PR #5202, opened 2023-06-05
Avoid allocations while parsing NuGetVersion from strings

### LINKED ISSUE #12630: Parsing NuGetVersion causes significant GC pressure
The code that parses NuGetVersion from strings indirectly uses .NET's Version.TryParse to parse the version portion of the string. When a typical restore can be thousands upon thousands of NuGetVersion objects floating around, this API on .NET Framework is not good enough as it indirectly uses String.Split under the covers and is very inefficient.

Erarndt:
## Description
The code that parses NuGetVersion from strings [indirectly uses .NET's Version.TryParse](https://github.com/NuGet/NuGet.Client/blob/a6fff16af6f44927a83851448a99beb1d2801600/src/NuGet.Clients/NuGet.VisualStudio/LegacyTypes/SemanticVersion.cs#L179) to parse the version portion of the string. This can be avoided to save lots of allocations in hot paths.

Fixes: NuGet/Home#12630

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
+++ b/src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
@@ -164,8 +164,8 @@ namespace NuGet.Versioning
 
                 long returnValue = 0;
 
-                // Parse the values in reverse and assign a tens place to them.
-                // When parsing "123456", this method adds 6 + 50 + 400 + 3000 + 20000 + 100000
+                // Parse the values digit by digit and multiplies by 10 to make space for the next digit.
+                // When parsing "123456", this method becomes 1 -> 10 + 2 -> 120 + 3 -> 1230 + 4 -> 12340 + 5 -> 123450 + 6 -> 123456
                 for (int i = start; i < end; i++)
                 {
                     // negative numbers are invalid for version strings so we only need to check for digits
--- a/src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
+++ b/src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
@@ -149,10 +149,10 @@ namespace NuGet.Versioning
                     end = s.Length;
                 }
 
-                return TryParseInt(s, start, end - 1, out versionNumber);
+                return TryParseInt(s, start, end, out versionNumber);
             }
 
-            // start and end are inclusive bounds for the section of string to parse
+            // start is inclusive bound and end is exclusive for the section of string to parse
             static bool TryParseInt(string s, int start, int end, out int value)
             {
                 // invalid section specified
@@ -163,11 +163,10 @@ namespace NuGet.Versioning
                 }
 
                 long returnValue = 0;
-                int multiplier = 1;
 
                 // Parse the values in reverse and assign a tens place to them.
                 // When parsing "123456", this method adds 6 + 50 + 400 + 3000 + 20000 + 100000
-                for (int i = end; i >= start; i--)
+                for (int i = start; i < end; i++)
                 {
                     // negative numbers are invalid for version strings so we only need to check for digits
                     char current = s[i];
@@ -179,8 +178,7 @@ namespace NuGet.Versioning
 
                     // subtract off ASCII value of '0' from our current character to get the digit's value
                     // e.g. '3' - '0' == 51 - 58 == 3
-                    returnValue += (current - '0') * multiplier;
-                    multiplier *= 10;
+                    returnValue = returnValue * 10 + (current - '0');
 
                     // Check for overflow. We can't get outside the bounds of long before exceeding int.MaxValue
                     // Intentionally avoid usage of 'checked' statement to avoid exception
--- a/src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
+++ b/src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
@@ -49,13 +49,7 @@ namespace NuGet.Versioning
                 {
                     string versionPart = versionString!;
 
-                    if (IndexOf(versionPart, '.') < 0)
-                    {
-                        // System.Version requires at least a 2 part version to parse.
-                        versionPart += ".0";
-                    }
-
-                    if (Version.TryParse(versionPart, out systemVersion))
+                    if (TryGetNormalizedVersion(versionPart, out systemVersion))
                     {
                         // labels
                         if (releaseLabels != null)
@@ -76,8 +70,6 @@ namespace NuGet.Versioning
                             return false;
                         }
 
-                        var ver = NormalizeVersionValue(systemVersion);
-
                         var originalVersion = value;
 
                         if (IndexOf(originalVersion, ' ') > -1)
@@ -90,7 +82,7 @@ namespace NuGet.Versioning
 #endif
                         }
 
-                        version = new NuGetVersion(version: ver,
+                        version = new NuGetVersion(version: systemVersion,
                             releaseLabels: releaseLabels,
                             metadata: buildMetadata ?? string.Empty,
                             originalVersion: originalVersion);
@@ -109,6 +101,70 @@ namespace NuGet.Versioning
 #else
                 return str.IndexOf(c);
 #endif
+            }
+        }
+
+        private static bool TryGetNormalizedVersion(string str, [NotNullWhen(true)] out Version? version)
+        {
+            if (GetNextSection(str, 0, out int endIndex, out int major))
+            {
+                int build = 0;
+                int revision = 0;
+
+                // check for all the possible parts of the version string. If endIndex is less than the end of
+                // the string, the input string was invalid (e.g. "1.2.3.4.5").
+                bool validString = GetNextSection(str, endIndex + 1, out endIndex, out int minor) &&
+                    GetNextSection(str, endIndex + 1, out endIndex, out build) &&
+                    GetNextSection(str, endIndex + 1, out endIndex, out revision) &&
+                    endIndex == str.Length;
+
+                if (validString)
+                {
+                    version = new Version(major, Math.Max(0, minor), Math.Max(0, build), Math.Max(0, revision));
+                    return true;
+                }
+            }
+
+            version = null;
+
+            return false;
+
+            // returns false if an invalid section was found while processing the string
+            static bool GetNextSection(string s, int start, out int end, out int versionNumber)
+            {
+                // check to see if we've processed the whole string
+                if (start >= s.Length)
+                {
+                    // we've reached the end. The section is empty but not invalid.
+                    end = s.Length;
+                    versionNumber = -1;
+
+                    return true;
+                }
+
+                end = s.IndexOf('.', start);
+                string versionSection;
+
+                if (end == -1)
+                {
+                    end = s.Length;
+                    versionSection = s.Substring(start);
+                }
+                else
+                {
+                    versionSection = s.Substring(start, end - start);
+                }
+
+                bool parseResult = int.TryParse(versionSection, out versionNumber);
+                if (versionNumber < 0)
+                {
+                    // negative numbers are invalid for version strings
+                    versionNumber = 0;
+
+                    return false;
+                }
+
+                return parseResult;
             }
         }
 
--- a/src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
+++ b/src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
@@ -155,8 +155,11 @@ namespace NuGet.Versioning
             // start and end are inclusive bounds for the section of string to parse
             static bool TryParseInt(string s, int start, int end, out int value)
             {
-                value = 0;
+                long returnValue = 0;
                 int multiplier = 1;
+
+                // Parse the values in reverse and assign a tens place to them.
+                // When parsing "123456", this method adds 6 + 50 + 400 + 3000 + 20000 + 100000
                 for (int i = end; i >= start; i--)
                 {
                     // negative numbers are invalid for version strings so we only need to check for digits
@@ -167,9 +170,22 @@ namespace NuGet.Versioning
                         return false;
                     }
 
-                    value += (current - '0') * multiplier;
+                    // subtract off ASCII value of '0' from our current character to get the digit's value
+                    // e.g. '3' - '0' == 51 - 58 == 3
+                    returnValue += (current - '0') * multiplier;
                     multiplier *= 10;
+
+                    // Check for overflow. We can't get outside the bounds of long before exceeding int.MaxValue
+                    // Intentionally avoid usage of 'checked' statement to avoid exception
+                    if (returnValue > int.MaxValue)
+                    {
+                        value = 0;
+                        return false;
+                    }
                 }
+
+                // Previous checks guarantee returnValue <= int.MaxValue
+                value = (int)returnValue;
 
                 return true;
             }
--- a/src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
+++ b/src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
@@ -155,6 +155,13 @@ namespace NuGet.Versioning
             // start and end are inclusive bounds for the section of string to parse
             static bool TryParseInt(string s, int start, int end, out int value)
             {
+                // invalid section specified
+                if (start >= end)
+                {
+                    value = 0;
+                    return false;
+                }
+
                 long returnValue = 0;
                 int multiplier = 1;
 
--- a/src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
+++ b/src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
@@ -104,7 +104,7 @@ namespace NuGet.Versioning
             }
         }
 
-        public static bool TryGetNormalizedVersion(string str, [NotNullWhen(true)] out Version? version)
+        private static bool TryGetNormalizedVersion(string str, [NotNullWhen(true)] out Version? version)
         {
             if (string.IsNullOrWhiteSpace(str))
             {
@@ -183,7 +183,6 @@ namespace NuGet.Versioning
                         intermediateVersionNumber = intermediateVersionNumber * 10 + currentChar - '0';
 
                         // Check for overflow. We can't get outside the bounds of intermediateVersionNumber, a long, before exceeding int.MaxValue
-                        // since we're
                         // Intentionally avoid usage of 'checked' statement to avoid exception
                         if (intermediateVersionNumber > int.MaxValue)
                         {
--- a/src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
+++ b/src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
@@ -106,7 +106,7 @@ namespace NuGet.Versioning
 
         private static bool TryGetNormalizedVersion(string str, [NotNullWhen(true)] out Version? version)
         {
-            if (GetNextSection(str, 0, out int endIndex, out int major))
+            if (!string.IsNullOrEmpty(str) && GetNextSection(str, 0, out int endIndex, out int major))
             {
                 int build = 0;
                 int revision = 0;
--- a/src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
+++ b/src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
@@ -143,28 +143,35 @@ namespace NuGet.Versioning
                 }
 
                 end = s.IndexOf('.', start);
-                string versionSection;
 
                 if (end == -1)
                 {
                     end = s.Length;
-                    versionSection = s.Substring(start);
                 }
-                else
+
+                return TryParseInt(s, start, end - 1, out versionNumber);
+            }
+
+            // start and end are inclusive bounds for the section of string to parse
+            static bool TryParseInt(string s, int start, int end, out int value)
+            {
+                value = 0;
+                int multiplier = 1;
+                for (int i = end; i >= start; i--)
                 {
-                    versionSection = s.Substring(start, end - start);
+                    // negative numbers are invalid for version strings so we only need to check for digits
+                    char current = s[i];
+                    if (current < '0' || current > '9')
+                    {
+                        value = 0;
+                        return false;
+                    }
+
+                    value += (current - '0') * multiplier;
+                    multiplier *= 10;
                 }
 
-                bool parseResult = int.TryParse(versionSection, out versionNumber);
-                if (versionNumber < 0)
-                {
-                    // negative numbers are invalid for version strings
-                    versionNumber = 0;
-
-                    return false;
-                }
-
-                return parseResult;
+                return true;
             }
         }
 
--- a/src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
+++ b/src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
@@ -104,95 +104,166 @@ namespace NuGet.Versioning
             }
         }
 
-        private static bool TryGetNormalizedVersion(string str, [NotNullWhen(true)] out Version? version)
+        public static bool TryGetNormalizedVersion(string str, [NotNullWhen(true)] out Version? version)
         {
-            if (!string.IsNullOrEmpty(str) && GetNextSection(str, 0, out int endIndex, out int major))
+            if (string.IsNullOrWhiteSpace(str))
             {
-                int build = 0;
-                int revision = 0;
-
-                // check for all the possible parts of the version string. If endIndex is less than the end of
-                // the string, the input string was invalid (e.g. "1.2.3.4.5").
-                bool validString = GetNextSection(str, endIndex + 1, out endIndex, out int minor) &&
-                    GetNextSection(str, endIndex + 1, out endIndex, out build) &&
-                    GetNextSection(str, endIndex + 1, out endIndex, out revision) &&
-                    endIndex == str.Length;
-
-                if (validString)
-                {
-                    version = new Version(major, Math.Max(0, minor), Math.Max(0, build), Math.Max(0, revision));
-                    return true;
-                }
+                version = null;
+                return false;
             }
 
-            version = null;
+            int minor = 0;
+            int build = 0;
+            int revision = 0;
 
-            return false;
+            // Check for all the possible parts of the version string. If lastParsedPosition is less than the end of
+            // the string, the input string was invalid (e.g. "1.2.3.4.5").
+            bool success = ParseSection(str, 0, out int lastParsedPosition, out int major) &&
+                ParseSection(str, lastParsedPosition, out lastParsedPosition, out minor) &&
+                ParseSection(str, lastParsedPosition, out lastParsedPosition, out build) &&
+                ParseSection(str, lastParsedPosition, out lastParsedPosition, out revision) &&
+                lastParsedPosition == str.Length;
 
-            // returns false if an invalid section was found while processing the string
-            static bool GetNextSection(string s, int start, out int end, out int versionNumber)
+            if (success)
             {
-                // check to see if we've processed the whole string
-                if (start >= s.Length)
-                {
-                    // we've reached the end. The section is empty but not invalid.
-                    end = s.Length;
-                    versionNumber = -1;
+                version = new Version(major, minor, build, revision);
+                return true;
+            }
+            else
+            {
+                version = null;
+                return false;
+            }
 
+            // Returns false if an invalid section was found while processing the string.
+            static bool ParseSection(string str, int start, out int end, out int versionNumber)
+            {
+                // Section is empty.
+                if (start == str.Length)
+                {
+                    end = start;
+                    versionNumber = 0;
                     return true;
                 }
 
-                end = s.IndexOf('.', start);
-
-                if (end == -1)
+                // Trim off leading whitespace
+                for (end = start; end < str.Length; ++end)
                 {
-                    end = s.Length;
+                    char currentChar = str[end];
+                    if (!char.IsWhiteSpace(currentChar))
+                    {
+                        if (IsDigit(currentChar))
+                        {
+                            break;
+                        }
+                        else
+                        {
+                            // Found a non-whitespace non-digit character. Invalid string.
+                            versionNumber = 0;
+                            return false;
+                        }
+                    }
                 }
 
-                return TryParseInt(s, start, end, out versionNumber);
-            }
-
-            // start is inclusive bound and end is exclusive for the section of string to parse
-            static bool TryParseInt(string s, int start, int end, out int value)
-            {
-                // invalid section specified
-                if (start >= end)
+                bool done = false;
+                bool digitFound = false;
+                long intermediateVersionNumber = 0;
+                // Handle number portion.
+                for (; end < str.Length; ++end)
                 {
-                    value = 0;
+                    // Negative numbers are invalid for version strings so we only need to check for digits.
+                    char currentChar = str[end];
+                    if (IsDigit(currentChar))
+                    {
+                        // Parse the values digit by digit and multiplies by 10 to make space for the next digit.
+                        // When parsing "123456", this method becomes 1 -> 10 + 2 -> 120 + 3 -> 1230 + 4 -> 12340 + 5 -> 123450 + 6 -> 123456
+                        // We subtract off ASCII value of '0' from our current character to get the digit's value
+                        // e.g. '3' - '0' == 51 - 48 == 3
+                        digitFound = true;
+                        intermediateVersionNumber = intermediateVersionNumber * 10 + currentChar - '0';
+
+                        // Check for overflow. We can't get outside the bounds of intermediateVersionNumber, a long, before exceeding int.MaxValue
+                        // since we're
+                        // Intentionally avoid usage of 'checked' statement to avoid exception
+                        if (intermediateVersionNumber > int.MaxValue)
+                        {
+                            versionNumber = 0;
+                            return false;
+                        }
+                    }
+                    else if (currentChar == '.')
+                    {
+                        ++end;
+                        // version string ended with '.'
+                        if (end == str.Length)
+                        {
+                            versionNumber = 0;
+                            return false;
+                        }
+
+                        done = true;
+                        break;
+                    }
+                    else if (char.IsWhiteSpace(currentChar))
+                    {
+                        break;
+                    }
+                    else
+                    {
+                        versionNumber = 0;
+                        return false;
+                    }
+                }
+
+                // We failed to find a number in the section, so the string is invalid.
+                if (!digitFound)
+                {
+                    versionNumber = 0;
                     return false;
                 }
 
-                long returnValue = 0;
-
-                // Parse the values digit by digit and multiplies by 10 to make space for the next digit.
-                // When parsing "123456", this method becomes 1 -> 10 + 2 -> 120 + 3 -> 1230 + 4 -> 12340 + 5 -> 123450 + 6 -> 123456
-                for (int i = start; i < end; i++)
+                if (end == str.Length)
                 {
-                    // negative numbers are invalid for version strings so we only need to check for digits
-                    char current = s[i];
-                    if (current < '0' || current > '9')
-                    {
-                        value = 0;
-                        return false;
-                    }
+                    done = true;
+                }
 
-                    // subtract off ASCII value of '0' from our current character to get the digit's value
-                    // e.g. '3' - '0' == 51 - 58 == 3
-                    returnValue = returnValue * 10 + (current - '0');
-
-                    // Check for overflow. We can't get outside the bounds of long before exceeding int.MaxValue
-                    // Intentionally avoid usage of 'checked' statement to avoid exception
-                    if (returnValue > int.MaxValue)
+                if (!done)
+                {
+                    // trailing whitespace
+                    for (; end < str.Length; ++end)
                     {
-                        value = 0;
-                        return false;
+                        char currentChar = str[end];
+                        if (!char.IsWhiteSpace(currentChar))
+                        {
+                            if (currentChar == '.')
+                            {
+                                ++end;
+                                // version string ended with '.'
+                                if (end == str.Length)
+                                {
+                                    versionNumber = 0;
+                                    return false;
+                                }
+
+                                break;
+                            }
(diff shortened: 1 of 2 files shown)


zivkan:
@Erarndt the linked issue you skipped in the checklist is used by our automated release notes tool, and we've got feedback from customers in the past that they like to know about perf improvements. So, I think having an issue in NuGet/Home linked from this PR will be appreciated by customers who learn that this perf improvement is benefiting them. Especially since NuGet does a LOT (really a LOT) of version string parsing in a typical restore.

And just general conversation, one day it'll be nice to get rid of the substrings, and use `span<char>`, but there's a bunch of extra work required there to get NuGet's ilmerged assemblies to work, and possibly some manual testing to make sure that msbuild restore still works correctly when NuGet is inserted into VS.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
```diff
@@ -112,6 +104,77 @@ int IndexOf(string str, char c)
...
+                if (end == -1)
+                {
+                    end = s.Length;
+                }
+
+                return TryParseInt(s, start, end - 1, out versionNumber);
+            }
+
+            // start and end are inclusive bounds for the section of string to parse
+            static bool TryParseInt(string s, int start, int end, out int value)
```
jeffkl: Should this method guard against an overflow? I highly doubt anyone would but technically someone could specify a version of `1.0.0.123456789123456789`.  I guess if the length is more than an `int` can be, just return false?

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
```diff
@@ -112,6 +104,77 @@ int IndexOf(string str, char c)
...
+                {
+                    // negative numbers are invalid for version strings so we only need to check for digits
+                    char current = s[i];
+                    if (current < '0' || current > '9')
+                    {
+                        value = 0;
+                        return false;
+                    }
+
+                    value += (current - '0') * multiplier;
```
jeffkl: I had to debug this method to fully understand it.  I like how it works but I think a few comments would help.

* It converts the char value to an integer by subtracting the char value of '0'.  So `'3' - '0'` is `51 - 48 = 3`
* It parses values in reverse and assigns a tens place to them. When parsing `"123456"`, this method adds `6 + 50 + 400 + 3000 + 20000 + 100000`

Pretty neat!

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
```diff
@@ -112,6 +104,77 @@ int IndexOf(string str, char c)
...
+            // start and end are inclusive bounds for the section of string to parse
+            static bool TryParseInt(string s, int start, int end, out int value)
+            {
+                value = 0;
+                int multiplier = 1;
+                for (int i = end; i >= start; i--)
+                {
+                    // negative numbers are invalid for version strings so we only need to check for digits
+                    char current = s[i];
+                    if (current < '0' || current > '9')
```
jeffkl: Would it be okay to call `char.IsDigit()` here instead?  It doesn't appear to have much overhead but not sure if you intentionally avoided calling it?

https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Char.cs,318

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
```diff
@@ -112,6 +104,77 @@ int IndexOf(string str, char c)
...
+                // check for all the possible parts of the version string. If endIndex is less than the end of
+                // the string, the input string was invalid (e.g. "1.2.3.4.5").
+                bool validString = GetNextSection(str, endIndex + 1, out endIndex, out int minor) &&
+                    GetNextSection(str, endIndex + 1, out endIndex, out build) &&
+                    GetNextSection(str, endIndex + 1, out endIndex, out revision) &&
+                    endIndex == str.Length;
+
+                if (validString)
+                {
+                    version = new Version(major, Math.Max(0, minor), Math.Max(0, build), Math.Max(0, revision));
```
jeffkl: I'm unsure if this matters, but our current code calls `Version.TryParse()` which returns a `Version` with only the appropriate version components set.  So `"1.0"` would return a version with `-1` as the build and revision.  This code now always calls the `Version` constructor with values of 0.  

To keep the current behavior, it would only need to specify `0` for the minor if its not specified but pass `-1` for the build and revision.

Erarndt: Could be wrong, but it looked like the call to `NormalizeVersionValue` ensures that build and revision are set to a minimum of zero. 


        internal static Version NormalizeVersionValue(Version version)
        {
            var normalized = version;

            if (version.Build < 0
                || version.Revision < 0)
            {
                normalized = new Version(
                    version.Major,
                    version.Minor,
                    Math.Max(version.Build, 0),
                    Math.Max(version.Revision, 0));
            }

            return normalized;
        }


`NormalizeVersionValue` allocates a new `Version` instance to do this that I was hoping to avoid.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
zivkan: I'd be concerned about unicode characters that return true for `char.IsDigit`, but `(c - '0')` will result in values outside of `0` to `9`.

Having written that, I'm now wondering if that means `Version.Parse` supports other unicode digits, and therefore if we have version strings that are currently parsable that will fail with this change. (though technically I believe those strings would be a violation of the SemVer spec. From memory it's specific about using ASCII '0' to '9'). But, unless the code to convert a single char into a number supports "all" unicode characters, I don't think it's safe to use `char.IsDigit`.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
Erarndt: Added checks for overflow.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
Erarndt: Updated with comments. Let me know if you think I should clarify anything.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
jeffkl: Ah sorry I didn't realize that!

jeffkl (commented):
There is a unit tests failure because the new code does not throw for a version like `1.2..`

```
NuGet.Versioning.Test.NuGetVersionTest.ParseThrowsIfStringIsNotAValidSemVer(versionString: "1.2..")
   Source: NuGetVersionTest.cs line 106
   Duration: 2 ms

  Message: 
    Assert.NotNull() Failure

  Stack Trace: 
    ExceptionAssert.Throws[TException](Action act, Action`1 condition) line 19
    ExceptionAssert.ThrowsArgumentException[TArgException](Action act, String paramName, String message) line 77
    ExceptionAssert.ThrowsArgumentException(Action act, String paramName, String message) line 72
    NuGetVersionTest.ParseThrowsIfStringIsNotAValidSemVer(String versionString) line 108
```

Erarndt:
> 



> There is a unit tests failure because the new code does not throw for a version like `1.2..`
> 
> ```
> NuGet.Versioning.Test.NuGetVersionTest.ParseThrowsIfStringIsNotAValidSemVer(versionString: "1.2..")
>    Source: NuGetVersionTest.cs line 106
>    Duration: 2 ms
> 
>   Message: 
>     Assert.NotNull() Failure
> 
>   Stack Trace: 
>     ExceptionAssert.Throws[TException](Action act, Action`1 condition) line 19
>     ExceptionAssert.ThrowsArgumentException[TArgException](Action act, String paramName, String message) line 77
>     ExceptionAssert.ThrowsArgumentException(Action act, String paramName, String message) line 72
>     NuGetVersionTest.ParseThrowsIfStringIsNotAValidSemVer(String versionString) line 108
> ```

Ah, that caught a bug with `TryParseInt`. It was erroneously returning `true` and `0` when `start >= end`

jeffkl:
@Erarndt I haven't thanked you yet for this contribution, so thank you very much for this!

I've been playing with the method for a bit today because you got my brain flowing, what do you think of the following implementation?  It does the following:

* Loop through each character sequentially (because it multiplies the current value by 10 to make room for the new ones place
* If the current character is not a digit or period, fails fast
* Allocates an `int[]` once the first character is found
* Handles overflows with `checked` (the very rare case so shouldn't add overhead in the success case)
* Handles when someone specifies `1..2` or `..1`

```c#
using System.Diagnostics.CodeAnalysis;
using Xunit;

namespace IntParseTest
{
    public class UnitTest1
    {
        [Theory]
        [InlineData("1", true, 1)]
        [InlineData("1.12", true, 1, 12)]
        [InlineData("1.12.123", true, 1, 12, 123)]
        [InlineData("1.12.123.1234", true, 1, 12, 123, 1234)]
        [InlineData("2147483647.2147483647.2147483647.2147483647", true, 2147483647, 2147483647, 2147483647, 2147483647)]
        [InlineData("2147483648.2.3.4", false)]
        [InlineData("1.2147483648.3.4", false)]
        [InlineData("1.2.2147483648.4", false)]
        [InlineData("1.2.3.2147483648", false)]
        [InlineData("1..2", false)]
        [InlineData("1.2..", false)]
        [InlineData("..1.2", false)]
        [InlineData("1. 2", false)]
        [InlineData("-1.2.3.4", false)]
        [InlineData("1.-2.3.4", false)]
        [InlineData("1.2.-3.4", false)]
        [InlineData("1.2.3.-4", false)]
        public void Test4(string versionString, bool expectedResult, int expectedMajor = 0, int expectedMinor = 0, int expectedBuild = 0, int expectedRevision = 0)
        {
            bool actualResult = TryParseVersion(versionString, out Version? version);

            Assert.Equal(expectedResult, actualResult);

            if (expectedResult)
            {
                Assert.Equal(expectedMajor, version!.Major);
                Assert.Equal(expectedMinor, version!.Minor);
                Assert.Equal(expectedBuild, version!.Build);
                Assert.Equal(expectedRevision, version!.Revision);
            }
            else
            {
                Assert.Null(version);
            }
        }

        private static bool TryParseVersion(string versionString, [NotNullWhen(true)] out Version? version)
        {
            version = null;

            // Used to store 4 integers representing the major, minor, build, and revision components of the version.
            int[]? components = null;

            // The current component being parsed
            int componentIndex = 0;

            for (int i = 0; i < versionString.Length; i++)
            {
                char currentChar = versionString[i];

                if (currentChar == '.')
                {
                    // Move to the next version component
                    componentIndex++;

                    // No digits have been found for the first component or the previous component did not contain any digits (like .1 or 1..2)
                    if (components == null || components[componentIndex - 1] == -1)
                    {
                        return false;
                    }

                    // Move to the next character
                    continue;
                }

                // Return false if any character is not a digit
                if (currentChar < '0' || currentChar > '9')
                {
                    return false;
                }

                if (components == null)
                {
                    // Initialize the component array for the first time when the first digit is encountered.
                    // Each value starts at -1 to indicate whether or not any digits have been encountered for that particular component.
                    components = new int[4] { 0, -1, -1, -1 };
                }

                try
                {
                    // Set the value to zero for the current component if this is the first digit encountered for that component
                    if (components[componentIndex] == -1)
                    {
                        components[componentIndex] = 0;
                    }

                    // Multiply the current value by 10 so the current digit can be added to the ones place
                    // Subtracting '0' from the current character converts it to the integer value it represents
                    // The checked statement tells the runtime to throw an OverflowException if the arithmetic operation overflows the integer value
                    components[componentIndex] = checked(components[componentIndex] * 10 + (currentChar - '0'));
                }
                catch (OverflowException)
                {
                    return false;
                }
            }

            // Make uninitialized components zero to normalize the version
            version = new Version(components![0], Math.Max(0, components[1]), Math.Max(0, components[2]), Math.Max(0, components[3]));

            return true;
        }
    }
}
```

Erarndt:
> > 
> 
> > There is a unit tests failure because the new code does not throw for a version like `1.2..`
> > ```
> > NuGet.Versioning.Test.NuGetVersionTest.ParseThrowsIfStringIsNotAValidSemVer(versionString: "1.2..")
> >    Source: NuGetVersionTest.cs line 106
> >    Duration: 2 ms
> > 
> >   Message: 
> >     Assert.NotNull() Failure
> > 
> >   Stack Trace: 
> >     ExceptionAssert.Throws[TException](Action act, Action`1 condition) line 19
> >     ExceptionAssert.ThrowsArgumentException[TArgException](Action act, String paramName, String message) line 77
> >     ExceptionAssert.ThrowsArgumentException(Action act, String paramName, String message) line 72
> >     NuGetVersionTest.ParseThrowsIfStringIsNotAValidSemVer(String versionString) line 108
> > ```
> 
> Ah, that caught a bug with `TryParseInt`. It was erroneously returning `true` and `0` when `start >= end`



> @Erarndt I haven't thanked you yet for this contribution, so thank you very much for this!
> 
> I've been playing with the method for a bit today because you got my brain flowing, what do you think of the following implementation? It does the following:
> 
> * Loop through each character sequentially (because it multiplies the current value by 10 to make room for the new ones place
> * If the current character is not a digit or period, fails fast
> * Allocates an `int[]` once the first character is found
> * Handles overflows with `checked` (the very rare case so shouldn't add overhead in the success case)
> * Handles when someone specifies `1..2` or `..1`
> 
> ```cs
> using System.Diagnostics.CodeAnalysis;
> using Xunit;
> 
> namespace IntParseTest
> {
>     public class UnitTest1
>     {
>         [Theory]
>         [InlineData("1", true, 1)]
>         [InlineData("1.12", true, 1, 12)]
>         [InlineData("1.12.123", true, 1, 12, 123)]
>         [InlineData("1.12.123.1234", true, 1, 12, 123, 1234)]
>         [InlineData("2147483647.2147483647.2147483647.2147483647", true, 2147483647, 2147483647, 2147483647, 2147483647)]
>         [InlineData("2147483648.2.3.4", false)]
>         [InlineData("1.2147483648.3.4", false)]
>         [InlineData("1.2.2147483648.4", false)]
>         [InlineData("1.2.3.2147483648", false)]
>         [InlineData("1..2", false)]
>         [InlineData("1.2..", false)]
>         [InlineData("..1.2", false)]
>         [InlineData("1. 2", false)]
>         [InlineData("-1.2.3.4", false)]
>         [InlineData("1.-2.3.4", false)]
>         [InlineData("1.2.-3.4", false)]
>         [InlineData("1.2.3.-4", false)]
>         public void Test4(string versionString, bool expectedResult, int expectedMajor = 0, int expectedMinor = 0, int expectedBuild = 0, int expectedRevision = 0)
>         {
>             bool actualResult = TryParseVersion(versionString, out Version? version);
> 
>             Assert.Equal(expectedResult, actualResult);
> 
>             if (expectedResult)
>             {
>                 Assert.Equal(expectedMajor, version!.Major);
>                 Assert.Equal(expectedMinor, version!.Minor);
>                 Assert.Equal(expectedBuild, version!.Build);
>                 Assert.Equal(expectedRevision, version!.Revision);
>             }
>             else
>             {
>                 Assert.Null(version);
>             }
>         }
> 
>         private static bool TryParseVersion(string versionString, [NotNullWhen(true)] out Version? version)
>         {
>             version = null;
> 
>             // Used to store 4 integers representing the major, minor, build, and revision components of the version.
>             int[]? components = null;
> 
>             // The current component being parsed
>             int componentIndex = 0;
> 
>             for (int i = 0; i < versionString.Length; i++)
>             {
>                 char currentChar = versionString[i];
> 
>                 if (currentChar == '.')
>                 {
>                     // Move to the next version component
>                     componentIndex++;
> 
>                     // No digits have been found for the first component or the previous component did not contain any digits (like .1 or 1..2)
>                     if (components == null || components[componentIndex - 1] == -1)
>                     {
>                         return false;
>                     }
> 
>                     // Move to the next character
>                     continue;
>                 }
> 
>                 // Return false if any character is not a digit
>                 if (currentChar < '0' || currentChar > '9')
>                 {
>                     return false;
>                 }
> 
>                 if (components == null)
>                 {
>                     // Initialize the component array for the first time when the first digit is encountered.
>                     // Each value starts at -1 to indicate whether or not any digits have been encountered for that particular component.
>                     components = new int[4] { 0, -1, -1, -1 };
>                 }
> 
>                 try
>                 {
>                     // Set the value to zero for the current component if this is the first digit encountered for that component
>                     if (components[componentIndex] == -1)
>                     {
>                         components[componentIndex] = 0;
>                     }
> 
>                     // Multiply the current value by 10 so the current digit can be added to the ones place
>                     // Subtracting '0' from the current character converts it to the integer value it represents
>                     // The checked statement tells the runtime to throw an OverflowException if the arithmetic operation overflows the integer value
>                     components[componentIndex] = checked(components[componentIndex] * 10 + (currentChar - '0'));
>                 }
>                 catch (OverflowException)
>                 {
>                     return false;
>                 }
>             }
> 
>             // Make uninitialized components zero to normalize the version
>             version = new Version(components![0], Math.Max(0, components[1]), Math.Max(0, components[2]), Math.Max(0, components[3]));
> 
>             return true;
>         }
>     }
> }
> ```

@jeffkl Happy to contribute! There's a lot to like about your implementation. It's cleaner, you process each char once, and your approach when accumulating the digit is clever (I'm totally stealing that). I've been trying to find a way to avoid the `int[]` allocation and I can't think of anything good unless we want to utilize an array pool or `stackalloc`. I think `stackalloc` would only make sense if we can use `Span<int>`, so that probably means pulling in the System.Memory package. Not sure if that's the route we'd want to take. Since allocations have been an issue in this path, my preference would be to eliminate as many as possible.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
```diff
@@ -112,6 +104,98 @@ int IndexOf(string str, char c)
...
+                {
+                    value = 0;
+                    return false;
+                }
+
+                long returnValue = 0;
+
+                // Parse the values in reverse and assign a tens place to them.
+                // When parsing "123456", this method adds 6 + 50 + 400 + 3000 + 20000 + 100000
+                for (int i = start; i < end; i++)
```
zivkan: It's no longer parsing in reverse character order.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
```diff
@@ -112,6 +104,98 @@ int IndexOf(string str, char c)
...
+                    // e.g. '3' - '0' == 51 - 58 == 3
+                    returnValue = returnValue * 10 + (current - '0');
+
+                    // Check for overflow. We can't get outside the bounds of long before exceeding int.MaxValue
+                    // Intentionally avoid usage of 'checked' statement to avoid exception
+                    if (returnValue > int.MaxValue)
+                    {
+                        value = 0;
+                        return false;
+                    }
```
zivkan: When I try something equivalent to this method with input `10000000000000000000`, `returnValue` has value `-8446744073709551616`, which is `< int.MaxValue`. It's also `< int.MinValue`, so the cast will throw on line 193. However, I'm concerned that there might be very special strings that parse into `> 0 && < int.MaxValue`, so I don't think that checking `>= 0` is the correct solution.

Adding tests for these edge cases, since we're no longer relying on System.Version's tests catching these edge cases, are important.

Anyway, I think the perf focus is overenthusiastic here, referring to avoiding `checked` and exceptions. An overflow is truly an exception circumstance. If a customer has a number too large when referencing a package, restore is going to fail, and the customer is going to have to fix their reference in order for restore to have the possibility to succeed. Similarly, someone trying to create a package will fail, and they'll have to change the number to create a package. A package source that returns a package with an unparsable version string will break all customers using it with the official NuGet client.

So, there's no scenario I can think of where an overflow can happen, resulting in `return false` in this method, but the user action still succeeds. Therefore I don't expect this to happen frequently since customers will fix the issue before merging anything into their main branch, and there being a high quantity of solution opens or CI builds where overflows happen during version parsing.

I think that using `checked` to detect overflows is not a perf concern, and significantly simplifies the logic to detect overflows.

zivkan (changes requested):
Since this PR no longer takes advantage of System.Version.Parse to catch input validation edge cases for us, it's important to ensure there are unit tests for as many of these edge cases as you can think of, such as `1..`, `1.2.3.4.5`, `-1.0`, `1.-1`, versions that overflow. I can't think of any more edge cases right now, but if you can, please make sure there are tests for them.

jeffkl:
> @jeffkl Happy to contribute! There's a lot to like about your implementation. It's cleaner, you process each char once, and your approach when accumulating the digit is clever (I'm totally stealing that). I've been trying to find a way to avoid the `int[]` allocation and I can't think of anything good unless we want to utilize an array pool or `stackalloc`. I think `stackalloc` would only make sense if we can use `Span<int>`, so that probably means pulling in the System.Memory package. Not sure if that's the route we'd want to take. Since allocations have been an issue in this path, my preference would be to eliminate as many as possible.

I tried another variation of my code where I use 4 integers and map them to the current component.  The code is pretty readable and it does not allocate but its much slower? Oh well...


|                  Method |      Mean |    Error |   StdDev |   Gen0 | Allocated |
|------------------------ |----------:|---------:|---------:|-------:|----------:|
|        ParseVersion4Ints | 124.02 ns | 2.718 ns | 7.887 ns | 0.0076 |      32 B |
|       ParseVersionArray |  55.45 ns | 0.981 ns | 0.870 ns | 0.0172 |      72 B |
| ParseVersionPullRequest |  57.82 ns | 0.990 ns | 1.355 ns | 0.0076 |      32 B |
|     ParseVersionCurrent | 101.54 ns | 1.681 ns | 2.001 ns | 0.0153 |      64 B |

So it would appear your implementation is the fastest and has the least allocations (the 32 bytes are the `Version` itself).

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
```diff
@@ -112,6 +104,98 @@ int IndexOf(string str, char c)
...
+
+                long returnValue = 0;
+
+                // Parse the values in reverse and assign a tens place to them.
+                // When parsing "123456", this method adds 6 + 50 + 400 + 3000 + 20000 + 100000
+                for (int i = start; i < end; i++)
+                {
+                    // negative numbers are invalid for version strings so we only need to check for digits
+                    char current = s[i];
+                    if (current < '0' || current > '9')
```
jeffkl: One more edge case here.  For values with spaces in them, `Version.TryParse()` appears to handle this gracefully.  This new code does not.

For example, if my version is `1. 2`, the new code will fail to parse because of the space.  (https://dotnetfiddle.net/i7hx5x)

It appears to only ignore leading and trailing whitespace though.  I'd hate to call `.Trim()` but do we want to have things work if the user specifies:

```xml
<PackageReference Include="PackageA" Version="1. 23 . 4" />
```

Note, our code works with that crazy version today...

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
jeffkl: > Adding tests for these edge cases, since we're no longer relying on System.Version's tests catching these edge cases, are important.

I'm growing more concerned about this as well now that I've discovered that `1. 23 . 4` can be parsed by our code today and will throw with this change.  I'll try to look at Version.TryParse() to see if we can determine all of the edge cases it was handling for us.

Erarndt: @jeffkl and @zivkan, I really want to thank you both for spending time helping me ensure that this code is up to snuff 😄. I took a step back and rewrote my approach to address some of the issues we found (incorporating some things I picked up from @jeffkl 's implementation) and added a battery of tests (over 100) to test all the cases that were brought up and any additional ones I can think of.

I admit, I wasn't anticipating that `Version.Parse` and `Version.TryParse` handled so many different cases when I started this. The documentation for `Version.Pars` is pretty light on details 😄 

"The input parameter must have the following format:

major.minor[.build[.revision]]

where major, minor, build, and revision are the string representations of the version number's four components: major version number, minor version number, build number, and revision number, respectively. Optional components are shown in square brackets ([ and ]). The components must appear in the specified order and must be separated by periods."

@zivkan, I'm not entirely opposed to using `checked`, but there is a tiny bit of overhead even in the good case (super small). My bigger concern is around the error case since exceptions are expensive and methods with the BCL `Try*` methods don't throw at all. Unless there's a good reason to, I think we should follow that pattern.

I'd be curious to see what test code you used with `10000000000000000000`. The calculations are done using `long`, so seeing any overflow seem impossible to me since the code terminates as soon as we exceed `int.MaxValue`.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
```diff
@@ -112,6 +104,169 @@ int IndexOf(string str, char c)
             }
         }
 
+        public static bool TryGetNormalizedVersion(string str, [NotNullWhen(true)] out Version? version)
```
zivkan: ~Make this `internal`. This is not an API I want external customers to have available. Our tests have `InternalsVisibleTo`.~

After reviewing the tests and reconsidering, this method should be an internal implementation detail that is not exposed either publicly, or to tests. Our contract is `NuGetVersion.Parse`, and equivalent, which is what I think should be tested.

test/NuGet.Core.Tests/NuGet.Versioning.Test/NuGetVersionTest.cs
```diff
@@ -363,5 +363,131 @@ public void TryParseReturnsFalseWhenUnableToParseString(string versionString)
...
+        [InlineData("   19   .   19   .   19   .   19   ", 19, 19, 19, 19)]
+        [InlineData("1. 1", 1, 1, 0, 0)]
+        [InlineData("1 .1", 1, 1, 0, 0)]
+        [InlineData("1 . 1 .0", 1, 1, 0, 0)]
+        [InlineData("1. 23 . 4", 1, 23, 4, 0)]
+        [InlineData("01. 23 . 4", 1, 23, 4, 0)]
+        [InlineData("01.1.1.1", 1, 1, 1, 1)]
+        [InlineData("1.01.1.1", 1, 1, 1, 1)]
+        [InlineData("1.1.01.1", 1, 1, 1, 1)]
+        [InlineData("1.1.01.01", 1, 1, 1, 1)]
```
zivkan: The semver spec says leading zeros are not allowed. Does `NuGetVersion` already accept it?  I think I remember a customer complaining that it's not allowed, although that might have been in the prerelease labels, not the major/minor/patch components.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
```diff
@@ -112,6 +104,169 @@ int IndexOf(string str, char c)
...
+                        // Parse the values digit by digit and multiplies by 10 to make space for the next digit.
+                        // When parsing "123456", this method becomes 1 -> 10 + 2 -> 120 + 3 -> 1230 + 4 -> 12340 + 5 -> 123450 + 6 -> 123456
+                        // We subtract off ASCII value of '0' from our current character to get the digit's value
+                        // e.g. '3' - '0' == 51 - 48 == 3
+                        digitFound = true;
+                        intermediateVersionNumber = intermediateVersionNumber * 10 + currentChar - '0';
+
+                        // Check for overflow. We can't get outside the bounds of intermediateVersionNumber, a long, before exceeding int.MaxValue
+                        // since we're
+                        // Intentionally avoid usage of 'checked' statement to avoid exception
```
zivkan: The comment is

test/NuGet.Core.Tests/NuGet.Versioning.Test/NuGetVersionTest.cs
```diff
@@ -363,5 +363,131 @@ public void TryParseReturnsFalseWhenUnableToParseString(string versionString)
...
+        [InlineData("19   .1 9   ")]
+        [InlineData("19   .19   .1 9   ")]
+        [InlineData("19   .19   .19   .1 9   ")]
+        [InlineData("   1 9   ")]
+        [InlineData("   19   .   1 9   ")]
+        [InlineData("   19   .   19   .   1 9   ")]
+        [InlineData("   19   .   19   .   19   .   1 9   ")]
+        public void CatchesInvalidVersionStrings(string versionString)
+        {
+            Assert.False(NuGetVersion.TryGetNormalizedVersion(versionString, out var version));
```
zivkan: I think these tests (and the others below), should use `NuGetVersion.Parse` or `NuGetVersion.TryParse` instead.  `TryGetNorializedVersion` is an internal implementation detail that doesn't really matter that much.

Also, a bunch of the test cases are already tested with `NuGetVersion.Parse`, so some of these test cases are duplicates. Adding the new test cases to existing tests will have the same test coverage with fewer tests overall.

test/NuGet.Core.Tests/NuGet.Versioning.Test/NuGetVersionTest.cs
Erarndt: `Version.Parse` accepts them, so I wanted to make sure the new approach did as well. A string like `Version.Parse("0001.  0022 . 03 ")` would drop all the leading zeros and handle the spaces (`1.22.3`)

kartheekp-ms:
@Erarndt. Thank you for the contribution. In the linked issue you have mentioned the following which makes me think that we are not disposing objects properly. As far as I know, these objects shouldn't be lingering once the NuGet operation, like "restore," is finished. This makes me believe that the problem might be more extensive. Personally, I'd rather use the .NET Framework API because it improves code clarity and reduces maintenance efforts for our NuGet team. However, I don't want to hold up the PR.

> When a typical restore can be thousands upon thousands of NuGetVersion objects floating around, this API on .NET Framework is not good enough as it indirectly uses String.Split under the covers and is very inefficient.

Erarndt:
> @Erarndt. Thank you for the contribution. In the linked issue you have mentioned the following which makes me think that we are not disposing objects properly. As far as I know, these objects shouldn't be lingering once the NuGet operation, like "restore," is finished. This makes me believe that the problem might be more extensive. Personally, I'd rather use the .NET Framework API because it improves code clarity and reduces maintenance efforts for our NuGet team. However, I don't want to hold up the PR.
> 
> > When a typical restore can be thousands upon thousands of NuGetVersion objects floating around, this API on .NET Framework is not good enough as it indirectly uses String.Split under the covers and is very inefficient.

HI @kartheekp-ms !

This isn't an issue of calling `Dispose()` or similar. In the execution of the `NugetVersion.TryParse`, there were several smaller allocations that became problematic due to how frequently this is called. `System.Version.TryParse` calls `string.Split` internally which allocates a `char[]` and an `int[]` the size of the original string, a new `string` for each substring, and, a `string[]` to return the items. In addition, `NugetVersion.TryParse` can allocate a new `string` here `versionPart += ".0"`, and another `System.Version` object in `NormalizeVersionValue`. All of these allocations produce measurable GC pressure and can be avoided.

test/NuGet.Core.Tests/NuGet.Versioning.Test/NuGetVersionTest.cs
nkolev92: I'm kind of tempted to have a *separate* PR that we merge in quickly(override checks once the unit tests pass) that just adds the tests to promote confidence that there are no functional changes in this refactoring.

test/NuGet.Core.Tests/NuGet.Versioning.Test/NuGetVersionTest.cs
```diff
@@ -363,5 +417,36 @@ public void TryParseReturnsFalseWhenUnableToParseString(string versionString)
             Assert.False(result);
```
kartheekp-ms: ~If my understanding is correct, the current implementation returns false when version number is "19". However, the with the proposed implementation it looks like it returns true. please verify.~

~The comment https://github.com/NuGet/NuGet.Client/pull/5202/files#diff-7e0ebda87c9c1a63b04e7f8650af6c88aeffbf8cc9a49684a602390cfb57132fL54 also suggests that `System.Version requires at least a 2 part version to parse.`~

Please ignore this comment, I think we are good in this case because ParseSection() returns 0 if there are no characters to read.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
```diff
@@ -112,6 +104,168 @@ int IndexOf(string str, char c)
...
+                    return true;
+                }
+
+                // Trim off leading whitespace
+                for (end = start; end < str.Length; ++end)
+                {
+                    char currentChar = str[end];
+                    if (!char.IsWhiteSpace(currentChar))
+                    {
+                        if (IsDigit(currentChar))
```
kartheekp-ms, line 155: How about using [`Char.IsDigit()`](https://learn.microsoft.com/en-us/dotnet/api/system.char.isdigit?view=net-7.0) API instead of a custom implementation?

davkean:
> > @Erarndt. Thank you for the contribution. In the linked issue you have mentioned the following which makes me think that we are not disposing objects properly. As far as I know, these objects shouldn't be lingering once the NuGet operation, like "restore," is finished. This makes me believe that the problem might be more extensive. Personally, I'd rather use the .NET Framework API because it improves code clarity and reduces maintenance efforts for our NuGet team. However, I don't want to hold up the PR.
> > > When a typical restore can be thousands upon thousands of NuGetVersion objects floating around, this API on .NET Framework is not good enough as it indirectly uses String.Split under the covers and is very inefficient.
> 
> HI @kartheekp-ms !
> 
> This isn't an issue of calling `Dispose()` or similar. In the execution of the `NugetVersion.TryParse`, there were several smaller allocations that became problematic due to how frequently this is called. `System.Version.TryParse` calls `string.Split` internally which allocates a `char[]` and an `int[]` the size of the original string, a new `string` for each substring, and, a `string[]` to return the items. In addition, `NugetVersion.TryParse` can allocate a new `string` here `versionPart += ".0"`, and another `System.Version` object in `NormalizeVersionValue`. All of these allocations produce measurable GC pressure and can be avoided.

I wanted to add more to this, while in common case I would agree with your assessment that it is easier to maintain to use .NET's version of `Version.TryParse` (and when .NET Core only, maybe you can do that). However, this is a very hot path in NuGet and found in 7% of high-GC sessions allocating up to 32 MB/s. This needs to be addressed. In the same way C# doesn't use the built-in XML parser to parse XML comments, NuGet needs to step up and take ownership of parsing its own version numbers.

nkolev92:
@Erarndt 

- Can you please create a separate PR with the tests only? Once the tests pass, I can help get that merged in quickly.
- After that's in, we can rebase this PR on top of that and it'll increase the confidence of the reviewers about the risk of the change.

jeffkl:
@Erarndt let me know if you want me to port your test changes and submit a new PR, if you don't have time I'd be happy to do it.

davkean:
@jeffkl Eric is away until next week.

jeffkl:
Okay I've opened a pull request to just add the new test cases and then I'll rebase this pull request once its merged: https://github.com/NuGet/NuGet.Client/pull/5240

jeffkl:
I've merged the change to update the tests so this commit now just migrates away from `Version.Parse()`.

jeffkl (approved):

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
zivkan: From: https://github.com/NuGet/NuGet.Client/pull/5202#discussion_r1218465203 (I don't know how well the link will work, since github historically broke deep links to resolved conversations)

> I'd be concerned about unicode characters that return true for `char.IsDigit`, but `(c - '0')` will result in values outside of `0` to `9`.
>
> Having written that, I'm now wondering if that means `Version.Parse` supports other unicode digits, and therefore if we have version strings that are currently parsable that will fail with this change. (though technically I believe those strings would be a violation of the SemVer spec. From memory it's specific about using ASCII '0' to '9'). But, unless the code to convert a single char into a number supports "all" unicode characters, I don't think it's safe to use `char.IsDigit`.

davkean: IsDigit returns true only for "decimal" digits.

davkean: ~~IsNumeric~~ IsNumber is the one that returns true for all unicode numerics.

zivkan: `char.IsDigit('१')` in C# Interactive returns `true`.

davkean: Indeed, basically this list: https://www.fileformat.info/info/unicode/category/Nd/list.htm. Hmm, I see a few bugs around the tree, assuming this is 0-9. Agreed, probably want to hard code this and add a unit test so it doesn't regress.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
```diff
@@ -112,6 +104,168 @@ int IndexOf(string str, char c)
...
+                return true;
+            }
+            else
+            {
+                version = null;
+                return false;
+            }
+
+            // Returns false if an invalid section was found while processing the string.
+            static bool ParseSection(string str, int start, out int end, out int versionNumber)
```
zivkan, line 139: I don't want to shift goalposts, wasting time & effort, so consider this more like a conversation than a request for change. But I discovered in another PR that System.Memory, and therefore `Span<char>` is available thoughout NuGet (the dll is already in our VS install, for our MSBuild task, and is ilmerged in nuget.exe).

I find this `ParseSection` method a bit long, so I wonder if using `Span<char>` could make the implementation either shorter, or split into smaller methods, making it easier to read & understand. Although, from a perf point of view, I imagine this forward only, read once implementation will be hard to beat.

zivkan (approved):

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
jeffkl: That's an interesting point.  I think I'll merge this for now and maybe later someone can experiment with spans to see if the code can be cleaner and/or more performant.

PR #5205, opened 2023-06-06
Convert LibraryRangeCacheKey to a struct

### LINKED ISSUE #12646: ResolverUtility.FindLibraryCachedAsync should use a struct as lookup
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1824495

drewnoakes:
## Bug

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1824495
Fixes: https://github.com/NuGet/Home/issues/12646

## Description

`LibraryRangeCacheKey ` is showing up in customer traces captured during GC pauses as contributing to many of those pauses. It can comfortably be a struct, avoiding the additional heap allocation.

This is a binary change to public API (unlikely to be a source change).

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.DependencyResolver.Core/LibraryRangeCacheKey.cs
+++ b/src/NuGet.Core/NuGet.DependencyResolver.Core/LibraryRangeCacheKey.cs
@@ -10,9 +10,9 @@ using NuGet.Shared;
 namespace NuGet.DependencyResolver
 {
     /// <summary>
-    /// Helper class to hold a library range and framework.
+    /// Helper type to hold a library range and framework.
     /// </summary>
-    public class LibraryRangeCacheKey : IEquatable<LibraryRangeCacheKey>
+    public readonly struct LibraryRangeCacheKey : IEquatable<LibraryRangeCacheKey>
     {
         public LibraryRangeCacheKey(LibraryRange range, NuGetFramework framework)
         {
@@ -32,7 +32,7 @@ namespace NuGet.DependencyResolver
 
         public override bool Equals(object obj)
         {
-            return Equals(obj as LibraryRangeCacheKey);
+            return obj is LibraryRangeCacheKey key && Equals(key);
         }
 
         public override int GetHashCode()
@@ -42,16 +42,6 @@ namespace NuGet.DependencyResolver
 
         public bool Equals(LibraryRangeCacheKey other)
         {
-            if (Object.ReferenceEquals(this, other))
-            {
-                return true;
-            }
-
-            if (Object.ReferenceEquals(null, other))
-            {
-                return false;
-            }
-
             return LibraryRange.Equals(other.LibraryRange)
                 && Framework.Equals(other.Framework);
         }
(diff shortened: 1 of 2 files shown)


src/NuGet.Core/NuGet.DependencyResolver.Core/LibraryRangeCacheKey.cs
```diff
@@ -10,9 +10,9 @@
 namespace NuGet.DependencyResolver
 {
     /// <summary>
-    /// Helper class to hold a library range and framework.
+    /// Helper type to hold a library range and framework.
     /// </summary>
-    public class LibraryRangeCacheKey : IEquatable<LibraryRangeCacheKey>
+    public readonly struct LibraryRangeCacheKey : IEquatable<LibraryRangeCacheKey>
```
zivkan, line 15: As mentioned in the PR description, this is a breaking change. I searched the type on GitHub, and all I can find are forks of this repo, I can't find any external usage of the type. Similarly, internally I couldn't find any usage. Finally, given my limited knowledge of NuGet.DependencyResolver.Core, I don't expect this to be a library that many people will use, and this type might not be needed by most users of the resolver. There's only 1 reference to the type outside of the assembly, which is a class used for caching, in `NuGet.Commands`.

FWIW, Prism's Allocation blame put this as NuGet's 6th highest cause of allocations in VS 17.6. 

So normally I'd be fairly conservative about breaking API changes, but this is a top cause of allocations and it's just really hard to imagine anyone actually using this API. I'm leaning towards yes, we should approve it. But if others have feedback, I'd like to hear it.

I'm really surprised that the public API analyzer didn't catch this as a breaking change. I wonder if we need to update the analyzer package version.

jeffkl: I can't find any usages of the type either.  I lean towards making the change as the risk seems low.

jeffkl (approved):

src/NuGet.Core/NuGet.DependencyResolver.Core/LibraryRangeCacheKey.cs
nkolev92: Is this type used outside of NuGet.DependencyResolver.Core? 
Like, is there a chance to make it internal altogether.

drewnoakes: > is there a chance to make it internal altogether

I'll try that and push an update.

Looking at a different NuGet GCPauseWatson issue with a teammate, again the challenge with fixing it was modifying public API. If there's an opportunity to make a whole bunch of NuGet code internal in one hit, that could give a lot more freedom to improve code quality, and avoid having to pay the risk analysis/monitoring every time you consider such a change. I don't have enough context to know whether that is possible though.

drewnoakes: Making `LibraryRangeCacheKey` internal would require making `ResolverUtility.FindLibraryCachedAsync` internal too and giving IVT to `NuGet.Commands` (or maybe linking the source into both projects?).

nkolev92: Ah, I see. It's ok if it's used in multiple assemblies then. We don't want IVT among our assemblies.

nkolev92 (approved):

src/NuGet.Core/NuGet.DependencyResolver.Core/LibraryRangeCacheKey.cs
nkolev92: @zivkan I'm guessing you're ok with this being merged?

zivkan (approved):

PR #5206, opened 2023-06-06
Use default '/' split symbol in ETW events.

### LINKED ISSUE #12631: ETW events should use default '/' instead of '_'
### NuGet Product Used

Visual Studio Package Manager Console

### Product Version

N/A

### Worked before?

N/A

### Impact

None

### Repro Steps & Context

There is an issue in performance tests, not sure if this impact any of NuGet telemetry monitors.

As per David Kean:
We should avoid converting event/property names in ETW provider
This is allocating a lot during solution close and other places where lots of events are being produced, which is causing GCs to occur in some runs, introducing noise. The statement that says ETW events can't have '/' seems to be incorrect, as Windows, VS & .NET all include events with '/'. Also stopped converting properties as these are raw strings.

### Verbose Logs

_No response_

DevPaulLiu:
## Bug
Fixes: https://github.com/NuGet/Home/issues/12631

## Description
Avoid converting event/property names in ETW provider

This is allocating a lot during solution close and other places where lots of events are being produced, which is causing GCs to occur in some runs, introducing noise. The statement that says ETW events can't have '/' seems to be incorrect, as Windows, VS & .NET all include events with '/'. Also stopped converting properties as these are raw strings.

## PR Checklist

- [ ] PR has a linked issue.
- [ ] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.VisualStudio.Common/Telemetry/EtwLogActivity.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Common/Telemetry/EtwLogActivity.cs
@@ -14,7 +14,7 @@ namespace NuGet.VisualStudio.Telemetry
         {
             if (VsEtwLogging.IsProviderEnabled(VsEtwKeywords.Ide, VsEtwLevel.Information))
             {
-                var fullName = (VSTelemetrySession.VSEventNamePrefix + activityName).ToLowerInvariant().Replace('/', '_');
+                var fullName = (VSTelemetrySession.VSEventNamePrefix + activityName).ToLowerInvariant();
                 _activity = VsEtwLogging.CreateActivity(fullName, VsEtwKeywords.Ide, VsEtwLevel.Information);
             }
         }


srdjanjovcic (approved):

zivkan:
> The statement that says ETW events can't have '/' seems to be incorrect

@DevPaulLiu what statement?  If there's a comment somewhere (there's none in the file you modified in this PR, and I don't think we'd have written that down in any of our markdown docs), that should be modified.

DevPaulLiu:
> > The statement that says ETW events can't have '/' seems to be incorrect
> 
> @DevPaulLiu what statement? If there's a comment somewhere (there's none in the file you modified in this PR, and I don't think we'd have written that down in any of our markdown docs), that should be modified.

Hi Andy, we have some performance testing, which leverage these telemetry events. And there is a change in VS that avoid converting event/property names in ETW provider. That's why we made this change. [Pull Request 399963](https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/399963): Avoid converting event/property names in ETW provider

Here are more details about the perf tests [Task 1811217](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1811217): Add back correct measurement for Inner Loop tests

zivkan:
I see. It's a copy/paste from somewhere else that loses context, so it no longer makes sense here.

The change will break any tooling that expects the old name, but the NuGet team doesn't have any automation that listens to those ETW events, so we're unaffected.

zivkan (approved):

davkean (approved):

PR #5207, opened 2023-06-06
Add dependencies for PVP flow

### LINKED ISSUE #12633: Enable PVP flow for Nuget.Client in .NET source-build
Contributes to https://github.com/dotnet/source-build/issues/3043

PVP flow is a method of flowing dependencies in source build that only overrides versions of packages that the repo has a declared dependency on in `Version.Details.xml`. This means that the package flow is closer to the Microsoft's way of building .NET.

To fully enable PVP flow, the repo project in the VMR will need to set `<PackageVersionPropsFlowType>DependenciesOnly</PackageVersionPropsFlowType>` once these changes flow in.

NikolaMilosavljevic:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12633

Regression? No

## Description

Contributes to https://github.com/dotnet/source-build/issues/3043

PVP flow is a method of flowing dependencies in source build that only overrides versions of packages that the repo has a declared dependency on in Version.Details.xml. This means that the package flow is closer to the Microsoft's way of building .NET.

To enable this for `nuget.client`, we just need to add dependencies on the packages that should be overridden in source build. In this case, it's `Microsoft.Build` packages. We originally thought that these are used as references, and introduced relevant SBRP packages with https://github.com/dotnet/source-build-reference-packages/pull/694. However, runtime packages are needed, and this PR enables the flow of live packages.

To fully enable PVP flow, the repo project in the VMR will need to set `<PackageVersionPropsFlowType>DependenciesOnly</PackageVersionPropsFlowType>` once these changes flow in.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -5,4 +5,31 @@
     See https://github.com/dotnet/arcade/issues/2396 for details.
     See https://github.com/dotnet/arcade/blob/master/Documentation/DependencyDescriptionFormat.md#details-file for schema.
   -->
+
+  <!--
+    Following dependencies are required for PVP flow.
+    PVP flow infra will automatically bump the version to the latest live one.
+  -->
+  <ProductDependencies>
+    <Dependency Name="Microsoft.Build" Version="17.3.1">
+      <Uri>https://github.com/dotnet/msbuild</Uri>
+      <Sha>80f618ad45d38475773fd1a6eaa059f118a0ad5a</Sha>
+    </Dependency>
+    <Dependency Name="Microsoft.Extensions.FileProviders.Abstractions" Version="6.0.0">
+      <Uri>https://github.com/dotnet/runtime</Uri>
+      <Sha>8470979eb44c2218025515234d3e01138bd74afb</Sha>
+    </Dependency>
+    <Dependency Name="Microsoft.Extensions.FileSystemGlobbing" Version="6.0.0">
+      <Uri>https://github.com/dotnet/runtime</Uri>
+      <Sha>8470979eb44c2218025515234d3e01138bd74afb</Sha>
+    </Dependency>
+    <Dependency Name="Microsoft.Web.Xdt" Version="3.0.0">
+      <Uri>https://github.com/aspnet/xdt</Uri>
+      <Sha>d4d088b6a9c793525b1a27a119cb66ba4587bb39</Sha>
+    </Dependency>
+    <Dependency Name="System.ComponentModel.Composition" Version="4.5.0">
+      <Uri>https://github.com/dotnet/corefx</Uri>
+      <Sha>30ab651fcb4354552bd4891619a0bdd81e0ebdbf</Sha>
+    </Dependency>
+  </ProductDependencies>
 </Dependencies>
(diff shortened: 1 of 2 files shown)


NikolaMilosavljevic:
@mmitche @MichaelSimons @mthalman

jeffkl (dismissed):

MichaelSimons (dismissed):

nkolev92:
I'm trying to better understand what this would mean for our repo. 

- Does this apply to the special source-build only?
- Do we need this version to be the same as the one we have in Directory.Packages.props? Does the version here matter?
- Is there any automatic update? I'm guessing no, cause we'd end up with a circular reference among NuGet/MSBuild if we tried that.
-

NikolaMilosavljevic:
> I'm trying to better understand what this would mean for our repo.
> 
> * Does this apply to the special source-build only?
> * Do we need this version to be the same as the one we have in Directory.Packages.props? Does the version here matter?
> * Is there any automatic update? I'm guessing no, cause we'd end up with a circular reference among NuGet/MSBuild if we tried that.

Good questions. This is specific for source-build. This repo does not use Arcade build infra which would enable dependency flow via `Version.Details.xml` file. However, full source-build would honor this new dependency while evaluating PVP requirements - it would enable passing live package versions to `nuget.client` build. Note that today, full source-build passes versions of all available live (source-built) packages. PVP work enables us to scope this to only the properties (and packages) that are consumed by the specific repo.

The version does not matter, and it is not necessary to keep it in sync with version in `Directory.Packages.props`.

No automatic updates as this repo is not using Arcade.

In essence, this work enables full .NET source-build to pass only required packages into specific repo build (`nuget.client` in this case).

NikolaMilosavljevic:
This is not ready to be merged yet due to issues in full source-build that might be related: https://github.com/dotnet/installer/pull/16603

NikolaMilosavljevic:
> This is not ready to be merged yet due to issues in full source-build that might be related: [dotnet/installer#16603](https://github.com/dotnet/installer/pull/16603)

Will update the PR later today. Few more dependencies are needed due to: https://github.com/NuGet/NuGet.Client/blob/dbffb6c650ac44014bc9e5608bcf1bee6e858f11/Directory.Packages.props#L18-L22

NikolaMilosavljevic:
I've force-pushed a fix for missing dependencies. It would be great to get another review and to get checks running. cc @jeffkl @MichaelSimons

MichaelSimons (approved):

eng/Version.Details.xml
```diff
@@ -5,4 +5,31 @@
     See https://github.com/dotnet/arcade/issues/2396 for details.
     See https://github.com/dotnet/arcade/blob/master/Documentation/DependencyDescriptionFormat.md#details-file for schema.
   -->
+
+  <!--
+    Following dependencies are required for PVP flow.
```
nkolev92, line 10: It'd be great to enhance the comment here to say that the versions don't matter for us as we don't have arcade dependency flow.

NikolaMilosavljevic: Fixed with https://github.com/NuGet/NuGet.Client/pull/5207/commits/c4e7913463255147da1d0f43577c93c32733e8ad

NikolaMilosavljevic:
@zivkan @nkolev92 can we rerun checks? I have updated the PR with a requested change.

jeffkl (approved):

nkolev92 (approved):

PR #5214, opened 2023-06-07
upgrade System.Security.Cryptography.Xml package version

kartheekp-ms:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2339

[`Microsoft.Build.Tasks.Core`](https://www.nuget.org/packages/Microsoft.Build.Tasks.Core/17.3.1#dependencies-body-tab) package has a direct dependency on [`System.Security.Cryptography.Xml`](https://www.nuget.org/packages/System.Security.Cryptography.Xml/#versions-body-tab) package which has few versions marked (6.0.0 and 4.7.0) as vulnerable.

| Target framework | Top-level Package Id | Top-level Version | Transitive Dependency Package Id | Transitive Dependency Version |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| netcoreapp3.1  | Microsoft.Build.Tasks.Core  | 16.8.0 | System.Security.Cryptography.Xml | 4.7.0 (vulnerable)|
| netstandard2.0  | Microsoft.Build.Tasks.Core  | 16.8.0 | System.Security.Cryptography.Xml | 4.7.0 (vulnerable)|
| netcoreapp5.0  | Microsoft.Build.Tasks.Core  | 16.11.0 | System.Security.Cryptography.Xml | 4.7.0 (vulnerable)|
| > net5.0  | Microsoft.Build.Tasks.Core  | 17.3.1 | System.Security.Cryptography.Xml | 6.0.0 (vulnerable)|

I initially thought of leveraging [CPM transitive-pinning](https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management#transitive-pinning) feature but there were lot of restore failures. Hence ended up add `System.Security.Cryptography.Xml` package as direct dependency for projects that have a direct dependency on `Microsoft.Build.Tasks.Core` or `Microsoft.Build.Runtime > Microsoft.Build.Tasks.Core` to ensure that non-vulnerable versions are not downloaded during restore.

Regression? Last working version: n/a

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] N/A 

- **Documentation**
  - [x] N/A

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,6 +1,7 @@
 <Project>
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+    <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
 
     <CryptographyPackagesVersion Condition="'$(CryptographyPackagesVersion)' == ''">5.0.0</CryptographyPackagesVersion>
     <MicrosoftCodeAnalysisPublicApiAnalyzersVersion Condition="'$(MicrosoftCodeAnalysisPublicApiAnalyzersVersion)' == ''">3.0.0</MicrosoftCodeAnalysisPublicApiAnalyzersVersion>
@@ -91,6 +92,7 @@
     <PackageVersion Include="System.Security.Cryptography.Cng" Version="$(CryptographyPackagesVersion)" />
     <PackageVersion Include="System.Security.Cryptography.Pkcs" Version="$(CryptographyPackagesVersion)" />
     <PackageVersion Include="System.Security.Cryptography.ProtectedData" Version="4.4.0" />
+    <PackageVersion Include="System.Security.Cryptography.Xml" Version="6.0.1" />
     <PackageVersion Include="System.Text.Encoding.Extensions" Version="$(SystemPackagesVersion)" />
     <PackageVersion Include="System.Threading" Version="$(SystemPackagesVersion)" />
     <PackageVersion Include="System.Threading.Tasks" Version="$(SystemPackagesVersion)" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -184,6 +184,7 @@
       <_allowBuildFromSourcePackage Include="System.Security.Cryptography.Cng" />
       <_allowBuildFromSourcePackage Include="System.Security.Cryptography.Pkcs" />
       <_allowBuildFromSourcePackage Include="System.Security.Cryptography.ProtectedData" />
+      <_allowBuildFromSourcePackage Include="System.Security.Cryptography.Xml" />
 
       <_sourceBuildUnexpectedPackage Include="@(PackageReference)" Exclude="@(_allowBuildFromSourcePackage)" />
     </ItemGroup>
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -15,6 +15,10 @@
     <!-- The last package version of MSBuild that works with net5.0 is 16.11.0.  Our assemblies are still compiled against MSBuild assembly version 15.1.0.0 which will work with all versions -->
     <MicrosoftBuildVersion Condition="'$(TargetFramework)' == 'netcoreapp5.0'">16.11.0</MicrosoftBuildVersion>
 
+    <!-- Default System.Security.Cryptography.Xml version -->
+    <SystemSecurityCryptographyXml Condition="'$(SystemSecurityCryptographyXml)' == ''">6.0.1</SystemSecurityCryptographyXml>
+    <SystemSecurityCryptographyXml Condition="'$(MicrosoftBuildVersion)' == '16.8.0' Or '$(MicrosoftBuildVersion)' == '16.11.0'">4.7.1</SystemSecurityCryptographyXml>
+
     <!-- Overridden by source build to ensure the same version is used across products, do not remove these properties -->
     <MicrosoftExtensionsFileProvidersAbstractionsPackageVersion Condition="'$(MicrosoftExtensionsFileProvidersAbstractionsPackageVersion)' == ''">6.0.0</MicrosoftExtensionsFileProvidersAbstractionsPackageVersion>
     <MicrosoftExtensionsFileSystemGlobbingPackageVersion Condition="'$(MicrosoftExtensionsFileSystemGlobbingPackageVersion)' == ''">6.0.0</MicrosoftExtensionsFileSystemGlobbingPackageVersion>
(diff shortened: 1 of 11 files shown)


Directory.Packages.props
```diff
@@ -179,6 +184,7 @@
       <_allowBuildFromSourcePackage Include="System.Security.Cryptography.Cng" />
       <_allowBuildFromSourcePackage Include="System.Security.Cryptography.Pkcs" />
       <_allowBuildFromSourcePackage Include="System.Security.Cryptography.ProtectedData" />
+      <_allowBuildFromSourcePackage Include="System.Security.Cryptography.Xml" />
```
kartheekp-ms, line 187: I added this item to fix Source build failure on CI. I hope that's okay. (I don't know how source build works internally)

src/NuGet.Clients/NuGet.VisualStudio.Common/NuGet.VisualStudio.Common.csproj
```diff
@@ -17,6 +17,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Build" ExcludeAssets="runtime" />
     <PackageReference Include="Microsoft.Build.Tasks.Core" ExcludeAssets="runtime" PrivateAssets="All" />
+    <PackageReference Include="System.Security.Cryptography.Xml" ExcludeAssets="runtime" PrivateAssets="All" />
```
zivkan: Sort alphabetically, ~and I think the MSBuild packages have `ExcludeAssets` for a reason (something would go wrong otherwise, we want MSBuild to be loaded from its default path). I guess since tests pass that using `ExcludeAssets` doesn't cause a problem, but otherwise I don't understand why ExcludeAssets is being used. The packages that are not Microsoft.Build.* below don't have Exclude assets.~

Perhaps there should be a comment to explain that System.Security.Cryptography.Xml is a transitive dependency of MIcrosoft.Build.Tasks.Core, and that's why it's using the same PrivateAssets value.

src/NuGet.Clients/NuGet.VisualStudio.Implementation/NuGet.VisualStudio.Implementation.csproj
```diff
@@ -31,6 +31,7 @@
     <PackageReference Include="Microsoft.Build" IncludeAssets="None" PrivateAssets="all"/>
     <PackageReference Include="Microsoft.Build.Framework" IncludeAssets="None" PrivateAssets="all"/>
     <PackageReference Include="Microsoft.Build.Tasks.Core" IncludeAssets="None" PrivateAssets="all"/>
+    <PackageReference Include="System.Security.Cryptography.Xml" IncludeAssets="None" PrivateAssets="All" />
```
zivkan: Sort alphabetically. Maybe add comment that it's a dependency of Microsoft.Build.Tasks.Core, so if that package gets upgraded, we can check if we can remove this PackageReference.

src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.csproj
```diff
@@ -45,6 +45,7 @@
     <PackageReference Include="Microsoft.Build.Framework" ExcludeAssets="runtime" GeneratePathProperty="true" />
     <PackageReference Include="Microsoft.Build.Tasks.Core" ExcludeAssets="runtime" GeneratePathProperty="true" />
     <PackageReference Include="Microsoft.Build.Utilities.Core" ExcludeAssets="runtime" GeneratePathProperty="true" />
+    <PackageReference Include="System.Security.Cryptography.Xml" ExcludeAssets="Runtime" GeneratePathProperty="true"/>
```
zivkan: This package's package path property isn't being used, so you don't need `GeneratePathProperty`.

~Similarly to another file, I'm not sure the ExcludeAssets is needed, whereas I think it is needed for the MSBuild packages, which is why they have them.~

Perhaps there should be a comment to explain that System.Security.Cryptography.Xml is a transitive dependency of MIcrosoft.Build.Tasks.Core, and that's why it's using the same ExcludeAssets value.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/NuGet.XPlat.FuncTest.csproj
```diff
@@ -8,6 +8,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Build.Runtime" />
+    <PackageReference Include="System.Security.Cryptography.Xml" />
```
zivkan: Microsoft.CodeAnalysis.CSharp is in the wrong place for this list to have been sorted alphabetically before, but I think it's still worthwhile to move the new package to the end to make it mostly sorted.

test/NuGet.Core.Tests/NuGet.Build.Tasks.Pack.Test/NuGet.Build.Tasks.Pack.Test.csproj
```diff
@@ -18,6 +18,7 @@
   <ItemGroup Condition=" '$(IsCore)' == 'true' ">
     <PackageReference Include="Microsoft.Build.Framework" PrivateAssets="All" />
     <PackageReference Include="Microsoft.Build.Tasks.Core" PrivateAssets="All" />
+    <PackageReference Include="System.Security.Cryptography.Xml" PrivateAssets="All" />
```
zivkan: sort, please.

test/TestUtilities/Test.Utility/Test.Utility.csproj
```diff
@@ -85,6 +85,7 @@
     <PackageReference Include="FluentAssertions" />
     <PackageReference Include="Microsoft.Build" ExcludeAssets="runtime" />
     <PackageReference Include="Microsoft.Build.Tasks.Core" ExcludeAssets="runtime" />
+    <PackageReference Include="System.Security.Cryptography.Xml" ExcludeAssets="runtime" />
```
zivkan: sort packages, please.

~Also, since this doesn't use `PrivateAssets="all"`, then any test project with a project reference to `Test.Utility` will end up with this package transitively. So, I'm not sure all the `PrivateAssets="all"` on all the other test projects makes a difference. Well, maybe it "reverts" the ExcludeAssets in this PackageReference. Note also that only MSBuild packages here have ExcludeAssets, the other packages don't, so I'm not sure that it's needed for this package. If it is, it's not clear why.~

Directory.Packages.props
```diff
@@ -15,6 +15,10 @@
     <!-- The last package version of MSBuild that works with net5.0 is 16.11.0.  Our assemblies are still compiled against MSBuild assembly version 15.1.0.0 which will work with all versions -->
     <MicrosoftBuildVersion Condition="'$(TargetFramework)' == 'netcoreapp5.0'">16.11.0</MicrosoftBuildVersion>
 
+    <!-- Default System.Security.Cryptography.Xml version -->
+    <SystemSecurityCryptographyXml Condition="'$(SystemSecurityCryptographyXml)' == ''">6.0.1</SystemSecurityCryptographyXml>
+    <SystemSecurityCryptographyXml Condition="'$(MicrosoftBuildVersion)' == '16.8.0' Or '$(MicrosoftBuildVersion)' == '16.11.0'">4.7.1</SystemSecurityCryptographyXml>
```
zivkan: The condition on MicrosoftBuildVersion is an interesting trick to make it so when we upgrade the MSBuild package version, this package will work differently. However, there's no logic on all the PackageReferences for the package, or on the PackageVersion either. In which case I expect there'd be a bunch of restore errors.

For this reason, I think the comment at the top of the block should explain it's a transitive dependency of MSBuild, and can probably be removed when MSBuild is updated to a newer version.

jeffkl (dismissed):
Please sort the package list as Andy requested

src/NuGet.Clients/NuGet.VisualStudio.Common/NuGet.VisualStudio.Common.csproj
kartheekp-ms: Thank you for the feedback. I addressed all the comments related to sorting and the PR is ready for review again.

zivkan (approved):

jeffkl (approved):

PR #5215, opened 2023-06-07
Add AsString() for LibraryDependencyTarget and LibraryIncludeFlags

### LINKED ISSUE #12638: Avoid repeated Enum.ToString() in PackageSpecWriter.SetDependencies
PackageSpecWriter.SetDependencies calls Enum.ToString() on two enum types and multiple fields:

LibraryDependencyTarget
LibraryIncludeFlags

This causes unavoidable stringbuilder allocations to produce the various combinations. Normally this never shows up, however, in NuGut this code path is called a lot and hence it needs to work around this.

Erarndt:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12638

## Description
This change introduces an `AsString()` extension to handle the string conversion. There is a `Dictionary` backing this method that avoids duplicates of the same string. The full set of potential flags is high, but it's expected that the actual number of distinct strings required at runtime will be significantly lower.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.LibraryModel/LibraryDependencyTargetUtils.cs
+++ b/src/NuGet.Core/NuGet.LibraryModel/LibraryDependencyTargetUtils.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 
@@ -135,7 +136,7 @@ namespace NuGet.LibraryModel
             return string.Join(",", flagStrings);
         }
 
-        private static readonly Dictionary<LibraryDependencyTarget, string> LibraryDependencyTargetCache = new();
+        private static readonly ConcurrentDictionary<LibraryDependencyTarget, string> LibraryDependencyTargetCache = new();
 
         /// <summary>
         /// Efficiently converts <see cref="LibraryDependencyTarget"/> to it's <see cref="string"/> representation.
@@ -147,7 +148,7 @@ namespace NuGet.LibraryModel
             if (!LibraryDependencyTargetCache.TryGetValue(includeFlags, out string enumAsString))
             {
                 enumAsString = includeFlags.ToString();
-                LibraryDependencyTargetCache[includeFlags] = enumAsString;
+                LibraryDependencyTargetCache.TryAdd(includeFlags, enumAsString);
             }
 
             return enumAsString;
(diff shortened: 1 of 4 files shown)


src/NuGet.Core/NuGet.LibraryModel/LibraryDependencyTargetUtils.cs
```diff
@@ -135,6 +135,24 @@ public static string GetFlagString(LibraryDependencyTarget flags)
...
+        /// Efficiently converts <see cref="LibraryDependencyTarget"/> to it's <see cref="string"/> representation.
+        /// </summary>
+        /// <param name="includeFlags">The <see cref="LibraryDependencyTarget"/> instance to get the <see cref="string"/> representation for.</param>
+        /// <returns>The <see cref="string"/> representation of <paramref name="includeFlags"/>.</returns>
+        public static string AsString(this LibraryDependencyTarget includeFlags)
+        {
+            if (!LibraryDependencyTargetCache.TryGetValue(includeFlags, out string enumAsString))
+            {
+                enumAsString = includeFlags.ToString();
+                LibraryDependencyTargetCache[includeFlags] = enumAsString;
```
zivkan: This is not threadsafe, and Nuget restores projects in parallel. This will cause problems.

src/NuGet.Core/NuGet.LibraryModel/LibraryIncludeFlagUtils.cs
```diff
@@ -118,5 +118,23 @@ public static LibraryIncludeFlags GetFlags(string flags, LibraryIncludeFlags def
...
+        /// Efficiently converts <see cref="LibraryIncludeFlags"/> to it's <see cref="string"/> representation.
+        /// </summary>
+        /// <param name="includeFlags">The <see cref="LibraryIncludeFlags"/> instance to get the <see cref="string"/> representation for.</param>
+        /// <returns>The <see cref="string"/> representation of <paramref name="includeFlags"/>.</returns>
+        public static string AsString(this LibraryIncludeFlags includeFlags)
+        {
+            if (!LibraryIncludeFlagsCache.TryGetValue(includeFlags, out string enumAsString))
+            {
+                enumAsString = includeFlags.ToString();
+                LibraryIncludeFlagsCache[includeFlags] = enumAsString;
```
zivkan: This is not threadsafe, and Nuget restores projects in parallel. This will cause problems.

src/NuGet.Core/NuGet.LibraryModel/LibraryDependencyTargetUtils.cs
Erarndt: Ah, my mistake. Unless I'm missing something, `ConcurrentDictionary<LibraryDependencyTarget, string>` should be sufficient here?

zivkan (approved):

PR #5216, opened 2023-06-07
Avoid several allocations in call to GetDirectoryContents

### LINKED ISSUE #12639: VirtualFileProvider.GetDirectoryContents Recreates a String
VirtualFileProvider splits up files into their respective path components in its constructor:

        public VirtualFileProvider(List<string> files)
        {
            _files = files.Select(file => file.Split(ForwardSlashChar)).ToList();
        }

And then goes to recreate these exact strings in VirtualFileProvider.GetDirectoryContents via string.Join:

            foreach (var file in _files)
            {
              ...
              var virtualFile = new VirtualFileInfo(string.Join(ForwardSlash, file));
              contents.Add(virtualFile);
            }

This can be avoided by keeping the original string around.

Erarndt:
## Bug
`VirtualFileProvider.GetDirectoryContents` is frequently called and produces significant GC pression. Reducing the number of allocations can significantly improve performance. 

Fixes: https://github.com/NuGet/Home/issues/12639

## Description
A large portion of allocations can be avoided by keeping the string passed to the `VirtualFileProvider` constructor around that is split and recreated repeatedly later. Additionally,  the `VirtualFileProvider` constructor previously only accepted a `List<string>` that only ever had one item. This required the single internal caller to create an intermediate list. Due to this `EnumerableDirectoryContents` was also updated to handle a single item.

## PR Checklist

- [X] PR has a linked issue.
- [X] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A


nkolev92 (approved):

nkolev92:
@Erarndt Mind rebasing? There's a dartlab outage that's making our E2E tests fail and the update is https://github.com/NuGet/NuGet.Client/commit/def121245e22c924a1d06164c3935cf35de7395e. 

Probably applies to the rest of the PRs as well.

Erarndt:
> @Erarndt Mind rebasing? There's a dartlab outage that's making our E2E tests fail and the update is [def1212](https://github.com/NuGet/NuGet.Client/commit/def121245e22c924a1d06164c3935cf35de7395e).
> 
> Probably applies to the rest of the PRs as well.

Updated the PRs

src/NuGet.Core/NuGet.Commands/RestoreCommand/ContentFiles/VirtualFileProvider.cs
```diff
@@ -83,18 +79,69 @@ public IChangeToken Watch(string filter)
...
+                    _current = _entry;
+                    _done = true;
+
+                    return true;
+                }
+            }
+
+            public void Reset()
+            {
+                _done = _entry is not null;
```
zivkan, line 142: This is the opposite of what the constructor does, so it feels like a bug. Should the constructor set `_entry`, and then call `Reset()`, to ensure that the implementation is the same? (reduce code duplication)

src/NuGet.Core/NuGet.Commands/RestoreCommand/ContentFiles/VirtualFileProvider.cs
```diff
@@ -83,18 +79,69 @@ public IChangeToken Watch(string filter)
...
 
             public bool Exists => true;
 
-            public IEnumerator<IFileInfo> GetEnumerator() => _entries.GetEnumerator();
+            public IEnumerator<IFileInfo> GetEnumerator() => new SingleFileInfoEnumerator(_entry);
+
+            IEnumerator IEnumerable.GetEnumerator() => new SingleFileInfoEnumerator(_entry);
+        }
+
+        private struct SingleFileInfoEnumerator : IEnumerator<IFileInfo>
```
zivkan, line 96: This struct will always be referenced through an `IEnumerator` interface, and therefore will always be boxed. From https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/types/boxing-and-unboxing#performance:

> To a lesser degree, the cast required for unboxing is also expensive computationally.

Also https://learn.microsoft.com/en-us/dotnet/framework/performance/performance-tips#boxing-and-unboxing says more or less similar things.

On one hand, it's not being boxed into `object`, so I'm not sure if the cast on unboxing is actually relevant. But maybe the internal implementation of boxing is that it's always put in an `object`, and therefore does need to get typecast on the other side?

Anyway, are we better off just making this a class?

src/NuGet.Core/NuGet.Commands/RestoreCommand/ContentFiles/VirtualFileProvider.cs
```diff
@@ -16,59 +16,55 @@ namespace NuGet.Commands
...
             // Remove the root identifier from the relative path
-            if (string.Equals(subPathParts.FirstOrDefault(), RootDir, StringComparison.Ordinal))
+            var subPathPartsOffset = subPathParts.Length > 0 && string.Equals(subPathParts[0], RootDir, StringComparison.Ordinal) ? 1 : 0;
+            _splitFile ??= _originalFile.Split(ForwardSlashChar);
+
+            int i = 0;
+            // Walk the path as long as both the file and subpath contain the same directories
+            while (i < _splitFile.Length - 1
+                && i + subPathPartsOffset < subPathParts.Length
+                && string.Equals(_splitFile[i], subPathParts[i + subPathPartsOffset], StringComparison.OrdinalIgnoreCase))
```
zivkan, line 42: ctrl-f for `_splitFile[` only finds this copy, which means that it's never assigned to, this `string.Equals` will always return false, and `.Take` elsewhere is always going to return nulls.

Am I reading the PR wrong?

src/NuGet.Core/NuGet.Commands/RestoreCommand/ContentFiles/VirtualFileProvider.cs
davkean: Just throw on reset like the auto-generated ones in C# do, its not worth implementing something that never gets called.

Erarndt:
@drewnoakes addresses this issue and more in https://github.com/NuGet/NuGet.Client/pull/5231

Going to close this as a result.

PR #3765, opened 2020-11-16
Restore: improve performance by avoiding linear search in the implementation of the "Nearest wins" rule

marcin-krystianc:
## Bug

Fixes: https://github.com/NuGet/Home/issues/10286
Regression: No

## Fix

The predicate returned by the `ChainPredicate` method has better algorithmic complexity than before. 
It uses dictionary lookup instead of linear search which as shown in the results below can have a significant impact on the overall restore time.

## Testing

Tests Added: No  
Reason for not adding tests: 
 - No new features are being added.
 - This part of code is already well tested

## Validation (Benchmarking)

Used solutions:
 - [NuGet.Client](https://github.com/NuGet/NuGet.Client)
 - [OrchardCore ](https://github.com/OrchardCMS/OrchardCore)
 - [Orleans](https://github.com/dotnet/orleans)
 - [SanitisedNet471](https://github.com/NuGet/Home/issues/10030#issue-701185886)
 
Test procedure:
- Vanilla [RunPerformanceTests.ps1](https://github.com/NuGet/NuGet.Client/blob/dev/scripts/perftests/RunPerformanceTests.ps1) script was used for running the individual benchmarks.
- As we are interested in the performance of the graph walking algorithm, the "force restore" scenario with static graph option was chosen.
- To achieve high confidence in the results each individual benchmark was run twice with high number of iterations (50 or 250)

Results:

|                        |  N-Runs |   Mean  |  Median |   Min   |   Max   | StdDev  |  Mean Change |
|------------------------|:-------:|:-------:|:-------:|:-------:|:-------:|:-------:|:-------:|
| NuGet.Client(Base)     |  2x250  |  16.57s |  16.57s |  15.72s |  17.57s |  0.320  |         |
| NuGet.Client(Branch)   |  2x250  |  16.46s |  16.47s |  15.46s |  17.49s |  0.314  |  -0.66% |
| Orleans(Base)          |  2x250  |  14.99s |  14.97s |  14.35s |  15.73s |  0.269  |         |
| Orleans(Branch)        |  2x250  |  14.88s |  14.87s |  13.93s |  15.49s |  0.265  |  -0.76% |
| OrchardCore(Base)      |  2x50   |  25.48s |  25.48s |  24.61s |  26.62s |  0.430  |         |
| OrchardCore(Branch)    |  2x50   |  23.84s |  23.84s |  22.82s |  25.03s |  0.471  |  -6.42% |
[...]

PR #5218, opened 2023-06-07
Reduce some allocations in CreateGraphNode.

### LINKED ISSUE #12641: CreateGraphNode has a high number of allocations
There are a high number of allocations during calls to `CreateGraphNode`. Finding ways to reduce this will enhance overall performance and responsiveness.

Erarndt:
## Bug

CreateGraphNode creates  enough allocations to be a performance issue, and any reduction in this path will be beneficial.

Fixes: https://github.com/NuGet/Home/issues/12641

## Description
Collapses two collections that track `LibraryDependency` instances into one `HashSet<LibraryDependency>`. Also minimizes resizing when adding parent nodes.

## PR Checklist

- [X] PR has a linked issue.
- [X] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A


nkolev92 (dismissed):

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
```diff
@@ -107,15 +106,8 @@ public async Task<GraphNode<RemoteResolveResult>> WalkAsync(LibraryRange library
...
-                        if (dependencies == null)
-                        {
-                            // Init dependency lists
-                            dependencies = new List<LibraryDependency>(1);
-                            runtimeDependencies = new HashSet<string>();
-                        }
-
-                        dependencies.Add(libraryDependency);
-                        runtimeDependencies.Add(libraryDependency.Name);
+                        runtimeDependencies ??= new HashSet<LibraryDependency>(LibraryDependencyNameComparer.OrdinalIgnoreCaseNameComparer);
```
kartheekp-ms, line 109: How about using the List instead of HashSet here. I don't see any call to HashSet.Contains in this logic? If that works, we can eliminate the call to `.ToList()` while creating `RemotResolveResult` object here https://github.com/NuGet/NuGet.Client/pull/5218/files#diff-7457360b2ac43e8123a03561b1b5e2ca1832954941f84e44086365d26f22e5faR147

zivkan: list allows duplicates, HashSet does not.

edit: Having looked at this PR more carefully, the old code checked `.Contains` before an `.Add` (around line 145), so is an O(n) operation, compared to HashSet's (theoretical) O(1)-ish. However, realistically n will always be small, and there are many real-world examples where algorithms with worse complexity is faster for small n (for example quick sort vs simple O(n^2) algorithms). So, it's really not clear what has better perf, and I say that as someone working on NuGet for years, not an external contributor who was directed here by a tool that detected a large number of allocations, but otherwise doesn't know the code.

So, I think Kartheek could be right about another opportunity to reduce allocations and improve perf. But I think it might require a data analysis, as there's (low) risk of significantly increasing CPU usage.

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
```diff
@@ -543,12 +529,35 @@ internal void AddParentsToNode(GraphNode<RemoteResolveResult> node)
             {
                 lock (_transitiveCentralPackageVersions)
                 {
-                    foreach (var parent in _transitiveCentralPackageVersions[node.Item.Key.Name])
+                    List<GraphNode<RemoteResolveResult>> graphNodes = _transitiveCentralPackageVersions[node.Item.Key.Name];
+                    if (node.ParentNodes is List<GraphNode<RemoteResolveResult>> parentNodes)
                     {
-                        node.ParentNodes.Add(parent);
+                        parentNodes.AddRange(graphNodes);
```
zivkan: NuGet has a helper extension `AddRange` for `ICollection<T>`, in `CollectionExtensions.cs`.

You can make this method use that extension method, and then do your typecast trick in the helper method, which could benefit other parts of the code, not just this one.

Also, it might be worthwhile adding a comment that `List<T>.AddRange` is optimized for at most one resize, as it might not be intuitive to all readers what the benefit is to typecasting is, when typecast has a cost and will ultimately enumerate the incoming collection anyway.

zivkan (dismissed):

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
```diff
@@ -138,14 +130,11 @@ public async Task<GraphNode<RemoteResolveResult>> WalkAsync(LibraryRange library
...
-            if (dependencies?.Count > 0)
+            if (runtimeDependencies?.Count > 0)
             {
                 foreach (var nodeDep in node.Item.Data.Dependencies)
                 {
-                    if (runtimeDependencies?.Contains(nodeDep.Name, StringComparer.OrdinalIgnoreCase) != true)
-                    {
-                        dependencies.Add(nodeDep);
-                    }
+                    runtimeDependencies.Add(nodeDep);
```
kartheekp-ms, line 137: In the previous implementation, `nodeDep` is added to the `dependencies` list only if it is present in the hashset `runtimeDependencies` already. In the new implementation, we are adding to the hashset directly. I am not sure if there is any untested code path hence an issue in this logic was not caught. Please let me know if I am missing something.

nkolev92: dependencies was a list, so it was being done that way to avoid duplicates in dependencies, which is now not possible because dependencies is a hashset

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
Erarndt: Updated the extension method

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
Erarndt: Since this is expected to always have very few elements, let's swap to use a list for now. The original code had both, so this is a strict improvement already.

nkolev92:
Making assumption about the sizes of dependency lists has yielded mixed results in the past. See https://github.com/NuGet/NuGet.Client/pull/3765. 

To me, it's not immediately obvious that the benefit in allocations is not diminished by the fact that we're now using multiple lists. 
https://github.com/NuGet/NuGet.Client/tree/dev/scripts/perftests can sometimes catch some of these problems. 

Wouldn't the *set* only be more efficient? In this we're only removing 1 collection, but we're largely still making the same operations. With the latest change, we're going from list + set to 2 lists.

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
kartheekp-ms: I am unresolving this conversation because the current changes proposed has `runtimeDependencies.Contains` call twice. As we it is an O(n) operation. Looking at the call stack in https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1824305 and https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1824313, I think the best thing to do is first merge the changes to `CollectionExtensions.cs` and the following logic in a separate PR and then we will just focus on `CreateGraphNode` method in this PR. What do you think?

```cs
lock (_transitiveCentralPackageVersions)
                {
                    List<GraphNode<RemoteResolveResult>> graphNodes = _transitiveCentralPackageVersions[node.Item.Key.Name];
                    node.ParentNodes.AddRange(graphNodes);
                }
```

Erarndt: @kartheekp-ms, I originally made the change based on the suggestions from you and @zivkan. Based on feedback from @nkolev92, it seems that we can't rely on this collection always being small.

I'm inclined to revert the most recent change and go back to the implementation using `HashSet<LibraryDependency>`. We keep the O(1) check to avoid duplicates and the `ToList()` call will create a precisely sized list, so we'd save potential List resizes. Ideally, we'd just use the `HashSet<LibraryDependency>` directly as @nkolev92 suggests, but that would cause changes in several other areas that I'm not sure we'd want to take.

Let me know what you think.

kartheekp-ms: > I'm inclined to revert the most recent change and go back to the implementation using HashSet<LibraryDependency>. We keep the O(1) check to avoid duplicates and the ToList() call will create a precisely sized list, so we'd save potential List resizes.

Sorry for the confusion. Let's revert back to the previous commit.

Erarndt: Reverted back. Let me know what you think.

nkolev92:
Hey folks, 

Given the number of changes we've had, I'm thinking we punt PRs that don't have direct unit tests for the next version. 
That way we'd minimize the risk of regression + if we do have regressions, we'd have an easier time finding them. 

Thoughts?

kartheekp-ms (dismissed):

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
```diff
@@ -543,12 +529,25 @@ internal void AddParentsToNode(GraphNode<RemoteResolveResult> node)
...
-                    {
-                        node.ParentNodes.Add(parent);
-                    }
+                    List<GraphNode<RemoteResolveResult>> graphNodes = _transitiveCentralPackageVersions[node.Item.Key.Name];
+                    node.ParentNodes.AddRange(graphNodes);
                 }
             }
         }
+
+        private class LibraryDependencyNameComparer : IEqualityComparer<LibraryDependency>
```
jeffkl, line 538: Is it worth adding a private constructor to these comparers where we only want callers to access the singleton?  I'm wondering if some future dev will accidentally instantiate it instead of using the singleton and we don't have any code analysis to catch it.  Just wondering your thoughts?

jeffkl (dismissed):

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
Erarndt: Good idea. Adding it.

nkolev92 (approved):

jeffkl (approved):

PR #5222, opened 2023-06-09
Reduce allocations in PackageSpec object creation logic using RuntimeGraph.Empty

### LINKED ISSUE #12649: PackageSpec should use an empty RuntimeGraph instead of a new one
[Internal tracking issue](https://devdiv.visualstudio.com/DevDiv/_queries/edit/1826714/?triage=true)

kartheekp-ms:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12649

## Description

Reduce the allocations caused by the PackageSpec using RuntimeGraph.Empty instead of creating a new RuntimeGraph object every time during object creation.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] N/A 

- **Documentation**
  - [x] N/A


nkolev92 (dismissed):
I get that the change is likely safe, but it's giving me a bit of pause given that there's no code guarantee that this shared object won't accidentally change.

kartheekp-ms:
> I get that the change is likely safe, but it's giving me a bit of pause given that there's no code guarantee that this shared object won't accidentally change.

I modified the default constructor of the `RuntimeGraph` class (by adding win-x64 to the RuntimeDescription list). Out of the four tests that use the default constructor, two of them failed. This suggests that we have test coverage for the default constructor's functionality.

```cs
        public RuntimeGraph()
            : this(new List<RuntimeDescription>() { new RuntimeDescription("win-x64") }, Enumerable.Empty<CompatibilityProfile>())
        {
        }
```

image

davkean:
> I get that the change is likely safe, but it's giving me a bit of pause given that there's no code guarantee that this shared object won't accidentally change.

RuntimeGraphs are immutable apart from the cache, and the cache should be removed as a result of https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1826702.

jeffkl (approved):

drewnoakes:
https://github.com/NuGet/NuGet.Client/pull/5232 removes those caches, and removes an (unused) setter that accidentally made the type non-immutable.

PR #5225, opened 2023-06-12
Onboard onto Dartlab build-to-build upgrade feature

kartheekp-ms:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/1584

## Description

I followed the instructions in our [internal wiki](https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/34040/Component-Repository) to start using the build-to-build upgrade feature. According to the information I found, instead of setting up a new machine from scratch, this feature allows me to use an existing machine from our pool that meets the required criteria. When we deploy the machine, this feature updates Visual Studio to the necessary version instead of installing it again, which saves time for running tests.

I want to give a shout-out to @vishavpandhi for helping me throughout the onboarding process. I learned that the build-to-build upgrade feature only works when we're testing on the `int.main` VS branch. However, if we need to go back to the previous release version because of a problem caused by another team, DartLab will install Visual Studio from scratch.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] N/A 

- **Documentation**
  - [x] N/A

### DIFF
--- a/build/config.props
+++ b/build/config.props
@@ -27,7 +27,7 @@
     <VsTargetMajorVersion>$([MSBuild]::Add(11, $(MajorNuGetVersion)))</VsTargetMajorVersion>
     <VsTargetBranch>main</VsTargetBranch>
     <VsTargetChannel>int.$(VsTargetBranch)</VsTargetChannel>
-    <VsTargetChannelForTests>int.d17.5</VsTargetChannelForTests>
+    <VsTargetChannelForTests>int.$(VsTargetBranch)</VsTargetChannelForTests>
 
     <!-- NuGet SDK VS package Semantic Version -->
     <NuGetSdkVsSemanticVersion>$(VsTargetMajorVersion).$(MinorNuGetVersion).$(PatchNuGetVersion)</NuGetSdkVsSemanticVersion>
--- a/build/config.props
+++ b/build/config.props
@@ -27,7 +27,7 @@
     <VsTargetMajorVersion>$([MSBuild]::Add(11, $(MajorNuGetVersion)))</VsTargetMajorVersion>
     <VsTargetBranch>main</VsTargetBranch>
     <VsTargetChannel>int.$(VsTargetBranch)</VsTargetChannel>
-    <VsTargetChannelForTests>int.$(VsTargetBranch)</VsTargetChannelForTests>
+    <VsTargetChannelForTests>int.d17.5</VsTargetChannelForTests>
 
     <!-- NuGet SDK VS package Semantic Version -->
     <NuGetSdkVsSemanticVersion>$(VsTargetMajorVersion).$(MinorNuGetVersion).$(PatchNuGetVersion)</NuGetSdkVsSemanticVersion>
--- a/eng/pipelines/templates/Apex_Tests_On_Windows.yml
+++ b/eng/pipelines/templates/Apex_Tests_On_Windows.yml
@@ -37,8 +37,8 @@ stages:
           value: ${{parameters.bootstrapperUrl}}
         - ${{parameters.variables}}
       runSettingsURI: ${{parameters.runSettingsURI}}
-      visualStudioBootstrapperURI: $(bootstrapperUrl)
-      baseVisualStudioBootstrapperURI: $(bootstrapperUrl)
+      visualStudioBootstrapperURI: $(bootstrapperUrl)      
+      baseVisualStudioBootstrapperURI: https://vsdrop.corp.microsoft.com/file/v1/$(VisualStudio.BaseBuild.ProductsDropName);${{parameters.visualStudioBootstrapperDropPath}}
       candidateBaselineBuilds: $(BaselineBuildCommitIds)
       testLabPoolName: VS-Platform
       dartLabEnvironment: ${{parameters.DartLabEnvironment}}
--- a/eng/pipelines/templates/Apex_Tests_On_Windows.yml
+++ b/eng/pipelines/templates/Apex_Tests_On_Windows.yml
@@ -35,7 +35,8 @@ stages:
       variables:
         - name: bootstrapperUrl
           value: ${{parameters.bootstrapperUrl}}
-        - ${{parameters.variables}}
+        - ${{if gt(length(parameters.variables), 0)}}:
+          - ${{parameters.variables}}
       runSettingsURI: ${{parameters.runSettingsURI}}
       visualStudioBootstrapperURI: $(bootstrapperUrl)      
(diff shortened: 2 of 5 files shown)


jeffkl (approved):
What sort of time savings is there?  It looks like machine deployment still takes a while in your PR build, will that time go down for subsequent builds?

kartheekp-ms:
> What sort of time savings is there? It looks like machine deployment still takes a while in your PR build, will that time go down for subsequent builds?

Great question Jeff. I forgot to add the details in the PR description. I was told that once all the agents in the machine pool have VS main installed then the machine deployment takes less time for subsequent builds. It looks like the agents in the pool are recycled after few days. I think we should let our PR builds run for few weeks in order to measure the accurate improvements of this change. If we ever have to run our tests on the previous VS release version because of a problem caused by another team, DartLab will install Visual Studio from scratch instead of upgrading Visual Studio.

donnie-msft (approved):
Happy to see this is done so soon!
I don't really follow how we update the baseline. I might be glossing over it. I just see where we read the baseline and store it for tests to use. My questions are:
1. If an AzDo Build fails (even not on the build step), do we still end up moving the baseline forward? In other words, how do we protect setting the baseline to a commit that's not fully green?
2. For releases, official builds, etc, do those still use whatever VS is installed on the agent? (ie, not the incremental build?)

nkolev92:
I'd like to understand how the baseline is determined as well.

vishavpandhi:
We maintain a pool of machines which already has `Visual Studio` installed. All of these machines have a tag on them that corresponds to the commit from which this Visual Studio was created.

To find the baseline for the current build, we look at its git commit history and find the last 100 commits.
Then we try to find machines in our pool which has Visual Studio (visual studio created from any of the above 100 commits) installed.

1. If we are able to find machines, then we will run Upgrade on this machine (upgrade it to the target build).
2. If we don't find any machine in the pool, then we will create machine from scratch. We also create 1 extra machine with tag=`Current Commit ID`, so that it can be used later.

In our testing, we saw 50% saving in machine deployment time

Currently, B2B upgrade is available only on the `main` branch and not `rel` branches.

In case, we did find a baseline, but this happens to be the same as the current/target build, we disable the Upgrade and create machine from scratch as it happened in [this pipeline run](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7897328&view=logs&j=8a73fcdf-163b-5991-5c80-83261817a503&t=95b5d0ad-391e-5588-1689-0449479bc8ba&l=1796)

`WARNING: Base version '17.7.33809.247' is greater than or equal to '17.7.33809.247'. Disabling upgrade.
`

kartheekp-ms:
> In case, we did find a baseline, but this happens to be the same as the current/target build, we disable the Upgrade and create machine from scratch as it happened in [this pipeline run](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7897328&view=logs&j=8a73fcdf-163b-5991-5c80-83261817a503&t=95b5d0ad-391e-5588-1689-0449479bc8ba&l=1796)
> 
> `WARNING: Base version '17.7.33809.247' is greater than or equal to '17.7.33809.247'. Disabling upgrade. `

I appreciate @vishavpandhi for providing additional details on how the base version is determined. If the base version matches the target version, my assumption is that DartLab should not install Visual Studio on the agent machine. This is because the agent already has the target version installed. However, according to the [logs](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7897328&view=logs&j=0a108fa2-1091-581d-dd54-eafaa15161a9&t=09932453-0ca4-55db-1d1a-f7920de8e48b&l=30), it seems that VS was indeed installed. Please let me if there is something I'm overlooking.

```
##Deploying artifact 1/4: Install Visual Studio Under Test
##Deploying artifact 2/4: Restart Machine
##Deploying artifact 3/4: Test Visual Studio Instance
##Deploying artifact 4/4: Add User Capabilities

```

kartheekp-ms:
> Currently, B2B upgrade is available only on the `main` branch and not `rel` branches.

 we usually run end-to-end tests using the main branch of Visual Studio. However, there are instances when a regression, caused by our partner team, requires us to pin the Visual Studio version to a specific release branch. In such cases, I anticipate that the pipeline should not result in failure. Instead, it should disable B2B upgrade features and resort to installing Visual Studio from the beginning for each build. Please inform me if I have misunderstood anything. cc @vishavpandhi

vishavpandhi:
Let's say Baseline was created by commit `A` and `Target` was created by commit `B`. This means that `A` is present in the `B` git commit history, so upgrade is possible. 
If `Baseline `and `Target `aren't same, then upgrade is possible. That's why we look in the machine pool, to find a machine which has commit `A`, so that we can upgrade it to `Target`.
If `Baseline `and `Target ` are the same, then upgrade is not possible, as it's the same VS version. So, there is no point in looking for a machine with commit `A`. In this case, we install VS from scratch.
This is why VS is installing on this machine in this [run](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7897328&view=logs&j=0a108fa2-1091-581d-dd54-eafaa15161a9&t=09932453-0ca4-55db-1d1a-f7920de8e48b&l=30)

In case when B2B upgrade is not possible (including `rel `branches), `B2B upgrade` will be disabled, and pipeline will continue with VS installing from scratch.
Are you seeing this failure anywhere? 
Can you send the link?

kartheekp-ms:
> Are you seeing this failure anywhere?
> Can you send the link?

Currently VS tests are executed against [main](https://github.com/NuGet/NuGet.Client/blob/c7cd11ae0afd9c69638587dd5e86826983b3fa21/build/config.props#L30) branch and hence there was no failure as of now. I asked this question because we don't DartLab to fail when we want to pin the VS version to a release branch for tests in the future.

vishavpandhi:
> > Are you seeing this failure anywhere?
> > Can you send the link?
> 
> Currently VS tests are executed against [main](https://github.com/NuGet/NuGet.Client/blob/c7cd11ae0afd9c69638587dd5e86826983b3fa21/build/config.props#L30) branch and hence there was no failure as of now. I asked this question because we don't DartLab to fail when we want to pin the VS version to a release branch for tests in the future.

This issue is already fixed. If you run this pipeline again against a `rel` branch, it should succeed else let me know and I'll take a look at it.

kartheekp-ms:
Team Triage: Merge this PR and look for wall clock improvements in deploy machine step for Apex and E2E tests.

PR #5226, opened 2023-06-13
Avoid allocating StringBuilder in some cases

### LINKED ISSUE #12664: NuGet: VersionRangeFormatter.GetNormalizedString bypassing StringBuilderCache via use of string.format
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1826374/

drewnoakes:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12664

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1826374/
Potentially contributes to: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1828422

## Description

- When `VersionRange` is `null`, `ToLockFileDependencyGroupString` can return the `Name` property unmodified, without copying via a StringBuilder.
- Make fewer `StringBuilder.Append` calls.
- `FloatRange.ToString` can append to an existing `StringBuilder` rather than allocating a temporary string, then appending that.
- Pool more `StringBuilder` objects.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Versioning/FloatRange.cs
+++ b/src/NuGet.Core/NuGet.Versioning/FloatRange.cs
@@ -381,47 +381,11 @@ namespace NuGet.Versioning
         /// </summary>
         public override string ToString()
         {
-            var result = string.Empty;
-            switch (_floatBehavior)
-            {
-                case NuGetVersionFloatBehavior.None:
-                    result = MinVersion.ToNormalizedString();
-                    break;
-                case NuGetVersionFloatBehavior.Prerelease:
-                    result = string.Format(VersionFormatter.Instance, "{0:V}-{1}*", MinVersion, _releasePrefix);
-                    break;
-                case NuGetVersionFloatBehavior.Revision:
-                    result = string.Format(CultureInfo.InvariantCulture, "{0}.{1}.{2}.*", MinVersion.Major, MinVersion.Minor, MinVersion.Patch);
-                    break;
-                case NuGetVersionFloatBehavior.Patch:
-                    result = string.Format(CultureInfo.InvariantCulture, "{0}.{1}.*", MinVersion.Major, MinVersion.Minor);
-                    break;
-                case NuGetVersionFloatBehavior.Minor:
-                    result = string.Format(CultureInfo.InvariantCulture, "{0}.*", MinVersion.Major);
-                    break;
-                case NuGetVersionFloatBehavior.Major:
-                    result = "*";
-                    break;
-                case NuGetVersionFloatBehavior.PrereleaseRevision:
-                    result = string.Format(CultureInfo.InvariantCulture, "{0}.{1}.{2}.*-{3}*", MinVersion.Major, MinVersion.Minor, MinVersion.Patch, _releasePrefix);
-                    break;
-                case NuGetVersionFloatBehavior.PrereleasePatch:
-                    result = string.Format(CultureInfo.InvariantCulture, "{0}.{1}.*-{2}*", MinVersion.Major, MinVersion.Minor, _releasePrefix);
-                    break;
-                case NuGetVersionFloatBehavior.PrereleaseMinor:
-                    result = string.Format(CultureInfo.InvariantCulture, "{0}.*-{1}*", MinVersion.Major, _releasePrefix);
-                    break;
-                case NuGetVersionFloatBehavior.PrereleaseMajor:
-                    result = string.Format(CultureInfo.InvariantCulture, "*-{1}*", MinVersion.Major, _releasePrefix);
-                    break;
-                case NuGetVersionFloatBehavior.AbsoluteLatest:
-                    result = "*-*";
-                    break;
-                default:
-                    break;
-            }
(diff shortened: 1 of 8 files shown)


src/NuGet.Core/NuGet.Versioning/FloatRange.cs
```diff
@@ -423,6 +424,51 @@ public override string ToString()
             return result;
         }
 
+        /// <summary>
+        /// Create a floating version string in the format: 1.0.0-alpha-*
+        /// </summary>
+        public void ToString(StringBuilder sb)
```
davkean, line 394: Does this have to be public for LibraryModel to call it?

src/NuGet.Core/NuGet.Versioning/FloatRange.cs
```diff
@@ -423,6 +424,51 @@ public override string ToString()
...
+            switch (_floatBehavior)
+            {
+                case NuGetVersionFloatBehavior.None:
+                    sb.Append(MinVersion.ToNormalizedString());
+                    break;
+                case NuGetVersionFloatBehavior.Prerelease:
+                    sb.AppendFormat(VersionFormatter.Instance, "{0:V}-{1}*", MinVersion, _releasePrefix);
+                    break;
+                case NuGetVersionFloatBehavior.Revision:
+                    sb.AppendFormat(CultureInfo.InvariantCulture, "{0}.{1}.{2}.*", MinVersion.Major, MinVersion.Minor, MinVersion.Patch);
```
davkean, line 405: Wondering aloud if is this is better than just manually writing out the pieces? I'm pretty sure AppendFormat here allocates a stringbuilder under the covers.

src/NuGet.Core/NuGet.Versioning/FloatRange.cs
drewnoakes: Yes. Apparently IVTs aren't a thing in this repo.

src/NuGet.Core/NuGet.Versioning/FloatRange.cs
drewnoakes: Looking at decompiled mscorlib, it looks like `StringBuilder.AppendFormat` will only allocate another `StringBuilder` if you use a custom format string (i.e. `{0:#,##0.##}`).

src/NuGet.Core/NuGet.Versioning/VersionRangeFormatter.cs
```diff
@@ -185,7 +185,11 @@ private string GetNormalizedString(VersionRange range)
 
             sb.Append(range.HasUpperBound && range.IsMaxInclusive ? ']' : ')');
 
-            return sb.ToString();
+            string result = sb.ToString();
+
+            StringBuilderPool.Shared.Return(sb);
+
+            return result;
```
drewnoakes: This pattern is very common. What would folks say about changing `Return` to also do the `ToString` so we end up with this instead?

```suggestion
            return StringBuilderPool.Shared.Return(sb);
```

The method could be renamed to `ToStringAndReturn` or something if that'd be clearer. Looking through the code, all instances of `Return` could be changed to this. There's one in a `finally` block, which I think could be inlined (which would help perf a little too.)

davkean: Yep sounds good, I probably added this in the first place?

drewnoakes: Yes, you added it in 680f9bd4e97db7cd7482584276886764de69d3cb.

I pushed this change in an update.

src/NuGet.Core/NuGet.Versioning/FloatRange.cs
nkolev92: Can we avoid duplicating the tostring logic?

For example, I think we could have the `ToString()` method call this one with it's own stringbuilder.

nkolev92 (commented):
Looks good. Just a suggestion to reduce duplication.

src/NuGet.Core/NuGet.Versioning/FloatRange.cs
drewnoakes: My hunch while writing this originally was that doing so would increase allocations in the `ToString()` method. But looking now inside mscorlib I see that `string.Format` acquires a pooled `StringBuilder` to perform the format. So we can do the same on our side and reduce the duplication. Update pushed.

drewnoakes:
~~Looks like I can't kick CI by toggling labels.~~ I'll push to the upstream repo in future to avoid this.

drewnoakes:
> Can we avoid duplicating the tostring logic?

Done.

nkolev92 (approved):

nkolev92:
Seems like the CI label worked?

PR #5227, opened 2023-06-13
Avoid string.Split allocations in LockFileFormat.ReadTargetLibrary

### LINKED ISSUE #12663: NuGet: LockFileFormat.ReadTargetLibrary using string.split on a simple pattern
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1826489

drewnoakes:
## Bug
Fixes: https://github.com/NuGet/Home/issues/12663
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1826489

## Description

The previous code would allocate a `char[]`, a `string[]` and the returned `string`s.

This new code will only allocate the strings.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileFormat.cs
+++ b/src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileFormat.cs
@@ -509,11 +509,17 @@ namespace NuGet.ProjectModel
         {
             var library = new LockFileTargetLibrary();
 
-            var parts = property.Split(new[] { '/' }, 2);
-            library.Name = parts[0];
-            if (parts.Length == 2)
+#pragma warning disable CA1307 // Specify StringComparison
+            int slashIndex = property.IndexOf('/');
+#pragma warning restore CA1307 // Specify StringComparison
+            if (slashIndex == -1)
             {
-                library.Version = NuGetVersion.Parse(parts[1]);
+                library.Name = property;
+            }
+            else
+            {
+                library.Name = property.Substring(0, slashIndex);
+                library.Version = NuGetVersion.Parse(property.Substring(slashIndex + 1));
             }
 
             var jObject = json as JObject;


drewnoakes:
Could do slightly better here is `NuGetVersion.Parse` took a `ReadOnlySpan<char>`.

src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileFormat.cs
```diff
@@ -509,11 +509,17 @@ private static LockFileTargetLibrary ReadTargetLibrary(string property, JToken j
         {
             var library = new LockFileTargetLibrary();
 
-            var parts = property.Split(new[] { '/' }, 2);
-            library.Name = parts[0];
-            if (parts.Length == 2)
+#pragma warning disable CA1307 // Specify StringComparison
+            int slashIndex = property.IndexOf('/');
```
davkean, line 513: Why did you choose to suppress rather than fix the warning?

davkean (approved):

src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileFormat.cs
drewnoakes: There's no fix that works for all target frameworks.

Options were:

- different code by tfm
- use an overload that accepts string rather than char
- suppress warning 

I chose the last option as I don't think it applies to this character and the other options have downsides.

davkean: Indeed, looks like they only added the overload for char in .NET Core. Note, this isn't just about the character being searched for, its what method they call underneath in response to the comparison. In this case, the default for `IndexOf` is ordinal (`Contains` is CurrentCulture!) so the behavior and perf is okay.

nkolev92 (approved):

PR #5229, opened 2023-06-13
Package Source Mapping VS Options dialog automatically scrolls to the package selected in PM UI Details Pane

### LINKED ISSUE #12608: VS Options shortcut from PMUI for PackageSourceMappings doesn't scroll to or select the Package
### NuGet Product(s) Affected

### Current Behavior

The shortcut (`Configure` button) in PM UI for Package Source Mapping settings opens the VS Options dialog. The default behavior is to show the list scrolled to the top with nothing selected.

### Desired Behavior

Would be handy if the dialog scrolled down to and selected the Package in the VS Options PackageSourceMapping dialog for the package that was selected in the PMUI.

### Additional Context

_No response_

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12608

## Description

- MEF used to share state of selected package between PM UI and VS Options
  - VS Options page looks for a new type, `INuGetUIOptionsContext`, which may have a selected package ID (from PM UI).
- `SearchTree` can now return the matched pattern for a given search term. (Previously it only returned Package Sources).
- Selected package ID is used to search for the first matching pattern (eg., `Microsoft.Build` may find pattern `Microsoft.*`).
- PackageSourceMapping page's `ListBox` selects and scrolls to the found pattern

### PM UI Details Pane -> `Configure` button
Exact pattern matched:
image

Globbing pattern matched:
image

### Tools -> Options (unchanged)
No selected package ID:
image

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added - Also manually tested Tools -> Options still works without a service provided & no selected package ID.
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [x] Documentation PR or issue filled https://github.com/NuGet/docs.microsoft.com-nuget/issues/3063
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/UserInterfaceService/NuGetUI.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/UserInterfaceService/NuGetUI.cs
@@ -9,6 +9,7 @@ using System.Threading.Tasks;
 using System.Windows;
 using Microsoft;
 using Microsoft.ServiceHub.Framework;
+using Microsoft.VisualStudio.ComponentModelHost;
 using Microsoft.VisualStudio.Shell;
 using NuGet.Commands;
 using NuGet.Common;
@@ -281,7 +282,7 @@ namespace NuGet.PackageManagement.UI
                 {
                     if (optionsPageToOpen == OptionsPage.PackageSourceMapping)
                     {
-                        UpdateSelectedPackageInNuGetUIOptionsContextService();
+                        SetSelectedPackageInNuGetUIOptionsContextService();
                     }
 
                     UIContext.OptionsPageActivator.ActivatePage(optionsPageToOpen, null);
@@ -293,13 +294,13 @@ namespace NuGet.PackageManagement.UI
             }
         }
 
-        private void UpdateSelectedPackageInNuGetUIOptionsContextService()
+        private void SetSelectedPackageInNuGetUIOptionsContextService()
         {
             NuGetUIThreadHelper.JoinableTaskFactory.RunAsync(async delegate
             {
-                await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
 #pragma warning disable ISB001 // Dispose of proxies
-                var nuGetUIOptionsContext = await UIContext.ServiceBroker.GetProxyAsync<INuGetUIOptionsContext>(NuGetServices.NuGetUIOptionsContextService);
+                IComponentModel componentModelMapping = await ServiceLocator.GetComponentModelAsync();
+                var nuGetUIOptionsContext = componentModelMapping.GetService<INuGetUIOptionsContext>();
                 nuGetUIOptionsContext.SelectedPackageId = SelectedPackageId;
 #pragma warning restore ISB001 // Dispose of proxies, disposed in disposing event or in ClearSettings
             }).PostOnFailure(nameof(NuGetUI), nameof(LaunchNuGetOptionsDialog));
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/UserInterfaceService/NuGetUI.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/UserInterfaceService/NuGetUI.cs
@@ -25,6 +25,7 @@ using NuGet.Protocol.Core.Types;
 using NuGet.Resolver;
 using NuGet.VisualStudio;
 using NuGet.VisualStudio.Internal.Contracts;
+using NuGet.VisualStudio.Telemetry;
 using StreamJsonRpc;
 using Task = System.Threading.Tasks.Task;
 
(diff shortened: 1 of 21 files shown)


src/NuGet.Clients/NuGet.PackageManagement.UI/UserInterfaceService/NuGetUI.cs
```diff
@@ -306,7 +328,14 @@ public void EndOperation()
 
         public bool ForceRemove { get; set; }
 
-        public PackageIdentity SelectedPackage { get; set; }
+        public string SelectedPackageId
```
donnie-msft, line 364: I don't think this property was being used. However, I prefer that this parent model be responsible for knowing the selected Package ID, so I'm using it in this PR. 
I think future refactoring can be done to use it instead of digging up the Details Pane model to find the same value.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
```diff
@@ -1094,6 +1094,7 @@ internal async Task UpdateDetailPaneAsync(CancellationToken cancellationToken)
                 EmitSearchSelectionTelemetry(selectedItem);
 
                 await _detailModel.SetCurrentPackageAsync(selectedItem, _topPanel.Filter, () => _packageList.SelectedItem);
+                Model.UIController.SelectedPackageId = selectedItem.Id;
```
donnie-msft, line 1100: The main PMUI control knows that a package was changed, so I think the (parent) PMUI model should also store the selected package ID value. 
Previously, we were only passing it to the Details Pane model.

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMapping.cs
```diff
@@ -38,6 +38,13 @@ public IReadOnlyList<string> GetConfiguredPackageSources(string packageId)
             return SearchTree.Value?.GetConfiguredPackageSources(packageId);
         }
 
+#pragma warning disable RS0016 // Add public types and members to the declared API
```
donnie-msft: TODO; remove this, add to public api

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/SearchTree.cs
```diff
@@ -117,10 +129,51 @@ public IReadOnlyList<string> GetConfiguredPackageSources(string term)
...
             if (i == term.Length && currentNode.PackageSources != null)
             {
-                return currentNode.PackageSources;
+                return currentNode;
+            }
+
+            return longestMatchingPrefixNode;
+        }
+
+        private string SearchPatternByTerm(string term)
```
donnie-msft, line 144: Wanted to find matching patterns, and return the pattern. Using a `StringBuilder` and same algorithm as `GetConfiguredPackageSources`. 
/cc @erdembayar - any concerns?

test/NuGet.Core.Tests/NuGet.Configuration.Test/SearchTreeTest.cs
```diff
@@ -222,9 +222,50 @@ public void SearchTree_InvalidSearchInput_Throws(string packagePatterns, string
...
+        [InlineData("   public    ,    nuget    ", "   nuget   ", "nuget")]
+        [InlineData("public,*", "Microsoft.Build", "*")]
+        [InlineData("public,Contoso.Opensource.*,*", "Microsoft.Build", "*")]
+        [InlineData("public,Contoso.Opensource.*", "Contoso.Opensource.", "contoso.opensource.*")]
+        [InlineData("public,Contoso.Opensource.*", "Contoso.Opensource.MVC", "contoso.opensource.*")]
+        [InlineData("public,Contoso.Opensource.*", "Contoso.Opensource.MVC.ASP", "contoso.opensource.*")]
+        [InlineData("public,Contoso.Opensource.* ", "Contoso.Opensource.MVC.ASP", "contoso.opensource.*")]
+        [InlineData(" public,Contoso.Opensource.*", "Contoso.Opensource.MVC.ASP", "contoso.opensource.*")]
+        [InlineData(" public,Contoso.Opensource.* ", " Contoso.Opensource.MVC.ASP ", "contoso.opensource.*")]
+        public void SearchPatternByTerm_ExistingTerm_PatternFound(string packagePatterns, string term, string expectedPatternMatch)
```
donnie-msft, line 254: Notice expected patterns are all lowercase since `SearchTree` lowercases them. Since these should be case insensitive, I think this is sufficient. The UI still renders the original casing.

jeffkl (approved):

PR #5231, opened 2023-06-14
Create SingleFileProvider

### LINKED ISSUE #12706: Create SingleFileProvider and use it for content files
- https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1824549
- https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1826414
- https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1826487

drewnoakes:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12706
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1824549
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1826414
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1826487

## Description

Previously, `ContentFileUtils.GetContentFileGroup` would construct a `VirtualFileProvider` class, passing only a single string to the constructor. That class performs a bunch of string chopping and multiple collection allocations. It then loops through them, performing a bunch of string operations and string concatenations in order to produce results within the `GetDirectoryContents` method.

This change adds a `SingleFileProvider` implementation of `IFileProvider` that's specialised for this single-file case. It doesn't need to do any string splits, and only retains the original string. When `GetDirectoryContents` is called, it uses `ReadOnlySpan<char>` to avoid allocating temporary strings. It also keeps a singleton instance of the "none found" result, to avoid allocating anything in such cases.

Unit tests have been added that cover both implementations when a single string exists, to ensure behavioral parity between implementations.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


drewnoakes:
Note that the `VirtualFileProvider` class is no longer used. Would you like me to remove it in this PR?

nkolev92:
@drewnoakes 

How do these changes compare to https://github.com/NuGet/NuGet.Client/pull/5216/files? 

There's lots of similarities between these 2, instinctively this one seems likely to allocate less here. 

Either way, I think we probably only need 1 of the 2 changes.

nkolev92 (dismissed):

davkean:
This is more extensive and tackles way more, we should stick with this one.

drewnoakes:
I missed @Erarndt's PR. I picked up another issue in `VirtualFileProvider` and did the work here, only later to discover there was another similar issue that I'd inadvertently addressed here. I didn't know Eric had made a PR for it. It seems we came to pretty similar conclusions about only needing a single file.

> Note that the VirtualFileProvider class is no longer used. Would you like me to remove it in this PR?

@nkolev92 what do you think about this? This PR means the `VirtualFileProvider` class is no longer used anywhere (other than the unit test I added). Should we remove it?

nkolev92:
I'm fine with removing VirtualFileProvider.

drewnoakes:
Pushed an update that removes it. Also extended the test coverage to include some case-insensitive examples.

nkolev92 (dismissed):

jeffkl (dismissed):

src/NuGet.Core/NuGet.Commands/RestoreCommand/ContentFiles/SingleFileProvider.cs
```diff
@@ -0,0 +1,109 @@
...
+    private readonly string _path;
+
+    public SingleFileProvider(string path)
+    {
+        _path = path;
+    }
+
+    public IDirectoryContents GetDirectoryContents(string subpath)
+    {
+        if (subpath.Length == 0)
```
kartheekp-ms: should this be `string.IsNullOrEmpty(subpath)`?

nkolev92:
I was about to merge this, but @kartheekp-ms added some feedback :D 

@drewnoakes This is mergeable as far as I am concerned. Once you've reached a conclusion with @kartheekp-ms you can merge. 

Re-request review if there are any new changes that need looked.

src/NuGet.Core/NuGet.Commands/RestoreCommand/ContentFiles/SingleFileProvider.cs
```diff
@@ -0,0 +1,109 @@
...
+
+    public IFileInfo GetFileInfo(string subpath)
+    {
+        return new VirtualFileInfo(subpath);
+    }
+
+    public IChangeToken Watch(string filter)
+    {
+        return NullChangeToken.Singleton;
+    }
```
kartheekp-ms, line 85: Visual Studio says there are 0 references to these methods. Can we remove this dead code?

src/NuGet.Core/NuGet.Commands/RestoreCommand/ContentFiles/SingleFileProvider.cs
drewnoakes:
```suggestion
        if (string.IsNullOrEmpty(subpath))
```

davkean: Looking at definition and callers, the intention is that this is non-null.

src/NuGet.Core/NuGet.Commands/RestoreCommand/ContentFiles/SingleFileProvider.cs
drewnoakes: These are on the `IFileProvider` interface so cannot be removed without breaking the compilation. That interface is defined and presumably used in `Microsoft.Extensions.FileProviders.Abstractions`.

drewnoakes:
> Re-request review if there are any new changes that need looked.

I pushed one small change in 0228c3fca58feda4f07931559eabcda4956590f4 which dismissed your reviews.

test/NuGet.Core.Tests/NuGet.Commands.Test/SingleFileProviderTests.cs
```diff
@@ -0,0 +1,45 @@
...
+    [InlineData("a/b/file.txt", "ROOT", "a", true)]
+    [InlineData("a/b/file.txt", "ROOT/a", "a/b", true)]
+    [InlineData("a/b/file.txt", "ROOT/a/b", "a/b/file.txt", false)]
+    [InlineData("a/b/file.txt", "a", "a/b", true)]
+    [InlineData("a/b/file.txt", "a/b", "a/b/file.txt", false)]
+    [InlineData("a/b/file.txt", "ROOT/A", "a/b", true)]
+    [InlineData("a/b/file.txt", "ROOT/A/B", "a/b/file.txt", false)]
+    [InlineData("a/b/file.txt", "A", "a/b", true)]
+    [InlineData("a/b/file.txt", "A/B", "a/b/file.txt", false)]
+    public void GetDirectoryContents(string source, string search, string expected, bool isDirectory)
```
kartheekp-ms, line 21: This test passed when I added `[InlineData("a/b/file.txt", "ROOT/", "a", true)]` because we trim `/` character from start of the string. 

However, it failed when I added `[InlineData("a/b/file.txt", "ROOT/a/", "a/b", true)]`, is that expected?

nkolev92: How did the previous implementation behave? Maybe we should've kept the VirtualFileProvider in this PR and then just removed in the next one to make reviewing easier.

src/NuGet.Core/NuGet.Commands/RestoreCommand/ContentFiles/ContentFileUtils.cs
```diff
@@ -129,17 +129,16 @@ internal static class ContentFileUtils
...
+                            fileProvider: new SingleFileProvider(relativePath),
                             fileInfo: rootDirectory,
                             parent: null);
 
                         // Currently Matcher only returns the file name not the full path, each file must be
                         // check individually.
                         var matchResults = matcher.Execute(globbingDirectory);
 
-                        if (matchResults.Files.Any())
+                        if (matchResults.HasMatches)
```
zivkan, line 141: In case you haven't already figured out what this code is doing, is it's matching `contentFiles/*` files in the zip to `<contentFiles>` entries in the package's `nuspec`.  I'm guessing the reason this code is using the file globbing library rather than doing exact string matches, is because it's possible for the nuspec to use a `**/*` (or other) globs.

Therefore, I belive it's possible for one glob in the nuspec to match multiple files in the zip's `contentFiles/` directory. Using NuGet.Insights, I can see there are some packages that indeed use `*` in the nuspec's contentFiles definitions.  Are there any tests to ensure that this still works?

From a quick read, it looks like the `foreach` loop on line 120 is enumerating the files in the zip and checking them one by one against the nuspec (line 105's foreach). So, I guess this PR is safe, since the code indeed checks one file at a time.

I wonder if it would be more efficient to create a single instance of an optimized virtual file provider outside line 105's foreach loop, and that way we can eliminate line 120's foreach loop, since the file globbing library can do its work?

A more impactful perf improvement would be to stop using the file globbing library when the nuspec entry does not contain any `*` or `?` characters.

test/NuGet.Core.Tests/NuGet.Commands.Test/SingleFileProviderTests.cs
drewnoakes: As with all the PRs where I've added unit tests and largely rewritten or replaced a component, I made sure the tests passed on the previous implementation. I'm certain these tests passed on the old code, and that this behaviour is consistent between implementations. If there is a difference, it would be a scenario that wasn't covered by one of the tests, though I was careful to write tests that covered everything I could see. If you want to verify the old behaviour for yourself, you can go back in the commit history here and check out the commit prior to where `VirtualFileProvider` was removed.

src/NuGet.Core/NuGet.Commands/RestoreCommand/ContentFiles/ContentFileUtils.cs
drewnoakes: That's a good observation and would be a nice perf improvement. Should we file a new issue for that, or would you want to see that in this PR too?

test/NuGet.Core.Tests/NuGet.Commands.Test/SingleFileProviderTests.cs
kartheekp-ms: I created a new class with the contents of `VirtualFileProvider` in my local copy of the repo. In the previous implementation, this test failed for `[InlineData("a/b/file.txt", "ROOT/", "a", true)]` input whereas in the new implementation, it passed. I am not sure which behavior is correct.

image

nkolev92:
Similar to https://github.com/NuGet/NuGet.Client/pull/5218, this PR might be a candidate for the next release as well.

I think it's more complicated, thus more risky than some of the already merged PRs.

nkolev92:
Happy to reevaluate, but given the # of changes + our insertions are still blocked + other features that are going into this release, I think we need to minimize the risk.

davkean:
@nkolev92 I'll like the QB to jump in around risk before you start punting - NuGet is the largest cause of allocations by far across the tree, we want these in 17.7.

nkolev92:
@davkean 

There are a bunch of PRs that are non-trivial. 
There's just *a lot* of changes within a short period of time and we haven't had an insertion in almost a week. 

If something does goes wrong, we'd need to able to figure out where it went wrong.

davkean:
Understood, but surely this isn't your last insertion for 17.7?

nkolev92:
No, we have daily insertions. 

We're gonna try to insert what's merged already. [Navigation telemetry for Package Source Mapping in PM UI & VS Option](https://github.com/NuGet/NuGet.Client/commit/b804ec78e6a9760e00352da19381d1b7e1f2838c) is the first change not merged in VS from our dev https://github.com/NuGet/NuGet.Client/commits/dev.

davkean:
So when you marked this as 'Merge next release' what do you mean? Next insertion or 17.8?

nkolev92:
That label is for next minor release, 17.8

src/NuGet.Core/NuGet.Commands/RestoreCommand/ContentFiles/ContentFileUtils.cs
zivkan: new issue is fine. I just wanted to raise awareness since you seem to be spending a lot of time of all these nuget perf improvements. I didn't know if you'd be motivated to do it right away, or leave it as something for the future.

I just want to make sure that nuspecs that use globbing don't break though. I don't have confidence in NuGet's existing unit tests, and Kartheek's comment above is an example of why: https://github.com/NuGet/NuGet.Client/pull/5231#discussion_r1237802635

test/NuGet.Core.Tests/NuGet.Commands.Test/SingleFileProviderTests.cs
drewnoakes: Definitely looks like an interesting case and a gap in the test. The previous implementation called `string.Split` on `/` without discarding empty items, so a `/` at the start or end would create an empty element which would then cause it to fail matching. The new implementation was more tolerant of that.

As an aside, when I was choosing test cases, I ran a bunch of code with a breakpoint in that method, looking for sample values to try and get good coverage. I didn't see any values starting or ending with `/`. But still, the goal here is behavioural parity, so we will fix this case.

I rolled my workspace back to 6438542faf1c8661e6d503bbe6fec086e65d732a, just before removing `VirtualFileProvider` and added some more test cases. These include strings starting with `/`, ending with `/` and containing double `//` characters. A few small tweaks to the `SingleFileProvider` implementation brought it in line with the older implementation.

Here's the diff on that commit, which I'll apply to the HEAD of this PR.
```diff
diff --git a/src/NuGet.Core/NuGet.Commands/RestoreCommand/ContentFiles/SingleFileProvider.cs b/src/NuGet.Core/NuGet.Commands/RestoreCommand/ContentFiles/SingleFileProvider.cs
index a1f9c7308..cb225442b 100644
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/ContentFiles/SingleFileProvider.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/ContentFiles/SingleFileProvider.cs
@@ -33,14 +33,17 @@ public IDirectoryContents GetDirectoryContents(string subpath)

         ReadOnlySpan<char> span = subpath.AsSpan();

+        if (span.EndsWith("/".AsSpan()))
+        {
+            return EnumerableDirectoryContents.Empty;
+        }
+
         // Remove the root identifier from the relative path
         if (span.StartsWith(RootPrefix.AsSpan(), StringComparison.Ordinal))
         {
-            span = span.Slice(RootPrefix.Length);
+            span = span.Slice(RootPrefix.Length).TrimStart('/');
         }

-        span = span.TrimStart('/');
-
         if (_path.AsSpan().StartsWith(span, StringComparison.OrdinalIgnoreCase))
         {
             if (span.Length == _path.Length)
diff --git a/test/NuGet.Core.Tests/NuGet.Commands.Test/FileProviderTests.cs b/test/NuGet.Core.Tests/NuGet.Commands.Test/FileProviderTests.cs
index bc1cc3cb4..73e4c7f5b 100644
--- a/test/NuGet.Core.Tests/NuGet.Commands.Test/FileProviderTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Commands.Test/FileProviderTests.cs
@@ -38,12 +38,19 @@ public void SingleSourceFile(string source, string search, string expected, bool

     [Theory]
     [InlineData("a/b/file.txt", "")]
+    [InlineData("a/b/file.txt", "ROOT/")]
+    [InlineData("a/b/file.txt", "ROOT/a/")]
     [InlineData("a/b/file.txt", "ROOT/z")]
     [InlineData("a/b/file.txt", "ROOT/aa")]
     [InlineData("a/b/file.txt", "ROOT/z/a")]
     [InlineData("a/b/file.txt", "ROOT/a/z")]
     [InlineData("a/b/file.txt", "ROOT/a/b/file")]
     [InlineData("a/b/file.txt", "ROOT/a/b/file.txt")]
+    [InlineData("a/b/file.txt", "/a/b/file.txt")]
+    [InlineData("a/b/file.txt", "/a/")]
+    [InlineData("a/b/file.txt", "/a")]
+    [InlineData("a/b/file.txt", "a/")]
+    [InlineData("a/b/file.txt", "a//b")]
     public void SingleSourceFile_NoMatch(string source, string search)
     {
         var contents = new VirtualFileProvider(new List<string> { source }).GetDirectoryContents(search);
```

drewnoakes: Fixed in 20fe16f4e5be27ec2641ac22ca309fe904adca06.

kartheekp-ms (approved):

drewnoakes:
@nkolev92 please reconsider this for 17.7! It fixes some of the biggest allocators reported by GCPauseWatson.

kartheekp-ms:
IMHO, considering that this pull request includes unit tests, I am comfortable with accepting this fix for version 17.7.

src/NuGet.Core/NuGet.Commands/RestoreCommand/ContentFiles/ContentFileUtils.cs
nkolev92: Can we get a new issue created with that idea?

zivkan:
It contains unit tests for the new class, but [there are no tests on `ContentFileUtils`](https://github.com/search?q=repo%3ANuGet%2FNuGet.Client+ContentFileUtils&type=code) where the new code is being used. Therefore, the new tests aren't going to catch regressions.  Well, there might be other tests that indirectly test ContentFileUtils, but I've been busy on other things, so I haven't searched for such tests.

drewnoakes:
Putting a breakpoint in the new code and running unit tests in the debugger shows a lot of use beyond just the new tests.

src/NuGet.Core/NuGet.Commands/RestoreCommand/ContentFiles/ContentFileUtils.cs
zivkan: * https://github.com/NuGet/Home/issues/12695
* https://github.com/NuGet/Home/issues/12696

zivkan (approved):
NuGet.Commands.Test has a file `ContentFilesTests.cs` with a bunch of tests, including ones with globbing patterns. While most of the tests I looked at didn't actually verify that the nuspec metadata was applied correctly, some did, so luckily there are checks for regressions.

davkean:
Given the insertion has gone through - it would be great to get this included back in 17.7. Looking back over 90 days, these 3 bugs at their worst combined produced 1.4 GB/s.

nkolev92 (approved):

PR #5232, opened 2023-06-14
Reduce allocations in RuntimeGraph

### LINKED ISSUE #12717: Reduce RuntimeGraph allocations as it's immutable
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1826712
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1826702

drewnoakes:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12717
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1826702
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1826712

## Description

- `RuntimeGraph` objects are immutable. The `Clone` operation can just return itself for such values. Similarly, the `Merge` operation can be optimised in cases where either operand is empty.
- Don't initialize the three `ConcurrentDictionary<>` fields on `RuntimeGraph` when empty.
- Share the `ReadOnlyDictionary<>` instances for `Runtimes` and `Supports` properties across all empty instances by ensuring there's only a single instance that's empty, via the `Empty` property and making the parameterless constructor private.
- Remove the `Supports` property setter as this type should be immutable. The property appeared unused.
- Null annotate the class and update the API files accordingly.
- Short circuit several methods when `IsEmpty` is `true`.
- Add code to ensure that `IsEmpty` is checked before trying to dereference cache fields, that are only initialised when not empty. To help constrain these call sites in a more obvious way, the `GetOrAdd`-style callbacks were moved to local functions.
- Convert dictionary key classes to `readonly struct` types to reduce heap allocations.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace NuGet.RuntimeModel
 {
-    public class RuntimeGraph : IEquatable<RuntimeGraph>
+    public sealed class RuntimeGraph : IEquatable<RuntimeGraph>
```
davkean, line 15: Note, this is a breaking change - but given how NuGet objects implement Clone where they assume no derived classes, they should probably be sealed by default.

davkean (dismissed):

drewnoakes:
Found some more wins, and a fix for https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1826702. I pushed an additional commit that invalidates @davkean's review. PR description updated accordingly.

src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/PackageBuilder.cs
```diff
@@ -809,7 +809,7 @@ private static void ValidateFileFrameworks(IEnumerable<IPackageFile> files)
                 set.Add(file);
             }
 
-            var managedCodeConventions = new ManagedCodeConventions(new RuntimeGraph());
+            var managedCodeConventions = new ManagedCodeConventions(RuntimeGraph.Empty);
```
drewnoakes, line 812: We now have a singleton empty instance.

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
```diff
@@ -8,31 +8,33 @@
 using System.Linq;
 using NuGet.Shared;
 
+#nullable enable
```
drewnoakes, line 11: This file has been annotated, resulting in a large diff in the API text files.

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
```diff
@@ -8,31 +8,33 @@
...
 
-        public static readonly RuntimeGraph Empty = new RuntimeGraph();
+        /// <summary>
+        /// Gets a singleton, immutable, empty instance of <see cref="RuntimeGraph"/>.
+        /// </summary>
+        public static readonly RuntimeGraph Empty = new();
 
         public IReadOnlyDictionary<string, RuntimeDescription> Runtimes { get; }
-        public IReadOnlyDictionary<string, CompatibilityProfile> Supports { get; set; }
+        public IReadOnlyDictionary<string, CompatibilityProfile> Supports { get; }
```
drewnoakes: This setter was removed. Note that `Runtimes` didn't have one to remove. These objects are otherwise immutable, and the optimisations here require that immutability. There were no usages of this setter in this repo.

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
```diff
@@ -54,10 +56,21 @@ private RuntimeGraph(Dictionary<string, RuntimeDescription> runtimes, Dictionary
...
         }
 
+        internal bool IsEmpty => Runtimes.Count == 0 && Supports.Count == 0;
+
         public RuntimeGraph Clone()
         {
+            if (IsEmpty)
+            {
+                return this;
+            }
```
drewnoakes, line 86: Optimise clone for empty.

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
```diff
@@ -67,6 +80,16 @@ public RuntimeGraph Clone()
...
         {
+            if (left.IsEmpty)
+            {
+                return right;
+            }
+
+            if (right.IsEmpty)
+            {
+                return left;
+            }
```
drewnoakes: Optimise merge when either is empty.

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
```diff
@@ -110,50 +132,52 @@ public static RuntimeGraph Merge(RuntimeGraph left, RuntimeGraph right)
...
         {
-            return _expandCache.GetOrAdd(runtime, r => new HashSet<string>(ExpandRuntimeInternal(r), StringComparer.Ordinal));
-        }
+            return _expandCache!.GetOrAdd(runtime, r => new HashSet<string>(ExpandRuntimeInternal(r), StringComparer.Ordinal));
 
-        /// <summary>
-        /// Expand runtimes in a BFS walk. This ensures that nearest RIDs are returned first.
-        /// Ordering is important for finding the nearest runtime dependency.
-        /// </summary>
-        private IEnumerable<string> ExpandRuntimeInternal(string runtime)
```
drewnoakes, line 124: This method was converted to a local function, as it should only be called on non-empty instances. All callers of `ExpandRuntimeCached` validate the instance is non-empty before calling. The method is otherwise unchanged.

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
```diff
@@ -173,18 +197,28 @@ public bool AreCompatible(string criteria, string provided)
...
+                return false;
+            }
+
             var key = new RuntimeCompatKey(criteria, provided);
 
-            return _areCompatible.GetOrAdd(key, k => AreCompatibleInternal(k));
-        }
+            return _areCompatible!.GetOrAdd(key, AreCompatibleInternal);
 
-        private bool AreCompatibleInternal(RuntimeCompatKey key)
```
drewnoakes, line 180: Ditto, local function.

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
```diff
@@ -173,18 +197,28 @@ public bool AreCompatible(string criteria, string provided)
...
+                return ExpandRuntimeCached(key.RuntimeName).Contains(key.Other);
+            }
         }
 
         public IEnumerable<RuntimePackageDependency> FindRuntimeDependencies(string runtimeName, string packageId)
         {
+            if (IsEmpty)
+            {
+                return Enumerable.Empty<RuntimePackageDependency>();
+            }
```
drewnoakes, line 259: Optimise when empty.

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
```diff
@@ -173,18 +197,28 @@ public bool AreCompatible(string criteria, string provided)
                 return true;
             }
 
+            if (IsEmpty)
+            {
+                return false;
+            }
```
drewnoakes, line 242: Optimise when empty.

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
```diff
@@ -309,7 +339,7 @@ public static void ReleaseList(List<T> list)
         /// <summary>
         /// RID + package id
         /// </summary>
-        private sealed class RuntimeDependencyKey : IEquatable<RuntimeDependencyKey>
+        private readonly struct RuntimeDependencyKey : IEquatable<RuntimeDependencyKey>
```
drewnoakes, line 381: Convert this private collection key to a struct to reduce heap allocations.

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
```diff
@@ -341,7 +371,7 @@ public override int GetHashCode()
         /// <summary>
         /// RID -> RID compatibility key
         /// </summary>
-        private sealed class RuntimeCompatKey : IEquatable<RuntimeCompatKey>
+        private readonly struct RuntimeCompatKey : IEquatable<RuntimeCompatKey>
```
drewnoakes, line 413: Convert this private collection key to a struct to reduce heap allocations.

drewnoakes (commented):
Highlighting some of the diff here to help reviewers.

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
```diff
@@ -8,31 +8,33 @@
...
-        private HashSet<string> _packagesWithDependencies;
+        private HashSet<string>? _packagesWithDependencies;
 
         public static readonly string RuntimeGraphFileName = "runtime.json";
 
-        public static readonly RuntimeGraph Empty = new RuntimeGraph();
+        /// <summary>
+        /// Gets a singleton, immutable, empty instance of <see cref="RuntimeGraph"/>.
+        /// </summary>
+        public static readonly RuntimeGraph Empty = new();
```
nkolev92, line 32: nit: personally I like not modifying lines for minor stylistic changes. It just makes the diff unnecessarily larger.

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
nkolev92: Unclear to me why https://github.com/NuGet/NuGet.Client/commit/3dc89fbb2b29ca8f05b797d8d49440931e45b348 added a setter. It wasn't used in that PR either.

nkolev92 (dismissed):
LGTM. 

Worth waiting for @zivkan and @jeffkl who are familiar with this code to review as well. 

I don't mind the setter breaking change, but maybe we just revert the sealed change?

drewnoakes:
> maybe we just revert the sealed change

The rationale here is that for any derived class the clone method cannot work correctly. I don't feel strongly either way and can remove `sealed` if preferred.

nkolev92:
> > maybe we just revert the sealed change
> 
> The rationale here is that for any derived class the clone method cannot work correctly. I don't feel strongly either way and can remove `sealed` if preferred.

Don't feel too strongly. 
I don't expect this to cause any problems. 

dotnet/runtime is the only known user of these types I know, https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.NETCore.Platforms/src/GenerateRuntimeGraph.cs.

drewnoakes:
Ah ok, then this PR will break that code. I removed the parameterless `RuntimeGraph` constructor used here:

https://github.com/dotnet/runtime/blob/64cab8fd72b72f5ce708e5866fa4323ddf2613b4/src/libraries/Microsoft.NETCore.Platforms/src/GenerateRuntimeGraph.cs#L152

It's an easy fix (`new RuntimeGraph()` -> `RuntimeGraph.Empty`) but it may still cause binary issues. I'll reinstate it as public.

nkolev92:
I think we should try not to break them. I know bringing the constructor back is gonna come with some drawbacks, but I think that's where we're at right now. 

You can consider marking the constructor as obsolete.

drewnoakes:
I pushed an update that reinstates the constructor and also reduces allocations across the various constructors by reusing empty immutable `ReadOnlyDictionary<>` instances where possible.

> I know bringing the constructor back is gonna come with some drawbacks

With the update the drawback is minimal enough that I didn't mark it as obsolete.

nkolev92 (dismissed):

nkolev92:
There's some compile errors.

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
```diff
@@ -8,98 +8,164 @@
...
-    public class RuntimeGraph : IEquatable<RuntimeGraph>
+    public sealed class RuntimeGraph : IEquatable<RuntimeGraph>
     {
-        private readonly ConcurrentDictionary<RuntimeCompatKey, bool> _areCompatible
-            = new ConcurrentDictionary<RuntimeCompatKey, bool>();
-
-        private readonly ConcurrentDictionary<string, HashSet<string>> _expandCache
-            = new ConcurrentDictionary<string, HashSet<string>>(StringComparer.Ordinal);
+        private static readonly ReadOnlyDictionary<string, RuntimeDescription> EmptyRuntimes = new(new Dictionary<string, RuntimeDescription>());
+        private static readonly ReadOnlyDictionary<string, CompatibilityProfile> EmptySupports = new(new Dictionary<string, CompatibilityProfile>());
```
drewnoakes, line 18: These singleton empty dictionaries are used in constructor chaining (to a new constructor) instead of passing `Enumerable.Empty<>` to the prior constructor which would then construct a new empty dictionary. Actually, it would do that twice, as it then wrapped it in `ReadOnlyDictionary<>`.

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
```diff
@@ -8,98 +8,164 @@
...
 
-        private readonly ConcurrentDictionary<RuntimeDependencyKey, List<RuntimePackageDependency>> _dependencyCache
-            = new ConcurrentDictionary<RuntimeDependencyKey, List<RuntimePackageDependency>>();
+        // These fields are null when IsEmpty is true
+        private readonly ConcurrentDictionary<RuntimeCompatKey, bool>? _areCompatible;
+        private readonly ConcurrentDictionary<string, HashSet<string>>? _expandCache;
+        private readonly ConcurrentDictionary<RuntimeDependencyKey, List<RuntimePackageDependency>>? _dependencyCache;
```
drewnoakes, line 23: These objects are heavy, so they're no longer allocated for empty graphs.

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
```diff
@@ -8,98 +8,164 @@
...
+            // TODO append .NoAllocEnumerate() here once https://github.com/NuGet/NuGet.Client/pull/5246 is available
+            foreach (var pair in left.Runtimes)
+            {
+                runtimes[pair.Key] = pair.Value.Clone();
             }
 
             // Merge the right-side runtimes
-            foreach (var runtime in right.Runtimes.Values)
+            // TODO append .NoAllocEnumerate() here once https://github.com/NuGet/NuGet.Client/pull/5246 is available
+            foreach (var pair in right.Runtimes)
```
drewnoakes: Calling `.Values` on a `Dictionary<>` allocates a `ValueCollection`. Better to just enumerate the key/value pairs together.

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
```diff
@@ -8,98 +8,164 @@
...
+                return right;
+            }
+
+            if (right.IsEmpty)
             {
-                runtimes[runtime.RuntimeIdentifier] = runtime.Clone();
+                return left;
+            }
+
+            var runtimes = new Dictionary<string, RuntimeDescription>(capacity: left.Runtimes.Count + right.Runtimes.Count);
```
drewnoakes, line 127: Specify the size on the dictionary to avoid resizing. In this case we know the exact size required, so use it.

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
```diff
@@ -8,98 +8,164 @@
...
                 {
-                    runtimes[runtime.RuntimeIdentifier] = runtime;
+                    runtimes[pair.Key] = pair.Value;
                 }
             }
 
+            var supports = new Dictionary<string, CompatibilityProfile>(capacity: right.Supports.Count + left.Supports.Count);
+
             // Copy over the right ones
-            var supports = new Dictionary<string, CompatibilityProfile>();
```
drewnoakes, line 92: Specify the size on the dictionary to avoid resizing. In this case it's an upper bound, as `left.Supports` is filtered to only those having restore contexts. If that's uncommon, we could reduce the left's contribution to the capacity, perhaps dividing it in half.

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
```diff
@@ -8,98 +8,164 @@
...
+            // TODO append .NoAllocEnumerate() here once https://github.com/NuGet/NuGet.Client/pull/5246 is available
```
jeffkl: Don't forget about this TODO

jeffkl (changes requested):
Looks good, I'll approve when https://github.com/NuGet/NuGet.Client/pull/5246 is merged and this PR is rebased with the TODO removed

nkolev92:
I'll wait on the other PR for review. You can ping me when the rebase happens.

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
drewnoakes: https://github.com/NuGet/NuGet.Client/pull/5246 has now merged. I've rebased and addressed these TODOs. cc @nkolev92

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
```diff
@@ -197,33 +270,29 @@ public IEnumerable<RuntimePackageDependency> FindRuntimeDependencies(string runt
...
+                    if (Runtimes.TryGetValue(expandedRuntime, out RuntimeDescription? runtimeDescription))
                     {
-                        return dependencySet.Dependencies.Values.AsList();
+                        if (runtimeDescription.RuntimeDependencySets.TryGetValue(key.PackageId, out var dependencySet))
+                        {
+                            return dependencySet.Dependencies.Values.AsList();
+                        }
                     }
                 }
+                return new List<RuntimePackageDependency>();
```
kartheekp-ms, line 291: nit: Can this be `return new List<RuntimePackageDependency>(0);`?

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
kartheekp-ms: Today, I learned about the concept of a readonly struct. According to the documentation, a `readonly struct` can undergo [boxing and unboxing conversions](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/struct#conversions) when it implements an interface. In this case, the `RuntimeCompatKey` is used as a key in a dictionary, and it implements the `IEquatable` interface. Therefore, I would anticipate that there might be some boxing operations involved when comparing keys for equality. Is it acceptable for these allocations to occur due to boxing?

drewnoakes: Structs are only boxed when they're called through the interface.

```c#
int i = 1;
i.CompareTo(2); // call (non boxing and cheap)
IComparable j = i; // box
j.CompareTo(2); // callvirt and boxed argument (expensive)
```

Here's the MSIL for the above code.

```
// i.CompareTo(2)
        IL_0002: ldloca.s 0
        IL_0004: ldc.i4.2
        IL_0005: call instance int32 [mscorlib]System.Int32::CompareTo(int32)
        IL_000a: pop
// IComparable j = i
        IL_000b: ldloc.0
        IL_000c: box [mscorlib]System.Int32
// j.CompareTo(2)
        IL_0011: ldc.i4.2
        IL_0012: box [mscorlib]System.Int32
        IL_0017: callvirt instance int32 [mscorlib]System.IComparable::CompareTo(object)
        IL_001c: pop
        IL_001d: ret
```

In the case of the dictionary, if `TKey` is a struct, the JIT generates non-boxing calls to the `Equals` and `GetHashCode` overridden from `object`.

We could probably remove the `IEquatable<>` interfaces from this type and there'd be no difference in behaviour.

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
drewnoakes: Specifying zero and specifying no capacity are equivalent, although passing zero does do slightly more work as it has to validate the argument and test whether it's zero. Passing no argument is clearer and marginally cheaper.

https://github.com/microsoft/referencesource/blob/51cf7850defa8a17d815b4700b67116e3fa283c2/mscorlib/system/collections/generic/list.cs#L49-L68

nkolev92 (dismissed):

davkean:
> Today, I learned about the concept of a readonly struct. According to the documentation, a readonly struct can undergo [boxing and unboxing conversions](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/struct#conversions) when it implements an interface. In this case, the RuntimeCompatKey is used as a key in a dictionary, and it implements the IEquatable interface. Therefore, I would anticipate that there might be some boxing operations involved when comparing keys for equality. Is it acceptable for these allocations to occur due to boxing?

No, there will be no allocations when comparing keys for equality, the call chain (through `Comparer<T>.Default`) to perform the comparison is generically specialized around 'T', not around the interface which avoids the box.

kartheekp-ms (dismissed):

drewnoakes:
> Looks good, I'll approve when #5246 is merged and this PR is rebased with the TODO removed

@jeffkl the TODO has been addressed. Please take another look or dismiss your review.

drewnoakes:
Merged in `dev` and resolved a few basic conflicts (we no longer have to clone runtimes due to #5249).

nkolev92 (approved):

drewnoakes:
> > Looks good, I'll approve when #5246 is merged and this PR is rebased with the TODO removed
> 
> @jeffkl the TODO has been addressed. Please take another look or dismiss your review.

@jeffkl gentle ping

zivkan (approved):

kartheekp-ms (approved):

jeffkl (approved):

PR #5238, opened 2023-06-15
Disable Install/Update button in Details Pane when a Package Source Mapping is Required

### LINKED ISSUE #12607: Install/Update buttons are enabled in Details Pane when the PackageSourceMapping is not found
### NuGet Product(s) Affected

### Current Behavior

Part of spec: https://github.com/NuGet/Home/blob/dev/proposed/2023/package-source-mapping-status-PMUI.md

Install/Update buttons are enabled in Details Pane when the PackageSourceMapping is not found.
This leads to NU1100 errors from Restore. 
Related: [NU1100 is used for both TFM incompatibility and for an unresolved PackageSourceMapping · Issue #12606 · NuGet/Home (github.com)](https://github.com/NuGet/Home/issues/12606)

### Desired Behavior

Install/Update buttons should be disabled in Details Pane when the PackageSourceMapping is not found

### Additional Context

_No response_

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12607

## Description

Part of spec: https://github.com/NuGet/Home/blob/dev/proposed/2023/package-source-mapping-status-PMUI.md

Introduce `CanInstallWithPackageSourceMapping` which is added to the logic determining whether an Install/Update button should be enabled in the Details Pane.

Editing `nuget.config` outside of VS will update the state of the buttons if PM UI is open in the background.

Adding tests was challenging, as there are so many dependencies on property values. The new property is tested, as well as ensuring that the `PropertyChanged` events are fired as expected.
I decided for the Solution Detail Model tests to just add an explanation of why I asserted some values will remain `False` (because chasing properties to make them be `True` is proving difficult). Ideally, this gives us a starting point for testing these models as we continue working toward an MVVM architecture.

### Examples of Install button state matching Package Source Mapping state:
image
image

### Examples of Update button state matching Package Source Mapping state:
image
image

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [x] Documentation PR or issue filled - https://github.com/NuGet/docs.microsoft.com-nuget/issues/3063
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
@@ -599,7 +599,7 @@ namespace NuGet.PackageManagement.UI
                     // Clear detailed view
                     PackageMetadata = null;
 
-                    if (_selectedVersion != null)
+                    if (_selectedVersion != null && _searchResultPackage != null)
                     {
                         var loadCts = new CancellationTokenSource();
                         var oldCts = Interlocked.Exchange(ref _selectedVersionCancellationTokenSource, loadCts);
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
@@ -121,6 +121,8 @@ namespace NuGet.PackageManagement.UI
             RecommenderVersion = recommenderVersion;
         }
 
+        public abstract void SetInstalledOrUpdateButtonIsEnabled();
+
         /// <summary>
         /// Sets the package to be displayed in the detail control.
         /// </summary>
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
@@ -823,6 +823,8 @@ namespace NuGet.PackageManagement.UI
 
         public PackageSourceMappingActionViewModel PackageSourceMappingViewModel { get; }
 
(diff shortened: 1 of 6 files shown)


nkolev92:
I feel like our focus should be to enable clicking install that adds a mapping instead blocking install altogether.

donnie-msft:
> I feel like our focus should be to enable clicking install that adds a mapping instead blocking install altogether.

The [spec ](https://github.com/NuGet/Home/pull/12616) for that is still in Draft mode. Appreciate any input there I can get.

The idea here is to guide customers to the Configuration of source mappings, rather than let them experience a Restore error by pressing "Install".

nkolev92:
@donnie-msft 
Sorry for the delay, I've reviewed your PR. 

Given the delay, I'm totally game to have a sync review of this in order to unblock you.

donnie-msft:
> @donnie-msft 
> Sorry for the delay, I've reviewed your PR. 
> 
> Given the delay, I'm totally game to have a sync review of this in order to unblock you.

Awesome! Could you review this PR in the meantime so we can get it shipped until that spec gets implemented?

martinrrm (approved):
LGTM it's great that it was easy to update the button when there is a change in settings!

nkolev92:
@donnie-msft I don't agree with this change in 17.7 preview 3. I left feedback in the design and I don't think it covered this behavior.

Can you please revert it until we can discuss it?

donnie-msft:
@nkolev92 sure, please set up a meeting with me an @skofman1 to talk about it. Anyone else can join as optional. Thanks

nkolev92:
Talked offline, was pointed to the spec describing the behavior, https://github.com/NuGet/Home/blob/dev/proposed/2023/package-source-mapping-status-PMUI.md.

PR #5243, opened 2023-06-16
Use Microsoft.IO.Redist on .NET Framework

drewnoakes:
## Bug

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1826488
Fixes: https://github.com/NuGet/Client.Engineering/issues/2315

## Description

This package providers lower allocation versions of several Path and Directory methods, including overloads that accept `ReadOnlySpan<char>` in many cases.

This change uses `Microsoft.IO.Path.GetRelativePath` in the `PathUtility.GetRelativePath` method on .NET Framework to help reduce allocations in Visual Studio.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -54,6 +54,7 @@
     <PackageVersion Include="Microsoft.Extensions.FileProviders.Abstractions" Version="$(MicrosoftExtensionsFileProvidersAbstractionsPackageVersion)" />
     <PackageVersion Include="Microsoft.Extensions.FileSystemGlobbing" Version="$(MicrosoftExtensionsFileSystemGlobbingPackageVersion)" />
     <PackageVersion Include="Microsoft.Internal.VisualStudio.Shell.Framework" Version="$(VSFrameworkVersion)" />
+    <PackageVersion Include="Microsoft.IO.Redist" Version="6.0.0" />
     <PackageVersion Include="Microsoft.Net.Compilers.netcore" Version="3.0.0-dev-61717-03" />
     <PackageVersion Include="Microsoft.Net.Compilers.Toolset" Version="3.9.0" />
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.1.0" />
--- a/src/NuGet.Core/NuGet.Common/NuGet.Common.csproj
+++ b/src/NuGet.Core/NuGet.Common/NuGet.Common.csproj
@@ -25,6 +25,10 @@
     <ProjectReference Include="..\NuGet.Frameworks\NuGet.Frameworks.csproj" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
+    <PackageReference Include="Microsoft.IO.Redist" />
+  </ItemGroup>
+
   <ItemGroup>
     <Compile Update="Strings.Designer.cs">
       <DesignTime>True</DesignTime>
--- a/NuGet.sln
+++ b/NuGet.sln
@@ -11,6 +11,7 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{4A4623E9-CE1F-4DF2-A4C2-513CE8377D2A}"
 	ProjectSection(SolutionItems) = preProject
 		.editorconfig = .editorconfig
+		Directory.Packages.props = Directory.Packages.props
 		build\NuGet.ruleset = build\NuGet.ruleset
 	EndProjectSection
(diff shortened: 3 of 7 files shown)


jeffkl (approved):

src/NuGet.Core/NuGet.Common/NuGet.Common.csproj
```diff
@@ -25,6 +25,10 @@
     <ProjectReference Include="..\NuGet.Frameworks\NuGet.Frameworks.csproj" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
+    <PackageReference Include="Microsoft.IO.Redist" />
```
jeffkl, line 29: The code looks good but adding this dependency means we'll need to ILMerge it into `NuGet.exe` and ship it in the NuGet VSIX.

nkolev92: Is there a binding redirect for this in VS? 
If no, then once we add, we might get an RPS failure too.

Given that we're already dealing with some RPS issues, it'd be great to hold on making this change until those are fixed.

drewnoakes: CPS uses this. I believe you should be fine. Perhaps @arkalyanms has some context here.

zivkan: Even if VS has a binding redirect/codebase for the dll, msbuild cli restore doesn't use any VS infrastucture. I see that msbuild's bin directory has the dll, so maybe NuGet doesn't need it, although if msbuild and nuget ever compile against different versions of the dll, I expect there will be errors unless nuget ships its own copy in its own directory.

So, I'm not sure it's "safe" to make this change, without changing nuget's vsix to include microsoft.io.redist.dll.

davkean: This assembly ships in C:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\IDE\PublicAssemblies and should not be included in NuGet's VSIX.

zivkan: MSBuild doesn't load assemblies from VS's PublicAssemblies, so anything used by our MSBuild tasks have to be included in NuGet's VSIX, or command  line (all CI) restores will fail. If the assembly ships in MSBuild's own bin directory, that might be an exception to the rule, but it needs to be tested (and if it works, we also need to test what happens when NuGet is compiled against a different assembly version than what MSBuild ships in its bin directory).

davkean: NuGet is an extension to MSBuild and VS. It should be inheriting dependencies of the dll in both cases. The fact that VS and MSBuild bits are mixed in a single location for NuGet is unusual and not how other features work, so the trickiness here is to pick a dependency that satisifies both VS and MSBuild.

zivkan: It took me longer than I expected it would, but msbuild restore still works with this change. And nuget.exe's ilmerge uses the bin directory's `*.dll`, so it'll get included too.

So, as far as I can tell, potential RPS/DDRITs regressions excluded, this PR probably won't break anything.

Directory.Packages.props
zivkan, line 1: On all the Windows CI agents, but also when building locally, there's a build error:

> src\NuGet.Clients\NuGet.MSSigning.Extensions\NuGet.MSSigning.Extensions.csproj(56,5): Error : Build dependencies are inconsistent with mergeinclude specified in ilmerge.props

So, NuGet.MSSigning.Extensions.csproj needs to be fixed somehow. Unfortunately ilmerge doesn't seem to provide a very useful error message

jeffkl:
@drewnoakes do you want to update this PR and address Andy's comment about NuGet.MSSigning.Extensions?

drewnoakes:
> @drewnoakes do you want to update this PR and address Andy's comment about NuGet.MSSigning.Extensions?

I don't have much context on the infra of this repo and have moved on to other work since the Perfathon wrapped up. If there's interest from the team for this change then feel free to push to this branch.

zivkan:
FWIW, the build failure also reproduced locally, it's not a CI-only issue. `msbuild nuget.sln`.

PR #5246, opened 2023-06-16
Enable non-allocating enumeration via collection interfaces

### LINKED ISSUE #12679: Add NoAllocEnumerateExtensions

drewnoakes:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12679

## Description

Interfaces like `IEnumerable<T>`, `IList<T>` and `IDictionary<TKey, TValue>` are used in many APIs. The upside of this is that the implementation of these APIs may change the concrete type of the collection as it sees fit without breaking consumers. The downside of this is that some operations through the interface are more expensive than if they were performed directly on the concrete type. Enumeration is perhaps the biggest challege here, as enumerating through these interfaces with `foreach` always requires a heap allocation.

Most concrete implementations these implementations will be types such as `List<T>` and `Dictionary<TKey,TValue>` which have a non-allocating struct enumerators.

Specifically:

```c#
List<int> list = new();
IList<int> ilist = list;

foreach (int i in list) {} // does not allocate
foreach (int i in ilist) {} // ❌ allocates a boxed enumerator
```

This change adds `NoAllocEnumerate()` extension methods that avoid any enumerator allocation when the concrete collection type is known to have such a struct enumerator at runtime.

```c#
List<int> list = new();
IList<int> ilist = list;

foreach (int i in list) {} // does not allocate
foreach (int i in ilist.NoAllocEnumerate()) {} // ✅ does not allocate
```

Note that this pattern can be extended to other commonly enumerated types that have struct enumerators. Here it's specialized for `IEnumerable<T>`, `IList<T>` and `IDictionary<TKey,TValue>` in order to limit the number of struct enumerators needed on the structs themselves. It seems like NuGet.Client mostly uses `List<T>` and `Dictionary<TKey,TValue>` (as opposed to other collections like `ImmutableArray<T>`, `ImmutableList<T>`, `ImmutableDictionary<TKey,TValue>`, etc.). The list of supported types can be extended as investigations and investment in this area continue.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/build/Shared/NoAllocEnumerateExtensions.cs
+++ b/build/Shared/NoAllocEnumerateExtensions.cs
@@ -1,0 +1,407 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+#nullable enable
+
+using System;
+using System.Collections.Generic;
+
+namespace NuGet;
+
+internal static class NoAllocEnumerateExtensions
+{
+    #region IList
+
+    /// <summary>
+    /// Avoids allocating an enumerator when enumerating an <see cref="IList{T}"/> where the concrete type
+    /// has a well known struct enumerator, such as for <see cref="List{T}"/>.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// Several collection types (e.g. <see cref="List{T}"/>) provide a struct-based enumerator type
+    /// (e.g. <see cref="List{T}.Enumerator"/>) which the compiler can use in <see langword="foreach" /> statements.
+    /// When using a struct-based enumerator, no heap allocation occurs during such enumeration.
+    /// This is in contrast to the interface-based enumerator <see cref="IEnumerator{T}"/> which will
+    /// always be allocated on the heap.
+    /// </para>
+    /// <para>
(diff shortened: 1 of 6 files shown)


build/Shared/NoAllocEnumerateExtensions.cs
```diff
@@ -0,0 +1,407 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+#nullable enable
+
+using System;
+using System.Collections.Generic;
+
+namespace NuGet;
+
+internal static class NoAllocEnumerateExtensions
```
jeffkl, line 11: I wish this sort of thing was built in.  I also really want an analyzer to ensure we keep using this sort of thing as we add more code.

jeffkl (approved):

nkolev92 (approved):

drewnoakes:
Once this is merged in I'll file a series of PRs that close off perfathon issues by using these extension methods.

PR #5249, opened 2023-06-17
Obsolete Clone methods on immutable types

### LINKED ISSUE #12669: Obsolete Clone methods on immutable types
Several NuGet.Client types are immutable yet define `Clone` methods. There is no need to clone them, as the cloned value will be equivalent in every way to the original. The downside of cloning is the extra memory allocation and GC pressure such clones cause.

The clone methods should be marked as `[Obsolete]` to discourage their use and promote better memory efficiency.

drewnoakes:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12669

## Description

These types are immutable. There is no need to clone them, as the cloned value will be equivalent in every way to the original. The downside of cloning is the extra memory allocation and GC pressure such clones cause.

These types are immutable, so their `Clone` methods have been changed to just return the current instance, and they've been marked as `[Obsolete]` to discourage their use and promote better memory efficiency.

Code that calls them has been updated to avoid calling the obsolete methods.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Frameworks/FrameworkRuntimePair.cs
+++ b/src/NuGet.Core/NuGet.Frameworks/FrameworkRuntimePair.cs
@@ -7,7 +7,10 @@ using NuGet.Shared;
 
 namespace NuGet.Frameworks
 {
-    public class FrameworkRuntimePair : IEquatable<FrameworkRuntimePair>, IComparable<FrameworkRuntimePair>
+    /// <remarks>
+    /// Immutable.
+    /// </remarks>
+    public sealed class FrameworkRuntimePair : IEquatable<FrameworkRuntimePair>, IComparable<FrameworkRuntimePair>
     {
         public NuGetFramework Framework { get; }
 
@@ -48,9 +51,10 @@ namespace NuGet.Frameworks
                 RuntimeIdentifier);
         }
 
+        [Obsolete("This type is immutable, so there is no need or point to clone it.")]
         public FrameworkRuntimePair Clone()
         {
-            return new FrameworkRuntimePair(Framework, RuntimeIdentifier);
+            return this;
         }
 
         public int CompareTo(FrameworkRuntimePair? other)
--- a/src/NuGet.Core/NuGet.ProjectModel/TargetFrameworkInformation.cs
+++ b/src/NuGet.Core/NuGet.ProjectModel/TargetFrameworkInformation.cs
@@ -73,7 +73,7 @@ namespace NuGet.ProjectModel
             Imports = new List<NuGetFramework>(cloneFrom.Imports);
             AssetTargetFallback = cloneFrom.AssetTargetFallback;
             Warn = cloneFrom.Warn;
-            DownloadDependencies = CloneList(cloneFrom.DownloadDependencies, static dep => dep.Clone());
+            DownloadDependencies = cloneFrom.DownloadDependencies.ToList();
             CentralPackageVersions = new Dictionary<string, CentralPackageVersion>(cloneFrom.CentralPackageVersions, StringComparer.OrdinalIgnoreCase);
             FrameworkReferences = new HashSet<FrameworkDependency>(cloneFrom.FrameworkReferences);
             RuntimeIdentifierGraphPath = cloneFrom.RuntimeIdentifierGraphPath;
--- a/src/NuGet.Core/NuGet.Versioning/VersionRangeFactory.cs
+++ b/src/NuGet.Core/NuGet.Versioning/VersionRangeFactory.cs
@@ -9,11 +9,10 @@ using System.Linq;
 
 namespace NuGet.Versioning
 {
-    /// <summary>
-    /// Static factory methods for creating version range objects.
-    /// </summary>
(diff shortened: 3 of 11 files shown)


src/NuGet.Core/NuGet.ProjectModel/TargetFrameworkInformation.cs
```diff
@@ -73,7 +73,7 @@ internal TargetFrameworkInformation(TargetFrameworkInformation cloneFrom)
             Imports = new List<NuGetFramework>(cloneFrom.Imports);
             AssetTargetFallback = cloneFrom.AssetTargetFallback;
             Warn = cloneFrom.Warn;
-            DownloadDependencies = CloneList(cloneFrom.DownloadDependencies, static dep => dep.Clone());
+            DownloadDependencies = cloneFrom.DownloadDependencies.ToList();
```
drewnoakes, line 76: Here we still need to clone the mutable collection, but no longer need to clone the elements within it.

src/NuGet.Core/NuGet.Versioning/VersionRangeFactory.cs
```diff
@@ -9,11 +9,10 @@
 
 namespace NuGet.Versioning
 {
-    /// <summary>
-    /// Static factory methods for creating version range objects.
-    /// </summary>
     public partial class VersionRange
     {
+        // Static factory methods for creating version range objects.
+
```
drewnoakes, line 15: This comment only applies to the members in this partial class's part, but by being a doc comment it is displayed in IntelliSense for the whole class, which seems unintended. Moved to a single line comment within the part to prevent that.

drewnoakes:
It would be great to give `CompatibilityProfile.Clone` the same treatment, which would then enable similarly avoiding `RuntimeGraph.Clone`. I didn't include that because `CompatibilityProfile.RestoreContexts` is a mutable `IList<>` of immutable elements. We could back it with a read only list that'd throw at runtime if changes were attempted, and there'd be no issue in this repo. However it's less safe than the changes already in this PR for any external consumers. What do you think?

src/NuGet.Core/NuGet.Frameworks/FrameworkRuntimePair.cs
```diff
@@ -7,7 +7,10 @@
 
 namespace NuGet.Frameworks
 {
-    public class FrameworkRuntimePair : IEquatable<FrameworkRuntimePair>, IComparable<FrameworkRuntimePair>
+    /// <remarks>
+    /// Immutable.
+    /// </remarks>
+    public sealed class FrameworkRuntimePair : IEquatable<FrameworkRuntimePair>, IComparable<FrameworkRuntimePair>
```
nkolev92, line 13: Why is this not being detected as a breaking chnage by the analyzer? I thought https://github.com/NuGet/NuGet.Client/pull/5234 might fix that. 

@zivkan any ideas?

nkolev92 (commented):
I think we have a breaking change that's not being detected here. I know we've made a bunch of breaking changes with sealing classes in other changes, but I think that was usually types that already had breaking changes. 

Not sure why the public API analyzer is not detecting.

What's your thought about removed the sealed off of the types being changed?

nkolev92:
It's also be great to create an issue for this. 

I forgot it in the previous PRs with breakign changes, but we have a label Type:BreakingChange, that we use in our release notes to notify customers.

src/NuGet.Core/NuGet.Frameworks/FrameworkRuntimePair.cs
drewnoakes: There are a lot of things the current public API analyzers don't catch. See https://github.com/dotnet/roslyn-analyzers/issues/3047 for more info. I think our best way of addressing that would be https://github.com/dotnet/roslyn-analyzers/issues/3901.

drewnoakes:
> What's your thought about removed the sealed off of the types being changed?

The benefit is small, but so is the risk. I don't feel strongly either way. Let me know if you'd like to unseal the types here.

> It's also be great to create an issue for this.

https://github.com/NuGet/Home/issues/12669 -- copied from the PR description.

drewnoakes:
> It would be great to give `CompatibilityProfile.Clone` the same treatment, which would then enable similarly avoiding `RuntimeGraph.Clone`. I didn't include that because `CompatibilityProfile.RestoreContexts` is a mutable `IList<>` of immutable elements. We could back it with a read only list that'd throw at runtime if changes were attempted, and there'd be no issue in this repo. However it's less safe than the changes already in this PR for any external consumers. What do you think?

@nkolev92 interested in your thoughts on this too.

nkolev92 (approved):

nkolev92:
Let's keep it as is for now. 

We've made a lot of changes over the last 15-20 PRs.

PR #5251, opened 2023-06-19
Avoid dictionary lookups in loops

### LINKED ISSUE #12666: Avoid value lookup in foreach loop over dictionary's keys
NuGet.Client has several places where the following pattern is seen:

```c#
foreach (var key in dictionary.Keys)
{
    var value = dictionary[key];
    // ...
}
```

This code is slower than it needs to be. It's possible to avoid looking up the value on each iteration of the loop by enumerating both the key and value at the same time:

```c#
foreach (KeyValuePair<TKey, TValue> pair in dictionary)
{
    var key = pair.Key;
    var value = pair.Value;
    // ...
}
```

This will be faster as the lookups are avoided, and also avoids allocating a `Dictionary<>.KeyCollection` class.

Using deconstruction, it's possible to write this more succinctly as:

```c#
foreach ((var key, var value) in dictionary)
{
    // ...
}
```

drewnoakes:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12666

## Description

Several places were looping through a dictionary's keys, then looking up each value by that key within the loop. This is inneficient as both the key and value are availble for enumeration, by enumerating the dictionary directly.

This pattern is particularly troublesome for large collections, and for `ConcurrentDictionary<>` which can incur a lot of overhead from locking when contention exists.

Another gain here is that the `Keys` property lazily allocates a `Dictionary<>.KeyCollection` instance, so these changes avoid that allocation too.

We add a `Deconstruct` extension method to make enumeration of dictionaries a little more ergonomic as well, allowing deconstructing directly into two local variables within the `foreach` statement itself. I had previously added `DeconstructionExtensions` for use in dependencies tree code, and this change moves it to the shared folder for use elsewhere in the solution.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


davkean (dismissed):

nkolev92 (commented):
ExtensibilityTelemetryCollection has a DeconstructionExtensions class as well. Can we remove that too?

drewnoakes:
> DeconstructionExtensions

Done.

nkolev92 (approved):

PR #5252, opened 2023-06-19
Reduce allocations in ContentFileUtils.GetContentFileGroup

### LINKED ISSUE #12668: GetContentFileFolderRelativeToFramework allocates too much
GetContentFileFolderRelativeToFramework allocates too many strings, enumerators and arrays. It can be made to allocate only the output string.

drewnoakes:
## Bug

Fixes https://github.com/NuGet/Home/issues/12668

## Description

This change adds a unit test for `ContentFileUtils.GetContentFileGroup` to ensure consistent behaviour, then rewrites the method to avoids a bunch of temporary string allocations, arrays and enumerators by using `ReadOnlySpan<char>` instead of `string`.

`SuppressAsserts` was added since there are Debug.Assert statements in the code tested by the new unit test, `GetContentFileFolderRelativeToFramework`.
eg,
image

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/test/NuGet.Core.Tests/NuGet.Commands.Test/ContentFileUtilsTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Commands.Test/ContentFileUtilsTests.cs
@@ -1,0 +1,48 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using System.Linq;
+using Xunit;
+
+namespace NuGet.Commands.Test;
+
+public sealed class ContentFileUtilsTests
+{
+    [Theory]
+    [InlineData("contentFiles/cs/net45/config/config.xml", "config/config.xml")]
+    [InlineData("contentFiles/any/any/config/config.xml", "config/config.xml")]
+    [InlineData("", "")]
+    [InlineData("hello", "hello")]
+    [InlineData("a/b/", "a/b/")]
+    [InlineData("a/b/c/", "")]
+    [InlineData("a/b/c", "a/b/c")]
+    [InlineData("///config/config.xml", "config/config.xml")]
+    [InlineData("/a/b/config/config.xml", "config/config.xml")]
+    public void GetContentFileFolderRelativeToFramework(string input, string expected)
+    {
+        using var _ = new SuppressAsserts();
+
+        string actual = ContentFileUtils.GetContentFileFolderRelativeToFramework(input);
+
+        Assert.Equal(expected, actual);
+    }
+
+    private sealed class SuppressAsserts : IDisposable
+    {
+        private readonly TraceListener[] _suppressedListeners;
+
+        public SuppressAsserts()
+        {
+            _suppressedListeners = Trace.Listeners.Cast<TraceListener>().ToArray();
+
+            Trace.Listeners.Clear();
+        }
+
+        public void Dispose()
+        {
+            Trace.Listeners.AddRange(_suppressedListeners);
+        }
+    }
+}
--- a/test/NuGet.Core.Tests/NuGet.Commands.Test/ContentFileUtilsTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Commands.Test/ContentFileUtilsTests.cs
@@ -29,6 +29,11 @@ public sealed class ContentFileUtilsTests
         Assert.Equal(expected, actual);
     }
 
+    /// <summary>
+    /// Suppresses the default trace listeners for a specific duration. Useful to ensure that
+    /// debug asserts don't block unit tests indefinitely by showing modal dialog messages
+    /// about assertion failures.
+    /// </summary>
     private sealed class SuppressAsserts : IDisposable
     {
         private readonly TraceListener[] _suppressedListeners;
--- a/test/NuGet.Core.Tests/NuGet.Commands.Test/ContentFileUtilsTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Commands.Test/ContentFileUtilsTests.cs
@@ -24,7 +24,7 @@ public sealed class ContentFileUtilsTests
     {
         using var _ = new SuppressAsserts();
 
-        string actual = ContentFileUtils.GetContentFileFolderRelativeToFramework(input);
+        string actual = ContentFileUtils.GetContentFileFolderRelativeToFramework(input.AsSpan());
 
         Assert.Equal(expected, actual);
     }
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/ContentFiles/ContentFileUtils.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/ContentFiles/ContentFileUtils.cs
@@ -207,7 +207,7 @@ namespace NuGet.Commands
 
                 if (copyToOutput)
                 {
-                    string destination = null;
+                    string destination;
 
                     if (flatten)
                     {
@@ -217,7 +217,7 @@ namespace NuGet.Commands
                     {
                         // Find path relative to the TxM
                         // Ex: contentFiles/cs/net45/config/config.xml -> config/config.xml
-                        destination = GetContentFileFolderRelativeToFramework(file);
+                        destination = GetContentFileFolderRelativeToFramework(file.AsSpan());
                     }
 
                     if (isPP)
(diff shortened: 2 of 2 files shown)


test/NuGet.Core.Tests/NuGet.Commands.Test/ContentFileUtilsTests.cs
```diff
@@ -0,0 +1,48 @@
...
+    public void GetContentFileFolderRelativeToFramework(string input, string expected)
+    {
+        using var _ = new SuppressAsserts();
+
+        string actual = ContentFileUtils.GetContentFileFolderRelativeToFramework(input.AsSpan());
+
+        Assert.Equal(expected, actual);
+    }
+
+    private sealed class SuppressAsserts : IDisposable
```
donnie-msft, line 37: I'd like to understand why this is necessary. Is there a lot of trace noise from our tests? I'm guessing this is suppressing the 2 ` Debug.Assert` we have in `ContentFileUtils`? 
If so, then:
1. Is it a preferable practice that we not to have the asserts and use a different approach? Are we doing something unusual?
2. If it's ok to have the asserts, then I'd like to have this suppression more globally, somehow.

donnie-msft:
We require a GitHub issue for each PR. I looked at the AzDO issue and I'm not sure how it relates to this fix... is this addressing the `MatcherContext.Match` starting capacity, or is it addressing a flooding of Tracing events, or something else?
Noting the actual problem being solved will help in future searches for similar bugfixes as well. Thanks

test/NuGet.Core.Tests/NuGet.Commands.Test/ContentFileUtilsTests.cs
drewnoakes: Without this the test blocks on a dialog, courtesy of the default trace listeners.

Either option you outline seems good to me. I don't have enough context on why the assert is there. It seems harmless enough, especially as it's debug only.

drewnoakes:
> We require a GitHub issue for each PR. I looked at the AzDO issue and I'm not sure how it relates to this fix... is this addressing the `MatcherContext.Match` starting capacity, or is it addressing a flooding of Tracing events, or something else? Noting the actual problem being solved will help in future searches for similar bugfixes as well. Thanks

I filed https://github.com/NuGet/Home/issues/12668.

That said, the issue doesn't add any context that isn't already in the commit messages or PR description.

donnie-msft:
Thanks, @drewnoakes for creating the issue.

Regarding:

> That said, the issue doesn't add any context that isn't already in the commit messages or PR description.

A great example of context you can add to your issue is the tracing issue you brought up in https://github.com/NuGet/NuGet.Client/pull/5252#discussion_r1234685302.  How did you get signal on it? Have you seen other repos with issues with the default trace listeners? Why are the *default* trace listeners by default causing problems with testing?

drewnoakes:
I see the assert/tracing issue as specific to the PR's implementation, not the original issue that the PR set out to fix. It only became an issue because I wanted to add unit tests and not hit any review blockers on the removal of the assert.

donnie-msft:
> I see the assert/tracing issue as specific to the PR's implementation, not the original issue that the PR set out to fix.

That's fair, but my perspective is this: there's a new custom class to override tracing behavior, which we only have 1 instance of in the repo, and if myself, someone on the team, or a community contributor tries to create a similar test, we're going to wonder what prompted this class. Likely, this test pattern will get copy/pasted without anyone understanding it, but also afraid to stop doing it. So, I'm trying to understand this now during the PR review.

What if you don't disable the tracing AND don't remove the asserts? Is there any problem we can document in the PR (even just one screenshot, error message, stack trace, or a sentence explaining why we have this workaround). 

Thank you for helping with all these perf issues!

test/NuGet.Core.Tests/NuGet.Commands.Test/ContentFileUtilsTests.cs
drewnoakes, line 37:
```suggestion
    /// <summary>
    /// Suppresses the default trace listeners for a specific duration. Useful to ensure that
    /// debug asserts don't block unit tests indefinitely by showing modal dialog messages
    /// about assertion failures.
    /// </summary>
    private sealed class SuppressAsserts : IDisposable
```

drewnoakes:
> What if you don't disable the tracing AND don't remove the asserts? Is there any problem we can document in the PR (even just one screenshot, error message, stack trace, or a sentence explaining why we have this workaround).

This was already called out on the PR above:

> Without this the test blocks on a dialog, courtesy of the default trace listeners.

As for:

> Likely, this test pattern will get copy/pasted

I would hope that if this ends up being used elsewhere then it'd be promoted to a top-level class and re-used rather than copy/pasted. Or maybe on the [third copy](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) :) Regardless it's only a few lines of code.

I added API docs to the class. Does that address your concerns?

donnie-msft:
> This was already called out on the PR above:
> 
> > Without this the test blocks on a dialog, courtesy of the default trace listeners.

It wasn't called out when I originally reviewed. It looks like you added it a few hours ago after I asked https://github.com/NuGet/NuGet.Client/pull/5252/commits/92e5da2395a339941dd2d72f14d8b80982b5f778.

drewnoakes:
I think I see the confusion. You'd asked if we could add an explanation on the PR, which I'd done the day before.

> Is there any problem we can document in the PR (even just one screenshot, error message, stack trace, or a sentence explaining why we have this workaround).

From this comment I assumed you were talking about the discussion here, rather than to the code itself. I cannot add a screenshot to the code.

Regardless, let me know if there's anything else you want on this PR.

donnie-msft:
I ran a build with `SuppressAsserts` removed to see if I could find out for myself what error you're referring to. 
https://github.com/NuGet/NuGet.Client/commit/167621e15213980f5d10fb8c0cd5f3d88a421478

I got a completely green build somehow, so none of the current tests hit the scenario. 
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7956747&view=results

> Without this the test blocks on a dialog, courtesy of the default trace listeners.

I'm trying to see the behavior that `SuppressAsserts` is guarding against. 

> I cannot add a screenshot to the code.

I'm looking for a screenshot of the blocking dialog (or just a name of it, etc) Which dialog are you referring to?

drewnoakes:
Did you run a debug build?

donnie-msft:
> Did you run a debug build?

No, could you do that and then screenshot whatever error you see? Just drop it in the PR description and we're done. Thanks

test/NuGet.Core.Tests/NuGet.Commands.Test/ContentFileUtilsTests.cs
```diff
@@ -0,0 +1,53 @@
...
+    [InlineData("contentFiles/cs/net45/config/config.xml", "config/config.xml")]
+    [InlineData("contentFiles/any/any/config/config.xml", "config/config.xml")]
+    [InlineData("", "")]
+    [InlineData("hello", "hello")]
+    [InlineData("a/b/", "a/b/")]
+    [InlineData("a/b/c/", "")]
+    [InlineData("a/b/c", "a/b/c")]
+    [InlineData("///config/config.xml", "config/config.xml")]
+    [InlineData("/a/b/config/config.xml", "config/config.xml")]
+    public void GetContentFileFolderRelativeToFramework(string input, string expected)
```
nkolev92, line 23: I'm happy we have tests for this now!

nkolev92 (approved):

drewnoakes:
Debug asserts only appear in debug build. They look like this:

https://images.app.goo.gl/w3B3MNFnsZvPGL5m7

donnie-msft:
> Debug asserts only appear in debug build. They look like this:
> 
> https://images.app.goo.gl/w3B3MNFnsZvPGL5m7

Which of our test(s) were you running where you saw a debug assert window appear?

donnie-msft (approved):
Since there was a lot of back and forth, I went ahead and edited the PR description to explain why the assert suppression is needed for this test.

drewnoakes:
I've been OOF today trying to answer your questions from my phone. I think there's enough info there to put it all together yourself.

PR #5259, opened 2023-06-19
Update .NET functional tests to make them more reliable

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2288

## Description

This change is pretty large but only touches test code.  The primary goal was to get the .NET functional tests working on Mac for .NET 8 because they've been crashing intermittently.  So this change includes the following improvements;

* Executions of .NET would return a result and each test would then verify the result.  Instead I've added two methods, `RunDotnetExpectSuccess()` and `RunDotnetExpectFailure()`.  The common code then verifies the outcome of `dotnet` in a uniform way and can include extra diagnostic information.
* Updated `build.proj` to include diagnostic tracing from `vstest.console` if System.Debug is selected for the pipeline
* Got rid of the `waitForExit` parameter on the test utility method `CommandRunner.Run()` because no calls were passing `false` and it makes me nervous that our code could run an EXE but not wait for it to return.
* Renamed the copy/pasted `MSBuildIntegrationTestsFixture` to `DotnetIntegrationTestFixture` in `Dotnet.Integration.Test` because it doesn't run `msbuild` it runs `dotnet`.
* Added a `Semaphore` to `CommandRunner` so parallel tests are only executing commands one at a time.
* Updated the `Dispose()` logic in `DotnetIntegrationTestFixture` because it was swalling an an exception looking for processes to kill.  

**NOTE: This change also enables the dotnet integration tests on Mac again which are currently disabled**

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetIntegrationTestFixture.cs
```diff
@@ -0,0 +1,506 @@
...
+            {
+                Stopwatch stopwatch = Stopwatch.StartNew();
+
+                result = CommandRunner.Run(TestDotnetCli, workingDirectory, args, environmentVariables: finalEnvironmentVariables);
+
+                stopwatch.Stop();
+
+                if (expectSuccess)
+                {
+                    Assert.True(result.ExitCode == 0, $"{TestDotnetCli} {args} should have succeeded but returned exit code {result.ExitCode} after {stopwatch.Elapsed.TotalSeconds:N1}s with the following output:{Environment.NewLine}{result.AllOutput}");
```
nkolev92: Use FluentAssertions to get the full error message.

The Xunit asserts truncate.

jeffkl: Done!

nkolev92 (dismissed):
Thanks for the walk through earlier :D

test/TestUtilities/Test.Utility/CommandRunner.cs
```diff
@@ -7,141 +7,125 @@
 using System.Diagnostics;
 using System.IO;
 using System.Text;
+using System.Threading;
 
 namespace NuGet.Test.Utility
 {
     public class CommandRunner
     {
+        private static readonly SemaphoreSlim Semaphore = new SemaphoreSlim(1, 1);
```
zivkan: Would it be "better" to change the relevant test projects xunit settings to not allow tests in that project to be run in parallel?

Our tests are already so slow, with these integration tests being the slowest of the slow, so it seems a real shame to limit them to 1 test at a time.

Having said that, xunit runs tests within a collection sequentially, and only runs different collections in parallel (each class is its own collection if no collection attribute is explicitly set). However, since all dotnet integration tests use the same collection, in order to get access to the singleton test fixture, all dotnet.integration.tests should already be running sequentially, so this semaphore shouldn't change any behaviour in that test project. I can't remember if NuGet.CommandLine.Tests use a collection, but if it doesn't, then this semaphore will make those tests significantly slower, when my understanding is you're trying to target dotnet.integration.tests.

jeffkl: Excellent question.  What bothers me the most is I never found a smoking gun as to why `dotnet` was crashing on Mac for .NET 8.  I added this as a big hammer to see if it would fix the problem (along with a few other improvements) and things started working better.  I'll push a commit with the semaphore removed and see it works.  Otherwise, I can move the semaphore to the `DotnetIntegrationTestFixture` instead so it doesn't affect other tests.

jeffkl:
I don't want to destabilize our pipeline so I'm holding off on merging this until we branch for 6.7.

donnie-msft (approved):
I ♥️ seeing unnecessary code going away!
Didn't look at every line. Thanks for your PR Description for summarizing all the refactoring and improvements. 
I wonder if the semaphore will reduce any flakiness we've been seeing.

test/TestUtilities/Test.Utility/CommandRunner.cs
```diff
@@ -15,133 +15,105 @@ public class CommandRunner
...
+
+            using var process = new Process
             {
-                WorkingDirectory = Path.GetFullPath(workingDirectory),
-                UseShellExecute = false,
-                CreateNoWindow = true,
-                RedirectStandardOutput = true,
-                RedirectStandardError = true,
-                RedirectStandardInput = inputAction != null
+                EnableRaisingEvents = true,
```
erdembayar, line 30: May I know why need to raise event here? 
What is the experience with or without this one? I believe it's improving code quality, I'm curious.

zivkan (approved):

PR #5260, opened 2023-06-20
Don't allocate as many comparer instances

### LINKED ISSUE #12680: NuGet.Client allocates many instances of comparers
Comparers are almost always immutable and well suited to reuse. A lot of code in NuGet.Client constructs a new instance when needed, then lets the GC clean it up.

Adding static `Instance` members to comparers would allow them to be reused with minimal code churn.

There are some instances where comparers come from a set of values, depending upon the value of an enum (like `VersionComparison`). There can be a singleton for each, and a method to look this up efficiently at runtime when the required behaviour can vary with state.

Where possible, public constructors should be marked obsolete to discourage their futher use. Where possible, they should be made private.

drewnoakes:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12680

## Description

Comparer classes tend not to have any instance state, or to come from a small pool of related comparers when they do. They also tend to be invoked in tight loops, where allocations can add up. This makes them great candidates for reuse.

The changes in this PR look at the most common comparer types and pools them to use singleton instances.

## PR Checklist

- [ ] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Core/NuGet.Versioning/VersionRangeComparer.cs
```diff
@@ -82,7 +98,7 @@ public bool Equals(VersionRangeBase? x, VersionRangeBase? y)
         /// Creates a hash code based on all properties of the range. This follows the rules of the
         /// version comparer when comparing the version bounds.
         /// </summary>
-        public int GetHashCode(VersionRangeBase obj)
+        public int GetHashCode(VersionRangeBase? obj)
```
zivkan: This doesn't look correct to me. ~If you were to (temporarily, as a test) add `net7.0` to NuGet.Versioning's TagetFramework, I would expect the nullability analyzer to warn you that the nullability of this method does not match the inherited signature (object.GetHashCode).~

edit: object.GetHashCode doesn't have parameters 🤦 Still, asking "what colour is a car that doesn't exist" isn't a sensible question, so I don't think that asking for a hash code of a null object is sensible either. no matter what hash code you assign to null, a non-null instance could theoretically have the same hash code, which is why I guess that System.Collection's dictionaries don't allow null keys.

src/NuGet.Core/NuGet.Versioning/VersionComparer.cs
```diff
@@ -80,7 +92,7 @@ public static int Compare(SemanticVersion? version1, SemanticVersion? version2,
         /// <summary>
         /// Gives a hash code based on the normalized version string.
         /// </summary>
-        public int GetHashCode(SemanticVersion version)
+        public int GetHashCode(SemanticVersion? version)
```
zivkan: Same as my comment in VersionRangeComparer. I don't think that GetHashCode methods should allow nulls. When I added null annotations, I didn't mark this as nullable, even though the implementation allows it, because I'd rather API consumers to use better API designs.

src/NuGet.Core/NuGet.Versioning/VersionRangeComparer.cs
drewnoakes: I'm on a phone so can't easily check right now. You can widen but not narrow from the base implementation, so this should be fine.

src/NuGet.Core/NuGet.Versioning/IVersionRangeComparer.cs
```diff
@@ -8,7 +8,7 @@ namespace NuGet.Versioning
     /// <summary>
     /// Version range comparer
     /// </summary>
-    public interface IVersionRangeComparer : IEqualityComparer<VersionRangeBase>
+    public interface IVersionRangeComparer : IEqualityComparer<VersionRangeBase?>
```
drewnoakes: @zivkan this change (and another like it) were the reason I had to make `IEqualityComparer.GetHashCode`'s argument nullable.

There was another change that required this type parameter to be nullable, but I can't find it in a browser. If you check out this branch and remove this one change, you'll find it.

zivkan: I checked out your branch on my devbox, changed `IEqualityComparer<VersionRangeBase?>` back to `IEqualityComparer<VersionRangeBase>`, and it builds without error or warning on my machine.

drewnoakes: Maybe it was the other instance of this that required it. I'll take a look on my PC in a bit, as I'm about to start my work day.

drewnoakes: Yeah it was the other instance, and I copied that here for consistency.

Changing `IVersionComparer` to remove both `?` causes compile errors:

> src\NuGet.Core\NuGet.Versioning\SemanticVersionBase.cs(134,43,134,48): error CS8604: Possible null reference argument for parameter 'y' in 'int IComparer<SemanticVersion>.Compare(SemanticVersion x, SemanticVersion y)'.
>src\NuGet.Core\NuGet.Versioning\SemanticVersionBase.cs(125,42,125,47): error CS8604: Possible null reference argument for parameter 'y' in 'bool IEqualityComparer<SemanticVersion>.Equals(SemanticVersion x, SemanticVersion y)'.

They both handle nulls correctly, so changing the advertised type seems fair to me.

zivkan: My comments about removing the `?` annotation were on `GetHashCode` methods, not `Equals`. It makes sense for `Equals` methods to accept nulls, which is why `IEqualityComparer<T>` marks the parameters as nullable in the `Equals` method: https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.iequalitycomparer-1.equals?view=net-7.0. So, you don't need to mark the type as nullable in the class declaration

drewnoakes: I understand what you're saying. It still didn't compile without that change however.

zivkan: In my PR when I added null annotations to the assembly, I ignored a bunch of warnings, because the .NET Framework is missing annotations and the compiler incorrectly assumes that nulls are not allowed: https://github.com/NuGet/NuGet.Client/pull/5097/files

https://github.com/NuGet/NuGet.Client/blob/4b6a87bfe90d247f961eba41d63dbf418a56fdcc/src/NuGet.Core/NuGet.Versioning/FloatRange.cs#L442-L445

Unsurprisingly, since I added these suppressions in my own PR, I think it would be better for you to add the same suppressions in your PR. This way in the future when we can drop .NET Framework, we're not left with API design bugs.

drewnoakes: I'm happy either way. I'm content with annotating the methods to reflect the values they actually accept, though I don't know why `GetHashCode` accepts a null. It seems like the alternative means every call site needs to know it's fine to suppress the warning, but if that's preferred then that works. Does 3b08080a5e26aac25531bc68d04aedbddd7d12a7 capture what you had in mind?

drewnoakes:
Merged upstream/dev to resolve merge conflicts.

nkolev92 (dismissed):

jeffkl (dismissed):

src/NuGet.Core/NuGet.Packaging/Core/comparers/PackageIdentityComparer.cs
```diff
@@ -48,6 +48,22 @@ public PackageIdentityComparer(IVersionComparer versionComparer)
...
+        internal static PackageIdentityComparer VersionReleaseMetadata { get; } = new PackageIdentityComparer(VersionComparison.VersionReleaseMetadata);
+
+        internal static PackageIdentityComparer Get(VersionComparison versionComparison)
+        {
+            return versionComparison switch
+            {
+                VersionComparison.Default => Default,
+                VersionComparison.Version => new PackageIdentityComparer(VersionComparison.Version),
+                VersionComparison.VersionRelease => new PackageIdentityComparer(VersionComparison.VersionRelease),
+                VersionComparison.VersionReleaseMetadata => new PackageIdentityComparer(VersionComparison.VersionReleaseMetadata),
```
zivkan: These are intended to return the static instances allocated above, right?

```suggestion
                VersionComparison.Version => Version,
                VersionComparison.VersionRelease => VersionRelease,
                VersionComparison.VersionReleaseMetadata => VersionReleaseMetadata,
```

src/NuGet.Core/NuGet.Versioning/PublicAPI.Shipped.txt
```diff
@@ -62,7 +62,7 @@ NuGet.Versioning.SemanticVersionConverter.SemanticVersionConverter() -> void
 NuGet.Versioning.VersionComparer
 NuGet.Versioning.VersionComparer.Compare(NuGet.Versioning.SemanticVersion? x, NuGet.Versioning.SemanticVersion? y) -> int
 NuGet.Versioning.VersionComparer.Equals(NuGet.Versioning.SemanticVersion? x, NuGet.Versioning.SemanticVersion? y) -> bool
-NuGet.Versioning.VersionComparer.GetHashCode(NuGet.Versioning.SemanticVersion! version) -> int
+NuGet.Versioning.VersionComparer.GetHashCode(NuGet.Versioning.SemanticVersion? version) -> int
```
zivkan: I still think we should prefer API design over implementation and not make this nullable. 0 is a valid hash code for a non-null instance, and I think we should encourage API customers to not request nonsensical values. When I checkout this branch locally and build there are no build errors.

src/NuGet.Core/NuGet.Versioning/PublicAPI.Shipped.txt
```diff
@@ -111,7 +111,7 @@ NuGet.Versioning.VersionRangeBase.Satisfies(NuGet.Versioning.NuGetVersion! versi
 NuGet.Versioning.VersionRangeBase.VersionRangeBase(NuGet.Versioning.NuGetVersion? minVersion = null, bool includeMinVersion = true, NuGet.Versioning.NuGetVersion? maxVersion = null, bool includeMaxVersion = false) -> void
 NuGet.Versioning.VersionRangeComparer
 NuGet.Versioning.VersionRangeComparer.Equals(NuGet.Versioning.VersionRangeBase? x, NuGet.Versioning.VersionRangeBase? y) -> bool
-NuGet.Versioning.VersionRangeComparer.GetHashCode(NuGet.Versioning.VersionRangeBase! obj) -> int
+NuGet.Versioning.VersionRangeComparer.GetHashCode(NuGet.Versioning.VersionRangeBase? obj) -> int
```
zivkan: I still think we should prefer API design over implementation and not make this nullable. 0 is a valid hash code for a non-null instance, and I think we should encourage API customers to not request nonsensical values. When I checkout this branch locally and build there are no build errors.

kartheekp-ms: > I still think we should prefer API design over implementation and not make this nullable. 0 is a valid hash code for a non-null instance.

I agree, but the current implementation already returns `0` when null value is passed. I think the changes proposed in the PR makes it explicit that this value can be null. Please let me know if I am missing some context here.

zivkan: My first sentence, that you quoted, states my position:

> I still think we should prefer API design over implementation and not make this nullable.

We already have a bunch of bad APIs (mostly too many public APIs) which makes some bug fixes and perf improvements difficult or impossible. We've had insertions into the .NET SDK fail in the past when we broke an API we didn't think anyone outside of NuGet would use, but the .NET SDK itself did. Let's not make things worse by annotating a badly designed API as allowing a null when it should have been designed & implemented not to.

Honestly, I think a `GetHashCode(T object)` itself is a bad design. If someone wanted GetHashCode for version without build metadata + GetHashCode without build metadata or prerelease, plus the "everything" GetHashCode, then `NuGetVersion.GetHashCode(HashCodeType)` (or even `GetHashCode(VersionComparer)`) would have been better, and would have been a similar API design to the existing `string.GetHashCode(StringComparer)`. But that's not what shipped 6+ years ago, and who knows how many people we'd break if we changed it now.

Anyway, we only added nullable annotations in 6.7, so they're brand new, not even out in GA yet. I wanted to take the opportunity to add some "API design" into the nullable annotations. I also understand wanting to annotate was was implemented, so don't take API design into consideration for existing APIs we annotate. Overall it's up to the team, I'm just one person. The only reason I didn't approve the PR is because a bunch of comparer allocations were missed in `PackageIdentityComparer.cs`.

src/NuGet.Core/NuGet.Packaging/Core/comparers/PackageIdentityComparer.cs
drewnoakes: Yes, my bad. Thanks for spotting.

src/NuGet.Core/NuGet.Versioning/PublicAPI.Shipped.txt
drewnoakes: I don't feel strongly either way, so let me know what the team decides and we can update. Feel free to push updates here too if you prefer.

I understand the desire to try and tighten up the API design via the annotations.

Equally if a consumer today is passing null, then when they update they'll start getting a warning for code that arguably "works",  which is inconvenient.

kartheekp-ms: Thanks for the detailed explanation, @zivkan. I agree with reverting the `GetHashCode` related changes from this PR.

zivkan (approved):

PR #5267, opened 2023-06-21
Reduce allocations in NuGet.DependencyResolver.Tracker

### LINKED ISSUE #12719: WalkTreeRejectNodesOfRejectedNodes constantly triggering resizes of its tracker collection
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1826391

drewnoakes:
## Bug

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1826391
Fixes: https://github.com/NuGet/Home/issues/12719

## Description

`NuGet.DependencyResolver.Tracker` keeps track of graph items as they are observed. The internal implementation has been optimised to pack data in a more compact fashion, as this type was identified as contributing to GC pauses in VS by GCPauseWatson.

Unit tests were added to document the previous behaviour, then the class was rewritten to use less memory.

Some of the optimisations:

- Don't create an `Entry` for every look up operation. Only create it when some state actually needs to be written (adding a graph item, or marking ambiguous).
- Don't allocate the `HashSet` for every entry in the tracker. Only allocate one once there is more than one item for a given library.
- Don't call `Contains` on the `HashSet` before calling `Add`. If the item already exists, `Add` is a no-op.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.DependencyResolver.Core/GraphModel/Tracker.cs
+++ b/src/NuGet.Core/NuGet.DependencyResolver.Core/GraphModel/Tracker.cs
@@ -33,6 +33,9 @@ namespace NuGet.DependencyResolver
             GetOrAddEntry(item).Ambiguous = true;
         }
 
+        /// <remarks>
+        /// Note, this method returns <see langword="true"/> for items that were never tracked.
+        /// </remarks>
         public bool IsBestVersion(GraphItem<TItem> item)
         {
             var entry = TryGetEntry(item);
--- a/src/NuGet.Core/NuGet.DependencyResolver.Core/GraphModel/Tracker.cs
+++ b/src/NuGet.Core/NuGet.DependencyResolver.Core/GraphModel/Tracker.cs
@@ -3,83 +3,136 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
+
+#nullable enable
 
 namespace NuGet.DependencyResolver
 {
     public class Tracker<TItem>
     {
-        private readonly Dictionary<string, Entry> _entries
-            = new Dictionary<string, Entry>(StringComparer.OrdinalIgnoreCase);
+        private readonly Dictionary<string, Entry> _entryByLibraryName = new(StringComparer.OrdinalIgnoreCase);
 
         public void Track(GraphItem<TItem> item)
         {
-            var entry = GetEntry(item);
-            if (!entry.List.Contains(item))
-            {
-                entry.List.Add(item);
-            }
+            GetOrAddEntry(item).AddItem(item);
         }
 
         public bool IsDisputed(GraphItem<TItem> item)
         {
-            return GetEntry(item).List.Count > 1;
+            return TryGetEntry(item)?.HasMultipleItems ?? false;
         }
 
         public bool IsAmbiguous(GraphItem<TItem> item)
         {
-            return GetEntry(item).Ambiguous;
+            return TryGetEntry(item)?.Ambiguous ?? false;
         }
 
         public void MarkAmbiguous(GraphItem<TItem> item)
         {
-            GetEntry(item).Ambiguous = true;
+            GetOrAddEntry(item).Ambiguous = true;
         }
 
         public bool IsBestVersion(GraphItem<TItem> item)
         {
-            var entry = GetEntry(item);
+            var entry = TryGetEntry(item);
 
-            var version = item.Key.Version;
-
-            foreach (var known in entry.List)
+            if (entry is not null)
             {
-                if (version < known.Key.Version)
+                var version = item.Key.Version;
+
+                foreach (var known in entry.Items)
                 {
-                    return false;
+                    if (version < known.Key.Version)
+                    {
+                        return false;
+                    }
                 }
             }
 
             return true;
         }
 
-        public IEnumerable<GraphItem<TItem>> GetDisputes(GraphItem<TItem> item) => GetEntry(item).List;
+        public IEnumerable<GraphItem<TItem>> GetDisputes(GraphItem<TItem> item)
+        {
+            return TryGetEntry(item)?.Items ?? Enumerable.Empty<GraphItem<TItem>>();
+        }
 
         internal void Clear()
         {
-            _entries.Clear();
+            _entryByLibraryName.Clear();
         }
 
-        private Entry GetEntry(GraphItem<TItem> item)
+        private Entry? TryGetEntry(GraphItem<TItem> item)
         {
-            Entry itemList;
-            if (!_entries.TryGetValue(item.Key.Name, out itemList))
-            {
-                itemList = new Entry();
-                _entries[item.Key.Name] = itemList;
-            }
-            return itemList;
+            _entryByLibraryName.TryGetValue(item.Key.Name, out Entry? entry);
+            return entry;
         }
 
-        private class Entry
+        private Entry GetOrAddEntry(GraphItem<TItem> item)
         {
-            public Entry()
+            if (!_entryByLibraryName.TryGetValue(item.Key.Name, out Entry? entry))
             {
-                List = new HashSet<GraphItem<TItem>>();
+                entry = new Entry();
+                _entryByLibraryName[item.Key.Name] = entry;
             }
 
-            public HashSet<GraphItem<TItem>> List { get; set; }
+            return entry;
+        }
+
+        private sealed class Entry
+        {
+            /// <summary>
+            /// This field can have one of three values:
+            ///
+            /// - null: when no graph items exist in this entry
+            /// - a graph item: when only a single graph item exists in this entry
+            /// - a hash set of graph items: when multiple graph items exist in this entry
+            ///
+            /// This packing exists in order to reduce the size of this object in
+            /// memory. For large graphs, this can amount to a significant saving in memory,
+            /// which helps overall performance.
+            /// </summary>
+            private object? _storage;
 
             public bool Ambiguous { get; set; }
+
+            public void AddItem(GraphItem<TItem> item)
+            {
+                if (_storage is null)
+                {
+                    _storage = item;
+                }
+                else if (_storage is GraphItem<TItem> existingItem)
+                {
+                    if (!existingItem.Equals(item))
+                    {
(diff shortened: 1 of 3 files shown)


drewnoakes:
Pinging @NuGet/nuget-client to make sure this PR doesn't get buried.

src/NuGet.Core/NuGet.DependencyResolver.Core/GraphModel/Tracker.cs
```diff
@@ -3,83 +3,136 @@
...
+                entry = new Entry();
+                _entryByLibraryName[item.Key.Name] = entry;
             }
-            return itemList;
+
+            return entry;
         }
 
-        private class Entry
+        private sealed class Entry
```
zivkan, line 86: Out of curiosity, what does adding `sealed` on a private class help with?

I know in Java all methods are virtual, so `sealed` allows the compiler to call methods directly, rather than looking them up in a v-table, which has a very marginal perf benefit. But in .NET, virtual methods must be marked explicitly. And given this is a nested private class, nothing else can inherit and extend the class.

I'm curious if regularly adding `sealed` to new code I write is worthwhile.

jeffkl (dismissed):

src/NuGet.Core/NuGet.DependencyResolver.Core/GraphModel/Tracker.cs
drewnoakes: I generally seal things by default, more for documentation than performance. I like knowing a type is locked down this way, and I don't have to think about inheritance. Sealing classes can impact perf positively, though more in the modern versions of .NET than in .NET Framework. And I don't think any of the usage of this `Entry` type would benefit from the perf side. Perhaps for the arrays within the dictionary, but that wasn't on my mind when I sealed it. We can revert if you prefer as I don't feel that strongly.

This blog post seems quite good on the perf implications of sealing classes: https://www.meziantou.net/performance-benefits-of-sealed-class.htm

src/NuGet.Core/NuGet.DependencyResolver.Core/GraphModel/Tracker.cs
```diff
@@ -3,83 +3,136 @@
...
+            ///
+            /// This packing exists in order to reduce the size of this object in
+            /// memory. For large graphs, this can amount to a significant saving in memory,
+            /// which helps overall performance.
+            /// </summary>
+            private object? _storage;
+
+            public bool Ambiguous { get; set; }
+
+            public void AddItem(GraphItem<TItem> item)
```
nkolev92, line 103: I know we're testing it indirectly, but wouldn't it be easier to get some times for this class explicitly and this particular method? 

It's where the major change is.

test/NuGet.Core.Tests/NuGet.DependencyResolver.Core.Tests/TrackerTests.cs
```diff
@@ -0,0 +1,188 @@
...
+    public void IsBestVersion()
+    {
+        var itemA1 = MakeItem("A", 1);
+        var itemA2 = MakeItem("A", 2);
+        var itemA3 = MakeItem("A", 3);
+        var itemB1 = MakeItem("B", 1);
+
+        var tracker = new Tracker<string>();
+
+        // NOTE this is strange behavior. An item that was never tracked is reported as the best version.
```
nkolev92, line 133: Isn't this a consequence of the API behaving as a GetOrAdd?

nkolev92 (commented):
LGTM. 

Few small comments.

test/NuGet.Core.Tests/NuGet.DependencyResolver.Core.Tests/TrackerTests.cs
drewnoakes: That's right. That's how the prior code was written, and it's been preserved here. I called it out here in a comment as the test looks strange to me. I'm not sure it's safe to change the behaviour, so left it as-is.

src/NuGet.Core/NuGet.DependencyResolver.Core/GraphModel/Tracker.cs
drewnoakes: I feel like it's thoroughly tested via the API of `Tracker`. This class is really an internal implementation detail of that API, similar to [`Dictionary<,>.Entry`](https://github.com/microsoft/referencesource/blob/51cf7850defa8a17d815b4700b67116e3fa283c2/mscorlib/system/collections/generic/dictionary.cs#L61-L66).

src/NuGet.Core/NuGet.DependencyResolver.Core/GraphModel/Tracker.cs
zivkan: Thanks for the link. I'm surprised how much of a difference (in percentage terms) casting can be, though I highly value composition over inheritance, which minimises the frequency that casting is used. I also don't understand why production code would use interfaces or inheritance if it then just casts to a (sealed) concrete type, which makes it feel more academic than practical to me. But it does make nice benchmark results.

src/NuGet.Core/NuGet.DependencyResolver.Core/GraphModel/Tracker.cs
```diff
@@ -3,83 +3,136 @@
...
+            ///
+            /// - null: when no graph items exist in this entry
+            /// - a graph item: when only a single graph item exists in this entry
+            /// - a hash set of graph items: when multiple graph items exist in this entry
+            ///
+            /// This packing exists in order to reduce the size of this object in
+            /// memory. For large graphs, this can amount to a significant saving in memory,
+            /// which helps overall performance.
+            /// </summary>
+            private object? _storage;
```
zivkan, line 99: Given casting has a perf cost, plus `object` obscures types, is it worthwhile having two fields instead, `private TItem? _single` and `private HashSet<TItem>? _collection` instead?  It gives more compile time static type analysis.

drewnoakes: The goal here was to reduce memory footprint, and a second field would increase size again. Type checking is not free, but it's still very fast. The scenario where this was flagged as contributing to GC, the CPU was spending a huge amount of time on GC. My sense is that reducing GC pressure would offset any overhead from type checking. To write a benchmark you'd need realistic data loads, which would take time to gather. I don't think it'd be worth the effort to investigate given how much low hanging fruit exists throughout NuGet.

test/NuGet.Core.Tests/NuGet.DependencyResolver.Core.Tests/TrackerTests.cs
nkolev92: I think it makes to not change the behavior, but I'm not sure if the comment here adds much. 
IMO, given that the behavior is get or add, would we better off to have a comment there?

drewnoakes: Sure, we could add a comment to `IsBestVersion` too, explaining the strange behaviour. I like tests to read clearly though on their own, hence the comment here. To me this test reads like:

```c#
List<int> numbers = new();
numbers.Contains(1); // true?
```

src/NuGet.Core/NuGet.DependencyResolver.Core/GraphModel/Tracker.cs
davkean: I agree with Drew's assessment here.

test/NuGet.Core.Tests/NuGet.DependencyResolver.Core.Tests/TrackerTests.cs
nkolev92: Sure, I don't mind a version of the test comment if we document the behavior in the class itself. That's where people are most likely to read about the behavior anyway.s

drewnoakes: Pushed an update.

nkolev92 (approved):

src/NuGet.Core/NuGet.DependencyResolver.Core/GraphModel/Tracker.cs
```diff
@@ -3,83 +3,139 @@
...
 
-            public bool Ambiguous { get; set; }
+            public IEnumerable<GraphItem<TItem>> Items
+            {
+                get
+                {
+                    if (_storage is null)
+                        return Enumerable.Empty<GraphItem<TItem>>();
+                    if (_storage is GraphItem<TItem> item)
+                        return new[] { item };
```
kartheekp-ms, line 135: Does this code snippet initialize a collection with a capacity of 1 since it only adds one element to it?

davkean: This returns an array of one element.

kartheekp-ms: Thank you. [docs](https://learn.microsoft.com/dotnet/csharp/programming-guide/arrays/single-dimensional-arrays#array-initialization) clearly states that `The length specifier isn't needed because it's inferred by the number of elements in the initialization list.`

```cs
string[] weekDays = new string[] { "Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat" };
```

kartheekp-ms (approved):

src/NuGet.Core/NuGet.DependencyResolver.Core/GraphModel/Tracker.cs
drewnoakes: Arrays can't differentiate between count and capacity. Collections like List and hash based containers do, where they allocate oversized arrays internally.

jeffkl (approved):

PR #5270, opened 2023-06-22
License Dialog looks at correct sources when Package Source Mapping is enabled

donnie-msft:
## Bug

## Description

When Package Source Mapping is enabled, and licenses are checked against enabled package sources, only the mapped sources will be used rather than all enabled sources.

When a local source (GPF, etc) contains a license, that license is still used, regardless of any package source mappings.

### To test when the package exists on a mapped source
1. I configured nuget.org as a package source
2. When the preview window appeared, I deleted `System.Linq` from my GPF.
3. I debugged and stepped into the License Dialog code to see that it finds nuget.org as a source
image

### To test when the package does not exist on any mapped source
_(Note: this means the GPF somehow had the package, yet the solution has a source mapping to a source where the package doesn't exist)_
1. I created `package source 1`, the source mapped to `System.Linq`. 
2. When the preview window appeared, I deleted `System.Linq` from my GPF to ensure this logic would attempt to get the license from the mapped source.
6. After the license is not obtained from the mapped source `package source 1`, the error below is shown in the error list (the same error which would happen today)
image

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception - See screenshot and explanation above
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
@@ -800,8 +800,8 @@ namespace NuGet.PackageManagement.UI
                 }
             }
 
-            IEnumerable<SourceRepository> sources = _sourceProvider.GetRepositories().Where(e => e.PackageSource.IsEnabled);
-            List<IPackageSearchMetadata> licenseMetadata = await GetPackageMetadataAsync(sources, licenseCheck, token);
+            IEnumerable<SourceRepository> allEnabledSources = _sourceProvider.GetRepositories().Where(e => e.PackageSource.IsEnabled);
+            List<IPackageSearchMetadata> licenseMetadata = licenseMetadata = await GetPackageMetadataAsync(allEnabledSources, licenseCheck, uiService.UIContext.PackageSourceMapping, token);
 
             TelemetryServiceUtility.StopTimer();
 
@@ -1016,6 +1016,7 @@ namespace NuGet.PackageManagement.UI
         private async Task<List<IPackageSearchMetadata>> GetPackageMetadataAsync(
             IEnumerable<SourceRepository> sources,
             IEnumerable<PackageIdentity> packages,
+            PackageSourceMapping packageSourceMapping,
             CancellationToken token)
         {
             var results = new List<IPackageSearchMetadata>();
@@ -1045,14 +1046,27 @@ namespace NuGet.PackageManagement.UI
                 {
                     // get remaining package's metadata from remote repositories
                     var remainingPackages = allPackages.Where(package => package != null && !completed.Any(packageMetadata => packageMetadata != null && packageMetadata.Identity.Equals(package)));
-#pragma warning disable CS8619 // Nullability of reference types in value doesn't match target type.
-                    IPackageSearchMetadata[] remoteResults = (await TaskCombinators.ThrottledAsync(
-                        remainingPackages,
-                        (p, t) => GetPackageMetadataAsync(sources, sourceCacheContext, p, t),
-                        token)).Where(metadata => metadata != null).ToArray();
-#pragma warning restore CS8619 // Nullability of reference types in value doesn't match target type.
+                    IPackageSearchMetadata[] remoteResults;
+                    if (!packageSourceMapping.IsEnabled)
+                    {
+                        remoteResults = await GetPackageMetadataThrottledAsync(sources, sourceCacheContext, remainingPackages, token);
+                        results.AddRange(remoteResults);
+                    }
+                    else // Only look at sources for the package's source mapping.
+                    {
+                        foreach (PackageIdentity package in remainingPackages)
+                        {
+                            IReadOnlyList<string>? sourcesMappedToPackage = packageSourceMapping.GetConfiguredPackageSources(package.Id);
+                            if (sourcesMappedToPackage is null)
+                            {
+                                break;
+                            }
 
-                    results.AddRange(remoteResults);
+                            IEnumerable<SourceRepository> enabledSourcesMappedToPackage = sources.Where(enabledSourceRepository => sourcesMappedToPackage.Contains(enabledSourceRepository.PackageSource.Name, StringComparer.CurrentCultureIgnoreCase));
+                            remoteResults = await GetPackageMetadataThrottledAsync(enabledSourcesMappedToPackage, sourceCacheContext, remainingPackages, token);
+                            results.AddRange(remoteResults);
+                        }
+                    }
                 }
             }
             // check if missing metadata for any package
@@ -1063,6 +1077,18 @@ namespace NuGet.PackageManagement.UI
             }
 
             return results;
+        }
+
+        private static async Task<IPackageSearchMetadata[]> GetPackageMetadataThrottledAsync(IEnumerable<SourceRepository> sources, SourceCacheContext sourceCacheContext, IEnumerable<PackageIdentity> packages, CancellationToken token)
+        {
+#pragma warning disable CS8619 // Nullability of reference types in value doesn't match target type.
+            IPackageSearchMetadata[] remoteResults = (await TaskCombinators.ThrottledAsync(
+                packages,
+                (p, t) => GetPackageMetadataAsync(sources, sourceCacheContext, p, t),
+                token)).Where(metadata => metadata != null).ToArray();
+#pragma warning restore CS8619 // Nullability of reference types in value doesn't match target type.
+
+            return remoteResults;
         }
 
         private static async Task<IPackageSearchMetadata?> GetPackageMetadataAsync(


src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
```diff
@@ -1065,6 +1079,18 @@ private PackageLicenseInfo GeneratePackageLicenseInfo(IPackageSearchMetadata met
...
         }
 
+        private static async Task<IPackageSearchMetadata[]> GetPackageMetadataThrottledAsync(IEnumerable<SourceRepository> sources, SourceCacheContext sourceCacheContext, IEnumerable<PackageIdentity> packages, CancellationToken token)
+        {
+#pragma warning disable CS8619 // Nullability of reference types in value doesn't match target type.
+            IPackageSearchMetadata[] remoteResults = (await TaskCombinators.ThrottledAsync(
+                packages,
+                (p, t) => GetPackageMetadataAsync(sources, sourceCacheContext, p, t),
+                token)).Where(metadata => metadata != null).ToArray();
+#pragma warning restore CS8619 // Nullability of reference types in value doesn't match target type.
```
zivkan, line 1089: There's a lot of code in here where nullability is being disabled. Can we use extra variables, or new lines, and limit the null checking suppression to a single statement, so the rest of this code benefits from null checking?

I'm wondering where the warning is being generated, and if we can either annotate the API so it doesn't generate a warning here, or if there's another way we can change the code so that the compiler's null checking works for us.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
zivkan, line 1: > When a local source (GPF, etc) contains a license, that license is still used, regardless of any package source mappings.

NuGet is designed around the assumption that packages are globally unique and immutable. Therefore, if the package already exists in the GPF, it should have the same licence as the package on the remote server.

If the two locations don't have the same licence, there are frankly a lot of questions that go far beyond just the licence acceptance window (what other content is different between the two packages with the same name and version). Specially on licences, the package in the GPF is going to be used in the local restore & build, and therefore the customer has accepted a different licence than what ends up getting used on their own machine. On the other hand, with the current behaviour, that you're changing in this PR, on other developers machines, or CI agents, the developer machine's local package licence is not being used in other contexts.

Would it be better to detect if the license on the remote server is different to the one in the GPF (only feasible for license expressions and license URLs, not embedded licenses), and then warn the customer when they're different?

Overall, I don't really understand the high level goal for making the licence acceptance window not follow NuGet's globally unique and immutable package axiom. Is there a broader goal that I'm not aware of?

donnie-msft: My change is not intending to change anything about the globally unique axiom. The goal is to avoid querying sources which are not mapped to a package in order to download its license.

The point I made about local sources is that the existing logic is maintained which first checks local sources.

Or are you saying this logic was completely unnecessary prior to this PR?

Let me know if the above doesn't resolve your concern, then I can schedule a meeting next week to dig deeper.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
donnie-msft: This code is just me extracting the existing code to a method. I didn't change anything about nullability. If you think I did, I'll take a look next week.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
donnie-msft: > Would it be better to detect if the license on the remote server is different to the one in the GPF (only feasible for license expressions and license URLs, not embedded licenses), and then warn the customer when they're different?

That's an interesting idea, but I feel out of scope for source mapping. Sounds like we'd have to diff nupkg content? Again, this PR doesn't care about content, it just aims to avoid querying sources which are not source mapped.

donnie-msft: > Overall, I don't really understand the high level goal for making the licence acceptance window not follow NuGet's globally unique and immutable package axiom. Is there a broader goal that I'm not aware of?

If there's a broader goal, it's existed for a while. Do you reckon this logic could have stemmed from packages.config era, since the packages folder hadn't been populated, yet, so licenses were queried first?

I think your comments overall are asking why not delete this logic and rely on the GPF's license content? If that's not right we should chat next week to help me understand your concern better.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
martinrrm: I think it's a good idea to try and fix this code debt, if it's too complex maybe we can create an issue, so we are aware of it.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
zivkan: > I think your comments overall are asking why not delete this logic and rely on the GPF's license content?

Yes, that's right. I think I misunderstood your comment that I quoted. For some reason I thought you were saying this PR is ignoring the GPF package and always getting the licence from the HTTP server, whereas when I read the comment now, it reads like you're reminding reviewers that this is how NuGet works.

But your "why not delete this logic and rely on the GPF license content" sounds like that this PR is indeed ignoring the GPF package.

So now I'm confused again 😆

donnie-msft: > For some reason I thought you were saying this PR is ignoring the GPF package and always getting the licence from the HTTP server

Glad we're on the same page now.
The reason I stated GPF is checked first is that the existing code;
1. Looked at local sources first, if license found,stop looking.
2. If license not obtained from local sources, look at **All enabled HTTP sources**

The key, and the only thing this PR attempts to accomplish, is to not let step 2 look at non-source-mapped sources for a package's license.

I don't know why we have this logic. I'm happy to delete it all later, but right now this is a bug fix. 
The bug is that logic in PM UI ignores source mapping.
My call-out was that I haven't changed the "locals first" algorithm in steps 1&2 above.

nkolev92:
Given that preview will usually install the package into the GPF, will the fix be exercised hit *without* deleting the package from the GPF in the middle of the package installation/update/

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
donnie-msft: What code debt are you referring to?

donnie-msft:
> Given that preview will usually install the package into the GPF, will the fix be exercised hit *without* deleting the package from the GPF in the middle of the package installation/update/

I don't know why we have this logic.The logic is present, so it could be dead code or code that gets hit seldom.  I'm happy to delete it all later, but right now this is a bug fix and out of scope of what I'm trying to do (make VS not stomp all over source mapping).

Is there currently any way to get signal on the impact of removing it? Since this code doesn't have test coverage, there's nothing to point to in a historical context.  I don't think there's any telemetry indicators here, either.

That said, I can start experimenting with removing it and doing a controlled scream test to understand if it serves some purpose. Also, writing tests around bits of it should help.

nkolev92:
I think there's some metadata that's only available on the server, such as vulnerability or deprecation, so maybe that `GetMetadataAsync` call does merging?

Not sure.

donnie-msft:
> I think there's some metadata that's only available on the server, such as vulnerability or deprecation, so maybe that `GetMetadataAsync` call does merging?
> 
> Not sure.
> 

Isn't it a problem to call `GetMetadataAsync` for sources that aren't source mapped? That's what this PR attempts to change.

nkolev92:
The source mapping change is fine, that comment was in reference to the idea of removing it altogether.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
martinrrm: For me when I see a "warning disable" is a sign that something we are doing is not good and at that time we decided that the best was just ignore it, probably this a needed disable to which I prefer using the file ones rather than source like this one.

And about the fix, I had the impression that we agreed that if we saw a code debt we were going to try and fix it if it was easy or create an issue (that probably will end in P3 icebox 😢 )

martinrrm (approved):
I'm not sure what GPF is, but code wise, this looks great.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
zivkan:
```suggestion
            IPackageSearchMetadata[] remoteResults = (await TaskCombinators.ThrottledAsync(
                packages,
                (p, t) => GetPackageMetadataAsync(sources, sourceCacheContext, p, t),
                token)).Where(metadata => metadata != null).Cast<IPackageSearchMetadata>().ToArray();
```

I had a quick look in the IDE. `GetPackageMetadataAsync` returns `IPackageSearchMetadata?`, so the null checker is warning about `IPackageSearchMetadata?[]` being assigned to `IPackageSearchMetadata[]`. Using `.Cast<IPackageSearchMetadata>()` we can remove the "is allowed to be null" annotation.

Another option that might work (I didn't test) is to add our own LINQ extension `IEnumerable<T> WhereNotNull<T>(this IEnumerable<T?> incoming) where T : class { foreach (T? t in incoming) { if (t != default) { yield return t; } } }`

Basically, the way the current code is written, the compiler can't figure out nullability. We have to learn to write code a little differently to let the compiler check for us.

donnie-msft:
superseded by https://github.com/NuGet/NuGet.Client/pull/5288

PR #5272, opened 2023-06-23
[Draft] Preview Window shows the Source Mappings being created for all packages being installed

donnie-msft:
## Bug

This Draft PR has been replaced by PR # 5355

## Description

- If Source Mapping is disabled, nothing changes

_If Source Mapping is enabled..._
- Always calculate new Source Mappings for all **added packages** which do not have an existing source mapping.
  - (ie, the delta from the source mappings in the existing `nuget.config` and the source mappings resulting from this Install's PreviewRestore) 
- Enable Install button even when status is "Requires" mapping 
image
- Works with Quick Install link (From packages list) :
image

- Preview Window, if enabled, will list the new mappings at the bottom of the list.
  - New section heading: "Solution". I chose this since Source Mapping applies to `nuget.config` which is a solution level concept.
    - ie, even in Project level PMUI, the text is "Solution"
image

## TODOs
- Do we need an `Options` (in details pane) to allow customizing whether Install/actions automatically create mappings?
  -  If not, I can remove the boolean to see if the Install button should be disabled due to source mapping (currently hard coded to always `true`)
- Force Preview Window to show (regardless of setting) when there's an Error case or we need Customer's consent:
  - Check `nupkg.metadata` for source for packages in GPF
  - If transitive not available on any source, probably this PR will let it fail back to the Error List. 
    - Possibly: somehow allow customer to modify sources / select a source which does have the missing transitive

## PR Checklist

- [ ] PR has a linked issue.
- [ ] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
@@ -1,6 +1,8 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+#nullable enable
+
 using System.Collections.Generic;
 
 namespace NuGet.PackageManagement.UI
@@ -13,22 +15,22 @@ namespace NuGet.PackageManagement.UI
 
         public IEnumerable<UpdatePreviewResult> Updated { get; }
 
-        public IEnumerable<string> NewSourceMappingsPackageIds { get; }
+        public IDictionary<string, IEnumerable<string>>? NewSourceMappings { get; }
 
-        public string Name { get; }
+        public string? Name { get; }
 
         public PreviewResult(
-            string projectName,
+            string? projectName,
             IEnumerable<AccessiblePackageIdentity> added,
             IEnumerable<AccessiblePackageIdentity> deleted,
             IEnumerable<UpdatePreviewResult> updated,
-            IEnumerable<string> newSourceMappingsPackageIds)
+            IDictionary<string, IEnumerable<string>>? newSourceMappings)
         {
             Name = projectName;
             Added = added;
             Deleted = deleted;
             Updated = updated;
-            NewSourceMappingsPackageIds = newSourceMappingsPackageIds;
+            NewSourceMappings = newSourceMappings;
         }
     }
 }
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
@@ -4,6 +4,7 @@
 #nullable enable
 
 using System.Collections.Generic;
+using System.Linq;
 
 namespace NuGet.PackageManagement.UI
 {
@@ -15,7 +16,7 @@ namespace NuGet.PackageManagement.UI
 
         public IEnumerable<UpdatePreviewResult> Updated { get; }
 
-        public IDictionary<string, IEnumerable<string>>? NewSourceMappings { get; }
+        public Dictionary<string, SortedSet<string>>? NewSourceMappings { get; }
 
         public string? Name { get; }
 
@@ -23,14 +24,21 @@ namespace NuGet.PackageManagement.UI
             string? projectName,
             IEnumerable<AccessiblePackageIdentity> added,
             IEnumerable<AccessiblePackageIdentity> deleted,
-            IEnumerable<UpdatePreviewResult> updated,
-            IDictionary<string, IEnumerable<string>>? newSourceMappings)
+            IEnumerable<UpdatePreviewResult> updated)
         {
             Name = projectName;
             Added = added;
             Deleted = deleted;
             Updated = updated;
+        }
+
+        public PreviewResult(Dictionary<string, SortedSet<string>>? newSourceMappings)
+        {
+            Name = "Solution";
             NewSourceMappings = newSourceMappings;
+            Added = Enumerable.Empty<AccessiblePackageIdentity>();
+            Deleted = Enumerable.Empty<AccessiblePackageIdentity>();
+            Updated = Enumerable.Empty<UpdatePreviewResult>();
         }
     }
 }
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
@@ -16,7 +16,7 @@ namespace NuGet.PackageManagement.UI
 
         public IEnumerable<UpdatePreviewResult> Updated { get; }
 
-        public Dictionary<string, SortedSet<string>>? NewSourceMappings { get; }
+        public IReadOnlyDictionary<string, SortedSet<string>>? NewSourceMappings { get; }
 
         public string? Name { get; }
 
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
@@ -13,7 +13,7 @@ namespace NuGet.PackageManagement.UI
 
         public IEnumerable<UpdatePreviewResult> Updated { get; }
 
-        public IEnumerable<AccessiblePackageIdentity> NewSourceMapping { get; }
+        public IEnumerable<string> NewSourceMappingsPackageIds { get; }
 
         public string Name { get; }
 
@@ -22,13 +22,13 @@ namespace NuGet.PackageManagement.UI
             IEnumerable<AccessiblePackageIdentity> added,
             IEnumerable<AccessiblePackageIdentity> deleted,
             IEnumerable<UpdatePreviewResult> updated,
-            IEnumerable<AccessiblePackageIdentity> newSourceMapping)
+            IEnumerable<string> newSourceMappingsPackageIds)
         {
             Name = projectName;
             Added = added;
             Deleted = deleted;
             Updated = updated;
-            NewSourceMapping = newSourceMapping;
+            NewSourceMappingsPackageIds = newSourceMappingsPackageIds;
         }
     }
 }
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
@@ -13,18 +13,22 @@ namespace NuGet.PackageManagement.UI
 
         public IEnumerable<UpdatePreviewResult> Updated { get; }
 
+        public IEnumerable<AccessiblePackageIdentity> NewSourceMapping { get; }
+
         public string Name { get; }
 
         public PreviewResult(
             string projectName,
             IEnumerable<AccessiblePackageIdentity> added,
             IEnumerable<AccessiblePackageIdentity> deleted,
-            IEnumerable<UpdatePreviewResult> updated)
+            IEnumerable<UpdatePreviewResult> updated,
+            IEnumerable<AccessiblePackageIdentity> newSourceMapping)
         {
             Name = projectName;
             Added = added;
             Deleted = deleted;
             Updated = updated;
+            NewSourceMapping = newSourceMapping;
         }
     }
 }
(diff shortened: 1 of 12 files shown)


src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
```diff
@@ -1007,9 +1012,54 @@ private PackageLicenseInfo GeneratePackageLicenseInfo(IPackageSearchMetadata met
...
+                .ToList();
+
+            if (addedPackagesWithNewSourceMappings.Count == 0)
+            {
+                return;
+            }
+
+            if (newSourceMappings is null)
+            {
+                newSourceMappings = new Dictionary<string, SortedSet<string>>(capacity: 1)
```
donnie-msft, line 1048: TODO:
'Capacity:1' is temporary as right now, only 1 source (selected in PM UI) can be automatically mapped. 

Eventually, any enabled source could be automatically mapped. Details are TBD.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
```diff
@@ -1007,9 +1012,54 @@ private PackageLicenseInfo GeneratePackageLicenseInfo(IPackageSearchMetadata met
...
+            if (newSourceMappings?.Count > 0)
+            {
+                var solutionSourceMappingResult = new PreviewResult(newSourceMappings);
+                results.Add(solutionSourceMappingResult);
+            }
+
             return results;
         }
 
+        private static void GetNewSourceMappingsFromAddedPackages(ref Dictionary<string, SortedSet<string>>? newSourceMappings, UserAction? userAction, List<AccessiblePackageIdentity> added, PackageSourceMapping packageSourceMapping)
```
donnie-msft, line 1024: The ref dictionary is so that Solution level can build up all new mappings across multiple projects.

Probably can rename the method somehow better...

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/DetailControl.xaml.cs
```diff
@@ -110,12 +114,16 @@ private void ProjectInstallButtonClicked(object sender, EventArgs e)
 
             if (model != null && model.SelectedVersion != null)
             {
+                var packageId = model.Id;
```
donnie-msft: Todo: no longer need variable

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
```diff
@@ -1007,9 +1012,54 @@ private PackageLicenseInfo GeneratePackageLicenseInfo(IPackageSearchMetadata met
...
+        {
+            string? newMappingSourceName = userAction?.SourceMappingSourceName;
+            if (newMappingSourceName is null || added.Count == 0 || packageSourceMapping is null)
+            {
+                return;
+            }
+
+            List<string> addedPackagesWithNewSourceMappings = added.Select(_ => _.Id)
```
donnie-msft: The key algorithm of this PR.
If the packages were added by preview restore, but didn't have an existing source mapping, then those mappings are by definition required to be created.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
```diff
@@ -986,7 +990,8 @@ private PackageLicenseInfo GeneratePackageLicenseInfo(IPackageSearchMetadata met
                         deleted.Add(new AccessiblePackageIdentity(uninstalled[packageId]));
                     }
                 }
-
+                // Everything added which didn't already have a source mapping will be mentioned in the Preview Window.
+                GetNewSourceMappingsFromAddedPackages(ref newSourceMappings, userAction, added, uiService.UIContext.PackageSourceMapping);
```
nkolev92, line 994: The source mapping object, is it the update one? Or is this the original-pre operation one?

src/NuGet.Clients/NuGet.PackageManagement.UI/UserAction.cs
```diff
@@ -26,8 +26,8 @@ private UserAction(NuGetProjectActionType action, string packageId, NuGetVersion
             ActiveTab = activeTab;
         }
 
-        private UserAction(NuGetProjectActionType action, string packageId, NuGetVersion? packageVersion, bool isSolutionLevel, ContractsItemFilter activeTab, VersionRange versionRange)
-            : this(action, packageId, packageVersion, isSolutionLevel, activeTab)
+        private UserAction(NuGetProjectActionType action, string packageId, NuGetVersion? packageVersion, bool isSolutionLevel, ContractsItemFilter activeTab, VersionRange versionRange, string? sourceMappingSourceName)
+            : this(action, packageId, packageVersion, isSolutionLevel, activeTab, sourceMappingSourceName)
```
nkolev92: naming: SourceName?

It's basicallyt he source the package got installed from or is it more?

nkolev92 (commented):
When the install button is enable on requires source mapping, should we make it more apparent that installation *will* map packages?

I like that the preview window will list the new mappings. 

Regarding whether or not to create a source mapping when clicking install, IMO we shouldn't have an option. 

A few scenarios:
- the wanted package is mapped and can be installed, 
- The wanted package is unmapped, clicking install attempts to install and map.
- The wanted package is unmapped, then you go manually map things and then see line 1.

Force preview could use a design/something we can do CD on. If I had to guess, I'd say *no* forcing it, but in that case we'd have to fail if we have an ambiguous scenario such as package on disk, but not from right source. 

> If transitive not available on any source, probably this PR will let it fail back to the Error List.

This sounds correct. There will be scenarios that the simple algorithm can't address. We need to track this via telemetry to know how big of a problem this will be.


> Possibly: somehow allow customer to modify sources / select a source which does have the missing transitive

This could end up being a chicken or egg problem. 
Can't lookup a package on a source because of a mapping, but also it'd be helpful if you could know it might be on a source that's unmapped. This is by design as of now.

btw, as you make progress and figure out scenarios, I'd love to see the "rollback" scenarios when installation does not succeed.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
```diff
@@ -13,10 +16,12 @@ public class PreviewResult
 
         public IEnumerable<UpdatePreviewResult> Updated { get; }
 
-        public string Name { get; }
+        public Dictionary<string, SortedSet<string>>? NewSourceMappings { get; }
```
kartheekp-ms: Can we use `IReadOnlyDictionary` type here?

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
```diff
@@ -13,10 +16,12 @@ public class PreviewResult
...
+
+        public string? Name { get; }
 
         public PreviewResult(
-            string projectName,
+            string? projectName,
```
kartheekp-ms, line 24: Curious why we need nullable to this parameter? Looking at the callers they are passing a value to this parameter.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
```diff
@@ -26,5 +31,14 @@ public class PreviewResult
             Deleted = deleted;
             Updated = updated;
         }
+
+        public PreviewResult(Dictionary<string, SortedSet<string>>? newSourceMappings)
+        {
+            Name = "Solution";
```
kartheekp-ms, line 37: does this feature work for solution level PM UI currently?

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
```diff
@@ -936,6 +938,8 @@ private PackageLicenseInfo GeneratePackageLicenseInfo(IPackageSearchMetadata met
             // Group actions by project
             var actionsByProject = expandedActions.GroupBy(action => action.ProjectId);
 
+            Dictionary<string, SortedSet<string>>? newSourceMappings = null;
```
kartheekp-ms, line 941: I assume we are using `SortedSet` type because we need the data in alphabetical order. Correct?

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
kartheekp-ms: may be the better name here would be `addedPackagesWithNoSourceMappings`?

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -91,5 +92,14 @@ internal static class PackageSourceMappingUtility
 
             mappingProvider.SavePackageSourceMappings(newAndExistingPackageSourceMappingItems);
         }
+
+        internal static string? GetNewSourceMappingSourceName(PackageSourceMapping packageSourceMapping, PackageSourceMoniker activePackageSourceMoniker)
+        {
+            string? sourceMappingSourceName = packageSourceMapping.IsEnabled
+                && activePackageSourceMoniker.IsAggregateSource == false
+                ? activePackageSourceMoniker.PackageSourceNames.First() : null;
```
kartheekp-ms, line 100: If I understand this logic correctly, if a package is found in multiple package sources, then we are adding a source mapping entry for the first source, correct?

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
donnie-msft: Yes, but that's not what this means. This string "Solution" means it's affecting the solution by updating the config, in this case, package source mappings.

Btw, I mentioned in pr description that a solution level PM UI preview window also gets these changes.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
donnie-msft: I also needed a simple way to get uniqueness as projects are iterated for their new mapped packages. Imagine a huge solution where the same package is being installed across a dozen projects, with dozens of transitive dependencies each. I figured Set is appropriate, and chose SortedSet since I also need the final rendered list sorted.

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
donnie-msft: This algorithm needs work based on my Todo section.  Basically, think of this as taking your PM UI package source drop-down selected source. Unless it's "All", then I'm returning null. The First() currently is meaningless since there's only 1 selected source possible

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
donnie-msft: It's the original source mappings, because the current design is a provider which is flagged to not save changes to disk during preview restore. 
https://github.com/NuGet/NuGet.Client/blob/77d0b0f78ae21813fc605fef3e2c62f92ae84a7e/src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs#L2916

Then after preview restore, it saves mappings once restore completes:
https://github.com/NuGet/NuGet.Client/blob/77d0b0f78ae21813fc605fef3e2c62f92ae84a7e/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs#L469

src/NuGet.Clients/NuGet.PackageManagement.UI/UserAction.cs
donnie-msft: The source that the UI passed into the action, which is the selected source in the dropdown.
Maybe `selectedSourceName` or `selectedSourceNameForSourceMapping`  ?

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
donnie-msft: Because one of the supplied values, `UniqueName`, can be null:

https://github.com/NuGet/NuGet.Client/blob/77d0b0f78ae21813fc605fef3e2c62f92ae84a7e/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/ProjectMetadataContextInfo.cs#L19

src/NuGet.Clients/NuGet.PackageManagement.UI/UserAction.cs
nkolev92: I like the first one. 2nd seems verbose.

PR #5274, opened 2023-06-23
NuGetAudit telemetry

zivkan:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2058
Fixes: https://github.com/NuGet/Client.Engineering/issues/2068

Regression? No

## Description

Measure restore audit performance

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
@@ -64,6 +64,7 @@ namespace NuGet.Commands
         private const string NoOpRestoreOutputEvaluationDuration = nameof(NoOpRestoreOutputEvaluationDuration);
         private const string NoOpRestoreOutputEvaluationResult = nameof(NoOpRestoreOutputEvaluationResult);
         private const string NoOpReplayLogsDuration = nameof(NoOpReplayLogsDuration);
+        private const string NoOpCacheFileAgeDays = nameof(NoOpCacheFileAgeDays);
 
         // lock file data names
         private const string EvaluateLockFileDuration = nameof(EvaluateLockFileDuration);
@@ -188,7 +189,8 @@ namespace NuGet.Commands
                     {
                         telemetry.StartIntervalMeasure();
                         bool noOp;
-                        (cacheFile, noOp) = EvaluateCacheFile();
+                        TimeSpan? cacheFileAge;
+                        (cacheFile, noOp, cacheFileAge) = EvaluateCacheFile();
                         telemetry.TelemetryEvent[NoOpCacheFileEvaluationResult] = noOp;
                         telemetry.EndIntervalMeasure(NoOpCacheFileEvaluateDuration);
                         if (noOp)
@@ -214,6 +216,7 @@ namespace NuGet.Commands
                                 telemetry.TelemetryEvent[RestoreSuccess] = _success;
                                 telemetry.TelemetryEvent[TotalUniquePackagesCount] = cacheFile.ExpectedPackageFilePaths?.Count ?? -1;
                                 telemetry.TelemetryEvent[NewPackagesInstalledCount] = 0;
+                                if (cacheFileAge.HasValue) { telemetry.TelemetryEvent[NoOpCacheFileAgeDays] = cacheFileAge.Value.TotalDays; }
 
                                 return new NoOpRestoreResult(
                                     _success,
@@ -750,10 +753,11 @@ namespace NuGet.Commands
             return (success, isLockFileValid, packagesLockFile);
         }
 
-        private (CacheFile cacheFile, bool noOp) EvaluateCacheFile()
+        private (CacheFile cacheFile, bool noOp, TimeSpan? cacheFileAge) EvaluateCacheFile()
         {
             CacheFile cacheFile;
             var noOp = false;
+            TimeSpan? cacheFileAge = null;
 
             var noOpDgSpec = NoOpRestoreUtilities.GetNoOpDgSpec(_request);
 
@@ -770,7 +774,7 @@ namespace NuGet.Commands
             // DgSpec doesn't contain log messages, so skip no-op if there are any, as it's not taken into account in the hash
             if (_request.AllowNoOp &&
                 !_request.RestoreForceEvaluate &&
-                File.Exists(_request.Project.RestoreMetadata.CacheFilePath))
+                CacheFileExists(_request.Project.RestoreMetadata.CacheFilePath, out cacheFileAge))
             {
                 cacheFile = FileUtility.SafeRead(_request.Project.RestoreMetadata.CacheFilePath, (stream, path) => CacheFileFormat.Read(stream, _logger, path));
 
@@ -801,7 +805,15 @@ namespace NuGet.Commands
                     _request.Project.RestoreMetadata.CacheFilePath = null;
                 }
             }
-            return (cacheFile, noOp);
+            return (cacheFile, noOp, cacheFileAge);
+
+            static bool CacheFileExists(string path, out TimeSpan? cacheFileAge)
+            {
+                FileInfo fileInfo = new FileInfo(path);
+                if (fileInfo.Exists) { }
+                cacheFileAge = DateTime.UtcNow - fileInfo.LastWriteTimeUtc;
+                return fileInfo.Exists;
+            }
         }
 
         private string GetAssetsFilePath(LockFile lockFile)
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
@@ -82,10 +82,32 @@ namespace NuGet.Commands
         private const string ValidateRestoreGraphsDuration = nameof(ValidateRestoreGraphsDuration);
         private const string CreateRestoreResultDuration = nameof(CreateRestoreResultDuration);
         private const string IsCentralPackageTransitivePinningEnabled = nameof(IsCentralPackageTransitivePinningEnabled);
-        private const string VulnerablePackageCheck = nameof(VulnerablePackageCheck);
 
         // PackageSourceMapping names
         private const string PackageSourceMappingIsMappingEnabled = "PackageSourceMapping.IsMappingEnabled";
+
+        // NuGetAudit names
+        private const string AuditEnabled = "Audit.Enabled";
+        private const string AuditLevel = "Audit.Level";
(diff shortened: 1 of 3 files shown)


src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -82,11 +83,34 @@ public class RestoreCommand
...
         // PackageSourceMapping names
         private const string PackageSourceMappingIsMappingEnabled = "PackageSourceMapping.IsMappingEnabled";
 
+        // NuGetAudit names
+        private const string AuditEnabled = "Audit.Enabled";
+        private const string AuditLevel = "Audit.Level";
+        private const string AuditMode = "Audit.Mode";
+        private const string AuditDataSources = "Audit.DataSources";
+        private const string AuditDirectVulnerabilitiesPackages = "Audit.Vulnerability.Direct.Packages";
+        private const string AuditDirectVulnerabilitiesTotal = "Audit.Vulnerability.Direct.Total";
```
nkolev92: I'm guessing this is count?

The guidance for telemetry tends to be include `Count`, `Duration` etc and similar to make it obv what it is.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -82,11 +83,34 @@ public class RestoreCommand
...
+        private const string AuditDirectVulnerabilitySev0 = "Audit.Vulnerability.Direct.Severity0";
+        private const string AuditDirectVulnerabilitySev1 = "Audit.Vulnerability.Direct.Severity1";
+        private const string AuditDirectVulnerabilitySev2 = "Audit.Vulnerability.Direct.Severity2";
+        private const string AuditDirectVulnerabilitySev3 = "Audit.Vulnerability.Direct.Severity3";
+        private const string AuditDirectVulnerabilitySevOther = "Audit.Vulnerability.Direct.SeverityOther";
```
nkolev92: is this a just in case thing?

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -99,36 +125,81 @@ private void ReplayErrors(AggregateException exceptions)
...
+            stopwatch.Stop();
+            CheckPackagesDurationSeconds = stopwatch.Elapsed.TotalSeconds;
 
             if (packagesWithKnownVulnerabilities == null) return;
 
+            stopwatch.Restart();
+
+            int directPackageCount = packagesWithKnownVulnerabilities.Values.Count(p => p.IsDirect);
+            DirectPackagesWithAdvisory = new(capacity: directPackageCount);
+            TransitivePackagesWithAdvisory = new(capacity: packagesWithKnownVulnerabilities.Count - directPackageCount);
```
nkolev92, line 140: do you account for the mode here?

nkolev92 (commented):
Do we get transitive data if the mode is direct?

Can we maybe do that?

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -455,6 +473,57 @@ public async Task<RestoreResult> ExecuteAsync(CancellationToken token)
...
+
+            if (audit.DirectPackagesWithAdvisory is not null) { AddPackagesList(telemetry, AuditDirectVulnerabilitiesPackages, audit.DirectPackagesWithAdvisory); }
+            telemetry.TelemetryEvent[AuditDirectVulnerabilitiesTotal] = audit.DirectPackagesWithAdvisory?.Count ?? 0;
+            telemetry.TelemetryEvent[AuditDirectVulnerabilitySev0] = audit.Sev0DirectMatches;
+            telemetry.TelemetryEvent[AuditDirectVulnerabilitySev1] = audit.Sev1DirectMatches;
+            telemetry.TelemetryEvent[AuditDirectVulnerabilitySev2] = audit.Sev2DirectMatches;
+            telemetry.TelemetryEvent[AuditDirectVulnerabilitySev3] = audit.Sev3DirectMatches;
+            telemetry.TelemetryEvent[AuditDirectVulnerabilitySevOther] = audit.OtherSevDirectMatches;
+
+            if (audit.TransitivePackagesWithAdvisory is not null) { AddPackagesList(telemetry, AuditTransitiveVulnerabilitiesPackages, audit.TransitivePackagesWithAdvisory); }
```
nkolev92, line 499: Is this ever actually null?

nkolev92: Might be just me, but I'm having a tough time following what does it mean when some of these props gets populated.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
zivkan: Yes. Since I didn't want to fail restore, or hide an advisory from the customer that otherwise matches a package they're using, because of invalid data, if the server provides an invalid severity, NuGet will just accept it and show the value to the customer. Maybe I should call it `Invalid`, rather than `Other`.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
zivkan: In your review comment you wrote:

> Do we get transitive data if the mode is direct?
> Can we maybe do that?

By not accounting for the mode setting, we always collect transitive package data 😆

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
zivkan: Yes. AuditUtility works in 3 steps:

1. Get vulnerability data from sources
2. Check restored packages for known vulnerabilities
3. Format warning messages, and telemetry data

So, if either steps 1 or 2 early exit, then the collections will never be initialized (in addition to avoiding a bunch of unnecessary CPU work, loops that will never match anything, and memory allocations)

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
nkolev92: Tricky, tricky :D

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
nkolev92: I like Invalid. Other implies it's an accepted severity.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -455,6 +473,57 @@ public async Task<RestoreResult> ExecuteAsync(CancellationToken token)
...
+
+            void AddPackagesList(TelemetryActivity telemetry, string eventName, List<string> packages)
+            {
+                List<TelemetryEvent> result = new List<TelemetryEvent>(packages.Count);
+                foreach (var package in packages)
+                {
+                    TelemetryEvent packageData = new TelemetryEvent(eventName: null);
+                    packageData.AddPiiData("id", package);
+                    result.Add(packageData);
+                }
```
kartheekp-ms, line 521: I guess we should invoke `token.ThrowIfCancellationRequested();` inside forearch loop just in case if caller requested cancellation.

nkolev92 (approved):

martinrrm (approved):

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -737,7 +807,26 @@ private bool ValidatePackagesSha512(PackagesLockFile lockFile, LockFile assetsFi
                     _request.Project.RestoreMetadata.CacheFilePath = null;
                 }
             }
-            return (cacheFile, noOp);
+            return (cacheFile, noOp, cacheFileAge);
+
+            static bool CacheFileExists(string path, out TimeSpan? cacheFileAge)
```
kartheekp-ms, line 812: I think this local function has 2 responsibilities, one to check if the file exists and second one to fetch the cache file age. Can we have separate methods? I am okay with doing this in a follow-up PR.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -99,36 +125,81 @@ private void ReplayErrors(AggregateException exceptions)
...
+            Dictionary<PackageIdentity, PackageAuditInfo>? packagesWithKnownVulnerabilities =
                 FindPackagesWithKnownVulnerabilities(knownVulnerabilities);
+            stopwatch.Stop();
+            CheckPackagesDurationSeconds = stopwatch.Elapsed.TotalSeconds;
 
             if (packagesWithKnownVulnerabilities == null) return;
 
+            stopwatch.Restart();
+
+            int directPackageCount = packagesWithKnownVulnerabilities.Values.Count(p => p.IsDirect);
```
kartheekp-ms, line 138: It seems we're using the Count function to determine the size of the lists we've created below. Is there an alternative approach to calculate the number of elements?

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
zivkan: This foreach loop does trivial work, so should take almost no time.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
zivkan: If we call one method to check if the file exists, and another to get the timestamp, then we're going to go to the kernel twice, which is bad for performance.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
zivkan: Nothing better seems obvious to me.

PR #5275, opened 2023-06-23
Sort installed packages by vulenrabilities, deprecation then a-z

### LINKED ISSUE #12693: Sort installed packages in PM UI by Vulnerabilities
"Promote" vulnerable and deprecated packages to the top of the list in the Installed tab in PM UI

martinrrm:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12693

## Description

"Promote" vulnerable and deprecated packages to the top of the list in the Installed tab in PM UI

image
image

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A  I wasn't able to test this, I found that testing a vulnerable/deprecated package in the InfiniteScroll is hard since we do another query for that metadata information. 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
@@ -743,6 +743,11 @@ namespace NuGet.PackageManagement.UI
             ItemsView.GroupDescriptions.Clear();
         }
 
+        internal void ClearPackageSorting()
+        {
+            ItemsView.SortDescriptions.Clear();
+        }
+
         internal void AddPackageLevelGrouping()
         {
             ItemsView.Refresh();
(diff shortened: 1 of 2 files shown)


zivkan:
The linked issue was created minutes before this PR, and doesn't contain any supporting data. Has this idea been validated in any way? 

If I'm looking for a specific package in the list, now that the list is no longer sorted alphabetically, I find it harder to find the package I care about. This PR's order makes sense when you want to make the vulnerable and deprecated more obvious, but when the customer is trying to achieve different things it's less intuitive. 

Also, consider the point of view of a customer with zero context. Maybe they'll figure out that packages without the warning icon are sorted alphabetically, but why are the packages with the warning icon not sorted alphabetically? There's no visual difference between vulnerable and deprecated. If the icon was different, or if there were subheadings, I think it would be a lot easier to understand.

It's unfortunate that PM UI doesn't currently have any configurable sort-by option. I think it might be nice to be able to choose "default" or "alphabetical"

martinrrm:
@zivkan Thanks for the feedback, I completely ignored to add more context of the decision. I'm going to create a [Client.Engineering issue](https://github.com/NuGet/Client.Engineering/issues/2361) to give more details.

About the ordering, now that you mentioned it, you're right, there is no way to differentiate between Vulnerable and Deprecated making the list appear that is not sorted, maybe we can show them together, so the user knows that those packages need attention? Also, in each group they are sorted alphabetical.

nkolev92:
@JonDouglas thoughts about having vuln and deprecated group together as the indicator is the same today making it confusing.

JonDouglas:
@nkolev92 Maybe consider https://github.com/NuGet/Home/issues/12506 to help distinguish. Could likely get away with icon changes for now?

martinrrm:
@JonDouglas I'll try to change icons to see if its more clear

JonDouglas:
@zivkan I published a branch with a small written draft that relates to your concerns. It happened to be a topic we have discussed over the last few weeks around this work & UX considerations. I'll likely have workflows, mockups, and further definition on it next week. Right now it is a basic write-up.

https://github.com/NuGet/Home/blob/vs-filters-sorts/proposed/2023/vs-filters.md

martinrrm:
Converted this to draft, since there is no clear direction of what to do, maybe the best is to wait for 17.8 and reconsider

nkolev92:
We should probably investigate the longer term approach for 17.8, do the work once

martinrrm:
Closing in favor of a better filtering design https://github.com/NuGet/Home/blob/6225561b2aa85ddfe55024344155c7ec82ff8973/proposed/2023/vs-filters.md

PR #5279, opened 2023-06-26
Reduce allocations in VersionRangeFormatter

### LINKED ISSUE #12707: Heavy allocations in NuGet.Commands.RestoreRunner.ExecuteAndCommitAsync|nuget.versioning.dll!NuGet.Versioning.VersionFormatter.Format
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1841687

drewnoakes:
## Bug

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1841687
Fixes: https://github.com/NuGet/Home/issues/12707

## Description

`VersionRangeFormatter` calls into `VersionFormatter` for almost all of its operations. The latter produces a `string` which the former then composes into its output. This intermediary `string` is a wasteful allocation that can be avoided. GCPauseWatson shows that `VersionRangeFormatter` calls into `VersionFormatter` often enough that these temporary strings contribute to GC pauses on customer machines in the real world.

This change creates a pathway for `VersionRangeFormatter` to pass a `StringBuilder` to `VersionFormatter` in order to avoid the temporary string allocation. It also changes `VersionRangeFormatter` to use a `StringBuilder` more consistently internally, and to pass that to other methods it invokes where possible.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Versioning/VersionRangeFormatter.cs
+++ b/src/NuGet.Core/NuGet.Versioning/VersionRangeFormatter.cs
@@ -11,9 +11,6 @@ namespace NuGet.Versioning
     /// </summary>
     public class VersionRangeFormatter : IFormatProvider, ICustomFormatter
     {
-        private const string ZeroN = "{0:N}";
-        private static readonly VersionFormatter VersionFormatter = VersionFormatter.Instance;
-
         /// <summary>
         /// A static instance of the <see cref="VersionRangeFormatter"/> class.
         /// </summary>
@@ -29,8 +26,7 @@ namespace NuGet.Versioning
                 throw new ArgumentNullException(nameof(arg));
             }
 
-            var range = arg as VersionRange;
-            if (range == null)
+            if (arg is not VersionRange range)
             {
                 throw ResourcesFormatter.TypeNotSupported(arg.GetType(), nameof(arg));
             }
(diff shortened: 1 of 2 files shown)


src/NuGet.Core/NuGet.Versioning/VersionRangeFormatter.cs
```diff
@@ -11,9 +11,6 @@ namespace NuGet.Versioning
     /// </summary>
     public class VersionRangeFormatter : IFormatProvider, ICustomFormatter
     {
-        private const string ZeroN = "{0:N}";
```
drewnoakes, line 14: This was used to tell `VersionFormatter` to run its `AppendNormalized` method. This PR just makes that method internal and calls it directly to avoid a bunch of indirection, in addition to the temporary string allocations.

src/NuGet.Core/NuGet.Versioning/VersionRangeFormatter.cs
```diff
@@ -40,32 +36,14 @@ public string Format(string? format, object? arg, IFormatProvider? formatProvide
...
+
+            for (var i = 0; i < format!.Length; i++)
             {
-                string formatted = Format(format[0], range);
-                return formatted;
+                Format(builder, format[i], range);
             }
-            else
-            {
-                var sb = StringBuilderPool.Shared.Rent(format.Length);
```
drewnoakes, line 51: This length was not correct. The format string here is typically a single character such as `N`, with length one. The output string is rarely one.

nkolev92 (approved):

PR #5286, opened 2023-06-29
Show icon in InfiniteScrollList when network error getting package info

zivkan:
## Bug

Maybe Fixes: https://github.com/NuGet/Client.Engineering/issues/2368

Regression? no

## Description

Looking at NuGet's fault telemetry, the vast majority of `TaskCanceledException` faults come from `PackageItemViewModel`'s `ReloadPackageVersionsAsync` and `ReloadPackageMetadataAsync` methods. Since both of these already suppress the fault when the cancelation token requests cancelation, it's likely that the telemetry is due to `HttpClient` timeouts.

Since the error means that package information was not retrieved successfully (either vulnerability/deprecation, or version information), show something to the customer. This is particularly important for missing deprecation or vulnerability information, otherwise it looks like the package doesn't have vulnerabilities or deprecation information, whereas in fact PM UI doesn't know if it does or not.

Injecting some random failures, here's what it looks like with the tooltip:
image

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception: I don't know how to test httptime exceptions in this codepath
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
@@ -994,4 +994,7 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
     <value>_OK</value>
     <comment>Button_Cancel and Button_OK must have underscore in different letters. See https://docs.microsoft.com/windows/apps/design/input/access-keys#choose-access-keys</comment>
   </data>
+  <data name="Label_NetworkError" xml:space="preserve">
+    <value>Error getting additional information about package. Refresh to try again.</value>
+  </data>
 </root>
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
@@ -995,6 +995,6 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
     <comment>Button_Cancel and Button_OK must have underscore in different letters. See https://docs.microsoft.com/windows/apps/design/input/access-keys#choose-access-keys</comment>
   </data>
   <data name="Label_NetworkError" xml:space="preserve">
-    <value>Error getting additional information about package. Refresh to try again.</value>
+    <value>Error getting additional information about the package. Refresh to try again.</value>
   </data>
 </root>
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
@@ -995,6 +995,6 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
     <comment>Button_Cancel and Button_OK must have underscore in different letters. See https://docs.microsoft.com/windows/apps/design/input/access-keys#choose-access-keys</comment>
   </data>
   <data name="Label_NetworkError" xml:space="preserve">
-    <value>An error occurred while fetching additional information about the package. Please refresh to try again.</value>
+    <value>An error occurred while fetching additional information about the package. Refresh to try again.</value>
   </data>
 </root>
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
@@ -995,6 +995,6 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
     <comment>Button_Cancel and Button_OK must have underscore in different letters. See https://docs.microsoft.com/windows/apps/design/input/access-keys#choose-access-keys</comment>
   </data>
   <data name="Label_NetworkError" xml:space="preserve">
-    <value>Error getting additional information about the package. Refresh to try again.</value>
+    <value>An error occurred while fetching additional information about the package. Please refresh to try again.</value>
   </data>
 </root>
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
@@ -732,10 +732,10 @@ namespace NuGet.PackageManagement.UI
                 {
                     await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(cancellationToken);
                     IsPackageWithNetworkErrors = true;
-        }
+                }
                 catch (OperationCanceledException)
                 {
-                    // task cancelation
+                    // if cancellationToken cancelled before the above is scheduled on UI thread, don't log fault telemetry
                 }
             }
         }
@@ -770,7 +770,7 @@ namespace NuGet.PackageManagement.UI
                 }
                 catch (OperationCanceledException)
                 {
-                    // task cancelation
+                    // if cancellationToken cancelled before the above is scheduled on UI thread, don't log fault telemetry
                 }
             }
         }
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
@@ -404,6 +404,20 @@ namespace NuGet.PackageManagement.UI
             get => IsPackageDeprecated || IsPackageVulnerable;
         }
 
+        private bool _isPackageWithNetworkErrors;
+        public bool IsPackageWithNetworkErrors
+        {
+            get => _isPackageWithNetworkErrors;
+            set
+            {
+                if (IsPackageWithNetworkErrors != value)
+                {
+                    _isPackageWithNetworkErrors = value;
+                    OnPropertyChanged(nameof(IsPackageWithNetworkErrors));
+                }
+            }
+        }
+
         private Uri _iconUrl;
         public Uri IconUrl
         {
@@ -711,6 +725,19 @@ namespace NuGet.PackageManagement.UI
             {
                 // UI requested cancellation
             }
+            catch (TaskCanceledException)
+            {
+                // HttpClient throws TaskCanceledExceptions for HTTP timeouts
+                try
+                {
+                    await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(cancellationToken);
+                    IsPackageWithNetworkErrors = true;
+        }
+                catch (OperationCanceledException)
+                {
+                    // task cancelation
+                }
+            }
         }
 
         private async Task ReloadPackageMetadataAsync()
@@ -732,6 +759,19 @@ namespace NuGet.PackageManagement.UI
             catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
             {
                 // UI requested cancellation.
+            }
+            catch (TaskCanceledException)
+            {
+                // HttpClient throws TaskCanceledExceptions for HTTP timeouts
+                try
+                {
+                    await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(cancellationToken);
+                    IsPackageWithNetworkErrors = true;
+                }
+                catch (OperationCanceledException)
+                {
+                    // task cancelation
+                }
             }
         }
 
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.Designer.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.Designer.cs
@@ -1096,6 +1096,15 @@ namespace NuGet.PackageManagement.UI {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to Error getting additional information about package. Refresh to try again..
+        /// </summary>
+        public static string Label_NetworkError {
+            get {
+                return ResourceManager.GetString("Label_NetworkError", resourceCulture);
+            }
+        }
+        
+        /// <summary>
         ///   Looks up a localized string similar to NuGet: {0}.
         /// </summary>
         public static string Label_NuGetWindowCaption {
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.Designer.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.Designer.cs
@@ -1096,7 +1096,7 @@ namespace NuGet.PackageManagement.UI {
         }
         
         /// <summary>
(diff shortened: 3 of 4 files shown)


src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
```diff
@@ -994,4 +994,7 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
     <value>_OK</value>
     <comment>Button_Cancel and Button_OK must have underscore in different letters. See https://docs.microsoft.com/windows/apps/design/input/access-keys#choose-access-keys</comment>
   </data>
+  <data name="Label_NetworkError" xml:space="preserve">
+    <value>Error getting additional information about package. Refresh to try again.</value>
```
donnie-msft: Not sure if this sounds better

```suggestion
    <value>Error getting additional information about the package. Refresh to try again.</value>
```

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
```diff
@@ -711,6 +725,19 @@ private async System.Threading.Tasks.Task ReloadPackageVersionsAsync()
...
                 // UI requested cancellation
             }
+            catch (TaskCanceledException)
+            {
+                // HttpClient throws TaskCanceledExceptions for HTTP timeouts
+                try
+                {
+                    await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(cancellationToken);
+                    IsPackageWithNetworkErrors = true;
+        }
```
donnie-msft: formatting

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
```diff
@@ -711,6 +725,19 @@ private async System.Threading.Tasks.Task ReloadPackageVersionsAsync()
             {
                 // UI requested cancellation
             }
+            catch (TaskCanceledException)
+            {
+                // HttpClient throws TaskCanceledExceptions for HTTP timeouts
+                try
```
donnie-msft, line 731: The nested `try` feels weird. What scenario is it catching?

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
```diff
@@ -711,6 +725,19 @@ private async System.Threading.Tasks.Task ReloadPackageVersionsAsync()
...
+                {
+                    await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(cancellationToken);
+                    IsPackageWithNetworkErrors = true;
+        }
+                catch (OperationCanceledException)
+                {
+                    // task cancelation
+                }
```
donnie-msft:
```suggestion
                if (ex.Token != cancellationToken)
                {
                    await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(cancellationToken);
                    IsPackageWithNetworkErrors = true;
                }
```

donnie-msft (commented):
Thanks for doing this, as I have had in my backlog to do something like this for async loading in PM UI.
It's early previews, so we have time to fix bugs, but I'm wondering if the new icon was tested in a few different ways... resizing the pane, with/without Updates Available, different tabs, etc? I'll try to test your VSIX today as well.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
nkolev92: The code is switching to the UI thread. It's a good practice to pass a cancellation token to that method to avoid needlessly switching to the UI thread.

nkolev92 (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
donnie-msft: I'm aware of that generally, but that doesn't explain the scenario here. The cancellation token is not the UI token, but it could be cancelled? Also, I suggested just checking the token here https://github.com/NuGet/NuGet.Client/pull/5286#discussion_r1247006277

martinrrm (dismissed):
This looks amazing!

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
zivkan: The purpose is to avoid any possibility of having more TaskCanceledExceptions, or OperationCanceledExceptions, being logged as fault telemetry.

This is async code, which might mean all work has to be scheduled on a thread, and this introduces latency. Plus, switching from a thread pool thread to the UI thread requires scheduling. So, it's theoretically possible for cancelation to be requested in the very small time between the http request timing out and the view model update being scheduled and executing on the UI thread. When VS is very busy, or if the customer has other processes running, then the CPU will be busy and it will take longer for the OS to schedule VS threads on CPU cores, and for the UI thread and thread pool scheduler to pick our work from the work queue.

If we write the code in a way to assume it can't throw, but it does in fact throw and see we telemetry that we want to investigate, it might be very difficult to figure out the root cause. We would read the code and we'll see that there's already code handling cancelation, so it might not be obvious that there's a timing bug between catching the http timeout and the exception handler completing successfully.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
zivkan: `ex.Token != canancellationToken` should never be false, because of the previous `catch` block a few lines higher. Also, as per the other comment conversation, the switch to main thread takes time, during which the cancelation token could be triggered, and we don't want to record more fault telemetry: https://github.com/NuGet/NuGet.Client/pull/5286#discussion_r1247512600

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
zivkan: It absolutely does! 🤘

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
donnie-msft: I see, so the concern is cancellation occurring in the middle of the switch to main. 
Why not just use `CancellationToken.None`?

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
donnie-msft: > If we write the code in a way to assume it can't throw, but it does in fact throw

I'd never suggest assuming anything can't throw. The other approach is not to care about cancellation. 

The catch above here already ignores UI cancellations: `// UI requested cancellation`
Therefore, this catch must be for non-UI cancellations. I thought that's exactly when we *do want to show* your new icon? 

How did you go about testing this? I'm tempted to try Clumsy or similar (despite previous bad experiences with it).

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
```diff
@@ -994,4 +994,7 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
     <value>_OK</value>
     <comment>Button_Cancel and Button_OK must have underscore in different letters. See https://docs.microsoft.com/windows/apps/design/input/access-keys#choose-access-keys</comment>
   </data>
-</root>
\ No newline at end of file
+  <data name="Label_NetworkError" xml:space="preserve">
+    <value>Error getting additional information about the package. Refresh to try again.</value>
```
jeffkl:
```suggestion
    <value>An error occurred while fetching additional information about the package. Please refresh to try again.</value>
```

jeffkl (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
zivkan: I don't know if PM UI correctly handles this, if not it's a performance bug, but if PM UI is closed while these HTTP requests are pending, or the customer clicks VS's close button (so shutdown starts), there's no benefit to successfully getting onto the UI thread, changing the view model and possibly causing the UI to then do redraw work.

I don't know how sensitive to performance all the various stakeholders are, so maybe I'm overthinking this. But on the other hand, VS is a HUGE app, with many components, with a lot of UI thread switching already. So even if it doesn't make a big difference here in isolation, maybe it contributes to "death by a thousand cuts"?  When doing performance work, I'm used to dealing with hot paths, particularly restore, where even little improvements are desirable.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
zivkan: This conversation has become a duplicate of what https://github.com/NuGet/NuGet.Client/pull/5286#discussion_r1247003359 has become, so I'm resolving this one to have a single conversation.

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
donnie-msft: I believe "please" isn't supposed to be used in certain cases. Are we sure we should take that part of the suggestion?

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
donnie-msft: > if PM UI is closed while these HTTP requests are pending

Weren't the HTTP requests already cancelled, because we're in the catch?

> or the customer clicks VS's close button (so shutdown starts), there's no benefit to successfully getting onto the UI thread,

Isn't this the entire point of JTF? I'm now noticing this method is invoked via JTF.RunAsync but it's not awaited. Is that what you're getting at?

> changing the view model and possibly causing the UI to then do redraw work.

I asked earlier about how you went about testing. Did you happen to test this *without* switching to the UI thread? WPF marshalls the property changed calls to the UI thread. There's some situations where it doesn't, particularly around bound collections.

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
zivkan: https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/p/please

It's unclear if the example they provide is something to avoid or acceptable usage. Windows login prompt just says "Try again", no please.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
zivkan: > Weren't the HTTP requests already cancelled, because we're in the catch?

There are two exception handlers from a single `try` block.

The first is when our own `CancelationToken` is in a state requesting cancelation, and was added months ago. This PR is adding a second, which is when we did not request cancelation.

I'll use the term "timing reasons" to mean anything to do with pre-emptive scheduling on a single CPU core, any timing related risks due to having multiple threads executing concurrently on multiple CPU cores, etc. So, because of these timing reasons, it's possible for our own token not to request cancelation when our code is choosing between exception handler 1 or 2, but our token is requesting cancelation before exception handler 2 is finally running on the UI thread.

>> or the customer clicks VS's close button (so shutdown starts), there's no benefit to successfully getting onto the UI thread,
> 
> Isn't this the entire point of JTF?

I feel like you and me are probably talking about different contexts, which makes me feel like we're misunderstanding each other. Otherwise, I don't understand the question being asked. I don't believe JTF to have a singular purpose, it has a few. But what you quoted from me is trying to say "updating the UI to show something that is no longer relevant (say, VS is closing down) does not benefit the customer", so I'm not sure how JTF even fits in this picture, other than the fact that this code happens to be calling a JTF method.

In case it helps, here is JTF's docs on requesting the main thread: https://github.com/microsoft/vs-threading/blob/5680786f95e51f65ab6286898725c537f5f3ac87/doc/cookbook_vs.md?plain=1#L59-L70. It documents that `OperationCanceledException` will be thrown on a `ThreadPool` thread.

> I'm now noticing this method is invoked via JTF.RunAsync but it's not awaited. Is that what you're getting at?

This code not running on an `await`ed task is relevant, because we log fault telemetry on such tasks via `PostOnFailure()`, and all these catch blocks in this method are for the express purpose of not logging fault telemetry on expected scenarios, but keep logging faults for unexpected failures.

> I asked earlier about how you went about testing. Did you happen to test this without switching to the UI thread?

Sorry, I forgot to reply to that.  I tested by creating a static `Random` instance, and then putting `if (_random.Next(5) == 0) { await Task.Delay(TimeSpan.FromSeconds(10)); throw TaskCanceledException(); }` as the first statement after the `try {`, before the `await _searchService.`.

Regardless of how the HTTP timeout was simulated, I didn't test the nested try-catch block's catch handler, because it's a no-op, I know what the outcome will be.

In order to happen in practise, the cancelation would need to be requested in probably sub-millisecond accuracy/accident, but that doesn't mean it'll never happen on customers machines.

nkolev92 (approved):

jzabroski:
Spoke with @JonDouglas about this. While this does surface errors that are non-obvious and that is appreciated, today is the first time I personally saw this kind of error. I guess I have a very fast connection and so don't notice this often, plus I rely on a large package cache, which I do not clear regularly. 

Is there a way to make this appear in the Visual Studio Errors List, instead of just the InfiniteScrollList ?  These tiny icons in various places in Visual Studio are easily my _least_ favorite aspect of using Visual Studio. Even the Solution Explorer has these error icons and I frequently have no idea what they mean. Rule #1 of UI development: Maximize recognition, minimize recall.

Unrelated, this is a good case study for why TaskCancelledException is terrible and I keep championing a zio-like library for .NET, which requires custom async results builders. - You would be able to return the root cause of the exception to the caller, and they could explicitly handle that, vs. digging inside stack traces to wonder what happened.

PR #5288, opened 2023-06-30
Only look at local package sources when prompting for License Acceptance

donnie-msft:
## Bug

## Description

When Package Source Mapping is enabled, existing logic looks up licenses against all enabled package sources. This essentially ignores the source mappings.
There shouldn't be a need to look at remote sources for license metadata. 

- This PR changes PM UI to look only at local sources for all projects when prompting for license acceptance.
- Add `GetAllPackageFoldersAsync` to search service to avoid copying the logic into UI directly.
- Remove analyzer suppression by adding a Cast for the metadata collection.

I tested by debugging and verifying the expected local sources were being passed:
image

When 0 local sources are found: same experience as usual, the action fails with the Error List showing:
image

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception - Using existing services to retrieve local package sources for the UI to use in the prompt. Debugging screenshot above.
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
```diff
@@ -1013,47 +1015,23 @@ private PackageLicenseInfo GeneratePackageLicenseInfo(IPackageSearchMetadata met
...
             using (var sourceCacheContext = new SourceCacheContext())
             {
-                // first check all the packages with local sources.
-#pragma warning disable CS8619 // Nullability of reference types in value doesn't match target type.
                 IPackageSearchMetadata[] completed = (await TaskCombinators.ThrottledAsync(
                     allPackages,
                     (p, t) => GetPackageMetadataAsync(localSources, sourceCacheContext, p, t),
-                    token)).Where(metadata => metadata != null).ToArray();
-#pragma warning restore CS8619 // Nullability of reference types in value doesn't match target type.
+                    token)).Cast<IPackageSearchMetadata>().Where(metadata => metadata != null).ToArray();
```
zivkan: Minor "reading understanding" improvement, though there's no functional difference.

```suggestion
                    token)).Where(metadata => metadata != null).Cast<IPackageSearchMetadata>().ToArray();
```

To me, casting to a non-nullable type tells me that there are already no nulls possible. If that happens before the `!= null` filter, that suggests to me that there can't be any nulls before that point, making the `!= null` check unnecessary (and in that case, `GetPackageMetadataAsync` has the wrong annotation).

By moving the cast to after the `!= null` filtering, it tells me that filtering the nulls is functionally important.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetPackageSearchService.cs
```diff
@@ -416,7 +416,7 @@ private async ValueTask<IInstalledAndTransitivePackages> GetInstalledAndTransiti
         /// <param name="projectContextInfos">A collection of projects</param>
         /// <param name="cancellationToken">Cancellation Token</param>
         /// <returns>A collection of Global package folder repositories combined with repositories found in packageFolders from PackageReference projects</returns>
-        private async Task<IReadOnlyList<SourceRepository>> GetAllPackageFoldersAsync(
+        public async ValueTask<IReadOnlyList<SourceRepository>> GetAllPackageFoldersAsync(
```
zivkan: Putting a breakpoint on the `await Task.WhenAll` call a few lines down, when opening PM UI, the breakpoint was hit 4 times, and all 4 times `tasks` contained 1 task, whose state was `WaitingForActivation`. In order words, it did not complete synchronously.

From [a (long) blog post about `ValueTask`](https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/)

> unless the performance implications outweigh the usability implications, Task / Task<TResult>are still preferred

and

>  either you expect synchronous completion to be a very common case, or you’re able to effectively pool objects for use with asynchronous completion

So, this doesn't appear to be one of the situations where `ValueTask` provides benefits.

All this to say I think keeping it as `Task<T>` is better.

donnie-msft: Thanks for the links. Was just blindly copying from the other methods in the service.

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/NuGetSearchServiceReconnector.cs
```diff
@@ -124,6 +124,11 @@ public void Dispose()
                 // do not dispose `_parent`, it's the responsibility of the object owning the instance.
             }
 
+            public Task<IReadOnlyList<SourceRepository>> GetAllPackageFoldersAsync(IReadOnlyCollection<IProjectContextInfo> projectContextInfos, CancellationToken cancellationToken)
+            {
+                return _parent._service.GetAllPackageFoldersAsync(projectContextInfos, cancellationToken);
```
jeffkl, line 129: This line looks odd to me to be accessing two private variables.  Maybe this class should just accept the `INuGetSearchService` instead of the `NuGetSearchServiceReconnector` since only the `_parent._service` is used?  I know its the existing pattern but does this count as "technical debt" and can be cleaned up while you're here?

jeffkl (approved):

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/NuGetSearchServiceReconnector.cs
donnie-msft: @zivkan thoughts?

donnie-msft: Decided it seems more than a trivial change. There's a comment about `Dispose` logic, so it's possible there was some bug this structure prevented. Anyway, I filed a TechDebt issue. Please comment on it & close it if there's reason for the nested variables.

https://github.com/NuGet/Home/issues/12722

jeffkl: Thanks for looking into it!

PR #5289, opened 2023-07-03
Update System.Security.Cryptography.Pkcs version

zivkan:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2372

Regression? no

## Description

* System.Security.Cryptography.Pkcs has a known vulnerability, and needs to be upgraded. Or at least, it's easier to upgrade than to figure out if we're using the package in a way that's vulnerable or not.
* System.Security.Cryptography.Cng hasn't been getting newer versions since 5.0.0, but is listed as a dependency of System.Security.Cryptography.Pkcs, and we only used .Cng where were also used .Pkcs`, so I removed the `.Cng` references
* Although `CryptographyPackagesVersion` is now only used by 1 PackageVersion (Pkcs), I didn't rename the property, because I assume that source-build uses it to override the version
* The newer version of .Pkcs (or its dependencies) now bring System.Buffer into the restore graph for NuGet.Packaging's netstandard2.0 build, so the internal copy of `ArrayPool` was causing compiler errors. Hence, stop using it in netstandard2.0, since we can reference it from the .NET library.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
 
-    <CryptographyPackagesVersion Condition="'$(CryptographyPackagesVersion)' == ''">5.0.0</CryptographyPackagesVersion>
+    <CryptographyPackagesVersion Condition="'$(CryptographyPackagesVersion)' == ''">6.0.4</CryptographyPackagesVersion>
     <MicrosoftCodeAnalysisPublicApiAnalyzersVersion Condition="'$(MicrosoftCodeAnalysisPublicApiAnalyzersVersion)' == ''">3.3.4</MicrosoftCodeAnalysisPublicApiAnalyzersVersion>
     <NewtonsoftJsonPackageVersion Condition="'$(NewtonsoftJsonPackageVersion)' == ''">13.0.1</NewtonsoftJsonPackageVersion>
     <SystemPackagesVersion Condition="'$(SystemPackagesVersion)' == ''">4.3.0</SystemPackagesVersion>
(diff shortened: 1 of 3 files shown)


jeffkl (approved):

martinrrm (approved):

nkolev92:
cc @NikolaMilosavljevic @MichaelSimons can confirm what this change does for source build. 

We *might* be able to change the property name

MichaelSimons:
> cc @NikolaMilosavljevic @MichaelSimons can confirm what this change does for source build.
> 
> We _might_ be able to change the property name

I think this upgrade is going to cause a prebuilt in source-build.  System.Security.Cryptography.Pkcs 6.0.4 will need to be added to https://github.com/dotnet/source-build-reference-packages.  Instructions for adding new packages are [here](https://github.com/dotnet/source-build-reference-packages/blob/main/README.md#adding-new-packages).

zivkan:
That doesn't address the question about the MSBuild property name used to override the package version in this repo, but ok: https://github.com/dotnet/source-build-reference-packages/pull/738

However, after creating the PR, I realize that the version of the package we're currently using, 5.0.0, isn't in the repo at the moment anyway 🤷

MichaelSimons:
> That doesn't address the question about the MSBuild property name used to override the package version in this repo

The property name doesn't matter here because source-build is able to utilize a reference package.  The property name becomes important if the package reference has to be lifted during source-build to the latest.  In that case, it has to follow a naming convention.

> However, after creating the PR, I realize that the version of the package we're currently using, 5.0.0, isn't in the repo at the moment anyway 🤷

Isn't it here: https://github.com/dotnet/source-build-reference-packages/tree/main/src/referencePackages/src/system.security.cryptography.pkcs/5.0.0?

zivkan:
> Isn't it here

Yes, yes it is.

I have no idea what was wrong with me this morning. Trying to rush to finish too many things I have time for before the end of the day. 🤦‍♂️

PR #5290, opened 2023-07-03
Select the correct installed version when version is not normalized in nuspec.

### LINKED ISSUE #12661: Deprecated info will flash for less than one second in the right penal when clicking package “Microsoft.Net.Http” with a non-deprecated version in the package list
### NuGet Product Used

### Product Version

NuGet Client Dev\6.7.0.83

### Worked before?

_No response_

### Impact

#### Repro Steps:
1. Open VS and create a F# Console Application (.Net Framework 4.8.1) project. 
2. Right-click the project in Solution Explorer and select “Manage NuGet Packages…” menu item to open PM UI. 
3. Click “Browse” tab and select the “nuget.org” from Package source drop-down list.
4. Search the package “FSharp.Net.Http” in the filter and click the package in the package list.
5. Install the package “FSharp.Net.Http” and go to the “Installed” tab.
6. Click the package “Microsoft.Net.Http” whose latest version is not a deprecated.

#### Expected:
Deprecated info should not show in the right penal since the latest version is not a deprecated.
#### Actual:
Deprecated info will flash for less than one second in the right penal as below record.
DeprecatedInfoFlash

#### Notes:
1.Repro rate: 100%.
2.It is not a regression bug since it reproes on 17.5.4 Preview 1.0 [33530.505.d17.5]+ Dev\6.5.0.154.
3.The issue will not repro when clicking the same package “Microsoft.Net.Http” in the "Browse' tab. It only reproes when the package is a transitive package.

martinrrm:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12661

## Description

`InstalledVersion` property in the PM UI is calculated with the metadata of the installed package. If the user packs with our official tooling a validation will be run and the version will be correct, but users can create their nuspec and zip the package themselves, this can lead to versions with the `revision` property.

In this case the package version `Microsoft.Net.Http 2.0.20710` has a nuspec version of `2.0.20710.0` which will break the combobox since we are comparing strings and select another version (latest one), this quick change of versions makes the UI change really fast and make the "flash" effect. `ToNormalizedString();` help us to remove that extra 0. 

installed-version-fix

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A  I'll ask vendors to test this again

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
@@ -732,7 +732,7 @@ namespace NuGet.PackageManagement.UI
             }
             else
             {
-                var installedVersion = _searchResultPackage?.AllowedVersions?.OriginalString ?? _searchResultPackage?.InstalledVersion.ToString();
+                var installedVersion = _searchResultPackage?.AllowedVersions?.OriginalString ?? _searchResultPackage?.InstalledVersion.ToNormalizedString();
                 SelectedVersion =
                     possibleVersions.FirstOrDefault(v => StringComparer.OrdinalIgnoreCase.Equals(v.Range?.OriginalString, installedVersion))
                     ?? possibleVersions.FirstOrDefault(v => v.IsValidVersion);


nkolev92 (approved):

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
```diff
@@ -732,7 +732,7 @@ private void SelectVersionPackageReferenceProject(NuGetVersion latestVersion)
             }
             else
             {
-                var installedVersion = _searchResultPackage?.AllowedVersions?.OriginalString ?? _searchResultPackage?.InstalledVersion.ToString();
+                var installedVersion = _searchResultPackage?.AllowedVersions?.OriginalString ?? _searchResultPackage?.InstalledVersion.ToNormalizedString();
```
zivkan, line 735: In the checklist, you listed "N/A" for testing. Why do you believe testing is not applicable?

I know many parts of NuGet are not written in ways that are testable, especially the VS parts. So I'd believe a test exception (not n/a) saying it's not testable. But, we should always look for opportunities to write automated tests that can prevent regressions.

zivkan (approved):

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
martinrrm: Its not that I believe is not applicable, I would love to add tests to everything UI related, especially this component that always have bugs (and I created 😢), but I don't see an easy way to test this. Hopefully when we fix APEX tests, we can add more tests to the UI that can catch these behaviors.

PR #5291, opened 2023-07-03
Combobox version is aware of labels when autoselecting version

### LINKED ISSUE #12724: Combobox versions autoselect doesn't work with deprecation/vulnerable information
### NuGet Product Used

### Product Version

VS 2022

### Worked before?

_No response_

### Impact

### Repro Steps & Context

When versions have a label the combobox is not able to select the correct version when using the "Enter key", this happens for floating/ranges too.

combobox-versions-additionalInfo-before

### Verbose Logs

_No response_

### LINKED ISSUE #12725: The version doesn’t display the “(Deprecated)” or “(Vulnerable)” label in version box when typing a deprecated or vulnerable package version in the “version” drop-down box
### NuGet Product Used

### Product Version

NuGet Client Private Dev\6.8.0.14

### Worked before?

_No response_

### Impact

#### Repro Steps:    

1. Create a C# Console App (.Net core) project.    
2. Right-click the project in Solution Explorer and select "Manage NuGet Packages…" menu item to open PM UI.                                         
3. Go to "Browse" tab and click a package (for example “NuGet.Core” or "Newtonsoft.Json").   
4. Click the "Version" drop-down box in the right panel.  
5. Type a version (for example “2.14.0” for package “NuGet.Core” or "12.0.1" for package "Newtonsoft.Json") which already exists in the list and press the "Enter key". 

#### Expected:     

The version will display “(Deprecated)” or “(Vulnerable)” label in version box. 

#### Actual:     

The version doesn’t display the “(Deprecated)” or “(Vulnerable)” label as below： 
screenshot2

#### Note:    
1. The repro rate is 100%.     
2. This is not a regression since it is a new feature change. 

### Verbose Logs

_No response_

martinrrm:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12724
Fixes: https://github.com/NuGet/Home/issues/12725

## Description

When trying to select a version using "Enter" key the auto selected version was not correctly set, this issue was happening to floating/ranges versions too.

With this PR the combobox is aware if the version has deprecated/vulnerable info and the comparison is made against `SelectedItem.Version` and not `SelectedItem.ToString()` which contains the label so users don't need to write the label to autoselect.

combobox-versions-additionalInfo-after

Fix for https://github.com/NuGet/Home/issues/12725
versions-show-label-combobox

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A  Will ask vendors to do some testing in this, we want to merge this to 17.7 too

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageDetailControlModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageDetailControlModel.cs
@@ -252,7 +252,7 @@ namespace NuGet.PackageManagement.UI
             if ((UserInput.StartsWith("(", StringComparison.OrdinalIgnoreCase) || UserInput.StartsWith("[", StringComparison.OrdinalIgnoreCase)) &&
                VersionRange.TryParse(UserInput, out VersionRange userRange))
             {
-                if (o != null && NuGetVersion.TryParse(o.ToString(), out NuGetVersion userVersion))
+                if (o != null && NuGetVersion.TryParse(version.Version.ToString(), out NuGetVersion userVersion))
                 {
                     if (userRange.Satisfies(userVersion))
                     {
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageDetailControlModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageDetailControlModel.cs
@@ -252,12 +252,9 @@ namespace NuGet.PackageManagement.UI
             if ((UserInput.StartsWith("(", StringComparison.OrdinalIgnoreCase) || UserInput.StartsWith("[", StringComparison.OrdinalIgnoreCase)) &&
                VersionRange.TryParse(UserInput, out VersionRange userRange))
             {
-                if (o != null && NuGetVersion.TryParse(version.Version.ToString(), out NuGetVersion userVersion))
+                if (o != null && userRange.Satisfies(version.Version))
                 {
-                    if (userRange.Satisfies(userVersion))
-                    {
-                        return true;
-                    }
+                    return true;
                 }
 
                 return false;
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
@@ -161,7 +161,7 @@ namespace NuGet.PackageManagement.UI
                         if (_versions.SelectedIndex >= 0 && isAVersion)
                         {
                             // Check if the user typed a custom version
-                            if (isUserInputValidNuGetVersionRange)
+                            if (isUserInputValidNuGetVersionRange && userRequestedVersionRange != null && _versions.Items?.CurrentItem != null)
                             {
                                 // Search for the best version
                                 NuGetVersion rangeBestVersion = userRequestedVersionRange.FindBestMatch(versions);
@@ -169,8 +169,8 @@ namespace NuGet.PackageManagement.UI
                                 bool isBestOption = rangeBestVersion.ToString() == selectedVersion.Version.ToString();
                                 if (isBestOption)
                                 {
-                                    PackageDetailControlModel.SelectedVersion = new DisplayVersion(userRequestedVersionRange, rangeBestVersion, additionalInfo: null, isVulnerable: false);
-                                    _versions.Text = comboboxText;
+                                    PackageDetailControlModel.SelectedVersion = new DisplayVersion(userRequestedVersionRange, rangeBestVersion, additionalInfo: null, isVulnerable: selectedVersion.IsVulnerable, isDeprecated: selectedVersion.IsDeprecated);
+                                    _versions.Text = PackageDetailControlModel.SelectedVersion.ToString();
                                 }
                                 else
                                 {
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
@@ -169,6 +169,7 @@ namespace NuGet.PackageManagement.UI
                                 bool isBestOption = rangeBestVersion.ToString() == selectedVersion.Version.ToString();
                                 if (isBestOption)
                                 {
+                                    // Add vulnerable/deprecated labels to non floating/range versions
                                     if (userRequestedVersionRange.OriginalString.StartsWith("(", StringComparison.OrdinalIgnoreCase) ||
                                         userRequestedVersionRange.OriginalString.StartsWith("[", StringComparison.OrdinalIgnoreCase) ||
                                         userRequestedVersionRange.IsFloating)
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
@@ -165,7 +165,8 @@ namespace NuGet.PackageManagement.UI
                             {
                                 // Search for the best version
                                 NuGetVersion rangeBestVersion = userRequestedVersionRange.FindBestMatch(versions);
-                                bool isBestOption = rangeBestVersion.ToString() == _versions.Items[_versions.SelectedIndex].ToString();
+                                DisplayVersion selectedVersion = _versions.Items.CurrentItem as DisplayVersion;
+                                bool isBestOption = rangeBestVersion.ToString() == selectedVersion.Version.ToString();
                                 if (isBestOption)
                                 {
                                     PackageDetailControlModel.SelectedVersion = new DisplayVersion(userRequestedVersionRange, rangeBestVersion, additionalInfo: null, isVulnerable: false);
@@ -248,10 +249,13 @@ namespace NuGet.PackageManagement.UI
             for (int i = 0; i < _versions.Items.Count; i++)
             {
                 DisplayVersion currentItem = _versions.Items[i] as DisplayVersion;
-                if (currentItem != null && (comboboxText == _versions.Items[i].ToString() || _versions.Items[i].ToString() == matchVersion?.ToString()))
+                if (currentItem != null &&
+                    _versions.Items[i] != null && // null separator
+                    (comboboxText.Trim() == currentItem.Version.ToNormalizedString() || // trim extra spaces and compare versions without labels
+                     matchVersion?.ToString() == currentItem.Version.ToNormalizedString()))
                 {
                     _versions.SelectedIndex = i; // This is the "select" effect in the dropdown
-                    PackageDetailControlModel.SelectedVersion = new DisplayVersion(userRange, matchVersion, additionalInfo: null, isVulnerable: currentItem.IsVulnerable);
+                    PackageDetailControlModel.SelectedVersion = new DisplayVersion(userRange, matchVersion, additionalInfo: null, isDeprecated: currentItem.IsDeprecated, isVulnerable: currentItem.IsVulnerable);
                 }
             }
         }
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
@@ -169,7 +169,16 @@ namespace NuGet.PackageManagement.UI
                                 bool isBestOption = rangeBestVersion.ToString() == selectedVersion.Version.ToString();
                                 if (isBestOption)
                                 {
-                                    PackageDetailControlModel.SelectedVersion = new DisplayVersion(userRequestedVersionRange, rangeBestVersion, additionalInfo: null, isVulnerable: selectedVersion.IsVulnerable, isDeprecated: selectedVersion.IsDeprecated);
+                                    if (userRequestedVersionRange.OriginalString.StartsWith("(", StringComparison.OrdinalIgnoreCase) ||
+                                        userRequestedVersionRange.OriginalString.StartsWith("[", StringComparison.OrdinalIgnoreCase) ||
+                                        userRequestedVersionRange.IsFloating)
+                                    {
+                                        PackageDetailControlModel.SelectedVersion = new DisplayVersion(userRequestedVersionRange, rangeBestVersion, additionalInfo: null);
+                                    }
+                                    else
+                                    {
+                                        PackageDetailControlModel.SelectedVersion = new DisplayVersion(userRequestedVersionRange, rangeBestVersion, additionalInfo: null, isVulnerable: selectedVersion.IsVulnerable, isDeprecated: selectedVersion.IsDeprecated);
+                                    }
                                     _versions.Text = PackageDetailControlModel.SelectedVersion.ToString();
                                 }
                                 else
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
@@ -161,7 +161,7 @@ namespace NuGet.PackageManagement.UI
                         if (_versions.SelectedIndex >= 0 && isAVersion)
                         {
                             // Check if the user typed a custom version
-                            if (isUserInputValidNuGetVersionRange && userRequestedVersionRange != null && _versions.Items?.CurrentItem != null)
+                            if (isUserInputValidNuGetVersionRange)
                             {
                                 // Search for the best version
                                 NuGetVersion rangeBestVersion = userRequestedVersionRange.FindBestMatch(versions);
@@ -259,10 +259,8 @@ namespace NuGet.PackageManagement.UI
             for (int i = 0; i < _versions.Items.Count; i++)
             {
                 DisplayVersion currentItem = _versions.Items[i] as DisplayVersion;
-                if (currentItem != null &&
-                    _versions.Items[i] != null && // null separator
-                    (comboboxText.Trim() == currentItem.Version.ToNormalizedString() || // trim extra spaces and compare versions without labels
-                     matchVersion?.ToString() == currentItem.Version.ToNormalizedString()))
+                if (currentItem != null && // null, this represent a bar in UI in the Versions combobox 
+                    (comboboxText.Trim() == currentItem.Version.ToNormalizedString() || matchVersion?.ToString() == currentItem.Version.ToNormalizedString())) // trim extra spaces and compare versions without labels
                 {
                     _versions.SelectedIndex = i; // This is the "select" effect in the dropdown
                     PackageDetailControlModel.SelectedVersion = new DisplayVersion(userRange, matchVersion, additionalInfo: null, isDeprecated: currentItem.IsDeprecated, isVulnerable: currentItem.IsVulnerable);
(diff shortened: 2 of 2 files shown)


nkolev92 (dismissed):

jeffkl (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageDetailControlModel.cs
```diff
@@ -252,7 +252,7 @@ private bool VersionsFilter(object o)
             if ((UserInput.StartsWith("(", StringComparison.OrdinalIgnoreCase) || UserInput.StartsWith("[", StringComparison.OrdinalIgnoreCase)) &&
                VersionRange.TryParse(UserInput, out VersionRange userRange))
             {
-                if (o != null && NuGetVersion.TryParse(o.ToString(), out NuGetVersion userVersion))
+                if (o != null && NuGetVersion.TryParse(version.Version.ToString(), out NuGetVersion userVersion))
```
zivkan: `version.Version` is already a `NuGetVersion`: https://github.com/NuGet/NuGet.Client/blob/4f976ea2b3f32926ed585db4b21f6773762892b7/src/NuGet.Clients/NuGet.PackageManagement.UI/DisplayVersion.cs#L114

There's no need to serialize it into a string and then parse it again.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
```diff
@@ -165,7 +165,8 @@ protected override void OnKeyUp(KeyEventArgs e)
                             {
                                 // Search for the best version
                                 NuGetVersion rangeBestVersion = userRequestedVersionRange.FindBestMatch(versions);
-                                bool isBestOption = rangeBestVersion.ToString() == _versions.Items[_versions.SelectedIndex].ToString();
+                                DisplayVersion selectedVersion = _versions.Items.CurrentItem as DisplayVersion;
+                                bool isBestOption = rangeBestVersion.ToString() == selectedVersion.Version.ToString();
```
zivkan: `value as SomeType` will be null is `value` cannot be cast to `SomeType`. Since you then don't check `selectedVersion` to see if it's null, then this will throw a NullReferenceException, which is hard to figure out why a bug exists. But if you cast with `(DisplayVersion)value`, then we'll get an exception that clearly says something like "WhateverType can't be cast to SomeType`, making it much easier to understand.

If `#nullable enable` was used in this file, the compiler would tell us about this logic bug.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
```diff
@@ -248,10 +249,13 @@ private void SetComboboxCurrentVersion(string comboboxText, IEnumerable<NuGetVer
             for (int i = 0; i < _versions.Items.Count; i++)
             {
                 DisplayVersion currentItem = _versions.Items[i] as DisplayVersion;
-                if (currentItem != null && (comboboxText == _versions.Items[i].ToString() || _versions.Items[i].ToString() == matchVersion?.ToString()))
+                if (currentItem != null &&
+                    _versions.Items[i] != null && // null separator
+                    (comboboxText.Trim() == currentItem.Version.ToNormalizedString() || // trim extra spaces and compare versions without labels
+                     matchVersion?.ToString() == currentItem.Version.ToNormalizedString()))
```
kartheekp-ms: I see that `currentItem.Version.ToNormalizedString()` is invoked twice in this code snippet. Can we just invoke the method once and cache the result so that we can use the cached value for comparison?

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
martinrrm: I created an issue for this https://github.com/NuGet/Home/issues/12738. First I was planning in adding that in this PR but I want to merge this in a QB insertion and adding that could be to much for the bar. Thanks for the reminder of adding that, I'll try to add that to every PR

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
```diff
@@ -161,15 +161,26 @@ protected override void OnKeyUp(KeyEventArgs e)
                         if (_versions.SelectedIndex >= 0 && isAVersion)
                         {
                             // Check if the user typed a custom version
-                            if (isUserInputValidNuGetVersionRange)
+                            if (isUserInputValidNuGetVersionRange && userRequestedVersionRange != null && _versions.Items?.CurrentItem != null)
```
kartheekp-ms: Looking at line no:159 (copied below), I think we can avoid `userRequestedVersionRange != null` null check because `isUserInputValidNuGetVersionRange` will be `true` only when `userRequestedVersionRange` is non null. Please let me know if I missed something.

```cs
bool isUserInputValidNuGetVersionRange = _versions.SelectedIndex >= 0 ? VersionRange.TryParse(comboboxText, out userRequestedVersionRange) : false;
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
kartheekp-ms: @martinrrm - I think Andy's suggestion is not to enable #nullable for this class file in this PR. The feedback is if `_versions.Items.CurrentItem as DisplayVersion;` cast fails then the following line `bool isBestOption = rangeBestVersion.ToString() == selectedVersion.Version.ToString();` will throw an exception because `selectedVersion` is `null`. [as operator reference](https://learn.microsoft.com/dotnet/csharp/language-reference/operators/type-testing-and-cast#as-operator), `The as operator explicitly converts the result of an expression to a given reference or nullable value type. If the conversion isn't possible, the as operator returns null.`

does the following line of code work (added [`Null-conditional operators ?.`](https://learn.microsoft.com/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and-) next to the `selectedVersion`)?

```cs
bool isBestOption = rangeBestVersion.ToString() == selectedVersion?.Version.ToString();
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
kartheekp-ms: This `if` block is nested inside another if block that executes only when `isAVersion` is true. For `isAVersion` to be `true` then `_versions.Items.CurrentItem != null`. Therefore, the same condition is not required to be evaluated in the nested if block.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
zivkan: Right, nullable checking will tell you the code is unsafe. I'm saying don't write unsafe code, even when the compiler doesn't warn you.

But I disagree about using `selectedVersion?.Version`.  I feel like it kind of misses the point I was trying to make. If our code only expects `.SelectedItem` to be of type `DisplayVersion` and never anything else, then using `as` means that when `.SelectedItem` is any different type, it'll go into any null handling and we won't understand why the customer is having problems. If you do an explicit `(DisplayVersion)whatever.SelectedItem` then we get a very clear exception message.

Only use `value as Type` casting when we expect `value` to be one of multiple types. Although `if (value is Type1 t1) { } else if (value is Type2 t2) { }` is a much better pattern, so frankly I don't see much use for the `as` keyword any longer.

aortiz-msft: This pattern is followed in several places in this file. Would it be okay if Martin fixes all instances in a subsequent PR?

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
```diff
@@ -248,10 +259,13 @@ private void SetComboboxCurrentVersion(string comboboxText, IEnumerable<NuGetVer
             for (int i = 0; i < _versions.Items.Count; i++)
             {
                 DisplayVersion currentItem = _versions.Items[i] as DisplayVersion;
-                if (currentItem != null && (comboboxText == _versions.Items[i].ToString() || _versions.Items[i].ToString() == matchVersion?.ToString()))
+                if (currentItem != null &&
```
aortiz-msft: I'm not finding the new layout readable. Could we group the conditions a bit better so it's easier to read?

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
```diff
@@ -248,10 +259,13 @@ private void SetComboboxCurrentVersion(string comboboxText, IEnumerable<NuGetVer
...
+                    _versions.Items[i] != null && // null separator
```
aortiz-msft: What'd does this comment is trying to say? I wonder if we could be a bit more specific/descriptive.

aortiz-msft (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
martinrrm: We use "null" in the UI to represent the bar that separates Latest/Installed versions with the rest of versions. 
image

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
zivkan: Sure, a PR dedicated to fixing the pattern is ok, but I'd still like to see new PRs avoid the bad patterns in new or modified code. The so called "boy scout rule".

martinrrm: Changed to use casting instead of as, there is a previous check for null, but the code is not clear. I really want to improve the code here in the next quality week. I did test the `#nullable` and there are a lot of improvements here for readability and nulls.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
```diff
@@ -248,10 +259,11 @@ private void SetComboboxCurrentVersion(string comboboxText, IEnumerable<NuGetVer
             for (int i = 0; i < _versions.Items.Count; i++)
             {
                 DisplayVersion currentItem = _versions.Items[i] as DisplayVersion;
-                if (currentItem != null && (comboboxText == _versions.Items[i].ToString() || _versions.Items[i].ToString() == matchVersion?.ToString()))
+                if (currentItem != null && // null, this represent a bar in UI in the Versions combobox 
+                    (comboboxText.Trim() == currentItem.Version.ToNormalizedString() || matchVersion?.ToString() == currentItem.Version.ToNormalizedString())) // trim extra spaces and compare versions without labels
                 {
                     _versions.SelectedIndex = i; // This is the "select" effect in the dropdown
-                    PackageDetailControlModel.SelectedVersion = new DisplayVersion(userRange, matchVersion, additionalInfo: null, isVulnerable: currentItem.IsVulnerable);
+                    PackageDetailControlModel.SelectedVersion = new DisplayVersion(userRange, matchVersion, additionalInfo: null, isDeprecated: currentItem.IsDeprecated, isVulnerable: currentItem.IsVulnerable);
```
donnie-msft, line 266: Can't we bail out of the loop once a version is selected? Does this select all matching versions? I believe what I see here is that it will select every version that matches, then the final state is the last match will be selected. If that's the desired behavior, there should be a more explicit way to do that.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
```diff
@@ -165,11 +165,22 @@ protected override void OnKeyUp(KeyEventArgs e)
...
+                                DisplayVersion selectedVersion = (DisplayVersion)_versions.Items.CurrentItem;
+                                bool isBestOption = rangeBestVersion.ToString() == selectedVersion.Version.ToString();
                                 if (isBestOption)
                                 {
-                                    PackageDetailControlModel.SelectedVersion = new DisplayVersion(userRequestedVersionRange, rangeBestVersion, additionalInfo: null, isVulnerable: false);
-                                    _versions.Text = comboboxText;
+                                    // Add vulnerable/deprecated labels to non floating/range versions
+                                    if (userRequestedVersionRange.OriginalString.StartsWith("(", StringComparison.OrdinalIgnoreCase) ||
+                                        userRequestedVersionRange.OriginalString.StartsWith("[", StringComparison.OrdinalIgnoreCase) ||
+                                        userRequestedVersionRange.IsFloating)
```
donnie-msft, line 175: The `IsFloating` check is probably a tiny bit cheaper than the string check, so move it up?

```suggestion
                                    if (userRequestedVersionRange.IsFloating ||
                                        userRequestedVersionRange.OriginalString.StartsWith("(", StringComparison.OrdinalIgnoreCase) ||
                                        userRequestedVersionRange.OriginalString.StartsWith("[", StringComparison.OrdinalIgnoreCase))
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
```diff
@@ -165,11 +165,22 @@ protected override void OnKeyUp(KeyEventArgs e)
...
+                                    {
+                                        PackageDetailControlModel.SelectedVersion = new DisplayVersion(userRequestedVersionRange, rangeBestVersion, additionalInfo: null);
+                                    }
+                                    else
+                                    {
+                                        PackageDetailControlModel.SelectedVersion = new DisplayVersion(userRequestedVersionRange, rangeBestVersion, additionalInfo: null, isVulnerable: selectedVersion.IsVulnerable, isDeprecated: selectedVersion.IsDeprecated);
```
donnie-msft, line 181:
```suggestion
                                        PackageDetailControlModel.SelectedVersion = new DisplayVersion(userRequestedVersionRange, rangeBestVersion, additionalInfo: null, selectedVersion.IsVulnerable, selectedVersion.IsDeprecated);
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageDetailControlModel.cs
```diff
@@ -252,12 +252,9 @@ private bool VersionsFilter(object o)
             if ((UserInput.StartsWith("(", StringComparison.OrdinalIgnoreCase) || UserInput.StartsWith("[", StringComparison.OrdinalIgnoreCase)) &&
                VersionRange.TryParse(UserInput, out VersionRange userRange))
             {
-                if (o != null && NuGetVersion.TryParse(o.ToString(), out NuGetVersion userVersion))
+                if (o != null && userRange.Satisfies(version.Version))
```
donnie-msft, line 255:
```suggestion
                if (version != null && userRange.Satisfies(version.Version))
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
```diff
@@ -248,10 +259,11 @@ private void SetComboboxCurrentVersion(string comboboxText, IEnumerable<NuGetVer
             for (int i = 0; i < _versions.Items.Count; i++)
             {
                 DisplayVersion currentItem = _versions.Items[i] as DisplayVersion;
-                if (currentItem != null && (comboboxText == _versions.Items[i].ToString() || _versions.Items[i].ToString() == matchVersion?.ToString()))
+                if (currentItem != null && // null, this represent a bar in UI in the Versions combobox 
+                    (comboboxText.Trim() == currentItem.Version.ToNormalizedString() || matchVersion?.ToString() == currentItem.Version.ToNormalizedString())) // trim extra spaces and compare versions without labels
```
kartheekp-ms, line 263:
```suggestion
string normalizedVersion = currentItem.Version.ToNormalizedString()
                    (comboboxText.Trim() ==  normalizedVersion|| matchVersion?.ToString() == normalizedVersion)) // trim extra spaces and compare versions without labels
```
The idea is to call `currentItem.Version.ToNormalizedString()` method only once.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
martinrrm: We can return once a version is selected

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
martinrrm: We need the named params since there are other parameters that `DisplayVersion` can have in this constructor.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
```diff
@@ -165,11 +165,22 @@ protected override void OnKeyUp(KeyEventArgs e)
                             {
                                 // Search for the best version
                                 NuGetVersion rangeBestVersion = userRequestedVersionRange.FindBestMatch(versions);
-                                bool isBestOption = rangeBestVersion.ToString() == _versions.Items[_versions.SelectedIndex].ToString();
+                                DisplayVersion selectedVersion = (DisplayVersion)_versions.Items.CurrentItem;
+                                bool isBestOption = rangeBestVersion.ToString() == selectedVersion.Version.ToString();
```
zivkan, line 169: are `rangeBestVersion` and `selectedVersion.Version` the same type?  Can we compare them directly, without allocating strings with `ToString()`?

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/ProjectView.xaml.cs
```diff
@@ -165,11 +165,22 @@ protected override void OnKeyUp(KeyEventArgs e)
...
                                 if (isBestOption)
                                 {
-                                    PackageDetailControlModel.SelectedVersion = new DisplayVersion(userRequestedVersionRange, rangeBestVersion, additionalInfo: null, isVulnerable: false);
-                                    _versions.Text = comboboxText;
+                                    // Add vulnerable/deprecated labels to non floating/range versions
+                                    if (userRequestedVersionRange.OriginalString.StartsWith("(", StringComparison.OrdinalIgnoreCase) ||
+                                        userRequestedVersionRange.OriginalString.StartsWith("[", StringComparison.OrdinalIgnoreCase) ||
```
zivkan, line 174: Just thinking out loud here, I'm undecided whether I recommend a change or not. Generally, I don't like string manipulation, I prefer strong data type checking. `VersionRange` has `HasLowerBound` and `IsMinInclusive` booleans, plus another set for upper/max. Those could be checked, rather than checking `OriginalString`.
Alternatively, maybe `.OriginalString[0] == '('` avoids needing to pass in the string comparison, but I'm not sure if the analyzer will flag that as using the current culture, and suggesting to use a culture aware API instead.

zivkan (approved):

aortiz-msft (approved):

martinrrm:
I'll work on this file's in quality week to fix the code debt

PR #5292, opened 2023-07-04
Display NuGet config files in VS Options dialog

dainizzz:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2299

## Description

Adds a new page in the Visual Studio options dialog that displays a list of the NuGet configuration files that are relevant to the user's solution.

Configuration files options page:
Screenshot of Configuration Files Options Window
Error message:
Screenshot of Error Message
## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception
No new APIs were added and it's difficult to write tests for the UI.
  - **OR**
  - [ ] N/A 

PLOC test
image

- **Documentation**
  
  - **OR**
  - [x] N/A


zivkan:
@dainizzz since you merged the dev branch into this branch rather than rebasing your branch, now github is showing those dev branch changes as changes in your branch, making it harder to see what will actually change in the PR.

Can you please reset your branch to before the merge (not what git calls a revert. A revert is a new commit. We want to completely remove the merge commit), and then use rebase + force push to resolve any merge conflicts.  Thanks.

martinrrm:
@dainizzz Can you add screenshots of the design to the description? Maybe the error message too

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsOptionsPage.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// immediately hits enter causing the window to close without firing the combobox LostFocus event.
+        /// </summary>
+        /// <param name="e">Event arguments.</param>
+        protected override void OnApply(PageApplyEventArgs e)
+        {
+            //bool wasApplied = PackageSourceMappingControl.ApplyChangedSettings();
+            //if (!wasApplied)
+            //{
+            //    e.ApplyBehavior = ApplyKind.CancelNoNavigate;
+            //}
```
jeffkl: Is this supposed to be removed?

dainizzz: Yes, this was a mistake! I'm removing it!

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsControl.xaml.cs
```diff
@@ -0,0 +1,69 @@
...
+            ConfigPathsWindow = new ConfigPathsWindowViewModel();
+            OpenConfigurationFile = new DelegateCommand(ExecuteOpenConfigurationFile, IsSelectedPath, NuGetUIThreadHelper.JoinableTaskFactory);
+            DataContext = this;
+            InitializeComponent();
+        }
+
+        private bool IsSelectedPath(object obj)
+        {
+            return _configurationPaths.SelectedItem != null;
+        }
```
kartheekp-ms: The parameter `obj` is unused. Is that expected? If yes, do we need that parameter? If not, can we just remove it.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsControl.xaml.cs
```diff
@@ -0,0 +1,69 @@
...
+using NuGet.Common;
+
+namespace NuGet.PackageManagement.UI.Options
+{
+    /// <summary>
+    /// Interaction logic for ConfigPathsControl.xaml
+    /// </summary>
+    public partial class ConfigPathsControl : UserControl
+    {
+        public ConfigPathsWindowViewModel ConfigPathsWindow { get; set; }
```
kartheekp-ms: The name `ConfigPathsWindow` doesn't seem correct to me. Any reason why this this property is public?

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsWindowViewModel.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            IComponentModel componentModelMapping = NuGetUIThreadHelper.JoinableTaskFactory.Run(ServiceLocator.GetComponentModelAsync);
+            var settings = componentModelMapping.GetService<Configuration.ISettings>();
+            IReadOnlyList<string> configPaths = settings.GetConfigFilePaths().ToList();
+            ConfigPaths.AddRange(CreateViewModels(configPaths));
+        }
+
+        private ObservableCollection<ConfigPathsViewModel> CreateViewModels(IReadOnlyList<string> configPaths)
+        {
+            var configPathsCollection = new ObservableCollection<ConfigPathsViewModel>();
```
kartheekp-ms:
```suggestion
            var configPathsCollection = new ObservableCollection<ConfigPathsViewModel>(configPaths.Count);
```
[doc](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.ireadonlycollection-1.count?view=net-7.0#system-collections-generic-ireadonlycollection-1-count)

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsWindowViewModel.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        public void OpenConfigFile(ConfigPathsViewModel selectedPath)
+        {
+            var componentModel = NuGetUIThreadHelper.JoinableTaskFactory.Run(ServiceLocator.GetComponentModelAsync);
+            var projectContext = componentModel.GetService<INuGetProjectContext>();
+            if (!File.Exists(selectedPath.ConfigPath))
+            {
+                var error = new FileNotFoundException(selectedPath.ConfigPath);
+                var errorEvent = new TelemetryEvent("ConfigPathsFileNotFoundException");
```
kartheekp-ms: I am curious to learn more about, why do we need telemetry for this exception?

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsWindowViewModel.cs
```diff
@@ -0,0 +1,59 @@
...
+            }
+
+            return configPathsCollection;
+        }
+
+        public void OpenConfigFile(ConfigPathsViewModel selectedPath)
+        {
+            var componentModel = NuGetUIThreadHelper.JoinableTaskFactory.Run(ServiceLocator.GetComponentModelAsync);
+            var projectContext = componentModel.GetService<INuGetProjectContext>();
+            if (!File.Exists(selectedPath.ConfigPath))
```
kartheekp-ms: I assume this line of code `File.Exists` is executed on UI thread. FYI. This operation is not cheap and caused UI delays in the past for other components inside Visual Studio.

kartheekp-ms (commented):
(still reviewing this PR. left few comments to start with)

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsWindowViewModel.cs
```diff
@@ -0,0 +1,59 @@
...
+            {
+                var error = new FileNotFoundException(selectedPath.ConfigPath);
+                var errorEvent = new TelemetryEvent("ConfigPathsFileNotFoundException");
+                TelemetryActivity.EmitTelemetryEvent(errorEvent);
```
donnie-msft:
```suggestion
                TelemetryUtility.PostFaultAsync(error, nameof(ConfigPathsWindowViewModel))
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsWindowViewModel.cs
jeffkl: Team Review:  Let's remove the telemetry event for now and add it later when the self assessment process is complete and we determine the event is needed.

jeffkl (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsWindowViewModel.cs
nkolev92: If you follow the codepath that's calling the file open you'll see that we already do that, https://github.com/NuGet/NuGet.Client/blob/4187eab2d176d6fccb6f0cfc45fc71fd2e750708/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VsCommonOperations.cs#L36-L51. 

So this File.Exists would happen anyways, so it's probably not adding anything perf wise. 

Another note for performance related things is that we shouldn't be too worried about paths that aren't hot. You can't always dodge doing work on the UI thread. 

This will only be triggered when someone *clicks*. Their focus is already there and the 1ms that the check could take is negligible.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsViewModel.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace NuGet.PackageManagement.UI.Options
+{
+    public class ConfigPathsViewModel
```
jebriede: This view model is for a single configuration path so a more appropriate name would be singular rather than ConfigPaths since this represents a single path.

```suggestion
    public class ConfigPathViewModel
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsWindowViewModel.cs
dainizzz: Hi Kartheek! I looked over the documentation, but I don't understand how to use `configPaths.Count` here. In this line, I'm just creating a new empty `ObservableCollection`. Later on, I create `ConfigPathsViewModel`s from the list of `configPaths`, which I add to the `ObservableCollection`. Do you mean that it would be better to use a for loop based on the count of `configPaths` instead of a foreach loop?

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsControl.xaml.cs
dainizzz: I'll try to come up with a more clear name! `ConfigPathsWindow` is public because it's used by `ConfigPathsControl.xaml`.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsWindowViewModel.cs
dainizzz: I'm checking if the file exists here to handle an edge case where a user has the options window open, selects a file, deletes or moves the selected file, then tries to open that same file. In that case, `fullPath` is not null (and can't be null since this command can only be executed when there is a `selectedPath`), so `OpenFile()` doesn't throw an exception. At the same time, the `if (dte.ItemOperations != nulll && File.Exists(fullPath))` condition is not met, since the file no longer exists at that location, so nothing happens. As a result, the user doesn't get any sort of error message when this occurs.

I'll add a comment explaining this.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsViewModel.cs
jebriede: I'd also consider removing this ViewModel class altogether since it's only wrapping a string and doesn't provide much value.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsControl.xaml.cs
```diff
@@ -0,0 +1,69 @@
...
+        public ConfigPathsViewModel SelectedPath { get; set; }
+        public ICommand OpenConfigurationFile { get; set; }
+
+        public ConfigPathsControl()
+        {
+            ConfigPathsWindow = new ConfigPathsWindowViewModel();
+            OpenConfigurationFile = new DelegateCommand(ExecuteOpenConfigurationFile, IsSelectedPath, NuGetUIThreadHelper.JoinableTaskFactory);
+            DataContext = this;
```
jebriede: Your DataContext for your View (which this class is your view), should be your ViewModel, not the View itself.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsControl.xaml.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// </summary>
+    public partial class ConfigPathsControl : UserControl
+    {
+        public ConfigPathsWindowViewModel ConfigPathsWindow { get; set; }
+        public ConfigPathsViewModel SelectedPath { get; set; }
+        public ICommand OpenConfigurationFile { get; set; }
+
+        public ConfigPathsControl()
+        {
+            ConfigPathsWindow = new ConfigPathsWindowViewModel();
```
jebriede: This variable name for the ViewModel is not very clear. The ViewModel does not represent the Window, so the name is misleading. Please rename to ConfigPathsViewModel.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsWindowViewModel.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Linq;
+using Microsoft.VisualStudio.ComponentModelHost;
+using Microsoft.VisualStudio.Services.Common;
+using NuGet.Common;
+using NuGet.ProjectManagement;
+using NuGet.VisualStudio;
+
+namespace NuGet.PackageManagement.UI.Options
+{
+    public class ConfigPathsWindowViewModel
```
jebriede: The "Window" part of the name should be removed. ViewModels aren't tied to their View implementation so the View Model should be named after what its purpose is, not which UI uses it.

One way of thinking about this is to think of how a ViewModel could technically be used by a Window or a Command Line utility or a web interface, etc.

```suggestion
    public class ConfigPathsViewModel
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsControl.xaml.cs
```diff
@@ -0,0 +1,69 @@
...
+            return _configurationPaths.SelectedItem != null;
+        }
+
+        private void ExecuteOpenConfigurationFile(object obj)
+        {
+            SelectedPath = (ConfigPathsViewModel)_configurationPaths.SelectedItem;
+            ConfigPathsWindow.OpenConfigFile(SelectedPath);
+            var evt = new NavigatedTelemetryEvent(NavigationType.Button, NavigationOrigin.Options_ConfigurationFiles_Open);
+            TelemetryActivity.EmitTelemetryEvent(evt);
+        }
```
jebriede: This should be handled by the ViewModel.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsControl.xaml.cs
```diff
@@ -0,0 +1,69 @@
...
+namespace NuGet.PackageManagement.UI.Options
+{
+    /// <summary>
+    /// Interaction logic for ConfigPathsControl.xaml
+    /// </summary>
+    public partial class ConfigPathsControl : UserControl
+    {
+        public ConfigPathsWindowViewModel ConfigPathsWindow { get; set; }
+        public ConfigPathsViewModel SelectedPath { get; set; }
+        public ICommand OpenConfigurationFile { get; set; }
```
jebriede: Once you have the ViewModel as the data context for this control, this ICommand should move to the ViewModel.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsControl.xaml.cs
```diff
@@ -0,0 +1,69 @@
...
+
+namespace NuGet.PackageManagement.UI.Options
+{
+    /// <summary>
+    /// Interaction logic for ConfigPathsControl.xaml
+    /// </summary>
+    public partial class ConfigPathsControl : UserControl
+    {
+        public ConfigPathsWindowViewModel ConfigPathsWindow { get; set; }
+        public ConfigPathsViewModel SelectedPath { get; set; }
```
jebriede: The state of which path is selected should be handled by the ViewModel.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsControl.xaml.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            ConfigPathsWindow = new ConfigPathsWindowViewModel();
+            OpenConfigurationFile = new DelegateCommand(ExecuteOpenConfigurationFile, IsSelectedPath, NuGetUIThreadHelper.JoinableTaskFactory);
+            DataContext = this;
+            InitializeComponent();
+        }
+
+        private bool IsSelectedPath(object obj)
+        {
+            return _configurationPaths.SelectedItem != null;
```
jebriede: State management should not be handled by the View and its controls. Instead, bind the SelectedItem of the list to a property on the ViewModel and use that in your ViewModel code to determine which item is selected.

The state management of the application logic should be in the ViewModel, which includes item is active or selected.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsControl.xaml.cs
```diff
@@ -0,0 +1,69 @@
...
+            SelectedPath = (ConfigPathsViewModel)_configurationPaths.SelectedItem;
+            ConfigPathsWindow.OpenConfigFile(SelectedPath);
+            var evt = new NavigatedTelemetryEvent(NavigationType.DoubleClick, NavigationOrigin.Options_ConfigurationFiles_ListItem);
+            TelemetryActivity.EmitTelemetryEvent(evt);
+        }
+
+        internal void InitializeOnActivated(CancellationToken cancellationToken)
+        {
+            ConfigPathsWindow.ConfigPaths.Clear();
+            ConfigPathsWindow.SetConfigPaths();
```
jebriede: The ViewModel should know what it needs to be initialized and that should be called from here.

Two ways I can think of handling this off the top of my head:
1) Create a new view model each time this InitializeOnActivated is called. If the ViewModel is passed in as a parameter to the View, this would not be ideal.
2) Have a method on the ViewModel for initializing or resetting and call it here.

That way, the idea that the collection on the VM is cleared and the config paths are reloaded is encapsulated in the ViewModel rather than the View needing to instruct it how to initialize.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsControl.xaml.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        private void ExecuteOpenExternalLink(object sender, ExecutedRoutedEventArgs e)
+        {
+            var hyperlink = e.OriginalSource as Hyperlink;
+            if (hyperlink != null && hyperlink.NavigateUri != null)
```
jebriede:
```suggestion
            if (e.OriginalSource is Hyperlink hyperlink && hyperlink.NavigateUri != null)
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsOptionsPage.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        /// <summary>
+        /// Gets the Windows Presentation Foundation (WPF) child element to be hosted inside the Options dialog page.
+        /// </summary>
+        /// <returns>The WPF child element.</returns>
+        protected override UIElement Child => _configPathsControl.Value;
+
+        public ConfigPathsOptionsPage()
+        {
+            _configPathsControl = new Lazy<ConfigPathsControl>(() => new ConfigPathsControl());
```
jebriede: What about creating a new instance of ViewModel for the control and passing it in the View's constructor?

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigPathsWindowViewModel.cs
```diff
@@ -0,0 +1,59 @@
...
+                TelemetryActivity.EmitTelemetryEvent(errorEvent);
+                MessageHelper.ShowErrorMessage(error.Message, Resources.ShowError_FileNotFound);
+            }
+            _ = projectContext.ExecutionContext.OpenFile(selectedPath.ConfigPath);
+        }
+
+        public ConfigPathsWindowViewModel()
+        {
+            ConfigPaths = new ObservableCollection<ConfigPathsViewModel>();
+        }
```
jebriede: The constructor should be at the top, as the first method in the class.

jebriede (changes requested):

src/NuGet.Clients/NuGet.Tools/NuGetPackage.cs
```diff
@@ -52,8 +52,9 @@ namespace NuGetVSExtension
         Window = "{34E76E81-EE4A-11D0-AE2E-00A0C90FFFC3}", // this is the guid of the Output tool window, which is present in both VS and VWD
         Orientation = ToolWindowOrientation.Right)]
     [ProvideOptionPage(typeof(GeneralOptionPage), "NuGet Package Manager", "General", 113, 115, true)]
-    [ProvideOptionPage(typeof(PackageSourceOptionsPage), "NuGet Package Manager", "Package Sources", 113, 114, true, Sort = 0)]
-    [ProvideOptionPage(typeof(PackageSourceMappingOptionsPage), "NuGet Package Manager", "Package Source Mapping", 113, 116, true, Sort = 1)]
+    [ProvideOptionPage(typeof(ConfigPathsOptionsPage), "NuGet Package Manager", "Config Paths", 113, 117, true, Sort = 0)]
```
donnie-msft: The page name here doesn't match the screenshot. I'd expect the 3rd parameter to be "Configuration Files". Is it actually showing that currently? If it is I wonder where it's pulling it from 🤔

donnie-msft (commented):
There's a lot of feedback already, so I'm not sure how much you'll be able to complete with the time you have remaining. Our repo has not been very strict on MVVM guidance, so I don't consider partial MVVM to be blocking. You may want to talk to management about what they want you to prioritize and discuss the cost/benefit of that if it'll take up too much of your remaining time.
Once the blocking feedback is addressed, I'll review again.

What is necessary to ship, though, are compliance considerations:
- Localization testing (PLOC) see [NuGet.Client UI Guidelines](https://github.com/NuGet/NuGet.Client/blob/b3c4e8094bbb924408180d700ae5b3d37713587c/docs/ui-guidelines.md#localizability-testing)
- [Accessibility Insights FastPass](https://accessibilityinsights.io/docs/windows/getstarted/fastpass/) - typically, I have just been asking for a screenshot of the entire UI element (window/pane).

Happy to review these tools with you if your time permits and you're interested in learning about them now.

src/NuGet.Clients/NuGet.Tools/NuGetPackage.cs
dainizzz: Hi Donnie! The name is coming from VSPackage.resx. I'll fix the naming!

jebriede (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesViewModel.cs
```diff
@@ -0,0 +1,67 @@
...
+        private void OpenConfigFile(string selectedPath)
+        {
+            var componentModel = NuGetUIThreadHelper.JoinableTaskFactory.Run(ServiceLocator.GetComponentModelAsync);
+            var projectContext = componentModel.GetService<INuGetProjectContext>();
+
+            // This check is performed in case the user moves or deletes a config file while they have it selected in the Options window.
+            if (!File.Exists(selectedPath))
+            {
+                var error = new FileNotFoundException(selectedPath);
+                MessageHelper.ShowErrorMessage(error.Message, Resources.ShowError_FileNotFound);
```
donnie-msft: What does this actually show? Did you intend to pass the exception, or the message? 
The `FileNotFoundException.Message` default message appears to be very similar to `Resources.ShowError_FileNotFound`. Do we need both? Is there a screenshot of how this error looks?

```suggestion
                MessageHelper.ShowErrorMessage(Resources.ShowError_FileNotFound);
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml.cs
```diff
@@ -0,0 +1,44 @@
...
+            _viewModel = configurationFiles;
+            DataContext = configurationFiles;
+        }
+
+        private void ListViewItem_MouseDoubleClick(object sender, MouseButtonEventArgs e)
+        {
+            _openButton.Command.Execute(null);
+        }
+
+        internal void InitializeOnActivated(CancellationToken cancellationToken)
```
donnie-msft: I see there are now 2 implementations of `InitializeOnActivated`. Can this one be removed and instead just use the the ViewModel's `InitializeOnActivated` ?

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            _viewModel.SetConfigPaths();
+        }
+
+        private void ExecuteOpenExternalLink(object sender, ExecutedRoutedEventArgs e)
+        {
+            if (e.OriginalSource is Hyperlink hyperlink && hyperlink.NavigateUri != null)
+            {
+                UIUtility.LaunchExternalLink(hyperlink.NavigateUri);
```
donnie-msft: This got me thinking, since we're talking about `nuget.config` files, I wonder if UNC paths are used for those? If so, my understanding is that this utility will simply do nothing. Maybe it's an edge case that's out of scope, but we might need an error message whenever we're unable to launch the config file.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Windows.Input;
+
+namespace NuGet.PackageManagement.UI.Options
+{
+    /// <summary>
+    /// Interaction logic for ConfigurationFilesControl.xaml
+    /// </summary>
+    public partial class ConfigurationFilesControl : UserControl
+    {
+        private ConfigurationFilesViewModel _viewModel;
```
donnie-msft, line 13: I believe we won't need this field if you agree with my other comment to remove `InitializeOnActivated`

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
```diff
@@ -990,11 +990,32 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
...
+    <value>Note: Configuration files are listed in order from highest to lowest priority.</value>
+  </data>
+  <data name="Text_ApplyingConfigurationFiles_DocumentLink" xml:space="preserve">
+    <value>Common NuGet configurations: How settings are applied</value>
+  </data>
+  <data name="VSOptions_Accessibility_ConfigurationFilesList" xml:space="preserve">
+    <value>Configuration files list</value>
+  </data>
+  <data name="VSOptions_Button_Open" xml:space="preserve">
+    <value>_Open</value>
```
donnie-msft, line 1003: Thank you for adding a shortcut key!

donnie-msft (commented):
Looking really good! A few minor questions. Also, I understand you'll be adding the Compliance info I requested soon - thank you for addressing that!

donnie-msft:
Are there any issues tracking creating docs for this new page? We typically don't have to create the docs until later (closer to when the feature goes GA).

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml
```diff
@@ -0,0 +1,67 @@
...
+      </ResourceDictionary.MergedDictionaries>
+      <Style TargetType="ListViewItem">
+        <EventSetter Event="MouseDoubleClick" Handler="ListViewItem_MouseDoubleClick"/>
+      </Style>
+    </ResourceDictionary>
+  </UserControl.Resources>
+  <UserControl.CommandBindings>
+    <CommandBinding
+      Command="{x:Static nuget:PackageManagerControlCommands.OpenExternalLink}"
+      Executed="ExecuteOpenExternalLink" />
```
zivkan: I know very little about WPF, but is this command binding, binding to a static class's static field, expected?

Since it's static, I'm concerned that this binding will overwrite whatever static instance is saved in that field, and therefore  whatever has binded (bound?) to it earlier will now execute the wrong command when clicked?

Maybe I'm off-base because of my limited WPF knowledge, but I feel like either:

1. Use a static routed command, but don't set the "executed" handler, it should be set up by the static itself, so it's always the same for everyone (the field should probably be `readonly` too, or a get-only property, to prevent these types of reassigning issues)

or

2. it should bind to a field/property on the viewmodel, not a static.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml
```diff
@@ -0,0 +1,67 @@
...
+      </Grid.ColumnDefinitions>
+      <Grid Grid.Column="0">
+        <Grid.RowDefinitions>
+          <RowDefinition Height="auto" />
+          <RowDefinition Height="auto" />
+        </Grid.RowDefinitions>
+        <TextBlock Grid.Row="0" TextWrapping="WrapWithOverflow" Text="{x:Static nuget:Resources.Text_ApplyingConfigurationFiles}"/>
+        <TextBlock Grid.Row="1" TextWrapping="WrapWithOverflow">
+          <Hyperlink
+              NavigateUri="https://learn.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#how-settings-are-applied"
```
zivkan:
```suggestion
              NavigateUri="https://learn.microsoft.com/nuget/consume-packages/configuring-nuget-behavior#how-settings-are-applied"
```

Docs links should never have the language code

 Honestly, considering that the docs site used to be docs.microsoft.com and is now learn.microsoft.com, I'm not sure it's a good idea for us to hardcode a link to the current hostname, in case it changes again. Plus, hardcoding it assumes that we'll never change the rest of the URL either. I think we should use an aka..ms link so that if the docs URL ever changes in the future, we can update the aka.ms link, and all versions of VS that use the link will Just Work.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml.cs
zivkan: Looking at references to this method name, it's only used by a doc link, not nuget.config files.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        internal void InitializeOnActivated(CancellationToken cancellationToken)
+        {
+            _viewModel.SetConfigPaths();
+        }
+
+        private void ExecuteOpenExternalLink(object sender, ExecutedRoutedEventArgs e)
+        {
+            if (e.OriginalSource is Hyperlink hyperlink && hyperlink.NavigateUri != null)
```
zivkan: Currently this method only handles when `e.OriginalSource` is of type `Hyperlink`. If this method gets called with any other type, it will silently do nothing. Therefore, if someone tries to use it in a new control, without testing, from users point of view they click something and nothing happens, and our telemetry won't show anything either.

As far as I can tell, this method is only called when a specific `RoutedCommand` is executed. Therefore, it shouldn't be executed when other links or buttons are clicked. Therefore, if we make this command handler throw an exception, or log fault telemetry, I don't expect it will have unintended consequences.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesViewModel.cs
```diff
@@ -0,0 +1,67 @@
...
+        private bool IsSelectedPath()
+        {
+            return SelectedPath != null;
+        }
+
+        private void ExecuteOpenConfigurationFile()
+        {
+            OpenConfigFile(SelectedPath);
+        }
+        internal void InitializeOnActivated(CancellationToken cancellationToken)
```
zivkan: nit: 1 whitespace line between methods

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesViewModel.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            SetConfigPaths();
+        }
+
+        public void SetConfigPaths()
+        {
+            IComponentModel componentModelMapping = NuGetUIThreadHelper.JoinableTaskFactory.Run(ServiceLocator.GetComponentModelAsync);
```
zivkan: 1. using `NuGetUIThreadHelper.JoinableTaskFactory` basically makes this untestable.
2. The component model is used twice, here in `SetConfigPaths`, and again in `OpenConfigFile`, but the component model is itself a singleton. Therefore, it could be obtained once, passed to this class's constructor. This has the side effect that `IComponentModel` can be mocked and therefore tests can be written.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesViewModel.cs
dainizzz: Here's a screenshot of what the error message looks like:
Options Error Message
`error.Message` is the file path, and `Resources.ShowError_FileNotFound` is "The following file could not be found:".

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml.cs
```diff
@@ -0,0 +1,44 @@
...
+        public ConfigurationFilesControl(ConfigurationFilesViewModel configurationFiles)
+        {
+            InitializeComponent();
+            _viewModel = configurationFiles;
+            DataContext = configurationFiles;
+        }
+
+        private void ListViewItem_MouseDoubleClick(object sender, MouseButtonEventArgs e)
+        {
+            _openButton.Command.Execute(null);
```
kartheekp-ms, line 24: Curious to understand why is `null` argument passed to the `Execute` method?

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml
jebriede: I agree with @zivkan. I'm used to using fwlinks in code to link to docs. I suppose an aka.ms would work as well. Either way, we should use a link that is redirected so we can update it if it every changes without having to ship a code update.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml.cs
kartheekp-ms: `cancellationToken` parameter is unused. Is that okay? Please let me know if I am missing something.

dainizzz: I tried seeing if I could use the `InitializeOnActivated` in the ViewModel instead of the one in the Control, but it's actually called from `OnActivateAsync` in ConfigurationFilesOptionsPage, and I'm having trouble accessing it from the ViewModel there.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesViewModel.cs
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
```
kartheekp-ms, line 4: May be add [`#nullable enable`](https://learn.microsoft.com/dotnet/csharp/nullable-references) here.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesViewModel.cs
```diff
@@ -0,0 +1,67 @@
...
+using NuGet.ProjectManagement;
+using NuGet.VisualStudio;
+
+namespace NuGet.PackageManagement.UI.Options
+{
+    public class ConfigurationFilesViewModel
+    {
+        public ObservableCollection<string> ConfigurationFilesCollection { get; private set; }
+        public string SelectedPath { get; set; }
+        public ICommand OpenConfigurationFile { get; set; }
```
kartheekp-ms: This property is initialized in the constructor and hence I think it can be marked as `readonly`.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesViewModel.cs
kartheekp-ms: How about doing a string.Format or something similar to avoid creating new instance of `FileNotFoundException` just to obtain the error message?
```cs
var message = string.Format(Resources.ShowError_FileNotFound, selectedPath);
MessageHelper.ShowErrorMessage(message);
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml
jebriede: The preferred way to handle this would be number 2 from what @zivkan suggested.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesViewModel.cs
jebriede: Since this is a property, the way to achieve this would be to remove the setter:

`public ICommand OpenConfigurationFile { get; }`

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml
martinrrm: We actually do this in 4 other places, example: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/DetailControl.xaml#L24-L26 and it works. We decided to use the same implementation that we currently use in the project, should we change this in all of the places?

aortiz-msft: Yes please. I'd like us to improve the MVVM delineation in PM UI as we do work in it.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml.cs
dainizzz: I checked the documentation, and it says this argument is "Data used by the command. If the command does not require data to be passed, this object can be set to null."

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml.cs
dainizzz: Martin and I took a look at [PackageSourceMappingOptionsControl.xaml.cs](https://github.com/NuGet/NuGet.Client/blob/972557a7258d77dbaed3b7a649d2f13a4083f94c/src/NuGet.Clients/NuGet.PackageManagement.UI/Options/PackageSourceMappingOptionsControl.xaml.cs#L46C81-L46C81) and we saw that `cancellationToken` was also not used there, so we felt it was OK to remove the parameter.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesOptionsPage.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.ComponentModel;
+using System.Runtime.InteropServices;
+using System.Threading;
+using System.Windows;
+using Microsoft.VisualStudio.Shell;
+
+namespace NuGet.PackageManagement.UI.Options
+{
+    [Guid("C17B308A-00BB-446E-9212-2D14E1005985")]
+
```
donnie-msft: nit: whitespace

donnie-msft (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml
jebriede: I spent some time looking at this with @dainizzz in further detail. I better understand why this RoutedCommand is defined here now.

The RoutedCommand is designed to handle an event that is passed up the element tree to somewhere a CommandBinding is defined and then uses the Execute and CanExecute methods to handle the command. This is the way the DetailsControl is set up to use it and it works... however, this adds extra complexity and indirection that isn't necessary.

The [Hyperlink](https://learn.microsoft.com/en-us/dotnet/api/system.windows.documents.hyperlink?view=windowsdesktop-7.0) object has an event [RequestNavigate](https://learn.microsoft.com/en-us/dotnet/api/system.windows.documents.hyperlink.requestnavigate?view=windowsdesktop-7.0#system-windows-documents-hyperlink-requestnavigate) that is responsible for handling the click of a hyperlink that has a NavigateUri attribute specified. The NavigateUri gets passed in the RequestNavigateEventArgs in its Uri property.

We should be handling this RequestNavigate event on the UI and delegating the responsibility of validating the Uri and launching the external process to the UIUtility.

Doing this allows us to get rid of our dependency on the PackageManagerControlCommands altogether, it removes the need to create a RoutedCommand and specify the CommandBinding on the UserControl, it removes the need for us to specify a Command on the Hyperlink and instead we can rely on the built-in event that's designed to handle URI navigation. The click of the hyperlink also doesn't have to walk up the element tree to find the user control's CommandBinding anymore, and the click of the hyperlink can be handled immediately by the RequestNavigate event handler. Additionally, it's reusable -- more than one hyperlink on the same UI can point to the same RequestNavigate handler.

And lastly, I asked myself if it's best to put the navigation of to the URI in the View or in the ViewModel and I think the View is the appropriate place. Why? Because if this same ViewModel were to be used for a CommandLine utility or a website, the way the URI would be navigated to would be vastly different. It would be dependent on how the UI was designed as to how it wanted to display the content from that URI, so the responsibility of how to navigate appropriately lies with the View.

However, an argument could be made for making the URI as a property in the ViewModel and binding to it in the View.

Example (not tested) in XAML:
```
<Hyperlink
    NavigateUri="{Binding HelpUri}"
    RequestNavigate="{Hyperlink_OnRequestNavigate}"/>
```

Then in the ViewModel (not tested) :
```
    public ConfigurationFilesViewModel()
    {
        // Constructor of ViewModel
        // ... other code here
        HelpUri = new Uri("https://aka.ms/nuget/how-settings-are-applied");
    }
    
    public Uri HelpUri { get; }  // Or a better name?
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        private void ExecuteOpenExternalLink(object sender, ExecutedRoutedEventArgs e)
+        {
+            if (e.OriginalSource is Hyperlink hyperlink && hyperlink.NavigateUri != null)
+            {
+                UIUtility.LaunchExternalLink(hyperlink.NavigateUri);
+
+                e.Handled = true;
+            }
+        }
```
jebriede: Based on my [comment above](https://github.com/NuGet/NuGet.Client/pull/5292/files#r1273996057), we should instead have a RequestNavigate handler for the Hyperlink that does the following:

```suggestion
        private void Hyperlink_OnRequestNavigate(object sender, RequestNavigateEventArgs e)
        {
            UIUtility.LaunchExternalLink(e.Uri.AbsoluteUri);
            e.Handled = true;
        }
```

And then the Hyperlink in the XAML should have the `RequestNavigate` attribute pointing to `Hyperlink_OnRequestNavigate` to hook up this event handler.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml
jebriede: I show my suggestion in code [here](https://github.com/NuGet/NuGet.Client/pull/5292/files#r1274003977)

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesViewModel.cs
```diff
@@ -21,11 +21,15 @@ public class ConfigurationFilesViewModel
...
+        private ISettings _settings;
+        private INuGetProjectContext _projectContext;
 
-        public ConfigurationFilesViewModel()
+        public ConfigurationFilesViewModel(ISettings settings, INuGetProjectContext projectContext)
         {
             ConfigurationFilesCollection = new ObservableCollection<string>();
             OpenConfigurationFile = new DelegateCommand(ExecuteOpenConfigurationFile, IsSelectedPath, NuGetUIThreadHelper.JoinableTaskFactory);
+            _settings = settings;
+            _projectContext = projectContext;
```
jebriede: Enforce the scenario where the params are null

```suggestion
            _settings = settings ?? throw new ArgumentNullException(nameof(settings));
            _projectContext = projectContext  ?? throw new ArgumentNullException(nameof(projectContext));
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        internal void InitializeOnActivated()
+        {
+            _viewModel.SetConfigPaths();
+        }
+
+        private void Hyperlink_RequestNavigate(object sender, System.Windows.Navigation.RequestNavigateEventArgs e)
+        {
+            if (sender is Hyperlink)
```
jebriede: We shouldn't have to check if the sender is a hyperlink since the method signature of this method requires that a RequestNavigateEventArgs be passed as the second argument and that's really what we care about. From the RequestNavigateEventArgs we want the Uri. In my comment I left previously I wrote what the code for this method should look like.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml.cs
```diff
@@ -0,0 +1,47 @@
...
+            {
+                UIUtility.LaunchExternalLink(e.Uri);
```
jebriede: As per the comment I left previously

```suggestion
                UIUtility.LaunchExternalLink(e.Uri.AbsoluteUri);
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml.cs
```diff
@@ -0,0 +1,47 @@
...
+                e.Handled = true;
+            }
+            else
+            {
+                throw new InvalidOperationException();
+            }
```
jebriede: This is not necessary. If the method passes RequestNavigateEventArgs, we can get the URI and assume the sender is a Hyperlink.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesViewModel.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        private bool IsSelectedPath()
+        {
+            return SelectedPath != null;
+        }
+
+        private void ExecuteOpenConfigurationFile()
+        {
+            OpenConfigFile(SelectedPath);
+        }
```
jebriede: It would be better to only have 1 method that handles the command to open a config file. The binding should pass the SelectedPath as the CommandParameter and the parameter should be read in the ExecuteOpenConfigurationFile method. The way this is written works, but it introduces an extra method call.

jebriede (commented):
Looks good overall! Left a few comments to reduce some code.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml.cs
dainizzz: I tried this but `LaunchExternalLink` takes a Uri argument, and `AbsoluteUri` is a string. I think it's OK to leave as-is because `LaunchExternalLink` calls `Process.Start(url.AbsoluteUri)`, so it ends up using `AbsoluteUri` anyway.

jebriede: Good catch! In that case, passing the Uri is the best. Thanks!

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesViewModel.cs
dainizzz: Do I still need to pass a command parameter if I always want to use whatever is in `SelectedPath` (i.e. the `SelectedItem` from `ListView`) to open the file?

martinrrm (dismissed):
LGTM

jeffkl (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// <summary>
+    /// Interaction logic for ConfigurationFilesControl.xaml
+    /// </summary>
+    public partial class ConfigurationFilesControl : UserControl
+    {
+        private ConfigurationFilesViewModel _viewModel;
+
+        public ConfigurationFilesControl(ConfigurationFilesViewModel configurationFiles)
+        {
+            InitializeComponent();
```
kartheekp-ms, line 17: `InitializeComponent` method invokes `_viewModel.SetConfigPaths();` method but the `_viewModel` itself is initialized in the next line i.e., line no. 18. If I am understanding correctly a null reference exception will be thrown. Please let me know if I am missing something.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesOptionsPage.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        /// <summary>
+        /// This method is called when VS wants to activate this page.
+        /// ie. when the user opens the tools options page.
+        /// </summary>
+        /// <param name="e">Cancellation event arguments.</param>
+        protected override void OnActivate(CancelEventArgs e)
+        {
+            // The UI caches the settings even though the tools options page is closed.
```
kartheekp-ms, line 41: What will happen if the user updates their settings after UI caches them upon activation? In other words, will the cache be invalidated if the user modifies these settings after opening config options page once?

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml.cs
```diff
@@ -0,0 +1,38 @@
...
+            _viewModel = configurationFiles;
+            DataContext = configurationFiles;
```
kartheekp-ms, line 19: I am curious, why assigning `configurationFiles` to 2 variables.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml.cs
martinrrm: I believe `InitializeComponent` doesn't calls that method, this class first get lazy initialized and when we want to load this page is when `InitializeOnActivated()` is called. 

`InitializeComponent` loads the XAML

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesOptionsPage.cs
martinrrm: `OnActivate` gets called every time you select `Configuration Files` window, so this method will always return the latest data that was saved. This dialog must be closed in case you want to modify the file in VS.

If you modify the file outside VS you must refresh this window

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml.cs
martinrrm: `DataContext` is the variable used by XAML to know what is the data of the element which is the `ViewModel`, but in this case the ViewModel is updated when the page is loaded, not when initialized. So we need to store the ViewModel in a variable and later update it when needed. 

Maybe we could cast DataContext to be of type `ConfigurationFilesViewModel` but I believe this is an error in MVVM design

donnie-msft: To add to Martin's point: The `DataContext` can also be set in XAML. Setting it here certainly looks redundant, but the `_viewModel` is more convenient than casting the `DataContext`.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesOptionsPage.cs
donnie-msft: I think if you load the options window, then modify the config externally, the options window will overwrite the file. Probably worth testing, but also seems low frequency/risk.

donnie-msft (dismissed):

martinrrm (approved):
Approving again, fixed a merge conflict.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/ConfigurationFilesControl.xaml.cs
jebriede: @martinrrm is correct. `InitializeComponent` is a method on the WPF partial class used to load the component. It's can't be referencing _viewModel since it doesn't know that variable exists.

donnie-msft (approved):

PR #5295, opened 2023-07-05
License Dialog looks at correct sources when Package Source Mapping is enabled

### LINKED ISSUE #12723: An error “unable to find metadata of PackageName.1.0.0” occurs when installing package with “packages.config” format
### NuGet Product Used

### Product Version

Dev\6.8.0.2

### Worked before?

_No response_

### Impact

#### Installation Steps:
1.OS: Windows-11-Enterprise-22H2.
2.Install the preinstall cmd from: \\vspreinstall\preinstall\preinstall.cmd.
3. Install  VS from [Pipelines - Run 6.8.0.2 artifacts (visualstudio.com)](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=8006212&view=artifacts&pathAsName=false&type=publishedArtifacts) -> VS15 -> Bootstrapper ->vs_enterprise.exe.

#### Repro Steps:
1. Open VS and create a C# Console Application (.Net Framework 4.8.1) project. 
2. Right-click the project in Solution Explorer and select “Manage NuGet Packages…” menu item to open PM UI. 
3. Click “Browse” tab and select the “nuget.org” from Package source drop-down list.
4. Search a package such as “Newtonsoft.Json” in the filter and click the package in the package list.
5. Select a package version such as “13.0.2” in the “Version” drop-down box and click “Install” button.
6. Select “packages.config” format in the prompt dialog and click “Apply” button in the Preview Changes dialog.

#### Expected:
The package can be installed successfully.

#### Actual:
An error “unable to find metadata of PackageName.1.0.0” shows in the Error List window and the package is failed to install as below screenshot.
image
#### Note:
1.The package failed to install is random since the same package could be installed successfully into the same project after a while.
2.It only reproes on legacy projects with “packages.config” format and doesn’t repro on “PackageReference” format.
3.It doesn’t repro on VS Main\33826.246 + NuGet Client Dev\6.7.0.109.

### Verbose Logs

_No response_

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12723
Fixes: https://github.com/NuGet/Client.Engineering/issues/2413

## Description

- UIActionEngine will now look for enabled Package Source Mapping and if found, limit License Metadata lookup of remote package sources to only the mapped sources.

Created a separate PR https://github.com/NuGet/NuGet.Client/pull/5339 for:
- Added some tests for `PackageDownloader` due to missing test coverage
- Enable nullable in `PackageSourceMapping` so that we can rely on the API providing empty collections instead of possibly null & empty collections.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
```diff
@@ -1015,26 +1012,35 @@ private PackageLicenseInfo GeneratePackageLicenseInfo(IPackageSearchMetadata met
...
+            results.AddRange(localMetadata);
 
-            using (var sourceCacheContext = new SourceCacheContext())
+            if (localMetadata.Length != allPackages.Length)
             {
-                IPackageSearchMetadata[] completed = (await TaskCombinators.ThrottledAsync(
-                    allPackages,
-                    (p, t) => GetPackageMetadataAsync(localSources, sourceCacheContext, p, t),
-                    token)).Where(metadata => metadata != null).Cast<IPackageSearchMetadata>().ToArray();
+                var packageSourceMapping = uiService.UIContext.PackageSourceMapping;
```
donnie-msft: TODO: move logic into `GetRemotePackageMetadataAsync`

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
```diff
@@ -1043,6 +1049,97 @@ private PackageLicenseInfo GeneratePackageLicenseInfo(IPackageSearchMetadata met
...
+                results.AddRange(remoteResults);
+            }
+
+            return results;
+        }
+
+        /// <summary>
+        /// Calculates a dictionary of package ID keys and source name values. If any package ID doesn't have a source mapping, returns a <c>null</c> dictionary.
+        /// </summary>
+        private Dictionary<string, List<string>>? GetSourceMappingsForPackages(IEnumerable<PackageIdentity> packages, PackageSourceMapping packageSourceMapping)
```
donnie-msft: Todo: no longer needed

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
```diff
@@ -1043,6 +1049,97 @@ private PackageLicenseInfo GeneratePackageLicenseInfo(IPackageSearchMetadata met
...
+            return completed;
+        }
+
+        private static async Task<IPackageSearchMetadata[]> GetPackageMetadataThrottledAsync(PackageSourceMapping packageSourceMapping, IEnumerable<SourceRepository> enabledSources, SourceCacheContext sourceCacheContext, IEnumerable<PackageIdentity> packages, CancellationToken token)
+        {
+            IPackageSearchMetadata[] completed = (await TaskCombinators.ThrottledAsync(
+                packages,
+                (p, t) =>
+                {
+                    var mappedSources = packageSourceMapping.GetConfiguredPackageSources(p.Id);
```
donnie-msft: Having the throttled delegate do the mapping lookup seemed like a better idea than precalculating the sources. Need to verify this works well.

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -32,8 +32,8 @@ internal static class PackageSourceMappingUtility
             List<string> addedPackageIdsWithoutExistingMappings = new();
             foreach (string addedPackageId in addedPackageIds)
             {
-                IReadOnlyList<string>? configuredSource = packageSourceMapping.GetConfiguredPackageSources(addedPackageId);
-                if (configuredSource is null || configuredSource.Count == 0)
+                IReadOnlyList<string> configuredSource = packageSourceMapping.GetConfiguredPackageSources(addedPackageId);
```
nkolev92, line 35: I think this is incorrect. Other places calling this method check for null as well. Looking at the implementation, it seems like nulls are possible.

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMapping.cs
```diff
@@ -20,25 +22,30 @@ public class PackageSourceMapping
...
         /// <exception cref="ArgumentException"> if <paramref name="packageId"/> is null, empty, or whitespace only.</exception>
         public IReadOnlyList<string> GetConfiguredPackageSources(string packageId)
         {
-            return SearchTree.Value?.GetConfiguredPackageSources(packageId);
+            if (SearchTree.Value is null)
+            {
+                return Array.Empty<string>();
+            }
+
+            return SearchTree.Value.GetConfiguredPackageSources(packageId);
```
nkolev92: The SearchTree method can return null as well.

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMapping.cs
```diff
@@ -20,25 +22,30 @@ public class PackageSourceMapping
         /// </summary>
         internal IReadOnlyDictionary<string, IReadOnlyList<string>> Patterns { get; }
 
-        private Lazy<SearchTree> SearchTree { get; }
+        private Lazy<SearchTree?> SearchTree { get; }
```
nkolev92: Would it make more sense to just have the full lazy be null. I think we can do that safely with the is enabled check.

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/SearchTree.cs
```diff
@@ -94,10 +96,16 @@ public string SearchForPattern(string term)
         /// <exception cref="ArgumentException"> if <paramref name="term"/> is null, empty, or whitespace only.</exception>
         public IReadOnlyList<string> GetConfiguredPackageSources(string term)
         {
-            return SearchNodeByTerm(term)?.PackageSources;
+            SearchNode? searchNodeResult = SearchNodeByTerm(term);
```
nkolev92: I see this makes my above comment invalid. 

I think if we're going to have a refactoring like this, it makes to do it in a separate PR. 

Furthermore, this *is* a binary breaking change. The behavior of the method is changing, even if undeclared. 

Is there a place where we differentiate between empty and null? Is that ever relevant or do they largely mean the same thing?

src/NuGet.Core/NuGet.Configuration/PublicAPI.Unshipped.txt
```diff
@@ -1,4 +1,4 @@
 #nullable enable
 NuGet.Configuration.PackageSourceMappingProvider.ShouldSkipSave.get -> bool
-~NuGet.Configuration.PackageSourceMapping.SearchForPattern(string packageId) -> string
+NuGet.Configuration.PackageSourceMapping.SearchForPattern(string! packageId) -> string?
```
nkolev92: Is `!` valid in the search for pattern? As a public API, I think it makes more sense for it to be null safe instead.

nkolev92 (commented):
The changeset is pretty large. I think there's some changes that are probably best done in a separate PR to make this one easier to review. 

I think the changes in NuGet.PackageManagementUI should be separate from the NuGet.Core changes.

ditto for the PackageDownloader tests.

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMapping.cs
donnie-msft: The `SearchTree` returns an empty list
https://github.com/NuGet/NuGet.Client/blob/1fbe83fe674911a7f1289dad525be6ff73b7d2e5/src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/SearchTree.cs#L100

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMapping.cs
donnie-msft: Then we have to null check the property itself. Not sure if I see the benefits clearly, yet...

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
donnie-msft: The `SearchTree` returns an empty list
https://github.com/NuGet/NuGet.Client/blob/1fbe83fe674911a7f1289dad525be6ff73b7d2e5/src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/SearchTree.cs#L100

src/NuGet.Core/NuGet.Configuration/PublicAPI.Unshipped.txt
donnie-msft: I'm not sure how it wouldn't be valid? I believe the fixit analyzer put `!` there.

Existing behavior throws an intentional error message:
https://github.com/NuGet/NuGet.Client/blob/1fbe83fe674911a7f1289dad525be6ff73b7d2e5/src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/SearchTree.cs#L110

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/SearchTree.cs
donnie-msft: > I think if we're going to have a refactoring like this, it makes to do it in a separate PR.

Can do: https://github.com/NuGet/NuGet.Client/pull/5339

> Is there a place where we differentiate between empty and null? Is that ever relevant or do they largely mean the same thing?

"Same thing", but I'm trying to consolidate these APIs to return empty to avoid unnecessary null checking plus empty checking by callers. 
I should be using Empty checks everywhere.

src/NuGet.Core/NuGet.Configuration/PublicAPI.Unshipped.txt
nkolev92: I missed the fact that this is an intentional behavior.

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
nkolev92: After I read the code, I realized that you changed that as well. Resolving this thread.

donnie-msft:
> The changeset is pretty large. I think there's some changes that are probably best done in a separate PR to make this one easier to review.
> 
> I think the changes in NuGet.PackageManagementUI should be separate from the NuGet.Core changes.
> 
> ditto for the PackageDownloader tests.

Created https://github.com/NuGet/NuGet.Client/pull/5339

aortiz-msft (dismissed):
:shipit:

nkolev92 (dismissed):

nkolev92 (approved):

PR #5296, opened 2023-07-05
Fix license display when it comes from a File

### LINKED ISSUE #10670: [Bug Bash] Only the embedded license content of the latest version can be loaded correctly in PM UI when there are multiple versions in the same package from local feeds
## Details about Problem
NuGet version: NuGet Client Dev\5.10.0.7170
VS Version: Main\31118.282
OS: Windows-10-Enterprise-20h1
## Repro steps and/or sample project
1.	Create a Class Library (.NET Framework 4.8) in VS and build the solution. 
2.	In the Developer Command Prompt for VS, cd to the project path and create a nuspec file by executing command “nuget spec”. 
3.	Make an image file (size limited to 1 MB with PNG format and 128x128 resolution) and locate in the package. 
4.	Make a License.txt and locate in the package. 
5.	Add following to nuspec file 
```
<package> 
  <metadata> 
    ... 
    <icon>images\icon.png</icon> 
    <license type="file">License.txt</license> 
  </metadata> 
  <files> 
    ... 
    <file src="[ProjectPath]\icon.png" target="images\" /> 
    <file src="[ProjectPath]\License.txt" target="" /> 
  </files> 
</package> 
```
6.	Pack the package to local path by executing command “nuget pack”. 
7.	Pack the same package with another version following step3-step6.
8.	Create a C# Console App (.NET Framework 4.8) in VS and open PM UI of the project/solution. 
9.	Add the package source as a local feed in Settings->NuGet Package Manager->Package Sources. 
10.	Select the new added local feed in the Package source dropdown box in the PM UI and click the package on the left panel.
11.	Choose a package version in the “Version” drop-down box on the right panel and click the “View License” link of every version. 
## Expected 
The embedded license content of every version can be loaded correctly in the PM UI.
## Actual
Only the embedded license content of the latest version can be loaded correctly as below.
image
## Notes
1.	Repro on both project-level PM UI and solution-level PM UI.
2.	Not a regression since it start to repro from the beginning of the embedded license feature in dev-rrelyea-remotefileservice\5.9.0.14330.

### LINKED ISSUE #12060: [Bug]: View License dialog does not display license content
### NuGet Product Used

### Product Version

NuGet Package Manager 6.3.0
Microsoft Visual Studio Enterprise 2022
Version 17.3.2
VisualStudio.17.Release/17.3.2+32819.101

### Worked before?

Unknown

### Impact

### Repro Steps & Context

Repro

- Open VS2022 
- Choose New Project
- Create new WPF Application from project template
- Choose .NET 6 (Long-term support)
- Right click project > Manage NuGet packages for project
- Search for `Microsoft.TestPlatform`
- Select `Microsoft.TestPlatform`
- Check project
- Click Install
- Click OK
- On License Acceptance dialog click View License link

Outcome

image

Expectations

License text is visible in dialog (previous behavior)

Context

Use of our NuGet package (https://www.nuget.org/packages/Esri.ArcGISRuntime.WPF) requires license acceptance. 

### Verbose Logs

_No response_

martinrrm:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12060
Fixes: https://github.com/NuGet/Home/issues/10670

## Description

[NuGetPackageFileService.GetEmbeddedLicenseAsync tries to get a URI from an in-memory cache](https://github.com/NuGet/NuGet.Client/blob/629ec930a61f1a52dfdd3c6713a4f6abd48a02df/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetPackageFileService.cs#L123), but the cache is never updated. With this change, I'm adding the correct information to the cache when the link is being created. This issue only occurs when the License comes from a file.

Fix for https://github.com/NuGet/Home/issues/10670:
display-license-local

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/PackageLicenseUtilities.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/PackageLicenseUtilities.cs
@@ -125,7 +125,7 @@ namespace NuGet.PackageManagement.UI
 
         private static Uri CreateEmbeddedLicenseUri(string packagePath, LicenseMetadata licenseMetadata)
         {
-            Uri baseUri = Convert(packagePath);
+            Uri baseUri = new Uri(packagePath, UriKind.Absolute);
 
             var builder = new UriBuilder(baseUri)
             {
@@ -133,21 +133,6 @@ namespace NuGet.PackageManagement.UI
             };
 
             return builder.Uri;
-        }
-
-        /// <summary>
-        /// Convert a string to a URI safely. This will return null if there are errors.
-        /// </summary>
-        private static Uri Convert(string uri)
-        {
-            Uri fullUri = null;
-
-            if (!string.IsNullOrEmpty(uri))
-            {
-                Uri.TryCreate(uri, UriKind.Absolute, out fullUri);
-            }
-
-            return fullUri;
         }
 
         private static void PopulateLicenseIdentifiers(NuGetLicenseExpression expression, IList<string> identifiers)
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/PackageLicenseUtilities.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/PackageLicenseUtilities.cs
@@ -9,6 +9,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using System.Windows.Documents;
 using Microsoft.ServiceHub.Framework;
+using NuGet.PackageManagement.VisualStudio;
 using NuGet.Packaging;
 using NuGet.Packaging.Core;
 using NuGet.Packaging.Licenses;
@@ -109,6 +110,9 @@ namespace NuGet.PackageManagement.UI
                     break;
 
                 case LicenseType.File:
+                    NuGetPackageFileService.AddLicenseToCache(
+                        packageIdentity,
+                        CreateEmbeddedLicenseUri(packagePath, metadata));
                     list.Add(new LicenseFileText(Resources.Text_ViewLicense, licenseFileHeader, packagePath, metadata.License, packageIdentity));
                     break;
 
@@ -117,6 +121,33 @@ namespace NuGet.PackageManagement.UI
             }
 
             return list;
+        }
+
+        private static Uri CreateEmbeddedLicenseUri(string packagePath, LicenseMetadata licenseMetadata)
+        {
+            Uri baseUri = Convert(packagePath);
+
+            var builder = new UriBuilder(baseUri)
+            {
+                Fragment = licenseMetadata.License
+            };
+
+            return builder.Uri;
+        }
+
+        /// <summary>
+        /// Convert a string to a URI safely. This will return null if there are errors.
+        /// </summary>
+        private static Uri Convert(string uri)
+        {
+            Uri fullUri = null;
+
+            if (!string.IsNullOrEmpty(uri))
+            {
+                Uri.TryCreate(uri, UriKind.Absolute, out fullUri);
+            }
+
+            return fullUri;
         }
 
         private static void PopulateLicenseIdentifiers(NuGetLicenseExpression expression, IList<string> identifiers)
--- a/test/NuGet.Clients.Tests/NuGet.PackageManagement.UI.Test/PackageLicenseUtilitiesTests.cs
+++ b/test/NuGet.Clients.Tests/NuGet.PackageManagement.UI.Test/PackageLicenseUtilitiesTests.cs
@@ -156,29 +156,27 @@ namespace NuGet.PackageManagement.UI.Test
         [Fact]
         public void PackageLicenseUtility_GeneratesLinkForFiles()
         {
-            using (TestDirectory testDir = TestDirectory.Create())
+            using (var pathContext = new SimpleTestPathContext())
             {
                 // Setup
-                // Create decoy nuget package
-                var zipPath = Path.Combine(testDir.Path, "file.nupkg");
-                CreateDummyPackage(zipPath: zipPath, licenseFile: "License.txt", licenseFileTargetElement: "");
+                var packageA1 = new SimpleTestPackageContext("AddLicenseToCache", "1.0.0");
+                string licenseFileLocation = "License.txt";
+                string licenseFileHeader = "header";
+                LicenseMetadata licenseData = new LicenseMetadata(LicenseType.File, licenseFileLocation, null, null, LicenseMetadata.CurrentVersion);
+                packageA1.AddFile(licenseFileLocation, StreamLicenseContents);
 
-                var licenseFileLocation = "License.txt";
-                var licenseFileHeader = "header";
-                var licenseData = new LicenseMetadata(LicenseType.File, licenseFileLocation, null, null, LicenseMetadata.CurrentVersion);
-
-                var packageIdentity = new PackageIdentity("AddLicenseToCache", NuGetVersion.Parse("1.0.0"));
+                PackageIdentity packageIdentity = new PackageIdentity("AddLicenseToCache", NuGetVersion.Parse("1.0.0"));
 
                 // Act
-                var links = PackageLicenseUtilities.GenerateLicenseLinks(
+                IReadOnlyList<IText> links = PackageLicenseUtilities.GenerateLicenseLinks(
                     licenseData,
                     licenseFileHeader,
-                    zipPath,
+                    pathContext.SolutionRoot,
                     packageIdentity: packageIdentity);
 
                 Assert.Equal(1, links.Count);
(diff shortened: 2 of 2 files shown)


jeffkl (dismissed):

zivkan (dismissed):

zivkan:
@martinrrm can you check if there is any telemetry emit before your change to let us know that getting an embedded licence has failed?  If not, please add some, because if the cache decides its too full and decides to evict the data in the short time between adding it to the cache and then reading it from the cache, we're still going to have problems.

If we're not going to fix the fundamental design flaw, we should at least have telemetry.

martinrrm:
@zivkan Sorry for the late response, we do emit telemetry for faults here. https://github.com/NuGet/NuGet.Client/blob/629ec930a61f1a52dfdd3c6713a4f6abd48a02df/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetPackageFileService.cs#L126-L128
image

donnie-msft:
Since this change is to the utilities, might the PR also fix https://github.com/NuGet/Home/issues/10670 ?
https://github.com/NuGet/NuGet.Client/blob/636dfa50504451bee7fb6cd70adca21db339bf99/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailedPackageMetadata.cs#L116

martinrrm:
@donnie-msft Confirmed that this PR also fixes that issue, added a gif to demonstrate.

jeffkl (dismissed):

test/NuGet.Clients.Tests/NuGet.PackageManagement.UI.Test/PackageLicenseUtilitiesTests.cs
```diff
@@ -181,5 +245,43 @@ public void PackageLicenseUtility_GenerateCorrectLink()
             Assert.Equal(license, licenseText.Text);
             Assert.Equal("https://licenses.nuget.org/MIT", licenseText.Link.AbsoluteUri);
         }
+
+        private static void CreateDummyPackage(
```
donnie-msft: This feels like something we already have, or if not, should make more globally available to tests. Perhaps a new test fixture.

test/NuGet.Clients.Tests/NuGet.PackageManagement.UI.Test/PackageLicenseUtilitiesTests.cs
```diff
@@ -181,5 +245,43 @@ public void PackageLicenseUtility_GenerateCorrectLink()
...
+            {
+                nuspec.Write(writer);
+                writer.Flush();
+                nuspecStream.Position = 0;
+                var pkgBuilder = new PackageBuilder(stream: nuspecStream, basePath: folderPath);
+                pkgBuilder.Save(nupkgStream);
+            }
+        }
+
+        private static string StreamLicenseContents = "I am  license";
```
donnie-msft: I'm not sure if this ought to be static. We have a lot of tests that share test class data, and singletons create potential for side effects from other tests.
Probably should also be `readonly`.
```suggestion
        private readonly string _streamLicenseContents = "I am license";
```

test/NuGet.Clients.Tests/NuGet.PackageManagement.UI.Test/PackageLicenseUtilitiesTests.cs
```diff
@@ -144,23 +156,75 @@ public void PackageLicenseUtility_BadUnlicensedGeneratesNoLinksAndAWarning()
...
 
-            // Act
-            var links = PackageLicenseUtilities.GenerateLicenseLinks(
-                licenseData,
-                licenseFileHeader,
-                licenseFileLocation,
-                packageIdentity: null);
+                var licenseFileLocation = "License.txt";
+                var licenseFileHeader = "header";
+                var licenseData = new LicenseMetadata(LicenseType.File, licenseFileLocation, null, null, LicenseMetadata.CurrentVersion);
```
donnie-msft: Please use named parameters

donnie-msft (dismissed):
🎉 Thanks for checking on the other bug. "2 bugs with 1 PR"!

I left some comments which I believe are techdebt. Ideally, fix them in this PR. Otherwise, please create a follow-up issue to address them.

test/NuGet.Clients.Tests/NuGet.PackageManagement.UI.Test/PackageLicenseUtilitiesTests.cs
zivkan: probably `const`, rather than `readonly`. If the value never changes, it doesn't need to be a per-instance field.

test/NuGet.Clients.Tests/NuGet.PackageManagement.UI.Test/PackageLicenseUtilitiesTests.cs
zivkan: `SimpleTestPathContext` creates a temp directory for tests to write to, but also a nuget.config file that adds a local folder as a package source.  If you look at references to the `PackageSource` property, you can probably find the helper methods that create test packages for Restore tests to use.

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageLicenseUtilities.cs
```diff
@@ -119,6 +123,33 @@ internal static IReadOnlyList<IText> GenerateLicenseLinks(LicenseMetadata metada
...
+                Fragment = licenseMetadata.License
+            };
+
+            return builder.Uri;
+        }
+
+        /// <summary>
+        /// Convert a string to a URI safely. This will return null if there are errors.
+        /// </summary>
+        private static Uri Convert(string uri)
```
aortiz-msft: Should this name be a bit more specific than just Convert? For example, ConvertPathToUri? While the types can be inferred from the arguments and return type, I find the name a bit generic. Alternatively, do we need this method at all since it's been used from a single location?

test/NuGet.Clients.Tests/NuGet.PackageManagement.UI.Test/PackageLicenseUtilitiesTests.cs
aortiz-msft: Const or private static readonly string would make sense for this use, I think.

aortiz-msft (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageLicenseUtilities.cs
aortiz-msft: @zivkan ^

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageLicenseUtilities.cs
```diff
@@ -119,6 +123,33 @@ internal static IReadOnlyList<IText> GenerateLicenseLinks(LicenseMetadata metada
             return list;
         }
 
+        private static Uri CreateEmbeddedLicenseUri(string packagePath, LicenseMetadata licenseMetadata)
+        {
+            Uri baseUri = Convert(packagePath);
+
+            var builder = new UriBuilder(baseUri)
```
zivkan: Your current implementation of `Convert` can return null, but [`System.UriBuilder`'s constructor that takes a `Uri` input throws ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.uribuilder.-ctor?view=net-7.0#system-uribuilder-ctor(system-uri)). So, this code isn't null safe.

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageLicenseUtilities.cs
zivkan: honestly, considering this is used in only one place, and both this method and the other one is short, I wouldn't bother having this function at all. I think it obscures understanding, not helping, as I commented where it's used, since it makes it harder to understand null safety.

If we keep this method, renaming it to something like `Uri ParseUriOrNull(string)` might be better, but then there will be questions about why not `bool TryParseUri(string)`, but then that's almost identical to the BCL's own `bool Uri.TryParse(string, UriKind)`.  But since nulls aren't being handled, the code might as well use the `Uri(string, UriKind)` constructor. So again I'm coming back to getting rid of this method as the better option 🤷

zivkan (approved):

donnie-msft (approved):

PR #5297, opened 2023-07-06
NuGetAuditMode all should only warn on resolved versions

### LINKED ISSUE #12730: NuGetAuditMode all warns about package versions that were upgraded (rejected)
### NuGet Product Used

dotnet.exe, MSBuild.exe

### Product Version

6.7

### Worked before?

no, new feature

### Impact

### Repro Steps & Context

Consider the following project:

```xml
  <PropertyGroup>
    <NuGetAudit>true</NuGetAudit>
    <NuGetAuditMode>all</NuGetAuditMode>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="NuGet.Packaging" Version="5.11.5" />
    <PackageReference Include="Microsoft.AspNetCore.JsonPatch" Version="5.0.1" />
  </ItemGroup>

</Project>
```

This results in two warnings about two different versions of Newtonsoft.Json:

> D:\src\test\vulnBug\vulnBug.csproj : warning NU1902: Package 'Newtonsoft.Json' 9.0.1 has a known moderate severity vulnerability, https://github.com/advisories/GHSA-5crp-9r3c-p9vr
> D:\src\test\vulnBug\vulnBug.csproj : warning NU1902: Package 'Newtonsoft.Json' 12.0.2 has a known moderate severity vulnerability, https://github.com/advisories/GHSA-5crp-9r3c-p9vr

The package `NuGet.Packaging` has a dependency on `Newtonsoft.Json` 9.0.1, whereas `Microsoft.AspNetCore.JsonPatch` has a dependency on version `12.0.2`. If you upgrade either package to a newer version, that has a dependency on a no longer vulnerable version of `Newtonsoft.Json`, NU1902 will still since there's a part of the restore graph that rejected the lower version of Newtonsoft.Json.

### Verbose Logs

_No response_

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12730

Regression? no, new feature to 6.7

## Description

See linked bug for example. While the data type `TargetRestoreGraph` has a property `ResolvedDependencies`, it's a flat list of all graph nodes, but without information about whether the version was accepted or rejected (upgraded by another node in the graph of the same package id with a higher version, or a direct package reference). Therefore, we need to instead walk `TargetRestoreGraph.Graphs`, so we can check if `node.Disposition == Disposition.Accepted`.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -307,6 +307,40 @@ private static (string severityLabel, NuGetLogCode code) GetSeverityLabelAndCode
                 }
             }
             return result;
+
+            IEnumerable<(LibraryIdentity package, bool isDirect)> WalkAcceptedPackages(RestoreTargetGraph graph)
```
nkolev92: We'll probably need a test for this method to make it easier to argue for a patch.

jeffkl (dismissed):

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -307,6 +307,40 @@ private static (string severityLabel, NuGetLogCode code) GetSeverityLabelAndCode
...
+            {
+                Stack<(GraphNode<RemoteResolveResult> node, bool isDirect)> stack = new();
+                foreach (var projectNode in graph.Graphs)
```
nkolev92: Why not use Flattened instead? 
It should already exclude the reject nodes + no computation.

I think you *might* need to differentiate between direct and transitive there somehow, but cheaper than an exhaustive graph walk.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -298,7 +299,8 @@ private static (string severityLabel, NuGetLogCode code) GetSeverityLabelAndCode
                                 affectedGraphs.Add(graph.TargetGraphName);
                             }
 
-                            if (!auditInfo.IsDirect && resolvedDependency.Parent == currentProject)
+                            if (!auditInfo.IsDirect &&
```
nkolev92, line 302: nit: Is it worth creating a hash set for this?

test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/CompatibilityIssueTests.cs
```diff
@@ -6,7 +6,7 @@
 using NuGet.Versioning;
 using Xunit;
 
-namespace NuGet.Commands.Test
+namespace NuGet.Commands.Test.RestoreCommandTests
```
nkolev92, line 9: Is this stylistic only?

nkolev92 (approved):
LGTM. 

I'm guessing your PR body needs an update given that now you have tests?

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
zivkan: I assume you mean on the following line, not the quoted line?

I don't know. Creating a hash set has a cost, and I expect most people will want to avoid using packages with vulnerabilities in their projects, so I don't expect this code path to be used very frequently. It would also be interesting to know what the 95th percentile is for the number of direct package references in a project is. If the number is small, then I suspect there will be little benefit to a hashset.

test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/CompatibilityIssueTests.cs
zivkan: Basically, yes.

Namespace hierarchy not matching the directory hierarchy really bugs me. My position is that if the files benefit from organization in subdirectories, so do namespaces. Plus VS creates files with namespaces matching the directory structure, so using anything different means one more (little) thing where we're fighting against the IDE.

Unfortunately, we have a class named `RestoreCommand`, in addition to a directory with the same name. So, using namespaces that match directory structure means that code gets confused when the namespace and class are "visible" at the same time, making it impossible to use the class, without also specifying the full namespace. In other words, `NuGet.Commands.RestoreCommand`, because otherwise using `RestoreCommand`, the compiler always thinks it's referring to `NuGet.Commands.Test.RestoreCommand` namespace, and even using `using RestoreCommand = NuGet.Commands.RestoreCommand` doesn't work.  Hence I renamed the directory to `RestoreCommandTests`, and updated all affected files to use the matching namespace.

nkolev92: Good to know. 
FWIW, for QB mode bugs, the file count + line count tends to be noticed, so for future reference PRs intended to be patch, should be as minimal as can be :)

PR #5300, opened 2023-07-06
Move Version property to config.props & output to buildinfo.json

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2167

## Description

After moving this, my understanding is that `$(Version)` will contain the full package # that we insert into VS, so that it can be used in our insertion PRs.
I have already modified the NuGet Client Release to modify the following files, but I need the full version# of VS to put here:
- `nuget.packageconfig`
- `NuGetVisualStudio.props`

I don't know how to test this until I have an Official build to try the release on.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/build/config.props
+++ b/build/config.props
@@ -100,6 +100,9 @@
   <Target Name="GetVsTargetMajorVersion">
     <Message Text="$(VsTargetMajorVersion)" Importance="High"/>
   </Target>
+  <Target Name="GetVersion">
+    <Message Text="$(Version)" Importance="High"/>
+  </Target>
   <Target Name="GetVsTargetBranch">
     <Message Text="$(VsTargetBranch)" Importance="High"/>
   </Target>
--- a/build/config.props
+++ b/build/config.props
@@ -37,6 +37,35 @@
     <VsTargetChannel Condition="'$(IsEscrowMode)' == 'true'">int.d$(VsTargetMajorVersion).$(MinorNuGetVersion)</VsTargetChannel>
   </PropertyGroup>
 
+ <PropertyGroup Condition=" '$(PreReleaseVersion)' == '' ">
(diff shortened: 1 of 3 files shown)


build/config.props
```diff
@@ -37,6 +37,35 @@
     <VsTargetChannel Condition="'$(IsEscrowMode)' == 'true'">int.d$(VsTargetMajorVersion).$(MinorNuGetVersion)</VsTargetChannel>
   </PropertyGroup>
 
+ <PropertyGroup Condition=" '$(PreReleaseVersion)' == '' ">
```
nkolev92, line 40: This is just step 1. 

You basically need to get the exact version to buildinfo.json, which is generated here: https://github.com/NuGet/NuGet.Client/blob/0ebb20188285663d7c8349df9ddfd71020c231c4/scripts/cibuild/ConfigureVstsBuild.ps1#L170-L189.

nkolev92: You need an equivalent target as the GetNuGetSdkVsSemanticVersion and the likes

donnie-msft: If I can get the version there, why do I need this change?

nkolev92: The targets and props are defined in config.props for all other properties. common.project.props has other imports and properties set that you don't want or need.

donnie-msft: How's this?
image

nkolev92: I think that's what you want.

nkolev92 (approved):

PR #5303, opened 2023-07-09
Reduce size of LibraryDependency class

drewnoakes:
## Bug

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1826369
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1835763

## Description

_This PR is @melytc's work as part of the perfathon. I put together the PR as she's OOF the coming week._

The `LibraryDependency` class has been identified as contributing to high GC on customer machines, allocating up to 37MB/s. This change reduces the memory footprint of that type in two ways:

1. Six integral fields are packed into a single 32-bit integer, reducing 12 bytes down to 4 bytes.
2. The backing collection for the `NoWarn` property is now lazily allocated, as it will usually be empty.

For the packing of fields, there are three Boolean and three enum-based fields. With some masking and shifting, these can all be efficiently stored within a single 32-bit integer space. Unit tests have been added to ensure the correctness of those fields now and in future. Comments have been added to the enums being packed to ensure any future modifications to those enums take this packing strategy into account. The packing of enums does leave some headroom, so a few more members could be added to each without requiring any code changes in `LibraryDependency`.

For the `NoWarn` collection, an additional `NoWarnCount` property was added that allows callers to test whether the collection is empty before calling the `NoWarn` getter (which would lazily allocate a backing collection). This seems to be a pragmatic approach to this issue given the backwards compatibility requirements. Several callers have been updated to check the count first.

This PR also null annotates `LibraryDependency` and some other types that needed annotations in the process of doing so.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
+++ b/src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
@@ -261,7 +261,7 @@ namespace NuGet.LibraryModel
                         continue;
                     }
 
-                    if (centralPackageVersions.TryGetValue(d.Name, out CentralPackageVersion centralPackageVersion))
+                    if (centralPackageVersions.TryGetValue(d.Name, out CentralPackageVersion? centralPackageVersion))
                     {
                         d.LibraryRange!.VersionRange = centralPackageVersion.VersionRange;
                     }
--- a/src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
+++ b/src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
@@ -3,10 +3,13 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using NuGet.Common;
 using NuGet.Shared;
 using NuGet.Versioning;
+
+#nullable enable
 
 namespace NuGet.LibraryModel
 {
@@ -58,7 +61,7 @@ namespace NuGet.LibraryModel
         /// <summary>
         /// Lazily allocated backing collection for the <see cref="NoWarn"/> property.
         /// </summary>
-        private IList<NuGetLogCode> _noWarn;
+        private IList<NuGetLogCode>? _noWarn;
 
         public bool GeneratePathProperty
         {
@@ -118,6 +121,7 @@ namespace NuGet.LibraryModel
         /// This property lazily allocates its backing collection. Callers should check <see cref="NoWarnCount"/>
         /// for a non-zero value before reading this property to avoid redundant allocations.
         /// </remarks>
+        [AllowNull]
         public IList<NuGetLogCode> NoWarn
         {
             // Lazily allocate the list if needed.
@@ -130,16 +134,16 @@ namespace NuGet.LibraryModel
         /// </summary>
         public int NoWarnCount => _noWarn?.Count ?? 0;
 
-        public LibraryRange LibraryRange { get; set; }
+        public LibraryRange? LibraryRange { get; set; }
 
-        public string Name => LibraryRange.Name;
+        public string Name => LibraryRange!.Name;
 
-        public string Aliases { get; set; }
+        public string? Aliases { get; set; }
 
         /// <summary>
         /// Gets or sets a value indicating a version override for any centrally defined version.
         /// </summary>
-        public VersionRange VersionOverride { get; set; }
+        public VersionRange? VersionOverride { get; set; }
 
         public LibraryDependency()
         {
@@ -149,13 +153,13 @@ namespace NuGet.LibraryModel
             LibraryRange libraryRange,
             LibraryIncludeFlags includeType,
             LibraryIncludeFlags suppressParent,
-            IList<NuGetLogCode> noWarn,
+            IList<NuGetLogCode>? noWarn,
             bool autoReferenced,
             bool generatePathProperty,
             bool versionCentrallyManaged,
             LibraryDependencyReferenceType libraryDependencyReferenceType,
-            string aliases,
-            VersionRange versionOverride)
+            string? aliases,
+            VersionRange? versionOverride)
         {
             LibraryRange = libraryRange;
             IncludeType = includeType;
@@ -171,10 +175,10 @@ namespace NuGet.LibraryModel
 
         private LibraryDependency(
             int flags,
-            LibraryRange libraryRange,
-            IList<NuGetLogCode> noWarn,
-            string aliases,
-            VersionRange versionOverride)
+            LibraryRange? libraryRange,
+            IList<NuGetLogCode>? noWarn,
+            string? aliases,
+            VersionRange? versionOverride)
         {
             _flags = flags;
             LibraryRange = libraryRange;
@@ -186,7 +190,7 @@ namespace NuGet.LibraryModel
         public override string ToString()
         {
             // Explicitly call .ToString() to ensure string.Concat(string, string, string) overload is called.
-            return LibraryRange.ToString() + " " + LibraryIncludeFlagUtils.GetFlagString(IncludeType);
+            return LibraryRange?.ToString() + " " + LibraryIncludeFlagUtils.GetFlagString(IncludeType);
         }
 
         public override int GetHashCode()
@@ -201,12 +205,12 @@ namespace NuGet.LibraryModel
             return hashCode.CombinedHash;
         }
 
-        public override bool Equals(object obj)
+        public override bool Equals(object? obj)
         {
             return Equals(obj as LibraryDependency);
         }
 
-        public bool Equals(LibraryDependency other)
+        public bool Equals(LibraryDependency? other)
         {
             if (other == null)
             {
@@ -228,7 +232,7 @@ namespace NuGet.LibraryModel
 
         public LibraryDependency Clone()
         {
-            var clonedLibraryRange = new LibraryRange(LibraryRange.Name, LibraryRange.VersionRange, LibraryRange.TypeConstraint);
+            var clonedLibraryRange = new LibraryRange(LibraryRange!.Name, LibraryRange.VersionRange, LibraryRange.TypeConstraint);
             var clonedNoWarn = _noWarn is null ? null : new List<NuGetLogCode>(_noWarn);
 
             return new LibraryDependency(_flags, clonedLibraryRange, clonedNoWarn, Aliases, VersionOverride);
@@ -249,18 +253,18 @@ namespace NuGet.LibraryModel
             }
             if (centralPackageVersions.Count > 0)
             {
-                foreach (LibraryDependency d in packageReferences.Where(d => !d.AutoReferenced && d.LibraryRange.VersionRange == null))
+                foreach (LibraryDependency d in packageReferences.Where(d => !d.AutoReferenced && d.LibraryRange!.VersionRange == null))
                 {
                     if (d.VersionOverride != null)
                     {
-                        d.LibraryRange.VersionRange = d.VersionOverride;
+                        d.LibraryRange!.VersionRange = d.VersionOverride;
 
                         continue;
                     }
 
                     if (centralPackageVersions.TryGetValue(d.Name, out CentralPackageVersion centralPackageVersion))
                     {
-                        d.LibraryRange.VersionRange = centralPackageVersion.VersionRange;
+                        d.LibraryRange!.VersionRange = centralPackageVersion.VersionRange;
                     }
 
                     d.VersionCentrallyManaged = true;
--- a/src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
+++ b/src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
@@ -223,10 +223,9 @@ namespace NuGet.LibraryModel
             }
 
             return _flags == other._flags &&
-                   EqualityUtility.EqualsWithNullCheck(LibraryRange, other.LibraryRange) &&
-                   _noWarn is null == other._noWarn is null &&
-                   (_noWarn is null || _noWarn.SequenceEqualWithNullCheck(other._noWarn)) &&
                    Aliases == other.Aliases &&
+                   EqualityUtility.EqualsWithNullCheck(LibraryRange, other.LibraryRange) &&
+                   (_noWarn ?? Enumerable.Empty<NuGetLogCode>()).SequenceEqualWithNullCheck(other._noWarn ?? Enumerable.Empty<NuGetLogCode>()) &&
                    EqualityUtility.EqualsWithNullCheck(VersionOverride, other.VersionOverride);
         }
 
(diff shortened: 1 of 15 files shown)


nkolev92:
@drewnoakes There are bunch of test failures that look like true failures. Can you please take a look?

drewnoakes:
I can't see any failures that seem related to this PR. Can you point to an example of what you consider a true failure?

**EDIT** I assume you're talking about the functional tests.

zivkan:
Every test I look at in Azure DevOps, and go to the history tab for the failed test, shows it only fails in this PR's branch, not in any other branch. So, while NuGet does have flakey tests, all the tests I looked at for this PR's build appear not to be flakey, it appears related to the change in this PR. Given the same test fails on Mac, Linux, Windows .NET Framework and Windows .NET (Core) also suggests it's not flakey, and related to this PR.

drewnoakes:
Thanks. I saw an earlier build number as the first instance of the failure, but that seems to have been an earlier build of this PR.

That failure is very high level. It's not quickly obvious where the failure could be l, or how to narrow it down from such a high level test.

I won't have time to look at this for a few days.

drewnoakes:
@zivkan @nkolev92 I found and fixed an issue. The tests are running now.

To be clear this bug was in my code, not @melytc's! I made a change to avoid materializing the `NoWarn` collection in `Equals` and messed it up. The last commit fixes it, and adds a test for that case.

zivkan:
source build retargets everything to build for net8.0, and this has caused a compile error: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=8059454&view=logs&j=cb98efd6-7062-56ef-f642-0176540bbe0a&t=d42c45d2-f6c8-5ae2-d854-abf6f8d78853&l=88

> /home/vsts/work/1/s/src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs(264,72): error CS8600: Converting null literal or possible null value to non-nullable type. [/home/vsts/work/1/s/src/NuGet.Core/NuGet.LibraryModel/NuGet.LibraryModel.csproj::TargetFramework=net8.0]

src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
```diff
@@ -3,70 +3,168 @@
...
+        /// <summary>
+        /// Lazily allocated backing collection for the <see cref="NoWarn"/> property.
+        /// </summary>
+        private IList<NuGetLogCode>? _noWarn;
 
-        public string Name => LibraryRange.Name;
+        public bool GeneratePathProperty
+        {
+            // This property is stored in the lowest bit (0b1) in position 00.
+            get => (_flags & 0b1) != 0;
```
nkolev92, line 69: I think readability should be a #1 priority for all code we write.
All the bitshifting makes this challenging.
Are the benefits really that huge that we need to resort to something as involved as this? 

We're talking about a few booleans here. The longer name for the dependency would have a bigger impact to the overall size of the type.

src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
```diff
@@ -3,70 +3,168 @@
...
+
+        /// <summary>
+        /// Gets the list of NoWarn codes for this library dependency.
+        /// </summary>
+        /// <remarks>
+        /// This property lazily allocates its backing collection. Callers should check <see cref="NoWarnCount"/>
+        /// for a non-zero value before reading this property to avoid redundant allocations.
+        /// </remarks>
+        [AllowNull]
+        public IList<NuGetLogCode> NoWarn
```
nkolev92, line 125: I like this change.

src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
```diff
@@ -3,70 +3,168 @@
...
+        {
+            // Lazily allocate the list if needed.
+            get => _noWarn ??= new List<NuGetLogCode>();
+            set => _noWarn = value;
+        }
+
+        /// <summary>
+        /// This internal field will help us avoid allocating a list when calling the count on a null.
+        /// </summary>
+        public int NoWarnCount => _noWarn?.Count ?? 0;
```
nkolev92, line 135: It calls it internal, but it's declared as public.

nkolev92 (commented):
I like the lazy initialization changes. 
The packing seems like a lot. 

I'd love to see how significant of a different this would make for some test repos like what we use in https://github.com/NuGet/NuGet.Client/tree/dev/scripts/perftests

drewnoakes:
> I like the lazy initialization changes.

It helps, but it makes all calling code more ugly and requires callers to know that they need to check this property before reading. Future code will easily break this invariant and undo any benefit from making it lazy. the `Freeze` approach in a recent PR could avoid that.

> The packing seems like a lot.

Prism shows a lot of memory devoted to instances of this type. Packing these fields cuts the size of the type significantly. By contrast to the lazy change however, it's completely enclosed within this type, doesn't require any change to the users of these types, and cannot be broken by external changes.

> We're talking about a few booleans here. The longer name for the dependency would have a bigger impact to the overall size of the type.

It's three booleans and three enums. Altogether that saves ten bytes per instance. For this to appear in Prism, there must be a lot of instances of this type in the real world.

src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
zivkan: I just did some testing locally.

## Data

First, I added a `static int counter` to the class, and in the constructor output to console the value of `Interlocked.Increment(counter)`. Restoring OrchardCore, the counter reached 24,862

Running the same restore in PerfView, here's the GC stats:

<TABLE border="1"><TBODY><TR><TH>Gen</TH><TH>Count</TH><TH>Max<BR>Pause</TH><TH>Max<BR>Peak MB</TH><TH>Max Alloc<BR>MB/sec</TH><TH>Total<BR>Pause</TH><TH>Total<BR>Alloc MB</TH><TH>Alloc MB/<BR>MSec GC</TH><TH>Survived MB/<BR>MSec GC</TH><TH>Mean<BR>Pause</TH><TH>Induced</TH></TR><TR><TD align="Center">ALL</TD><TD align="Center">258</TD><TD align="Center">53.5</TD><TD align="Center">921.7</TD><TD align="Center">5,220.699</TD><TD align="Center">4,550.4</TD><TD align="Center">2,596.0</TD><TD align="Center">0.6</TD><TD align="Center">0.501</TD><TD align="Center">17.6</TD><TD align="Center">0</TD></TR><TR bgcolor="#bdedff"><TD align="Center">0</TD><TD align="Center">155</TD><TD align="Center">42.1</TD><TD align="Center">605.3</TD><TD align="Center">5,220.699</TD><TD align="Center">1,926.2</TD><TD align="Center">1,577.7</TD><TD align="Center">0.7</TD><TD align="Center">0.236</TD><TD align="Center">12.4</TD><TD align="Center">0</TD></TR><TR bgcolor="#82caff"><TD align="Center">1</TD><TD align="Center">96</TD><TD align="Center">41.8</TD><TD align="Center">604.0</TD><TD align="Center">1,233.922</TD><TD align="Center">2,432.6</TD><TD align="Center">970.7</TD><TD align="Center">0.9</TD><TD align="Center">0.453</TD><TD align="Center">25.3</TD><TD align="Center">0</TD></TR><TR bgcolor="#56a5ec"><TD align="Center">2</TD><TD align="Center">7</TD><TD align="Center">53.5</TD><TD align="Center">921.7</TD><TD align="Center">173.133</TD><TD align="Center">191.7</TD><TD align="Center">47.6</TD><TD align="Center">0.1</TD><TD align="Center">15.743</TD><TD align="Center">27.4</TD><TD align="Center">0</TD></TR></TBODY></TABLE>

By comparison, here's the GC stats for the same restore from the baseline (commit tagged 6.8.0.13, which was the first tagged commit when I did `git log` on this branch):

<TABLE border="1"><TBODY><TR><TH>Gen</TH><TH>Count</TH><TH>Max<BR>Pause</TH><TH>Max<BR>Peak MB</TH><TH>Max Alloc<BR>MB/sec</TH><TH>Total<BR>Pause</TH><TH>Total<BR>Alloc MB</TH><TH>Alloc MB/<BR>MSec GC</TH><TH>Survived MB/<BR>MSec GC</TH><TH>Mean<BR>Pause</TH><TH>Induced</TH></TR><TR><TD align="Center">ALL</TD><TD align="Center">259</TD><TD align="Center">79.1</TD><TD align="Center">575.3</TD><TD align="Center">10,176.142</TD><TD align="Center">3,789.5</TD><TD align="Center">2,603.4</TD><TD align="Center">0.7</TD><TD align="Center">0.727</TD><TD align="Center">14.6</TD><TD align="Center">0</TD></TR><TR bgcolor="#bdedff"><TD align="Center">0</TD><TD align="Center">154</TD><TD align="Center">18.7</TD><TD align="Center">573.8</TD><TD align="Center">10,176.142</TD><TD align="Center">1,553.9</TD><TD align="Center">1,579.7</TD><TD align="Center">0.6</TD><TD align="Center">0.291</TD><TD align="Center">10.1</TD><TD align="Center">0</TD></TR><TR bgcolor="#82caff"><TD align="Center">1</TD><TD align="Center">97</TD><TD align="Center">34.3</TD><TD align="Center">575.3</TD><TD align="Center">1,628.654</TD><TD align="Center">1,985.4</TD><TD align="Center">968.7</TD><TD align="Center">0.8</TD><TD align="Center">0.516</TD><TD align="Center">20.5</TD><TD align="Center">0</TD></TR><TR bgcolor="#56a5ec"><TD align="Center">2</TD><TD align="Center">8</TD><TD align="Center">79.1</TD><TD align="Center">574.0</TD><TD align="Center">123.293</TD><TD align="Center">250.2</TD><TD align="Center">55.1</TD><TD align="Center">0.1</TD><TD align="Center">14.118</TD><TD align="Center">31.3</TD><TD align="Center">0</TD></TR></TBODY></TABLE>

Finally, I ran `dotnet restore --force -nr:false -p:NuGetRestoreTargets="D:\src\NuGet.Client\artifacts\NuGet.Build.Tasks.Console\bin\Release\netcoreapp5.0\NuGet.targets"` 10 times each, and used PowerShell's `Measure-Command` to get the `TotalSeconds` duration:

|Run|Baseline|PR|
|--|--|--|
|1|13.5290796|14.6853872|
|2|14.0730449|13.7910359|
|3|13.1487936|13.9927866|
|4|13.2918021|14.2871618|
|5|13.9346949|12.7434424|
|6|14.3410409|14.0126256|
|7|14.0464915|13.7106395|
|8|13.9592139|14.2359064|
|9|13.340791|13.9556673|
|10|12.5458226|13.7064334|
|mean|13.6210775|13.91210861|
|stddev|0.51874489|0.482214791|

## My opinion

I believe that GC stats are non-deterministic, so a sample size of 1 is completely insufficient to measure any small difference. What I think it does show is that there isn't a significant difference to overall memory allocations (baseline has 0.3% fewer total MB allocated). Collecting this is too slow for me to do manually to try to get more compelling data for or against.

The PR adds a comment saying that the bit packing saves 8 bytes per instance. Given the approx 25k instances measured,  that's theoretically 200k saved per restore. I didn't try to isolate the bit packing changes separately from the NoWarn & LibraryRange changes. A quick search on Stack Overflow found me an (unverified) claim that objects in .NET have an overhead of 8 bytes, plus the size of the types that the class contains, plus possibly padding. So, avoiding the NoWarn lists should be a significantly higher impact than the bit packing.

Finally, it's odd that the total restore duration was slower with these changes, but maybe we just got unlucky. A sample size of 10 isn't very large. I don't remember first year university statistics, so I can't remember how significant the fact that both the baseline and PR mean durations are within a single standard deviation of each other.

> Prism shows a lot of memory devoted to instances of this type.

I don't speak for my whole team, but I'm certainly willing to take simple, easy to understand changes based only on "prism shows a lot of memory [on] this type". So the lazy initialization of the NoWarn list is fine to me without further evidence.

However, reducing readability to something that is difficult to maintain I think should require a higher bar. Unless I made a mistake in my measurements, I don't think the data supports the changes as being meaningful.

> I think readability should be a # 1 priority for all code we write

I don't agree with this, as written. If a particular bit of code is a performance bottleneck AND in a user scenario where performance is very important, then we should consider more difficult to understand code as an acceptable tradeoff. Looking at ASP.NET Core's TechEmpower benchmark improvements over the years shows that making unreadable code changes can have enormous performance impacts, and save customers a lot of money in compute reductions. In NuGet, that would be equivalent to reduced CI agent minutes/hours. Or theoretically fewer GC pauses in VS, increasing it's "responsiveness".

However, as previously mentioned, I don't see evidence in my measurements that this PR does reduce restore duration. The GC stats are probably meaningless because of the single sample for the baseline and PR, but if somehow they are representative, while the single sample shows a reduction in gen2 GC duration, it had an increase in gen 0 and gen 1 GC duration, which happens much more frequently, while the number of GCs was effectively the same. Therefore from a VS UI latency/responsiveness point of view, the changes appear counter-productive. But I expect this just to be a sample size problem. Reducing total allocation by several megabytes, it doesn't make sense for GC to be slower while also the same frequency. Also, trying to infer VS GC stats from a CLI restore is probably foolish. 

In any case, there just isn't compelling data that the reduced readability/maintainability has actually improved performance in any metric.

ghost:
This PR has been automatically marked as stale because it has no activity for **7 days**. It will be closed if no further activity occurs **within another 90 days** of this comment. If it is closed, you may reopen it anytime when you're ready again, as long as you don't delete the branch.

davkean:
@nkolev92 @drewnoakes @zivkan 

Please reconsider this PR; in dumps taken from a large partner's internal restore, I can see that there are almost half a million of these objects in memory. 

image

This is just a snapshot at a point in time so doesn't include all of its allocations over the lifetime of the restore. The memory at its peak, quadrupled than what the above dump showed, so there's likely more than 2 million of these objects in memory at any point throughout its restore.

When you have that many instances of these objects in memory, every byte counts.

davkean:
I'm reopening for consideration.

nkolev92:
The interesting thing that @zivkan's tests show that the difference was really small: https://github.com/NuGet/NuGet.Client/pull/5303#discussion_r1274806397.  So this PR might not be enough to really make a dent. 

Another idea would be to intern this type (within the context of restore). 

A more breaking approach would be to do https://github.com/NuGet/Home/issues/13077, and then we can make this type immutable, and change all the initialization code around it.

I think there's probably parts of this PR that are less controversial than others. 
For example, I think eliminating the empty list creation would make a bigger dent than the bool packing, and is shorter. Maybe we can start with that too.

aortiz-msft:
Closing as per last comment.

PR #5304, opened 2023-07-10
Improve hashing and equality allocations/performance

### LINKED ISSUE #12746: Improve hashing and equality allocations/performance
NuGet.Client has a bunch of redundant allocations in the computation of equality and hash codes.

It should be possible to avoid a lot of `OrderBy` calls in `GetHashCode` calls to `HashCodeCombiner` by using XOR on unordered items.

The `NoAllocEnumerate` method can be improved for `IList<T>` to never allocate a fallback enumerator.

Calls to `HashCodeCombiner.CheckInitialized` can be removed, as more modern C# allows field initializers on structs.

drewnoakes:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12746

## Description

Improve the performance of hashing and equality checks.

- Create a `HashCodeCombiner.AddUnorderedSequence` that avoids needing calls to `OrderBy` that cause buffers to be allocated and sort algorithms to run.
- Improve `NoAllocEnumerate` for `IList<T>` to no longer allocate a fallback enumerator, and instead enumerate the list directly by index.
- Use `NoAllocEnumerate` in `HashCodeCombiner`.
- Improve `EqualityUtility.OrderedEquals` when sequence contains a single item, again, avoiding `OrderBy`.
- Fix incorrect calls to `HashCodeCombiner.GetHashCode` where callers should be using `CombinedHash` instead. The former boxes the struct in order to call the virtual `ValueType.GetHashCode`. Convert `HashCodeCombiner` to a `ref struct` to prevent this occurring in future.
- Fix places where strings were being treated as sequences of characters when adding to a hash.
- Remove `HashCodeCombiner.CheckInitialized`. This method has no practical utility in the existing code. It's possible to define the initial seed in a field initializer in newer versions of C#. Calling it from every method reduces that chance that calls can be inlined, and introduces branches that hurt perf.
- Null annotate some APIs.
- Add unit test class for `HashCodeCombiner`.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/build/Shared/HashCodeCombiner.cs
+++ b/build/Shared/HashCodeCombiner.cs
@@ -42,8 +42,8 @@ namespace NuGet.Shared
             AddHashCode(b ? 1 : 0);
         }
 
-        internal void AddObject<TValue>(TValue? o, IEqualityComparer<TValue> comparer)
-            where TValue : class
+        internal void AddObject<T>(T? o, IEqualityComparer<T> comparer)
+            where T : class
         {
             CheckInitialized();
             if (o != null)
--- a/build/Shared/HashCodeCombiner.cs
+++ b/build/Shared/HashCodeCombiner.cs
@@ -17,13 +17,13 @@ namespace NuGet.Shared
         // seed from String.GetHashCode()
         private const long Seed = 0x1505L;
 
-        private bool _initialized;
-        private long _combinedHash;
+        private long _combinedHash = Seed;
 
-        internal int CombinedHash
+        public HashCodeCombiner()
         {
-            get { return _combinedHash.GetHashCode(); }
         }
+
+        internal int CombinedHash => _combinedHash.GetHashCode();
 
         private void AddHashCode(int i)
         {
@@ -32,20 +32,17 @@ namespace NuGet.Shared
 
         internal void AddObject(int i)
         {
-            CheckInitialized();
             AddHashCode(i);
         }
 
         internal void AddObject(bool b)
         {
-            CheckInitialized();
             AddHashCode(b ? 1 : 0);
         }
 
         internal void AddObject<T>(T? o, IEqualityComparer<T> comparer)
             where T : class
         {
-            CheckInitialized();
             if (o != null)
             {
                 AddHashCode(comparer.GetHashCode(o));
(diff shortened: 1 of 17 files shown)


build/Shared/HashCodeCombiner.cs
```diff
@@ -5,47 +5,43 @@
...
-        private long _combinedHash;
+        private long _combinedHash = Seed;
 
-        internal int CombinedHash
+        public HashCodeCombiner()
         {
-            get { return _combinedHash.GetHashCode(); }
         }
 
+        internal int CombinedHash => _combinedHash.GetHashCode();
```
nkolev92, line 25: I think in general when making refactorings, making minor stylistic changes only makes things harder to review sa every time it is reviewed, people would review something that's a no-op.

nkolev92 (approved):
LGTM!

nkolev92:
Can you please create a matching issue?

Thanks!

build/Shared/HashCodeCombiner.cs
drewnoakes: From memory the IDE showed a diagnostic suggesting this change. I try to reduce the number of diagnostics in the gutter/margin as part of refactoring/tidy ups.

If this style is not preferred, perhaps remove the diagnostic from the solution in the `.editorconfig` file. As it stands, it seems like a request to change this.

drewnoakes:
https://github.com/NuGet/Home/issues/12746

build/Shared/HashCodeCombiner.cs
nkolev92: In #9, we touch on that, https://github.com/NuGet/NuGet.Client/blob/dev/docs/coding-guidelines.md. 

The general idea is that it's preferred to use that in new code, but part of the internal feedback for the turnaround of some of these PRs is the sheer number of line changes.

nkolev92:
The CI is failing due to a machine timeout :(
I retriggered that and hopefully it passes. 

@NuGet/nuget-client can someone shepherd this?

PR #5308, opened 2023-07-12
Address component government alerts

erdembayar:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2372

## Description
Recently https://github.com/NuGet/NuGet.Client/pull/5289 upgraded  System.Security.Cryptography.Pkcs to non-vulnerable version, [still GC alert persists](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=8053545&view=logs&j=cb98efd6-7062-56ef-f642-0176540bbe0a&t=f95ba3d8-e30a-5687-3fc7-3bc14bb0710b&l=3310). 

- Problem was https://www.nuget.org/packages/Microsoft.Build.Tasks.Core/17.3.1 defined in [Directory.Packages.props](https://github.com/NuGet/NuGet.Client/blob/b3c4e8094bbb924408180d700ae5b3d37713587c/Directory.Packages.props#L12) is still bringing same vulnerable version 6.0.1 into graph. 
image
- Added override to version into respective projects to fix this.
- In `NuGet.Build.Tasks.Pack.csproj` we need to remove `ExcludeAssets="runtime" />` because this one is doing ILmerge, without this ilmerge fails with [strong name signature problem](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=8056049&view=logs&j=d80b2849-30d4-52dd-74cd-d6536ba98fe0&t=c9202bea-0353-5c27-a52b-68d3d866d2ce&l=1112), see [here](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=8056049&view=logs&j=d80b2849-30d4-52dd-74cd-d6536ba98fe0&t=86d56ec7-bfd0-5d56-4f04-a8bddd2f234f&l=235).

GC alert is no longer present in CI build after this fix: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=8058016&view=logs&j=cb98efd6-7062-56ef-f642-0176540bbe0a&t=f95ba3d8-e30a-5687-3fc7-3bc14bb0710b&l=2327

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Build.Tasks.Console/NuGet.Build.Tasks.Console.csproj
+++ b/src/NuGet.Core/NuGet.Build.Tasks.Console/NuGet.Build.Tasks.Console.csproj
@@ -34,6 +34,7 @@
     <PackageReference Include="Microsoft.Build.Utilities.Core" ExcludeAssets="Runtime" />
     <!-- System.Security.Cryptography.Xml is a dependency of Microsoft.Build.Tasks.Core. Therefore, we have assigned the same ExcludeAssets value. -->
     <PackageReference Include="System.Security.Cryptography.Xml" ExcludeAssets="Runtime" />
+    <PackageReference Include="System.Security.Cryptography.Pkcs" ExcludeAssets="runtime" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == '$(NETFXTargetFramework)' ">
--- a/src/NuGet.Core/NuGet.Build.Tasks.Console/NuGet.Build.Tasks.Console.csproj
+++ b/src/NuGet.Core/NuGet.Build.Tasks.Console/NuGet.Build.Tasks.Console.csproj
@@ -34,13 +34,16 @@
     <PackageReference Include="Microsoft.Build.Utilities.Core" ExcludeAssets="Runtime" />
     <!-- System.Security.Cryptography.Xml is a dependency of Microsoft.Build.Tasks.Core. Therefore, we have assigned the same ExcludeAssets value. -->
     <PackageReference Include="System.Security.Cryptography.Xml" ExcludeAssets="Runtime" />
-    <PackageReference Include="System.Security.Cryptography.Pkcs" ExcludeAssets="runtime" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == '$(NETFXTargetFramework)' ">
     <Reference Include="Microsoft.Build.Utilities.v4.0" Aliases="MicrosoftBuildUtilitiesv4" />
   </ItemGroup>
-
+  
+  <ItemGroup Condition=" '$(IsCore)' == 'true' ">
+    <PackageReference Include="System.Security.Cryptography.Pkcs" ExcludeAssets="Compile"/>
+  </ItemGroup>
+  
   <ItemGroup>
     <None Include="App.config" Condition=" '$(TargetFramework)' == '$(NETFXTargetFramework)' " />
     <None Include="app.manifest" />
--- a/src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.csproj
+++ b/src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.csproj
@@ -47,6 +47,7 @@
     <PackageReference Include="Microsoft.Build.Utilities.Core" ExcludeAssets="runtime" GeneratePathProperty="true" />
     <!-- System.Security.Cryptography.Xml is a dependency of Microsoft.Build.Tasks.Core. Therefore, we have assigned the same ExcludeAssets value. -->
     <PackageReference Include="System.Security.Cryptography.Xml" ExcludeAssets="Runtime" />
+    <PackageReference Include="System.Security.Cryptography.Pkcs" ExcludeAssets="runtime" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.csproj
+++ b/src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.csproj
@@ -47,7 +47,7 @@
     <PackageReference Include="Microsoft.Build.Utilities.Core" ExcludeAssets="runtime" GeneratePathProperty="true" />
     <!-- System.Security.Cryptography.Xml is a dependency of Microsoft.Build.Tasks.Core. Therefore, we have assigned the same ExcludeAssets value. -->
     <PackageReference Include="System.Security.Cryptography.Xml" ExcludeAssets="Runtime" />
-    <PackageReference Include="System.Security.Cryptography.Pkcs" ExcludeAssets="runtime" />
+    <PackageReference Include="System.Security.Cryptography.Pkcs" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/NuGet.Core/NuGet.Build.Tasks/NuGet.Build.Tasks.csproj
+++ b/src/NuGet.Core/NuGet.Build.Tasks/NuGet.Build.Tasks.csproj
@@ -50,7 +50,7 @@
     <PackageReference Include="Microsoft.Build.Utilities.Core" ExcludeAssets="runtime" />
     <!-- System.Security.Cryptography.Xml is a dependency of Microsoft.Build.Tasks.Core. Therefore, we have assigned the same ExcludeAssets value. -->
     <PackageReference Include="System.Security.Cryptography.Xml" ExcludeAssets="runtime" />
-    <PackageReference Include="System.Security.Cryptography.Pkcs" ExcludeAssets="runtime" />
+    <PackageReference Include="System.Security.Cryptography.Pkcs" ExcludeAssets="compile" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/NuGet.Core/NuGet.Build.Tasks/NuGet.Build.Tasks.csproj
+++ b/src/NuGet.Core/NuGet.Build.Tasks/NuGet.Build.Tasks.csproj
@@ -50,6 +50,7 @@
     <PackageReference Include="Microsoft.Build.Utilities.Core" ExcludeAssets="runtime" />
     <!-- System.Security.Cryptography.Xml is a dependency of Microsoft.Build.Tasks.Core. Therefore, we have assigned the same ExcludeAssets value. -->
     <PackageReference Include="System.Security.Cryptography.Xml" ExcludeAssets="runtime" />
+    <PackageReference Include="System.Security.Cryptography.Pkcs" ExcludeAssets="runtime" />
   </ItemGroup>
 
   <ItemGroup>


src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.csproj
```diff
@@ -47,6 +47,7 @@
     <PackageReference Include="Microsoft.Build.Utilities.Core" ExcludeAssets="runtime" GeneratePathProperty="true" />
     <!-- System.Security.Cryptography.Xml is a dependency of Microsoft.Build.Tasks.Core. Therefore, we have assigned the same ExcludeAssets value. -->
     <PackageReference Include="System.Security.Cryptography.Xml" ExcludeAssets="Runtime" />
+    <PackageReference Include="System.Security.Cryptography.Pkcs" />
```
erdembayar, line 50: No `ExcludeAssets="runtime"` here otherwise IlMerge fails.

donnie-msft (commented):
I believe this looks OK, although I'm not sure why the ILMerge issue came up because of this change? I suppose the worst side effect is there's a change in `nuget.exe` payload with this package, but I hope someone else understands the cause and its implications better than I do.

Ideally, @zivkan would be the best to approve in case he knows more having recently looked at this package and CG.

Also, there's a lot of build links in the description, but those will all be lost in a short timeframe, as they expire. Consider putting whatever detail you want directly into the PR.

zivkan (dismissed):
I missed this in my own PR because I didn't expect CG to be "green" when there's an alert in the PR build. I see that in official builds that the component detection task is orange.

@erdembayar can you please figure out why official and PR builds behave differently (I'm sure it's related to some build variable), and fix the PR pipeline to also be orange, so we can more easily see when our PRs don't fix CG issues correctly.

src/NuGet.Core/NuGet.Build.Tasks.Console/NuGet.Build.Tasks.Console.csproj
```diff
@@ -34,6 +34,7 @@
     <PackageReference Include="Microsoft.Build.Utilities.Core" ExcludeAssets="Runtime" />
     <!-- System.Security.Cryptography.Xml is a dependency of Microsoft.Build.Tasks.Core. Therefore, we have assigned the same ExcludeAssets value. -->
     <PackageReference Include="System.Security.Cryptography.Xml" ExcludeAssets="Runtime" />
+    <PackageReference Include="System.Security.Cryptography.Pkcs" ExcludeAssets="runtime" />
```
zivkan: Since NuGet.Build.Tasks.Console has a PackageReference on NuGet.Buld.Tasks, it shouldn't be necessary to add this package reference in both. Adding to the other one will cause this project, .Console, to get the higher version transitively.

src/NuGet.Core/NuGet.Build.Tasks/NuGet.Build.Tasks.csproj
```diff
@@ -50,6 +50,7 @@
     <PackageReference Include="Microsoft.Build.Utilities.Core" ExcludeAssets="runtime" />
     <!-- System.Security.Cryptography.Xml is a dependency of Microsoft.Build.Tasks.Core. Therefore, we have assigned the same ExcludeAssets value. -->
     <PackageReference Include="System.Security.Cryptography.Xml" ExcludeAssets="runtime" />
+    <PackageReference Include="System.Security.Cryptography.Pkcs" ExcludeAssets="runtime" />
```
zivkan: Actually, I don't think that excluding runtime is correct. Thoughts @jeffkl @nkolev92 ?

Since this is a direct package reference, it will override the transitive package reference from nuget.packaging, and therefore the dll will be excluded. I'm actually surprised tests don't fail at runtime because of this missing dll. Since this package reference is only being added for .NET Core, the biggest risk is if NuGet is the only component causing this package to be included in the .NET SDK, then if we merge this PR, when we insert the pkcs dll will be excluded from the SDK and then NuGet stops working. The way our `dotnet.integration.tests` patch the SDK for testing, it won't catch this type of error.

So, I think leave out the `ExcludeAssets` is the correct solution (honestly, all this would be simpler if transitive packages were only runtime by default, no buildTransitive, no compile). But most of our code has no business calling cryptography APIs, other than NuGet.Packaging, so `ExcludeAssets="compile"` wouldn't be terrible.

src/NuGet.Core/NuGet.Build.Tasks.Console/NuGet.Build.Tasks.Console.csproj
erdembayar: I tried to remove this one then GC alert popped up again, see [here](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=8061706&view=logs&j=cb98efd6-7062-56ef-f642-0176540bbe0a&t=f95ba3d8-e30a-5687-3fc7-3bc14bb0710b&l=2339) for CI build and see [here](https://github.com/NuGet/NuGet.Client/commit/84ffcc8136b957257af514f28ec975e84d321195) for code diff.
Not sure why it's happening, probably if don't pin the version then it still download the v6.0.1 transitive package before it's overloaded by higher version v6.0.4 from `Nuget.Build.Tasks` during the restore process.

src/NuGet.Core/NuGet.Build.Tasks/NuGet.Build.Tasks.csproj
erdembayar: I saw all other package references in this file are already using `ExcludeAssets="runtime"` then followed the convention. Of course I can change to `ExcludeAssets="compile"`. Before doing that I'll wait feedback from @jeffkl @nkolev92

nkolev92: I agree. I don't see a reason why `Exclude="Runtime"` is added here. 
I'm guessing the tests are passing because the SDK carries versions of this assembly. 

What's the reasoning behind adding the `Exclude="Runtime"` here @erdembayar

erdembayar: > What's the reasoning behind adding the Exclude="Runtime" here @erdembayar

@nkolev92 
This convention was here before my change, according to blame it looks @kartheekp-ms  added line 51-52, he might know the reasoning.
https://github.com/NuGet/NuGet.Client/blob/b3c4e8094bbb924408180d700ae5b3d37713587c/src/NuGet.Core/NuGet.Build.Tasks/NuGet.Build.Tasks.csproj#L49-L52

erdembayar: In addition to above it looks like line 49-50 was added by @donnie-msft in https://github.com/NuGet/NuGet.Client/pull/4761. Do you remember why we're using `ExcludeAssets="runtime"` here?

donnie-msft: I don't have to remember, because it's in the first sentence of the PR you linked:

>  MSBuild does not support runtime scenarios where the Microsoft.Build DLL is placed in an output directory.

erdembayar:
> @erdembayar can you please figure out why official and PR builds behave differently (I'm sure it's related to some build variable), and fix the PR pipeline to also be orange, so we can more easily see when our PRs don't fix CG issues correctly.

~~It was env var `cfsNugetWarnLevel` making the difference. I copied it from `official` to `private` pipeline.
Now it's properly raising the GC warning in `private` pipeline with `dev` branch.~~
Env vars in both private and public pipelines looks almost identical, not sure what is making one to flag and not the other one. I'll inquire offline.

src/NuGet.Core/NuGet.Build.Tasks/NuGet.Build.Tasks.csproj
nkolev92: > This convention was here before my change, according to blame it looks @kartheekp-ms added line 51-52, he might know the reasoning.

It's not a convention. The MSBuild packages are special, and System.Security.Cryptography.xml seems to be a result of that as well.

erdembayar: > > This convention was here before my change, according to blame it looks @kartheekp-ms added line 51-52, he might know the reasoning.
> 
> It's not a convention. The MSBuild packages are special, and System.Security.Cryptography.xml seems to be a result of that as well.

I don't know much about MSBuild special packages. Do you want me to ask `dotnet/msbuild folks` to take a look at this?

src/NuGet.Core/NuGet.Build.Tasks.Console/NuGet.Build.Tasks.Console.csproj
nkolev92: Ah, the unused package CG alerts. 
I reached out to the CG folks again. We had a conversation about that some time ago, I'm gonna try to revive it.

erdembayar: Could we modify the NuGet dependency graph calculation to prevent the download of unused packages like this? While it might be tricky, I believe it's still doable. It might not be ideal as a default experience, but it could be offered as an opt-in feature.

nkolev92: It's actually a super easy change, but it makes all subsequent true restore slower as it'd need to redownload packages to compute the graph.

An analyzer should not make assumptions about what's used or not. It should only look at the used ones.

src/NuGet.Core/NuGet.Build.Tasks/NuGet.Build.Tasks.csproj
erdembayar: @nkolev92 
What makes `System.Security.Cryptography.Pkcs` different from ` System.Security.Cryptography.xml` already here? Both of them are transitive packages here. 
It looks @zivkan  and @jeffkl  approved the original https://github.com/NuGet/NuGet.Client/pull/4761, do you have any knowledge about this?

zivkan: The MSBuild packages are special, because `NuGet.Build.Tasks[.*]` are class libraries that implement MSBuild's Task (ITask?) interface. Therefore, the only way to run these libraries is to invoke MSbuild and have a `<UsingTask` element call into our library. But since msbuild.exe (or dotnet.exe) is the process dynamically loading our assembly, Microsoft.Build.*.dll is already loaded by MSBuild. Therefore, we don't have a need to ship those dlls with our task.

When I look at `{vs install directory}\MSBuild` and search for `System.Security.Cryptography.*.dll`, I don't see the pkcs dll. I also don't see the .xml assembly either, but is S.S.C.X a transitive dependency of NuGet.Packaging? My guess is that it's not.

However, this is about .NET (Core) TFMs, not .NET Framework, so looking at MSBuild's VS install isn't really representative.

If you really want to validate, you can build your branch locally, including packing the nupkgs (which build.ps1 does by default), the clone dotnet/sdk, add NuGet.Client's nupkgs directory as a package source, update the nuget.client package version to your local nupkg version number in the props file (eng\Package.Versions.props?), then build the SDK locally, and check if the .pkcs dll still makes it into the built SDK output directory.  But I think just removing the `ExcludeAssets="compile"` for safety would be quicker, as well as more technically correct of our package graph.

erdembayar: Ok. I changed to `ExcludeAssets="compile"` here.

erdembayar:
> 
> @erdembayar can you please figure out why official and PR builds behave differently (I'm sure it's related to some build variable), and fix the PR pipeline to also be orange, so we can more easily see when our PRs don't fix CG issues correctly.

I inquired about this, below is answer:

```
This is happening due to Production/Non-Production classification differences in both pipelines.  The partial task failure (orange build) will only occur if there’s a due date on the alert. 
Due dates are only applicable to production build. 
```

That means we couldn't up light up orange build in private pipeline unless we classify as production, I believe we're not going to do it, because it would cause ton of other unnecessary compliance warnings.

src/NuGet.Core/NuGet.Build.Tasks.Console/NuGet.Build.Tasks.Console.csproj
nkolev92: Is `runtime` supposed to be there?

Can we get a `before` and  `after` of the deps.json file?

That's the most fullproof way of ensuring we didn't change something major.

erdembayar: Is it ok I get it from my local? Or does it have to be from CI pipeline?

erdembayar: Here is comparison from my local build with debug config, I can see 3 changes, please note 3 screenshots here:

Left side is from this branch, right side is from dev branch.
image

image

image

nkolev92: Local is fine. 

Seems like we're missing the PKCS.dll on the left side, which could be an issue if pkcs.dll for whatever reason doesn't get composed into the SDK. 

tldr; We probably have an extra `exclude=runtime` somewhere in the graph.

zivkan: this `PackageReference` is still using `ExcludeAssets="runtime"`. It's only the other file that was changed to `compile`.

zivkan (approved):

src/NuGet.Core/NuGet.Build.Tasks.Console/NuGet.Build.Tasks.Console.csproj
```diff
@@ -39,7 +39,11 @@
   <ItemGroup Condition=" '$(TargetFramework)' == '$(NETFXTargetFramework)' ">
     <Reference Include="Microsoft.Build.Utilities.v4.0" Aliases="MicrosoftBuildUtilitiesv4" />
   </ItemGroup>
-
+  
+  <ItemGroup Condition=" '$(IsCore)' == 'true' ">
+    <PackageReference Include="System.Security.Cryptography.Pkcs" ExcludeAssets="Compile"/>
```
erdembayar, line 44: @nkolev92 
Here is the difference between `current` branch and `dev` for deps.json file.
Left side `current branch` and right is `dev` branch.

image

src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.csproj
erdembayar: Here is the difference between `current` branch and `dev` for deps.json file.
Please note here 2 screenshots:
Left side `current branch` and right is `dev` branch.

1. 
image
2. 
image

src/NuGet.Core/NuGet.Build.Tasks/NuGet.Build.Tasks.csproj
```diff
@@ -50,6 +50,7 @@
     <PackageReference Include="Microsoft.Build.Utilities.Core" ExcludeAssets="runtime" />
     <!-- System.Security.Cryptography.Xml is a dependency of Microsoft.Build.Tasks.Core. Therefore, we have assigned the same ExcludeAssets value. -->
     <PackageReference Include="System.Security.Cryptography.Xml" ExcludeAssets="runtime" />
+    <PackageReference Include="System.Security.Cryptography.Pkcs" ExcludeAssets="compile" />
```
erdembayar, line 53: Here is the difference between `current` branch and `dev` for deps.json file.
Left side `current branch` and right is `dev` branch.

1. 
image
2.
image

erdembayar (commented):
@nkolev92 
I added deps.json file comparison for each changes after latest change. Please review again.

aortiz-msft (approved):
LGTM

PR #5309, opened 2023-07-12
Update coding guidelines to avoid using as keyword for type casting

zivkan:
## Bug

Fixes: N/A

## Description

Update coding guidelines to avoid using `as` keyword for type casting, as it causes `NullReferenceException` to be thrown instead of `InvalidCastException`, making debugging much harder.

In the future, if we can enable [nullable referene types](https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references) throughout the repo, then the compiler will tell us that the code is not null safe. But enabling nullable everywhere takes a lot of effort to fix all the existing issues and annotating APIs. So, this coding guideline is a reminder for us to write better null safe code until we can have the compiler automate null safety for us.

## PR Checklist

- [x] ~PR has a meaningful title~ no product change
- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/docs/coding-guidelines.md
+++ b/docs/coding-guidelines.md
@@ -239,7 +239,36 @@ For example the following are correct:
     }
     ```
 
-Exceptions are allowed when multiple type names coming from different namespaces are available.
+    Exceptions are allowed when multiple type names coming from different namespaces are available.
+
+1. Do not use `as` to cast types
+
+   These are correct:
+
+   ```cs
+   ListItem item = (ListItem)selectedItem; // InvalidCastException if wrong type
+   string name = item.Name;
+
+   if (obj is Type1 t1)
+   {
+       return t1.Value;
+   }
+   else if (obj is Type2 t2)
+   {
+       return t2.Value;
+   }
+   else
+   {
+        throw new InvalidOperationException($"Unexpected type {sender.GetType().Name}");
+   }
+   ```
+
+   This is incorrect:
+
+   ```cs
+   ListItem item = selectedItem as ListItem;
+   string name = item.Name; // NullReferenceException if selectedItem was not ListItem
+   ```
 
 Many of the guidelines, wherever possible, and potentially some not listed here, are enforced by an [EditorConfig](https://editorconfig.org "EditorConfig homepage") file (`.editorconfig`) at the root of the repository.
 
--- a/docs/coding-guidelines.md
+++ b/docs/coding-guidelines.md
@@ -248,7 +248,9 @@ For example the following are correct:
    ```cs
    ListItem item = (ListItem)selectedItem; // InvalidCastException if wrong type
    string name = item.Name;
+   ```
 
+   ```cs
    if (obj is Type1 t1)
    {
        return t1.Value;


docs/coding-guidelines.md
```diff
@@ -239,7 +239,36 @@ For example the following are correct:
...
+
+1. Do not use `as` to cast types
+
+   These are correct:
+
+   ```cs
+   ListItem item = (ListItem)selectedItem; // InvalidCastException if wrong type
+   string name = item.Name;
+
+   if (obj is Type1 t1)
```
erdembayar, line 254: What is this `obj` here? Above 2 lines are related to this one? Looks like 2 different use cases. If it's 2 different cases could you please separate them?

docs/coding-guidelines.md
```diff
@@ -239,7 +239,36 @@ For example the following are correct:
     }
     ```
 
-Exceptions are allowed when multiple type names coming from different namespaces are available.
+    Exceptions are allowed when multiple type names coming from different namespaces are available.
+
+1. Do not use `as` to cast types
```
nkolev92, line 244: While I agree that objectively, `as` casting is probably not necessary, this almost feels like one of those things that's too opinionated. I don't mind accepting it, but that was my first instict.

donnie-msft (commented):
What if I don't want `InvalidCastException` to be thrown? If we `nullable enable` more of our code, will this help address the NREs?

docs/coding-guidelines.md
zivkan: Just about everything in the coding guidelines is opinionated!

In cases where nulls are silently ignored, it can be a correctness issue. I'm thinking of something like `var item = CurrentSelection as Package; if (item is not null) { DoSomething(item); }` to handle the case when a button is clicked when no package is selected. If the type is wrong, customers will have selected an item, but nothing happens and the customer doesn't know why, and there are no errors to help us understand.

In cases of `var item = CurrentSelection as Package; var name = item.Name;` then a `NullReferenceException` is much harder to debug from telemetry compared to an `InvalidCastException` that tells use "Type1 cannot be cast to Type2".

So, I don't understand why this guideline would be less appropriate than most of the other opinionated guidelines in this document. Many others are nothing more than stylistic, this one actually reduces debugging complexity, hence increases productivity.

zivkan:
> What if I don't want InvalidCastException to be thrown? 

Then handle it, like in my second "this is correct" snippet. How is a NullReferenceException any better? NREs are harder to debug, since they provide less information.

> If we nullable enable more of our code, will this help address the NREs?

Yes, enabling nullable would tell the author that the code they're writing is not null safe. But in cases like `DoMethod(value as SomeType)` we'll only get correct nullable warnings if `DoMethod` is already annotated, which is why I've been starting at the bottom of the project dependency graph and trying to work my way up (but it takes a LOT of effort): https://github.com/NuGet/Client.Engineering/issues/2255

docs/coding-guidelines.md
zivkan: I'm not sure why GitHub's not marking the code snippet as "outdated", but I put the two samples in different code blocks, even though that pattern isn't used elsewhere in the document.

docs/coding-guidelines.md
```diff
@@ -239,7 +239,38 @@ For example the following are correct:
...
+   {
+        throw new InvalidOperationException($"Unexpected type {sender.GetType().Name}");
+   }
+   ```
+
+   This is incorrect:
+
+   ```cs
+   ListItem item = selectedItem as ListItem;
+   string name = item.Name; // NullReferenceException if selectedItem was not ListItem
```
erdembayar, line 272: How about conditional accessing? Should we add to correct section above? I feel it's more simpler in my opinion, this way no need to deal with exception.

```
ListItem item = selectedItem as ListItem;
if(item != null)
{
   string name = item.Name;
}
```

erdembayar: Looks like it's already in another thread, resolving.

docs/coding-guidelines.md
aortiz-msft: This guideline makes sense to me. My ask is that we keep an eye where this guideline has an unintended consequence and revisit the guideline if needed.

aortiz-msft (approved):

docs/coding-guidelines.md
```diff
@@ -239,7 +239,38 @@ For example the following are correct:
...
+   else
+   {
+        throw new InvalidOperationException($"Unexpected type {sender.GetType().Name}");
+   }
+   ```
+
+   This is incorrect:
+
+   ```cs
+   ListItem item = selectedItem as ListItem;
```
kartheekp-ms, line 271: Roslyn has an analyzer, [Remove unnecessary cast (IDE0004)](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/style-rules/ide0004) that flags instances when a cast is unnecessary. Perhaps the first step we should take is to change the severity of this analyzer to 'error.' I suspect that we are performing type casting as a safety measure, even in cases where these casts are implicit from the compiler's perspective.

image

aortiz-msft:
Team Review: This is approved to be merged, but please update the PR description to show that this is a step we are taking before we turn on nullable checks on the codebase.

docs/coding-guidelines.md
```diff
@@ -239,7 +239,38 @@ For example the following are correct:
...
+
+   These are correct:
+
+   ```cs
+   ListItem item = (ListItem)selectedItem; // InvalidCastException if wrong type
+   string name = item.Name;
+   ```
+
+   ```cs
+   if (obj is Type1 t1)
```
donnie-msft, line 254: My understanding is this **return/return/throw** pattern is just an example, not a prescriptive pattern for all scenarios. With that understanding, I think it's worth trying and hopefully will help readibility.

donnie-msft (approved):
Thank you for steering the team towards better coding patterns!

PR #5310, opened 2023-07-12
Reduce size of LockFileTargetLibrary

drewnoakes:
## Bug

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1827466

## Description

This class has many fields of `List<T>`, most of which will remain empty throughout the lifetime of the class. Having a field for each of these takes space, and so does the empty list allocated for each.

This commit removes the fields and replaces them with a sparse dictionary such that only those properties that are actually populated will be allocated in memory.

We also introduce the concept of "freezing" this object, after which accessing properties will no longer return a lazily constructed mutable `List<T>`, and instead will return `Array.Empty<T>` which cannot be modified. This prevents code that reads these properties from inadvertently materializing these lazy collections.

Null annotations are added to `LockFileTargetLibrary`.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileTargetLibrary.cs
+++ b/src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileTargetLibrary.cs
@@ -8,80 +8,204 @@ using NuGet.Packaging.Core;
 using NuGet.Shared;
 using NuGet.Versioning;
 
+#nullable enable
+
 namespace NuGet.ProjectModel
 {
     public class LockFileTargetLibrary : IEquatable<LockFileTargetLibrary>
     {
-        public string Name { get; set; }
+        #region Property packing
 
-        public string Framework { get; set; }
+        private readonly record struct PropertyKey(string PropertyName);
 
-        public NuGetVersion Version { get; set; }
+        private static readonly PropertyKey DependenciesKey = new(nameof(Dependencies));
+        private static readonly PropertyKey FrameworkAssembliesKey = new(nameof(FrameworkAssemblies));
+        private static readonly PropertyKey FrameworkReferencesKey = new(nameof(FrameworkReferences));
+        private static readonly PropertyKey RuntimeAssembliesKey = new(nameof(RuntimeAssemblies));
+        private static readonly PropertyKey ResourceAssembliesKey = new(nameof(ResourceAssemblies));
+        private static readonly PropertyKey CompileTimeAssembliesKey = new(nameof(CompileTimeAssemblies));
+        private static readonly PropertyKey NativeLibrariesKey = new(nameof(NativeLibraries));
+        private static readonly PropertyKey BuildKey = new(nameof(Build));
+        private static readonly PropertyKey BuildMultiTargetingKey = new(nameof(BuildMultiTargeting));
+        private static readonly PropertyKey ContentFilesKey = new(nameof(ContentFiles));
+        private static readonly PropertyKey RuntimeTargetsKey = new(nameof(RuntimeTargets));
+        private static readonly PropertyKey ToolsAssembliesKey = new(nameof(ToolsAssemblies));
+        private static readonly PropertyKey EmbedAssembliesKey = new(nameof(EmbedAssemblies));
+        private static readonly PropertyKey PackageTypeKey = new(nameof(PackageType));
 
-        public string Type { get; set; }
+        private readonly Dictionary<PropertyKey, object> _propertyValues = new();
 
-        public IList<PackageDependency> Dependencies { get; set; } = new List<PackageDependency>();
+        private bool _isFrozen;
 
-        public IList<string> FrameworkAssemblies { get; set; } = new List<string>();
+        private IList<T> GetListProperty<T>(PropertyKey key)
+        {
+            if (!_propertyValues.TryGetValue(key, out object? value))
+            {
+                if (_isFrozen)
+                {
+                    return Array.Empty<T>();
+                }
 
-        public IList<string> FrameworkReferences { get; set; } = new List<string>();
+                var list = new List<T>();
+                _propertyValues[key] = list;
+                return list;
+            }
 
-        public IList<LockFileItem> RuntimeAssemblies { get; set; } = new List<LockFileItem>();
+            return (IList<T>)value;
+        }
 
-        public IList<LockFileItem> ResourceAssemblies { get; set; } = new List<LockFileItem>();
+        private void SetListProperty<T>(PropertyKey key, IList<T> list)
+        {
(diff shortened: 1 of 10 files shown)


drewnoakes:
Putting this up for discussion. I think, given NuGet's everything-is-mutable APIs that this "freezing" idea might be a pragmatic way to improve performance and tighten up some guarantees. Perhaps we can get on a call and discuss.

drewnoakes:
Looks like there are some build failures due to `IsExternalInit` not being on an allow list.

drewnoakes:
Looks like we can't use `IsExternalInit` here. Should be easy to patch that up. Will address tomorrow as it's late here.

src/NuGet.Core/NuGet.ProjectModel/NuGet.ProjectModel.csproj
```diff
@@ -15,6 +15,10 @@
     <TargetFramework>$(NETFXTargetFramework)</TargetFramework>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="IsExternalInit" PrivateAssets="all" />
```
nkolev92: In order for us to take a dependency on a new package within our NuGet.Core libraries, we need to ensure that it can be source built. Now this is a source only package so I'm guessing this is ok. 

@NikolaMilosavljevic @MichaelSimons 

Worth understanding if there any licensing disclosure considerations as well. 
I think we'd need to disclose it 3rd party notices, but I'm not sure this being a source packages affects that.

drewnoakes: We don't really need to use the package. It only adds a single file, which is trivial. I've done the equivalent thing in a subsequent commit and removed the package from the PR.

src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileTargetLibrary.cs
```diff
@@ -8,80 +8,204 @@
...
+        private static readonly PropertyKey NativeLibrariesKey = new(nameof(NativeLibraries));
+        private static readonly PropertyKey BuildKey = new(nameof(Build));
+        private static readonly PropertyKey BuildMultiTargetingKey = new(nameof(BuildMultiTargeting));
+        private static readonly PropertyKey ContentFilesKey = new(nameof(ContentFiles));
+        private static readonly PropertyKey RuntimeTargetsKey = new(nameof(RuntimeTargets));
+        private static readonly PropertyKey ToolsAssembliesKey = new(nameof(ToolsAssemblies));
+        private static readonly PropertyKey EmbedAssembliesKey = new(nameof(EmbedAssemblies));
+        private static readonly PropertyKey PackageTypeKey = new(nameof(PackageType));
+
+        private readonly Dictionary<PropertyKey, object> _propertyValues = new();
```
nkolev92, line 36: dictionary has some overhead. I know measuring an object is tricky, but I'm curious about the difference between before and after.

src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileTargetLibrary.cs
```diff
@@ -8,80 +8,204 @@
...
 
-        public NuGetVersion Version { get; set; }
+        private void SetListProperty<T>(PropertyKey key, IList<T> list)
+        {
+            System.Diagnostics.Debug.Assert(!_isFrozen, "Attempting to set a property on a frozen instance.");
+            _propertyValues[key] = list;
+        }
 
-        public string Type { get; set; }
+        public void Freeze()
```
nkolev92, line 63: I have considered this concept for the `PackageSpec` type all up, but I hadn't seen it adopted much. 
Have you seen examples without some of the VS/dotnet repos of this?

nkolev92 (approved):
LGTM. 
A few small comments.

src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileTargetLibrary.cs
drewnoakes: There are 14 properties here, most of which will only ever have a default value.

Some napkin math...

If _no_ properties are allocated, then this approach:

- Removes backing fields for the properties, 14 * 8 = 112 bytes
- Removes empty list allocations, 14 * (24 + 8 + 4 + 4 + 8) = 672 bytes
- Adds a dictionary field and instance, 8 + 24 + 8 + 8 + 4 + 4 + 4 + 4 + 8 + 8 + 8 + 8 = 96 bytes
- Total saving of 688 bytes per instance of `LockFileTargetLibrary`

The incremental cost of each property in the collection is then:

- Entry in dictionary, 4 + 4 + 4 + 8 + 8 = 28 bytes
- List allocation, 24 + 8 + 4 + 4 + 8 = 48 bytes
- Total of 76 bytes

So to break even, we need to store 9 of the 14 properties on the object with a non-empty value.

(Looking at these numbers, we might save a little more by changing `PropertyKey` to be a `byte` rather than a `string` wrapper, though that might also have no impact due to padding.)

src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileTargetLibrary.cs
drewnoakes: The closest pattern to this that I've seen is [dependency properties](https://learn.microsoft.com/en-us/dotnet/desktop/wpf/advanced/dependency-properties-overview?view=netframeworkdesktop-4.8) in WPF and [`Freezable`](https://learn.microsoft.com/en-us/dotnet/desktop/wpf/advanced/freezable-objects-overview?view=netframeworkdesktop-4.8) objects.

src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileTargetLibrary.cs
drewnoakes: Also given N is likely to be low, `Dictionary<>` could be replaced with a more space efficient collection, such as an array of ordered entries and binary search, rather than a hash based container.

src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileTargetLibrary.cs
drewnoakes: `PackageSpec` does look like a good candidate for this pattern as well.

PR #5314, opened 2023-07-14
Pass SettingsLoadContext to LoadSettingsForSpecificConfigs

### LINKED ISSUE #12737: Lock contention thread pool issues caused by LoadSettings not passing settingsLoadingContext to LoadSettingsForSpecificConfigs
### NuGet Product Used

Other/NA

### Product Version

Visual Studio 17.6

### Worked before?

_No response_

### Impact

Other

### Repro Steps & Context

Thread pool exhaustion issue eventType=threadpool&failureHash=8c6a84c3-cc30-9e23-84e0-7b34342ef23c (PRISM)

I believe the source of this problem is that LoadSettings function shown below hasn't passed settingsLoadingContext to LoadSettingsForSpecificConfigs. This leads the machine level settings file is not cached during project evaluation, so each time it is read again, and when we have multiple evaluations on the same time in the thread pool, it leads bad contention to wait on this to be read and blocks multiple threads at the same time.

It seems important to cache the machine level cache as it is commonly used.

image

### Verbose Logs

_No response_

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12737

## Description

When the `SettingsLoadContext` class was added, it was missed that it could be passed from `Settings.LoadDefaultSettings()` to `Settings.LoadSettingsForSpecificConfigs()`.  There a few code paths that call this including the NuGet-based MSBuild project SDK resolver and static graph-based restore.  This leads to the machine-wide NuGet.config being read multiple times instead of being cached.

* Update `SettingsLoadingContext.GetOrCreateSettingsFile()` to guard against being called with a non-rooted path since it expects only rooted paths.  
* Update call to `SettingsLoadingContext.GetOrCreateSettingsFile()` to pass a rooted path

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception 
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs
+++ b/src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs
@@ -443,7 +443,8 @@ namespace NuGet.Configuration
                 validSettingFiles,
                 machineWideSettings,
                 loadUserWideSettings,
-                useTestingGlobalPath);
+                useTestingGlobalPath,
+                settingsLoadingContext);
         }
 
         /// <summary>
--- a/src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs
+++ b/src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs
@@ -774,6 +774,11 @@ namespace NuGet.Configuration
             {
                 if (settingsLoadingContext != null)
                 {
+                    if (!Path.IsPathRooted(settingsPath) && !string.IsNullOrWhiteSpace(settingsRoot))
+                    {
+                        settingsPath = Path.Combine(settingsRoot, settingsPath);
+                    }
+
                     return settingsLoadingContext.GetOrCreateSettingsFile(settingsPath, isMachineWideSettings, isAdditionalUserWideConfig);
                 }
 
--- a/src/NuGet.Core/NuGet.Configuration/Settings/SettingsLoadingContext.cs
+++ b/src/NuGet.Core/NuGet.Configuration/Settings/SettingsLoadingContext.cs
@@ -57,6 +57,11 @@ namespace NuGet.Configuration
                 throw new ArgumentNullException(nameof(filePath));
             }
 
+            if (!Path.IsPathRooted(filePath))
+            {
+                throw new ArgumentException("SettingsLoadingContext requires a rooted path", nameof(filePath));
+            }
+
             Lazy<SettingsFile> settingsLazy = _cache.GetOrAdd(
                 filePath,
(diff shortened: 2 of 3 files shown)


lifengl (approved):
looks good to me.

jeffkl:
@nkolev92 this seems to have introduced a behavior change, any ideas on what to do about it?


```diff
        "configFilePaths": [
          "D:\\nuget\\.test\\work\\ee0ae8f4\\9a1a6dd0\\NuGet.Config",
          "D:\\nuget\\NuGet.Config",
          "C:\\Users\\jeffkl\\AppData\\Roaming\\NuGet\\NuGet.Config",
-         "C:\\Program Files (x86)\\NuGet\\Config\\Microsoft.VisualStudio.FallbackLocation.config",
+         "D:\\nuget\\.test\\work\\ee0ae8f4\\9a1a6dd0\\solution\\Microsoft.VisualStudio.FallbackLocation.config",
-         "C:\\Program Files (x86)\\NuGet\\Config\\Microsoft.VisualStudio.Offline.config"
+         "D:\\nuget\\.test\\work\\ee0ae8f4\\9a1a6dd0\\solution\\Microsoft.VisualStudio.Offline.config"
        ],
```

jeffkl:
Oh I wonder if its because the SettingsLoadContext is static and being shared for each test.  I'll see if I need to reset it for each test 😢

nkolev92:
It's got to be the statics. 
Given that this is a big change, I almost feel like existing tests shouldn't be covering it correct?

Because if they are, it's because the tests aren't correct, or if they aren't, then the change isn't correct. 

Anyway, I'm guessing the tests that you'll need to fix are likely the answer.

zivkan:
> I'll see if I need to reset it for each test

Considering MSBuild reuses nodes and therefore the same process can be used in different projects/solutions, doesn't this mean it's a real product bug, and not just some testing issue?  Making tests run sequentially and then resetting the static in between will mask the bug.

aortiz-msft (commented):
LGTM. Are you planning on adding UTs for the change?

ghost:

jeffkl:
Turns out there's a bug in our code, the `SettingsLoadingContext` requires a full path and this change introduces a call that is only being passed a file name when loading machine wide settings.  I'll push a commit to fix this.

zivkan (approved):

aortiz-msft (approved):
:shipit:

PR #5315, opened 2023-07-17
Fix NuGetAudit severity level mapping (#5313)

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12743

Regression? No, no feature

## Description

Cherry picks #5313 onto `release-6.7.x` branch.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
@@ -72,6 +72,11 @@ namespace NuGet.Commands.Restore.Utility
             stopwatch.Stop();
             DownloadDurationSeconds = stopwatch.Elapsed.TotalSeconds;
 
+            if (allVulnerabilityData?.Exceptions is not null)
+            {
+                ReplayErrors(allVulnerabilityData.Exceptions);
+            }
+
             if (allVulnerabilityData is null || !AnyVulnerabilityDataFound(allVulnerabilityData.KnownVulnerabilities))
             {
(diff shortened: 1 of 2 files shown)


src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -368,31 +369,31 @@ private int ParseAuditLevel()
 
             if (auditLevel == null)
             {
-                return 1;
+                return 0;
```
aortiz-msft, line 372: >return 0; [](http://example.com/codeflow?start=16&length=9)

Could we add a comment on when auditLevel is expected to be null? I'm guessing this is the default case, so it would be good to comment that and also comment why the default case was decided to be equivalent to low.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -368,31 +369,31 @@ private int ParseAuditLevel()
...
             }
 
             if (string.Equals("low", auditLevel, StringComparison.OrdinalIgnoreCase))
```
aortiz-msft, line 375: >low [](http://example.com/codeflow?start=31&length=3)

I'm a bit worried that these are not constants defined somewhere. A typo in a later change could make the strings to be mismatched when written/read from the properties.

PR #5317, opened 2023-07-17
[ci] don't sign on build or pack

zivkan:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2393

Regression? no

## Description

Set MSBuild property on build and pack steps to skip microbuild signing, since we have a separate step to batch sign. It doesn't affect the build step since we've somehow disabled sign on build another way. But the pack step decreased in duration from over 40 minutes (typically 45 to 50 minutes), to just 20 seconds. The batch sign of nupkgs afterwards takes just 3 minutes, since the nupkgs get signed in parallel, rather than sequentially during pack (which is itself is probably a bug, but signing in another task is a better and easier fix)

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/eng/pipelines/templates/Build_and_UnitTest.yml
+++ b/eng/pipelines/templates/Build_and_UnitTest.yml
@@ -120,7 +120,7 @@ steps:
   inputs:
     solution: "build\\build.proj"
     configuration: "$(BuildConfiguration)"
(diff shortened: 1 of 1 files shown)


erdembayar (approved):
Wow, saving 40 min from each CI build, that is awesome.

aortiz-msft (approved):
👏

PR #5318, opened 2023-07-18
Make dotnet list package http communication thread safe

### LINKED ISSUE #12755: dotnet list package errors with Object reference not set to an instance of an object
### NuGet Product Used

dotnet.exe

### Product Version

all?

### Worked before?

_No response_

### Impact

### Repro Steps & Context

On very rare occasions (although somehow our CI test `DotnetListPackageTests.ListPackage_WithHttpSource_Warns` is hitting it very frequently), `dotnet list package`, with one of the arguments that makes HTTP requests (`--outdated`, `--vulnerable`, or `--deprecated`) will output a NullReferenceException message, followed by the help message

> error: Object reference not set to an instance of an object.
> 
> Usage: NuGet.CommandLine.XPlat.dll package list [arguments] [options]

### Verbose Logs

_No response_

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12755

Regression? no, I don't think so

## Description

1. Fixed thread safety issue

`ListPackageCommandRunner` was fanning out, getting package metadata from all package sources in parallel, and then adding all versions to a single `List<T>`. There was no synchronization to ensure that two package sources that return results at exactly the same time don't cause problems.

The way the data is fetched and the in-memory model updated was more or less rewritten. Where work is fanned out concurrently, all the parallel work has zero side effects, and returns results to its orchestrator, which collects the results single-threaded.

2. Removed redundant code that fans out to package sources a second time

List package was getting "current version" and "all version" metadata seperately. However, both of these come from the same protocol resource (registration). So, rather than getting more or less the same data twice, just get it once.

3. Changed some data types from `IList` and `IEnumerable` to `List`, so they can be enumerated without allocating an enumerator on the heap, but more importantly, to be able to get `.Count`.
4. Changed the test that's failing most often to use an environment variable to use the SimpleTestPathContext's http cache directory. This avoids polluting the real http cache directory when running tests locally.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception: I'm hoping existing tests cover everything. This PR isn't intended to change any functional behaviour, just address a multi-threaded timing bug that's impossible to reliably test.
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/ListPackageCommandRunner.cs
```diff
@@ -112,33 +111,33 @@ internal async Task<(int, ListPackageReportModel)> GetReportDataAsync(ListPackag
...
-                    if (packages.Any())
+                    if (frameworks.Count > 0)
                     {
                         if (listPackageArgs.ReportType != ReportType.Default)  // generic list package is offline -- no server lookups
                         {
                             PopulateSourceRepositoryCache(listPackageArgs);
                             WarnForHttpSources(listPackageArgs, projectModel);
-                            await GetRegistrationMetadataAsync(packages, listPackageArgs);
-                            await AddLatestVersionsAsync(packages, listPackageArgs);
+                            var metadata = await GetPackageMetadataAsync(frameworks, listPackageArgs);
```
erdembayar, line 132: IIRC we use only explicit type?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/ListPackageCommandRunner.cs
```diff
@@ -265,75 +264,67 @@ public static bool FilterPackages(IEnumerable<FrameworkPackages> packages, ListP
...
-            AddPackagesToDict(packages, packagesVersionsDict);
+            List<string> allPackages = GetAllPackageIdentifiers(targetFrameworks, listPackageArgs.IncludeTransitive);
+            var packageMetadataById = new Dictionary<string, List<IPackageSearchMetadata>>(capacity: allPackages.Count);
 
-            //Prepare requests for each of the packages
-            var getLatestVersionsRequests = new List<Task>();
-            foreach (var package in packagesVersionsDict)
+            // We're (probably) going to make a bunch of HTTP requests, so limit to max 4 packages in parallel to avoid being unkind to servers.
+            // Note, each package will also run all sources in parallel, so the max concurrent HTTP requests is higher.
+            int taskCount = Math.Min(allPackages.Count, 4);
```
erdembayar: Should this `4` would be defined as const on top?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/ListPackageCommandRunner.cs
```diff
@@ -265,75 +264,67 @@ public static bool FilterPackages(IEnumerable<FrameworkPackages> packages, ListP
...
-                        package.Key,
-                        listPackageArgs,
-                        packagesVersionsDict));
+                tasks[packageIndex] = GetPackageVersionsAsync(allPackages[packageIndex], listPackageArgs);
             }
 
-            // Make requests in parallel.
-            await RequestNuGetResourcesInParallelAsync(getLatestVersionsRequests);
+            // throttling steady state
+            while (packageIndex < allPackages.Count)
```
erdembayar: Do we need to reset `packageIndex` to 0 before we do this?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/ListPackageCommandRunner.cs
```diff
@@ -265,75 +264,67 @@ public static bool FilterPackages(IEnumerable<FrameworkPackages> packages, ListP
...
+                {
+                    if (tasks[i].IsCompleted)
+                    {
+                        KeyValuePair<string, List<IPackageSearchMetadata>> result = await tasks[i];
+                        packageMetadataById[result.Key] = result.Value;
 
-            //Save latest versions within the InstalledPackageReference
-            await GetVersionsFromDictAsync(packages, packagesVersionsDict, listPackageArgs);
-        }
+                        tasks[i] = GetPackageVersionsAsync(allPackages[packageIndex++], listPackageArgs);
```
erdembayar: IIRC we already have other throttling pattern code elsewhere. Should we reuse instead of introducing new one?

erdembayar:
> Changed some data types from IList and IEnumerable to List, so they can be enumerated without allocating an enumerator on the heap, but more importantly, to be able to get .Count.

Just wondering what prevents us using List more often than `IList`, `IEnumerable` if former one is more memory efficient? When to choose one best suited for scenario?

erdembayar (commented):
Few questions

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/ListPackageCommandRunner.cs
zivkan: why? It's only used in one place, and there's a comment explaining it.  Constants are useful when it's used in multiple places, so when the value changes, it doesn't need to be changed in multiple places. But that doesn't apply here.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/ListPackageCommandRunner.cs
zivkan: no, because then the first 4 packages would be fetched twice.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/ListPackageCommandRunner.cs
zivkan: yes, ideally. I spent a short time searching and didn't find anything useful.

There's a `SemaphoreThottle`, but I don't like the API, because it just has a `Wait()` (and I don't understand what the type provides over using `Semaphore` or `SemaphoreSlim` directly). But I don't want to start all tasks in parallel, and have the tasks internally throttle. I'd rather throttle the creation of the tasks that are being awaited. Uses less memory.

But I have a much higher tolerance for boilerplate code than most people seem to. To me, reusable code has two primary purposes. Firstly, to avoid implementation bugs. If someone else has already implemented it, hopefully it's bug free. Secondly, when requirements change, there's a single place to change and the entire product now implements the change.  This method is 30 lines, so hopefully fits on most peoples screens without scrolling, so not unreasonable to read and look for bugs. It doesn't contain business logic that needs to be applied in other areas of the product. So to me, it's not an important use case for reusing code 🤷

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/ListPackageCommandRunner.cs
donnie-msft: I find this opinionated, but even for 1 use, anything that's a special constant I like to put high up in a class as a constant.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetListPackageTests.cs
```diff
@@ -473,13 +478,13 @@ public async Task ListPackage_WithHttpSource_Warns()
 
             using var mockServer = new FileSystemBackedV3MockServer(pathContext.PackageSource);
             mockServer.Start();
-            pathContext.Settings.AddSource("http-source", mockServer.ServiceIndexUri);
+                pathContext.Settings.AddSource("http-source", mockServer.ServiceIndexUri);
```
donnie-msft: is the whitespace correct?

donnie-msft (dismissed):
Thanks for figuring this out! I was trying to figure out the test failure as well.
Tested your branch and the test didn't fail after 23minutes!
image

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetListPackageTests.cs
zivkan: 🤦  https://developercommunity.visualstudio.com/t/Make-it-more-obvious-when-compare-with/10418789

donnie-msft (approved):

PR #5319, opened 2023-07-19
Add vulnerability severity enum to NuGet.Protocol public API

### LINKED ISSUE #12781: Changing PackageVulnerabilityInfo severity from int to enum
### NuGet Product(s) Affected

NuGet SDK

### Current Behavior

In NuGet.Protocol 6.7, `PackageVulnerabilityInfo.Severity` is an `int`. This is effectively what the underlying protocol requires, but it makes it harder for developers using NuGet.Protocol (and is a contributing factor for why https://github.com/NuGet/Home/issues/12743 happened)

### Desired Behavior

`.Severity` should be an enum, so that it's easy for developers building tooling around vulnerabilities to know what all the expected values are, and don't make mistakes in mapping to the underlying `int` values.

However, this is a breaking change in NuGet.Protocol.

### Additional Context

_No response_

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12781

Regression? no, new feature

## Description

Address feedback on release-6.7.x cherry-pick PR: https://github.com/NuGet/NuGet.Client/pull/5315

Add enum with expected package vulnerability severity levels, mapping them to the specific int values documented in the protocol: https://learn.microsoft.com/en-us/nuget/api/vulnerability-info#vulnerability-page

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception: existing tests cover the changed code
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
@@ -177,11 +177,11 @@ namespace NuGet.Commands.Restore.Utility
 
                     foreach (var advisory in auditInfo.GraphsPerVulnerability.Keys)
                     {
-                        int severity = advisory.Severity;
-                        if (severity == (int)PackageVulnerabilitySeverity.Low) { Sev0DirectMatches++; }
-                        else if (severity == (int)PackageVulnerabilitySeverity.Moderate) { Sev1DirectMatches++; }
-                        else if (severity == (int)PackageVulnerabilitySeverity.High) { Sev2DirectMatches++; }
-                        else if (severity == (int)PackageVulnerabilitySeverity.Critical) { Sev3DirectMatches++; }
+                        PackageVulnerabilitySeverity severity = (PackageVulnerabilitySeverity)advisory.Severity;
+                        if (severity == PackageVulnerabilitySeverity.Low) { Sev0DirectMatches++; }
+                        else if (severity == PackageVulnerabilitySeverity.Moderate) { Sev1DirectMatches++; }
+                        else if (severity == PackageVulnerabilitySeverity.High) { Sev2DirectMatches++; }
+                        else if (severity == PackageVulnerabilitySeverity.Critical) { Sev3DirectMatches++; }
                         else { InvalidSevDirectMatches++; }
                     }
                 }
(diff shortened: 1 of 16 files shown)


zivkan:
@aortiz-msft if we can commit to servicing NuGet 6.7 before it ships GA, we can also change this public API and possibly remove a bunch of the `(int)` casts in this PR: https://github.com/NuGet/NuGet.Client/blob/b03901bb24c64692c85a949c996bf057ba506b92/src/NuGet.Core/NuGet.Protocol/Model/PackageVulnerabilityInfo.cs#L28

martinrrm (dismissed):

martinrrm:
@zivkan Are these [NuGetLogCode's comments](https://github.com/NuGet/NuGet.Client/blob/59e4937ea73f840f1373dce721bd8c68f285eb63/src/NuGet.Core/NuGet.Common/Errors/NuGetLogCode.cs#L350-L367) correct? All of the code's say "Low severity".

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -177,10 +178,10 @@ private void CheckPackageVulnerabilities(IReadOnlyList<IReadOnlyDictionary<strin
...
                     {
                         int severity = advisory.Severity;
-                        if (severity == 0) { Sev0DirectMatches++; }
-                        else if (severity == 1) { Sev1DirectMatches++; }
-                        else if (severity == 2) { Sev2DirectMatches++; }
-                        else if (severity == 3) { Sev3DirectMatches++; }
+                        if (severity == (int)PackageVulnerabilitySeverity.Low) { Sev0DirectMatches++; }
+                        else if (severity == (int)PackageVulnerabilitySeverity.Moderate) { Sev1DirectMatches++; }
+                        else if (severity == (int)PackageVulnerabilitySeverity.High) { Sev2DirectMatches++; }
+                        else if (severity == (int)PackageVulnerabilitySeverity.Critical) { Sev3DirectMatches++; }
```
kartheekp-ms: I see almost identical code in the same class file lines 195 to 198. Can we extract these lines to a separate method and invoke in these places it to avoid duplication? I am okay with creating a follow-up PR to address this comment.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -374,19 +375,19 @@ private int ParseAuditLevel()
```
kartheekp-ms, line 375: Should we return `PackageVulnerabilitySeverity.Low` in this case?

zivkan: Very weird that the compiler didn't warn

src/NuGet.Core/NuGet.Protocol/Model/PackageVulnerabilitySeverity.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.Protocol
+{
+    public enum PackageVulnerabilitySeverity
+    {
+        Low = 0,
```
aortiz-msft, line 8: Should we also have a value for Unknown, just in case the service doesn't give us the severity for some reason?  #Resolved

aortiz-msft (changes requested):

src/NuGet.Core/NuGet.Protocol/Model/PackageVulnerabilitySeverity.cs
aortiz-msft: I see you added Unknown. Resolving.

aortiz-msft (approved):
The code looks a lot more readable to me. Thank you for making these changes.

PR #5321, opened 2023-07-20
Fix case sensitivity of runtime dependency sets during merge

### LINKED ISSUE #12757: NuGet doesn't restore RID-specific assets for Microsoft.DotNet.ILCompiler
### NuGet Product Used

dotnet.exe

### Product Version

8.0.100-preview.7.23361.25

### Worked before?

8.0.100-preview.7.23361.20

### Impact

### Repro Steps & Context

csproj:
```xml

  <PropertyGroup>
    <OutputType>Exe</OutputType>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.Dotnet.ILCompiler"
                      Version="8.0.0-preview.6.23316.3" />
  </ItemGroup>

</Project>
```

Run a RID-specific restore operation, for example `dotnet restore -r linux-x64`. Observe that `project.assets.json` doesn't contain an entry for `runtime.linux-x64.microsoft.dotnet.ilcompiler`. This was working previously.

I tracked this down to the following SDK update: https://github.com/dotnet/sdk/pull/33913, and narrowed it down a bit more to the nuget commit range https://github.com/nuget/nuget.client/compare/13f2b6af6e...ba16cb02 by building SDK with different nuget package versions.

This is blocking AOT compilation for .NET (see https://github.com/dotnet/runtime/issues/89040).

### Verbose Logs

_No response_

drewnoakes:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12757

Regression? Last working version: Regressed in #5248

## Description

PR #5248 reworked how construction of `RuntimeDescription` instances happens, partly to reduce the number of redundant collection copies that occurred.

As an example of that innefficiency, the prior `RuntimeDesciption.Merge` method would build a dictionary of merged `RuntimeDependencySet` instances, then call a `RuntimeDescription` constructor that would copy those into a new dictionary. The PR added a new private constructor that allowed the copy to be avoided.

That change introduced a bug however. It turns out that during the redundant copy a case-insensitive key comparer was specified, while the `Merge` method's dictionary was case-sensitive. So if two different `RuntimeDependencySet` identifiers differed only by case, then the collection created internally within the `Merge` method would contain both instances, then the constructor call would merge them during the subsequent `ToDictionary` call.

The fix is to ensure that the `Merge` method uses the correct `StringComparer` for `RuntimeDependencySet` identifiers, to preserve the prior behaviour.

Several unit tests have been added to validate the behaviour of `RuntimeDescription.Merge`.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeDescription.cs
+++ b/src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeDescription.cs
@@ -104,7 +104,7 @@ namespace NuGet.RuntimeModel
             }
 
             // Merge dependency sets
-            var newSets = new Dictionary<string, RuntimeDependencySet>();
+            var newSets = new Dictionary<string, RuntimeDependencySet>(StringComparer.OrdinalIgnoreCase);
             foreach (var dependencySet in left.RuntimeDependencySets.Values)
             {
                 newSets[dependencySet.Id] = dependencySet;
--- a/test/NuGet.Core.Tests/NuGet.Packaging.Test/RuntimeModelTests/RuntimeDescriptionTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Packaging.Test/RuntimeModelTests/RuntimeDescriptionTests.cs
@@ -1,0 +1,93 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+#nullable enable
+
+using System;
+using System.Linq;
+using Xunit;
+
+namespace NuGet.RuntimeModel;
+
+public sealed class RuntimeDescriptionTests
+{
+    [Fact]
+    public void Merge_DifferentIdsThrows()
+    {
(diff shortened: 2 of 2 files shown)


zivkan (approved):

lewing:
Is there a reason this isn't flowing to 17.8, we either need it for .NET 8.0-preview7 or we will have to roll back to a previous version https://github.com/dotnet/sdk/pull/34126#issuecomment-1644062795

zivkan:
~The official build didn't complete before our daily insertion.~  ~yes it did, I somehow got to the wrong link when I was trying to follow the commit's build 😕~

~for reasons I don't understand, our Azure DevOps selected the previous build when running today's daily insertion.~

I think my first comment was right. The build for this commit finished 22 minutes after our daily insertion pipeline ran.

lewing:
I assume it will flow tomorrow?

zivkan:
yes, I expect it to.

PR #5324, opened 2023-07-21
Use singleton NuGetFrameworkSorter instance

### LINKED ISSUE #12680: NuGet.Client allocates many instances of comparers

drewnoakes:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12680

## Description

Continues the approach taken in #5260.

Comparer classes tend not to have any instance state, or to come from a small pool of related comparers when they do. They also tend to be invoked in tight loops, where allocations can add up. This makes them great candidates for reuse.

This code makes all use of `NuGetFrameworkSorter` occur through a singleton instance, avoiding allocating temporary objects when comparisons are needed.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.PowerShellCmdlets/Utils/InstalledPackageEnumerator.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.PowerShellCmdlets/Utils/InstalledPackageEnumerator.cs
@@ -241,7 +241,7 @@ namespace NuGetConsole.Host.PowerShell
                 // Process each framework/id/version once to avoid duplicate work
                 // Packages may have different dependency orders depending on the framework, but there is 
                 // no way to fully solve this across an entire solution so we make a best effort here.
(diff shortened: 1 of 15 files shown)


aortiz-msft (approved):
LGTM. Needs an approval from @zivkan , @nkolev92, or @jeffkl

zivkan (approved):

PR #5325, opened 2023-07-21
Remove allocations in hash code computations

### LINKED ISSUE #12746: Improve hashing and equality allocations/performance

drewnoakes:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12746

## Description

Builds on the work started in #5304.

Rather than ordering items to add them to the hash code in a stable fashion, we re-use the new `AddUnorderedSequence` method. That method does not need to order its inputs, meaning we can avoid allocating buffers for the inputs, and running sort algorithms on those items. Instead we can just enumerate the items once while computing the hash, keeping running state on the stack.

We also add `NoAllocEnumerate` calls within `HashCodeCombiner` to avoid allocating enumerators where possible.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/test/NuGet.Core.Tests/NuGet.Packaging.Test/Core/comparers/PackageDependencyInfoComparerTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Packaging.Test/Core/comparers/PackageDependencyInfoComparerTests.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Linq;
 using NuGet.Packaging.Core;
 using NuGet.Shared;
 using NuGet.Versioning;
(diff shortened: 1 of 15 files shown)


test/NuGet.Core.Tests/NuGet.Packaging.Test/Core/comparers/PackageDependencyInfoComparerTests.cs
```diff
@@ -120,8 +119,8 @@ public void GetHashCode_WhenArgumentIsValid_ReturnsHashCode()
         {
             var combiner = new HashCodeCombiner();
 
-            combiner.AddObject(PackageIdentityComparer.Default.GetHashCode(DependencyInfoA));
-            combiner.AddObject(DependencyInfoA.Dependencies.First());
+            combiner.AddObject(DependencyInfoA, PackageIdentityComparer.Default);
+            combiner.AddUnorderedSequence(DependencyInfoA.Dependencies);
```
drewnoakes, line 123: The test had to change here because `AddUnorderedSequence` includes the count of items in the hash (which is required to prevent errors when using XOR), and the previous code just added each item. So the test could previously just add the single dependency and have it match, but now it needs to call `AddUnorderedSequence` so that the count can be included too.

zivkan (approved):

PR #5333, opened 2023-07-25
Add nullable annotations to NuGet.Common

### LINKED ISSUE #12775: Add nullable annotations to NuGet.Common
NuGet.Common is [a package on nuget.org](https://www.nuget.org/packages/NuGet.Common/), and is part of the [NuGet Client SDK](https://learn.microsoft.com/nuget/reference/nuget-client-sdk). We are adding null annotations, so projects using [nullable reference types](https://learn.microsoft.com/dotnet/csharp/nullable-references) will benefit from compiler null checking, leading to improved software reliability (fewer NullReferenceExceptions).

However, projects that use NuGet.Common's APIs, and have `nullable` enabled, may encounter compiler errors when upgrading to a version of the package with annotations.

The work to annotate the package is intended not to have binary (ABI) breaking changes, and we won;t be changing any method signatures. However, while annotating we might find bugs that need small changes in behaviour in order to fix any compiler errors preventing NuGet itself from building.

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12775

## Description

- Add `<Nullable>enable</Nullable>` to NuGet.Common and test project.
- Annotate all public APIs
- Fix all remaining compiler errors

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception: no functionality changes, existing tests are sufficient
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Core/NuGet.Common/PathUtil/NuGetEnvironment.cs
```diff
@@ -1,6 +1,9 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+// BCL annotations on Environment.GetEnvironmentVariable makes this file difficult to annotate in .NET 5+
+#nullable disable
```
zivkan, line 5: This file nearly made me give up. .NET Core 1.0 didn't have `Environment.GetFolderPath`, so this file has `#if` to compile different code for .NET Framework and everything else. On .NET Framework it uses `GetFolderPath`, which is annotated in .NET 5+ to not return nulls. But in our .NET Standard (and .NET (Core) if we add those TFMs) it will use `Environment.GetEnvironmentVariable`, which can return null.

However, .NET Core 2.0 added `GetFolderPath`, so I think that porting the .NET Standard code to use that. However, it's significantly higher risk than the rest of this PR, so I want to do it in a separate PR in case it needs to be reverted.

src/NuGet.Core/NuGet.Common/Telemetry/TelemetryActivity.cs
```diff
@@ -29,16 +29,24 @@ public class TelemetryActivity : IDisposable
...
 
         private TelemetryActivity(Guid parentId, TelemetryEvent telemetryEvent, Guid operationId)
         {
             if (telemetryEvent != null)
             {
+                if (telemetryEvent.Name is null)
+                {
+                    throw new ArgumentException(paramName: nameof(telemetryEvent), message: "Property 'Name' must not be null");
+                }
                 _telemetryActivity = NuGetTelemetryService?.StartActivity(telemetryEvent.Name);
```
zivkan, line 42: This exposes an API design problem. We have code creating instances of `TelemetryEvent` with a null name, but this code will throw a NullReferenceException if such an instance is used here.

src/NuGet.Core/NuGet.Common/Telemetry/TelemetryActivity.cs
```diff
@@ -29,16 +29,24 @@ public class TelemetryActivity : IDisposable
...
             }
+            else
+            {
+                Debug.Fail("Looking at all references to the static Create methods, I don't think this code path is possible.");
```
zivkan, line 46: I didn't feel comfortable deleting the `if != null` statement, but looking at usage it should never be null. I thought maybe this `Debug.Fail` will give us confidence in a few weeks if none of us hit this debug point, so we can make the change later.

martinrrm (dismissed):
I didn't review the tests but I trust they are good since CI is passing.

src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/Formatters/ILogMessageFormatter.cs
```diff
@@ -121,16 +121,18 @@ private ILogMessageFormatter()
...
 
                                 for (var i = 0; i < targetGraphsCount; ++i)
                                 {
                                     string targetGraph = reader.ReadString();
 
-                                    targetGraphs.Add(targetGraph);
+                                    list.Add(targetGraph);
                                 }
+
+                                targetGraphs = list;
```
aortiz-msft, line 135: Would you mind helping me understand why the naming change is needed?

aortiz-msft:
Reviewed until src/NuGet.Core/NuGet.Common/Migrations/Migration1.cs. Might need an extra day to go through the rest.

jeffkl (dismissed):

src/NuGet.Core/NuGet.Common/PathUtil/FileUtility.cs
```diff
@@ -92,7 +92,12 @@ public static void Replace(Action<string> writeSourceFile, string destFilePath)
                 throw new ArgumentNullException(nameof(destFilePath));
             }
 
-            var tempPath = GetTempFilePath(Path.GetDirectoryName(destFilePath));
+            string? destFileDirectory = Path.GetDirectoryName(destFilePath);
+            if (destFileDirectory is null)
+            {
+                throw new ArgumentException(paramName: destFilePath, message: "Path.GetDirectoryName(destFilePath) returned null");
```
aortiz-msft, line 98: >Path.GetDirectoryName(destFilePath) returned null [](http://example.com/codeflow?start=79&length=49)

Does this need to be localized? At the minimum, we should have a string defined somewhere so that the exception text is guaranteed to be the same for folks who are parsing the text. Same for all other instances.

src/NuGet.Core/NuGet.Common/PathUtil/PathUtility.cs
```diff
@@ -455,30 +473,21 @@ private static bool CheckIfFileSystemIsCaseInsensitive()
 
         private static bool CheckCaseSenstivityRecursivelyTillDirectoryExists(string path, out bool ignoreResult)
```
aortiz-msft: Could we correct the typo here as part of this change?

aortiz-msft: Also, it's not 100% clear to me what this method is used for. I know it's a pre-existing method, but could we clarify the method name further or add a comment to line 487?

src/NuGet.Core/NuGet.Common/Telemetry/TelemetryActivity.cs
aortiz-msft: Sounds good to me.

test/NuGet.Core.Tests/NuGet.Common.Test/SynchronizationTest.cs
```diff
@@ -149,23 +149,6 @@ public async Task ConcurrencyUtilityDoesntBlocksInProc()
             Assert.False(timeout.IsCancellationRequested);
         }
 
-        private async Task WaitForLockToEngage(SyncdRunResult result)
```
aortiz-msft, line 152: Are these methods not used anywhere? The removal of unused methods should be noted in the PR description.

aortiz-msft (changes requested):
:clock1:

src/NuGet.Core/NuGet.Common/Logging/NullLogger.cs
```diff
@@ -8,7 +8,7 @@ namespace NuGet.Common
 {
     public class NullLogger : LoggerBase
     {
-        private static ILogger _instance;
+        private static ILogger? _instance;
 
         public static ILogger Instance
```
nkolev92, line 13: tbh this optimization seems pointless

src/NuGet.Core/NuGet.Common/Telemetry/TelemetryEvent.cs
```diff
@@ -8,36 +8,36 @@ namespace NuGet.Common
...
-        private IDictionary<string, object> _properties;
-        private IDictionary<string, object> _piiProperties;
+        private IDictionary<string, object?> _properties;
+        private IDictionary<string, object?> _piiProperties;
 
         /// <summary> Creates a new instance of <see cref="TelemetryEvent"/>. </summary>
         /// <param name="eventName"> Event name. </param>
-        public TelemetryEvent(string eventName) :
-            this(eventName, new Dictionary<string, object>())
+        public TelemetryEvent(string? eventName) :
```
nkolev92: you mentioned some places create `null` events. Can that be changed so we can have a more realistic annotation this type?

src/NuGet.Core/NuGet.Common/Telemetry/TelemetryEvent.cs
```diff
@@ -8,36 +8,36 @@ namespace NuGet.Common
...
+            this(eventName, new Dictionary<string, object?>())
         {
         }
 
         /// <summary> Creates a new instance of <see cref="TelemetryEvent"/>. </summary>
         /// <param name="eventName"> Event name. </param>
         /// <param name="properties"> Properties to add to the event. </param>
-        public TelemetryEvent(string eventName, Dictionary<string, object> properties)
+        public TelemetryEvent(string? eventName, Dictionary<string, object?> properties)
```
nkolev92: Why is the value nullable?

src/NuGet.Core/NuGet.Common/Telemetry/TelemetryEvent.cs
```diff
@@ -55,26 +55,26 @@ public TelemetryEvent(string eventName, Dictionary<string, object> properties)
         }
 
         /// <summary> Complex data properties. </summary>
-        public IDictionary<string, object> ComplexData { get; } = new Dictionary<string, object>();
+        public IDictionary<string, object?> ComplexData { get; } = new Dictionary<string, object?>();
```
nkolev92, line 59: same here.

nkolev92 (commented):
Looks pretty good. Nicely done. 
TelemetryEvent is really the only type where I wish we can do better.

src/NuGet.Core/NuGet.Common/Logging/NullLogger.cs
zivkan: What optimization? I was annotating exactly what was implemented

src/NuGet.Core/NuGet.Common/PathUtil/FileUtility.cs
zivkan: I don't expect any customers to ever see this message, so no, I don't think it needs to be localized. It's only feasible if someone using `NuGet.Common` (part of the Client SDK) as part of their app calls the API with bad input (or if someone modifying NuGet starts calling this in a way that is buggy). Frankly, I think it's a bad API, NuGet shouldn't be in the business of proving and supporting trivial file utility APIs.

I also think it's a bad idea to parse exception messages. .NET Core changed some exception messages from what they were on the .NET Framework (I think ArgumentNullException's message is an example), and exception messages that are localized will have unexpected results when running on machines not using the same language as the developer who wrote code trying to parse the exception message. Even for tests this is a bad idea, since I don't want to exclude people running their devbox in something other than `en-US` from being able to run tests locally.

test/NuGet.Core.Tests/NuGet.Common.Test/SynchronizationTest.cs
zivkan: no, they're not used, which saved me from spending time figuring out how to annotate them and fix compiler errors, because from memory they were non-trivial to annotate. I was very happy when I noticed VS tell me 0 references!

src/NuGet.Core/NuGet.Common/Telemetry/TelemetryEvent.cs
zivkan: Say we have telemetry saying `telemetryEvent["SomeMSBuildProperty"] = value` to tell us what value customers are using for some feature. If the project doesn't define a value for that MSBuild property, it makes sense for the property to be null, and it also makes sense for us to want telemetry telling us about the null value, rather than omitting the telemetry property completely.

src/NuGet.Core/NuGet.Common/PathUtil/PathUtility.cs
zivkan: In all these PRs when I'm enabling nullable checks and annotating all the APIs, I wanted to minimize changes, firstly so the PR doesn't take an entire week to implement if I have to carefully consider every modified method & variable name, but also to minimize risk of introducing regressions. Especially when being threatened [multiple](https://github.com/NuGet/NuGet.Client/pull/5155#discussion_r1182635568) [times](https://github.com/NuGet/Home/issues/12618#issuecomment-1570626749) to revert the entire PR in case of bug, without investigating whether the introduced bug is trivial to fix without reverting. Although, sometimes adding the nullable checks uncovers bugs so then there's a question on fixing the bug vs suppressing the nullable checks, but now I'm talking theoretical, not about this specific comment.

Anyway, I'm all for applying [the "boy scout rule"](https://accesto.com/blog/Boy-scout-rule-in-6-examples-the-basic-principle-of-web-development/), but since enabling nullable involves annotating literally hundreds of APIs, I'm not sure how to apply the boy scout rule while also being pragmatic about scope creep.

Anyway, I think this method, `CheckCaseSenstivityRecursivelyTillDirectoryExists` was simple enough, risk is low enough, so I split it into two methods, so it should be easier to understand now.

src/NuGet.Core/NuGet.Common/Telemetry/TelemetryEvent.cs
zivkan: updated

src/NuGet.Core/NuGet.Common/Logging/NullLogger.cs
nkolev92: the lazy creation of the null logger. It's not really heavy so no need to delay it.
My bad, I should've pointed out that it was the existing code. 

Up to you whether you change it or not.

zivkan: ah, since I had to annotate hundreds of APIs, I didn't pay attention to most of what I was annotating. Trying to resolve all the compiler errors, and minimize risk of regression by avoiding unnecessary code changes.

nkolev92 (approved):

src/NuGet.Core/NuGet.Common/PathUtil/PathUtility.cs
```diff
@@ -442,43 +460,39 @@ private static bool CheckIfFileSystemIsCaseInsensitive()
...
                 foreach (var path in listOfPathsToCheck)
                 {
-                    bool ignore;
-                    var result = CheckCaseSenstivityRecursivelyTillDirectoryExists(path, out ignore);
-                    if (!ignore)
+                    string? firstParentDirectory = GetFirstParentDirectoryThatExists(path);
+                    if (firstParentDirectory is not null)
                     {
-                        isCaseInsensitive &= result;
+                        isCaseInsensitive &= CheckIfFileSystemIsCaseInsensitive(firstParentDirectory);
```
aortiz-msft, line 466: >isCaseInsensitive &= CheckIfFileSystemIsCaseInsensitive(firstParentDirectory); [](http://example.com/codeflow?start=24&length=78)

Just for my own learning, why wouldn't it be sufficient to check any one of these paths? I.e. why not return immediately?

src/NuGet.Core/NuGet.Common/PathUtil/FileUtility.cs
```diff
@@ -92,7 +92,12 @@ public static void Replace(Action<string> writeSourceFile, string destFilePath)
                 throw new ArgumentNullException(nameof(destFilePath));
             }
 
-            var tempPath = GetTempFilePath(Path.GetDirectoryName(destFilePath));
+            string? destFileDirectory = Path.GetDirectoryName(destFilePath);
+            if (destFileDirectory is null)
```
aortiz-msft, line 96: Should we also raise an exception if the destFileDirectory is an Empty string? [https://learn.microsoft.com/en-us/dotnet/api/system.io.path.getdirectoryname?view=net-7.0](https://learn.microsoft.com/en-us/dotnet/api/system.io.path.getdirectoryname?view=net-7.0)

src/NuGet.Core/NuGet.Common/PathUtil/FileUtility.cs
aortiz-msft: I'm okay with the answer. Suggestion for message: "Not a valid file path".

src/NuGet.Core/NuGet.Common/PathUtil/PathUtility.cs
aortiz-msft: Thank you for splitting the method. It's much easier to read now.

aortiz-msft (approved):
:shipit:

src/NuGet.Core/NuGet.Common/PathUtil/PathUtility.cs
zivkan: I don't know.

Technically, the whole thing is flawed, because individual filesystems can be case sensitive or not. On Windows I can configure a specific partition to be case sensitive, even though Windows' default is case insensitive filesystems. I expect there are filesystems on Linux where you can enable insensitivity.  But I imagine that the number of people who change the defaults is very close to zero, and NuGet wouldn't be the only software that breaks when those defaults are changed.

src/NuGet.Core/NuGet.Common/PathUtil/FileUtility.cs
zivkan: It looks like an exception will be thrown either way (either by us, or by the BCL). The difference is that the `ArgumentException`'s `ParamName` value will be a different name, a name that doesn't exist in our API. But given our code is open source and this API only has a single path parameter, it's trivial to discover which parameter had an input error.

So, it's still the caller's responsibility to pass in sensible input. It doesn't cause the API to produce incorrect results. Since pushing changes resets approvals, unfortunately I don't feel motivated to add such a minor change (same goes for reverting my "see if this fixes E2E tests" commit)

aortiz-msft: you could do a follow up PR :)

PR #3495, opened 2020-07-02
Add package source transaction object to use when saving sources

sbanni:
## Bug

Fixes: https://github.com/NuGet/Home/issues/9711
Regression: No  

## Fix

Details: Create a transaction object that we can use to add additional capabilities for use when saving package sources 

## Testing/Validation

Tests Added: No
Reason for not adding tests:  
Validation:  Tested local scenario and VSOe scenario

PR #5334, opened 2023-07-26
Use Package Source name as key when updating nuget.config, to avoid duplicates

### LINKED ISSUE #10015: 16.10: remove package source 1.0 service. remove obsolete APIs (in nuget.configuration that we added in 16.8)
Per @zivkan comment in PR: https://github.com/NuGet/NuGet.Client/pull/3649/files/e70389d28297073fc21d7d03a28c27143c66f290#r485233215

Consider approach for codespaces and these apis:
NuGet.Configuration.IPackageSourceProvider2
NuGet.Configuration.IPackageSourceProvider2.SavePackageSources(System.Collections.Generic.IEnumerable<NuGet.Configuration.PackageSource> sources, NuGet.Configuration.PackageSourceUpdateOptions packageSourceUpdateOptions) -> void
NuGet.Configuration.PackageSourceProvider.SavePackageSources(System.Collections.Generic.IEnumerable<NuGet.Configuration.PackageSource> sources, NuGet.Configuration.PackageSourceUpdateOptions sourceUpdateSettings) -> void
NuGet.Configuration.PackageSourceUpdateOptions
NuGet.Configuration.PackageSourceUpdateOptions.PackageSourceUpdateOptions(bool updateCredentials, bool updateEnabled) -> void
NuGet.Configuration.PackageSourceUpdateOptions.UpdateCredentials.get -> bool
NuGet.Configuration.PackageSourceUpdateOptions.UpdateEnabled.get -> bool
static readonly NuGet.Configuration.PackageSourceUpdateOptions.Default -> NuGet.Configuration.PackageSourceUpdateOptions

### LINKED ISSUE #12776: Disabling nuget.org in VS options creates a second nuget.org in nuget.config
### NuGet Product Used

### Product Version

17.8 preview 1

### Worked before?

_No response_

### Impact

### Repro Steps & Context

From an internal email:

1.	Launch VS 17.8 preview 1 and Simulate offline in VM by uncheck all the package sources and click “OK” in VS.
 
    image

2.	After step1,  extra feed will be added in NuGet.Config file.

    image

### Verbose Logs

_No response_

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12776
Fixes: https://github.com/NuGet/Home/issues/10015

Regression? no

## Description

When updating nuget.config's package sources (for example, to disable package one), the "name" should be a "key" used to match two SourceItems. Same name, then update the source, whether it's a change in `value` or `protocolVersion`.

Changed some tests, because in my opinion the old behaviour didn't make any sense.

Deleted `IPackageSourceProvider2` and related code, because it was a codespaces change that 1. didn't work and 2. the people working on it marked it as obsolete rather than deleting it, and then never got around to deleting it as they had planned. This seems like a good a time as any to delete it, since it's impacting this PR's other changes.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


martinrrm (dismissed):
LGTM. I agree that old behavior is weird. `Name` has the `key` attribute in XML element, using `protocolVersion` as a second key feels wrong.

jebriede (dismissed):
This is a good change since the protocol version shouldn't make the entry unique from one that doesn't have that attribute specified. @zivkan: What happens when a package source has a protocolVersion="3" and a user opens the Tools -> Options -> Package Source page, then just clicks OK? With this change, it should no longer be adding a second entry in the nuget.config, but will it update the package source item and remove the protocolVersion attribute? Or will it leave that attribute alone? It would be good to know what is happening in that scenario to see if it's expected behavior or if we have to make changes there as well.

jebriede:
> This is a good change since the protocol version shouldn't make the entry unique from one that doesn't have that attribute specified. @zivkan: What happens when a package source has a protocolVersion="3" and a user opens the Tools -> Options -> Package Source page, then just clicks OK? With this change, it should no longer be adding a second entry in the nuget.config, but will it update the package source item and remove the protocolVersion attribute? Or will it leave that attribute alone? It would be good to know what is happening in that scenario to see if it's expected behavior or if we have to make changes there as well.

I just tested this branch with a nuget.config that has just nuget.org with protocolVersion="3" and if I make an unrelated change in the Package Source Options dialog, such as adding another package source and click OK, it removes protocolVersion="3" from the nuget.org entry in the nuget.config. This feels wrong to me. If we aren’t supporting the PackageVersion property in the Package Sources UI, we should probably not be removing it either then unless this was a conscious design decision we made at some point.

aortiz-msft (changes requested):
> I just tested this branch with a nuget.config that has just nuget.org with protocolVersion="3" and if I make an unrelated change in the Package Source Options dialog, such as adding another package source and click OK, it removes protocolVersion="3" from the nuget.org entry in the nuget.config. This feels wrong to me. If we aren’t supporting the PackageVersion property in the Package Sources UI, we should probably not be removing it either then unless this was a conscious design decision we made at some point.

Looks like we need a CompareTo method for PackageSource so we can safely ignore the ProtocolVersion in this case as part of the comparison.

zivkan:
Unfortunately, a new `CompareTo` method isn't going to solve all problems. The bug the PR fixes (a duplicate source being added to the config) is more or less the same code that causes `protocolVersion` to be removed with this PR's current fix.

The package source options page ultimately calls into [`PackageSourceProvier.SavePackageSources`](https://github.com/NuGet/NuGet.Client/blob/d7c80b2e4c2c05b7819c7ae46c2133e9a181f9c9/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs#L659) with a list of all package sources. The options page has stripped the `protocolVersion` out of the package source, so from NuGet.Configuration's point of view, it's doing the correct thing removing the `protocolVersion` from source, since that's presisely what the VS code is asking it to do.

To fix this, we'd have to at the very minimum make the VS options page remember the `protocolVersion` from package sources loaded from config. A better solution would be to add `protocolVersion` to the VS options page, but that's a much, much bigger change. Which should be done eventually, but do we want to migrate the page to WPF before or after making such a change?

I'll investigate having options just remember the previous `protocolVersion`. Hopefully it's not a major change, but this is one of the older parts of our code (after all it's still using WinForms) that hasn't had as much refactoring/modernization over the years.

jeffkl (dismissed):

src/NuGet.Core/NuGet.Configuration/PublicAPI.Shipped.txt
```diff
@@ -327,8 +327,8 @@ const NuGet.Configuration.PackageSource.DefaultProtocolVersion = 2 -> int
 ~override NuGet.Configuration.AddItem.Clone() -> NuGet.Configuration.SettingBase
 ~override NuGet.Configuration.AddItem.DisallowedValues.get -> System.Collections.Generic.IReadOnlyDictionary<string, System.Collections.Generic.IReadOnlyCollection<string>>
 ~override NuGet.Configuration.AddItem.ElementName.get -> string
-~override NuGet.Configuration.AddItem.Equals(object other) -> bool
-override NuGet.Configuration.AddItem.GetHashCode() -> int
+~override sealed NuGet.Configuration.AddItem.Equals(object other) -> bool
+override sealed NuGet.Configuration.AddItem.GetHashCode() -> int
```
zivkan, line 324: These are breaking changes in the Client SDK, with comments in `AddItem.cs` explaining why it's important. The bug prevented by making these methods `sealed` is subtle and these methods need to be implemented in a specific way that goes against normal coding conventions, hence the risk of defects is high (especially when the person implementing tests validate buggy behaviour 🤦)

src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/ContextInfos/PackageSourceContextInfo.cs
```diff
@@ -22,20 +23,31 @@ public PackageSourceContextInfo(string source, string name)
         }
 
         public PackageSourceContextInfo(string source, string name, bool isEnabled)
+            : this(source, name, isEnabled, PackageSource.DefaultProtocolVersion)
```
donnie-msft, line 26: is this whitespace correct?

donnie-msft (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetSourcesService.cs
```diff
@@ -21,49 +21,53 @@ public sealed class NuGetSourcesService : INuGetSourcesService
...
+            Assumes.NotNull(packageSourceProvider);
 
             _options = options;
             _serviceBroker = serviceBroker;
             _authorizationServiceClient = authorizationServiceClient;
-            _sharedServiceState = state;
-            _sharedServiceState.SourceRepositoryProvider.PackageSourceProvider.PackageSourcesChanged += PackageSourceProvider_PackageSourcesChanged;
+            _packageSourceProvider = packageSourceProvider;
+#pragma warning disable CS0618 // Type or member is obsolete
+            _packageSourceProvider2 = packageSourceProvider as IPackageSourceProvider2;
```
nkolev92: You must have accidentally removed the change listener: PackageSourcesChanged.

nkolev92: Can we always use `IPackageSourceProvider2`. It's always available within VS no?

nkolev92 (commented):
Looks reasonable. There's 1 bug I see, but there's also some compile errors rn.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetSourcesService.cs
zivkan: Your question movivated me to finally implement https://github.com/NuGet/Home/issues/10015 / https://github.com/NuGet/Home/issues/10098

src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/ContextInfos/PackageSourceContextInfo.cs
zivkan: yes, no problem reported in VS or in the static analysis CI job (about this file, there was an error reported in another file in a previous CI run)

aortiz-msft: Looks correct to me.

src/NuGet.Core/NuGet.Configuration/Settings/Items/SourceItem.cs
```diff
@@ -62,23 +46,5 @@ public override SettingBase Clone()
...
-            }
-
-            if (ReferenceEquals(this, source))
-            {
-                return true;
-            }
-
-            return string.Equals(Key, source.Key, StringComparison.Ordinal) &&
-                string.Equals(ProtocolVersion, source.ProtocolVersion, StringComparison.Ordinal);
-        }
```
heng-liu, line 82: There are a few [tests ](https://github.com/NuGet/NuGet.Client/blob/dev/test/NuGet.Core.Tests/NuGet.Configuration.Test/SettingsFileParsingTests/SourceItemTests.cs#L45) in SourceItemTests using [DeepEquals](https://github.com/NuGet/NuGet.Client/blob/dev/test/NuGet.Core.Tests/NuGet.Configuration.Test/SettingsTestUtils.cs#L81). Seems this method is assumed to compare if two sourceItems are exactly the same, including key, value, protocolVersion, allowInsecureConnects.
This DeepEquals will call into this Equals method.
With this change, the tests will not be able to compare every attributes.

Since I'm working on adding `allowInsecureConnects` to SourceItem, I'll change tests to validate the other properties when using DeepEquals.

zivkan: The code deleted only compares `Key` (source name) and `ProtocolVersion`, it never compared `Value` (source URL). So, `DeepEquals` would have issues anyway if it depends on `SourceItems.Equals`. In fact, the same is true for `AddItem`, so anything that uses it, like `<fallbackFolders>` and `<disabledPackageSources>`, has the same risk.

nkolev92 (approved):

jeffkl (approved):

donnie-msft (approved):

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
```diff
@@ -709,8 +705,8 @@ public void SavePackageSources(IEnumerable<PackageSource> sources, PackageSource
                         oldPackageSource,
                         existingDisabledSourceItem,
                         existingCredentialsItem,
-                        updateEnabled: sourceUpdateSettings.UpdateEnabled,
-                        updateCredentials: sourceUpdateSettings.UpdateCredentials,
+                        updateEnabled: true,
+                        updateCredentials: true,
                         shouldSkipSave: true,
                         isDirty: ref isDirty);
```
aortiz-msft, line 711: Should the ProtocolVersion be updated as well?

nkolev92: That's part of the PackageSource object on line 708.

zivkan: The method that these lines of code are part of (call site, not call target), `SavePackageSources(IEnumerable<PackageSource> sources)`, looks at each package source in the incoming list, either finds the match in the existing `nuget.config` and then call `UpdatePackageSource`, or if no match is found, call `AddPackageSource`. Finally, `SavePackageSources` will remove any sources that were in the existing nuget.config, but were not matched in the incoming `sources` list.

The lines this comment was made on are the parameters to that `UpdatePackageSource`, and the first two parameters `source` and `oldPackageSource`, are of type `PackageSource`, which has the property `ProtocolVersion`, as Nikolche pointed out.

The diff that this comment is pointing out (removing `sourceUpdateSettings` and hardcoding `true` for `updateEnabled` and `updateCredentials` is related to bundling a fix for https://github.com/NuGet/Home/issues/10015 in the same PR, to avoid confusion like https://github.com/NuGet/NuGet.Client/pull/5334#discussion_r1281096639. I didn't use git's revert feature, but it should effectively be a revert of https://github.com/NuGet/NuGet.Client/pull/3495, since the goals it tried to achieve didn't work, and actually introduced a new mode of failure (well, https://github.com/NuGet/NuGet.Client/commit/b5b44526dea0379ebd6c8e51e8a041c06d5845ca#diff-db51a3ac45fc35243298d74958a007f9073a8b044e3510325cef068b810a6d9eR101 and https://github.com/NuGet/NuGet.Client/commit/b5b44526dea0379ebd6c8e51e8a041c06d5845ca#diff-2cde950543b0ac9b52f8a0b7c8e358a3fb7804f670a12e077827c6dfcd814459R33 introduced the mode of failure, and https://github.com/NuGet/NuGet.Client/commit/86c78b646f39fbebe1349a45deb59bc1e1c60fd9 introduced the regression that this PR is fixing), which I had to fix in this PR in `NuGetSourcesService`.

aortiz-msft (approved):

PR #5302, opened 2023-07-08
Introduce a --exclude-project option to list command

Greybird:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12585

Regression? Last working version:

## Description
`list` command behaves differently depending on the presence of the `--deprecated`, `--outdated`, or `--vulnerable` flag:
* if any of these flags is present, projects are filtered out of transitive dependencies,
* if not, projects are kept in the transitive dependencies.

As 
* filtering the projects out of the list of transitive dependencies is only dependent on the type of reports requested (see [ListPackageCommandRunner](https://github.com/NuGet/NuGet.Client/blob/d9d5a38c6e0c3ccc36b1b21cb8fa468474690080/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/ListPackageCommandRunner.cs#L118C157-L118C222)),
* the output (console, json, or other) does not display the type of dependency (it is lost during the model construction in [GetResolvedVersion](https://github.com/NuGet/NuGet.Client/blob/d9d5a38c6e0c3ccc36b1b21cb8fa468474690080/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs#L658C49-L658C68), not being included in [FrameworkPackages](https://github.com/NuGet/NuGet.Client/blob/d9d5a38c6e0c3ccc36b1b21cb8fa468474690080/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/FrameworkPackages.cs#L13-L45)/[InstalledPackageReference](https://github.com/NuGet/NuGet.Client/blob/d9d5a38c6e0c3ccc36b1b21cb8fa468474690080/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/InstalledPackageReference.cs#L17-L33))
there is in practice no way to request a complete list of _actual_ nuget packages.

I have no context on why the `list` command reports projects as transitive dependencies, but it is a fairly common need to only list nuget packages, to automate some process related to them. I tend to think that this is also a very common expectation for the command. The inability to filter projects adds noise to the output in these cases.

This PR adds a `--exclude-project` flag, allowing to override the behavior of the command when no `--deprecated`, `--outdated`, or `--vulnerable` flag is passed (a warning is reported if any of these flags is passed with `--exclude-project`), to actually alter the default project filtering behavior.

[...]

PR #5335, opened 2023-07-26
Remove projects from list package output

### LINKED ISSUE #12585: dotnet list package --format json --include-transitive includes project references as transitive packages.
### NuGet Product Used

dotnet.exe

### Product Version

7.0.203

### Worked before?

no

### Impact

### Repro Steps & Context

If I have a project that references another project, then the beforementioned command states that referenced project is a transitive package dependency for referencing project.

I'm not sure if this is really bug, perhaps it's just a need of some differentiation between project and package in output, for example, additional "type" property in "transitivePackages" node item, saying if it's a "Project" reference or a "Package".

Steps to reproduce:
1. Download the archive from attachment and unpack it somewhere
2. Run dotnet restore in unpacked folder
3. Run dotnet list package --format json --include-transitive

The output says that "DotnetListJsonTransitivesIssue" project has "ReferencedProject" in its transitive packages (or at least doesn't provide an information if this is really a package from some nuget source or just a project). 

Also, if it should include projects in output, in this very case the "ReferencedProject" should be within the "topLevelPackages" node

[DotnetListJsonTransitivesIssue.zip](https://github.com/NuGet/Home/files/11465827/DotnetListJsonTransitivesIssue.zip)

### Verbose Logs

```shell
{
  "version": 1,
  "parameters": "--include-transitive",
  "projects": [
    {
      "path": "C:/Users/User/source/repos/exp/DotnetListJsonTransitivesIssue/DotnetListJsonTransitivesIssue/DotnetListJsonTransitivesIssue.csproj",
      "frameworks": [
        {
          "framework": "net7.0",
          "transitivePackages": [
            {
              "id": "ReferencedProject",
              "resolvedVersion": "1.0.0"
            }
          ]
        }
      ]
    },
    {
      "path": "C:/Users/User/source/repos/exp/DotnetListJsonTransitivesIssue/ReferencedProject/ReferencedProject.csproj",
      "frameworks": [
        {
          "framework": "net7.0"
        }
      ]
    }
  ]
}
```

Greybird:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12585
Other PR: https://github.com/NuGet/NuGet.Client/pull/5302
Design spec: https://github.com/NuGet/Home/pull/12735

## Description
`list` command, when used with the `--include-transitive` flag, but without any of the `--deprecated`, `--outdated` or `--vulnerable` ones, reports directly referenced projects as transitive dependencies.
This PR removes this behavior, and adds a test to ensure that with or without `--include-transitive` flag, a project is never returned by the `list` command

## PR Checklist

- [X] PR has a linked issue.
- [X] Described changes

- **Tests**
  - [X] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [X] N/A


zivkan (dismissed):

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
```diff
@@ -652,11 +652,10 @@ internal static bool IsPackageReferenceProject(Project project)
         /// <param name="userInputFrameworks">A list of frameworks</param>
         /// <param name="assetsFile">Assets file for all targets and libraries</param>
         /// <param name="transitive">Include transitive packages/projects in the result</param>
-        /// <param name="includeProjects">Include project references in top-level and transitive package lists</param>
```
nkolev92, line 655: So in some places, we seem to want to include projects only in the default `dotnet list package`, but in other places, we seem to filter them out. 

The spec did not clearly call out the whole `dotnet list package` lists projects, but `--outdated`, `--vulnerable` or `--deprecated` does not. 

But maybe there is something to be discussed here?

zivkan: Sorry, I don't understand why we ever want to include projects in `dotnet list package`. There's already a `dotnet list reference` command to show project references.

nkolev92: 🤷 The feature design does not explicitly call out either one. The implementation tries to do both. 

I'm ok with listing just the packages, given the list reference command is available, but it just made me wonder whether anyone might truly depend on this.

Greybird:
I see that the checks are having test failures, but they are passing locally, and I seem to have no way to get access to the results. If an action is needed from me, please ping me!

zivkan:
The "Apex Tests on Windows" are just flakey, ignore those.

For Windows func tests, the test `Dotnet.Integration.Test.DotnetListPackageTests.DotnetListPackage_ProjectReference_Succeeds` failed with error message:

```text
Assert.Contains() Failure
Not found: ProjectB
In value:  Project 'ProjectA' has the following package references
[net46]: No packages were found for this framework.
```

[The test only runs on Windows](https://github.com/NuGet/NuGet.Client/blob/a39baac1e0fc3126a767b7261beb3804a28e4a97/test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetListPackageTests.cs#L289-L294), so if you're using a different OS, then running test project will skip that test, although it's not clear to me why this test is Windows only. It looks extremely generic and should work the same on Linux & Mac.

But [the test also looks fundamentally flawed to me](https://github.com/NuGet/NuGet.Client/blob/a39baac1e0fc3126a767b7261beb3804a28e4a97/test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetListPackageTests.cs#L325). `git blame` says I wrote the test! I added it when I fixed a bug, but now I believe that it was just an easy way to hit the bug (use `--outdated` on a package that doesn't have any known versions on any package source). Though I was newish on the team at the time, so I'm not sure I would have questioned the weird behaviour at the time. Anyway, IMO the test should be changed to validate that `dotnet list package --show-transitive` does NOT show project references. This will ensure that nobody re-introduces this behaviour accidentally in the future.

Greybird:
@zivkan Thanks for the info!
As I had added a `DotnetListPackage_DoesNotReturnProjects` test to validate no projects were returned, I chose to remove `DotnetListPackage_ProjectReference_Succeeds `, and to add tests cases to mine to further validate the situation for no flags/`--include-transitive` for all (currently supported) report types.

zivkan (approved):

nkolev92 (approved):

martinrrm (approved):

PR #5339, opened 2023-08-02
Make PackageSourceMapping API nullable enabled and improve test coverage

### LINKED ISSUE #12794: `PackageSourceMapping` API doesn't follow best practices for returning lists
Ideally, return empty lists, not `null` lists which must be checked for `null` & empty by callers.
There's also missing test coverage for `PackageDownloader`.

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12794

## Description

_Predecessor to https://github.com/NuGet/NuGet.Client/pull/5295_

- Enable nullable compiler checks, and make changes to callers as needed.
- Add missing test coverage for `PackageDownloader`

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMapping.cs
+++ b/src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMapping.cs
@@ -22,7 +22,7 @@ namespace NuGet.Configuration
         /// </summary>
         internal IReadOnlyDictionary<string, IReadOnlyList<string>> Patterns { get; }
 
-        private Lazy<SearchTree?> SearchTree { get; }
+        private Lazy<SearchTree>? SearchTree { get; }
 
         /// <summary>
         /// Indicate if any packageSource exist in package source mapping section
@@ -37,7 +37,7 @@ namespace NuGet.Configuration
         /// <exception cref="ArgumentException"> if <paramref name="packageId"/> is null, empty, or whitespace only.</exception>
         public IReadOnlyList<string> GetConfiguredPackageSources(string packageId)
         {
-            if (SearchTree.Value is null)
+            if (SearchTree?.Value is null)
             {
                 return Array.Empty<string>();
             }
@@ -47,14 +47,14 @@ namespace NuGet.Configuration
 
         public string? SearchForPattern(string packageId)
         {
-            return SearchTree.Value?.SearchForPattern(packageId);
+            return SearchTree?.Value.SearchForPattern(packageId);
         }
 
         public PackageSourceMapping(IReadOnlyDictionary<string, IReadOnlyList<string>> patterns)
         {
             Patterns = patterns ?? throw new ArgumentNullException(nameof(patterns));
             IsEnabled = Patterns.Count > 0;
-            SearchTree = new Lazy<SearchTree?>(() => GetSearchTree());
+            SearchTree = IsEnabled ? new Lazy<SearchTree>(() => GetSearchTree()) : null;
         }
 
         /// <summary>
@@ -82,16 +82,9 @@ namespace NuGet.Configuration
             return new PackageSourceMapping(patterns);
         }
 
-        private SearchTree? GetSearchTree()
+        private SearchTree GetSearchTree()
         {
-            SearchTree? patternsLookup = null;
-
-            if (IsEnabled)
-            {
-                patternsLookup = new SearchTree(this);
-            }
-
-            return patternsLookup;
+            return new SearchTree(this);
         }
     }
 }
--- a/src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMapping.cs
+++ b/src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMapping.cs
@@ -1,6 +1,8 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+#nullable enable
+
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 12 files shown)


src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMapping.cs
```diff
@@ -20,25 +22,30 @@ public class PackageSourceMapping
         /// </summary>
         internal IReadOnlyDictionary<string, IReadOnlyList<string>> Patterns { get; }
 
-        private Lazy<SearchTree> SearchTree { get; }
+        private Lazy<SearchTree?> SearchTree { get; }
```
nkolev92: Would it be more natural to have the lazy be null altogether? This way you have an extra object even if PSM is not enabled (majority of the scenarios right now).

donnie-msft: Have a look at https://github.com/NuGet/NuGet.Client/pull/5339/commits/24fb4a63b6bd924de8ec6ab7548886846fd62acb

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMapping.cs
```diff
@@ -77,14 +84,7 @@ public static PackageSourceMapping GetPackageSourceMapping(ISettings settings)
...
         {
-            SearchTree patternsLookup = null;
-
-            if (IsEnabled)
-            {
-                patternsLookup = new SearchTree(this);
-            }
-
-            return patternsLookup;
+            return new SearchTree(this);
```
nkolev92, line 87: nit: now that this method is only 1 line, it can probably be inlined.

nkolev92 (approved):

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMapping.cs
donnie-msft: I thought about that but believe the callstacks would be easier to look at this way.

nkolev92: The stack would just show the `SearchTree` constructor anyways, as that's where the work is actually being done.

donnie-msft: True... well, in future PR I might remove it if I don't extract it into an interface to make testing easier.

aortiz-msft (commented):
LGTM. (PR was merged while I was reviewing.)

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMapping.cs
```diff
@@ -20,34 +22,39 @@ public class PackageSourceMapping
         /// </summary>
         internal IReadOnlyDictionary<string, IReadOnlyList<string>> Patterns { get; }
 
-        private Lazy<SearchTree> SearchTree { get; }
+        private Lazy<SearchTree>? SearchTree { get; }
```
zivkan, line 25: Having a lazy that is also nullable feels weird. Although looking below, I see it's because this value will be null when `IsEnabled` is false. Maybe worth a comment?

zivkan, line 25: Also note that `SearchTree` does not meet our naming style. Private, non-static fields should be `_camelCased`, not `PascalCased`.  Isn't VS putting a squiggle under the name?

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMapping.cs
```diff
@@ -20,34 +22,39 @@ public class PackageSourceMapping
...
 
         /// <summary>
         /// Indicate if any packageSource exist in package source mapping section
         /// </summary>
         public bool IsEnabled { get; }
```
zivkan, line 30: I can't remember the exact name of the attribute, but there's an attribute something like `NotNullMemberWhen`, which tells the nullable checkers that when this property is `true`, then the referenced member is not null, so you can write code like `if (IsEnabled) { SearchTree.Value.Method(); }` and it won't force you to check the nullability of `SearchTree`. Although, `SearchTree` is private, and it seems that we don't have code here that checks `IsEnabled`, so maybe it's more of an idealistic suggestion than a practical one.

zivkan (commented):
I didn't have time to review yesterday because of CI failures I was trying to investigate in my own PRs, so this PR was merged before I had a look to comment 😞

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMapping.cs
donnie-msft: I agree it feels weird, but after I received feedback (https://github.com/NuGet/NuGet.Client/pull/5339#discussion_r1282180240), I saw the perspective of just letting it be `null` when we *know* there's no `SearchTree` due to `IsEnabled` being `false`. 

I like that this class is immutable, meaning that I can make assertions like this Lazy can be declared once in the constructor, and then there's no concern over state changes leading it to become stale.

donnie-msft:
> I didn't have time to review yesterday because of CI failures I was trying to investigate in my own PRs, so this PR was merged before I had a look to comment 😞

Could you let PR authors know if you're wanting to review something?

src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMapping.cs
donnie-msft: > Also note that SearchTree does not meet our naming style. Private, non-static fields should be _camelCased, not PascalCased. 

According to our coding guideline # 3, this only applies to fields. `SearchTree` is a private property. I'm not sure why the creator of the property made it a property, but nothing has changed in this PR.

> Isn't VS putting a squiggle under the name?

Nope.
Here's showing how it would do that for a field:

image

PR #5341, opened 2023-08-02
Add unit test cases for PatternExpression.Match Method

Nigusu-Allehu:
## Bug

Fixes: Add unit test cases for the PatternExpression.Match method

## Description

Adds test cases to the Match function inside PatternExpression. It also aims to test the TryMatch function which is a helper function to the Match method. 
## PR Checklist

- [ ] PR has a linked issue.
- [ ] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A


test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,198 @@
...
+using System.Collections.Generic;
+using NuGet.ContentModel;
+using NuGet.ContentModel.Infrastructure;
+using Xunit;
+
+namespace NuGet.Client.Test
+{
+    public class PatternExpressionMatchTests
+    {
+
```
nkolev92, line 14: You have redudant new lines.

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,198 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using NuGet.ContentModel;
+using NuGet.ContentModel.Infrastructure;
+using Xunit;
+
+namespace NuGet.Client.Test
```
nkolev92, line 10: The namespace should make the project name at minimum.

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,198 @@
...
+{
+    public class PatternExpressionMatchTests
+    {
+
+        
+        [Fact]
+        public void LiteralSegment_OnlyShouldMatch()
```
nkolev92, line 17: Consider following the test name guidelines: https://github.com/NuGet/NuGet.Client/blob/dev/docs/coding-guidelines.md#unit-test-method-naming. 

In general, you want the first part of a test method to have the method under test.

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,198 @@
...
+            var pattern = new PatternDefinition("content/file.txt");
+            var expression = new PatternExpression(pattern);
+            var path = "content/file.txt";
+            var result = expression.Match(path, propertyDefinitions);
+            Assert.NotNull(result);
+            Assert.Equal(path, result.Path);
+        }
+
+        [Fact]
+        public void LiteralSegment_OnlyShouldNotMatch()
```
nkolev92, line 33: This method and the previous match in all but 1 argument.

Consider using `Theory` instead of  `Fact` for the test method to combine them.

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,198 @@
...
+            };
+            var pattern = new PatternDefinition("content/file.txt");
+            var expression = new PatternExpression(pattern);
+            var path = "content/file2.txt";
+            var result = expression.Match(path, propertyDefinitions);
+            Assert.Null(result);
+        }
+
+        [Fact]
+        public void LiteralSegment_hasTokenSegmentShouldMatch()
```
nkolev92, line 48: Our coding guidelines call out pascal case for method names, https://github.com/NuGet/NuGet.Client/blob/dev/docs/coding-guidelines.md#unit-test-method-naming rule 3.

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,198 @@
...
+                {"name", new ContentPropertyDefinition("any", parser: (o, t) => o) }
+            };
+            //this pattern has both a token segment and a litteral segment
+            var pattern = new PatternDefinition("content/{name}");
+            var expression = new PatternExpression(pattern);
+            var path = "content/file.txt";
+            var match_result = expression.Match(path, propertyDefinitions);
+            object result;
+            match_result.Properties.TryGetValue("name", out result);
+            Assert.NotNull(result);
```
nkolev92: Assert both not null and value is redundant.

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,198 @@
...
+            };
+            //this pattern has tfm as a token segment and is not match only
+            //As a result tfm_raw property will be saved for the content item token matching with tfm
+            var pattern = new PatternDefinition("contentItem/{tfm}");
+            var expression = new PatternExpression(pattern);
+            var path = "contentItem/this/is/tfm/raw/file.txt";
+            var expected_result = "this/is/tfm/raw/file.txt";
+            var match_result = expression.Match(path, propertyDefinitions);
+            Assert.NotNull(match_result);
+            object result;
```
nkolev92, line 119: you can inline it. 

```cs
            match_result.Properties.TryGetValue("tfm_raw", out object result);
```

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,198 @@
...
+            match_result.Properties.TryGetValue("tfm_raw", out result);
+            Assert.Equal(expected_result, result);
+        }
+
+        [Fact]
+        public void TokenSegment_multipleTokens()
+        {
+            Dictionary<string, ContentPropertyDefinition> propertyDefinitions =
+                new Dictionary<string, ContentPropertyDefinition>() {
+        {"name", new ContentPropertyDefinition("any", parser: (o, t) => o) },
```
nkolev92, line 129: Formatting. This needs to be further to the right.

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,198 @@
...
+        {"version", new ContentPropertyDefinition("any", parser: (o, t) => o) }
+            };
+            var pattern = new PatternDefinition("contentItem/{name}/{version}");
+            var expression = new PatternExpression(pattern);
+            var path = "contentItem/mypackage/1.0.0";
+            var expected_name = "mypackage";
+            var expected_version = "1.0.0";
+            var match_result = expression.Match(path, propertyDefinitions);
+            Assert.NotNull(match_result);
+            object name_result;
```
nkolev92, line 139: same as the other ones, you can inline the results.

nkolev92: We also use pascal case all through out. `name_result` is snake case.

nkolev92 (commented):
Please follow the PR template. 

It recommends having an issue. 

Please creating an issue in Client/Engineering.

martinrrm:
@Nigusu-Allehu I think you will need to create another branch that follow the pattern `dev-<alias>-<description/feature>`, this is because our CI is configured so that branches with dev-* are picked automatically to run the automated tests

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,198 @@
...
+        {"name", new ContentPropertyDefinition("any", parser: (o, t) => o) }
+            };
+            var pattern = new PatternDefinition("contentItem/{name}/{name}");
+            var expression = new PatternExpression(pattern);
+            var path = "contentItem/mypackage/file.dll";
+            Assert.Throws<ArgumentException>(() => expression.Match(path, propertyDefinitions));
+        }
+
+
+
```
donnie-msft, line 194: nit: extra spaces

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,198 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
```
donnie-msft, line 3: Consider enabling nullable since this is a new test.

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,198 @@
...
+        [Fact]
+        public void LiteralSegment_hasTokenSegmentShouldNotMatch()
+        {
+            // We do not have a parser for the token segment name.
+            //As a result it should not be able to match and return a content item
+            Dictionary<string, ContentPropertyDefinition> propertyDefinitions =
+                new Dictionary<string, ContentPropertyDefinition>() {
+                {"name", new ContentPropertyDefinition("any") }
+            };
+            //this pattern has both a token segment and a litteral segment
```
donnie-msft:
```suggestion
            // this pattern has both a token segment and a literal segment
```

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,198 @@
...
+            var pattern = new PatternDefinition("content/{name}");
+            var expression = new PatternExpression(pattern);
+            var path = "content/file.txt";
+            var match_result = expression.Match(path, propertyDefinitions);
```
donnie-msft: I think for locals, we stick to camel casing. Is that right @nkolev92 ?

```suggestion
            var matchResult = expression.Match(path, propertyDefinitions);
```

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,198 @@
...
+            Assert.NotNull(match_result);
+            object result;
+            match_result.Properties.TryGetValue("tfm_raw", out result);
+            Assert.Equal(expected_result, result);
+        }
+
+        [Fact]
+        public void TokenSegment_multipleTokens()
+        {
+            Dictionary<string, ContentPropertyDefinition> propertyDefinitions =
```
donnie-msft:
```suggestion
            var propertyDefinitions =
```

donnie-msft (commented):
Glad you're working on more tests!
Some of the formatting nitpicks, you might be able to tackle or see easily by running our static analysis command. The CLI runs this on the entire solution:
`dotnet format whitespace --verify-no-changes NuGet.sln`
https://github.com/NuGet/NuGet.Client/blob/dev/eng/pipelines/templates/Static_Source_Analysis.yml#L77

It may be a little more efficient for you to just run it against the project you're working on:
`dotnet format whitespace --verify-no-changes .\test\NuGet.Core.Tests\NuGet.PackageManagement.Test\NuGet.PackageManagement.Test.csproj`

note that this output from the command doesn't mean you introduced a problem (our CI shows this as well)

> Warnings were encountered while loading the workspace. Set the verbosity option to the 'diagnostic' level to log warnings.

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,198 @@
...
+
+        
+        [Fact]
+        public void LiteralSegment_OnlyShouldMatch()
+        {
+            // match literal segments given a pattern
+            Dictionary<string, ContentPropertyDefinition> propertyDefinitions =
+                new Dictionary<string, ContentPropertyDefinition>() {
+                {"type", new ContentPropertyDefinition("text") }
+            };
```
dtivel: https://github.com/NuGet/NuGet.Client/blob/dev/docs/coding-guidelines.md#c-coding-style

1.  We use Allman style braces.
2.  There should be 1 space between a curly brace and whatever follows on the same line.
3.  [Optional] Since this is a `new` expression the data type of the right-hand side (RHS) is obvious.  You could have the explicit type just once (LHS or RHS).

```suggestion
            Dictionary<string, ContentPropertyDefinition> propertyDefinitions = new()
            {
                { "type", new ContentPropertyDefinition("text") }
            };
```

This feedback applies throughout the PR.

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,198 @@
...
+            var pattern = new PatternDefinition("content/file.txt");
+            var expression = new PatternExpression(pattern);
+            var path = "content/file.txt";
```
dtivel: Avoid string duplication.

```suggestion
            var path = "content/file.txt";
            var pattern = new PatternDefinition(path);
            var expression = new PatternExpression(pattern);
```

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,198 @@
...
+            var result = expression.Match(path, propertyDefinitions);
```
dtivel, line 27: Use explicit type instead of `var`.  `ContentItem`?

Check other tests too.

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,198 @@
...
+        [Fact]
+        public void LiteralSegment_hasTokenSegmentShouldMatch()
+        {
+            //We have a parser for the token "name" which takes in any type of string and returns it.
+            //Therefore file.txt should be added to the result content item
+            Dictionary<string, ContentPropertyDefinition> propertyDefinitions =
+                new Dictionary<string, ContentPropertyDefinition>() {
+                {"name", new ContentPropertyDefinition("any", parser: (o, t) => o) }
+            };
+            //this pattern has both a token segment and a litteral segment
```
dtivel, line 56: litteral -> literal

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
dtivel: Can simplify:

```suggestion
            Assert.True(match_result.Properties.TryGetValue("name", out object result));
```

This doesn't test that `result` is non-null, but the next assertion does that.

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,198 @@
...
+            Dictionary<string, ContentPropertyDefinition> propertyDefinitions =
+                new Dictionary<string, ContentPropertyDefinition>() {
+        {"name", new ContentPropertyDefinition("any", parser: (o, t) => o) },
+        {"version", new ContentPropertyDefinition("any", parser: (o, t) => o) }
+            };
+            var pattern = new PatternDefinition("contentItem/{name}/{version}");
+            var expression = new PatternExpression(pattern);
+            var path = "contentItem/mypackage/1.0.0";
+            var expected_name = "mypackage";
+            var expected_version = "1.0.0";
```
dtivel:
```suggestion
            var expectedName = "mypackage";
            var expectedVersion = "1.0.0";
            var path = $"contentItem/{expectedName}/{expectedVersion}";
```

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,198 @@
...
+        {
+            Dictionary<string, ContentPropertyDefinition> propertyDefinitions =
+                new Dictionary<string, ContentPropertyDefinition>()
+                {
+                    // Empty propertyDefinitions, no token definition available
+                };
+            var pattern = new PatternDefinition("contentItem/{name}");
+            var expression = new PatternExpression(pattern);
+            var path = "contentItem/mypackage/file.txt";
+            Assert.Throws<Exception>(() => expression.Match(path, propertyDefinitions));
```
dtivel, line 177: Be careful about asserting that something throws.  In the simple case (this), you aren't verifying anything beyond the throw.  It would be a stronger verification to assert that the thrown exception (e.g.:  type, parameter name, message, etc.) is what you expect.  In this case, [the exception type is `System.Exception`](https://github.com/NuGet/NuGet.Client/blob/23f51111dc157ce393c9dc96fb87450ec2aec676/src/NuGet.Core/NuGet.Packaging/ContentModel/Infrastructure/Parser.cs#L173), but you can still verify the message:

```suggestion
            Exception exception = Assert.Throws<Exception>(() => expression.Match(path, propertyDefinitions));
            Assert.Equal("Unable to find property definition for {name}", exception.Message);
```

Apply this feedback wherever you call `Assert.Throws<T>(...)`

zivkan: I dont' think it's a good idea to assert exception text messages. [It causes barriers to people who use locales other than English from contributing to our repo](https://github.com/NuGet/NuGet.Client/pull/5002#issuecomment-1679707317)

dtivel: I think we agree that the following are desirable:

* eliminating barriers for others to contribute
* improving test quality

In the context of this PR, until proven otherwise, I don't know that the two are mutually exclusive.

In the general case, it wasn't clear to me what the root problems were in the comment you linked.  Can you summarize here?

In most product code, we `string.Format(CultureInfo.CurrentCulture, ...)` to produce a culture-specific formatted message.  If we are following all localization guidance, why wouldn't a test that checks the message succeed in non-en-US locales?  My guess is that we aren't following all localization guidance, either in the product or in the test or both.  For example, the product uses `string.Format(CultureInfo.CurrentCulture, ...)` with a localizable resource string but a test uses a hardcoded en-US string for the expected result.  I would expect that to fail for non-en-US locales.  This is a test bug, and it should be fixed, so users don't have to switch locales to run tests.

There are a couple problems related to this test:

1.  The product code throws `System.Exception`, which is discouraged, but it is what it is.  It would be better if it threw a `ParserException` exception with a property that indicates the problematic token, much like `ArgumentException.ParamName`.  However, changing the exception type is out of scope for this PR and would change the behavior of a [public API](https://github.com/NuGet/NuGet.Client/blob/23f51111dc157ce393c9dc96fb87450ec2aec676/src/NuGet.Core/NuGet.Packaging/PublicAPI/netcoreapp5.0/PublicAPI.Shipped.txt#L67).  (I suspect throwable exceptions are outside the purview of the API analyzer, but I don't really know.)
2.  The thrown exception doesn't use a resource string.  This isn't localizable.

If the test formats the string like the product, would that fail on non-en-US locales?

```C#
Assert.Equal(string.Format(CultureInfo.CurrentCulture, "Unable to find property definition for {{{0}}}", "{name}"), exception.Message);
```

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
zivkan: This is actually part of our coding standards: https://github.com/NuGet/NuGet.Client/blob/4d815bf71aa1ce5fe8d9d9da6573303d7093c2f1/docs/coding-guidelines.md?plain=1#L214

If you saw another test file, even in the same directory, using the bad pattern, please don't copy it. In my opinion, NuGet.Client is full of bad patterns due to having been refactored many times, split into separate repos, then being merged back into one again. It's not a good repo for someone new to the industry to learn from,

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
zivkan: Looking at the first comment more carefully, I disagree with this:

> In the simple case (this), you aren't verifying anything beyond the throw. It would be a stronger verification to assert that the thrown exception (e.g.: type, parameter name, message, etc.) is what you expect.

Automated tests greatest value is catching regressions. Do you believe that changing an error message is a regression? I don't. Therefore asserting a message has a very specific string doesn't help catch logic errors, but it does slow down changes improving error messages, and therefore meets my definition of technical debt (slows us down without providing value).

I've worked on one project that had a LOT of unit tests asserting all the implementation details that the team could think of for most classes in the app. This means that every time you add a new feature and make minor changes to existing code, or fix a bug, there's a bunch of failing tests because a bunch of internal state that is irrelevant to customer experience was being asserted, slowing down team velocity. It's well intended, no doubt, but unfortunately counter-productive. Refactoring was an absolute nightmare as we effectively just deleted the existing tests, refactored the code, then wrote new tests, in which case the tests were useless because hey didn't help catch regressions. I think the most valuable tests validate contracts, not implementation details.

So, I don't agree that we should have stronger verification of error messages. Here, the important thing is that an error is raised when the invalid input is provided. In my opinion the exact text of the error message is not important.

I agree that it would be better if the exception was `ArgumentException`, so we could assert that `.ParamName == "whatever"`. If the current exception message has the param name (and expanding to general cases, not just this exception message, if the error message contains a value that was passed in as a method parameter), then I'm fine with checking that `exception.Message.Should().Contain("paramName")`. But I need convincing that asserting exact exception message text is valuable, rather than tech debt.

> In most product code, we string.Format(CultureInfo.CurrentCulture, ...) to produce a culture-specific formatted message. If we are following all localization guidance, why wouldn't a test that checks the message succeed in non-en-US locales?

If the test is running on a binary that contains localization, then any `StringResource.Whatever` will return the most appropriate string for the current locale, but then a test that validates `string.Equals(actualString, "Expected string in English")` is obviously going to fail for non-English locales.

Additionally, the `CultureInfo` passed to `string.Foramt` only applies culture formatting to numbers and dates. Maybe more, but certainly not string resources obtained from resx files. Side note, PM UI's package published date is the only place in the entire codebase where we actually use culture-aware formatting. I don't know of a single place where we use format to provide culture aware to format numbers (thousands separator, whole numbers from fraction). So passing a `CultureInfo` to `string.Format` proabbly doesn't provide NuGet.Client with any benefit, since the one place we format dates is in XAML, not C# code. But someone enabled a code analysis warning that makes sense in other products, so that's why we have `string.Format(CultureInfo.CurrentCulture,` littered throughout our repo. I didn't argue when it was being added because I need to pick my battles, and it isn't harmful, only adds boilerplate.

Anyway, in case you're not familiar with resx string localization, have a quick look at [one of the `.Designer.cs` files](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Common/Strings.Designer.cs). The class has [a private static CultureInfo](https://github.com/NuGet/NuGet.Client/blob/4d815bf71aa1ce5fe8d9d9da6573303d7093c2f1/src/NuGet.Core/NuGet.Common/Strings.Designer.cs#L29), and [every string resource property passes it to `ResourceManager.GetString`](https://github.com/NuGet/NuGet.Client/blob/4d815bf71aa1ce5fe8d9d9da6573303d7093c2f1/src/NuGet.Core/NuGet.Common/Strings.Designer.cs#L68).  Since the usual pattern for using parameterized string resources is `string.Format(someCulture, Strings.Error_Something, arg1, arg2)`, this means that the string is retrieved from the resource manager BEFORE being passed to `string.Format`, so the culture passed to `string.Format` cannot affect the string resource localization used.

> However, changing the exception type is out of scope for this PR and would change the behavior of a public API. (I suspect throwable exceptions are outside the purview of the API analyzer, but I don't really know.)

You're right, the public API analyzer doesn't track what exceptions are being thrown, so changing exceptions wouldn't be caught by tooling.

And changing from `Exception` to `ArgumentException` would only break someone doing `exception.GetType() == typeof(Exception)`, or equivalent. But I hope you agree that doesn't sound like a very valuable thing to do outside this specific test (and a test validating the contract of which exception type is thrown seems like a very valid test to me). Customers using the NuGet CLient SDK and doing `catch (Exception)` will continue to catch the exception if we change the exception type. However, if we were later to change it from `ArgumentException` to anything else (that does not extend `ArgumentException`) then that would be a breaking change with "realistic" risk.

Having written that, `PatternExpression` doesn't sound like a useful API for customers using the NuGet Client SDK in their own app. It sounds like something that should be internal, an implementation detail of some package related API. However, we have no way to know if anyone actually uses it. But since it sounds like a "useless" API outside of NuGet.Packaging.dll, I'd be reasonably open to making a breaking change in this type, but I don't know if other people in the team feel the same. We have had failed insertions in the past because of broken APIs, and other times where insertion didn't catch the break, and another product team had to change their code when NuGet flowed into their repo (or a feature without automated tests failed, but these days the .NET SDK has a LOT of tests)

dtivel: @Nigusu-Allehu, @zivkan and I chatted offline and agreed on the following:

* Change [this exception](https://github.com/NuGet/NuGet.Client/blob/23f51111dc157ce393c9dc96fb87450ec2aec676/src/NuGet.Core/NuGet.Packaging/ContentModel/Infrastructure/Parser.cs#L173) to `System.ArgumentException`.
* Pass the current message to `ArgumentException'`s constructor for the `message` argument and the token for the `paramName` argument.  Here the token is `"{name}"`.
* The test should assert that `ArgumentException` is thrown and that `ArgumentException.ParamName` == `"{name}"`.

The spirit of the feedback is that a negative test should try to verify both failure (here, that an exception is thrown) and that the failure reason is expected (here, that the exception is the one you expect).

PR #5343, opened 2023-08-03
Add an allowInsecureConnections property in NuGet.Config

### LINKED ISSUE #12786: Opt-out of "HTTPs everywhere" : add an allowInsecureConnections property in NuGet.Config
We need to add an `allowInsecureConnections` property into `packageSources` section in NuGet.Config files, as below:
```
<!-- Allows insecure connections on a specific http source -->
<packageSources>
    <add key="Contoso" value="http://contoso.com/packages/" allowInsecureConnections="true" />
</packageSources>
```
Spec: https://github.com/NuGet/Home/blob/dev/proposed/2023/InsecureConnectionsDisableCertificateValidation.md#package-source-nuget-config

heng-liu:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12786

Regression? No.

## Description
Add an `allowInsecureConnections` property into `packageSources` section in NuGet.Config files, as below:
```
<!-- Allows insecure connections on a specific http source -->
<packageSources>
    <add key="Contoso" value="http://contoso.com/packages/" allowInsecureConnections="true" />
</packageSources>
```
The default value will be false.
Spec: https://github.com/NuGet/Home/blob/dev/proposed/2023/InsecureConnectionsDisableCertificateValidation.md#package-source-nuget-config

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [x] Documentation PR or issue filled  Tracking issue: https://github.com/NuGet/docs.microsoft.com-nuget/issues/3116
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetSourcesService.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetSourcesService.cs
@@ -93,6 +93,7 @@ namespace NuGet.PackageManagement.VisualStudio
                     if (packageSource.Name.Equals(packageSourceContextInfo.Name, StringComparison.InvariantCulture)
                         && packageSource.Source.Equals(packageSourceContextInfo.Source, StringComparison.InvariantCulture)
                         && packageSource.ProtocolVersion == packageSourceContextInfo.ProtocolVersion
+                        && packageSource.AllowInsecureConnections == packageSourceContextInfo.AllowInsecureConnections
                         && packageSource.IsEnabled == packageSourceContextInfo.IsEnabled)
                     {
                         newPackageSources.Add(packageSource);
@@ -111,6 +112,7 @@ namespace NuGet.PackageManagement.VisualStudio
                             ClientCertificates = packageSource.ClientCertificates,
                             Description = packageSource.Description,
                             ProtocolVersion = packageSourceContextInfo.ProtocolVersion,
+                            AllowInsecureConnections = packageSourceContextInfo.AllowInsecureConnections,
                             MaxHttpRequestsPerSource = packageSource.MaxHttpRequestsPerSource,
                         };
 
--- a/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/Formatters/PackageSourceContextInfoFormatter.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/Formatters/PackageSourceContextInfoFormatter.cs
@@ -74,7 +74,7 @@ namespace NuGet.VisualStudio.Internal.Contracts
             Assumes.NotNullOrEmpty(source);
             Assumes.NotNullOrEmpty(name);
 
-            return new PackageSourceContextInfo(source, name, isEnabled, protocolVersion)
+            return new PackageSourceContextInfo(source, name, isEnabled, protocolVersion, allowInsecureConnections)
             {
                 IsMachineWide = isMachineWide,
                 Description = description,
--- a/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/Formatters/PackageSourceContextInfoFormatter.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/Formatters/PackageSourceContextInfoFormatter.cs
@@ -34,7 +34,7 @@ namespace NuGet.VisualStudio.Internal.Contracts
             string? description = null;
             int originalHashCode = 0;
             int protocolVersion = PackageSource.DefaultProtocolVersion;
-            bool allowInsecureConnections = PackageSource.DefaultAllowInsecureConnections;
+            bool allowInsecureConnections = false;
 
             int propertyCount = reader.ReadMapHeader();
             for (int propertyIndex = 0; propertyIndex < propertyCount; propertyIndex++)
--- a/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/Formatters/PackageSourceContextInfoFormatter.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/Formatters/PackageSourceContextInfoFormatter.cs
@@ -13,6 +13,7 @@ namespace NuGet.VisualStudio.Internal.Contracts
         private const string SourcePropertyName = "source";
         private const string IsEnabledPropertyName = "isenabled";
         private const string ProtocolVersionPropertyName = "protocolversion";
+        private const string AllowInsecureConnectionsPropertyName = "allowInsecureConnections";
         private const string IsMachineWidePropertyName = "ismachinewide";
         private const string NamePropertyName = "name";
         private const string DescriptionPropertyName = "description";
@@ -33,6 +34,7 @@ namespace NuGet.VisualStudio.Internal.Contracts
             string? description = null;
             int originalHashCode = 0;
             int protocolVersion = PackageSource.DefaultProtocolVersion;
+            bool allowInsecureConnections = PackageSource.DefaultAllowInsecureConnections;
 
             int propertyCount = reader.ReadMapHeader();
             for (int propertyIndex = 0; propertyIndex < propertyCount; propertyIndex++)
@@ -60,6 +62,9 @@ namespace NuGet.VisualStudio.Internal.Contracts
                     case ProtocolVersionPropertyName:
                         protocolVersion = reader.ReadInt32();
                         break;
+                    case AllowInsecureConnectionsPropertyName:
+                        allowInsecureConnections = reader.ReadBoolean();
+                        break;
                     default:
                         reader.Skip();
                         break;
@@ -79,11 +84,13 @@ namespace NuGet.VisualStudio.Internal.Contracts
 
         protected override void SerializeCore(ref MessagePackWriter writer, PackageSourceContextInfo value, MessagePackSerializerOptions options)
         {
-            writer.WriteMapHeader(count: 7);
+            writer.WriteMapHeader(count: 8);
             writer.Write(SourcePropertyName);
             writer.Write(value.Source);
             writer.Write(ProtocolVersionPropertyName);
             writer.Write(value.ProtocolVersion);
+            writer.Write(AllowInsecureConnectionsPropertyName);
+            writer.Write(value.AllowInsecureConnections);
             writer.Write(IsEnabledPropertyName);
             writer.Write(value.IsEnabled);
             writer.Write(IsMachineWidePropertyName);
--- a/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSource.cs
+++ b/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSource.cs
@@ -19,7 +19,7 @@ namespace NuGet.Configuration
         /// </summary>
         public const int DefaultProtocolVersion = 2;
 
-        internal const bool DefaultAllowInsecureConnections = false;
+        public const bool DefaultAllowInsecureConnections = false;
 
         private int _hashCode;
         private string _source;
--- a/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSource.cs
+++ b/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSource.cs
@@ -19,7 +19,7 @@ namespace NuGet.Configuration
         /// </summary>
         public const int DefaultProtocolVersion = 2;
 
-        public const bool DefaultAllowInsecureConnections = false;
+        internal const bool DefaultAllowInsecureConnections = false;
 
         private int _hashCode;
         private string _source;
--- a/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSource.cs
+++ b/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSource.cs
@@ -19,6 +19,8 @@ namespace NuGet.Configuration
         /// </summary>
         public const int DefaultProtocolVersion = 2;
 
+        public const bool DefaultAllowInsecureConnections = false;
+
         private int _hashCode;
         private string _source;
         private bool _isHttp;
@@ -102,6 +104,11 @@ namespace NuGet.Configuration
         public int ProtocolVersion { get; set; } = DefaultProtocolVersion;
 
         /// <summary>
+        /// Gets or sets the allowInsecureConnections of the source. Defaults to false.
+        /// </summary>
+        public bool AllowInsecureConnections { get; set; } = DefaultAllowInsecureConnections;
+
+        /// <summary>
         /// Whether the source is using the HTTP protocol, including HTTPS.
         /// </summary>
         public bool IsHttp => _isHttp;
@@ -152,7 +159,13 @@ namespace NuGet.Configuration
             {
                 protocolVersion = $"{ProtocolVersion}";
             }
-            return new SourceItem(Name, Source, protocolVersion);
+
+            string? allowInsecureConnections = null;
+            if (AllowInsecureConnections != DefaultAllowInsecureConnections)
+            {
+                allowInsecureConnections = $"{AllowInsecureConnections}";
+            }
+            return new SourceItem(Name, Source, protocolVersion, allowInsecureConnections);
         }
 
         public bool Equals(PackageSource? other)
@@ -189,6 +202,7 @@ namespace NuGet.Configuration
                 ClientCertificates = ClientCertificates?.ToList(),
                 IsMachineWide = IsMachineWide,
                 ProtocolVersion = ProtocolVersion,
+                AllowInsecureConnections = AllowInsecureConnections,
             };
         }
     }
--- a/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
+++ b/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
@@ -199,6 +199,7 @@ namespace NuGet.Configuration
             }
 
             packageSource.ProtocolVersion = ReadProtocolVersion(setting);
+            packageSource.AllowInsecureConnections = ReadAllowInsecureConnections(setting);
 
             return packageSource;
         }
@@ -211,6 +212,16 @@ namespace NuGet.Configuration
             }
 
             return PackageSource.DefaultProtocolVersion;
+        }
+
+        private static bool ReadAllowInsecureConnections(SourceItem setting)
+        {
+            if (bool.TryParse(setting.AllowInsecureConnections, out var allowInsecureConnections))
+            {
+                return allowInsecureConnections;
+            }
+
+            return PackageSource.DefaultAllowInsecureConnections;
         }
 
         private static int AddOrUpdateIndexedSource(
--- a/src/NuGet.Core/NuGet.Configuration/PublicAPI.Shipped.txt
+++ b/src/NuGet.Core/NuGet.Configuration/PublicAPI.Shipped.txt
@@ -281,6 +281,7 @@ NuGet.Configuration.SettingsUtility
 NuGet.Configuration.SourceItem
 ~NuGet.Configuration.SourceItem.ProtocolVersion.get -> string
 ~NuGet.Configuration.SourceItem.ProtocolVersion.set -> void
+~NuGet.Configuration.SourceItem.SourceItem(string key, string value, string protocolVersion) -> void
 NuGet.Configuration.StoreClientCertItem
 NuGet.Configuration.StoreClientCertItem.FindType.get -> System.Security.Cryptography.X509Certificates.X509FindType
 ~NuGet.Configuration.StoreClientCertItem.FindValue.get -> string
--- a/src/NuGet.Core/NuGet.Configuration/PublicAPI.Shipped.txt
+++ b/src/NuGet.Core/NuGet.Configuration/PublicAPI.Shipped.txt
@@ -281,7 +281,6 @@ NuGet.Configuration.SettingsUtility
 NuGet.Configuration.SourceItem
 ~NuGet.Configuration.SourceItem.ProtocolVersion.get -> string
 ~NuGet.Configuration.SourceItem.ProtocolVersion.set -> void
-~NuGet.Configuration.SourceItem.SourceItem(string key, string value, string protocolVersion = "") -> void
 NuGet.Configuration.StoreClientCertItem
 NuGet.Configuration.StoreClientCertItem.FindType.get -> System.Security.Cryptography.X509Certificates.X509FindType
 ~NuGet.Configuration.StoreClientCertItem.FindValue.get -> string
--- a/src/NuGet.Core/NuGet.Configuration/Settings/Items/SourceItem.cs
+++ b/src/NuGet.Core/NuGet.Configuration/Settings/Items/SourceItem.cs
@@ -21,12 +21,40 @@ namespace NuGet.Configuration
             set => AddOrUpdateAttribute(ConfigurationConstants.ProtocolVersionAttribute, value);
         }
 
-        public SourceItem(string key, string value, string protocolVersion = "")
+        public string AllowInsecureConnections
+        {
+            get
+            {
+                if (Attributes.TryGetValue(ConfigurationConstants.AllowInsecureConnections, out var attribute))
+                {
+                    return Settings.ApplyEnvironmentTransform(attribute);
+                }
+
+                return null;
+            }
+            set => AddOrUpdateAttribute(ConfigurationConstants.AllowInsecureConnections, value);
+        }
+
+        public SourceItem(string key, string value)
+            : this(key, value, protocolVersion: "", allowInsecureConnections: "")
+        {
+        }
+
+        public SourceItem(string key, string value, string protocolVersion)
+            : this(key, value, protocolVersion, allowInsecureConnections: "")
+        {
+        }
+
+        public SourceItem(string key, string value, string protocolVersion = "", string allowInsecureConnections = "")
             : base(key, value)
         {
             if (!string.IsNullOrEmpty(protocolVersion))
             {
                 ProtocolVersion = protocolVersion;
+            }
+            if (!string.IsNullOrEmpty(allowInsecureConnections))
+            {
+                AllowInsecureConnections = allowInsecureConnections;
             }
         }
 
@@ -37,7 +65,7 @@ namespace NuGet.Configuration
 
         public override SettingBase Clone()
         {
-            var newSetting = new SourceItem(Key, Value, ProtocolVersion);
+            var newSetting = new SourceItem(Key, Value, ProtocolVersion, AllowInsecureConnections);
 
             if (Origin != null)
             {
--- a/src/NuGet.Core/NuGet.Configuration/Settings/Items/SourceItem.cs
+++ b/src/NuGet.Core/NuGet.Configuration/Settings/Items/SourceItem.cs
@@ -45,7 +45,7 @@ namespace NuGet.Configuration
         {
         }
 
-        public SourceItem(string key, string value, string protocolVersion = "", string allowInsecureConnections = "")
+        public SourceItem(string key, string value, string protocolVersion, string allowInsecureConnections)
             : base(key, value)
         {
             if (!string.IsNullOrEmpty(protocolVersion))
--- a/src/NuGet.Core/NuGet.Configuration/Utility/ConfigurationContants.cs
+++ b/src/NuGet.Core/NuGet.Configuration/Utility/ConfigurationContants.cs
@@ -11,6 +11,8 @@ namespace NuGet.Configuration
 
         public static readonly string AllowUntrustedRoot = "allowUntrustedRoot";
 
+        public static readonly string AllowInsecureConnections = "AllowInsecureConnections";
+
         public static readonly string ApiKeys = "apikeys";
 
         public static readonly string Author = "author";
--- a/src/NuGet.Core/NuGet.Configuration/Utility/ConfigurationContants.cs
+++ b/src/NuGet.Core/NuGet.Configuration/Utility/ConfigurationContants.cs
@@ -11,7 +11,7 @@ namespace NuGet.Configuration
 
         public static readonly string AllowUntrustedRoot = "allowUntrustedRoot";
 
-        public static readonly string AllowInsecureConnections = "AllowInsecureConnections";
+        public static readonly string AllowInsecureConnections = "allowInsecureConnections";
 
         public static readonly string ApiKeys = "apikeys";
 
--- a/test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Services/NuGetSourcesServiceTests.cs
+++ b/test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Services/NuGetSourcesServiceTests.cs
@@ -130,5 +130,42 @@ namespace NuGet.PackageManagement.VisualStudio.Test
             savedSources!.Count.Should().Be(1);
             savedSources[0].ProtocolVersion.Should().Be(3);
         }
+
+        [Fact]
+        public async Task Save_SourceWithDifferentAllowInsecureConnections_SavesNewValue()
+        {
+            PackageSource packageSource = new(name: "Source-Name", source: "Source-Path")
+            {
+                ProtocolVersion = 3,
+                AllowInsecureConnections = false
+            };
+
+            Mock<IPackageSourceProvider> packageSourceProvider = new();
+            packageSourceProvider.Setup(psp => psp.LoadPackageSources())
+                .Returns(new[] { packageSource });
+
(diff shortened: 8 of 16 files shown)


src/NuGet.Core/NuGet.Configuration/Utility/ConfigurationContants.cs
```diff
@@ -11,6 +11,8 @@ public static class ConfigurationConstants
 
         public static readonly string AllowUntrustedRoot = "allowUntrustedRoot";
 
+        public static readonly string AllowInsecureConnections = "AllowInsecureConnections";
```
aortiz-msft: Should this be camelcased like "allowUntrustedRoot"?

src/NuGet.Core/NuGet.Configuration/PublicAPI.Shipped.txt
```diff
@@ -288,7 +288,6 @@ NuGet.Configuration.SettingsUtility
 NuGet.Configuration.SourceItem
 ~NuGet.Configuration.SourceItem.ProtocolVersion.get -> string
 ~NuGet.Configuration.SourceItem.ProtocolVersion.set -> void
-~NuGet.Configuration.SourceItem.SourceItem(string key, string value, string protocolVersion = "") -> void
```
aortiz-msft, line 284: Why not keep this and/or the new overload?: 
`NuGet.Configuration.SourceItem.SourceItem(string key, string value, string protocolVersion = "", string allowInsecureConnections = "")`

src/NuGet.Core/NuGet.Configuration/Utility/ConfigurationContants.cs
aortiz-msft: Should this be camelCased like the above?

aortiz-msft (approved):

aortiz-msft (dismissed):
Should AsSourceItem_WorksCorrectly() and Clone_CopiesAllPropertyValuesFromSource() be updated in NuGet.Client\test\NuGet.Core.Tests\NuGet.Configuration.Test\PackageSourceTests.cs

src/NuGet.Core/NuGet.Configuration/PublicAPI.Shipped.txt
heng-liu: Thanks for reviewing!
Adding multiple overloads with optional parameters violates a rule in analyzer : 
Symbol 'Saurceltem' violates the backcompat requirement: 'Do not add multiple overloads with optional parameters'. See '[https://github.com/dotnet/roslyn/blob/main/docs/Adding%20Optional%20Parameters%20in%20Public%20API.md](https://github.com/dotnet/roslyn/blob/main/docs/Adding%20Optional%20Parameters%20in%20Public%20API.md)' for details.

src/NuGet.Core/NuGet.Configuration/Utility/ConfigurationContants.cs
heng-liu: Good catch! Fixed. Thanks!

zivkan (dismissed):
PR template not filled out.

In particular, since you're changing nuget.config, we need a docs issue to update the nuget.config reference doc at a minimum. If we want this documented someone in addition to the nuget.config reference, then the docs issue should say where else it should be documented.

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSource.cs
```diff
@@ -19,6 +19,8 @@ public class PackageSource : IEquatable<PackageSource>
         /// </summary>
         public const int DefaultProtocolVersion = 2;
 
+        public const bool DefaultAllowInsecureConnections = false;
```
zivkan: I don't see this as being very useful as a public API: https://github.com/NuGet/NuGet.Client/blob/dev/docs/nuget-sdk.md#changes-to-apis

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
```diff
@@ -698,7 +709,8 @@ public void SavePackageSources(IEnumerable<PackageSource> sources, PackageSource
 
                 if (existingSettingsLookup != null &&
                     existingSettingsLookup.TryGetValue(source.Name, out existingSourceItem) &&
-                    ReadProtocolVersion(existingSourceItem) == source.ProtocolVersion)
+                    ReadProtocolVersion(existingSourceItem) == source.ProtocolVersion &&
+                    ReadAllowInsecureConnections(existingSourceItem) == source.AllowInsecureConnections)
```
zivkan: This is buggy, and exactly what I'm fixing in https://github.com/NuGet/NuGet.Client/pull/5334

Unfortunately, although I addressed all feedback 2 days ago, nobody has reviewed or approved the PR since then, so it's still not merged 😞

src/NuGet.Core/NuGet.Configuration/PublicAPI.Shipped.txt
zivkan: In general, I'd say the solution to analyzer violations for public APIs is to suppress the analyzer for that method, so we don't break backwards compatibility for our package.

However, I expect that the multiple overloads added for the constructor, this specific PR is both API and ABI backwards compatible. It's just that recompiling the same code with different versions of this dll will result in a different method signature being called.

It's probably worthwhile hand editing PublicAPI.Shipped.txt and PublicAPI.Unshipped.txt to keep this 3 string overload in this file, so we know it's an existing API that we should probably keep backwards compat for, not a new API. The Public API analyzer's code fix suggestions are very simplistic, so I find myself often having to manually hand edit these files when it's not the first time I'm adding a new public API.

src/NuGet.Core/NuGet.Configuration/Settings/Items/SourceItem.cs
```diff
@@ -23,13 +23,41 @@ public string ProtocolVersion
...
+            : this(key, value, protocolVersion: "", allowInsecureConnections: "")
+        {
+        }
+
+        public SourceItem(string key, string value, string protocolVersion)
+            : this(key, value, protocolVersion, allowInsecureConnections: "")
+        {
+        }
+
+        public SourceItem(string key, string value, string protocolVersion = "", string allowInsecureConnections = "")
```
zivkan: The optional parameters no longer provide value.  If someone has source code calling `new SourceItem(key, value)`, it's going to use the 2 string overload, not this one with the default parameters. Same for the 3 string `new Source(key, value, protocolVersion)` overload, it's not going to use this 4 string overload. Therefore, the only way to call this overload is to explicitly pass in all 4 parameters, at which point none of them are using the default values.

Well, that's not entirely true. It's possible to call `new SourceItem(key, value, allowInsecureConnections: "true")`, but since this is a new API, there's obviously not going to be any existing code doing this.

It also avoid potential future issues with public APIs and default parameter analyzer violations.

src/NuGet.Core/NuGet.Configuration/Settings/Items/SourceItem.cs
```diff
@@ -23,13 +23,41 @@ public string ProtocolVersion
             set => AddOrUpdateAttribute(ConfigurationConstants.ProtocolVersionAttribute, value);
         }
 
-        public SourceItem(string key, string value, string protocolVersion = "")
+        public string AllowInsecureConnections
```
zivkan, line 24: You should check what happens when the nuget.config already has a value for `allowInsecureConnections`, and then you edit sources in VS's options page. I think VS will strip out the `allowInsecureConnections` value, so I think you need to add this property into `PackageSourceContextInfo`, its MsgPack formatter, and possibly `NuGetSourcesService.GetPackageSourcesToUpdate`, as I did in this PR: https://github.com/NuGet/NuGet.Client/pull/5334

test/NuGet.Core.Tests/NuGet.Configuration.Test/SettingsFileParsingTests/SourceItemTests.cs
zivkan, line 1: There are two tests validating that sourceItem can be parsed successfully with `allowInsecureConnections`, but there is no test validating that it gets serialized back to nuget.config correctly

test/NuGet.Core.Tests/NuGet.Packaging.Test/RuntimeModelTests/RuntimeDescriptionTests.cs
```diff
@@ -45,7 +45,7 @@ public void Merge_PrefersRightImports()
     [Fact]
     public void Merge_UsesLeftImportsIfRightEmpty()
     {
-        RuntimeDescription l = new(runtimeIdentifier: "rid", Array.Empty<string>(),  Array.Empty<RuntimeDependencySet>());
+        RuntimeDescription l = new(runtimeIdentifier: "rid", Array.Empty<string>(), Array.Empty<RuntimeDependencySet>());
```
zivkan: You need to rebase your branch. This was already fixed: https://github.com/NuGet/NuGet.Client/pull/5336/files

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
nkolev92: image

We didn't know it was ready for a new review :)

aortiz-msft:
> Should AsSourceItem_WorksCorrectly() and Clone_CopiesAllPropertyValuesFromSource() be updated in NuGet.Client\test\NuGet.Core.Tests\NuGet.Configuration.Test\PackageSourceTests.cs

@heng-liu  I didn't see an update on this in the latest iteration.

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
donnie-msft: @zivkan can you reach out to @aortiz-msft directly?
image

test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs
```diff
@@ -822,6 +822,56 @@ public void LoadPackageSources_ReadsSourcesWithProtocolVersionFromPackageSourceS
...
         }
 
+        [Theory]
+        [InlineData(true, null, null)]
+        [InlineData(true, "3", null)]
+        [InlineData(true, "3", "false")]
+        [InlineData(true, "2", "true")]
+        [InlineData(false, null, "true")]
+        [InlineData(false, "", "true")]
+        public void LoadPackageSources_ReadsSourcesWithAllowInsecureConnectionsFromPackageSourceSections(bool useStaticMethod, string protocolVersion, string allowInsecureConnections)
```
donnie-msft: I know people like keeping code to a minimum, but I would say this is a good time to create multiple tests. When there's significant branching in Assertions, that makes it hard to understand what it means when the test fails. Also, the name of the method doesn't indicate the criteria and success conditions.

eg, we could have (just example wordings):

`LoadPackageSources_WithNullProtocolVersion_LoadsDefault`
`LoadPackageSources_WithNullAllowInsecureConnections_LoadsDefault`
...
`LoadPackageSources_WithNotNullProtocolVersion_LoadsValue`

test/NuGet.Core.Tests/NuGet.Configuration.Test/SettingsFileParsingTests/SourceItemTests.cs
```diff
@@ -54,14 +54,20 @@ public void SourceItem_ParsedSuccessfully()
...
+        <add key='nuget5' value='http://serviceIndex.test5/api/index.json' allowInsecureConnections='false' protocolVersion='2' />
     </packageSources>
 </configuration>";
 
             var expectedValues = new List<SourceItem>()
             {
                 new SourceItem("nugetorg","http://serviceIndexorg.test/api/index.json"),
-                new SourceItem("nuget2","http://serviceIndex.test/api/index.json", "3" ),
+                new SourceItem("nuget2","http://serviceIndex.test2/api/index.json", "3" ),
+                new SourceItem("nuget3","http://serviceIndex.test3/api/index.json", protocolVersion: "3", allowInsecureConnections: "true" ),
```
kartheekp-ms, line 69: what happens if we specify `allowInsecureConnections` in a different case? I assume that setting will be applied correctly but wanted to make sure.

test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs
aortiz-msft: +1 to above comment.

src/NuGet.Core/NuGet.Configuration/PublicAPI.Shipped.txt
heng-liu: Thanks! I manually edited the PublicAPI files.

heng-liu:
> PR template not filled out.
> 
> In particular, since you're changing nuget.config, we need a docs issue to update the nuget.config reference doc at a minimum. If we want this documented someone in addition to the nuget.config reference, then the docs issue should say where else it should be documented.

Thanks! Created a doc issue and fill out the PR template.

test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs
heng-liu: Thanks! Fixed.

test/NuGet.Core.Tests/NuGet.Packaging.Test/RuntimeModelTests/RuntimeDescriptionTests.cs
heng-liu: Thanks! Fixed.

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
heng-liu: Thanks for reminding me about this recent change! This is fixed.

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSource.cs
heng-liu: Thanks! Fixed. It's no more a public API.

test/NuGet.Core.Tests/NuGet.Configuration.Test/SettingsFileParsingTests/SourceItemTests.cs
heng-liu: Thanks for reviewing! Yes, it works for a different case. I changed one line to use Uppercase.

test/NuGet.Core.Tests/NuGet.Configuration.Test/SettingsFileParsingTests/SourceItemTests.cs
heng-liu: Thanks! Added a test case `SourceItem_AsXNode_ReturnsExpectedXNode`.

heng-liu:
> Should AsSourceItem_WorksCorrectly() and Clone_CopiesAllPropertyValuesFromSource() be updated in NuGet.Client\test\NuGet.Core.Tests\NuGet.Configuration.Test\PackageSourceTests.cs

Thanks! Updated the test case in the above test cases.
However, the test [AsSourceItem_WorksCorrectly()](https://github.com/NuGet/NuGet.Client/blob/dev-hengliu-optoutHttpWarning-addAllowInsecureConnectionsProperty/test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceTests.cs#L44) in PackageSourceTests testing `AsSourceItem` seems kind of confusing.
As the PackageSource.AsSourceItem() is not guarantied to be equal to a new SouceItem for the following reasons:
1. If the ProtocalVersion or the AllowInsecureConnections are the default value (int 2, bool false), then the two parameters passed into SouceItem constructor will be null: https://github.com/NuGet/NuGet.Client/blob/dev-hengliu-optoutHttpWarning-addAllowInsecureConnectionsProperty/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSource.cs#L157-L168
2. "True" and "true" will be treated as different attribute values when comparing the two sourceItems. 
The is okey as [DeepEquals](https://github.com/NuGet/NuGet.Client/blob/dev/test/NuGet.Core.Tests/NuGet.Configuration.Test/SettingsTestUtils.cs#L81) is only used in our test.
PackageSourceProvider will handle the normalization at [ReadAllowInsecureConnections](https://github.com/NuGet/NuGet.Client/blob/dev-hengliu-optoutHttpWarning-addAllowInsecureConnectionsProperty/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs#L217-L225) We have new tests in [PackageSourceProviderTests](https://github.com/NuGet/NuGet.Client/blob/dev-hengliu-optoutHttpWarning-addAllowInsecureConnectionsProperty/test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs#L825-L920) verified that.
So this test just run a special case to check if AsSourceItem() works. 
To avoid confusion caused, I added some comments to this test case.
Pls let me know if anyone thinks we should remove this test.

src/NuGet.Core/NuGet.Configuration/Settings/Items/SourceItem.cs
heng-liu: Thanks! That's very helpful! Fixed and added tests. Manually tested in VS, the non-default value of  `allowInsecureConnections` will not be stripped out when updating in VS options page. The default value(false) and invalid value(e.g. "random", will be changed into default value) will be removed after updating, as an expected normalization work. (Just like the protocal Version "2" will be stripped out after updating).

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSource.cs
heng-liu: Sorry, had to change it back into a public API, as [PackageSourceContextInfoFormatter.cs](https://github.com/NuGet/NuGet.Client/blob/dev-hengliu-optoutHttpWarning-addAllowInsecureConnectionsProperty/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/Formatters/PackageSourceContextInfoFormatter.cs#L37) needs this.

heng-liu:
I spent some time on investigating the tricky test AsSourceItem_WorksCorrectly() and sorry for the delay!
All comments are addressed. Please take another look. Thanks!

src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/Formatters/PackageSourceContextInfoFormatter.cs
```diff
@@ -33,6 +34,7 @@ private PackageSourceContextInfoFormatter()
             string? description = null;
             int originalHashCode = 0;
             int protocolVersion = PackageSource.DefaultProtocolVersion;
+            bool allowInsecureConnections = PackageSource.DefaultAllowInsecureConnections;
```
zivkan: this variable is updated when reading  all the properties, but it's not used when creating a new instance of `PackageSourceContextInfo` around like 77.

Since no tests failed, this means there's also missing and/or buggy tests. A good way to avoid these types of accidents is to write tests first, that fail, and then change the product code so the test passes (test driven development, TDD).

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSource.cs
zivkan: I might be having an overly strong, or just unpopular, opinion, so consider this a nitpick. I'm curious what other peoples opinions are. But, I still don't think it's a good/useful public API.

`default(bool)` is `false`, so providing a different way to get the same const value isn't useful. Providing a const for a non-obvious value makes sense, which is why `DefaultProtocolVersion` isn't so bad. But I think it should be obvious to anyone who understands basic security principals that allowing insecure connections is not default. And therefore I can't imagine this ever changing in the future, and changing defaults is a good reason to have pubic APIs.

But on the topic of changing defaults, from an API design point of view, I think a const is problematic for providing default values. The default protocol version above has the same problem. If in the future we want to change the defaults, then it will only impact projects that update their package version and recompile their project. Consts get "burned in" at compile time. So, if I have `Zivkan.Utils` that compiled against this `false` default, then someone using my package uses my package, but a newer version of `NuGet.Configuration` that changed the default, my `Zivkan.Utils` library will continue using the old, now incorrect, value at runtime. If we want to provide defaults as public APIs, get-only properties, or `static readonly` fields will cause assemblies that reference `NuGet.Configuraiton` to read the updated defaults at runtime.

The first BCL example to use as prior art that I can think of is `HttpClient.DefaultRequestVersion` and `HttpRequestMessage.Version`. The docs say what the default is if you don't explicitly set the value youself, but there is no static or const API to get the default. But maybe the default HTTP protocol version isn't a good example, because it's not serialized to disk.

The first example I could think of a value that is serialized to disk is how before .NET Core 1.0, all MSBuild project files had to have XML namespaces. But MSBuild's APIs don't appear to have a public API to get the default namespace. All examples I see in a quick search are `private` or `internal`, and they've duplicated it in a few assemblies, rather than having a single public API that's referenced from all other assemblies: https://github.com/search?q=repo%3Adotnet%2Fmsbuild%20http%3A%2F%2Fschemas.microsoft.com%2Fdeveloper%2Fmsbuild%2F2003&type=code

I think it's perfectly acceptable to use `bool allowInsecureConnections = false`, rather than this const across assembly boundaries, in `PackageSourceContextInfoFormatter`.

src/NuGet.Core/NuGet.Configuration/Settings/Items/SourceItem.cs
zivkan: This comment hasn't been addressed. I feel strongly that the default values should be removed, so we don't get those analyzer warnings if trying to add more overloads in the future. Honestly, in my opinion, public APIs should never use default values. Leave it for private and internal APIs only.

test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs
```diff
@@ -822,6 +822,103 @@ public void LoadPackageSources_ReadsSourcesWithProtocolVersionFromPackageSourceS
...
+            Assert.Equal("Source", loadedSource.Name);
+            Assert.Equal("https://some-source.test", loadedSource.Source);
+            AssertCredentials(loadedSource.Credentials, "Source", "source-user", "source-password");
+            Assert.Equal(PackageSource.DefaultAllowInsecureConnections, loadedSource.AllowInsecureConnections);
+        }
+
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void LoadPackageSources_ReadsSourcesWithRandomAllowInsecureConnectionsFromPackageSourceSections_LoadsDefault(bool useStaticMethod)
```
zivkan:
```suggestion
        public void LoadPackageSources_ReadsSourcesWithInvalidAllowInsecureConnectionsFromPackageSourceSections_LoadsDefault(bool useStaticMethod)
```

having a test name saying "with random ..." makes it sound like you're going to use `System.Random` to auto-generate random values to test with (basically, [fuzz testing](https://en.wikipedia.org/wiki/Fuzzing)), which is not what this test is doing.

test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs
```diff
@@ -822,6 +822,103 @@ public void LoadPackageSources_ReadsSourcesWithProtocolVersionFromPackageSourceS
...
+            var sourceItem = new SourceItem("Source", "https://some-source.test", protocolVersion: null, allowInsecureConnections: allowInsecureConnections);
+
+            settings.Setup(s => s.GetSection("packageSources"))
+                .Returns(new VirtualSettingSection("packageSources",
+                    sourceItem));
+
+            settings
+                .Setup(s => s.GetSection("packageSourceCredentials"))
+                .Returns(new VirtualSettingSection("packageSourceCredentials",
+                    new CredentialsItem("Source", "source-user", "source-password", isPasswordClearText: true, validAuthenticationTypes: null)));
```
zivkan: Is there a reason I can't think of why credentials are being tested in addition to the allow insecure connection setting?  The test name doesn't say anything about credentials, and I would expect that other tests already validate credentials. Could all the tests be 6 lines shorter by removing the credentials stuff? (and anything else that's already being tested elsewhere/is not relevant to the name of the test)

I found the credentials distracting, making it slower to read and understand the tests.

test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs
```diff
@@ -822,6 +822,103 @@ public void LoadPackageSources_ReadsSourcesWithProtocolVersionFromPackageSourceS
...
+
+            settings.Setup(s => s.GetConfigFilePaths())
+                .Returns(new List<string>());
+            // Act
```
zivkan: nitpick: empty line between these two. Same issue in ther other 2 tests as well.

test/NuGet.Core.Tests/NuGet.Configuration.Test/SettingsFileParsingTests/SourceItemTests.cs
```diff
@@ -76,18 +83,58 @@ public void SourceItem_ParsedSuccessfully()
                 var section = settingsFile.GetSection("packageSources");
                 section.Should().NotBeNull();
 
-                var children = section.Items.ToList();
+                var children = section.Items.Select(c => c as SourceItem).ToList();
```
zivkan: [please don't use `as` to cast, unless you handle nulls](https://github.com/NuGet/NuGet.Client/pull/5309) 😭 

```suggestion
                var children = section.Items.Cast<SourceItem>().ToList();
```

test/NuGet.Core.Tests/NuGet.Configuration.Test/SettingsFileParsingTests/SourceItemTests.cs
```diff
@@ -76,18 +83,58 @@ public void SourceItem_ParsedSuccessfully()
...
+                        new XAttribute("protocolVersion", "2"),
+                        new XAttribute("allowInsecureConnections", "false")),
+                    new XElement("add",
+                        new XAttribute("key", "nuget3"),
+                        new XAttribute("value", "http://serviceIndex.test3/api/index.json"),
+                        new XAttribute("allowInsecureConnections", "true")),
+                    new XElement("add",
+                        new XAttribute("key", "nuget4"),
+                        new XAttribute("value", "http://serviceIndex.test4/api/index.json"),
+                        new XAttribute("protocolVersion", "3"))));
```
zivkan, line 132: I find this more difficult to read than `const string expectedXml = @"<configuration> ...`. Whether you use `XElement.Parse` or serialize the output of `AsXNode` to a string and then compare strings, I don't think it matters too much. A good way to choose between the two is to make the test fail, see what the exception message is, and then pick the one that is easier to understand.

src/NuGet.Core/NuGet.Configuration/Settings/Items/SourceItem.cs
heng-liu: Thanks! Fixed. Removed the public API with optional arguments.

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSource.cs
heng-liu: Thanks! Fixed.

test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs
heng-liu: Thanks! Fixed.

test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs
heng-liu: Thanks! Fixed.

test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs
heng-liu: Thanks! Fixed.

test/NuGet.Core.Tests/NuGet.Configuration.Test/SettingsFileParsingTests/SourceItemTests.cs
heng-liu: Thanks! Fixed.

test/NuGet.Core.Tests/NuGet.Configuration.Test/SettingsFileParsingTests/SourceItemTests.cs
heng-liu: Thanks! Fixed.

src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/Formatters/PackageSourceContextInfoFormatter.cs
heng-liu: Good catch! Fixed both the src and test.
The original test is just comparing the name and source, changed it into comparing the hashcode.

heng-liu:
Thanks for all the comments! I think all the comments are addressed. Pls help review it again. Thanks!

src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/Formatters/PackageSourceContextInfoFormatter.cs
```diff
@@ -13,6 +13,7 @@ internal sealed class PackageSourceContextInfoFormatter : NuGetMessagePackFormat
         private const string SourcePropertyName = "source";
         private const string IsEnabledPropertyName = "isenabled";
         private const string ProtocolVersionPropertyName = "protocolversion";
+        private const string AllowInsecureConnectionsPropertyName = "allowInsecureConnections";
```
donnie-msft, line 16: I like your use of camel-casing better, but should this be lowercase to be consistent?

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Services/NuGetSourcesServiceTests.cs
```diff
@@ -130,5 +130,42 @@ public async Task Save_SourceWithDifferentProtocolVersion_SavesNewValue()
...
+            List<PackageSource>? savedSources = null;
+            packageSourceProvider.Setup(psp => psp.SavePackageSources(It.IsAny<IEnumerable<PackageSource>>()))
+                .Callback((IEnumerable<PackageSource> newSources) => { savedSources = newSources.ToList(); });
+
+            var target = new NuGetSourcesService(options: default,
+                Mock.Of<IServiceBroker>(),
+                new AuthorizationServiceClient(Mock.Of<IAuthorizationService>()),
+                packageSourceProvider.Object);
+
+            List<PackageSourceContextInfo> updatedSources = new(1)
```
donnie-msft, line 156: nit: 
```suggestion
            List<PackageSourceContextInfo> updatedSources = new(capacity: 1)
```

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetSourcesService.cs
```diff
@@ -93,6 +93,7 @@ private IReadOnlyList<PackageSource> GetPackageSourcesToUpdate(IReadOnlyList<Pac
                     if (packageSource.Name.Equals(packageSourceContextInfo.Name, StringComparison.InvariantCulture)
                         && packageSource.Source.Equals(packageSourceContextInfo.Source, StringComparison.InvariantCulture)
                         && packageSource.ProtocolVersion == packageSourceContextInfo.ProtocolVersion
+                        && packageSource.AllowInsecureConnections == packageSourceContextInfo.AllowInsecureConnections
```
donnie-msft, line 96: The comment above, which I don't think is useful, is now inaccurate.
1. `// If key properties have not changed, we don't need to do anything`
2. `// If identifying properties have not changed, we don't need to do anything`
3. `// If Name/Source/IsEnabled/ProtocolVersion/AllowInsecureConnections has not changed, we don't need to do `anything
4. delete the comment

I vote for 4 or 1.

donnie-msft (approved):

jeffkl (approved):

aortiz-msft (approved):

PR #5347, opened 2023-08-08
Enable NuGetAudit by default for all PackageReference projects

zivkan:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2381

Regression? No

## Description

* Delete the `default` value for `NuGetAudit` in *NuGet.Targets*
* Change `AuditUtils`, so that null or empty strings cause implicit opt-in. This means that all PackageReference projects will be implicitly enabled, both SDK and non-SDK style projects.
* Allow `AuditUtils` to accept nulls for `RestoreAuditProperties`, since it's now possible for audit to be enabled when none of the audit properties are set.
* Disable `NuGetAudit` for a bunch of tests that were failing (asserting zero warnings, for eample)
* Change test `ExecuteAsync_WithSinglePackage_PopulatesCorrectTelemetry` so it's much easier to figure out when there are missing, or additional telemetry properties, and what the property names are.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [x] Documentation PR or issue filled: https://github.com/NuGet/docs.microsoft.com-nuget/issues/3117
  - **OR**
  - [ ] N/A


src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
zivkan, line 1: This PR's change has a less than ideal experience for non-SDK style projects because of the following bug, but that bug is not a regression of this PR
- https://github.com/NuGet/Home/issues/12800

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -431,9 +431,9 @@ internal enum EnabledValue
             ExplicitOptOut
         }
 
-        public static EnabledValue ParseEnableValue(string value)
+        public static EnabledValue ParseEnableValue(string? value)
```
jeffkl: Is it worth simplifying this method to always return "enabled" unless the value is false or disabled since its either on (by default) or you have to turn it off?

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -431,9 +431,9 @@ internal enum EnabledValue
...
         {
-            if (string.Equals(value, "default", StringComparison.OrdinalIgnoreCase))
+            if (string.IsNullOrEmpty(value))
```
jeffkl: Remind me what happens if I set the value to something like "Foo"?  This returns `EnabledValue.Undefined` and the functionality is disabled?

zivkan: Yes, currently the invalid value would disable audit.

This reminds me I wasn't sure what we should should when the value is `Foo` and I wanted to ask others opinion if we should default to implicit opt in, and output a message, similar to when we parse `NuGetAuditMode`?

https://github.com/NuGet/NuGet.Client/blob/3dbfac4cf9633d012ff22536cdb895e2f4e4b487/src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs#L419-L423

nkolev92 (commented):
What happens with the warning for audit being enabled when no sources support auditing?

test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommandTests.cs
```diff
@@ -2135,7 +2135,8 @@ public async Task RestoreCommand_RestoreFloatingVersionWithIgnoreFailingLocalSou
                 }");
 
                 var specPath = Path.Combine(projectDir, "TestProject", "project.json");
-                var spec = JsonPackageSpecReader.GetPackageSpec(configJson.ToString(), "TestProject", specPath);
+                var spec = JsonPackageSpecReader.GetPackageSpec(configJson.ToString(), "TestProject", specPath)
+                    .EnsureProjectJsonRestoreMetadata();
```
aortiz-msft, line 2139: > .EnsureProjectJsonRestoreMetadata(); [](http://example.com/codeflow?start=19&length=37)

NIT: Could we have these on the same line as above? They look odd on their own line.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
aortiz-msft: IMO false should disable or opt-out of the feature. Any other value should be the same as the default behavior which is "enabled". Accordingly, it'd be great if we could simplify the logic as suggested in the other comment.

aortiz-msft (dismissed):
:shipit:

zivkan:
> What happens with the warning for audit being enabled when no sources support auditing?

@nkolev92 github continues to annoy me with a lack of features. If you don't mind asking these types of questions in a "conversation" on a file? (just pick a random file if there isn't one file that is an obvious best choice for such a question)  That way others can easily see responses, and the "conversation" can be marked as resolved, making it obvious that it's not still an open question. I don't know when GitHub added it, but a few weeks ago I noticed that the right hand side of the file has a comment button, allowing you to comment on a file, without choosing a specific line of code, if that helps: 
image

Anyway, NU1905 still gets reported only when customers explicitly opt on, which is why `ParseAuditEnabled` has different return values for `ExplicitOptIn` and `ImplicitOptIn`. So, no change in behaviour. This PR also modifies the test `Check_WithNoVulnerabilitySources_NU1905Warning`, which validates that it only gets output when explicitly opt in, not implicit opt-in.

test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommandTests.cs
zivkan: I disagree for multiple reasons. I'm sure this response is over the top, but here it is anyway.

Firstly, this type of "method chaining" is particularly common for "fluent interfaces" which is common/popular enough to have its own wikipedia article: https://en.wikipedia.org/wiki/Fluent_interface
Most of the examples have each chained method on a different line.

Here's an example where putting each chained method on a new line makes it easier to see that package id, version, and icon is being "built". If it was one line line, it wouldn't be so obvious how many different metadata items are being set:
https://github.com/NuGet/NuGet.Client/blob/3dbfac4cf9633d012ff22536cdb895e2f4e4b487/test/TestUtilities/Test.Utility/Protocol/LocalPackageSearchMetadataFixture.cs#L24-L28

Secondly, consider other scenarios where method chaining is used, but with nested expressions. Look at this parenthesis soup (I count 7 pairs): https://github.com/NuGet/NuGet.Client/blob/3dbfac4cf9633d012ff22536cdb895e2f4e4b487/test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Services/NuGetPackageSearchServiceTests.cs#L274
Maybe this is just a bad example, and we shouldn't say a technique/style is bad just because of some bad examples, but what is the final `Build()` being called on? The `NuGetVersion.Parse` result? On the `new PackageIdentity` instance? On the result of `PackageSearchMetadataBuilder.FromIdentity`? On `Task.FromResult`? It would be so much easier to understand if it was on multiple lines and indented appropriately, so we could see which methods are being chained to what. Of course, there are multiple ways this could have been written differently to only use 2 to 4 short lines. However, if you search the repo for `))).`, I see 138 instances, so it appears that chaining & nesting method calls isn't uncommon in our code, or 1540 instances of `)).`. LINQ really makes it more common, since most LINQ methods take a delegate as a parameter, and it's very common to call methods within an anonymous method. But language features added to C# in the last 5 years can help if we can get people in the team to use new patterns, rather than copying the existing code. Anyway, my point is that chained methods on fluent interfaces on new lines makes understanding nested calls almost trivial, although it does need to be indented correctly.

Finally, there are people who think that having an [enforced max line length](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#max_line_length) is a good idea. For better or worse, [dotnet's format style rules doesn't support it](https://github.com/dotnet/format/blob/main/docs/Supported-.editorconfig-options.md?plain=1#L13), but [resharper has had its own setting for years](https://learn.microsoft.com/en-us/answers/questions/1007988/line-length-enforcement-for-visual-studio-and-dotn), which the team I worked for before joining Microsoft used.

Similarly, while I don't, some people code on a monitor in vertical orientation, significantly reducing the horizontal space. Needing to scroll left and right is much more harmful to readability. There are definitely times I'm tempted to switch to vertical, because most code is not so wide, and having more lines would be better.

And last note on this, the hardware provided to me when I go to the office is limited to 1920x1080. Even if I got a higher resolution monitor, I'd still need a laptop docking station capable of driving it. I'm not going to write lines that don't fit on the screen I'm writing it on. Although most of these examples in this PR would JUST fit on a single line (considering I have Solution Explorer open, eating some horizontal space). Unless I switch to vertical monitor orientation, in which case it wouldn't have a chance to fit.

If you made it this far, thank you for reading my essay.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
zivkan: I changed `ParseEnableValue` to use the same pattern as `ParseAuditMode` and `ParseAuditLevel`, and to run audit when the enable value is not explicit opt-out (so, run audit implicitly, when the value is invalid)

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
zivkan: We have behaviour differences between explicit opt-in and implicit opt-in, so no, that won't work. I think for telemetry we want to know the difference between "not set" and "invalid value" for implicit opt-in, so I'm keeping those different values (if we see a large number of customers with an invalid value, we might want to investigate why).

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -447,14 +447,19 @@ public static EnabledValue ParseEnableValue(string value)
             {
                 return EnabledValue.ExplicitOptOut;
             }
-            return EnabledValue.Undefined;
+
+            string messageText = string.Format(Strings.Error_InvalidNuGetAuditValue, value, "true, false");
+            RestoreLogMessage message = RestoreLogMessage.CreateError(NuGetLogCode.NU1014, messageText);
```
zivkan, line 452: I'm having second thoughts about invalid values for `NuGetAudit`, `NuGetAuditLevel` and `NuGetAuditMode` causing a restore error. Does anyone else think it should be just a warning instead, so customers can NoWarn it if they really want to keep the invalid value and use default behaviour?

aortiz-msft (dismissed):
🛳

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
nkolev92: In a lot of other features, bad values are even silently ignored. I don't mind this. 
Worst case scenario, people run into issues when using newer tooling and older tooling combiantions because the newer would recognize newer options and the older one wouldn't, but we've had discussions that alerting the users of such discrepancies might be ok in some cases.

src/NuGet.Core/NuGet.Commands/Strings.resx
```diff
@@ -1092,4 +1092,10 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
     <value>NuGetAudit is enabled, but no package sources contain known vulnerability data.</value>
     <comment>Do not translate NuGetAudit</comment>
   </data>
+  <data name="Error_InvalidNuGetAuditValue" xml:space="preserve">
+    <value>Invalid NuGetAudit value '{0}'. Expected values: {1}</value>
```
nkolev92: Should it be, allowed instead or valid values instead of expected?

Expected <-> Actual, Valid <-> Invalid is my thinking.

test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommandTests.cs
nkolev92: It was only a nit @zivkan, dang :D

nkolev92 (dismissed):

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -433,7 +433,7 @@ internal enum EnabledValue
 
         public static EnabledValue ParseEnableValue(string? value, string projectFullPath, ILogger logger)
         {
-            if (string.IsNullOrEmpty(value))
+            if (string.IsNullOrEmpty(value) || string.Equals(value, "default", StringComparison.OrdinalIgnoreCase))
```
aortiz-msft, line 436: Could we please add a UT that covers the "default" value?

aortiz-msft (dismissed):

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
zivkan: I really think it's bad UX silently ignoring bad inputs. How many issues in Home are from customers trying to do something, it not working as they expected, and there's zero feedback about why? It doesn't guide customers to the path of success.

I'll change this to a warning, so unfortunately I'll need another approval 😞

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
zivkan: There was in the same commit that made this change: 
* https://github.com/NuGet/NuGet.Client/pull/5347/commits/d5c96e674961de753d4a94857183f53c42a856ea#diff-a988ebe0a6bbbdbb0f848e0b2668aecc2652241d10d6d64e5a8c2e4335062c26R37
* https://github.com/NuGet/NuGet.Client/pull/5347/files#diff-a988ebe0a6bbbdbb0f848e0b2668aecc2652241d10d6d64e5a8c2e4335062c26R37

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -416,8 +416,8 @@ private NuGetAuditMode ParseAuditMode()
                 return NuGetAuditMode.All;
             }
 
-            string messageText = string.Format(Strings.Error_InvalidNuGetAuditModeValue, auditMode, "direct, all");
-            RestoreLogMessage message = RestoreLogMessage.CreateError(NuGetLogCode.NU1014, messageText);
+            string messageText = string.Format(Strings.Warning_InvalidNuGetAuditModeValue, auditMode, "direct, all");
+            RestoreLogMessage message = RestoreLogMessage.CreateWarning(NuGetLogCode.NU1906, messageText);
```
nkolev92: I see you changed your mind and added a warning. 
Can you elaborate on that? I liked the error myself. 

Also, NU1906 is not the correct number, it needs to be NU15XY since that's an input warning

aortiz-msft: Waiting on this to approve.

zivkan: I was trying to decide whether I thought it was worth potentially breaking customers builds (well, for those who don't use `TreatWarningsAsErrors`). Especially in the case where someone might already have the property defined for some reason. Or if in the futre we add new allowed values for one of the enums, but then a CI agent or a developer still using an old VS/.NET SDK restores the project.

Today, I'm leaning back towards error is better, so I reverted the change to make it a warning.

jeffkl (approved):

src/NuGet.Core/NuGet.Commands/Strings.resx
```diff
@@ -1092,4 +1092,10 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
     <value>NuGetAudit is enabled, but no package sources contain known vulnerability data.</value>
     <comment>Do not translate NuGetAudit</comment>
   </data>
+  <data name="Error_InvalidNuGetAuditValue" xml:space="preserve">
+    <value>Invalid NuGetAudit value '{0}'. Valid values: {1}</value>
+    <comment>Don't translate 'NuGetAudit'
+{0} - is the value from the customer's project file
+{1} is the list of valid values "true, false"</comment>
```
dtivel, line 1099: How is this better than making `true, false` part of the resource string with a comment not to translate `true` and `false`?

zivkan: Firstly, I lack confidence that [humans actually translate our strings](https://www.reddit.com/r/CasualUK/comments/12cwylk/microsoft_has_mistranslated_zip_files_as_postcode/?rdt=48612). We also have [evidence that the localization team doesn't see our comments](https://github.com/NuGet/NuGet.Build.Localization/pulls?q=is%3Apr+lock+is%3Aclosed). Apparently we have to do some stuff in the loc repo, and looking at those PR diffs, I can't figure out what I would need to do to lock the `true, false` strings correctly.

So, this way is better because 1. it doesn't require us to figure out [the bizzare syntax and tooling used by our loc repo](https://github.com/NuGet/NuGet.Build.Localization/pull/613/files). 2. It reduces (eliminates) risk that we do it wrong, and despite out best efforts, French customers get told to set `<NuGetAudit>faux</NuGetAudit>`.

nkolev92 (approved):

src/NuGet.Core/NuGet.Commands/Strings.resx
zivkan: We'll update in another PR if we can figure out the locking syntax. It's really unfortunate that the tooling is so difficult.

PR #5350, opened 2023-08-08
Display requested version when CPM enabled in dotnet list package

### LINKED ISSUE #12765: dotnet list package doesn't list requested versions when using CPM
### NuGet Product Used

dotnet.exe

### Product Version

8.0.0-preview.5.23280.8

### Worked before?

_No response_

### Impact

None

### Repro Steps & Context

On a project that uses CPM, `dotnet list package` (both format text and json have the same problem), the requested version is not listed:

```text
Project 'NuGet.PackageManagement.UI.Test' has the following package references
   [net472]:
   Top-level Package                                     Requested   Resolved
   > FluentAssertions                                                6.6.0
   > Microsoft.NET.Test.Sdk                                          17.1.0
   > Microsoft.VisualStudio.Sdk.TestFramework.Xunit                  17.2.7
   > Microsoft.VisualStudioEng.MicroBuild.Core                       1.0.0
   > Moq                                                             4.18.1
   > xunit                                                           2.4.2
   > xunit.runner.visualstudio                                       2.4.5
```

### Verbose Logs

_No response_

martinrrm:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12765

## Description

`dotnet list package` retrieves the information from the project or asset files, since CPM disables the ability of using the `Version` property in `PackageReference` tag, `Requested` version information is missing when doing the command.

This PR checks if the version is CentrallyManaged and populates the `Requested` information from the Directory.Packages.props file instead of the project file.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
@@ -750,8 +750,7 @@ namespace NuGet.CommandLine.XPlat
                         {
                             try
                             { // In case proj and assets file are not in sync and some refs were deleted
-                                var isPackageCentrallyManaged = tfmInformation.CentralPackageVersions.Any(cpv => cpv.Key.Equals(topLevelPackage.Name, StringComparison.Ordinal));
-                                if (isPackageCentrallyManaged)
+                                if (assetsFile.PackageSpec.RestoreMetadata.CentralPackageVersionsEnabled)
                                 {
                                     var packageCentralVersion = tfmInformation.CentralPackageVersions.Where(cpv => cpv.Key.Equals(topLevelPackage.Name, StringComparison.Ordinal)).First();
                                     installedPackage = new InstalledPackageReference(topLevelPackage.Name)
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
@@ -750,7 +750,19 @@ namespace NuGet.CommandLine.XPlat
                         {
                             try
                             { // In case proj and assets file are not in sync and some refs were deleted
-                                installedPackage = projPackages.Where(p => p.Name.Equals(topLevelPackage.Name, StringComparison.Ordinal)).First();
+                                var isPackageCentrallyManaged = tfmInformation.CentralPackageVersions.Any(cpv => cpv.Key.Equals(topLevelPackage.Name, StringComparison.Ordinal));
+                                if (isPackageCentrallyManaged)
+                                {
+                                    var packageCentralVersion = tfmInformation.CentralPackageVersions.Where(cpv => cpv.Key.Equals(topLevelPackage.Name, StringComparison.Ordinal)).First();
+                                    installedPackage = new InstalledPackageReference(topLevelPackage.Name)
+                                    {
+                                        OriginalRequestedVersion = packageCentralVersion.Value.VersionRange.MinVersion.ToString(),
+                                    };
+                                }
+                                else
+                                {
+                                    installedPackage = projPackages.Where(p => p.Name.Equals(topLevelPackage.Name, StringComparison.Ordinal)).First();
+                                }
                             }
                             catch (Exception)
                             {
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
@@ -756,13 +756,23 @@ namespace NuGet.CommandLine.XPlat
                                 // If the project is using CPM and it's not using VersionOverride, get the version from Directory.Package.props file
                                 if (assetsFile.PackageSpec.RestoreMetadata.CentralPackageVersionsEnabled && !projectPackage.IsVersionOverride)
                                 {
-                                    ProjectRootElement directoryBuildPropsRootElement = GetDirectoryBuildPropsRootElement(project);
-                                    ProjectItemElement packageInCPM = directoryBuildPropsRootElement.Items.Where(i => i.ItemType == PACKAGE_VERSION_TYPE_TAG && i.Include.Equals(topLevelPackage.Name, StringComparison.OrdinalIgnoreCase)).FirstOrDefault();
-
-                                    installedPackage = new InstalledPackageReference(topLevelPackage.Name)
+                                    if (topLevelPackage.VersionOverride?.OriginalString is not null)
                                     {
-                                        OriginalRequestedVersion = topLevelPackage.VersionOverride?.OriginalString ?? packageInCPM.Metadata.FirstOrDefault(i => i.Name.Equals("Version", StringComparison.OrdinalIgnoreCase)).Value,
-                                    };
+                                        installedPackage = new InstalledPackageReference(topLevelPackage.Name)
+                                        {
+                                            OriginalRequestedVersion = topLevelPackage.VersionOverride?.OriginalString,
+                                        };
+                                    }
+                                    else
+                                    {
+                                        ProjectRootElement directoryBuildPropsRootElement = GetDirectoryBuildPropsRootElement(project);
+                                        ProjectItemElement packageInCPM = directoryBuildPropsRootElement.Items.Where(i => (i.ItemType == PACKAGE_VERSION_TYPE_TAG || i.ItemType.Equals("GlobalPackageReference")) && i.Include.Equals(topLevelPackage.Name, StringComparison.OrdinalIgnoreCase)).FirstOrDefault();
+
+                                        installedPackage = new InstalledPackageReference(topLevelPackage.Name)
+                                        {
+                                            OriginalRequestedVersion = packageInCPM.Metadata.FirstOrDefault(i => i.Name.Equals("Version", StringComparison.OrdinalIgnoreCase)).Value,
+                                        };
+                                    }
                                 }
                                 else
                                 {
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
@@ -758,7 +758,7 @@ namespace NuGet.CommandLine.XPlat
                                         {
                                             installedPackage = new InstalledPackageReference(topLevelPackage.Name)
                                             {
-                                                OriginalRequestedVersion = packageCentralVersion.Value.VersionRange.MinVersion.ToString(),
+                                                OriginalRequestedVersion = topLevelPackage.VersionOverride?.MinVersion.ToString() ?? packageCentralVersion.Value.VersionRange.MinVersion.ToString(),
                                             };
                                             break;
                                         }
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
@@ -648,14 +648,14 @@ namespace NuGet.CommandLine.XPlat
         /// Prepares the dictionary that maps frameworks to packages top-level
         /// and transitive.
         /// </summary>
-        /// <param name="project"> Project </param>
+        /// <param name="project">Project to get the resoled versions from</param>
         /// <param name="userInputFrameworks">A list of frameworks</param>
         /// <param name="assetsFile">Assets file for all targets and libraries</param>
         /// <param name="transitive">Include transitive packages/projects in the result</param>
         /// <returns>FrameworkPackages collection with top-level and transitive package/project
         /// references for each framework, or null on error</returns>
         internal List<FrameworkPackages> GetResolvedVersions(
-            Project project, IEnumerable<string> userInputFrameworks, LockFile assetsFile, bool transitive, bool includeProjects)
+            Project project, IEnumerable<string> userInputFrameworks, LockFile assetsFile, bool transitive)
         {
             if (userInputFrameworks == null)
             {
@@ -728,7 +728,7 @@ namespace NuGet.CommandLine.XPlat
 
                 //The packages for the framework that were retrieved with GetRequestedVersions
                 var frameworkDependencies = tfmInformation.Dependencies;
-                var projPackages = GetPackageReferencesFromTargets(projectPath, tfmInformation.ToString());
+                var projectPackages = GetPackageReferencesFromTargets(projectPath, tfmInformation.ToString());
                 var topLevelPackages = new List<InstalledPackageReference>();
                 var transitivePackages = new List<InstalledPackageReference>();
 
@@ -751,9 +751,9 @@ namespace NuGet.CommandLine.XPlat
                         {
                             try
                             { // In case proj and assets file are not in sync and some refs were deleted
-                                var projectPackage = projPackages.Where(p => p.Name.Equals(topLevelPackage.Name, StringComparison.Ordinal)).First();
+                                var projectPackage = projectPackages.Where(p => p.Name.Equals(topLevelPackage.Name, StringComparison.Ordinal)).First();
 
-                                // If the project is using CPM and its not using VersionOverride, get the version from Directory.Package.props file
+                                // If the project is using CPM and it's not using VersionOverride, get the version from Directory.Package.props file
                                 if (assetsFile.PackageSpec.RestoreMetadata.CentralPackageVersionsEnabled && !projectPackage.IsVersionOverride)
                                 {
                                     ProjectRootElement directoryBuildPropsRootElement = GetDirectoryBuildPropsRootElement(project);
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
@@ -743,7 +743,7 @@ namespace NuGet.CommandLine.XPlat
                     if (matchingPackages.Any())
                     {
                         var topLevelPackage = matchingPackages.Single();
-                        InstalledPackageReference installedPackage;
+                        InstalledPackageReference installedPackage = default;
 
                         //If the package is not auto-referenced, get the version from the project file. Otherwise fall back on the assets file
                         if (!topLevelPackage.AutoReferenced)
@@ -752,11 +752,17 @@ namespace NuGet.CommandLine.XPlat
                             { // In case proj and assets file are not in sync and some refs were deleted
                                 if (assetsFile.PackageSpec.RestoreMetadata.CentralPackageVersionsEnabled)
                                 {
-                                    var packageCentralVersion = tfmInformation.CentralPackageVersions.Where(cpv => cpv.Key.Equals(topLevelPackage.Name, StringComparison.Ordinal)).First();
-                                    installedPackage = new InstalledPackageReference(topLevelPackage.Name)
+                                    foreach (KeyValuePair<string, CentralPackageVersion> packageCentralVersion in tfmInformation.CentralPackageVersions)
                                     {
-                                        OriginalRequestedVersion = packageCentralVersion.Value.VersionRange.MinVersion.ToString(),
-                                    };
+                                        if (packageCentralVersion.Key.Equals(topLevelPackage.Name, StringComparison.Ordinal))
+                                        {
+                                            installedPackage = new InstalledPackageReference(topLevelPackage.Name)
+                                            {
+                                                OriginalRequestedVersion = packageCentralVersion.Value.VersionRange.MinVersion.ToString(),
+                                            };
+                                            break;
+                                        }
+                                    }
                                 }
                                 else
                                 {
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
@@ -756,23 +756,13 @@ namespace NuGet.CommandLine.XPlat
                                 // If the project is using CPM and it's not using VersionOverride, get the version from Directory.Package.props file
                                 if (assetsFile.PackageSpec.RestoreMetadata.CentralPackageVersionsEnabled && !projectPackage.IsVersionOverride)
                                 {
-                                    if (topLevelPackage.VersionOverride?.OriginalString is not null)
-                                    {
-                                        installedPackage = new InstalledPackageReference(topLevelPackage.Name)
-                                        {
-                                            OriginalRequestedVersion = topLevelPackage.VersionOverride?.OriginalString,
-                                        };
-                                    }
-                                    else
-                                    {
-                                        ProjectRootElement directoryBuildPropsRootElement = GetDirectoryBuildPropsRootElement(project);
-                                        ProjectItemElement packageInCPM = directoryBuildPropsRootElement.Items.Where(i => (i.ItemType == PACKAGE_VERSION_TYPE_TAG || i.ItemType.Equals("GlobalPackageReference")) && i.Include.Equals(topLevelPackage.Name, StringComparison.OrdinalIgnoreCase)).FirstOrDefault();
+                                    ProjectRootElement directoryBuildPropsRootElement = GetDirectoryBuildPropsRootElement(project);
+                                    ProjectItemElement packageInCPM = directoryBuildPropsRootElement.Items.Where(i => (i.ItemType == PACKAGE_VERSION_TYPE_TAG || i.ItemType.Equals("GlobalPackageReference")) && i.Include.Equals(topLevelPackage.Name, StringComparison.OrdinalIgnoreCase)).FirstOrDefault();
 
-                                        installedPackage = new InstalledPackageReference(topLevelPackage.Name)
-                                        {
-                                            OriginalRequestedVersion = packageInCPM.Metadata.FirstOrDefault(i => i.Name.Equals("Version", StringComparison.OrdinalIgnoreCase)).Value,
-                                        };
-                                    }
+                                    installedPackage = new InstalledPackageReference(topLevelPackage.Name)
+                                    {
+                                        OriginalRequestedVersion = packageInCPM.Metadata.FirstOrDefault(i => i.Name.Equals("Version", StringComparison.OrdinalIgnoreCase)).Value,
+                                    };
                                 }
                                 else
                                 {
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
@@ -648,23 +648,23 @@ namespace NuGet.CommandLine.XPlat
         /// Prepares the dictionary that maps frameworks to packages top-level
         /// and transitive.
         /// </summary>
-        /// <param name="projectPath"> Path to the project to get versions for its packages </param>
+        /// <param name="project"> Project </param>
         /// <param name="userInputFrameworks">A list of frameworks</param>
         /// <param name="assetsFile">Assets file for all targets and libraries</param>
         /// <param name="transitive">Include transitive packages/projects in the result</param>
         /// <returns>FrameworkPackages collection with top-level and transitive package/project
         /// references for each framework, or null on error</returns>
         internal List<FrameworkPackages> GetResolvedVersions(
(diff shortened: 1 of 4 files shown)


src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
```diff
@@ -751,7 +751,19 @@ internal static bool IsPackageReferenceProject(Project project)
                         {
                             try
                             { // In case proj and assets file are not in sync and some refs were deleted
-                                installedPackage = projPackages.Where(p => p.Name.Equals(topLevelPackage.Name, StringComparison.Ordinal)).First();
+                                var isPackageCentrallyManaged = tfmInformation.CentralPackageVersions.Any(cpv => cpv.Key.Equals(topLevelPackage.Name, StringComparison.Ordinal));
```
nkolev92: You're using `Any` here and then you're using `Where`. It's unnecessary.

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
```diff
@@ -751,7 +751,19 @@ internal static bool IsPackageReferenceProject(Project project)
...
+                                if (isPackageCentrallyManaged)
```
nkolev92: All packages are expected to be centrally managed, unless there's `VersionOverride`. Not sure if that check adds something.

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
```diff
@@ -751,7 +751,19 @@ internal static bool IsPackageReferenceProject(Project project)
...
+                                {
+                                    var packageCentralVersion = tfmInformation.CentralPackageVersions.Where(cpv => cpv.Key.Equals(topLevelPackage.Name, StringComparison.Ordinal)).First();
+                                    installedPackage = new InstalledPackageReference(topLevelPackage.Name)
+                                    {
+                                        OriginalRequestedVersion = packageCentralVersion.Value.VersionRange.MinVersion.ToString(),
+                                    };
+                                }
+                                else
+                                {
+                                    installedPackage = projPackages.Where(p => p.Name.Equals(topLevelPackage.Name, StringComparison.Ordinal)).First();
```
nkolev92: Does `VersionOverride` work today?

nkolev92 (commented):
How does VersionOverride work?

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
martinrrm: Yess it works for CPM, in the Samples repo we have an example I have been using to test this

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
martinrrm: You're right, I can just check if the CPM is enabled. 

I did some testing and even if the user has `VersionOverride` it will appear as centrally managed since it's going to be declared in the Directory.Packages.props file too.

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
nkolev92: Does `dotnet list package` work with it? Does it display the correct version?

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
```diff
@@ -751,7 +751,18 @@ internal static bool IsPackageReferenceProject(Project project)
                         {
                             try
                             { // In case proj and assets file are not in sync and some refs were deleted
-                                installedPackage = projPackages.Where(p => p.Name.Equals(topLevelPackage.Name, StringComparison.Ordinal)).First();
+                                if (assetsFile.PackageSpec.RestoreMetadata.CentralPackageVersionsEnabled)
+                                {
+                                    var packageCentralVersion = tfmInformation.CentralPackageVersions.Where(cpv => cpv.Key.Equals(topLevelPackage.Name, StringComparison.Ordinal)).First();
```
aortiz-msft: >tfmInformation.CentralPackageVersions.Where(cpv => cpv.Key.Equals(topLevelPackage.Name, StringComparison.Ordinal)).First(); [](http://example.com/codeflow?start=64&length=123)

Wouldn't this first iterate through the entire collection to find all matches, instantiating a new collection for them, then return the first element of the new collection? A for loop where we break at the first match would be more efficient.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetListPackageTests.cs
```diff
@@ -84,6 +84,94 @@ public async Task DotnetListPackage_NoRestore_Fail()
             }
         }
 
+        [PlatformFact(Platform.Windows)]
```
aortiz-msft, line 86: Should we also add a test for a global reference?

aortiz-msft (changes requested):
:clock1:

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
martinrrm: This is an example, the version in Directory.Packages.props is 13.0.0 and the VersionOverride is 9.0.1
image

nkolev92: That's incorrect in that case. We need to have requested be `9.0.1`.

martinrrm: I see your point and I believe its valid. I was looking at it the other way. In this example: For me I know that I'm requesting 13.0.0 but I want that my `VersionOverride` version is the one that gets Resolved. 

@jeffkl Can you help us clarify the expected behavior here?

nkolev92: VersionOverride changes what you're requesting. 
It's basically overriding what the PackageVersion wanted.

martinrrm: Thanks for the clarification! I'll change the behavior of that

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetListPackageTests.cs
martinrrm: Talked with Jeff about GlobalReferences and the feature its implemented, I'll add tests for that

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
nkolev92: Maybe we can add a visual indicator that there's a version override?  
Similar to say "Auto Referenced packages". 

I'm just brainstorming here, you can wait for Jeff to confirm :)

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetListPackageTests.cs
aortiz-msft: I'm signing off now assuming these tests will be added before merging.

aortiz-msft (dismissed):

aortiz-msft (dismissed):
Thank you for adding the test for GlobalPackageReference.

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
```diff
@@ -744,14 +744,31 @@ internal static bool IsPackageReferenceProject(Project project)
...
                         if (!topLevelPackage.AutoReferenced)
                         {
                             try
                             { // In case proj and assets file are not in sync and some refs were deleted
-                                installedPackage = projPackages.Where(p => p.Name.Equals(topLevelPackage.Name, StringComparison.Ordinal)).First();
+                                if (assetsFile.PackageSpec.RestoreMetadata.CentralPackageVersionsEnabled)
+                                {
+                                    foreach (KeyValuePair<string, CentralPackageVersion> packageCentralVersion in tfmInformation.CentralPackageVersions)
+                                    {
+                                        if (packageCentralVersion.Key.Equals(topLevelPackage.Name, StringComparison.Ordinal))
```
dtivel: Does CPM require package ID's to be case sensitive?

Suppose in Directory.packages.props you have one casing (e.g.:  NuGet.Common), but in your project-level override you have a different casing (e.g.:  nuget.common).  Would CPM support this?  It looks like this line of code does not.

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
```diff
@@ -744,14 +744,31 @@ internal static bool IsPackageReferenceProject(Project project)
...
+                                        {
+                                            installedPackage = new InstalledPackageReference(topLevelPackage.Name)
+                                            {
+                                                OriginalRequestedVersion = topLevelPackage.VersionOverride?.MinVersion.ToString() ?? packageCentralVersion.Value.VersionRange.MinVersion.ToString(),
```
dtivel: The value of `VersionOverride` can be `null`, which you handled.  However, it looks like [`MinVersion`](https://github.com/NuGet/NuGet.Client/blob/216cdfabbeefef8ce52d16d424a1a3cbc8a262d0/src/NuGet.Core/NuGet.Versioning/VersionRangeBase.cs#L27) can also be `null`, which is why [`HasLowerBound`](https://github.com/NuGet/NuGet.Client/blob/216cdfabbeefef8ce52d16d424a1a3cbc8a262d0/src/NuGet.Core/NuGet.Versioning/VersionRangeBase.cs#L42) exists.

Is it not possible to specify a [version range](https://github.com/NuGet/NuGet.Client/blob/216cdfabbeefef8ce52d16d424a1a3cbc8a262d0/src/NuGet.Core/NuGet.Versioning/VersionRangeBase.cs#L27) for `VersionOverride` or is it assumed that no one will?

```XML
<PackageReference Include="abc" VersionOverride="[,2.0)" />
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
nkolev92: All package ids are case insensitive, so this is for sure a bug.

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
nkolev92: The rest of the code handles that by displaying the range. I'm guessing original version. We should follow that pattern.

```xml

  <PropertyGroup>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="NuGet.Common" Version="[,6.6.1]" />
    <PackageReference Include="NuGet.Commands" Version="6.5.0" />
    <PackageReference Include="NuGet.Protocol" Version="[6.5.0,)" />
  </ItemGroup>
</Project>

```

```console
PS C:\Code\Temp\DLP> dotnet list package 
Project 'DLP' has the following package references
   [net8.0]: 
   Top-level Package      Requested   Resolved
   > NuGet.Commands       6.5.0       6.5.0
   > NuGet.Common         [,6.6.1]    3.2.0
   > NuGet.Protocol       [6.5.0,)    6.5.0

```

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
nkolev92: Package ID is case insensitive, so this comparer needs updated.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetListPackageTests.cs
```diff
@@ -84,6 +84,144 @@ public async Task DotnetListPackage_NoRestore_Fail()
...
+        public async Task DotnetListPackage_WithCPM()
+        {
+            using (var pathContext = _fixture.CreateSimpleTestPathContext())
+            {
+                var projectA = XPlatTestUtils.CreateProject(ProjectName, pathContext, "net46");
+
+                var packageX = XPlatTestUtils.CreatePackage();
```
nkolev92: Readability would be improved if it's obvious the id/version that this method creates.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetListPackageTests.cs
```diff
@@ -84,6 +84,144 @@ public async Task DotnetListPackage_NoRestore_Fail()
...
+                File.WriteAllText(Path.Combine(pathContext.SolutionRoot, ProjectName, string.Concat(ProjectName, ".csproj")), projectContent);
+
+                _fixture.RunDotnetExpectSuccess(Path.Combine(pathContext.SolutionRoot, projectA.ProjectName),
+                    $"restore {projectA.ProjectName}.csproj");
+
+                CommandRunnerResult listResult = _fixture.RunDotnetExpectSuccess(Directory.GetParent(projectA.ProjectPath).FullName,
+                    $"list {projectA.ProjectPath} package");
+
+                // Assert Requested version is 2.0.0, but was override by VersionOverride tag
+                Assert.True(ContainsIgnoringSpaces(listResult.AllOutput, "2.0.0"));
```
nkolev92: I feel like this is incorrect. 

VersionOverride overrides the requested version and the command should reflect.

nkolev92 (changes requested):

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
martinrrm: You're right, I was under the impression that Ranges weren't allowed, floating are not allowed. 

I'll have to investigate this more, I was using `MinVersion` because `VersionRange.OriginalString` is always a range even for if you're not using one.

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
dtivel: Does [this existing statement](https://github.com/NuGet/NuGet.Client/pull/5350/files#diff-90c9045bb743a57a0e781f15088f964285ff4adf24e006ed4cd3a68d7e520674L754) have a similar bug?  With @martinrrm's changes, the line moved down a bit, but it's still intact.  I didn't scrutinize the code closely, but since we're talking about case insensitivity correctness, the surrounding code is up for discussion.  Not sure why it was case-sensitive to begin with.

nkolev92: Yeah, I think that's a bug too.

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
aortiz-msft: IIRC, CPM does not allow ranges overall because the intent is to give repo maintainers full control over which dependencies are used..

jeffkl (dismissed):

martinrrm:
I'm still trying to fix the tests 😢 I'm not 100% sure why its failing.

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
```diff
@@ -648,30 +648,31 @@ internal static bool IsPackageReferenceProject(Project project)
         /// Prepares the dictionary that maps frameworks to packages top-level
         /// and transitive.
         /// </summary>
-        /// <param name="projectPath"> Path to the project to get versions for its packages </param>
+        /// <param name="project"> Project </param>
         /// <param name="userInputFrameworks">A list of frameworks</param>
         /// <param name="assetsFile">Assets file for all targets and libraries</param>
         /// <param name="transitive">Include transitive packages/projects in the result</param>
```
aortiz-msft, line 654: Missing new param documentation.

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
```diff
@@ -648,30 +648,31 @@ internal static bool IsPackageReferenceProject(Project project)
...
         /// <returns>FrameworkPackages collection with top-level and transitive package/project
         /// references for each framework, or null on error</returns>
         internal List<FrameworkPackages> GetResolvedVersions(
-            string projectPath, IEnumerable<string> userInputFrameworks, LockFile assetsFile, bool transitive)
+            Project project, IEnumerable<string> userInputFrameworks, LockFile assetsFile, bool transitive, bool includeProjects)
         {
```
aortiz-msft, line 659: Where is IncludeProjects used?

aortiz-msft:
>                 var projPackages = GetPackageReferencesFromTargets(projectPath, tfmInformation.ToString());

I know this is pre-existing, but we shouldn't abbreviate in variable names. Would you please rename to projectPackages?

---
Refers to: src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs:731 in 345ea37. [](commit_id = 345ea37ea6ca7730391f550d3e887e9d0164cb7b, deletion_comment = False)

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
```diff
@@ -743,14 +744,30 @@ internal static bool IsPackageReferenceProject(Project project)
...
-                        InstalledPackageReference installedPackage;
+                        InstalledPackageReference installedPackage = default;
 
                         //If the package is not auto-referenced, get the version from the project file. Otherwise fall back on the assets file
                         if (!topLevelPackage.AutoReferenced)
                         {
                             try
                             { // In case proj and assets file are not in sync and some refs were deleted
-                                installedPackage = projPackages.Where(p => p.Name.Equals(topLevelPackage.Name, StringComparison.Ordinal)).First();
+                                var projectPackage = projPackages.Where(p => p.Name.Equals(topLevelPackage.Name, StringComparison.Ordinal)).First();
```
aortiz-msft, line 754: >var projectPackage = projPackages.Where(p => p.Name.Equals(topLevelPackage.Name, StringComparison.Ordinal)).First(); [](http://example.com/codeflow?start=32&length=116)

Why not use a foreach and break at the first result instead?

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
```diff
@@ -743,14 +744,30 @@ internal static bool IsPackageReferenceProject(Project project)
...
+
+                                // If the project is using CPM and its not using VersionOverride, get the version from Directory.Package.props file
```
aortiz-msft: >its [](http://example.com/codeflow?start=67&length=3)

typo: replace "its" with "it's"

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
```diff
@@ -743,14 +744,30 @@ internal static bool IsPackageReferenceProject(Project project)
...
+                                if (assetsFile.PackageSpec.RestoreMetadata.CentralPackageVersionsEnabled && !projectPackage.IsVersionOverride)
+                                {
+                                    ProjectRootElement directoryBuildPropsRootElement = GetDirectoryBuildPropsRootElement(project);
+                                    ProjectItemElement packageInCPM = directoryBuildPropsRootElement.Items.Where(i => i.ItemType == PACKAGE_VERSION_TYPE_TAG && i.Include.Equals(topLevelPackage.Name, StringComparison.OrdinalIgnoreCase)).FirstOrDefault();
+
+                                    installedPackage = new InstalledPackageReference(topLevelPackage.Name)
+                                    {
+                                        OriginalRequestedVersion = topLevelPackage.VersionOverride?.OriginalString ?? packageInCPM.Metadata.FirstOrDefault(i => i.Name.Equals("Version", StringComparison.OrdinalIgnoreCase)).Value,
```
aortiz-msft: >topLevelPackage.VersionOverride?.OriginalStrin [](http://example.com/codeflow?start=67&length=46)

Would doing this check at the beginning save us work from lines 757 - 760?

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
```diff
@@ -858,10 +875,15 @@ private static IEnumerable<InstalledPackageReference> GetPackageReferencesFromTa
             newProject.Build(new[] { CollectPackageReferences }, new List<Microsoft.Build.Framework.ILogger> { }, out var targetOutputs);
 
             return targetOutputs.First(e => e.Key.Equals(CollectPackageReferences, StringComparison.OrdinalIgnoreCase)).Value.Items.Select(p =>
-                new InstalledPackageReference(p.ItemSpec)
+            {
+                var isVersionOverride = p.GetMetadata("VersionOverride") != string.Empty;
```
aortiz-msft, line 879: >VersionOverride [](http://example.com/codeflow?start=55&length=15)

I'm a bit concerned that these properties are not defined centrally somewhere. Would you please create a new issue for this?

martinrrm: I created the issue https://github.com/NuGet/Home/issues/12865, assigned to myself for next quality week.

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
```diff
@@ -743,14 +744,40 @@ internal static bool IsPackageReferenceProject(Project project)
...
 
                         //If the package is not auto-referenced, get the version from the project file. Otherwise fall back on the assets file
                         if (!topLevelPackage.AutoReferenced)
                         {
                             try
                             { // In case proj and assets file are not in sync and some refs were deleted
-                                installedPackage = projPackages.Where(p => p.Name.Equals(topLevelPackage.Name, StringComparison.Ordinal)).First();
+                                var projectPackage = projectPackages.Where(p => p.Name.Equals(topLevelPackage.Name, StringComparison.Ordinal)).First();
+
+                                // If the project is using CPM and it's not using VersionOverride, get the version from Directory.Package.props file
```
nkolev92, line 756: I feel like this is contradicting line 759 https://github.com/NuGet/NuGet.Client/pull/5350/files#diff-90c9045bb743a57a0e781f15088f964285ff4adf24e006ed4cd3a68d7e520674R759 which is handling verison overide cases.

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
```diff
@@ -743,14 +744,40 @@ internal static bool IsPackageReferenceProject(Project project)
...
+                                        ProjectItemElement packageInCPM = directoryBuildPropsRootElement.Items.Where(i => (i.ItemType == PACKAGE_VERSION_TYPE_TAG || i.ItemType.Equals("GlobalPackageReference")) && i.Include.Equals(topLevelPackage.Name, StringComparison.OrdinalIgnoreCase)).FirstOrDefault();
+
+                                        installedPackage = new InstalledPackageReference(topLevelPackage.Name)
+                                        {
+                                            OriginalRequestedVersion = packageInCPM.Metadata.FirstOrDefault(i => i.Name.Equals("Version", StringComparison.OrdinalIgnoreCase)).Value,
+                                        };
+                                    }
+                                }
+                                else
+                                {
```
nkolev92, line 768: This gets the version from versionoverride right?

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetListPackageTests.cs
```diff
@@ -83,6 +83,211 @@ public async Task DotnetListPackage_NoRestore_Fail()
...
+                    $"list {projectA.ProjectPath} package");
+
+                // Assert Requested version is 0.1.0, but 1.0.0 was resolved
+                Assert.True(ContainsIgnoringSpaces(listResult.AllOutput, "0.1.0"));
+                Assert.True(ContainsIgnoringSpaces(listResult.AllOutput, "1.0.0"));
+            }
+        }
+
+        [PlatformFact(Platform.Windows)]
+        public async Task DotnetListPackage_VersionRanges_WithCPM()
```
nkolev92, line 87: This test and the previous are technically the same so you might be able to use a Theory to avoid duplication. 

FWIW, I'm not sure how relevant it is to have both, having this one is probably more than sufficient.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetListPackageTests.cs
```diff
@@ -83,6 +83,211 @@ public async Task DotnetListPackage_NoRestore_Fail()
...
+                    $"list {projectA.ProjectPath} package");
+
+                // Assert Requested version is [0.1.0,), but 1.0.0 was resolved
+                Assert.True(ContainsIgnoringSpaces(listResult.AllOutput, "[0.1.0,)"));
+                Assert.True(ContainsIgnoringSpaces(listResult.AllOutput, "1.0.0"));
+            }
+        }
+
+        [PlatformFact(Platform.Windows)]
+        public async Task DotnetListPackage_WithCPM_WithOverrideVersion()
```
nkolev92:
```suggestion
public async Task DotnetListPackage_WithCPM_WithVersionOverride()
```

Correctly naming the test will help find tests with for VersionOverride easier.

nkolev92 (commented):
Looks great. 

1 question and some suggestions for the tests.

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
zivkan: FYI, I use github's website to look at PR comments, but I believe that you use VSCode. This is what I see: 
image

I assume you're referring to the code UNDER the line the comment is on, but github shows the 4 lines of code above. Your comment doesn't make sense with the info that github provides, so I'd like to ask you to please adapt to how github presents PR comments, so that it makes more sense for reviewers.

jeffkl (dismissed):

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
nkolev92: I was referring to the code below yeah. 

I'll try to remember that for the future.

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
martinrrm: Removing lines 759 and its corresponding if block, VersionOverride is already handled like you pointed out below

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs
```diff
@@ -648,30 +648,31 @@ internal static bool IsPackageReferenceProject(Project project)
         /// Prepares the dictionary that maps frameworks to packages top-level
         /// and transitive.
         /// </summary>
-        /// <param name="projectPath"> Path to the project to get versions for its packages </param>
+        /// <param name="project">Project to get the resoled versions from</param>
```
jeffkl, line 651:
```suggestion
        /// <param name="project">Project to get the resolved versions from</param>
```

jeffkl (approved):

nkolev92 (approved):

PR #5351, opened 2023-08-09
Add ArgumentException for PackageDownload missing version

### LINKED ISSUE #12212: A PackageDownload without a version causes a NullReferenceException
### NuGet Product Used

dotnet.exe, MSBuild.exe

### Product Version

All

### Worked before?

_No response_

### Impact

### Repro Steps & Context

1. Add a `<PackageDownload />` item with no version
2. Restore from the command-line
3. Get error

```
C:\Program Files\Visual Studio 2022 Preview\Common7\IDE\CommonExtensions\Microsoft\NuGet\NuGet.targets(132,5):
     error : Object reference not set to an instance of an object.
```

Sample project:
```xml
  <PropertyGroup>
  </PropertyGroup>
  <ItemGroup>
    <PackageDownload Include="NuGet.Commands" />
  </ItemGroup>
</Project>
```

The exception is thrown here: 
https://github.com/NuGet/NuGet.Client/blob/9f4941056c6b10c4f8c42bf680bbe8ce7316fa2b/src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs#L673

We should log a better error or prevent this from happening in the [GetRestorePackageDownloadsTask](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Build.Tasks/GetRestorePackageDownloadsTask.cs)

### Verbose Logs

_No response_

Nigusu-Allehu:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12212 

## Description

Adds an ArgumentException . The Exception is raised whenever we have a PackageDownload without a version.
## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


nkolev92 (commented):
The change seems reasonable. A few other comments: 

- We need a test. This should be something that's easily testable. 
- We have a few different implementations of restore. Commandline restore can be run 2 ways, the standard way + /p:RestoreUseStaticGraphEvaluation="true" and Visual Studio. 

I think it's a great idea to verify they are all consistent.

Nigusu-Allehu:
> The change seems reasonable. A few other comments:
> 
> * We need a test. This should be something that's easily testable.
> * We have a few different implementations of restore. Commandline restore can be run 2 ways, the standard way + /p:RestoreUseStaticGraphEvaluation="true" and Visual Studio.
> 
> I think it's a great idea to verify they are all consistent.

Thank you for the feedback. I have added a test case and made sure the errors are consistent.

src/NuGet.Clients/NuGet.SolutionRestoreManager/Resources.resx
```diff
@@ -180,4 +180,4 @@ To prevent NuGet from restoring packages during build, open the Visual Studio Op
   <data name="SolutionIsNotSaved" xml:space="preserve">
     <value>Solution is not saved. Please save your solution before managing NuGet packages.</value>
   </data>
-</root>
\ No newline at end of file
+</root>
```
aortiz-msft: It looks like only an empty line was added to this file. Please revert if that's the case.

src/NuGet.Clients/NuGet.SolutionRestoreManager/Resources.Designer.cs
```diff
@@ -1,4 +1,4 @@
-﻿//------------------------------------------------------------------------------
```
aortiz-msft: It doesn't look like there were meaningful changes to this file. Please revert if that's the case.

src/NuGet.Clients/NuGet.SolutionRestoreManager/Resources.resx
```diff
@@ -180,4 +180,4 @@ To prevent NuGet from restoring packages during build, open the Visual Studio Op
   <data name="SolutionIsNotSaved" xml:space="preserve">
     <value>Solution is not saved. Please save your solution before managing NuGet packages.</value>
   </data>
-</root>
```
aortiz-msft: It looks like only an empty line was added to this file. Please revert if that's the case.

src/NuGet.Core/NuGet.Build.Tasks/Strings.Designer.cs
```diff
@@ -1,4 +1,4 @@
-﻿//------------------------------------------------------------------------------
+//------------------------------------------------------------------------------
```
aortiz-msft: Same.

aortiz-msft (dismissed):

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
```diff
@@ -676,6 +676,10 @@ internal static void AddPackageDownloads(PackageSpec spec, IEnumerable<IMSBuildI
             {
                 var id = item.GetProperty("Id");
                 var versionString = item.GetProperty("VersionRange");
+                if (versionString == null)
```
dtivel:
```suggestion
                if (string.IsNullOrEmpty(versionString))
```

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
```diff
@@ -676,6 +676,10 @@ internal static void AddPackageDownloads(PackageSpec spec, IEnumerable<IMSBuildI
...
+                {
+                    throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_PackageDownload_OnlyExactVersionsAreAllowed, ""));
```
dtivel: I think this can be more useful/actionable than it currently is.

Consider the original output:

```
C:\Program Files\Visual Studio 2022 Preview\Common7\IDE\CommonExtensions\Microsoft\NuGet\NuGet.targets(132,5):
     error : Object reference not set to an instance of an object.
```

Your change will result in the following:

```
C:\Program Files\Visual Studio 2022 Preview\Common7\IDE\CommonExtensions\Microsoft\NuGet\NuGet.targets(132,5):
     error : '' is not an exact version like '[1.0.0]'. Only exact versions are allowed with PackageDownload.
```

_Ideally_, the error message would be something like:

```
The PackageDownload reference for package ID 'NuGet.Commands' in file MyProject.csproj (line 34) has no version.  An exact version (e.g.:  [1.0.0]) is required.
```

Then, it's actionable.  A user could easily open the right file, go to the right line, and apply the right fix.

I don't expect this fix to do all of that, to plumb through file and line locations.  However, we _can_ improve the error message by providing the package ID, which I think [we already have](https://github.com/NuGet/NuGet.Client/pull/5351/files#diff-69bcc4cc18dd020ef093e419c357cd2a3452094ec27be239a03c665b32de863bL677).  That is searchable.

I recommend a new resource string like so:

```
The PackageDownload reference for package ID 'NuGet.Commands' has no version.  An exact version (e.g.:  [1.0.0]) is required.
```

test/NuGet.Core.Tests/NuGet.Commands.Test/MSBuildRestoreUtilityTests.cs
```diff
@@ -3398,6 +3398,27 @@ public void MSBuildRestoreUtility_AddPackageDownloads_InvalidVersionRange(string
             // Act & Assert
             var exception = Assert.Throws<ArgumentException>(() => MSBuildRestoreUtility.AddPackageDownloads(spec, msbuildItems));
         }
+        [Fact]
```
dtivel, line 3403: Add exactly 1 empty line between methods.

```suggestion

        [Fact]
```

test/NuGet.Core.Tests/NuGet.Commands.Test/MSBuildRestoreUtilityTests.cs
```diff
@@ -3398,6 +3398,27 @@ public void MSBuildRestoreUtility_AddPackageDownloads_InvalidVersionRange(string
...
+        public void MSBuildRestoreUtility_AddPackageDownloads_NoVersion()
+        {
+            // Arrange
+            var spec = MSBuildRestoreUtility.GetPackageSpec(new[]
```
dtivel: See coding style rule 10 [here](https://github.com/NuGet/NuGet.Client/blob/dev/docs/coding-guidelines.md#c-coding-style).

Throughout this method avoid `var` unless the data type is really obvious.  Replace `var` on this line and lines 3414, 3420.

test/NuGet.Core.Tests/NuGet.Commands.Test/MSBuildRestoreUtilityTests.cs
```diff
@@ -3398,6 +3398,27 @@ public void MSBuildRestoreUtility_AddPackageDownloads_InvalidVersionRange(string
...
+            packageX.Setup(p => p.GetProperty("Type")).Returns("DownloadDependency");
+            packageX.Setup(p => p.GetProperty("Id")).Returns("x");
+
+            var msbuildItems = new[]
+            {
+                packageX.Object
+            };
+
+            // Act & Assert
+            var exception = Assert.Throws<ArgumentException>(() => MSBuildRestoreUtility.AddPackageDownloads(spec, msbuildItems));
```
dtivel:
```suggestion
            ArgumentException exception = Assert.Throws<ArgumentException>(() => MSBuildRestoreUtility.AddPackageDownloads(spec, msbuildItems));
            string expectedMessage = string.Format(CultureInfo.CurrentCulture, Strings.Error_PackageDownload_MissingVersion, packageId);
            Assert.Equal(expectedMessage, exception.Message);
```

test/NuGet.Core.Tests/NuGet.Commands.Test/MSBuildRestoreUtilityTests.cs
```diff
@@ -3398,6 +3398,27 @@ public void MSBuildRestoreUtility_AddPackageDownloads_InvalidVersionRange(string
...
+        {
+            // Arrange
+            var spec = MSBuildRestoreUtility.GetPackageSpec(new[]
+            {
+                CreateItems(new Dictionary<string, string>())
+            });
+
+            var packageX = new Mock<IMSBuildItem>();
+            packageX.Setup(p => p.GetProperty("Type")).Returns("DownloadDependency");
+            packageX.Setup(p => p.GetProperty("Id")).Returns("x");
```
dtivel:
```suggestion
            const string packageId = "x";
            packageX.Setup(p => p.GetProperty("Type")).Returns("DownloadDependency");
            packageX.Setup(p => p.GetProperty("Id")).Returns(packageId);
```

dtivel (changes requested):

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
aortiz-msft: +1 to what was mentioned by @dtivel above.

src/NuGet.Clients/NuGet.SolutionRestoreManager/Resources.resx
aortiz-msft: @Nigusu-Allehu  - There's still an unnecessary empty line here.

aortiz-msft (changes requested):

src/NuGet.Clients/NuGet.SolutionRestoreManager/Resources.resx
```diff
@@ -124,8 +124,8 @@
     <value>Error occurred while restoring NuGet packages: {0}</value>
   </data>
   <data name="Error_PackageDownload_OnlyExactVersionsAreAllowed" xml:space="preserve">
-    <value>'{0}' is not an exact version like '[1.0.0]'. Only exact versions are allowed with PackageDownload.</value>
-    <comment>0 - the version string that's not exact</comment>
+    <value>In the package '{1}', '{0}' is not an exact version like '[1.0.0]'. Only exact versions are allowed with PackageDownload.</value>
```
dtivel: I like that you included the package ID.

It's unusual that arguments are not in ordinal order (e.g.:  `...{1}...{0}...` vs. `...{0}...{1}...`), but I won't lose sleep over it.

The wording needs improvement.  "In the package" is incorrect and potentially misleading.  I already suggested something [here](https://github.com/NuGet/NuGet.Client/pull/5351#discussion_r1291893217), but working with what you have, the simplest possible change that makes sense might be "For package '{1}', '{0}' is not..."

Maybe others have a better suggestion.

src/NuGet.Clients/NuGet.SolutionRestoreManager/Resources.resx
```diff
@@ -124,8 +124,8 @@
     <value>Error occurred while restoring NuGet packages: {0}</value>
   </data>
   <data name="Error_PackageDownload_OnlyExactVersionsAreAllowed" xml:space="preserve">
-    <value>'{0}' is not an exact version like '[1.0.0]'. Only exact versions are allowed with PackageDownload.</value>
-    <comment>0 - the version string that's not exact</comment>
+    <value>For package '{1}', '{0}' is not an exact version like '[1.0.0]'. Only exact versions are allowed with PackageDownload.</value>
```
nkolev92: For package sounds off to me. 
How about:

> Package '{1}' '{0}' does not have an exact version like '[1.0.0]'. Only exact versions are allowed with PackageDownload.

src/NuGet.Clients/NuGet.SolutionRestoreManager/Resources.resx
```diff
@@ -124,7 +124,7 @@
     <value>Error occurred while restoring NuGet packages: {0}</value>
   </data>
   <data name="Error_PackageDownload_OnlyExactVersionsAreAllowed" xml:space="preserve">
-    <value>For package '{1}', '{0}' is not an exact version like '[1.0.0]'. Only exact versions are allowed with PackageDownload.</value>
```
nkolev92: `Package is` sounds off to me. 
I prefer the package does not suggestion.

You can wait for feedback from others though. maybe they have better ideas thna mine.

Nigusu-Allehu: If for example we are trying to PackageDownload "Nuget.Commands" with not an exact version like "v1", what we would get is `For package 'Nuget.Commands', 'v1' is not an exact version like '[1.0.0]'....` . I think you are right `Package 'Nuget.Commands' 'v1' does not have an exact version like [1.0.0]` sounds better.

nkolev92: Another idea would be:

> Package 'Nuget.Commands v1' does not have an exact version like [1.0.0]

' ' is not allowed in either version or id, so it's good separator.

nkolev92 (dismissed):

test/NuGet.Core.Tests/NuGet.Build.Tasks.Console.Test/MSBuildStaticGraphRestoreTests.cs
```diff
@@ -126,7 +127,33 @@ public void GetPackageDownloads_WhenNotExactVersion_ThrowsException(string versi
                     var _ = MSBuildStaticGraphRestore.GetPackageDownloads(project).ToList();
                 };
 
-                act.Should().Throw<ArgumentException>().WithMessage($"'{expected ?? VersionRange.Parse(version).OriginalString}' is not an exact version like '[1.0.0]'. Only exact versions are allowed with PackageDownload.");
+                act.Should().Throw<ArgumentException>().WithMessage($"Package 'PackageA {expected ?? VersionRange.Parse(version).OriginalString}' does not have an exact version like '[1.0.0]'. Only exact versions are allowed with PackageDownload.");
```
nkolev92:
```suggestion
                act.Should().Throw<ArgumentException>().WithMessage($"Package 'PackageA {expected ?? version}' does not have an exact version like '[1.0.0]'. Only exact versions are allowed with PackageDownload.");
```

I think the parsing is redundant if you'll use OriginalString.
It's supposed to give you version anyways.

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
```diff
@@ -214,14 +214,19 @@ internal static IEnumerable<DownloadDependency> GetPackageDownloads(IMSBuildProj
                 string id = projectItemInstance.Identity;
 
                 // PackageDownload items can contain multiple versions
-                foreach (var version in MSBuildStringUtility.Split(projectItemInstance.GetProperty("Version")))
+                string versionRanges = projectItemInstance.GetProperty("Version");
+                if (string.IsNullOrEmpty(versionRanges))
+                {
+                    throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_PackageDownload_OnlyExactVersionsAreAllowed, "", id));
+                }
+                foreach (var version in MSBuildStringUtility.Split(versionRanges))
```
aortiz-msft: There should be an empty line after a closing bracket.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
```diff
@@ -676,6 +676,10 @@ internal static void AddPackageDownloads(PackageSpec spec, IEnumerable<IMSBuildI
             {
                 var id = item.GetProperty("Id");
                 var versionString = item.GetProperty("VersionRange");
+                if (string.IsNullOrEmpty(versionString))
+                {
+                    throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_PackageDownload_OnlyExactVersionsAreAllowed, "", id));
+                }
                 var versions = versionString.Split(splitChars, StringSplitOptions.RemoveEmptyEntries);
```
aortiz-msft: same

test/NuGet.Clients.Tests/NuGet.SolutionRestoreManager.Test/VsSolutionRestoreServiceTests.cs
```diff
@@ -2387,6 +2388,44 @@ public void ToPackageSpec_TwoProjectReferencesToSameProject_DeduplicatesProjectR
         private delegate void TryGetProjectNamesCallback(string projectPath, out ProjectNames projectNames);
         private delegate bool TryGetProjectNamesReturns(string projectPath, out ProjectNames projectNames);
 
+        [Fact]
+        public void ToPackageSpec_NoVersion_ThrowsException()
+        {
```
aortiz-msft, line 2393: Is this the simplest scenario to validate against? Using the simplest possible scenario will make it easier for others to debug the test and understand what's going on if the test starts to fail.

test/NuGet.Clients.Tests/NuGet.SolutionRestoreManager.Test/VsSolutionRestoreServiceTests.cs
```diff
@@ -2387,6 +2388,34 @@ public void ToPackageSpec_TwoProjectReferencesToSameProject_DeduplicatesProjectR
...
+                        new VsReferenceProperty("Version", null)
+                    }))
+                    },
+                    frameworkReferences: emptyReferenceItems,
+                    projectProperties: Array.Empty<IVsProjectProperty>())
+            };
+            string expected = string.Format(CultureInfo.CurrentCulture, Resources.Error_PackageDownload_OnlyExactVersionsAreAllowed, "", packageName);
+
+            // Assert
+            ArgumentException exception = Assert.Throws<ArgumentException>(() => VsSolutionRestoreService.ToPackageSpec(projectName, targetFrameworks, originalTargetFrameworkstr: string.Empty, msbuildProjectExtensionsPath: string.Empty));
```
zivkan, line 2417: Since this throws an exception, it means that NuGet won't restore any other package in the project. If this project has never had a successful restore, it means that none of the packages this project references will have intellisense until this user-input error is fixed.

An alternative, is to use the "additional errors" code I added years ago. This means that `ToPackageSpec` should complete without throwing an exception (in this case, I'd leave the bad PackageDownload out of the PackageSpec that's returned), but the project adapter cache should have the error message pre-cached. Then, restore will automatically add the message as an error, but it allows restore to run to set up all the other packages in the project.

I don't know how other people in the team feel about this, but I think NuGet has really bad handling of input errors. Nobody wants a compiler that stops at the first error, without reporting all the other errors in the file, and I don't think Nuget should either.

But I know this would be a lot more complicated for someone just starting in the team to implement.

nkolev92: > An alternative, is to use the "additional errors" code I added years ago

This will use the AdditionalErrors code you added. 

https://github.com/NuGet/NuGet.Client/blob/fd095eca62b6d63ba5a6a68cfdfba63c23a243e1/src/NuGet.Clients/NuGet.SolutionRestoreManager/VsSolutionRestoreService.cs#L221-L242

> Since this throws an exception, it means that NuGet won't restore any other package in the project. If this project has never had a successful restore, it means that none of the packages this project references will have intellisense until this user-input error is fixed.

It's not quite the case with PackageDownload, but restore does not really work in a way where it allows for a partial restore. Any change in the top level packages can massively affect the graph. 

> I don't know how other people in the team feel about this, but I think NuGet has really bad handling of input errors. Nobody wants a compiler that stops at the first error, without reporting all the other errors in the file, and I don't think Nuget should either.

See above Re:Partial restore. 
I don't think this is worth it.

aortiz-msft:
@zivkan - Are you planning on reviewing the latest iteration?

nkolev92 (dismissed):

test/NuGet.Clients.Tests/NuGet.SolutionRestoreManager.Test/VsSolutionRestoreServiceTests.cs
zivkan: > Any change in the top level packages can massively affect the graph.

Due to how DependencyResolver works, we still extract all the same packages to the global packages folder. So, it's not really doing any work that wouldn't be done if all errors were fixed. But yes, I concede that the transitive packages can be very different.

> See above Re:Partial restore.
> I don't think this is worth it.

Even if we don't want to skip the packages and do a restore, I still wish that we thought about at least detecting all input errors and then fail the nomination (and equivalent in CLI restores), rather than bailing out on the first error. On my (now broken) Surface Book 2, it took MSBuild ~12 seconds to evaluate all the projects in NuGet.sln. If I have 2 input errors, and I have to wait 12+ seconds each time, rather than once and let me fix both errors, I feel frustrated.

zivkan:
> Are you planning on reviewing the latest iteration?

No, just commenting on other comments. Our product keeps optimizing for easy implementations, rather than good error experiences. I wish we'd consider better error handling.

Nigusu-Allehu:
Hey @dtivel @aortiz-msft - I was wondering if you are planning on reviewing the latest iteration.

test/NuGet.Clients.Tests/NuGet.SolutionRestoreManager.Test/VsSolutionRestoreServiceTests.cs
nkolev92: > I still wish that we thought about at least detecting all input errors and then fail the nomination (and equivalent in CLI restores), rather than bailing out on the first error

I don't mind this, I'm just not sure this is common enough to warrant the amount work it'd take :) 

Maybe you can create an issue about this? I think it's something worth tracking, and the best way to prove/disprove the sentiment.

src/NuGet.Clients/NuGet.SolutionRestoreManager/VSNominationUtilities.cs
```diff
@@ -509,7 +509,10 @@ private static IEnumerable<VersionRange> GetVersionRangeList(IVsReferenceItem it
         {
             char[] splitChars = new[] { ';' };
             string versionString = GetPropertyValueOrNull(item, "Version");
-
+            if (string.IsNullOrEmpty(versionString))
+            {
+                throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.Error_PackageDownload_OnlyExactVersionsAreAllowed, "", item.Name));
+            }
             if (versionString != null)
```
dtivel, line 517: 1 blank line after the curly brace and the next statement.

src/NuGet.Clients/NuGet.SolutionRestoreManager/VSNominationUtilities.cs
```diff
@@ -509,7 +509,10 @@ private static IEnumerable<VersionRange> GetVersionRangeList(IVsReferenceItem it
         {
             char[] splitChars = new[] { ';' };
             string versionString = GetPropertyValueOrNull(item, "Version");
-
+            if (string.IsNullOrEmpty(versionString))
+            {
+                throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.Error_PackageDownload_OnlyExactVersionsAreAllowed, "", item.Name));
```
dtivel: In this case and elsewhere, the error message for package ID NuGet.Common will be:

`Package 'NuGet.Common ' does not have an exact version like '[1.0.0]'. Only exact versions are allowed with PackageDownload.`

Note the trailing whitespace after the package ID.  It's unpolished.

From the start of this PR it is almost like there is an aversion to creating new resource strings.  Why not have 2 resource strings?

* one that takes package ID
* one that takes package ID and version

test/NuGet.Clients.Tests/NuGet.SolutionRestoreManager.Test/VsSolutionRestoreServiceTests.cs
```diff
@@ -2387,6 +2388,34 @@ public void ToPackageSpec_TwoProjectReferencesToSameProject_DeduplicatesProjectR
...
+                    projectReferences: emptyReferenceItems,
+                    packageDownloads: new List<IVsReferenceItem>
+                    { new VsReferenceItem(packageName, new VsReferenceProperties(new []
+                    {
+                        new VsReferenceProperty("Version", null)
+                    }))
+                    },
+                    frameworkReferences: emptyReferenceItems,
+                    projectProperties: Array.Empty<IVsProjectProperty>())
+            };
```
dtivel: Indentation is off.  Also, curly braces should both be on their own line or on the same line (if the body is small).

```suggestion
            {
                new VsTargetFrameworkInfo2(
                    targetFrameworkMoniker: "net5.0",
                    packageReferences: emptyReferenceItems,
                    projectReferences: emptyReferenceItems,
                    packageDownloads: new List<IVsReferenceItem>
                    {
                        new VsReferenceItem(
                            packageName,
                            new VsReferenceProperties(new [] { new VsReferenceProperty("Version", value: null) }))
                    },
                    frameworkReferences: emptyReferenceItems,
                    projectProperties: Array.Empty<IVsProjectProperty>())
            };
```

test/NuGet.Core.Tests/NuGet.Build.Tasks.Console.Test/MSBuildStaticGraphRestoreTests.cs
```diff
@@ -126,7 +127,33 @@ public void GetPackageDownloads_WhenNotExactVersion_ThrowsException(string versi
                     var _ = MSBuildStaticGraphRestore.GetPackageDownloads(project).ToList();
                 };
 
-                act.Should().Throw<ArgumentException>().WithMessage($"'{expected ?? VersionRange.Parse(version).OriginalString}' is not an exact version like '[1.0.0]'. Only exact versions are allowed with PackageDownload.");
+                act.Should().Throw<ArgumentException>().WithMessage($"Package 'PackageA {expected ?? version}' does not have an exact version like '[1.0.0]'. Only exact versions are allowed with PackageDownload.");
```
dtivel: Based on feedback elsewhere, I think this is more robust:

```C#
string expectedMessage = string.Format(Strings.Error_PackageDownload_OnlyExactVersionsAreAllowed, expected ?? version, "PackageA");

act.Should().Throw<ArgumentException>().WithMessage(expectedMessage);
```

This way, the test's expected message isn't hard-coded to en-US strings.

You did this already in new tests, like [here](https://github.com/NuGet/NuGet.Client/pull/5351/files#diff-071c19f558be6f4bbcd2a336dcde0c31f3f56d38f013a92fed62e6f5206bd598R156).

dtivel:
@Nigusu-Allehu, I added more feedback.  Please ping when ready for re-review.

src/NuGet.Clients/NuGet.SolutionRestoreManager/VSNominationUtilities.cs
Nigusu-Allehu: @nkolev92 what do you think?

nkolev92: I don't mind having separate resource strings.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
Nigusu-Allehu: Added two Error strings

- one that takes package ID `Package 'PackageA' does not have an exact version like '[1.0.0]'. Only exact versions are allowed with PackageDownload.`
- one that takes package ID and version `Package 'PackageA 1.*' does not have an exact version like '[1.0.0]'. Only exact versions are allowed with PackageDownload.`

test/NuGet.Clients.Tests/NuGet.SolutionRestoreManager.Test/VsSolutionRestoreServiceTests.cs
Nigusu-Allehu: Since the agreement is not to make the updates in this PR, I have marked the conversations as resolved. Please let me know if you disagree.

src/NuGet.Clients/NuGet.SolutionRestoreManager/VSNominationUtilities.cs
Nigusu-Allehu: I have added separate resource strings..

test/NuGet.Core.Tests/NuGet.Build.Tasks.Console.Test/MSBuildStaticGraphRestoreTests.cs
Nigusu-Allehu: Thank you! I have removed the hard coded en-US strings.

dtivel (dismissed):

nkolev92 (dismissed):

aortiz-msft:
>                 Action act = () =>

We should not abbreviate variable names like this one. Is this an abbreviation for "actual", "action", something else?

---
Refers to: test/NuGet.Core.Tests/NuGet.Build.Tasks.Console.Test/MSBuildStaticGraphRestoreTests.cs:152 in 4c44a16. [](commit_id = 4c44a164f3f31feb971ca117c734f102f4c50931, deletion_comment = False)

aortiz-msft (dismissed):
:shipit:

aortiz-msft (dismissed):

src/NuGet.Clients/NuGet.SolutionRestoreManager/Resources.resx
```diff
@@ -124,8 +124,12 @@
     <value>Error occurred while restoring NuGet packages: {0}</value>
   </data>
   <data name="Error_PackageDownload_OnlyExactVersionsAreAllowed" xml:space="preserve">
-    <value>'{0}' is not an exact version like '[1.0.0]'. Only exact versions are allowed with PackageDownload.</value>
-    <comment>0 - the version string that's not exact</comment>
+    <value>Package '{1} {0}' does not have an exact version like '[1.0.0]'. Only exact versions are allowed with PackageDownload.</value>
```
donnie-msft: It'd be easier to read if this were 0, 1, in chronological order. Is there any reason for mapping the parameters as 1, 0?

test/NuGet.Core.Tests/NuGet.Build.Tasks.Console.Test/MSBuildStaticGraphRestoreTests.cs
```diff
@@ -125,8 +127,34 @@ public void GetPackageDownloads_WhenNotExactVersion_ThrowsException(string versi
...
+                    }
+                };
+
+                Action action = () =>
+                {
+                    var _ = MSBuildStaticGraphRestore.GetPackageDownloads(project).ToList();
+                };
 
-                act.Should().Throw<ArgumentException>().WithMessage($"'{expected ?? VersionRange.Parse(version).OriginalString}' is not an exact version like '[1.0.0]'. Only exact versions are allowed with PackageDownload.");
+                action.Should().Throw<ArgumentException>().WithMessage(string.Format(CultureInfo.CurrentCulture, Strings.Error_PackageDownload_NoVersion, packageName));
```
donnie-msft: Nice, you're also contributing towards making our tests localized instead of hardcoded to `en-us`!

test/NuGet.Core.Tests/NuGet.Commands.Test/MSBuildRestoreUtilityTests.cs
```diff
@@ -3399,6 +3400,27 @@ public void MSBuildRestoreUtility_AddPackageDownloads_InvalidVersionRange(string
             var exception = Assert.Throws<ArgumentException>(() => MSBuildRestoreUtility.AddPackageDownloads(spec, msbuildItems));
         }
 
+        [Fact]
+        public void MSBuildRestoreUtility_AddPackageDownloads_NoVersion_ThrowsException()
+        {
+            // Arrange
+            PackageSpec spec = MSBuildRestoreUtility.GetPackageSpec(new[] { CreateItems(new Dictionary<string, string>()) });
+            var packageX = new Mock<IMSBuildItem>();
+            const string packageId = "x";
```
donnie-msft, line 3409: Good idea making this `const`, I don't always remember to do that when writing tests.

donnie-msft (dismissed):

src/NuGet.Clients/NuGet.SolutionRestoreManager/Resources.resx
Nigusu-Allehu: You are right. I have updated it now.

Nigusu-Allehu:
I modified `>Package '{1} {0}' does not have an exact version like '[1.0.0]'. Only exact versions are allowed with PackageDownload.` so that the parameters are in chronological order `{0} {1}` to make it easier to read. @donnie-msft @dtivel thanks for the recommendation.

src/NuGet.Core/NuGet.Commands/Strings.resx
```diff
@@ -775,8 +775,12 @@ For more information, visit https://docs.nuget.org/docs/reference/command-line-r
     <comment>{0} is the path to the project file. {1} is the restore duration.</comment>
   </data>
   <data name="Error_PackageDownload_OnlyExactVersionsAreAllowed" xml:space="preserve">
-    <value>'{0}' is not an exact version like '[1.0.0]'. Only exact versions are allowed with PackageDownload.</value>
-    <comment>0 - the version string that's not exact</comment>
+    <value>Package '{0} {1}' does not have an exact version like '[1.0.0]'. Only exact versions are allowed with PackageDownload.</value>
+    <comment>1 - the version string that's not exact. 0 - The package that does not have an exact version. Do not translate 'PackageDownload'.</comment>
```
donnie-msft: nit: the comments could be chronological. Not sure if it could confuse localization team at all.

test/NuGet.Core.Tests/NuGet.Build.Tasks.Console.Test/MSBuildStaticGraphRestoreTests.cs
```diff
@@ -125,8 +127,34 @@ public void GetPackageDownloads_WhenNotExactVersion_ThrowsException(string versi
...
+                act.Should().Throw<ArgumentException>().WithMessage(expectedMessage);
+            }
+        }
+
+        [Fact]
+        public void GetPackageDownloads_NoVersion_ThrowsException()
+        {
+            string packageName = "PackageA";
+            using (var testDirectory = TestDirectory.Create())
+            {
```
donnie-msft: Consider dropping the block, which means less indentation...

```suggestion
            using var testDirectory = TestDirectory.Create();
```

test/NuGet.Core.Tests/NuGet.Commands.Test/MSBuildRestoreUtilityTests.cs
```diff
@@ -3399,6 +3400,27 @@ public void MSBuildRestoreUtility_AddPackageDownloads_InvalidVersionRange(string
...
+            packageX.Setup(p => p.GetProperty("Type")).Returns("DownloadDependency");
+            packageX.Setup(p => p.GetProperty("Id")).Returns(packageId);
+
+            var msbuildItems = new[]
```
donnie-msft: I'd prefer to see the type somehow: either an explicit type instead of var, or an array type declaration.

donnie-msft (dismissed):

src/NuGet.Core/NuGet.Commands/Strings.resx
Nigusu-Allehu: I agree. I have addressed it

donnie-msft (dismissed):

dtivel (dismissed):

donnie-msft (approved):

PR #5355, opened 2023-08-11
Automatically create source mappings on Install for `PackageReference` & show in Preview Window

### LINKED ISSUE #11366: Create Package Source Mappings during Installation/update through PM UI
Refer to https://github.com/NuGet/Home/issues/11365. 

This may require some clarity, but there aree quality mockups in https://github.com/NuGet/Home/blob/dev/implemented/2021/PackageSourceMapping.md

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Home/issues/11366
Spec: https://github.com/NuGet/Home/pull/12810

## Description

For PackageReference projects, when the top-level package doesn't have a source mapping, and a single source (not `All`) is selected, create a source mapping to that source for the top-level and any transitive dependencies that are brought in.
  - ie, the delta from the source mappings in the existing `nuget.config` and the source mappings resulting from this Install's PreviewRestore) 
  - If a transitive is not source mapped and not available on that source, the action will fail as it does today.
  - New status to indicate when this is going to happen, and an Enabled Install button:
image

- Preview Window, if enabled, will list the new mappings at the bottom of the list.
  - New section heading: "Solution". I chose this since Source Mapping applies to `nuget.config` which is a solution level concept.
    - ie, even in Project level PMUI, the text is "Solution"
image
image

- All projects affected must be `PackageReference`
  - eg, Packages.config behavior is unchanged.
  - Install button remains enabled & status indicates a required mapping:
image

## Summary of Conditions

Project type:
**Applicable**: Only `PackageReference` project(s) in scope
**Not Applicable**: Non-`PackageReference` project in scope

Selected source:
**Applicable**: Single source selected in Package Source dropdown.
**Not Applicable**: Aggregate source (`All`) selected in Package Source dropdown.

|State | Project Type (*) | Selected Source (*) | Change by PR|
|--|--|--|--|
|Source Mapping is off | - | - | None |
|Source Mapping is required | Not Applicable | - | None (must manually create mappings) |
|Source Mapping is required | Applicable | Not Applicable | None (must manually create mappings) |
|Source Mapping is required | Applicable | Applicable| Automatically map to selected source & show changes in Preview Window |
|Source Mapping found | Applicable | Applicable & selection is the mapped source | None (existing mapping remains) |
|Source Mapping found | Applicable | Applicable & selection is not the mapped source | Automatically create another mapping to selected source & show changes in Preview Window |

_(*) = see description above_

### Table: Behavior for Packages not Installed to the GPF

_Selected Source: A_

_Enabled Sources: A, B_

_Package is not already mapped_

Type|Package found on Source|Automatic Mapping Source| Implemented by PR?|
|--|--|--|--|
Direct|A|A|✅|
Transitive|A|A|✅|
Transitive|B|Error*|(✅)|

(*) Error responses:
- The existing error for NU1101 will be shown.

### Table: Behavior for Packages Installed to the GPF

_Selected Source: A_

_Enabled Sources: A, B_

_Package is not already mapped_

Type|GPF Source|Automatic Mapping Source| Implemented by PR?|
|--|--|--|--|
Direct|A|A| ✅|
Direct|B|A| ✅|
Direct|C|A| ✅|
Transitive|A|A| ✅|
Transitive|B|B| ❌ |
Transitive|C|Error*|❌|

(*) Error responses:
- The package is available in the GPF, but the source it came from is not one of your configured sources.

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
@@ -4,6 +4,7 @@
 #nullable enable
 
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Linq;
 
 namespace NuGet.PackageManagement.UI
@@ -16,7 +17,7 @@ namespace NuGet.PackageManagement.UI
 
         public IEnumerable<UpdatePreviewResult> Updated { get; }
 
-        public IReadOnlyDictionary<string, SortedSet<string>>? NewSourceMappings { get; }
+        public ImmutableDictionary<string, SortedSet<string>>? NewSourceMappings { get; }
 
         public string? Name { get; }
 
@@ -35,7 +36,7 @@ namespace NuGet.PackageManagement.UI
         public PreviewResult(Dictionary<string, SortedSet<string>>? newSourceMappings)
         {
             Name = Resources.Label_Solution;
-            NewSourceMappings = newSourceMappings;
+            NewSourceMappings = newSourceMappings?.ToImmutableDictionary();
             Added = Enumerable.Empty<AccessiblePackageIdentity>();
             Deleted = Enumerable.Empty<AccessiblePackageIdentity>();
             Updated = Enumerable.Empty<UpdatePreviewResult>();
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
@@ -19,10 +19,10 @@ namespace NuGet.PackageManagement.UI
 
         public ImmutableDictionary<string, SortedSet<string>>? NewSourceMappings { get; }
 
-        public string? Name { get; }
+        public string Name { get; }
 
         public PreviewResult(
-            string? projectName,
+            string projectName,
             IEnumerable<AccessiblePackageIdentity> added,
             IEnumerable<AccessiblePackageIdentity> deleted,
             IEnumerable<UpdatePreviewResult> updated)
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
@@ -1,7 +1,10 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+#nullable enable
+
 using System.Collections.Generic;
+using System.Linq;
 
 namespace NuGet.PackageManagement.UI
 {
@@ -13,10 +16,12 @@ namespace NuGet.PackageManagement.UI
 
         public IEnumerable<UpdatePreviewResult> Updated { get; }
 
-        public string Name { get; }
+        public IReadOnlyDictionary<string, SortedSet<string>>? NewSourceMappings { get; }
+
+        public string? Name { get; }
 
         public PreviewResult(
-            string projectName,
+            string? projectName,
             IEnumerable<AccessiblePackageIdentity> added,
             IEnumerable<AccessiblePackageIdentity> deleted,
             IEnumerable<UpdatePreviewResult> updated)
@@ -25,6 +30,15 @@ namespace NuGet.PackageManagement.UI
             Added = added;
             Deleted = deleted;
             Updated = updated;
+        }
+
+        public PreviewResult(Dictionary<string, SortedSet<string>>? newSourceMappings)
+        {
+            Name = Resources.Label_Solution;
+            NewSourceMappings = newSourceMappings;
+            Added = Enumerable.Empty<AccessiblePackageIdentity>();
+            Deleted = Enumerable.Empty<AccessiblePackageIdentity>();
+            Updated = Enumerable.Empty<UpdatePreviewResult>();
         }
     }
 }
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
@@ -1010,7 +1010,7 @@ namespace NuGet.PackageManagement.UI
                 }
 #pragma warning restore CS8600 // Converting null literal or possible null value to non-nullable type.
 
-                var result = new PreviewResult(projectName, added, deleted, updated);
+                var result = new PreviewResult(projectName!, added, deleted, updated);
 
                 results.Add(result);
             }
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
@@ -78,7 +78,7 @@ namespace NuGet.PackageManagement.UI
                     uiService.RemoveDependencies,
                     uiService.ForceRemove,
                     newMappingID: userAction.PackageId,
-                    newMappingSource: userAction.SourceMappingSourceName,
+                    newMappingSource: userAction.SelectedSourceName,
                     cancellationToken),
                 cancellationToken);
         }
@@ -383,7 +383,7 @@ namespace NuGet.PackageManagement.UI
                     TelemetryServiceUtility.StartOrResumeTimer();
 
                     IReadOnlyList<ProjectAction> actions = await resolveActionsAsync(projectManagerService);
-                    IReadOnlyList<PreviewResult> results = await GetPreviewResultsAsync(projectManagerService, actions, cancellationToken);
+                    IReadOnlyList<PreviewResult> results = await GetPreviewResultsAsync(projectManagerService, actions, userAction, uiService, cancellationToken);
 
                     if (operationType == NuGetOperationType.Uninstall)
                     {
@@ -905,6 +905,8 @@ namespace NuGet.PackageManagement.UI
         internal static async ValueTask<IReadOnlyList<PreviewResult>> GetPreviewResultsAsync(
             INuGetProjectManagerService projectManagerService,
             IReadOnlyList<ProjectAction> projectActions,
+            UserAction? userAction,
+            INuGetUI uiService,
             CancellationToken cancellationToken)
         {
             var results = new List<PreviewResult>();
@@ -934,6 +936,8 @@ namespace NuGet.PackageManagement.UI
 
             // Group actions by project
             var actionsByProject = expandedActions.GroupBy(action => action.ProjectId);
+
+            Dictionary<string, SortedSet<string>>? newSourceMappings = null;
 
             // Group actions by operation
             foreach (IGrouping<string, ProjectAction> actions in actionsByProject)
@@ -986,6 +990,11 @@ namespace NuGet.PackageManagement.UI
                     }
                 }
 
+                if (userAction?.SelectedSourceName != null)
+                {
+                    // Everything added which didn't already have a source mapping will be mentioned in the Preview Window.
+                    PackageSourceMappingUtility.GetNewSourceMappingsFromAddedPackages(ref newSourceMappings, userAction.SelectedSourceName, added, uiService.UIContext.PackageSourceMapping);
+                }
 
                 IProjectMetadataContextInfo projectMetadata = await projectManagerService.GetMetadataAsync(actions.Key, cancellationToken);
 
@@ -1004,6 +1013,12 @@ namespace NuGet.PackageManagement.UI
                 var result = new PreviewResult(projectName, added, deleted, updated);
 
                 results.Add(result);
+            }
+
+            if (newSourceMappings?.Count > 0)
+            {
+                var solutionSourceMappingResult = new PreviewResult(newSourceMappings);
+                results.Add(solutionSourceMappingResult);
             }
 
             return results;
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
@@ -993,7 +993,7 @@ namespace NuGet.PackageManagement.UI
                 if (userAction?.SelectedSourceName != null)
                 {
                     // Everything added which didn't already have a source mapping will be mentioned in the Preview Window.
-                    PackageSourceMappingUtility.GetNewSourceMappingsFromAddedPackages(ref newSourceMappings, userAction.SelectedSourceName, added, uiService.UIContext.PackageSourceMapping);
+                    PackageSourceMappingUtility.AddNewSourceMappingsFromAddedPackages(ref newSourceMappings, userAction.SelectedSourceName, added, uiService.UIContext.PackageSourceMapping);
                 }
 
                 IProjectMetadataContextInfo projectMetadata = await projectManagerService.GetMetadataAsync(actions.Key, cancellationToken);
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PreviewWindowModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PreviewWindowModel.cs
@@ -53,6 +53,19 @@ namespace NuGet.PackageManagement.UI
                     }
                     sb.AppendLine("");
                 }
+                if (r.NewSourceMappings?.Any() == true)
+                {
+                    foreach (var newSourceMapping in r.NewSourceMappings)
+                    {
+                        sb.AppendLine(string.Format(Resources.Label_CreatingSourceMappings, newSourceMapping.Key));
+                        sb.AppendLine("");
+                        foreach (string newSourceMappingPackageId in newSourceMapping.Value)
+                        {
+                            sb.AppendLine(newSourceMappingPackageId);
+                        }
+                        sb.AppendLine("");
+                    }
+                }
             }
             return sb.ToString();
         }
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PreviewWindowModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PreviewWindowModel.cs
@@ -53,7 +53,7 @@ namespace NuGet.PackageManagement.UI
                     }
                     sb.AppendLine("");
                 }
-                if (r.NewSourceMappings?.Any() == true)
+                if (r.NewSourceMappings?.Count > 0)
                 {
                     foreach (var newSourceMapping in r.NewSourceMappings)
                     {
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
@@ -3,10 +3,10 @@
 
 #nullable enable
 
-using System;
 using System.Collections.Generic;
 using System.Linq;
 using NuGet.Configuration;
+using NuGet.PackageManagement.VisualStudio;
 
 namespace NuGet.PackageManagement.UI
 {
@@ -18,7 +18,7 @@ namespace NuGet.PackageManagement.UI
             PackageSourceMappingProvider sourceMappingProvider,
             IReadOnlyList<PackageSourceMappingSourceItem> existingPackageSourceMappingSourceItems)
         {
-            if (userAction?.SourceMappingSourceName is null || addedPackageIds is null)
+            if (userAction?.SelectedSourceName is null || addedPackageIds is null)
             {
                 return;
             }
@@ -46,7 +46,7 @@ namespace NuGet.PackageManagement.UI
                 .ToArray();
 
             CreateAndSavePackageSourceMappings(
-                userAction.SourceMappingSourceName,
+                userAction.SelectedSourceName,
                 packageIdsNeedingNewSourceMappings,
                 sourceMappingProvider,
                 existingPackageSourceMappingSourceItems);
@@ -90,6 +90,53 @@ namespace NuGet.PackageManagement.UI
             }
 
             mappingProvider.SavePackageSourceMappings(newAndExistingPackageSourceMappingItems);
+        }
+
+        internal static string? GetNewSourceMappingSourceName(PackageSourceMapping packageSourceMapping, PackageSourceMoniker activePackageSourceMoniker)
+        {
+            string? sourceMappingSourceName = packageSourceMapping.IsEnabled
+                && !activePackageSourceMoniker.IsAggregateSource
+                ? activePackageSourceMoniker.PackageSourceNames.First() : null;
+
+            return sourceMappingSourceName;
+        }
+
+        internal static void GetNewSourceMappingsFromAddedPackages(ref Dictionary<string, SortedSet<string>>? newSourceMappings, string newMappingSourceName, List<AccessiblePackageIdentity> added, PackageSourceMapping packageSourceMapping)
+        {
+            if (newMappingSourceName is null || added.Count == 0 || packageSourceMapping is null)
+            {
+                return;
+            }
+
+            List<string> addedPackagesWithNoSourceMappings = added.Select(_ => _.Id)
+                .Where(addedPackage =>
+                {
+                    IReadOnlyList<string> configuredSources = packageSourceMapping.GetConfiguredPackageSources(addedPackage);
+                    return configuredSources == null || configuredSources.Count == 0;
+                })
+                .Distinct()
+                .ToList();
+
+            if (addedPackagesWithNoSourceMappings.Count == 0)
+            {
+                return;
+            }
+
+            if (newSourceMappings is null)
+            {
+                newSourceMappings = new Dictionary<string, SortedSet<string>>(capacity: 1)
+                {
+                    { newMappingSourceName, new SortedSet<string>(addedPackagesWithNoSourceMappings) }
+                };
+            }
+            else if (newSourceMappings.TryGetValue(newMappingSourceName, out SortedSet<string>? newMappingPackageIds))
+            {
+                newMappingPackageIds.UnionWith(addedPackagesWithNoSourceMappings);
+            }
+            else
+            {
+                newSourceMappings.Add(newMappingSourceName, new SortedSet<string>(addedPackagesWithNoSourceMappings));
+            }
         }
     }
 }
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
@@ -108,34 +108,30 @@ namespace NuGet.PackageManagement.UI
                 return;
             }
 
-            List<string> addedPackagesWithNoSourceMappings = added.Select(package => package.Id)
-                .Where(addedPackage =>
-                {
-                    IReadOnlyList<string> configuredSources = packageSourceMapping.GetConfiguredPackageSources(addedPackage);
-                    return configuredSources == null || configuredSources.Count == 0;
-                })
-                .Distinct()
-                .ToList();
+            foreach (AccessiblePackageIdentity addedPackage in added)
+            {
+                IReadOnlyList<string> configuredSources = packageSourceMapping.GetConfiguredPackageSources(packageId: addedPackage.Id);
 
-            if (addedPackagesWithNoSourceMappings.Count == 0)
-            {
-                return;
-            }
-
-            if (newSourceMappings is null)
-            {
-                newSourceMappings = new Dictionary<string, SortedSet<string>>(capacity: 1)
+                if (configuredSources.Count > 0)
                 {
-                    { newMappingSourceName, new SortedSet<string>(addedPackagesWithNoSourceMappings) }
-                };
-            }
-            else if (newSourceMappings.TryGetValue(newMappingSourceName, out SortedSet<string>? newMappingPackageIds))
-            {
-                newMappingPackageIds.UnionWith(addedPackagesWithNoSourceMappings);
-            }
-            else
-            {
-                newSourceMappings.Add(newMappingSourceName, new SortedSet<string>(addedPackagesWithNoSourceMappings));
+                    continue;
+                }
+
+                if (newSourceMappings is null)
+                {
+                    newSourceMappings = new Dictionary<string, SortedSet<string>>(capacity: 1)
+                    {
+                        { newMappingSourceName, new SortedSet<string>(new List<string>(capacity: added.Count) { addedPackage.Id }) }
+                    };
+                }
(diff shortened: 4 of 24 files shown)


src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -91,5 +91,52 @@ internal static class PackageSourceMappingUtility
 
             mappingProvider.SavePackageSourceMappings(newAndExistingPackageSourceMappingItems);
         }
+
+        internal static string? GetNewSourceMappingSourceName(PackageSourceMapping packageSourceMapping, PackageSourceMoniker activePackageSourceMoniker)
+        {
+            string? sourceMappingSourceName = packageSourceMapping.IsEnabled
+                && !activePackageSourceMoniker.IsAggregateSource
+                ? activePackageSourceMoniker.PackageSourceNames.First() : null;
```
aortiz-msft, line 99: NIT: the indentation looks off. I'm curious what the C# style analyzer would recommend. Maybe we should consider start using it for our codebase. Anyway, my suggestion is: 

<pre>
string? sourceMappingSourceName = packageSourceMapping.IsEnabled 
                                      && !activePackageSourceMoniker.IsAggregateSource ?
                                          activePackageSourceMoniker.PackageSourceNames.First() : null;
</pre>

aortiz-msft (commented):
I couldn't get past the line I commented on today.

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
donnie-msft: According to our coding guideline, we indent with 4 spaces.
`                                      &&` doesn't seem right to me. 
I'm not sure which C# analyzer would be most applicable, but the `case` statement also uses 4 spaces:
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/csharp-formatting-options#csharp_indent_case_contents

Maybe others can clarify what the team prefers.

nkolev92: We already have whitespace analyzers, but they can't catch stuff like that. 
The indentation is correct. An extra level of indentation for certain types is not something we have configured. not sure if we can. 

For context; 

- https://github.com/NuGet/NuGet.Client/blob/216cdfabbeefef8ce52d16d424a1a3cbc8a262d0/eng/pipelines/templates/Static_Source_Analysis.yml#L77 Static source analysis leg runs `dotnet format`
 - https://github.com/NuGet/NuGet.Client/blob/216cdfabbeefef8ce52d16d424a1a3cbc8a262d0/build/common.project.props#L108 - Turns on style analyzers for build, even if they're only so accurate these days (there's some notable roslyn bugs)

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -91,5 +91,52 @@ internal static class PackageSourceMappingUtility
...
+        }
+
+        internal static void GetNewSourceMappingsFromAddedPackages(ref Dictionary<string, SortedSet<string>>? newSourceMappings, string newMappingSourceName, List<AccessiblePackageIdentity> added, PackageSourceMapping packageSourceMapping)
+        {
+            if (newMappingSourceName is null || added.Count == 0 || packageSourceMapping is null)
+            {
+                return;
+            }
+
+            List<string> addedPackagesWithNoSourceMappings = added.Select(_ => _.Id)
```
jeffkl: I have never seen this pattern to use a throw away in Linq when getting the property.  This looks more _normal_ to me.  Just curious what everyone else's thoughts on this pattern?

```suggestion
            List<string> addedPackagesWithNoSourceMappings = added.Select(i => i.Id)
```

jeffkl (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
donnie-msft: Good callout! I've been trying to use throw-aways more often, when I don't care about the variable name and don't really need one. To me, arbitrary letters are unnecessary noise. (`i` probably isn't what I'd choose, but maybe `p` for "package" or the word itself?)

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -91,5 +91,52 @@ internal static class PackageSourceMappingUtility
...
+                .Where(addedPackage =>
+                {
+                    IReadOnlyList<string> configuredSources = packageSourceMapping.GetConfiguredPackageSources(addedPackage);
+                    return configuredSources == null || configuredSources.Count == 0;
+                })
+                .Distinct()
+                .ToList();
```
aortiz-msft: How many collections are allocated here? It looks like n of added packages (configuredSources), then the call to Distinct() would create a new collection, as well as the call to ToList(). Could we save allocations by using a for loop and AddRange() instead?

aortiz-msft (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
aortiz-msft: It did look a bit unusual at first, but ultimately I find the code readable.

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
aortiz-msft: There are two parts to my comment. One is starting the indentation for a line with an assignment after the = operator. The other is not started a line with the ? operator and instead having the ? operator at the end of the previous line. The latter is more important IMO. Nikolche has a couple of suggestions that he'll add here.

nkolev92: Personally, I like having the operators at the end. IMO it makes it more readable.

Example: 
```cs
string? sourceMappingSourceName = packageSourceMapping.IsEnabled && 
     !activePackageSourceMoniker.IsAggregateSource ?
          activePackageSourceMoniker.PackageSourceNames.First() : null;
```
Or

```cs
string? sourceMappingSourceName = packageSourceMapping.IsEnabled && 
     !activePackageSourceMoniker.IsAggregateSource ?
          activePackageSourceMoniker.PackageSourceNames.First() : 
          null;
```

I feel like given that `&&` and `?, : ` operators are of different priority, they should be intended differently. 

It's be great if we can configure the style analyzers to enforce it.

donnie-msft: > The indentation is correct.

Thanks, @nkolev92 !

> It's be great if we can configure the style analyzers to enforce it.

nkolev92 & @aortiz-msft , Glad to see the team having interest in standardizing more coding guidelines, and especially setting up analyzers. Could you please take that conversation offline so we don't distract from the PR review? I'll add my comments on the examples provided to the issue/PR/Teams thread, whichever you prefer to start. Thanks!

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
donnie-msft: >  then the call to Distinct() would create a new collection

That's not how I understand `Distinct()` to work. It operates on an `IEnumerable`, and does not allocate anything, but simply tells the query to do it once iterated.

> as well as the call to ToList().

This is 1 allocation (there may be some internal temp arrays necessary as the collection grows, but that would be the same for `AddRange`, I believe as well).

> Could we save allocations by using a for loop and AddRange() instead?

Are you talking about this PR or are you talking about implementation of `GetConfiguredPackageSources()`? 
If you're talking about [GetConfiguredPackageSources](https://github.com/NuGet/NuGet.Client/blob/65e74ebcf721c07672394c598dccea2d86a4479d/src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMapping.cs#L38), that implementation works on the [SearchTree](https://github.com/NuGet/NuGet.Client/blob/65e74ebcf721c07672394c598dccea2d86a4479d/src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/SearchTree.cs) by looking for term matches in the linked list in [SearchNodeByTerm](https://github.com/NuGet/NuGet.Client/blob/65e74ebcf721c07672394c598dccea2d86a4479d/src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/SearchTree.cs#L108), which as far as I can tell doesn't allocate any collections as part of its search.

zivkan: > That's not how I understand Distinct() to work. It operates on an IEnumerable, and does not allocate anything, but simply tells the query to do it once iterated.

I'm not sure how any algorithm could determine a distinct list of items without any allocations. Here's the .NET 8 implementation: https://github.com/dotnet/runtime/blob/a39262bd2d8543e5e6636978d54a084669e1a8e5/src/libraries/System.Linq/src/System/Linq/Distinct.cs#L62-L79
It allocates a `HashSet<T>`, which itself has an internal collection that might grow depending on the number of insertions.

On .NET Framework, the method also uses `yield return`, meaning calling the method creates a finite state machine on the heap: https://referencesource.microsoft.com/#System.Core/System/Linq/Enumerable.cs,836

> This is 1 allocation (there may be some internal temp arrays necessary as the collection grows, but that would be the same for AddRange, I believe as well).

You can avoid temporary allocations (resizing) if it's possible to pre-compute the expected result list size. Then you use the constructor where you pass the capacity as a parameter. Then you can use `AddRange` (although using `IEnumerable<T>` causes allocations), so a `for` loop (or a `foreach` loop, but only when the collection you're iterating has a `struct` enumerator) can avoid all temporary allocations.

Even when you can't calculate the expected result size accurately, if you can estimate it and be close enough most of the time, it's still going to be a win, because `List<T>()`'s default constructor creates a list with zero capacity, the first add would allocate a capacity of 4, then it would double in size each time it got full. So, if you know you're going to have ballpark 10 items, using the default list capacity will cause 2 resizes. I'm not sure if newer versions of .NET have changed this, but VS runs on the .NET Framework anyway.

Having written all that, I haven't yet looked at this PR's code, I'm only writing in general terms (since "the same for AddRange" is also a general statement, not specific to this PR). For our most performance sensitive code, like restore, it's worthwhile to write extra lines of code to reduce allocations and increase performance, even when it increases code complexity by a "reasonable" amount (which unfortunately is subjective). However, for non-perf critical code, it's not worthwhile making the code a lot more complex.

I strongly encourage everyone to always practise basic performance best-practises, even in code paths that are not at all perf sensitive, so that when you later work on something that actually is on a hot path and is performance sensitive, you're already well practised in thinking about how to write code that reduces allocations, or is otherwise "high performance".

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
```diff
@@ -13,10 +16,12 @@ public class PreviewResult
 
         public IEnumerable<UpdatePreviewResult> Updated { get; }
 
-        public string Name { get; }
+        public IReadOnlyDictionary<string, SortedSet<string>>? NewSourceMappings { get; }
```
zivkan: Consider using `ImmutableDictionary`, rather than `IReadOnlyDictionary`. The benefit is that it's possible to enumerate the dictionary with zero allocations, which is not possible with `IReadOnlyDictionary`. edit: and the dictionary is being enumerated in `PreviewWindowModel.cs`

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PreviewWindowModel.cs
```diff
@@ -53,6 +53,19 @@ public override string ToString()
                     }
                     sb.AppendLine("");
                 }
+                if (r.NewSourceMappings?.Any() == true)
```
zivkan: https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/24181/Garbage-Collection?anchor=%E2%9C%94%EF%B8%8F-**do**-favor-%60length%60-and-%60count%60-checks-over-%60any%60

```suggestion
                if (r.NewSourceMappings?.Count > 0)
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
zivkan: From the docs of the C# language feature "discards": https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/discards

> Discards are placeholder variables that are intentionally unused in application code.

also

> Discards make the intent of your code clear. A discard indicates that our code never uses the variable.

Therefore, doing `_.Id` violates that intent.

The reason it works is because 1. when defining lamba functions, it's not necessary to specify types, when the compiler is capable of inferring the type (which it is in this code snippet). 2. C# allowed variables named `_` long before the discard feature was created, and to minimize risk of breaking customers with existing codebases, the language/compiler team continue to allow variables named `_`.

Therefore, this is not actually using the discard language feature, it's creating a variable named `_`.

Also from the docs:

> When _ is a valid discard, attempting to retrieve its value or use it in an assignment operation generates compiler error CS0103, "The name '_' doesn't exist in the current context". This error is because _ isn't assigned a value, and may not even be assigned a storage location.

So, depending on how familiar someone is with the discard language feature, using `_` as a variable name could be confusing (why didn't the compiler error out with CS0103?). Following the usual "use meaningful names" guideline (which I see isn't actually in our coding guidelines document) suggests that `_` isn't a good variable name as I can't infer any information at all about the data type or meaning of the variable from the name alone.

```suggestion
            List<string> addedPackagesWithNoSourceMappings = added.Select(package => package.Id)
```

though I'd expect most people would accept `p` as an acceptable variable name too. I think Jeff's suggestion `i` is not a good suggestion, because it further obfuscates what `i` is (is it an integer/index? is it a source? is it a package?).

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -91,5 +91,52 @@ internal static class PackageSourceMappingUtility
...
+        internal static string? GetNewSourceMappingSourceName(PackageSourceMapping packageSourceMapping, PackageSourceMoniker activePackageSourceMoniker)
+        {
+            string? sourceMappingSourceName = packageSourceMapping.IsEnabled
+                && !activePackageSourceMoniker.IsAggregateSource
+                ? activePackageSourceMoniker.PackageSourceNames.First() : null;
+
+            return sourceMappingSourceName;
+        }
+
+        internal static void GetNewSourceMappingsFromAddedPackages(ref Dictionary<string, SortedSet<string>>? newSourceMappings, string newMappingSourceName, List<AccessiblePackageIdentity> added, PackageSourceMapping packageSourceMapping)
```
zivkan: From an API design point of view, `Get` suggests a return value. I think `Add` better represents that the method is adding to a dictionary (well, unless `null` was passed in, which which case it creates a dictionary first)

```suggestion
        internal static void AddNewSourceMappingsFromAddedPackages(ref Dictionary<string, SortedSet<string>>? newSourceMappings, string newMappingSourceName, List<AccessiblePackageIdentity> added, PackageSourceMapping packageSourceMapping)
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
zivkan: ok, having looked though this PR, and better understanding what this method does, I have a few ideas.

One option is to create the list with a suitable capacity, and then use `AddRange`, as Arturo suggested. The upper bound of how big the result will be is `added.Count`, assuming none of the new packages have a mapping and all package IDs are unique. While it might feel like it's a waste of memory to create an array (`List<T>`'s backing store is ultimately an array), since the `addedPackagesWithNoSourceMappings` is not used outside of this method, it should hopefully be cleaned up in gen0 GC, so not overly expensive. I'm sure that would be less customer impact than requiring multiple allocations when the customer installs a package like `Microsoft.VisualStudio.SDK` and none of the packages have a mapping yet.

However, a second option is to avoid using the temporary `addedPackagesWithNoSourceMappings` list. Use a `foreach` loop on `added` (which is of type `List<T>` and therefore `foreach` is zero allocation), and then add to the dictionary directly. It means the `if (newSourceMappings is null)` check will happen multiple times, but that will have less perf impact that creating a temporary list (especially if it ever gets resized), which then needs to get GC'd.

To be honest, I'm not sure why the `Distinct()` is used in the LINQ expression. Isn't the list of `added` packages already unique? If not, that's very unintuitive to me, even in multi-targeting projects. But in that case, one option is to use your own `HashSet<T>`, as `Distinct()` does. However, [`SortedSet<T>.Add`](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.sortedset-1.add?view=net-7.0) does not throw an exception when a duplicate is added. So, I don't see any perf benefit to calculating duplicates in this method. Only using `newSourceMappings.Add` achieves the same as using a `HashSet` (either explicitly, or through `.Distinct()`), except without the allocations.

So, I believe this method could avoid all temporary allocations.

donnie-msft: > I'm not sure how any algorithm could determine a distinct list of items without any allocations. Here's the .NET 8 implementation:

I'm not sure either. But I was almost certain that the Iterator pattern was used in LINQ. If I look at what I believe is the code behind the extension method I'm using, I see it returns a `DistinctIterator`, which I thought means it's not iterated at this point in time?
https://github.com/microsoft/referencesource/blob/master/System.Core/System/Linq/Enumerable.cs#L826

Will reply to other comments later...

donnie-msft: > looks like n of added packages (configuredSources), then the call to Distinct() would create a new collection

I read this last night as N allocations for distinct which I'm now seeing it means 1 allocation? Sure. When I'm at my desk I'll look again why I thought I needed distinct here.

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
aortiz-msft: Good catch. I missed that.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
donnie-msft: Switched this out. My understanding is that it will enumerate the dictionary up front (in the ctor).

BTW, don't forget that XAML also binds to this property and that VS's "Find references" feature does not identify those.

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PreviewWindowModel.cs
donnie-msft: 👍

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
donnie-msft: Thanks for explaining that. I realize now I started this habit to avoid redundant naming, and it's actually just been an underscore variable name all along 😁.

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
donnie-msft: Absolutely! As I reworked the implementation, the name got a little stale.

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
donnie-msft: Thanks for the suggestion! @zivkan, I believe I've implemented your "second option" above. I added a single test case for the method in `PackageSourceMappingUtility` before changing it. Let me know if this is looks like what you meant. 

Looking back at my use of `Distinct`, it must have been a relic from before I started using a `SortedSet`, as I can't figure out why I'd need it. 

> Isn't the list of added packages already unique? If not, that's very unintuitive to me, even in multi-targeting projects.

I believe when I saw the loop for project actions here, I was thinking that when multiple install actions occur, which can be the same package to multiple projects, that I could see the same top-level or transitive package in a single call to `GetPreviewResultsAsync` . See loops here: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs#L945

zivkan: LGTM! No more easy to spot perf improvements. 👍

zivkan (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
```diff
@@ -13,10 +17,12 @@ public class PreviewResult
...
         public IEnumerable<UpdatePreviewResult> Updated { get; }
 
-        public string Name { get; }
+        public ImmutableDictionary<string, SortedSet<string>>? NewSourceMappings { get; }
+
+        public string? Name { get; }
 
         public PreviewResult(
-            string projectName,
+            string? projectName,
```
nkolev92: When does this not apply? :D

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
```diff
@@ -383,7 +383,7 @@ private static string GenerateUpgradeReport(string? projectName, string backupPa
                     TelemetryServiceUtility.StartOrResumeTimer();
 
                     IReadOnlyList<ProjectAction> actions = await resolveActionsAsync(projectManagerService);
-                    IReadOnlyList<PreviewResult> results = await GetPreviewResultsAsync(projectManagerService, actions, cancellationToken);
+                    IReadOnlyList<PreviewResult> results = await GetPreviewResultsAsync(projectManagerService, actions, userAction, uiService, cancellationToken);
```
nkolev92, line 386: you could probably pass down the PSM object here, but sounds like it might be more helpful to have the service later. 

Then again, having apprx 20 args is unfriendly as well. 

I'd say if you end up using like 2 things off of the ui service, I'd just pass those down, but otherwise I don't feel too strongly.

src/NuGet.Core/NuGet.Configuration/Settings/Items/PackageSourceMappingSourceItem.cs
```diff
@@ -157,7 +157,7 @@ internal override void Update(SettingItem other)
 
             base.Update(other);
 
-            Dictionary<PackagePatternItem, PackagePatternItem> otherPatterns = packageSourceMappingSourceItem.Patterns.ToDictionary(c => c, c => c);
+            Dictionary<PackagePatternItem, PackagePatternItem> otherPatterns = packageSourceMappingSourceItem.Patterns.GroupBy(item => item).ToDictionary(c => c.Key, c => c.Key);
```
nkolev92, line 160: Do we have a test? Sounds like we might, if we don't, we need one.

This is a low level API, so it should be easily testable.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
```diff
@@ -1678,6 +1678,13 @@ await PreviewBuildIntegratedProjectActionsAsync(buildIntegratedProject, actions,
 
             if (buildIntegratedProjectsToUpdate.Count != 0)
             {
+                // Only automatically create Source Mappings when there's exclusively BuildIntegratedProjects.
```
nkolev92, line 1681: This doesn't make me feel warm and fuzzy. Maybe have a bolean that checks when something is incompatible and maybe passes a null for new mapping id/new source there, instead of nulling it out here. 

idk if that's any better, but just an idea.

nkolev92 (commented):
A few small comments. 

I feel like the bug bash will help catching any potential issues.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
donnie-msft: When `UniqueName` is null: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs#L1000

Apparently, this is optional metadata for a project: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/ProjectMetadataContextInfo.cs#L49

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
donnie-msft: I'm hoping to remove this soon, but thinking of how to improve it, anyway... I'd rather not have to iterate the projects and determine types separately. 
Leave it for now?

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
donnie-msft: Attempted to refactor this, but my tests broke since `NuGet.Configuration/PackageSourceMapping` class is not Mockable. Therefore, my tests verifying # times the object is referenced now need to look at counts for particular method invocations (ie, `GetConfiguredPackageSources`). To do that will require refactoring of the class, and I don't believe that's worth doing here.
https://github.com/NuGet/Home/issues/12825

src/NuGet.Core/NuGet.Configuration/Settings/Items/PackageSourceMappingSourceItem.cs
donnie-msft: Seems existing tests did not cover duplicates, so I added `PackageSourceMappingSourceItemTests.Update_WithDuplicatePatterns_WritesSinglePattern`

Verified it threw the exception when the unchanged code was given a config with existing duplicate `PackagePatterns`, or when attempting to add duplicate `PackagePatterns`. (the exception is shown in screenshot):
image

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
nkolev92: What does not mockable mean?

It's a class with content. YOu can preload it with different data and pass it down. Isn't that sufficient?

donnie-msft: > What does not mockable mean?

Mockable is defined here: https://github.com/moq/moq/wiki/FAQ#moq---non-overridable-members-may-not-be-used-in-setup--verification-expressions

> It's a class with content. YOu can preload it with different data and pass it down. Isn't that sufficient?

No, because as I mentioned, I'm testing # calls to the object, not the values of the object.

aortiz-msft (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
nkolev92: I see. FWIW, in the first link the value can't be null.

nkolev92: So for when we create preview result, it seems to suggest it's always non-null. The first code work arounds that using unknown project and not a null.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
nkolev92: I'm not I follow why the call count is relevant and why it needs to be mockable. 

There's also a point in the doc you linked: 

> In this case there's no need to mock **PagingOptions **because it's easy to use a real one.

I think PackageSourceMapping is one of those classes where it's easy to use a real one.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
nkolev92: Yeah. I think we should add a tracking issue, if there is one.

donnie-msft: https://github.com/NuGet/Home/issues/12832

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
donnie-msft: > I see. FWIW, in the first link the value can't be null.

image

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
donnie-msft: I'd be happy to learn how to count method calls without using Mocks. However, since I've created a follow-up issue, can we discuss that in that issue? If ultimately it needs to be closed, that's fine. It's really not relevant to this PR, though.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
nkolev92: Yes and If that value is null, projectName gets assigned to Resources.UnknownProject. 

projectName is what gets passed to previewresult and that's non null.

donnie-msft: 🤦‍♂️ you're right - fixed & did a quick runtime test:
image

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
donnie-msft: @nkolev92 we talked offline today, and I think I understood we're good to discuss this separately as I suggested. Let me know if it's intended to be blocking feedback or not.

zivkan (approved):

aortiz-msft (approved):

nkolev92 (approved):

PR #5361, opened 2023-08-15
Pass through the cache context to the vulnerability resource

nkolev92:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2432

## Description

The VulnerabilityInformationProvider creates redundant SourceCacheContext. We should reuse the same SourceCacheContext   throughout the whole operation. 

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception - Tests already exist for the enum changes. 
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommandProviders.cs
```diff
@@ -39,11 +39,12 @@ public class RestoreCommandProviders
         private static IReadOnlyList<IVulnerabilityInformationProvider> CreateVulnerabilityInfoProviders(IReadOnlyList<IRemoteDependencyProvider> remoteProviders)
         {
             List<IVulnerabilityInformationProvider> providers = new(remoteProviders.Count);
+            using var sourceCacheContext = new SourceCacheContext();
```
donnie-msft, line 42: To match naming in the `SourceRepositoryDependencyProvider` and other places..
```suggestion
            using var cacheContext = new SourceCacheContext();
```

src/NuGet.Core/NuGet.Commands/RestoreCommand/VulnerabilityInformationProvider.cs
```diff
@@ -15,13 +15,15 @@ namespace NuGet.Commands
     internal sealed class VulnerabilityInformationProvider : IVulnerabilityInformationProvider
     {
         private readonly SourceRepository _source;
+        private readonly SourceCacheContext _sourceCacheContext;
```
donnie-msft, line 18: Same as other feedback.

donnie-msft (dismissed):

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -304,7 +304,7 @@ public async Task<RestoreResult> ExecuteAsync(CancellationToken token)
                 }
 
                 AuditUtility.EnabledValue enableAudit = AuditUtility.ParseEnableValue(_request.Project.RestoreMetadata?.RestoreAuditProperties?.EnableAudit);
-                telemetry.TelemetryEvent[AuditEnabled] = AuditUtility.GetString(enableAudit);
+                telemetry.TelemetryEvent[AuditEnabled] = enableAudit.ToString();
```
zivkan: I had the `GetString` methods intentionally, because `someEnum.ToString()` causes several allocations. I'm not 100% sure, but I think of .NET Framework at least, it causes boxing on the enum value, and then uses reflection to figure out the name, but allocates it in a new string each time.

The code you deleted allocates zero, because the strings returned by the method are const, so it keeps returning the same const instance.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommandProviders.cs
nkolev92: There's heavy usage of both tbh. 

image

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
nkolev92: I'll take a look and profile this quickly.

martinrrm (dismissed):

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommandProviders.cs
aortiz-msft: FWIW I like sourceCacheContext better.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
zivkan: I couldn't take the suspense of waiting another day to get an answer, so I wrote a quick benchmark myself:

<details>
<summary>code</summary>
<div>

```cs
using BenchmarkDotNet.Jobs;
using System.Net;

[MemoryDiagnoser]
[SimpleJob(RuntimeMoniker.Net481)]
[SimpleJob(RuntimeMoniker.Net80)]
public class Program
{
    {
    }

    [Benchmark]
    public string EnumToString()
    {
        return HttpStatusCode.OK.ToString();
    }

    [Benchmark(Baseline = true)]
    public string CustomGetString()
    {
        return GetString(HttpStatusCode.OK);

        static string GetString(HttpStatusCode code)
        {
            return code switch
            {
                HttpStatusCode.Continue => nameof(HttpStatusCode.Continue),
                HttpStatusCode.SwitchingProtocols => nameof(HttpStatusCode.SwitchingProtocols),
#if !NETFRAMEWORK
                HttpStatusCode.Processing => nameof(HttpStatusCode.Processing),
                HttpStatusCode.EarlyHints => nameof(HttpStatusCode.EarlyHints),
#endif
                HttpStatusCode.OK => nameof(HttpStatusCode.OK),
                _ => code.ToString(),
            };
        }
    }

}
```

</div>
</details>

|          Method |                  Job |              Runtime |        Mean |     Error |    StdDev |  Ratio | RatioSD |   Gen0 | Allocated | Alloc Ratio |
|---------------- |--------------------- |--------------------- |------------:|----------:|----------:|-------:|--------:|-------:|----------:|------------:|
|    EnumToString |             .NET 8.0 |             .NET 8.0 |  11.7090 ns | 0.2553 ns | 0.2132 ns |      ? |       ? | 0.0014 |      24 B |           ? |
| CustomGetString |             .NET 8.0 |             .NET 8.0 |   0.1623 ns | 0.0427 ns | 0.1260 ns |      ? |       ? |      - |         - |           ? |
|                 |                      |                      |             |           |           |        |         |        |           |             |
|    EnumToString | .NET Framework 4.8.1 | .NET Framework 4.8.1 | 413.9441 ns | 2.4596 ns | 2.3007 ns | 729.05 |    6.14 | 0.0153 |      96 B |          NA |
| CustomGetString | .NET Framework 4.8.1 | .NET Framework 4.8.1 |   0.5678 ns | 0.0037 ns | 0.0035 ns |   1.00 |    0.00 |      - |         - |          NA |

Given that a custom, zero-allocation method is 729 times faster than `.ToString()` on .NET Framework, can I call myself a 729x developer? 10x developers, keep trying harder😆 


But seriously, while writing the benchmark, I remembered that during VS's perfathon, we got a related PR contribution, also related to enum `ToString` allocations: 
* https://github.com/NuGet/NuGet.Client/pull/5215
* https://github.com/NuGet/Home/issues/12638

jeffkl (changes requested):
Looks good otherwise but let's bring back the `Enum.ToString()` optimization

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
aortiz-msft: Thank you for the analysis!

nkolev92: Another reference for the curious folks as well. 

https://github.com/dotnet/runtime/issues/57748

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -450,28 +450,6 @@ public static EnabledValue ParseEnableValue(string value)
             return EnabledValue.Undefined;
         }
```
aortiz-msft, line 459: In light of the learnings shared by Andy, I think it would be appropriate to add a comment on why these methods are important from a performance perspective to avoid someone else trying to make the same change in the future.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommandProviders.cs
nkolev92: Resolving as this is stylistic without a established pattern.

aortiz-msft (changes requested):

nkolev92:
@jeffkl @aortiz-msft Addressed all feedback.

jeffkl (dismissed):

zivkan (dismissed):

aortiz-msft (dismissed):

zivkan (approved):

PR #5363, opened 2023-08-17
Opt out of "HTTPS everywhere": Do not warn http sources in source command when allowInsecureConnections is set to true

heng-liu:
## Bug

Fixes:https://github.com/NuGet/Home/issues/12791

## Description
Suppress the Http warning in source command when allowInsecureConnections is set to true.

Note: only `disable`, `enable`, `list` would be able to apply allowInsecureConnections.
For `add` and `update`, the `allowInsecureConnections` will always be false, the default value. As we do not accept the additional properties when add or update source (same with property protocalVersion). 
Based on our [discussion](https://teams.microsoft.com/l/message/19:994f8ef52dd948039177020177583736@thread.v2/1690499497810?context=%7B%22contextType%22%3A%22chat%22%7D), adding -allowInsecureConnections option to nuget sources add/update commands could be done later.

BTW, we have a PR about adding -protocolVersion option to nuget sources add/update commands in https://github.com/NuGet/NuGet.Client/pull/5002
We could probably add allowInsecureConnections option to nuget sources add/update commands as well.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


jeffkl (approved):

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
```diff
@@ -366,6 +503,76 @@ public void Sources_NoWarnWhenDisableHttpSource()
...
 
+        [Theory]
+        [InlineData("http://source.test", "true")]
+        [InlineData("http://source.test", "TRUE")]
+        [InlineData("http://source.test", "FALSE")]
+        [InlineData("http://source.test", "invalidString")]
+        [InlineData("http://source.test", "")]
+        [InlineData("https://source.test", "TRUE")]
+        [InlineData("https://source.test", "false")]
+        public void Sources_DisableHttpSourceWithAllowInsecureConnections_NoWarns(string source, string allowInsecureConnections)
```
aortiz-msft: For this test, I wonder if there's value in testing anything other than [InlineData("http://source.test", "false")]. 

@zivkan, @nkolev92 -  thoughts?

aortiz-msft (approved):

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
nkolev92: I agree. 

A lot of this is just testing the parsing of the allowinsecureconnections .  All of that should be unit testable. 

Also https://github.com/NuGet/Client.Engineering/issues/133. 

All of these tests are really really slow. 
I'm guessing this PR adds over a minute of test time altogether. 

I don't see why we need to test every combination.

nkolev92 (changes requested):
I think the test are overly enthusiastic. 
They're testing so many different combinations, but in the end, in the product code, we only have a boolean. So we're effectively testing boolean parsing in a lot of these cases. 

I'm guessing there are plenty of unit tests regarding the parsing of `AllowInsecureConnections` from the config. 

We don't need to duplicate that here.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
heng-liu: Thanks for reviewing! Makes sense. I removed several tests.

heng-liu:
Thanks for all the reviews! I removed some unnecessary tests. Pls take another look. Thanks!

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/NuGetSourcesCommandTest.cs
```diff
@@ -156,6 +157,78 @@ public void SourcesCommandTest_EnableSource_WarnWhenUsingHttp()
             }
         }
 
+        [Theory]
+        [InlineData("http://source.test", "true", false)]
+        [InlineData("http://source.test", "false", true)]
```
nkolev92: I think even this is redundant. I'm guessing there's already a test that asserts a warning is raised...when no disabled. 

Same feedback applied.

heng-liu: Thanks! Yes, there are tests testing about enabling and disabling http source already.
I removed the two newly added tests entirely.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
```diff
@@ -303,6 +359,82 @@ public void Sources_WarnWhenEnableHttpSource()
             }
         }
 
+        [Theory]
+        [InlineData("http://source.test", "true", false)]
```
nkolev92: Same as the other tests?

heng-liu: Yes.. sorry, just removed those tests and checked no newly added tests testing enabling and disabling http source.

aortiz-msft (approved):

nkolev92 (approved):

PR #5364, opened 2023-08-17
Make GlobalJsonReader in NuGetSdkResolver a static singleton to improve caching

### LINKED ISSUE #12819: NuGetSdkResolver loads global.json multiple times during project load
### NuGet Product Used

MSBuild.exe

### Product Version

All

### Worked before?

_No response_

### Impact

### Repro Steps & Context

The NuGetSdkResolver uses a class to read `global.json`.  I added a caching mechanism to this class so that it would not read `global.json` more than once unless it changed on disk.  However, during project evaluation MSBuild is calling the NuGetSdkResolver in a way that means that multiple instances of the `GlobalJsonReader` class are instantiated.  We should use a static singleton instead.

https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/NuGetSdkResolver.cs#L43

I found this by manually debugging the SDK resolver while investigating something else.

### Verbose Logs

_No response_

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12819

## Description

While investigating another issue with the NuGetSdkResolver, I noticed that it was reading `global.json` more than one time which is unexpected.  It turns out that the way Project System loads projects can case multiple instances of the NuGetSdkResolver to be instantiated.  This leads to each NuGetSdkResolver instance also creating its own instance of `GlobalJsonReader`.  Since `GlobalJsonReader` has a file cache based on file path and last write time, its safe to make it static and reuse the cache for the whole process. 

This change makes use of a static singleton to make sure that a single process only ever loads `global.json` once unless it changes on disk.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception - Existing tests were updated, no new tests needed.
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/GlobalJsonReader.cs
+++ b/src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/GlobalJsonReader.cs
@@ -24,14 +24,21 @@ namespace Microsoft.Build.NuGetSdkResolver
         public const string GlobalJsonFileName = "global.json";
 
         /// <summary>
-        /// Tthe section of global.json contains MSBuild project SDK versions.
+        /// The name of the section in global.json that contains MSBuild project SDK versions.
         /// </summary>
         public const string MSBuildSdksPropertyName = "msbuild-sdks";
 
         /// <summary>
         /// Represents a thread-safe cache for files based on their full path and last write time.
         /// </summary>
-        private readonly ConcurrentDictionary<FileInfo, (DateTime LastWriteTime, Lazy<Dictionary<string, string>> Lazy)> _fileCache = new ConcurrentDictionary<FileInfo, (DateTime, Lazy<Dictionary<string, string>>)>(FileSystemInfoFullNameEqualityComparer.Instance);
+        private static readonly ConcurrentDictionary<FileInfo, (DateTime LastWriteTime, Lazy<Dictionary<string, string>> Lazy)> FileCache = new ConcurrentDictionary<FileInfo, (DateTime, Lazy<Dictionary<string, string>>)>(FileSystemInfoFullNameEqualityComparer.Instance);
+
+
+        private GlobalJsonReader()
+        {
+        }
+
+        public static GlobalJsonReader Instance { get; } = new GlobalJsonReader();
 
         /// <summary>
         /// Occurs when a file is read.
@@ -70,7 +77,7 @@ namespace Microsoft.Build.NuGetSdkResolver
             }
 
             // Add a new file to the cache if it doesn't exist.  If the file is already in the cache, read it again if the file has changed
-            (DateTime _, Lazy<Dictionary<string, string>> Lazy) cacheEntry = _fileCache.AddOrUpdate(
+            (DateTime _, Lazy<Dictionary<string, string>> Lazy) cacheEntry = FileCache.AddOrUpdate(
                 globalJsonPath,
                 key => (key.LastWriteTime, new Lazy<Dictionary<string, string>>(() => ParseMSBuildSdkVersions(key.FullName, context))),
                 (key, item) =>
--- a/src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/NuGetSdkResolver.cs
+++ b/src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/NuGetSdkResolver.cs
@@ -40,7 +40,7 @@ namespace Microsoft.Build.NuGetSdkResolver
         /// Initializes a new instance of the NuGetSdkResolver class.
         /// </summary>
         public NuGetSdkResolver()
-            : this(new GlobalJsonReader())
+            : this(GlobalJsonReader.Instance)
         {
         }
 
--- a/test/NuGet.Core.Tests/Microsoft.Build.NuGetSdkResolver.Test/GlobalJsonReader_Tests.cs
+++ b/test/NuGet.Core.Tests/Microsoft.Build.NuGetSdkResolver.Test/GlobalJsonReader_Tests.cs
@@ -47,7 +47,7 @@ namespace Microsoft.Build.NuGetSdkResolver.Test
 
                 var context = new MockSdkResolverContext(testDirectory);
 
-                var globalJsonReader = new GlobalJsonReader();
+                var globalJsonReader = GlobalJsonReader.Instance;
 
                 string actualGlobalJsonPath = null;
 
@@ -94,7 +94,7 @@ namespace Microsoft.Build.NuGetSdkResolver.Test
 }}
 }}");
 
-                var globalJsonReader = new GlobalJsonReader();
+                var globalJsonReader = GlobalJsonReader.Instance;
 
                 string actualGlobalJsonPath = null;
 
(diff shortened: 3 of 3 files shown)


src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/GlobalJsonReader.cs
```diff
@@ -24,14 +24,21 @@ internal sealed class GlobalJsonReader : IGlobalJsonReader
...
-        /// Tthe section of global.json contains MSBuild project SDK versions.
+        /// The name of the section in global.json that contains MSBuild project SDK versions.
         /// </summary>
         public const string MSBuildSdksPropertyName = "msbuild-sdks";
 
         /// <summary>
         /// Represents a thread-safe cache for files based on their full path and last write time.
         /// </summary>
-        private readonly ConcurrentDictionary<FileInfo, (DateTime LastWriteTime, Lazy<Dictionary<string, string>> Lazy)> _fileCache = new ConcurrentDictionary<FileInfo, (DateTime, Lazy<Dictionary<string, string>>)>(FileSystemInfoFullNameEqualityComparer.Instance);
+        private static readonly ConcurrentDictionary<FileInfo, (DateTime LastWriteTime, Lazy<Dictionary<string, string>> Lazy)> FileCache = new ConcurrentDictionary<FileInfo, (DateTime, Lazy<Dictionary<string, string>>)>(FileSystemInfoFullNameEqualityComparer.Instance);
```
zivkan, line 34: Do we have a plan for "memory leaks" for customers who close and open a LOT of different solutions/repos in a single VS session?

I'm ok with this as a temporary workaround, but are we planning to work with our partners to improve the API, so NuGet (and anything else that implements MSBuild's resolver API) can use a different "cache" per ProjectCollection instance?

For as long as I've been a software developer, global variables have been considered an anti-pattern, so if MSBuild API is forcing us to use one, then it's an API design flaw which would be nice to get fixed.

jeffkl: > Do we have a plan for "memory leaks" for customers who close and open a LOT of different solutions/repos in a single VS session?

Can you expand more?  The class instances will eventually get garbage collected so I'm not sure if there's necessarily a memory leak.

> I'm ok with this as a temporary workaround, but are we planning to work with our partners to improve the API, so NuGet (and anything else that implements MSBuild's resolver API) can use a different "cache" per ProjectCollection instance?

This particular API we've been discussing on improving it but the MSBuild team has resource constraints like everybody else so I doubt there will be any major improvements in the near term.

jeffkl: I also went back and forth between these options:

1. Make GlobalJsonReader have a singleton **and** make the file cache static
2. Just make the file cache `static`
3. Just make access to `GlobalJsonReader` via the singleton

I decided on option # 1 just to be sure, does anyone have a different preference?

zivkan: > Can you expand more? The class instances will eventually get garbage collected so I'm not sure if there's necessarily a memory leak.

Being static, the class instances will not eventually get garbage collected, unless there's a `FileCache.Remove(FileInfo)` somewhere, or a `FileCache.Clear()`, which I don't see. Unless there's something else non-obvious going on. Like, does MSBuild unload the assembly/AppDomain when a solution is closed?

Taking a quick look at VS telemetry, unless I have an error in my query, in the last 24 hours, there are hundreds of sessions where over 200 different solutions were opened (I made sure to de-duplicate the solution path hash, so it was definitely not the same solution opened over and over). That does seem like a crazy number of solutions, so it's probably some kind of automation, not a human. But it's still a scenario that we have evidence is really happening, and if each of them had a global.json, then this code will now have a dictionary with over 200 items.

I don't know if DDRITs has performance tests for a "open more than 1 solution" scenario, and even if it does, I don't know if either of the solutions have a global.json, and even if both do, the cached data might be under the threshold to trigger a regression. But there's still a small chance that this will cause a DDRITs regression, because increased memory usage, due to the dictionary being static and items never getting removed.

jeffkl: This class contains a static `SettingsLoadingContext` as well, which keeps a cache of of `NuGet.config` files and so far that hasn't been flagged.  I have to admit the reason that the NuGetSdkResolver doesn't flag anything in DDRITs is because its never called in that scenario, the .NET SDK resolver returns a value first.

https://github.com/NuGet/NuGet.Client/blob/58e93986c821f71812dd9bcfed405d819b166de6/src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/NuGetSdkResolver.cs#L33-L35

If a user's solution contains no projects which use the NuGetSdkResolver, then this won't matter.  If they do, this will hold a dictionary of `FileInfo`, `DateTime`, and `Lazy<Dictionary<string, string>>` where the dictionary contains the name of an SDK and the version with less than 3 items 99% of the time.  If the user is loading solution files within the same repository, the cache would be reused as `global.json` should exist at the root of the repo and not one for each solution.  

In my opinion, having the dictionary hanging around until you close Visual Studio is better than reading `global.json` once per project load since that is expected to be very fast.

zivkan: > In my opinion, having the dictionary hanging around until you close Visual Studio is better than reading global.json once per project load since that is expected to be very fast.

I agree, and that's [basically what I said](https://github.com/NuGet/NuGet.Client/pull/5364#discussion_r1297500393):

> I'm ok with this as a temporary workaround, but are we planning to work with our partners to improve the API, so NuGet (and anything else that implements MSBuild's resolver API) can use a different "cache" per ProjectCollection instance?

Just because nobody has yet reported the memory leak as a bug doesn't mean that it's not a memory leak.

If MSBuild/project system doesn't have an issue to tell them that their design bug is forcing NuGet to choose between two performance bugs (either re-reading files multiple times, or a memory leak), then they might not be aware of the design bug.  Maybe someone familiar with MSBuild knows a trivial solution and can create a PR quickly. If we don't ask we much less likely to get an improvement.

All I'm asking for is a tracking issue so we have documentation that this is a known API design flaw

July's perfathon showed that many perf issues could be fixed by people outside the product team. I don't know if this issue is too complex to be fixed by someone who isn't an msbuild expert. I don't know if VS calls [ProjectCollection.UnloadAllProjects](https://learn.microsoft.com/en-us/dotnet/api/microsoft.build.evaluation.projectcollection.unloadallprojects?view=msbuild-17-netcore) on solution close. If it does, then MSBuild sending all SDK resolvers a notification will allow NuGet's to do `FileCache.Clear()`, and reset `SettingsLoadContext` to a new instance, then that doesn't sound like an unreasonably complex solution to these memory leaks.

aortiz-msft (approved):
LGTM but a follow-up issue/discussion with MsBuild is warranted as per @zivkan 's comment.

src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/GlobalJsonReader.cs
jeffkl: I have filed https://github.com/dotnet/msbuild/issues/9164

I also talked to someone on CPS and they were not concerned with the extra memory usage and it not being cleared.

zivkan (approved):

PR #4019, opened 2021-04-26
Cancel no longer needed HTTP requests from PM UI when package list changes

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/10801

Regression? Last working version:

## Description

* Add a cancellation token to the view model, passed to the search service
* Request cancellation on the token when the view model is no longer used
* Added new `NuGetSearchServiceReconnector` to abstract away the brokered service implementation details, and avoid multiple places needing to know about the brokered service, subscribe to the `AvailabilityChanged` event, unsubscribe from the event, keep a JTF just for responding to the availability change, etc.

## PR Checklist

- [X] PR has a meaningful title
- [X] PR has a linked issue.
- [X] Described changes

- **Tests**
  - [X] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [ ] Documentation PR or issue filled
  - **OR**
  - [X] N/A

PR #5368, opened 2023-08-19
Remove IReconnectingNuGetSearchService interface as it's redundant

### LINKED ISSUE #12826: Remove IReconnectingNuGetSearchService interface as it's redundant

nkolev92:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12826

## Description

This interface is redundant. All usages of the interface can be  replace with INuGetSearchService as they have the exact same methods. Note that the improvements from https://github.com/NuGet/NuGet.Client/pull/4019/files#diff-0d4ee4bec3aa57378df7629bd264908ae71c25c6049f0e8102c94a51f3180e54 still apply. 

cc @zivkan @jebriede 

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception - No-op changes. Existing tests will catch issues
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


jebriede (approved):
Great change which leverages the INuGetSearchService interface regardless of the underlying implementation and removes the extra interface. 🚀

zivkan (approved):
@nkolev92 
> This interface is redundant

@jebriede 
> Great change which leverages the INuGetSearchService interface regardless of the underlying implementation and removes the extra interface. 🚀

These comments are missing the intent of why I added it in the first place.

Code using `INuGetSearchService` directly (theretically) needs to handle service connection interruptions and reconnect. Using the brokered service directly, if the client disconnects from the server, it must reconnect, otherwise all methods calls will throw.

The problem with `INuGetSearchService service = new ReconnectingNuGetSearchService(..)`, is that when I see `service.DoSomething()` elsewhere in the code, I have no idea if `service` uses the automatically reconnecting service, or if it has a bug because it's not handling the scenario that the client disconnected from the server.

By having `service` use interface `IReconnectingNuGetSearchService`, we can be absolutely certain that it's not using the brokered service directly.

In other words, this pull request, in my opinion, decreases confidence that the code using `INuGetSearchService` handles disconnection events.

However, we run PM UI in-proc, and VS codespaces never shipped, so now it's a scenario that is no longer possible. The entire reconnecting service class can probably be deleted, not just the interface that it previously implemented.

donnie-msft (approved):
Removing the interface is pretty trivial, and I'm wondering if we have plans to remove all the service implementation added for code spaces? If not, this should be called out as a step on that work.

nkolev92:
>  I'm wondering if we have plans to remove all the service implementation added for code spaces? If not, this should be called out as a step on that work.

We're not actively removing it, so not part of a larger effort. I think this change is a good change even with codespaces.

PR #5372, opened 2023-08-22
make the method PatterExpression.Match internal

Nigusu-Allehu:
## Bug

Fixes:

### Update
- As the method is an internal helper function, we have agreed to make the method internal.
Adds documentation comments that indicate the unit tests in NuGet.Core.UnitTests/NuGet.Packaging.Test/ContentModelTests cover PatterExpression.Match
- The comments added will allow the test cases to be visible in visual studio from the references list of PatternExpression.Match 
### Update 2.0
- We agreed in team review it is rather a better idea to make the method an internal method. The reason is because the method is just a utility method being used by other public methods that already have unit test cases.
## PR Checklist

- [ ] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A


test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,185 @@
...
+            //this pattern has tfm as a token segment and is not match only
+            //As a result tfm_raw property will be saved for the content item token matching with tfm
+            PatternDefinition pattern = new PatternDefinition("contentItem/{tfm}");
+            PatternExpression expression = new PatternExpression(pattern);
+            var path = "contentItem/this/is/tfm/raw/file.txt";
+            var expectedResult = "this/is/tfm/raw/file.txt";
+            ContentItem matchResult = expression.Match(path, propertyDefinitions);
+            Assert.NotNull(matchResult);
+            matchResult.Properties.TryGetValue("tfm_raw", out object result);
+            Assert.Equal(expectedResult, result);
```
zivkan: This feels wrong to me, or at least very confusing, since `contentItem/this/is/tfm/raw/file.txt` doesn't really conform to the `contentItem/` convention that package authors have to use, which makes it harder for me to understand what's going on. I think the problem is the pattern, `contentItem/{tfm}`, it's an unrealistic pattern that we don't use in our product code. 

Looking through `ManagedCodeConventions.cs`, all instances of `{tfm}` is followed by more pattern, often `{any?}`. So the important behaviour for `PatternExpression` to support is that a pattern `contentFiles/{codeLanguage}/{tfm}/{any?}` will match `contentFiles/any/net8.0/subdirectory/file.ext` and that `{codeLanguage}` maps to `any` (not `any/net8.0`), `{tfm}` (or `tfm_raw`) maps to `net8.0` (and not `net8.0/subdirectory`), and the {any?}` maps to `subdirectory/file.ext` (and not just `file.ext`).

But `/` is not a valid character for a TFM, so asserting that `tfm_raw`'s value is `this/is/tfm/raw/file.txt` confuses me, given my knowledge of TFMs, and how asset selection works.

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,185 @@
...
+using NuGet.ContentModel;
+using NuGet.ContentModel.Infrastructure;
+using Xunit;
+
+namespace NuGet.Packaging.Test
+{
+    public class PatternExpressionMatchTests
+    {
+        [Fact]
+        public void Match_LiteralSegment_ShouldMatch()
```
zivkan: I feel like this test isn't testing something very useful, and in fact is probably exposing an API design bug in the current implementation, rather than something we should fail a test and prevent a refactoring from "improving".

When I look at `ManagedCodeConventions.cs`, I don't see any `new PatternDefinition` that doesn't also use `{name}`. This test is testing the "pattern" `content/file.txt`, which means it's not actually detecting a pattern, it's detecting a single string.  Someone using `PatternExpression` to find a string literal is just losing performance, since they could iterate the file list and use `string.Equals` much more efficiently.

As I've gotten older and more experienced, my testing philosophy has changed a few times through my career, but currently my philosophy is "test the contract, not the implementation". Ideally, it should be possible for someone to delete the product code, and re-implement it from scratch, and once they get the code passing all these tests, the product works. As described above, I don't think this is asserting behaviour that we **need** from a re-implementation, or a refactoring of the product code. Nothing's perfect, especially not NuGet.Client, but we can still try to work towards whatever we think is best, given our current knowledge and experience.

I think it would be very reasonable for `PatternDefinition` to throw an `ArgumentException` saying that the input string isn't a valid pattern, when it doesn't contain any `{name}` variables (or mismatched number of `{` and `}`, or if we have invalid characters in the name, like `{some/name}`). To be clear, I'm not suggesting that you add tests that validate this, or change the product code to throw these exceptions. I just think from an API design point of view, it would make sense for the constructor to do so. Therefore, I don't think we should write tests that prevent it, unless we have a good/strong backwards compatibility reason to do so.

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,185 @@
...
+
+        [Fact]
+        public void Match_LiteralSegment_ShouldNotMatch()
+        {
+            // should not match the literal segment pattern because file.txt != file2.txt
+            Dictionary<string, ContentPropertyDefinition> propertyDefinitions = new()
+            {
+                { "type", new ContentPropertyDefinition("text") }
+            };
+            PatternDefinition pattern = new PatternDefinition("content/file.txt");
```
zivkan: similar to my other comments, I find this pattern too unrealistic for what our product code needs `PatternDefinition` for, so I think it would be better if the pattern was something like `content/{any?}`, and the filename it tries to match doesn't start with `content/`, so maybe `include/file.txt`.

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,185 @@
...
+            var path = "content/file.txt";
+            ContentItem matchResult = expression.Match(path, propertyDefinitions);
+            Assert.Null(matchResult);
+        }
+
+        [Fact]
+        public void Match_TokenSegmentMatchOnly_ShouldNotAddProperty()
+        {
+            // We have a parser that would match any field however the field "file.txt" here should not be added to the key "name" in the content Item
+            // because we used ? in the pattern for the token name and as a result the token is a matchonly token.
```
zivkan: What does "matchonly token" mean?

I'm not saying this is wrong. The test is passing, so clearly it's how the code works. But I find it confusing. When I look at the real patterns we use in our product code, `ManagedCodeConventions.cs`, we have a bunch of patterns that end in `{any?}`. Sometimes we use it like `contentFiles/{codeLanguage}/{tfm}/{any?}`, which I would expect to match both `contentFiles/any/any/file.txt` and also `contentFiles/any/any/subdirectory/file.txt`, but this test is asserting that it doesnt' match files in subdirectories?

edit: actually, it looks like the test is asserting that the expression matches the file, but the `{name}` "variable" didn't match anything. I think this comes back to my previous comment about "test the contract, not the implementation". I don't understand what user-experience this test is validating, it feels like an edge case that doesn't make sense.

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,185 @@
...
+        public void Match_TokenSegmentCustomDelimiter_ShouldMatch()
+        {
+            Dictionary<string, ContentPropertyDefinition> propertyDefinitions = new()
+            {
+                { "name", new ContentPropertyDefinition("any", parser: (o, t) => o) },
+                { "version", new ContentPropertyDefinition("any", parser: (o, t) => o) }
+            };
+            PatternDefinition pattern = new PatternDefinition("contentItem/{name}--{version}");
+            PatternExpression expression = new PatternExpression(pattern);
+            var path = "contentItem/mypackage--1.0.0/file.dll";
```
zivkan: I think this might be another example of unrealistic inputs. In `ManagedCodeConventions.cs` we have some patterns where we don't want the pattern to match files that match the pattern, except are actually in subdirectories. For example, `lib/{tfm}/{assembly}`. This not should match `lib/net7.0/en-au/my.resources.dll`, but it should match `lib/net7.0/my.dll`. Other patterns, that end in `{any?}` usually do want to match files in subdirectories.

I'm not sure what this test is trying to test. I mean, the has has "custom delimiter" in the name. THat's not something NuGet itself uses, so I think it's questionable usefulness. But, testing the pattern `contentItem/{name}--{version}` against the file `contentItems/mypackage--1.0.0/file.dll`, I'm not sure what to expect, so I don't know if this is correct, or if both the product and test have a bug.

zivkan (commented):
I don't know the product code that this test is validating, so the problem might just be with my own expectations. But with the knowledge I have of NuGet as a product, I'm not sure about these tests.

Also, I think there might be missing tests, validating that `lib/{tfm}/{assembly}` does not match `lib/net7.0/directory/my.dll`. Looking at the tests below, I have the impression that the file will actually match the pattern. Again, I don't know this code, so maybe the pattern does match, and there's some code somewhere else that removes the files in a subdirectory from the asset selection. But it feels like something I'd expect to be in the pattern matching code this code is testing.

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
aortiz-msft: I think we should have a UT for the simplest positive scenario or "golden path". Which test pattern and expression would be that?

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,185 @@
...
+            var path = "content/file.txt";
+            PatternDefinition pattern = new PatternDefinition(path);
+            PatternExpression expression = new PatternExpression(pattern);
+            ContentItem result = expression.Match(path, propertyDefinitions);
+            Assert.NotNull(result);
+            Assert.Equal(path, result.Path);
+        }
+
+        [Fact]
+        public void Match_LiteralSegment_ShouldNotMatch()
```
aortiz-msft: I'm not finding this name helpful to understand what the test is validating without reading its implementation. What about Match_LiteralSegment_VerifyNullIfNotFound?

aortiz-msft: Similarly, can we name other tests in a more specific way that shows the expected result/value? For example, for positive cases, Match_LiteralSegment_ReturnsContentItemCorrectly.

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
```diff
@@ -0,0 +1,185 @@
...
+        [Fact]
+        public void Match_LiteralSegmentAndTokenSegment_ShouldMatch()
+        {
+            //We have a parser for the token "name" which takes in any type of string and returns it.
+            //Therefore file.txt should be added to the result content item
+            Dictionary<string, ContentPropertyDefinition> propertyDefinitions = new()
+            {
+                { "name", new ContentPropertyDefinition("any", parser: (o, t) => o) }
+            };
+            //this pattern has both a token segment and a literal segment
```
aortiz-msft: NIT: Generally, an empty line should precede a line comment and succeed a closing bracket.

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/PatternExpressionMatchTests.cs
zivkan: [The pattern `lib/{assembly}`](https://github.com/NuGet/NuGet.Client/blob/0f33089a7bf04ff7a9ac94637d4a2462b942f777/src/NuGet.Core/NuGet.Packaging/ContentModel/ManagedCodeConventions.cs#L438), matching something like `lib/contoso.dll`, is the shortest I can find in ManagedCodeConventions.cs. [`build/{msbuild}`](https://github.com/NuGet/NuGet.Client/blob/0f33089a7bf04ff7a9ac94637d4a2462b942f777/src/NuGet.Core/NuGet.Packaging/ContentModel/ManagedCodeConventions.cs#L496) is also short.

Nigusu-Allehu:
As the method is tested thoroughly by the tests in NuGet.Core.UnitTests/NuGet.Packaging.Test/ContentModelTests, these tests have been marked redundant and unnecessary. Therefore, I have added a comment in the public Match method indicating that it has already been tested and where.

aortiz-msft (changes requested):
Should the comment be placed before the method definition?

Nigusu-Allehu:
@zivkan @aortiz-msft Please let me know what you think of the latest iteration

donnie-msft (commented):
Are we just adding a comment now? I'd suggest the PR be renamed, if so. And, can we avoid this comment, and instead just reference `Match` directly from the test... happy to chat offline about what I mean.

Nigusu-Allehu:
> Are we just adding a comment now? I'd suggest the PR be renamed, if so. And, can we avoid this comment, and instead just reference `Match` directly from the test... happy to chat offline about what I mean.

That is a great idea! Thank you, Donnie!

test/NuGet.Core.Tests/NuGet.Packaging.Test/ContentModelTests/ContentModelBuildTests.cs
```diff
@@ -4,14 +4,16 @@
...
 using NuGet.ContentModel;
 using NuGet.Frameworks;
 using NuGet.RuntimeModel;
 using Xunit;
 
 namespace NuGet.Client.Test
 {
+    /// <summary>
+    /// These unit test cases cover <see cref="ContentModel.Infrastructure.PatternExpression.Match(string, IReadOnlyDictionary{string, ContentPropertyDefinition})"/>
+    /// </summary>
```
zivkan: Two things::

1. My assumption is that you created your previous PR, that is now closed, because you looked at `PatternExpression.cs`, and couldn't find tests, so you spent time writing tests. While this comment makes it possible to use "find in all files" to look for the class name, I think it would be more discoverable if there was a comment in `PatternExpression.cs` and `PatternDefinition.cs`, saying that tests for the classes are located in `NuGet.Packaging.Test/ContentModelTests/`

2. The `///` comments are shown by Intellisense when you hover over types symbols, and some other scenarios. While it's not a "problem" here, because there's no reason for anyone to reference a test class from any other file, I don't think it's useful information to show in intellisense. So, if you add comments to `PatternDefinition.cs` and `PatternExpression.cs`, I recommend just using `//` (two slash) comments.

Nigusu-Allehu: These are two good points. @donnie-msft what do you think?

donnie-msft: 1. Finding references is far superior to me than text based searching. If Andy is suggesting that it's important to have a comment in both the caller and receiver, then you can consider having both comments, but I see no reason to remove this one, regardless.

2. I think it's useful information to link a type being tested directly to the test, when otherwise, there's no way to know without following the call stack manually. If you use 2 slashes, you'll not be able to reference the type in the comment in a way that VS can detect it.

src/NuGet.Core/NuGet.Packaging/ContentModel/Infrastructure/Parser.cs
```diff
@@ -63,8 +63,8 @@ private void Initialize(string pattern)
                 scanIndex = endToken + 1;
             }
         }
-        
-        //This method is tested by the unit test cases in test/NuGet.Core.UnitTests/NuGet.Packaging.Test/ContentModelTests
+
+
```
donnie-msft: Just have 1 blank line

donnie-msft (dismissed):
👍

aortiz-msft (changes requested):
Looking at the latest iteration, it might make sense to make the Match method internal or private.  I believe you were going to look into that.

aortiz-msft (approved):

PR #5374, opened 2023-08-23
Ensure top-level package isn't duplicated in new source mappings

### LINKED ISSUE #12839: Package Source Mapping utility always appends package ID
### NuGet Product Used

### Product Version

17.8

### Worked before?

_No response_

### Impact

None

### Repro Steps & Context

The logic in https://github.com/NuGet/NuGet.Client/blob/a39baac/src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs#L45 always Appends the package ID.
I'm not sure if this causes a problem today, but in a subsequent PR, I want to have an accurate count of newly mapped packages. So, I'd like this to only append package ID if it's not already included.

### Verbose Logs

_No response_

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12839

## Description

I'm not sure if this causes a problem today, but in a subsequent PR, I want to have an accurate count of newly mapped packages. So, I'd like this to only append package ID if it's not already included.

## PR Checklist

- [X] PR has a linked issue.
- [X] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [X] Test exception - existing tests look at the end result, which shouldn't change with this refactoring.
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [X] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
@@ -3,7 +3,6 @@
 
 #nullable enable
 
-using System;
 using System.Collections.Generic;
 using System.Linq;
 using NuGet.Configuration;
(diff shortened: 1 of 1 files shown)


src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -41,9 +43,12 @@ internal static class PackageSourceMappingUtility
...
             // Always include the Package ID being installed since it takes precedence over any globbing.
-            string[] packageIdsNeedingNewSourceMappings = addedPackageIdsWithoutExistingMappings
-                .Append(userAction.PackageId)
-                .ToArray();
+            if (!topLevelPackageIdWasAdded)
+            {
+                addedPackageIdsWithoutExistingMappings.Add(userAction.PackageId);
+            }
+
+            string[] packageIdsNeedingNewSourceMappings = addedPackageIdsWithoutExistingMappings.ToArray();
```
zivkan: `.ToArray()` causes an allocation. Can you change `CreateAndSavePackageSouyrceMappings` to use ` List<string>` instead of `string[]`? It's private, and this assembly isn't part of our public API anyway, so the variable shouldn't need to be an array, unless it calls a public API later on that does need to be an array.

donnie-msft: Sure, I can go ahead and refactor that in this PR.

aortiz-msft (dismissed):
:shipit:

donnie-msft:
Thanks for the reviews. I've integrated this change into https://github.com/NuGet/NuGet.Client/pull/5377 which supersedes this PR.

PR #5375, opened 2023-08-24
Update VS Options add/remove package source icons to VS2022 styling

### LINKED ISSUE #12840: VS Options add/remove package source icons aren't using VS2022 styling
I've been spending a lot of time lately in the NuGet Package source feed location and the icons are not current UX for VS2022.
55f95066-71e1-4c34-8318-b73ff766583e

These should be updated using the add/cancel (and I see the image list uses moveUp/Down as well but don't know where those are used) using the known monikers from VS2022.  In doing so it will result in:
image

It is a small change but makes the UI up to standard images used if still going to use images in this dialog.

The Up/Down are also updated:
image

timheuer:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12840

Regression? No

## Description
This updates only the imagelist16/32/64 base64 encoded image for up/down/cancel/add images used in Package Source UI in VS2022.  Source of new images are from the VS2022 KnownMonikers for MoveUp, MoveDown, Add, Cancel and match visuals of these type of actions elsewhere in VS2022.

Before:
55f95066-71e1-4c34-8318-b73ff766583e

After:
image

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


zivkan:
~Using the "KnownMonikers Explorer" extension, I see that VS already has icons for Add, Cancel, MoveUp, and MoveDown.~ You pointed out in the PR description that these icons already exist in VS as KnownMonikers. VS extensibility also have docs on how to use it in WinForms: https://learn.microsoft.com/en-us/visualstudio/extensibility/image-service-and-catalog?view=vs-2022#how-do-i-update-winforms-ui

So, should we delete these icons from NuGet's resx, and use VS's built-in icons?

timheuer:
> So, should we delete these icons from NuGet's resx, and use VS's built-in icons?

Honestly probably, but also I PR'd probably the smallest impact change...just the base64, no code, no DPI change, etc. -- I'd imagine changing to other means would be more code and then y'all would question the value of the change.  I felt this was simplest to get there.

aortiz-msft (approved):
Thank you for the contribution.

donnie-msft (approved):
Tried it out and looks good! Thanks for helping us out!
image

donnie-msft (changes requested):
The issue this closes mentions arrow icons, which I'm not sure I see in my testing. Left a comment here https://github.com/NuGet/Home/issues/12840#issuecomment-1694056618

Once that's clarified, I'll merge. Thanks again!

donnie-msft (approved):

PR #5377, opened 2023-08-25
Telemetry to Count Top Level & Transitive source mappings automatically created

### LINKED ISSUE #12839: Package Source Mapping utility always appends package ID

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2450
Fixes: https://github.com/NuGet/Home/issues/12839

## Description

- Telemetry will now have 2 integer properties for **counts of added top-level and transitive package source mappings**.
 - Top-level values can either be 0 or 1 in the current implementation (may become >1 in future iterations)
 - In scenarios when automatic mapping is not possible, the properties are not created (will be missing from telemetry).
  - **See description of Issue for telemetry samples**
  - Implements more of Assessment: https://github.com/NuGet/Client.Engineering/blob/main/designs/telemetry/assessment-source-mapping.md
- Secondarily, some refactoring is done in the utility class to return the counts
- Thirdly, the utility class implementation has changed
  - Determine in `ConfigureNewPackageSourceMapping` whether the top-level is already mapped, to avoid telemetry emitting `count:1` for top-level when it's already mapped.
  - Remove an unnecessary array.
  
## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
@@ -12,53 +12,80 @@ namespace NuGet.PackageManagement.UI
 {
     internal static class PackageSourceMappingUtility
     {
+        /// <summary>
+        /// Determines if new package source mappings should be written to the <paramref name="sourceMappingProvider"/> which is returned as a count.
+        /// </summary>
+        /// <param name="userAction"></param>
+        /// <param name="addedPackageIds"></param>
+        /// <param name="sourceMappingProvider"></param>
+        /// <param name="existingPackageSourceMappingSourceItems"></param>
+        /// <param name="countCreatedTopLevelSourceMappings">For Top-Level packages: <c>null</c> if not applicable; 0 if none needed to be added; > 0 is the count of new package source mappings added.</param>
+        /// <param name="countCreatedTransitiveSourceMappings">For Transitive packages: <c>null</c> if not applicable; 0 if none needed to be added; > 0 is the count of new package source mappings added.</param>
         internal static void ConfigureNewPackageSourceMapping(
             UserAction? userAction,
             IReadOnlyList<string>? addedPackageIds,
             PackageSourceMappingProvider sourceMappingProvider,
-            IReadOnlyList<PackageSourceMappingSourceItem> existingPackageSourceMappingSourceItems)
+            IReadOnlyList<PackageSourceMappingSourceItem> existingPackageSourceMappingSourceItems,
+            out int? countCreatedTopLevelSourceMappings,
+            out int? countCreatedTransitiveSourceMappings)
         {
+            countCreatedTopLevelSourceMappings = null;
+            countCreatedTransitiveSourceMappings = null;
+
             if (userAction?.SelectedSourceName is null || addedPackageIds is null)
             {
                 return;
             }
 
+            countCreatedTopLevelSourceMappings = 0;
+            countCreatedTransitiveSourceMappings = 0;
+
+            string topLevelPackageId = userAction.PackageId;
             Dictionary<string, IReadOnlyList<string>> patternsReadOnly = existingPackageSourceMappingSourceItems
                 .ToDictionary(pair => pair.Key, pair => (IReadOnlyList<string>)(pair.Patterns.Select(p => p.Pattern).ToList()));
 
             PackageSourceMapping packageSourceMapping = new(patternsReadOnly);
 
             // Expand all patterns/globs so we can later check if this package ID was already mapped.
-            List<string> addedPackageIdsWithoutExistingMappings = new();
+            List<string> addedPackageIdsWithoutExistingMappings = new(capacity: addedPackageIds.Count + 1);
+
             foreach (string addedPackageId in addedPackageIds)
             {
                 IReadOnlyList<string> configuredSource = packageSourceMapping.GetConfiguredPackageSources(addedPackageId);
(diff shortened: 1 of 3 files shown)


src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -12,53 +12,80 @@ namespace NuGet.PackageManagement.UI
...
                 IReadOnlyList<string> configuredSource = packageSourceMapping.GetConfiguredPackageSources(addedPackageId);
-                if (configuredSource.Count == 0)
+
+                // Top-level package was looked up.
+                if (addedPackageId == topLevelPackageId)
+                {
+                    // The top-level package is not already mapped to the selected source.
+                    if (configuredSource.Count == 0 || !configuredSource.Contains(userAction.SelectedSourceName))
+                    {
+                        countCreatedTopLevelSourceMappings = 1;
```
donnie-msft, line 62: The vision I have here is when multiple top-levels can be passed in and mapped, the count will become:
```suggestion
                        countCreatedTopLevelSourceMappings++;
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -12,53 +12,80 @@ namespace NuGet.PackageManagement.UI
...
+        /// <param name="existingPackageSourceMappingSourceItems"></param>
+        /// <param name="countCreatedTopLevelSourceMappings">For Top-Level packages: <c>null</c> if not applicable; 0 if none needed to be added; > 0 is the count of new package source mappings added.</param>
+        /// <param name="countCreatedTransitiveSourceMappings">For Transitive packages: <c>null</c> if not applicable; 0 if none needed to be added; > 0 is the count of new package source mappings added.</param>
         internal static void ConfigureNewPackageSourceMapping(
             UserAction? userAction,
             IReadOnlyList<string>? addedPackageIds,
             PackageSourceMappingProvider sourceMappingProvider,
-            IReadOnlyList<PackageSourceMappingSourceItem> existingPackageSourceMappingSourceItems)
+            IReadOnlyList<PackageSourceMappingSourceItem> existingPackageSourceMappingSourceItems,
+            out int? countCreatedTopLevelSourceMappings,
```
nkolev92, line 29: The 2 out parameters instinctively made me try to think of a better method separation. 

Would calling CreateAndSavePackageSourceMappings in a different method allow you to maybe just return the new source mapping items before saving them?
```cs
CreateAndSavePackageSourceMappings(
                userAction.SelectedSourceName,
                addedPackageIdsWithoutExistingMappings,
                sourceMappingProvider,
                existingPackageSourceMappingSourceItems);
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -12,53 +12,80 @@ namespace NuGet.PackageManagement.UI
...
                     addedPackageIdsWithoutExistingMappings.Add(addedPackageId);
                 }
             }
 
-            // Get all newly added package IDs that were not previously Source Mapped.
-            // Always include the Package ID being installed since it takes precedence over any globbing.
-            string[] packageIdsNeedingNewSourceMappings = addedPackageIdsWithoutExistingMappings
-                .Append(userAction.PackageId)
-                .ToArray();
+            countCreatedTransitiveSourceMappings = addedPackageIdsWithoutExistingMappings.Count - countCreatedTopLevelSourceMappings;
```
aortiz-msft, line 73: The problem with this calculation is that, if there's a bug in the logic in this function, telemetry would not help us understand what the bug is. Instead, I suggest incrementing the counter after line 61 (within the else if) and reporting that.

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
aortiz-msft: That makes sense to me, but I suggest addressing this comment in a subsequent PR so we can merge this PR in the next couple of days.

aortiz-msft (approved):

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
donnie-msft: Chatted with Nikolche offline: There's no simple refactoring that's meaningful right now at improving readability without significant changes. The goal of the utility from earlier feedback was to not continue growing the logic of within the UIActionEngine.

nkolev92 (approved):

PR #5378, opened 2023-08-27
Add a retry in MockServer Start method

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2451

## Description

Added a retry to `MockServer.Start()` because some tests fail intermittently when opening the connection.  Since the creation of the object must be in the retry loop, I also had to remove the `Listener` property which was being used by a few tests to set the `AuthenticationScheme` which I moved to the constructor instead.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/test/TestUtilities/Test.Utility/MockServer.cs
+++ b/test/TestUtilities/Test.Utility/MockServer.cs
@@ -27,9 +27,7 @@ namespace Test.Utility
         private Task _listenerTask;
         private bool _disposed = false;
         private AuthenticationSchemes _authenticationSchemes;
-#pragma warning disable CA2213 // Disposable fields should be disposed
         private HttpListener _listener;
-#pragma warning restore CA2213 // Disposable fields should be disposed
 
         public string BasePath { get; }
         private PortReserverOfMockServer PortReserver { get; }
@@ -92,7 +90,7 @@ namespace Test.Utility
                         throw;
                     }
 
-                    Thread.Sleep(50);
+                    Thread.Sleep(200);
                 }
             }
             while (_listener == null);
@@ -509,7 +507,7 @@ namespace Test.Utility
                 // Disposing the PortReserver
                 PortReserver.Dispose();
 
-                _listener = null;
+                (_listener as IDisposable)?.Dispose();
 
                 _disposed = true;
             }
--- a/test/TestUtilities/Test.Utility/MockServer.cs
+++ b/test/TestUtilities/Test.Utility/MockServer.cs
@@ -64,8 +64,6 @@ namespace Test.Utility
         /// </summary>
         public void Start()
         {
-            // Try creating and starting
-
             int attempts = 1;
             do
             {
--- a/test/TestUtilities/Test.Utility/MockServer.cs
+++ b/test/TestUtilities/Test.Utility/MockServer.cs
@@ -26,9 +26,12 @@ namespace Test.Utility
     {
         private Task _listenerTask;
         private bool _disposed = false;
+        private AuthenticationSchemes _authenticationSchemes;
+#pragma warning disable CA2213 // Disposable fields should be disposed
+        private HttpListener _listener;
+#pragma warning restore CA2213 // Disposable fields should be disposed
 
         public string BasePath { get; }
-        public HttpListener Listener { get; }
         private PortReserverOfMockServer PortReserver { get; }
         public RouteTable Get { get; }
         public RouteTable Put { get; }
@@ -43,19 +46,13 @@ namespace Test.Utility
         /// <summary>
         /// Initializes an instance of MockServer.
         /// </summary>
-        public MockServer()
+        /// <param name="authenticationSchemes">The optional <see cref="AuthenticationSchemes" /> to use.</param>
+        public MockServer(AuthenticationSchemes authenticationSchemes = AuthenticationSchemes.Anonymous)
         {
+            _authenticationSchemes = authenticationSchemes;
             BasePath = $"/{Guid.NewGuid().ToString("D")}";
 
             PortReserver = new PortReserverOfMockServer(BasePath);
-
-            // tests that cancel downloads and exit will cause the mock server to throw, this should be ignored.
-            Listener = new HttpListener()
-            {
-                IgnoreWriteExceptions = true
-            };
-
-            Listener.Prefixes.Add(PortReserver.BaseUri);
 
             Get = new RouteTable(BasePath);
             Put = new RouteTable(BasePath);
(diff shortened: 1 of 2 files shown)


test/TestUtilities/Test.Utility/MockServer.cs
```diff
@@ -26,9 +26,12 @@ public class MockServer : IDisposable
     {
         private Task _listenerTask;
         private bool _disposed = false;
+        private AuthenticationSchemes _authenticationSchemes;
+#pragma warning disable CA2213 // Disposable fields should be disposed
+        private HttpListener _listener;
+#pragma warning restore CA2213 // Disposable fields should be disposed
```
zivkan: This method already has a `Dispose` method, which you're using to set `_listener = null`.  Is there a reason you're not calling `_listener.Dispose()` before setting the field to null?

zivkan (dismissed):
LGTM, although it seems trivial to avoid needing to suppress the CA2213 analyzer. Am I missing something? If so, perhaps it's worth commenting why it's being suppressed, rather than followed.

test/TestUtilities/Test.Utility/MockServer.cs
```diff
@@ -69,7 +66,37 @@ public MockServer()
...
+                catch (Exception)
+                {
+                    _listener = null;
+
+                    if (attempts++ >= 5)
+                    {
+                        throw;
+                    }
+
+                    Thread.Sleep(50);
```
aortiz-msft: This doesn't seem enough of a delay before a retry. I would do 200ms which would be a total of 1 sec with 5 retries.

test/TestUtilities/Test.Utility/MockServer.cs
```diff
@@ -43,20 +46,14 @@ public class MockServer : IDisposable
         /// <summary>
         /// Initializes an instance of MockServer.
         /// </summary>
-        public MockServer()
+        /// <param name="authenticationSchemes">The optional <see cref="AuthenticationSchemes" /> to use.</param>
+        public MockServer(AuthenticationSchemes authenticationSchemes = AuthenticationSchemes.Anonymous)
```
aortiz-msft, line 49: This param name should probably be singular: authenticationScheme

aortiz-msft (dismissed):

test/TestUtilities/Test.Utility/MockServer.cs
jeffkl: For some reason `HttpListener` does not have a `Dispose()` method?  Since its just test code I suppressed the warning.  I've pushed a commit to cast it to `IDisposable` instead which appears to make things happy.

test/TestUtilities/Test.Utility/MockServer.cs
jeffkl: In this case `AuthenticationSchemes` can be a bitwise combination and support multiple authentication schemes at the same time so it should be plural.

test/TestUtilities/Test.Utility/MockServer.cs
```diff
@@ -69,7 +64,37 @@ public MockServer()
...
         {
-            Listener.Start();
+            // Try creating and starting
+
+            int attempts = 1;
+            do
+            {
+                try
+                {
+                    // tests that cancel downloads and exit will cause the mock server to throw, this should be ignored.
```
aortiz-msft, line 73: We should log each attempt to know when the retries are taking place and helping or not. I'm going to be Requesting Changes for this comment.

aortiz-msft (changes requested):

test/TestUtilities/Test.Utility/MockServer.cs
```diff
@@ -69,7 +64,37 @@ public MockServer()
         /// </summary>
         public void Start()
         {
-            Listener.Start();
+            // Try creating and starting
```
donnie-msft: Nit: move to the summary comment block

donnie-msft (dismissed):

test/TestUtilities/Test.Utility/MockServer.cs
jeffkl: Where do you suggest these are logged?  All of this is in test infrastructure, I'd have to pass around an `ITestOutputHelper` which is a bigger change.  Since they're just tests, I'm not sure if anyone would ever see a log output anyway.

aortiz-msft: We decided to not address this comment for now. We will revisit if we keep seeing issues after the retry logic is merged.

aortiz-msft (dismissed):

jeffkl:
Wow okay so now `HttpListener.Dispose()` is throwing?


### Error message
```
System.Net.Sockets.SocketException : Address already in use
```
### Stack trace
```
  at System.Net.Sockets.Socket.Bind (System.Net.EndPoint localEP) [0x00043] in <3d05db74c0154ea0ba79ffba6c4417f9>:0 
  at System.Net.EndPointListener..ctor (System.Net.HttpListener listener, System.Net.IPAddress addr, System.Int32 port, System.Boolean secure) [0x00047] in <3d05db74c0154ea0ba79ffba6c4417f9>:0 
  at System.Net.EndPointManager.GetEPListener (System.String host, System.Int32 port, System.Net.HttpListener listener, System.Boolean secure) [0x0009d] in <3d05db74c0154ea0ba79ffba6c4417f9>:0 
  at System.Net.EndPointManager.RemovePrefixInternal (System.String prefix, System.Net.HttpListener listener) [0x00040] in <3d05db74c0154ea0ba79ffba6c4417f9>:0 
  at System.Net.EndPointManager.RemoveListener (System.Net.HttpListener listener) [0x00024] in <3d05db74c0154ea0ba79ffba6c4417f9>:0 
  at System.Net.HttpListener.Close (System.Boolean force) [0x00006] in <3d05db74c0154ea0ba79ffba6c4417f9>:0 
  at System.Net.HttpListener.System.IDisposable.Dispose () [0x00009] in <3d05db74c0154ea0ba79ffba6c4417f9>:0 
  at Test.Utility.MockServer.Dispose () [0x00023] in <a52e08fd343d4edda3084f3c5ffaa0ad>:0 
  at NuGet.CommandLine.Test.NuGetListCommandTest.ListCommand_SimpleV3_NoListEndpoint () [0x00130] in <02721ee766f04f2f84eb2e6a1258dfcc>:0 
  at (wrapper managed-to-native) System.Reflection.RuntimeMethodInfo.InternalInvoke(System.Reflection.RuntimeMethodInfo,object,object[],System.Exception&)
  at System.Reflection.RuntimeMethodInfo.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0006a] in <bab7d1a00376483b944db50cdc31e41d>:0 
```

nkolev92:
> Wow okay so now HttpListener.Dispose() is throwing?

Reading the code, this seems intentional 😮

aortiz-msft (dismissed):

donnie-msft (commented):
> Wow okay so now HttpListener.Dispose() is throwing?
> 
> Error message
> System.Net.Sockets.SocketException : Address already in use

The docs suggest `Close()` is preferred; wonder if that will resolve the exception?
https://learn.microsoft.com/en-us/dotnet/api/system.net.httplistener.system-idisposable-dispose?view=net-7.0#remarks

donnie-msft (approved):
I see you're already running CI for 485edaa5e68317ff2f16f38c6ce87bc2d2561b1f 🍀
I'm going to be optimistic and go ahead and approve :)
Another benefit of `Close` is that network diagnostics will get emitted from that. May be useful if we ever need to debug the tests.

zivkan (approved):

PR #5384, opened 2023-08-28
Show vulnerability indicators for transitive vulnerable packages

### LINKED ISSUE #8756: Show vulnerabilities in transitive packages for PackageReference type projects in PMUI
A developer should be able to see a package vulnerability indicator for any known vulnerability in a transitive package in their dependency graph. 95% of vulnerable dependencies are transitive ones and we should bring awareness to this to help the ecosystem shift left.

image

jebriede:
## Bug

Fixes: https://github.com/NuGet/Home/issues/8756

## Description

Fetch vulnerabilities database for all enabled sources from VulnerabilityInfoResource and check if transitive packages are vulnerable to show warning indicator by Installed tab and by transitive package in package list.

Updated the install / uninstall / update action code paths that log telemetry to include a count of transitive vulnerable packages and maximum severity of each package.

## PR Checklist

- [x] PR has a linked issue.
- [ ] Described changes

- **Tests**
  - [x] Automated tests added
Updated existing test for telemetry although the test does not fully cover the scenario so we should consider adding additional tests in the future. Code copied from AuditUtility does not have tests as we are going to refactor the code to be shareable and put it under test.

  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
@@ -557,6 +557,8 @@ namespace NuGet.PackageManagement.UI
                         nuGetUI?.RecommenderVersion,
                         nuGetUI?.TopLevelVulnerablePackagesCount ?? 0,
                         nuGetUI?.TopLevelVulnerablePackagesMaxSeverities?.ToList() ?? new List<int>(),
+                        nuGetUI?.TransitiveVulnerablePackagesCount ?? 0,
+                        nuGetUI?.TransitiveVulnerablePackagesMaxSeverities?.ToList() ?? new List<int>(),
                         existingPackages,
                         addedPackages,
                         removedPackages,
@@ -609,6 +611,8 @@ namespace NuGet.PackageManagement.UI
             (string modelVersion, string vsixVersion)? recommenderVersion,
             int topLevelVulnerablePackagesCount,
             List<int> topLevelVulnerablePackagesMaxSeverities,
+            int transitiveVulnerablePackagesCount,
+            List<int> transitiveVulnerablePackagesMaxSeverities,
             HashSet<Tuple<string, string, string?>>? existingPackages,
             List<Tuple<string, string>>? addedPackages,
             List<string>? removedPackages,
@@ -645,6 +649,8 @@ namespace NuGet.PackageManagement.UI
 
             actionTelemetryEvent["TopLevelVulnerablePackagesCount"] = topLevelVulnerablePackagesCount;
             actionTelemetryEvent.ComplexData["TopLevelVulnerablePackagesMaxSeverities"] = topLevelVulnerablePackagesMaxSeverities;
+            actionTelemetryEvent["TransitiveVulnerablePackagesCount"] = transitiveVulnerablePackagesCount;
+            actionTelemetryEvent.ComplexData["TransitiveVulnerablePackagesMaxSeverities"] = transitiveVulnerablePackagesMaxSeverities;
 
             // log the installed package state
             if (existingPackages?.Count > 0)
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/DetailControl.xaml.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/DetailControl.xaml.cs
@@ -192,8 +192,16 @@ namespace NuGet.PackageManagement.UI
                     nugetUi.RecommendedCount = model.RecommendedCount;
                     nugetUi.RecommendPackages = model.RecommendPackages;
                     nugetUi.RecommenderVersion = model.RecommenderVersion;
-                    nugetUi.TopLevelVulnerablePackagesCount = model.IsPackageVulnerable ? 1 : 0;
-                    nugetUi.TopLevelVulnerablePackagesMaxSeverities = new List<int>() { model.PackageVulnerabilityMaxSeverity };
+                    if (model is PackageDetailControlModel packageModel && packageModel.PackageLevel == PackageLevel.Transitive)
+                    {
+                        nugetUi.TransitiveVulnerablePackagesCount = model.IsPackageVulnerable ? 1 : 0;
+                        nugetUi.TransitiveVulnerablePackagesMaxSeverities = new List<int>() { model.PackageVulnerabilityMaxSeverity };
+                    }
+                    else
+                    {
+                        nugetUi.TopLevelVulnerablePackagesCount = model.IsPackageVulnerable ? 1 : 0;
+                        nugetUi.TopLevelVulnerablePackagesMaxSeverities = new List<int>() { model.PackageVulnerabilityMaxSeverity };
+                    }
                 });
         }
     }
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
@@ -23,6 +23,7 @@ using NuGet.PackageManagement.UI.ViewModels;
 using NuGet.PackageManagement.VisualStudio;
 using NuGet.Packaging.Core;
 using NuGet.ProjectManagement;
+using NuGet.Protocol.Core.Types;
 using NuGet.Resolver;
 using NuGet.Versioning;
 using NuGet.VisualStudio;
@@ -64,6 +65,7 @@ namespace NuGet.PackageManagement.UI
         private string _settingsKey;
         private IServiceBroker _serviceBroker;
         private bool _disposed = false;
+        private IPackageVulnerabilityService _packageVulnerabilityService;
 
         private PackageManagerInstalledTabData _installedTabTelemetryData;
 
@@ -158,6 +160,10 @@ namespace NuGet.PackageManagement.UI
             {
                 controller.PackageManagerControl = this;
             }
+
+            ISourceRepositoryProvider sourceRepositoryProvider = await ServiceLocator.GetComponentModelServiceAsync<ISourceRepositoryProvider>();
+            var sourceRepositories = sourceRepositoryProvider.GetRepositories();
+            _packageVulnerabilityService = new PackageVulnerabilityService(sourceRepositories, _uiLogger);
 
             var solutionManager = Model.Context.SolutionManagerService;
             solutionManager.ProjectAdded += OnProjectChanged;
@@ -896,7 +902,8 @@ namespace NuGet.PackageManagement.UI
                     (NuGet.VisualStudio.Internal.Contracts.ItemFilter)_topPanel.Filter,
                     searchText: searchText,
                     includePrerelease: IncludePrerelease,
-                    useRecommender: useRecommender);
+                    useRecommender: useRecommender,
+                    _packageVulnerabilityService);
 
                 var loadingMessage = string.IsNullOrWhiteSpace(searchText)
                     ? Resx.Resources.Text_Loading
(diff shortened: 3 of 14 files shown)


src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
```diff
@@ -991,13 +998,32 @@ private async Task<(int, int)> GetInstalledVulnerableAndDeprecatedPackagesCountA
...
-            PackageCollection installedPackages = await loadContext.GetInstalledPackagesAsync();
-
-            var installedPackageMetadata = await Task.WhenAll(
-                installedPackages.Select(p => GetPackageMetadataAsync(p, packageSources, token)));
-
             int vulnerablePackagesCount = 0;
             int deprecatedPackagesCount = 0;
+            PackageCollection installedPackageCollection = null;
+
+            if (Model.IsSolution)
```
aortiz-msft, line 1006: A comment clarifying that transitive deps are only displayed in Project-level PM UI would be useful here.

aortiz-msft (dismissed):

nkolev92:
Your fixes statement in your PR description is broken. You need a link in  the `Fixes: ` line.

jebriede:
> Your fixes statement in your PR description is broken. You need a link in the `Fixes: ` line.

image

There's a link in fixes. Can you clarify what you mean? Do you want the raw url without the description or what are you seeing differently?

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
```diff
@@ -638,6 +642,8 @@ internal static List<TelemetryEvent> ToTelemetryPackageList(List<Tuple<string, s
 
             actionTelemetryEvent["TopLevelVulnerablePackagesCount"] = topLevelVulnerablePackagesCount;
             actionTelemetryEvent.ComplexData["TopLevelVulnerablePackagesMaxSeverities"] = topLevelVulnerablePackagesMaxSeverities;
+            actionTelemetryEvent["TransitiveVulnerablePackagesCount"] = transitiveVulnerablePackagesCount;
+            actionTelemetryEvent.ComplexData["TransitiveVulnerablePackagesMaxSeverities"] = transitiveVulnerablePackagesMaxSeverities;
```
nkolev92, line 653: MaxSeverities being a list is...weird :D 
Max makes me think 1. I'm not sure what I'd be doing with a list anyways.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/DetailControl.xaml.cs
```diff
@@ -186,8 +186,16 @@ private void ExecuteUserAction(UserAction action, NuGetActionType actionType)
                     nugetUi.RecommendedCount = model.RecommendedCount;
                     nugetUi.RecommendPackages = model.RecommendPackages;
                     nugetUi.RecommenderVersion = model.RecommenderVersion;
-                    nugetUi.TopLevelVulnerablePackagesCount = model.IsPackageVulnerable ? 1 : 0;
-                    nugetUi.TopLevelVulnerablePackagesMaxSeverities = new List<int>() { model.PackageVulnerabilityMaxSeverity };
+                    if (model is PackageDetailControlModel packageModel && packageModel.PackageLevel == PackageLevel.Transitive)
+                    {
+                        nugetUi.TransitiveVulnerablePackagesCount = model.IsPackageVulnerable ? 1 : 0;
+                        nugetUi.TransitiveVulnerablePackagesMaxSeverities = new List<int>() { model.PackageVulnerabilityMaxSeverity };
```
nkolev92, line 198: I don't see why this needs to be a list.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
```diff
@@ -159,6 +161,10 @@ private async ValueTask InitializeAsync(PackageManagerModel model, INuGetUILogge
                 controller.PackageManagerControl = this;
             }
 
+            ISourceRepositoryProvider sourceRepositoryProvider = await ServiceLocator.GetComponentModelServiceAsync<ISourceRepositoryProvider>();
+            var sourceRepositories = sourceRepositoryProvider.GetRepositories();
+            _packageVulnerabilityService = new PackageVulnerabilityService(sourceRepositories, _uiLogger);
```
nkolev92, line 166: Sources can change. You probably want to listen to settings changes and reload the sources.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/SourceRepositoryExtensions.cs
```diff
@@ -227,6 +229,21 @@ public static async Task<IPackageSearchMetadata> GetPackageMetadataForIdentityAs
...
+        public static async Task<GetVulnerabilityInfoResult> GetVulnerabilityInfoAsync(this SourceRepository sourceRepository, CancellationToken cancellationToken)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+            var vulnerabilityResource = await sourceRepository.GetResourceAsync<IVulnerabilityInfoResource>(cancellationToken);
+            if (vulnerabilityResource is null)
+            {
+                return null;
+            }
+
+            using (var sourceCacheContext = new SourceCacheContext())
```
nkolev92, line 241: Share the SourceCacheContext among source repositories.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/PackageVulnerabilityService.cs
```diff
@@ -0,0 +1,139 @@
...
+        private GetVulnerabilityInfoResult _vulnerabilities;
+        private INuGetUILogger _logger;
+
+        public PackageVulnerabilityService(IEnumerable<SourceRepository> sourceRepositories, INuGetUILogger logger)
+        {
+            _sourceRepositories = sourceRepositories ?? throw new ArgumentNullException(nameof(sourceRepositories));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger)); ;
+        }
+
+        public async Task<IEnumerable<PackageVulnerabilityMetadataContextInfo>> GetVulnerabilityInfoAsync(PackageIdentity packageId, CancellationToken cancellationToken)
```
nkolev92: Please use List to avoid constant allocations.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/PackageVulnerabilityService.cs
```diff
@@ -0,0 +1,139 @@
...
+            }
+
+            return packageVulnerabilities?.Select(pvi => new PackageVulnerabilityMetadataContextInfo(pvi.Url, (int)pvi.Severity));
+        }
+
+        private void ReplayErrors(AggregateException exceptions)
+        {
+            foreach (Exception exception in exceptions.InnerExceptions)
+            {
+                _logger.Log(ProjectManagement.MessageLevel.Warning, Strings.Error_VulnerabilityDataFetch, exception.Message);
```
nkolev92, line 135: Where does this go? OutputWindow? 
Some of the loggers in this space log to the activity log which is not very helpful.

nkolev92 (dismissed):
Seems reasonable. 

I have suggestions, but nothing that can't be addressed in a future PR.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
jebriede: Good observation. The history behind this is that some operations allow a list of packages. One such operation is Update. Install does not (currently). The telemetry is designed to put a count of vulnerable packages in the operation plus a list of the max severity per vulnerable package that was installed. The count of the list of max severities should equal the count of vulnerable packages in the operation.

There can be more than one vulnerability per package, but the count of vulnerable packages will be 1 per package that has at least 1 vulnerability. So the MaxSeverity is supposed to be the highest severity vulnerability per package.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/DetailControl.xaml.cs
jebriede: Rationale here: https://github.com/NuGet/NuGet.Client/pull/5384/files#r1308087183

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
nkolev92: So it's the max severity of each vulnerable package individually? 

I hate naming. 🤦

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/DetailControl.xaml.cs
nkolev92: Same as https://github.com/NuGet/NuGet.Client/pull/5384#discussion_r1308087183

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/PackageVulnerabilityService.cs
jebriede: @nkolev92 According to [this](https://github.com/NuGet/NuGet.Client/blob/545b3ce20365c91883518393c23988409988f20f/src/NuGet.Clients/NuGet.VisualStudio.Common/UserInterfaceService/INuGetUILogger.cs#L19-L23), it logs to the Error List for warnings and errors. This is logged as a warning. Do you think there's any action here or can this comment be resolved?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/SourceRepositoryExtensions.cs
jebriede: The static extension methods operate on a single source repository. In order to share SourceCacheContext among various source repositories, we would need to supply the SourceCacheContext along with each of the method calls in the SourceRepositoryExtensions. The current precedent set by this static class and its methods is to create a new SourceCacheContext per each call, which is consistent with what we're doing here.

An alternative design would be to store a static reference to the SourceCacheContext but then we'd need methods on this static class that manages the lifecycle of the SourceCacheContext. Ideally, we would have a better way to manage a collection of source repositories and their operations and share the SourceCacheContext across multiple repos.

I'm making note of your suggestion as an improvement that should be made, but it's out of scope of this feature as this simply follows an existing pattern using an existing mechanism which can be improved upon, but was not introduced in this change.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
jebriede: Made note of this for a follow-up. Thanks!

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
jebriede: Done, thanks!

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/PackageVulnerabilityService.cs
jebriede: Done, thanks!

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/SourceRepositoryExtensions.cs
nkolev92: > The current precedent set by this static class and its methods is to create a new SourceCacheContext per each call, which is consistent with what we're doing here.

I see. That's a poor pattern. See https://github.com/NuGet/NuGet.Client/commit/9ea2defeabc6ab8687f11298e314c197e95a6b8f. 
The caching context should be shared so that any caching decisions don't need to be changed in 10 different places. 
I think we should change that, I've created an issue for that https://github.com/NuGet/Home/issues/12859. 


> simply follows an existing pattern using an existing mechanism which can be improved upon, but was not introduced in this change.

I don't think we should be propagating bad patterns, even if it's a preexisting pattern. However, I think it's reasonable to say that all of those discussions should be handled on a case by case basis, as the cost/impact is different for each one individually.

My suggestion is to just pass down the source cache context, but I'm not too concerned about this single change, so feel free to make a call.

nkolev92 (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/SourceRepositoryExtensions.cs
jebriede: @nkolev92 thanks for creating the follow-up issue!

nkolev92 (approved):

aortiz-msft (approved):

PR #5387, opened 2023-08-30
Revert "Pass through the cache context to the vulnerability resource (#5361)"

zivkan:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2461

Regression? yes Last working version: before https://github.com/NuGet/NuGet.Client/commit/9ea2defeabc6ab8687f11298e314c197e95a6b8f

## Description

Preview install (which happens both for PM UI install, but also when other VS components call our IVsPackageInstaller.InstallPackage API), ignores the http-cache, so that newly published packages do not fail to install: 
https://github.com/NuGet/NuGet.Client/blob/e4a9922edc29c36e6458bc5d3b68a04e0ea14b91/src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs#L2935-L2941

(also see other references to `SourceCacheContext.MaxAge`)

However, that "newly published package" scenario does not apply to the vulnerability database. There's no reason to force customers to redownload the vulnerability database, so this reverts the previous commit.

This PR isn't a straight revert, it has two modifications:

- Keep the comments about the enum helpers in `AuditUtils`, because they still apply
- Add a big comment in `VulnerabilityInformationProvider.cs` explaining why we're using a different cache context.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception: reverting a commit that didn't have tests. Under time pressure to meet shipping dates, but we can consider some kind of complicated functional test to validate this scenario in the future.
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommandProvidersCache.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommandProvidersCache.cs
@@ -120,7 +120,7 @@ namespace NuGet.Commands
             var vulnerabilityInformationProviders = new List<IVulnerabilityInformationProvider>(sources.Count);
             foreach (SourceRepository source in sources)
             {
-                IVulnerabilityInformationProvider provider = _vulnerabilityInformationProviders.GetOrAdd(source, s => new VulnerabilityInformationProvider(s, cacheContext, log));
+                IVulnerabilityInformationProvider provider = _vulnerabilityInformationProviders.GetOrAdd(source, s => new VulnerabilityInformationProvider(s, log));
                 vulnerabilityInformationProviders.Add(provider);
             }
 
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommandProviders.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommandProviders.cs
@@ -39,12 +39,11 @@ namespace NuGet.Commands
         private static IReadOnlyList<IVulnerabilityInformationProvider> CreateVulnerabilityInfoProviders(IReadOnlyList<IRemoteDependencyProvider> remoteProviders)
         {
             List<IVulnerabilityInformationProvider> providers = new(remoteProviders.Count);
-            using var sourceCacheContext = new SourceCacheContext();
             for (int i = 0; i < remoteProviders.Count; i++)
             {
                 if (remoteProviders[i].SourceRepository != null)
                 {
-                    providers.Add(new VulnerabilityInformationProvider(remoteProviders[i].SourceRepository, sourceCacheContext, NullLogger.Instance));
+                    providers.Add(new VulnerabilityInformationProvider(remoteProviders[i].SourceRepository, NullLogger.Instance));
                 }
             }
             return providers;
@@ -151,7 +150,7 @@ namespace NuGet.Commands
             var vulnerabilityInfoProviders = new List<IVulnerabilityInformationProvider>(remoteProviders.Count);
             foreach (SourceRepository source in sources)
             {
-                var provider = new VulnerabilityInformationProvider(source, cacheContext, log);
+                var provider = new VulnerabilityInformationProvider(source, log);
                 vulnerabilityInfoProviders.Add(provider);
             }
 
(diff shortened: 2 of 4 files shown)


src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommandProvidersCache.cs
```diff
@@ -120,7 +120,7 @@ public class RestoreCommandProvidersCache
             var vulnerabilityInformationProviders = new List<IVulnerabilityInformationProvider>(sources.Count);
             foreach (SourceRepository source in sources)
             {
-                IVulnerabilityInformationProvider provider = _vulnerabilityInformationProviders.GetOrAdd(source, s => new VulnerabilityInformationProvider(s, cacheContext, log));
+                IVulnerabilityInformationProvider provider = _vulnerabilityInformationProviders.GetOrAdd(source, s => new VulnerabilityInformationProvider(s, log));
```
nkolev92, line 123: This is the only true problem correct? Because the SourceCacheContext here is the one passed down from "up above" which can be the NuGet Package Manager. 

IMO, we should still create a sourcecachecontext here, and not in the VulnerabilityInformationProvider. 

That way this change is basially a single line.

donnie-msft: 👀 sounds fine to me, too

donnie-msft (approved):

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommandProvidersCache.cs
zivkan: Reverting back to the dev branch and and having `RestoreCommandProvidersCache` create its own `SourceCacheContext` caused [CA2000](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2000). But this code only creates an instance of the `VulnerabilityInformationProvider`, which means it's not appropriate to dispose the `SourceCacheContext` when `RestoreCommandProvidersCache.GetOrAdd` returns. This class is part of our public APIs, so there's no good solution.

I opted to make the `SourceCacheContext` a field, but then I had to make the class implement `IDisposable` to avoid [CA1001](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1001). Now CA2000 and CA1001 triggered everywhere that creates `RestoreCommandProvidersCache` instances. It wasn't too bad, until `NuGet.PackageManagement.NuGetPackageManager` has a field holding an instance, so now it "needs" to become `IDisposable` as well. Looking at references to `NuGetPackageManager` constructors, there are 18 calls, although 3 are probably just from other overloads. This is where I decided this path was too much work.

Another idea I had to is change `RestoreCommandProvidersCache.GetOrAdd` to pass in 2 `SourceCacheContext`s, one just for vulnerabilities. However, looking at how the current cache context is passed in, it's coming from `RestoreArgs.CacheContext`. So do I need to add `VulnerabilityCacheContext` to `RestoreArgs`? That didn't seem like a good option.

My last idea is to add more properties to `SourceCacheContext`. I didn't investigate how much needs to change to make this works, but I expect it'll be a really bad change because ultimately `HttpSourceCacheContext` would need a `bool` or something to tell it to either use the existing properties, or to use the new (vulnerability only?) properties, almost all of which are public APIs.

In conclusion, I don't see a way to implement your proposal in a low-impact way (well, unless we don't dispose a disposable object, and let the GC dispose it, and suppress CA2000 in `RestoreCommandProvidersCache`).

If you, or someone else, wants to reduce memory allocations by reducing new instances of `SourceCacheContext`, then they can spend the effort solve the above problems. We're only creating one instance per source, not per project, and downloading over HTTP the vulnerability database has a higher customer impact than any potential GC from extra allocations of the `SourceCacheContext` class.

nkolev92: thanks for trying

PR #5388, opened 2023-08-30
Signing:  enable chain build retry policy by default (Windows only)

### LINKED ISSUE #12592: Signing:  enable `X509Chain.Build(...)` retry behavior by default
### NuGet Product(s) Affected

NuGet.exe, dotnet.exe

### Current Behavior

As of NuGet 6.0, NuGet has an environment variable which can be used to opt-in to `X509Chain.Build(...)` retry behavior on Windows only when the only failure is that the root certificate is untrusted.  See https://github.com/NuGet/NuGet.Client/pull/4259 for more context.

### Desired Behavior

We should enable this retry behavior by default with reasonable values and the same restrictions (i.e.:  untrusted status on Windows) while still allowing users to override or opt-out of the default.

For example, if we choose `3,1000` as the default, a user could override by setting the environment variable with different values:

```
set NUGET_EXPERIMENTAL_CHAIN_BUILD_RETRY_POLICY=10,3000
```

A user could also opt out (meaning, no retry at all) with a different value (TBD, but possibly just `0`):

```
set NUGET_EXPERIMENTAL_CHAIN_BUILD_RETRY_POLICY=0
```

A [GitHub search](https://github.com/search?q=NUGET_EXPERIMENTAL_CHAIN_BUILD_RETRY_POLICY&type=code) reveals `3,1000` as the most common value, with `9,2000` being next most common.

### Additional Context

The root problem is that a particular API call isn't retried, and it should be.  Whether or not Windows adds a retry, retrying is the only mitigation.  We can broadly mitigate this issue for NuGet users by retrying by default.

CC @joelverhagen

dtivel:
## Bug

Fixes:  https://github.com/NuGet/Home/issues/12592

Regression? Last working version:  no

## Description

This change enables the chain build retry policy described [here](https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu3028#retry-untrusted-root-failures) by default on Windows only.  The default value of `3,1000` means that chain building will be retried up to 3 times with a 1-second wait in-between retries.  This retry behavior only applies narrowly to chain builds which fail with an untrusted root status and only on Windows.  See [this](https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu3028#issue) for more context.

If users have already set the `NUGET_EXPERIMENTAL_CHAIN_BUILD_RETRY_POLICY` environment variable, its value will be honored as before this change.

Up until now, this option has been opt-in, but this change makes it opt-out.  To opt-out, simply set the environment variable value to `0`, like so:

```
set NUGET_EXPERIMENTAL_CHAIN_BUILD_RETRY_POLICY=0
```

## PR Checklist

- [X] PR has a linked issue.
- [X] Described changes

- **Tests**
  - [X] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [X] Documentation PR or issue filled:  https://github.com/NuGet/docs.microsoft.com-nuget/issues/3133
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Packaging/Signing/ChainBuilding/X509ChainBuildPolicyFactory.cs
+++ b/src/NuGet.Core/NuGet.Packaging/Signing/ChainBuilding/X509ChainBuildPolicyFactory.cs
@@ -8,7 +8,9 @@ namespace NuGet.Packaging.Signing
 {
     internal static class X509ChainBuildPolicyFactory
     {
+        private const string DefaultValue = "3,1000";
         // These fields are non-private only to facilitate testing.
+        internal const string DisabledValue = "0";
         internal const string EnvironmentVariableName = "NUGET_EXPERIMENTAL_CHAIN_BUILD_RETRY_POLICY";
         internal const char ValueDelimiter = ',';
 
@@ -45,9 +47,9 @@ namespace NuGet.Packaging.Signing
 
             if (RuntimeEnvironmentHelper.IsWindows)
             {
-                string value = reader.GetEnvironmentVariable(EnvironmentVariableName);
+                string value = reader.GetEnvironmentVariable(EnvironmentVariableName) ?? DefaultValue;
 
-                if (string.IsNullOrWhiteSpace(value))
+                if (string.Equals(value, DisabledValue, StringComparison.Ordinal))
                 {
                     return DefaultX509ChainBuildPolicy.Instance;
                 }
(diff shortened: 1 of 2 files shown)


nkolev92 (approved):

donnie-msft (approved):

aortiz-msft (approved):
Looks good. Is this needed for .NET 8 GA?

heng-liu (approved):
Shall we update the document at https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu3028 when this is released?

dtivel:
> Looks good. Is this needed for .NET 8 GA?

@aortiz-msft, I defer to you on this.  I feel that this change is pretty safe and will reduce friction for NuGet users on Windows, but it can go in whatever release makes sense to you.  As with any change, the more preview testing, the better.

@heng-liu, yep.  I already created https://github.com/NuGet/docs.microsoft.com-nuget/issues/3133.  (The PR template needed it.)

PR #5389, opened 2023-08-30
Add -NoHttpCache option and RestoreNoHttpCache property.

### LINKED ISSUE #9180: Rename no-cache to no-http-cache
Across all the tooling, dotnet.exe/nuget.exe/msbuild.exe, we have a --no-cache, -NoCache, RestoreNoCache switch to turn off our http cache caching, which includes a few things: 

- Caching the resources (thing caching repository signatures, servide index)
- Caching the versions list
- Caching the nupkgs 

This whole experience is rather confusing. A lot of .NET Customers think of the global packages folder as a cache and they expect RestoreNoCache to exclude the global packages folder. In reality this is a nuget http cache and should be clearly indicated as such. 

Related to https://github.com/NuGet/Home/issues/3116. 

cc @aortiz-msft @NuGet/nuget-client

Nigusu-Allehu:
## Bug  

Fixes: https://github.com/NuGet/Home/issues/9180

## Description

Spec: https://github.com/NuGet/Home/pull/12846
This PR will
- Add a new option `-NoHttpCache` for nuget.exe and `RestoreNoHttpCache` property for msbuild
- It will remove `-NoCache` from appearing in the help dialogue for Nuget
- It will log a message suggesting `-NoCache` or `RestoreNoCache`  are just aliases for `-NoHttpCache` and `RestoreNoHttpCache`
- The dotnet side for the option will be implemented here: https://github.com/dotnet/sdk/pull/35107
## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Clients/NuGet.CommandLine/NuGetCommand.resx
```diff
@@ -191,6 +191,9 @@ nuget push -?</value>
   <data name="CommandNoCache" xml:space="preserve">
     <value>Disable using the machine cache as the first package source.</value>
   </data>
+  <data name="CommandNoHttpCache" xml:space="preserve">
+    <value>Disable using Http cache as the package source.</value>
```
jeffkl:
```suggestion
    <value>Disable using the HTTP cache and contact all configured package sources for live information.</value>
```

src/NuGet.Clients/NuGet.CommandLine/NuGetCommand.resx
```diff
@@ -250,6 +253,10 @@ nuget list -verbose -allversions</value>
   <data name="ListCommandVerboseListDescription" xml:space="preserve">
     <value>Displays a detailed list of information for each package.</value>
   </data>
+  <data name="Log_RestoreNoCacheInformation" xml:space="preserve">
+    <value>Keep in mind that NoCache is solely for disabling Http Caching, not the global packages folder. For the same functionality with a more descriptive name, you can use NoHttpCache instead.</value>
```
jeffkl:
```suggestion
    <value>The NoCache option is only for disabling the HTTP cache and not the global packages folder. It is recommended to use the NoHttpCache option instead to disable the use of the HTTP cache and the global packages folder.</value>
```

src/NuGet.Clients/NuGet.CommandLine/NuGetCommand.resx
```diff
@@ -250,6 +253,10 @@ nuget list -verbose -allversions</value>
...
+    <comment>Do not transate NoCache and NoHttpCache</comment>
```
jebriede: Typo in "translate"

```suggestion
    <comment>Do not translate NoCache and NoHttpCache</comment>
```

src/NuGet.Clients/NuGet.CommandLine/Program.cs
```diff
@@ -129,6 +129,10 @@ public static int MainCore(string workingDirectory, string[] args)
                 // Parse the command
                 var command = parser.ParseCommandLine(args) ?? p.HelpCommand;
                 command.CurrentDirectory = workingDirectory;
+                if ((command is RestoreCommand restoreCommand && restoreCommand.NoCache == true) || (command is InstallCommand installCommand && installCommand.NoCache == true))
```
jebriede: Unless NoCache is a nullable bool, the "== true" is redundant.

```suggestion
                if ((command is RestoreCommand restoreCommand && restoreCommand.NoCache) || (command is InstallCommand installCommand && installCommand.NoCache))
```

jebriede: We are checking if the command is either a RestoreCommand or InstallCommand but in both cases we are looking at the same property `NoCache` which is implemented on the base class `DownloadCommandBase` so we can just treat the command as the abstract base class DownloadCommandBase and check if NoCache is set to true for it, rather than casting to each of the sub types to check the same property.

src/NuGet.Core/NuGet.Build.Tasks/RestoreTask.cs
```diff
@@ -140,11 +145,16 @@ private async Task<bool> ExecuteAsync(Common.ILogger log)
 
             EmbedInBinlog = GetFilesToEmbedInBinlog(dgFile);
 
+            if (RestoreNoCache)
+            {
+                log.LogWarning(Strings.Log_RestoreNoCacheInformation);
```
jeffkl: Team Review:  Do not introduce a new warning in a minor release.

jebriede: Is this going to cause the message to be output twice in some scenarios because we are outputting it once in Program.cs and once in RestoreTask? I haven't tested this so I'm not sure if this will, just wanting to understand why we log this warning once here and once when parsing the command.

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/ExtractOptionsTests.cs
```diff
@@ -0,0 +1,45 @@
...
+            ICommandManager manager = new CommandManager();
+            CommandLineParser commandLineParser = new CommandLineParser(manager);
+
+            //execute extract options
+            commandLineParser.ExtractOptions(command, args.GetEnumerator());
+
+            //assert
+            Assert.False(command.NoHttpCache);
+        }
+
```
jebriede: nit: remove extra blank line.

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/ExtractOptionsTests.cs
```diff
@@ -0,0 +1,45 @@
...
+            //assert
+            Assert.True(command.NoHttpCache);
+        }
+
+        [Fact]
+        public void ExtractOption_DoNotPassOptionNoHttpCache_NoHttpCacheShouldNotBeSet()
+        {
+            //setup
+            RestoreCommand command = new RestoreCommand();
+            List<string> args = new();
```
jebriede: With this change, you won't need `args.GetEnumerator()` on line 38.

```suggestion
            IEnumerable<string> args = Enumerable.Empty<string>();
```

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/ExtractOptionsTests.cs
```diff
@@ -0,0 +1,45 @@
...
+            ICommandManager manager = new CommandManager();
+            CommandLineParser commandLineParser = new CommandLineParser(manager);
```
jebriede: Since this is a unit test for the CommandLineParser, we should make sure that its dependencies are all mocked. It would be best if we passed a mock CommandManager.

If we don't, a more concise way to express this, since we never use the CommandManager in this test, is to set this up in one line.

```suggestion
            CommandLineParser commandLineParser = new CommandLineParser(new CommandManager());
```

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/ExtractOptionsTests.cs
```diff
@@ -0,0 +1,45 @@
...
+            commandLineParser.ExtractOptions(command, args.GetEnumerator());
+
+            //assert
+            Assert.True(command.NoHttpCache);
+        }
+
+        [Fact]
+        public void ExtractOption_DoNotPassOptionNoHttpCache_NoHttpCacheShouldNotBeSet()
+        {
+            //setup
```
jebriede: Nit

```suggestion
            // Arrange
```

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/ExtractOptionsTests.cs
```diff
@@ -0,0 +1,45 @@
...
+            RestoreCommand command = new RestoreCommand();
+            List<string> args = new();
+            ICommandManager manager = new CommandManager();
+            CommandLineParser commandLineParser = new CommandLineParser(manager);
+
+            //execute extract options
```
jebriede: Nit

```suggestion
            // Act
```

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/ExtractOptionsTests.cs
```diff
@@ -0,0 +1,45 @@
...
+            commandLineParser.ExtractOptions(command, args.GetEnumerator());
+
+            //assert
```
jebriede: Nit

```suggestion
            // Assert
```

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/ExtractOptionsTests.cs
```diff
@@ -0,0 +1,45 @@
...
+            Assert.True(command.NoHttpCache);
+        }
+
+        [Fact]
+        public void ExtractOption_DoNotPassOptionNoHttpCache_NoHttpCacheShouldNotBeSet()
```
jebriede, line 29: This test currently tests that RestoreCommand's default options are to set NoCache to false. However, the test name suggests that not passing an option to ExtractOption, will result in NoHttpCache being not set. Furthermore, we are testing ExtractOption and we cannot guarantee that the RestoreCommand default behavior will never change so we need to be explicit about how that is set up.

Let's change the name to reflect the fact we're testing ExtractOption with a command that has NoCache set to false, results in the NoHttpCache set to false.

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/ExtractOptionsTests.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            //setup
+            RestoreCommand command = new RestoreCommand();
```
jebriede: Ideally, we should not be relying on another object for a unit test here. Instead, we should be using a mock of the DownloadCommandBase (or RestoreCommand) and make sure it works appropriately given the proper parameters, for any kind of command of that type.

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/ExtractOptionsTests.cs
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Xunit;
+
+namespace NuGet.CommandLine.Test
+{
+    public class ExtractOptionsTests
```
jebriede: The name of a test class should correspond to the object or unit it is testing. In this case, we are testing the CommandLineParser. The test cases (or test methods), should test the individual methods on the object and should reflect the name of the method we are testing.

In other words, please rename this test class and its file to CommandLineParserTests.

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/ExtractOptionsTests.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+
+        [Fact]
+        public void ExtractOption_PassOptionNoHttpCache_NoHttpCacheShouldBeSet()
+        {
+            //setup
```
jebriede: Nit: we tend to use Arrange, Act, Assert as the three steps.
```suggestion
            // Arrange
```

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/ExtractOptionsTests.cs
```diff
@@ -0,0 +1,45 @@
...
+            RestoreCommand command = new RestoreCommand();
+            List<string> args = new() { "-NoHttpCache" };
+            ICommandManager manager = new CommandManager();
+            CommandLineParser commandLineParser = new CommandLineParser(manager);
+
+            //execute extract options
```
jebriede: Nit: we tend to use Arrange, Act, Assert as the three steps.

```suggestion
            // Act
```

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/ExtractOptionsTests.cs
```diff
@@ -0,0 +1,45 @@
...
+            commandLineParser.ExtractOptions(command, args.GetEnumerator());
+
+            //assert
```
jebriede: Nit: we tend to use Arrange, Act, Assert as the three steps.

```suggestion
            // Assert
```

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/ExtractOptionsTests.cs
```diff
@@ -0,0 +1,45 @@
...
+    public class ExtractOptionsTests
+    {
+
+        [Fact]
+        public void ExtractOption_PassOptionNoHttpCache_NoHttpCacheShouldBeSet()
+        {
+            //setup
+            RestoreCommand command = new RestoreCommand();
+            List<string> args = new() { "-NoHttpCache" };
+            ICommandManager manager = new CommandManager();
```
jebriede: Same feedback here as below regarding using a mock CommandManager or combining into one line.

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/ExtractOptionsTests.cs
```diff
@@ -0,0 +1,45 @@
...
+namespace NuGet.CommandLine.Test
+{
+    public class ExtractOptionsTests
+    {
+
+        [Fact]
+        public void ExtractOption_PassOptionNoHttpCache_NoHttpCacheShouldBeSet()
+        {
+            //setup
+            RestoreCommand command = new RestoreCommand();
```
jebriede: Same feedback regarding RestoreCommand below, applies here.

src/NuGet.Core/NuGet.Build.Tasks/Strings.resx
```diff
@@ -200,4 +200,8 @@
     <comment>0 - the path to the EXE being launched
 1 - The command-line arguments</comment>
   </data>
+  <data name="Log_RestoreNoCacheInformation" xml:space="preserve">
+    <value>Keep in mind that RestoreNoCache is solely for disabling Http Caching, not the global packages folder. For the same functionality with a more descriptive name, you can use RestoreNoHttpCache instead.</value>
+    <comment>Do not transate RestoreNoCache and RestoreNoHttpCache</comment>
```
jebriede:
```suggestion
    <comment>Do not translate RestoreNoCache and RestoreNoHttpCache</comment>
```

src/NuGet.Clients/NuGet.CommandLine/NuGetCommand.resx
jebriede: Please reach out to @albarry4 for help with wording.

jebriede (changes requested):
Great job! Left some comments and suggested test improvements. Please feel free to reach out for clarification on any of my comments!

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/ExtractOptionsTests.cs
jebriede: Let's make it explicit in the test case name what we are expecting. Unless we are relying on the default behavior of the command class when no option for NoHttpCache is being passed (in which case this test would need to change), we should change `ShouldNotBeSet` to tell us what exactly that means, which is that we are expecting it to be set to false.

First part of test case name should match method under test.

[Optional] Also suggesting update to middle part.

```suggestion
        public void ExtractOptions_WhenNotPassingOptionNoHttpCache_NoHttpCacheShouldBeFalse()
```

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/ExtractOptionsTests.cs
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Xunit;
+
+namespace NuGet.CommandLine.Test
```
jebriede: What happens if we pass a NoHttpCache option to ExtractOptions and a command that does not have the NoHttpCache option? Will the code fail gracefully? Should we add a test for that scenario?

src/NuGet.Core/NuGet.Build.Tasks/RestoreTask.cs
Nigusu-Allehu: I think it will not be outputted twice. Program.cs is for nuget.exe. And the RestoreNoCache check is for the situation when a user uses msbuild and passes the property RestoreNoCache set to true.

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
```diff
@@ -118,7 +118,7 @@ public async Task<bool> RestoreAsync(string entryProjectFilePath, IDictionary<st
                     dependencyGraphSpec: dependencyGraphSpec,
                     interactive: IsOptionTrue(nameof(RestoreTaskEx.Interactive), options),
                     recursive: IsOptionTrue(nameof(RestoreTaskEx.Recursive), options),
-                    noCache: IsOptionTrue(nameof(RestoreTaskEx.NoCache), options),
+                    noCache: IsOptionTrue(nameof(RestoreTaskEx.NoCache), options) || IsOptionTrue(nameof(RestoreTaskEx.NoHttpCache), options),
```
kartheekp-ms, line 121: I think it would be better for readability if we rename function parameters from `noCache` to `noHttpCache`. For e.g., https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Build.Tasks/BuildTasksUtility.cs#L141 and https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Build.Tasks/BuildTasksUtility.cs#L158

src/NuGet.Clients/NuGet.CommandLine/NuGetCommand.resx
```diff
@@ -192,7 +192,7 @@ nuget push -?</value>
     <value>Disable using the machine cache as the first package source.</value>
   </data>
   <data name="CommandNoHttpCache" xml:space="preserve">
-    <value>Disable using Http cache as the package source.</value>
+    <value>Disable using the Http cache and contact all configured package source sources for live information.</value>
```
jebriede: Removing extra word "source"

```suggestion
    <value>Disable using the Http cache and contact all configured package sources for live information.</value>
```

src/NuGet.Core/NuGet.Build.Tasks/RestoreTask.cs
nkolev92: fwiw, as implemented, I don't think this warning will fail the build even with treat warnings as errors

src/NuGet.Clients/NuGet.CommandLine/NuGetCommand.resx
```diff
@@ -250,6 +253,10 @@ nuget list -verbose -allversions</value>
   <data name="ListCommandVerboseListDescription" xml:space="preserve">
     <value>Displays a detailed list of information for each package.</value>
   </data>
+  <data name="Log_RestoreNoCacheInformation" xml:space="preserve">
+    <value>The NoCache option is only for disabling the HTTP cache and not the global packages folder. It is recommended to use the NoHttpCache option instead to disable the use of the HTTP cache.</value>
```
nkolev92: We're probably gonna duplicate these, but I'm leaving feedback on one.

IMO, we don't need to be this verbose, we can just say `NoCache` is deprecated in favor of the appropriately named `NoHttpCache`, or something along those lines. 

Curious what others on the team think.

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/CommandLineParserTests.cs
```diff
@@ -0,0 +1,42 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Xunit;
+
+namespace NuGet.CommandLine.Test
+{
+    public class CommandLineParserTests
+    {
+
```
nkolev92, line 21: nit: extra line.

src/NuGet.Clients/NuGet.CommandLine/NuGetCommand.resx
zivkan: Alternatively, "NoCache has been renamed to NoHttpCache. The behaviour remains the same".

Unfortunately my suggestion might cause some customers to contact us asking us if this means that NoHttpCache ignores other "caches", and we'll have to inform them no, NoCache was always only the http cache.

If we're concerned about that, we could try "NoCache has been renamed to NoHttpCache, as NoCache was often misunderstood to behave differently."

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
Nigusu-Allehu: Good point Kartheek. I have looked into that; however, it ended up making the PR longer and more confusing. I am thinking of addressing it in a sperate PR.

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/ExtractOptionsTests.cs
Nigusu-Allehu: I have combined them into one line

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/ExtractOptionsTests.cs
Nigusu-Allehu: That's a good question. It will fail gracefully : "Unknow option: -NoHttpCache". I have covered it in a new test case.

src/NuGet.Clients/NuGet.CommandLine/Commands/HelpCommand.cs
```diff
@@ -144,6 +144,7 @@ public void ViewHelpForCommand(string commandName)
 
                 foreach (KeyValuePair<OptionAttribute, PropertyInfo> o in options)
                 {
+                    if (o.Value.Name == "NoCache") { continue; }
```
nkolev92: A more general purpose solution would be to enhance our commandoption class by adding a bool property that indicates to not show this option in the help command, and then just add that check here. Basically something similar to https://github.com/dotnet/sdk/pull/35107/files#diff-15c09ffd510898e37c1ae177803c544ee05ad9bfa2a1b8a47b60284539e3bcd3R151. 

Say add an property bool named `Hidden`.

nkolev92 (commented):
One small suggestion based on a idea from your PR on the dotnet/sdk side, https://github.com/dotnet/sdk/pull/35107.

src/NuGet.Core/NuGet.Build.Tasks/RestoreTask.cs
```diff
@@ -140,11 +145,18 @@ private async Task<bool> ExecuteAsync(Common.ILogger log)
 
             EmbedInBinlog = GetFilesToEmbedInBinlog(dgFile);
 
+            if (RestoreNoCache)
```
nkolev92, line 148: If we log this every time + implement https://github.com/dotnet/sdk/pull/35107, we're gonna end up warning twice in the dotnet.exe case. 

One simple way to fix it would be to use conditional compilation here, but we can also just log twice. I don't see that being a major concern.

src/NuGet.Core/NuGet.Build.Tasks/Strings.resx
```diff
@@ -200,4 +200,8 @@
     <comment>0 - the path to the EXE being launched
 1 - The command-line arguments</comment>
   </data>
-</root>
+  <data name="Log_RestoreNoCacheInformation" xml:space="preserve">
+    <value>NoCache is deprecated in favor of the appropriately named NoHttpCache.</value>
```
nkolev92: This contains `NoCache`, but the property name is `RestoreNoCache`. We should use the full property names.

Nigusu-Allehu: Good catch, thank you!

src/NuGet.Core/NuGet.Build.Tasks/RestoreTask.cs
Nigusu-Allehu: From my understanding, the dotnet sdk calls MSBuild with the property RestoreNoCache, as a result, I decided to not add a log message in the sdk implementation to avoid logging the same message twice. So, I think, we would just have the one warning from MSBuild. I am not sure where the second log message comes from.

nkolev92: mb, I made some wrong assumptions there. You can ignore this.

src/NuGet.Clients/NuGet.CommandLine/Commands/OptionAttribute.cs
```diff
@@ -8,6 +8,7 @@ public sealed class OptionAttribute : Attribute
     {
         private string _description;
 
+        public bool Hidden { get; set; }
```
jebriede: It's helpful to have bool properties start with "Is", "Can", "Has", etc. when it adds value and I believe this is a case where `IsHidden` is more clear than `Hidden`. See [property naming guidelines](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members?redirectedfrom=MSDN#names-of-properties)

```suggestion
        public bool IsHidden { get; set; }
```

src/NuGet.Clients/NuGet.CommandLine/NuGetCommand.resx
```diff
@@ -192,7 +192,7 @@ nuget push -?</value>
     <value>Disable using the machine cache as the first package source.</value>
   </data>
   <data name="CommandNoHttpCache" xml:space="preserve">
-    <value>Disable using the Http cache and contact all configured package source sources for live information.</value>
+    <value>Disable using the Http cache and contact all configured package sources for live information.</value>
```
jebriede: Fixing grammar and fixing casing of HTTP abbreviation.

```suggestion
    <value>Disable the use of the HTTP cache and contact all configured package sources for live information.</value>
```

src/NuGet.Clients/NuGet.CommandLine/Program.cs
```diff
@@ -131,7 +131,9 @@ public static int MainCore(string workingDirectory, string[] args)
                 command.CurrentDirectory = workingDirectory;
                 if (command is DownloadCommandBase downloadCommandBase && downloadCommandBase.NoCache)
                 {
-                    console.WriteWarning(NuGetCommand.Log_RestoreNoCacheInformation);
+                    //Inform users that NoCache option is just for disabling HttpCache and
+                    //suggest them to use NoHttpCache instead, which does the same thing.
```
jebriede: Comments should explain what is not immediately understandable by reading the code. The next line is a Console.LogInformation so it's clear we're outputting a line to the console. We can also see what message is being output but following the reference to NuGetCommand.Log_RestoreNoCacheInformation.

Consider a more concise comment that explains briefly what context is missing:

```suggestion
                    // NoCache option is deprecated. Users should use NoHttpCache instead.
```

src/NuGet.Core/NuGet.Build.Tasks/RestoreTask.cs
```diff
@@ -147,7 +147,9 @@ private async Task<bool> ExecuteAsync(Common.ILogger log)
 
             if (RestoreNoCache)
             {
-                log.LogWarning(Strings.Log_RestoreNoCacheInformation);
+                //Inform users that NoCache option is just for disabling HttpCache and
+                //suggest them to use NoHttpCache instead, which does the same thing.
+                log.LogInformation(Strings.Log_RestoreNoCacheInformation);
```
jebriede, line 152: Can we avoid logging the same thing in [`Program`](https://github.com/NuGet/NuGet.Client/pull/5389/files/4354092afb811f229ff31199df13b7eeeb8375c8..4b5b16770883d57563538b31c1da58dc539f5448#diff-b8b8d924bebf157e6d56ed043f3447d6df86eb521cf75770ea4022908d4903deR136) and here in `RestoreTask`?

src/NuGet.Clients/NuGet.CommandLine/Commands/OptionAttribute.cs
```diff
@@ -33,12 +34,25 @@ private set
...
         }
 
         public OptionAttribute(Type resourceType, string descriptionResourceName)
         {
             ResourceType = resourceType;
             DescriptionResourceName = descriptionResourceName;
+            Hidden = false;
+        }
+        public OptionAttribute(string description, bool hidden) :
+            this(description)
```
jebriede:
```suggestion
        public OptionAttribute(string description, bool hidden)
            : this(description)
```

src/NuGet.Clients/NuGet.CommandLine/Commands/OptionAttribute.cs
```diff
@@ -33,12 +34,25 @@ private set
...
+        {
+            Hidden = hidden;
+        }
+
+        public OptionAttribute(Type resourceType, string descriptionResourceName, bool hidden) :
+            this(resourceType, descriptionResourceName)
```
jebriede:
```suggestion
        public OptionAttribute(Type resourceType, string descriptionResourceName, bool hidden)
            : this(resourceType, descriptionResourceName)
```

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/CommandLineParserTests.cs
```diff
@@ -0,0 +1,73 @@
...
+using NuGet.Commands;
+using System;
+using Moq;
+using System.Reflection;
+
+namespace NuGet.CommandLine.Test
+{
+    public class CommandLineParserTests
+    {
+        private Mock<CommandManager> _mockCommandManager = new Mock<CommandManager>();
```
jebriede: When testing, mock the dependency by interface type, not concrete class:

```suggestion
        private Mock<ICommandManager> _mockCommandManager = new Mock<ICommandManager>();
```

src/NuGet.Clients/NuGet.CommandLine/CommandManager.cs
```diff
@@ -50,7 +50,7 @@ public ICommand GetCommand(string commandName)
             return matchedCommand;
         }
 
-        public IDictionary<OptionAttribute, PropertyInfo> GetCommandOptions(ICommand command)
+        public virtual IDictionary<OptionAttribute, PropertyInfo> GetCommandOptions(ICommand command)
```
jebriede: We don't need to make this virtual if we mock the ICommandManager instead of the concrete CommandManager class in the tests.

src/NuGet.Clients/NuGet.CommandLine/NuGetCommand.resx
```diff
@@ -250,6 +253,10 @@ nuget list -verbose -allversions</value>
   <data name="ListCommandVerboseListDescription" xml:space="preserve">
     <value>Displays a detailed list of information for each package.</value>
   </data>
+  <data name="Log_RestoreNoCacheInformation" xml:space="preserve">
+    <value>NoCache is deprecated in favor of the appropriately named NoHttpCache.</value>
```
jebriede: Can we check on wording of this? I'm not sure "appropriately named NoHttpCache" makes sense without explaining what NoCache does. Consider a few alternatives:
 
```suggestion
    <value>NoCache is deprecated. Please use NoHttpCache instead.</value>
```

```suggestion
    <value>NoCache is deprecated and has been renamed to NoHttpCache. Please use NoHttpCache instead.</value>
```

src/NuGet.Core/NuGet.Build.Tasks/Strings.resx
```diff
@@ -200,4 +200,8 @@
     <comment>0 - the path to the EXE being launched
 1 - The command-line arguments</comment>
   </data>
-</root>
+  <data name="Log_RestoreNoCacheInformation" xml:space="preserve">
+    <value>RestoreNoCache is deprecated in favor of the appropriately named RestoreNoHttpCache.</value>
```
jebriede: Same feedback about wording as in my other comment.

jebriede (commented):
Good update! Thanks for my prior addressing my feedback. I left a few other comments.

src/NuGet.Core/NuGet.Build.Tasks/RestoreTask.cs
nkolev92: RestoreTask is not called in NuGet.exe

jebriede (approved):
Great work! Thanks for addressing the feedback.

nkolev92 (approved):

jeffkl (approved):

PR #4556, opened 2022-04-08
Warn when http sources are used in PackageReference restore operations

nkolev92:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/1358

Regression? Last working version:

## Description

Pretty straightforward. Contains the same warning as https://github.com/NuGet/NuGet.Client/pull/4552, but obviously with a code since it's restore.

## PR Checklist

- [x] PR has a meaningful title
- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [x] Documentation PR or issue filled https://github.com/NuGet/docs.microsoft.com-nuget/issues/2717
  - **OR**
  - [ ] N/A

PR #5390, opened 2023-08-30
Do not warn http sources in package reference restore when allowInsecureConnections is set to true

### LINKED ISSUE #12787: Opt-out of "HTTPS everywhere" : do not warn for HTTP sources in PackageReference restore scenarios in all NuGet tools when allowInsecureConnections  is true
When `allowInsecureConnections ` property in packageSources section is set to true in NuGet.Config files,  suppress warnings for HTTP sources in PackageReference restore scenarios in all NuGet tools.

The PR enable the HTTP warnings in this scenario is: https://github.com/NuGet/NuGet.Client/pull/4556

heng-liu:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12787

## Description
- When `allowInsecureConnections  `property in `packageSources `section is set to true in NuGet.Config files, suppress warnings for HTTP sources in PackageReference restore scenarios in all NuGet tools.
The PR enable the HTTP warnings in this scenario is: https://github.com/NuGet/NuGet.Client/pull/4556

- Updated to use `_request.DependencyProviders.RemoteProviders`, instead of `_request.Project.RestoreMetadata` to check , as the first one respect the settings.
**Context**: The original value of `allowInsecureConnections ` for `RestoreRequest.Project.RestoreMetadata.Sources` is always false, as the value is from: 1.https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs#L200 which doesn't contain any additional attributes like `allowInsecureConnections `. (dotnet restore)
2.https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.CommandLine/MsBuildUtility.cs#L210 which is loading an output.dg file(generated by GenerateRestoreGraphFileTask) which also doesn't contain any additional attributes like `allowInsecureConnections `. (NuGet.exe restore)

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
@@ -211,6 +211,9 @@ namespace NuGet.Commands
             // Standard properties
             restoreArgs.ApplyStandardProperties(request);
 
+            // Apply AllowInsecureConnections attribute
+            restoreArgs.ApplyAllowInsecureConnectionsAttribute(sources, request);
+
             // Add project references
             request.ExternalProjects = projectReferenceClosure.ToList();
 
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
@@ -211,7 +211,7 @@ namespace NuGet.Commands
             // Standard properties
             restoreArgs.ApplyStandardProperties(request);
 
-            // Apply AllowInsecureConnections attribute
+            // Apply AllowInsecureConnections attribute to RestoreRequest.Project.RestoreMetadata.Sources
             restoreArgs.ApplyAllowInsecureConnectionsAttribute(sources, request);
 
             // Add project references
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
@@ -168,6 +168,8 @@ namespace NuGet.Commands
             var globalPath = GetPackagesPath(restoreArgs, projectPackageSpec);
             var settings = _settings ?? Settings.LoadImmutableSettingsGivenConfigPaths(projectPackageSpec.RestoreMetadata.ConfigFilePaths, settingsLoadingContext);
             var sources = restoreArgs.GetEffectiveSources(settings, projectPackageSpec.RestoreMetadata.Sources);
+            //Update the PackageSource to get the attributes from settings(e.g. AllowInsecureConnections). So that the RestoreRequest to be created will have the updated RestoreRequest.Project.RestoreMetadata.Sources.
+            project.PackageSpec.RestoreMetadata.Sources = sources.Select(s => s.PackageSource).ToList();
             var clientPolicyContext = ClientPolicyContext.GetClientPolicy(settings, restoreArgs.Log);
             var packageSourceMapping = PackageSourceMapping.GetPackageSourceMapping(settings);
             var updateLastAccess = SettingsUtility.GetUpdatePackageLastAccessTimeEnabledStatus(settings);
@@ -210,9 +212,6 @@ namespace NuGet.Commands
 
             // Standard properties
             restoreArgs.ApplyStandardProperties(request);
-
-            // Apply AllowInsecureConnections attribute to RestoreRequest.Project.RestoreMetadata.Sources
-            restoreArgs.ApplyAllowInsecureConnectionsAttribute(sources, request);
 
             // Add project references
             request.ExternalProjects = projectReferenceClosure.ToList();
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
@@ -168,8 +168,6 @@ namespace NuGet.Commands
             var globalPath = GetPackagesPath(restoreArgs, projectPackageSpec);
             var settings = _settings ?? Settings.LoadImmutableSettingsGivenConfigPaths(projectPackageSpec.RestoreMetadata.ConfigFilePaths, settingsLoadingContext);
             var sources = restoreArgs.GetEffectiveSources(settings, projectPackageSpec.RestoreMetadata.Sources);
-            //Update the PackageSource to get the attributes from settings(e.g. AllowInsecureConnections). So that the RestoreRequest to be created will have the updated RestoreRequest.Project.RestoreMetadata.Sources.
-            project.PackageSpec.RestoreMetadata.Sources = sources.Select(s => s.PackageSource).ToList();
             var clientPolicyContext = ClientPolicyContext.GetClientPolicy(settings, restoreArgs.Log);
             var packageSourceMapping = PackageSourceMapping.GetPackageSourceMapping(settings);
             var updateLastAccess = SettingsUtility.GetUpdatePackageLastAccessTimeEnabledStatus(settings);
@@ -215,6 +213,9 @@ namespace NuGet.Commands
 
             // Add project references
             request.ExternalProjects = projectReferenceClosure.ToList();
+
+            //Update the RestoreRequest.Project.RestoreMetadata.Sources to get the attributes from settings(e.g. AllowInsecureConnections).
+            request.Project.RestoreMetadata.Sources = sources.Select(s => s.PackageSource).ToList();
 
             // The lock file is loaded later since this is an expensive operation
             var summaryRequest = new RestoreSummaryRequest(
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
@@ -168,6 +168,8 @@ namespace NuGet.Commands
             var globalPath = GetPackagesPath(restoreArgs, projectPackageSpec);
             var settings = _settings ?? Settings.LoadImmutableSettingsGivenConfigPaths(projectPackageSpec.RestoreMetadata.ConfigFilePaths, settingsLoadingContext);
             var sources = restoreArgs.GetEffectiveSources(settings, projectPackageSpec.RestoreMetadata.Sources);
+            //Update the PackageSource to get the attributes from settings(e.g. AllowInsecureConnections). So that the RestoreRequest to be created will have the updated RestoreRequest.Project.RestoreMetadata.Sources.
+            project.PackageSpec.RestoreMetadata.Sources = sources.Select(s => s.PackageSource).ToList();
             var clientPolicyContext = ClientPolicyContext.GetClientPolicy(settings, restoreArgs.Log);
             var packageSourceMapping = PackageSourceMapping.GetPackageSourceMapping(settings);
             var updateLastAccess = SettingsUtility.GetUpdatePackageLastAccessTimeEnabledStatus(settings);
@@ -213,9 +215,6 @@ namespace NuGet.Commands
 
             // Add project references
             request.ExternalProjects = projectReferenceClosure.ToList();
-
-            //Update the RestoreRequest.Project.RestoreMetadata.Sources to get the attributes from settings(e.g. AllowInsecureConnections).
-            request.Project.RestoreMetadata.Sources = sources.Select(s => s.PackageSource).ToList();
 
             // The lock file is loaded later since this is an expensive operation
             var summaryRequest = new RestoreSummaryRequest(
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
@@ -168,8 +168,6 @@ namespace NuGet.Commands
             var globalPath = GetPackagesPath(restoreArgs, projectPackageSpec);
             var settings = _settings ?? Settings.LoadImmutableSettingsGivenConfigPaths(projectPackageSpec.RestoreMetadata.ConfigFilePaths, settingsLoadingContext);
             var sources = restoreArgs.GetEffectiveSources(settings, projectPackageSpec.RestoreMetadata.Sources);
-            //Update the PackageSource to get the attributes from settings(e.g. AllowInsecureConnections). So that the RestoreRequest to be created will have the updated RestoreRequest.Project.RestoreMetadata.Sources.
-            project.PackageSpec.RestoreMetadata.Sources = sources.Select(s => s.PackageSource).ToList();
             var clientPolicyContext = ClientPolicyContext.GetClientPolicy(settings, restoreArgs.Log);
             var packageSourceMapping = PackageSourceMapping.GetPackageSourceMapping(settings);
             var updateLastAccess = SettingsUtility.GetUpdatePackageLastAccessTimeEnabledStatus(settings);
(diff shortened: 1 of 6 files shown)


src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
```diff
@@ -211,6 +211,9 @@ private static ExternalProjectReference GetExternalProject(PackageSpec rootProje
             // Standard properties
             restoreArgs.ApplyStandardProperties(request);
 
+            // Apply AllowInsecureConnections attribute to RestoreRequest.Project.RestoreMetadata.Sources
```
nkolev92: Use Sources from line 170. I think we already handle that deduping there to ensure that the PackageSourceCredentials stays on the PackageSource object.

nkolev92: You probably should try tracking how the sources are passed down to the restorecommand, and that'd probably need to replace the sources used instead of the ones in the request.

test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommandTests.cs
```diff
@@ -4087,10 +4089,22 @@ public async Task Restore_WithHttpSource_Warns()
...
+            string expectedWarning = "You are running the 'restore' operation with an 'HTTP' source, 'http://api.source/index.json'. Non-HTTPS access will be removed in a future version. Consider migrating to an 'HTTPS' source.";
+            string unExpectedWarning = "You are running the 'restore' operation with an 'HTTP' source, 'https://api.source/index.json'. Non-HTTPS access will be removed in a future version. Consider migrating to an 'HTTPS' source.";
+
+            if (isHttpWarningExpected)
+            {
+                result.LockFile.LogMessages.Should().HaveCount(1);
+                IAssetsLogMessage logMessage = result.LockFile.LogMessages[0];
+                logMessage.Code.Should().Be(NuGetLogCode.NU1803);
+                Assert.Contains(expectedWarning, logMessage.Message);
+                Assert.DoesNotContain(unExpectedWarning, logMessage.Message);
```
donnie-msft: I like that this test is thorough, but is it clearer to just assert `Equals` on the expected warning, then drop the unexpected warning completely? 

```suggestion
```

test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommandTests.cs
```diff
@@ -4087,10 +4089,22 @@ public async Task Restore_WithHttpSource_Warns()
             // Assert
             result.Success.Should().BeTrue(because: logger.ShowMessages());
             result.LockFile.Libraries.Should().HaveCount(0);
-            result.LockFile.LogMessages.Should().HaveCount(1);
-            IAssetsLogMessage logMessage = result.LockFile.LogMessages[0];
-            logMessage.Code.Should().Be(NuGetLogCode.NU1803);
-            logMessage.Message.Should().Be("You are running the 'restore' operation with an 'HTTP' source, 'http://api.source/index.json'. Non-HTTPS access will be removed in a future version. Consider migrating to an 'HTTPS' source.");
+
+            string expectedWarning = "You are running the 'restore' operation with an 'HTTP' source, 'http://api.source/index.json'. Non-HTTPS access will be removed in a future version. Consider migrating to an 'HTTPS' source.";
```
donnie-msft: To reduce copying the HTTP source as well as emphasize that your test is expecting the HTTP source URL, I think a variable here would be clearer:

```suggestion
            string expectedWarning = $"You are running the 'restore' operation with an 'HTTP' source, '{testHttpSourceUrl}'. Non-HTTPS access will be removed in a future version. Consider migrating to an 'HTTPS' source.";
```

src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
donnie-msft: > Use Sources from line 170

I think it is using that one? Do you just mean edit the comment, or the code as well?

donnie-msft: This comment doesn't help me, and in fact makes me wonder why `RestoreRequest` is mentioned. Can it either be elaborated as to why `RestoreRequest` is important, or just delete it?

Also, if you do feel a comment is needed, consider adding a comment-block to the method name, that way you can reference the attribute directly, something like `/// Will apply <see cref="sourceRepository.PackageSource.AllowInsecureConnections"> attribute to <see cref="RestoreRequest.Project.RestoreMetadata.Sources"> for sources in settings in the <paramref name="projectDgSpec" />`.
```suggestion
```

nkolev92: I don't think it's using it. 
GetEffectiveSources ensures that the PackageSource is the ones that comes form the settings, one that has the PackageSourceCredentials and the likes attached. This uses the one from the package spec on the request.

test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommandTests.cs
zivkan: If you mean why not do `Assert.Equals("NU1234: This is the expected message")`, then that's high risk of causing problems for people who wish to contribute to NuGet but don't use English as their locale: https://github.com/NuGet/Home/issues/12820. The risk will significantly increase if we eventually do localization more like the rest of github.com/dotnet repos, since local builds can have the satellite assemblies.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
heng-liu: Thanks for reviewing @nkolev92 and @donnie-msft !
I used the `sources` in line 170, to update the `RestoreRequest.Project.RestoreMetadata.Sources`, but only updated the `AllowInsecureConnections` flag of sources, not updating other attributes.
I just changed it into updating the whole `Sources`.

test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommandTests.cs
donnie-msft: > If you mean why not do Assert.Equals("NU1234: This is the expected message")

No, I do not mean that. I expect we would continue using the localized variable.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
heng-liu: Thanks for the question! The `RestoreRequest` is important as we check if the source is http at [here](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs#L242-L246). We need to check additional `source.AllowInsecureConnections`, but the `RestoreRequest.Project.RestoreMetadata.Sources` is not updated to reflect the value in settings. So we have to either updated the `AllowInsecureConnections` flag for `RestoreRequest.Project.RestoreMetadata.Sources`, or updated the whole `RestoreRequest.Project.RestoreMetadata.Sources`.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
```diff
@@ -214,6 +214,9 @@ private static ExternalProjectReference GetExternalProject(PackageSpec rootProje
             // Add project references
             request.ExternalProjects = projectReferenceClosure.ToList();
 
+            //Update the RestoreRequest.Project.RestoreMetadata.Sources to get the attributes from settings(e.g. AllowInsecureConnections).
+            request.Project.RestoreMetadata.Sources = sources.Select(s => s.PackageSource).ToList();
```
nkolev92: This was my bad @heng-liu, I should've done this investigation the first time I provided feedback. 

I think we need a slight change here. 
I *think* my original implementation that I added here: https://github.com/NuGet/NuGet.Client/commit/4d2c5a4ca8e38aa62d3c1d3357076c2a3f701c9c is not ideal. 

Some background: 

- PackageSpec, or rather `Project` is considered to be the `model` for restore, the input. This will be created differently on the commandline (static graph vs standard) vs Visual Studio. It has a sources property, which is the processed sources based on the project path, `RestoreSources` msbuild property etc. It's the actual end sources. The general expectation is that the PackageSpec *does* not get modified. It is basically what is used for no-op.
- RestoreRequest has a list of `DependencyProviders`. 
These dependency providers are "shared" across projects, created within this classl,https://github.com/NuGet/NuGet.Client/blob/41e98d33e40293d8276015d7a01e4d90610fa9d8/src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs#L175-L182. It basically ensures that restore for every project shares the same providers, the same SourceRepository and like resources. These dependency provides also happen to have repositories, and PackageSource that *should* match the ones from the settings file. Hence the whole GetEffectiveSources thing. 

When I added the original implementation, I used `_request.Project.RestoreMetadata.Sources`. I think that's incorrect. Those are *not* how the sources are used, they are used through `request.DependencyProviders` which is effectively *based* on the PackageSource list *from* the package spec. 

I should've used that and it's already used in a bunch of places, like for example, detecting http sources for telemetry, https://github.com/NuGet/NuGet.Client/blob/41e98d33e40293d8276015d7a01e4d90610fa9d8/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs#L155. 

Basically, PackageSpec on the request is expected to be immutable, see: https://github.com/NuGet/Home/issues/9041#issuecomment-1703410411. We basically clone it a bunch of time because of fear of these types of changes. 

Now this is a thing you've never caught yourself, and it's not *incorrect* to do this, but I think it'll be a smaller change if you just change it to use the dependencyproviders. 

You can try that change and see how it goes, tests might break. Like, not sure if everything is mocked 100%. If that becomes the case, please lmk and I can help or just do the change from using the package spec metadata to the dependencyprovider once, since I caused the pain anyways :)

nkolev92: I think the review has been re-requested mistakenly,

heng-liu: Thanks @nkolev92 for the detailed contexts! (I was wondering why the tests for no-op failed previously.)
I changed to use RemoteProviders in a testing branch and the [build ](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=8355913&view=results) shows the newly changed test failed. There is no warning in result.LockFile.LogMessages so it failed at [this line](https://github.com/NuGet/NuGet.Client/blob/dev-hengliu-optoutHttpWarnings-packageReferrenceRestore-test4/test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommandTests.cs#L4098)
However, I can see the warning from the console when running the commandline.
Do you have any suggestions about fixing this test? Thanks!

src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
donnie-msft: This comment and line of code was deleted, so resolving.

test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommandTests.cs
donnie-msft: Looking at the latest commits and this thread, this hasn't been addressed.

test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommandTests.cs
donnie-msft: Looking at the latest commits and this thread, this hasn't been addressed.

donnie-msft (commented):
I re-reviewed as requested by @jeffkl, but after looking at the latest commits and my unresolved threads, there are unaddressed comments by me.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
nkolev92: https://github.com/NuGet/NuGet.Client/blob/29f292950dfa50aea030bd74c3277e983eff711e/test/TestUtilities/Test.Utility/Commands/ProjectTestHelpers.cs#L223 has a bug. It's using a hardcoded list where it's using the `PathContext` source instead of the ones from the package spec. It should be using the PackageSpec ones. I'll do some quick testing to see if I can have a fix.

nkolev92: I'd add a new CreateRestoreRequestMethod, where the source is `spec.RestoreMetadata.Sources`. 

That makes the test pass. Other tests would fail if you make the change in project test helpers directly. 
I know how to fix the ones I wrote, but there's many other tests that have package specs that are incomplete.

nkolev92: Actually if you replace that line with: 
```cs
            var sources = spec.RestoreMetadata.Sources.Any() ?
                spec.RestoreMetadata.Sources :
                new List<PackageSource> { new PackageSource(pathContext.PackageSource) };
```

All tests seem to pass.

nkolev92: If you end up choosing that, please create an issue to fix the tests and assign to me. I think our test helpers code should use *as much* of the production code as possible. I'm hoping to do some changes in the ProjectTestHelpers that'll help us avoid these problems.

nkolev92: I spent about an hour doing some refactoring: https://github.com/NuGet/NuGet.Client/tree/dev-nkolev92-cleanupTestHelpers. Feel free to make the change to fix your tests.

nkolev92: I have https://github.com/NuGet/NuGet.Client/pull/5407 out.

test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommandTests.cs
heng-liu: Thanks for reviewing! Sorry for taking long to address this. This is fixed.

test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommandTests.cs
heng-liu: Thanks! This is fixed.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
heng-liu: Thanks @nkolev92 for your help! Please take another look.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -239,11 +239,12 @@ public async Task<RestoreResult> ExecuteAsync(CancellationToken token)
                     _success = false;
                 }
 
-                if (_request.Project?.RestoreMetadata != null)
+                if (_request.DependencyProviders.RemoteProviders != null)
```
nkolev92, line 242: Excellent!

Thanks for fixing my mistake :)

nkolev92 (approved):
🚢

PR #5394, opened 2023-09-01
Remove NuGetOperationType

### LINKED ISSUE #12866: Remove the NuGetOperationType from NuGet.PackageManagement, use NuGetProjectActionType instead
While working conditional PR updating, I noticed myself doing a lot of duplicated work. See more details for motivation: https://github.com/NuGet/NuGet.Client/compare/dev-nkolev92-conditionalPackageReferenceUpdating?expand=1

For example, I had to do lots of conversions with NuGetProjectActionType and NuGetOperationType.
This is redundant, and makes our code more complicated.

I looked at the references and the only place where the NuGetOperationType is explicitly used is as an API is ActionTelemetryEvent, which is not used a lot.

Consolidating these types simplifies the NuGet code and makes future updates simpler.

nkolev92:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12866

## Description

While working conditional PR updating, I noticed myself doing *a lot* of duplicated work. See more details for motivation: https://github.com/NuGet/NuGet.Client/compare/dev-nkolev92-conditionalPackageReferenceUpdating?expand=1

For example, I had to do lots of conversions with NuGetProjectActionType and NuGetOperationType. 
This is redundant, and makes our code more complicated. 

I looked at the references and the only place where the NuGetOperationType is explicitly used is as an API is ActionTelemetryEvent, which is not used a lot.

Now this is a breaking change to a public API, which is why this refactoring is coming on its own instead as part of the PR that will follow, but the reasoning is that we should reducing duplication and unnecessary conversion. The likelihood of this API breaking anyone is super low. 

**Note** - Note that NuGet.PackageManagement is *not* part of the .NET SDK. It is the broadest package we ship and is primarily used by component that package all of the NuGet functionality themselves, rather than as an extension of NuGet.

If people are ok with this change, and I'm looking at @zivkan primarily, I'll create a bug matching the reasoning in this description, and mark with `Type:BreakingChange`. 

## PR Checklist

- [ ] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added - The tests were changed to simply use the new type.
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.PackageManagement/Telemetry/ActionsTelemetryEvent.cs
+++ b/src/NuGet.Core/NuGet.PackageManagement/Telemetry/ActionsTelemetryEvent.cs
@@ -14,7 +14,7 @@ namespace NuGet.PackageManagement
         public ActionsTelemetryEvent(
             string operationId,
             string[] projectIds,
-            NuGetOperationType operationType,
+            NuGetProjectActionType operationType,
             DateTimeOffset startTime,
             NuGetOperationStatus status,
             int packageCount,
@@ -26,27 +26,6 @@ namespace NuGet.PackageManagement
 
         public const string NugetActionEventName = "NugetAction";
 
-        public NuGetOperationType OperationType => (NuGetOperationType)base[nameof(OperationType)];
-    }
-
-    /// <summary>
-    /// Define nuget operation type values.
-    /// </summary>
-    public enum NuGetOperationType
-    {
-        /// <summary>
-        /// Install package action.
-        /// </summary>
-        Install = 0,
-
-        /// <summary>
-        /// Update package action.
-        /// </summary>
-        Update = 1,
-
-        /// <summary>
-        /// Uninstall package action.
-        /// </summary>
-        Uninstall = 2,
+        public NuGetProjectActionType OperationType => (NuGetProjectActionType)base[nameof(OperationType)];
     }
 }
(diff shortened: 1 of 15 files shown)


src/NuGet.Core/NuGet.PackageManagement/Telemetry/ActionsTelemetryEvent.cs
```diff
@@ -26,27 +26,6 @@ public class ActionsTelemetryEvent : ActionEventBase
...
-        /// <summary>
-        /// Update package action.
-        /// </summary>
-        Update = 1,
-
-        /// <summary>
-        /// Uninstall package action.
-        /// </summary>
-        Uninstall = 2,
+        public NuGetProjectActionType OperationType => (NuGetProjectActionType)base[nameof(OperationType)];
```
dtivel, line 29: If I am understanding correctly, `OperationType` is emitted telemetry.  If so, why not keep `NuGetOperationType` and remove `NuGetProjectActionType`.  If the current change --- removing `NuGetOperationType` --- makes more sense, why not update parameter names (`operationType` -> `actionType`)?  This property can remain as-is for telemetry.

nkolev92: NuGetProjectActionType is used more heavily overall. 80+ references vs 20ish of NuGetOperationType. 

NuGetProjectActionType  is also used more in public types. More notable in the NuGetProjectAction which itself is a type that's used a lot.

dtivel (approved):

zivkan (approved):
I'm not too concerned about NuGet.PackageManagement.

While sometimes similar but different APIs can be useful for strongly typed language type safety, it's not clear to me these two enums provide any API correctness benefits, so consolidating them is fine with me.

PR #5395, opened 2023-09-01
Make MigrationRunner tests more reliable

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2452

## Description

These tests all run against the live Migrations directory but that causes them to be unreliable.  I've added an `internal` overload that accepts a migration directory to test the migration functionality against a temporary location.  This allows the tests to run in parallel and be reliable.

1. Refactored the tests to use `FluentAssertions`.
2. Refactored `GetMigrationsDirectory()` to not create the directory and simplify branching logic.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Common/Migrations/MigrationRunner.cs
+++ b/src/NuGet.Core/NuGet.Common/Migrations/MigrationRunner.cs
@@ -70,20 +70,16 @@ namespace NuGet.Common.Migrations
 
         internal static string GetMigrationsDirectory()
         {
-            string migrationsDirectory;
             if (RuntimeEnvironmentHelper.IsWindows)
             {
-                migrationsDirectory = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "NuGet", "Migrations");
-            }
-            else
-            {
-                var XdgDataHome = Environment.GetEnvironmentVariable("XDG_DATA_HOME");
-                migrationsDirectory = string.IsNullOrEmpty(XdgDataHome)
-                    ? Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".local", "share", "NuGet", "Migrations")
-                    : Path.Combine(XdgDataHome, "NuGet", "Migrations");
+                return Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "NuGet", "Migrations");
             }
 
-            return migrationsDirectory;
+            var XdgDataHome = Environment.GetEnvironmentVariable("XDG_DATA_HOME");
+
+            return string.IsNullOrEmpty(XdgDataHome)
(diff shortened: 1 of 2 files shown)


nkolev92 (approved):

src/NuGet.Core/NuGet.Common/Migrations/MigrationRunner.cs
```diff
@@ -14,6 +14,12 @@ public static class MigrationRunner
         public static void Run()
         {
             string migrationsDirectory = GetMigrationsDirectory();
+
+            Run(migrationsDirectory);
```
zivkan, line 18: This only changes the directory it looks for a file that says whether or not the migrations have already run on the current machine.

However, the linked issue is because the migration itself is trying to delete a directory that had a new file created in it since the recursive delete started. From memory, the only directory that gets deleted by the migration is the `v3-cache` directory on Linux & Mac, which is not this migrations directory. The old `v3-cache` directory is obtained by reading envionrment variables and/or `Environment.GetFolderPath`.

Therefore, I don't believe this PR fixes the linked issue.

jeffkl: The stack trace from the failed execution is:

```
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.Directory.DeleteHelper(String fullPath, String userPath, Boolean recursive, Boolean throwOnTopLevelDirectoryNotFound, WIN32_FIND_DATA& data)
   at System.IO.Directory.Delete(String fullPath, String userPath, Boolean recursive, Boolean checkHost)
   at NuGet.Common.Test.MigrationRunnerTests.Run_WhenExecutedOnSingleThreadThenOneMigrationFileIsCreated_Success() in D:\a\_work\1\s\test\NuGet.Core.Tests\NuGet.Common.Test\MigrationRunnerTests.cs:line 25
```

So I believe the code for each test gets the production migration directory and deletes it which can fail under certain circumstances.

https://github.com/NuGet/NuGet.Client/blob/41e98d33e40293d8276015d7a01e4d90610fa9d8/test/NuGet.Core.Tests/NuGet.Common.Test/MigrationRunnerTests.cs#L20-L22

So the point of this pull request is to have an overload so each test can run against a different directory so they don't stomp on each other.  I don't think there's anything wrong with the actual migration code, its just the test itself.

zivkan: Oops!  When I looked at the stack, my brain removed `Tests` from the name, and I thought it was the migration code throwing.  Yes, this PR should fix the actual call stack reported, not the imaginary call stack I was talking about 😆

zivkan (approved):

aortiz-msft (approved):
🎉🥳

PR #2229, opened 2018-05-10
Fixing the string formatting on Package metadata published date for Japanese

mishra14:
## Bug
Fixes: [57506](https://devdiv.visualstudio.com/DevDiv/NuGet/_workitems/edit/457506)
Regression: No  

## Fix
Details: Fixes the string formatting over the published date in the package metadata page. This PR adds a custom formatting for ja-JP as requested by loicalization vendors.

#### English -
en

#### Turkish-
tr

#### Japanese-
jp

## Testing/Validation
Tests Added: No  
Reason for not adding tests:  UI text changes.
Validation done:  Manual validation in ENG, TRK and JP.

PR #5396, opened 2023-09-01
Publish Date is shown with Regional Settings using CurrentCulture

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2283

## Description

1. Regional formatting is now honored. For example, the [Developer Community ticket](https://developercommunity.visualstudio.com/t/Date-format-in-nuget-options-is-in-US-fo/10353388) states that the United Kingdom date format wasn't being honored, which it now will be (see screenshots).
    1. The existing code relies on the `CurrentUICulture`, and unfortunately, if that's not explicitly set by something (every XAML page or globally, I believe), then the default is `en-us`. That's why existing behavior was culture agnostic.
  
1. Removes `DateTimeConverter` and tests. Apparently, this was created just to hard code Japanese to force a day of week upon that culture. I haven't found any evidence that this is appropriate. Just because it's expected in one culture does not mean it's expected in another culture.

1. I tested various combinations of Windows Language/Country/Regional settings, as well as Visual Studio International Settings. Results are below.

Notice that Language is not important, whereas Regional settings always determine the rendered formatting.

### Language: English with Country/Regional format: United Kingdom (English)

`Date Published` is in English/UK date format (matches regional settings)
image

### Language: Japanese with Country/Regional format: United States (English)
`Date Published` is in English/US date format (matches regional settings)
image

### Language: Japanese with Country/Regional format: Japan (Japanese)
`Date Published` is in Japanese/Japan date format (matches regional settings)
image

### Windows Language: English; Regional: Japanese/Japan; VS Language: Japanese
`Date Published` is in Japanese/Japan date format (matches regional settings)
image

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception - Removed custom logic and tests. Tested using Windows/VS international settings.
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/DateTimeConverter.cs
zivkan, line 1: Looking at this file's commit history in the dev branch, and finding the first commit's PR:
* https://github.com/NuGet/NuGet.Client/pull/2229

It has a link to an internal bug: https://devdiv.visualstudio.com/DevDiv/NuGet/_workitems/edit/457506

Therefore, deleting this file appears to regress the bug that was previously fixed.

donnie-msft: That's correct, and I also used the same technique to find the exact same bug. As mentioned in this PR description, I feel it's wrong to tell a culture that it must match another culture. Can you find evidence that we should use special behavior for Japanese? I couldn't find any other code doing this in VS, but it is hard to search.

zivkan: Your PR description says "I haven't found any evidence that this is appropriate". The bug filed on us in 2018 is evidence. If what you meant was "I saw the bug from 2018, but I don't think it is correct", then your PR description wasn't sufficiently clear. What you wrote is ambiguous and my interpretation of your PR description is not incompatible with you being unaware of the bug.

Your PR description also says "Just because it's expected in one culture does not mean it's expected in another culture.", but to be honest, I'm not really sure what this is trying to say. If you mean that some parts of Japan show the weekday, but other don't, that's something I not considered until writing this comment. Since the .NET data type for specifying locale is `CultureInfo`, I interpreted your comment as meaning something like "just because it's expected in Japanese, does not mean it's expected in English", which still doesn't make a lot of sense because the deleted code only shows weekday for `jp-JP`. But again, you used the word culture in the context of a class that checks the value of a `CultureInfo` instance.

Additionally, a quick web search came up with Wikipedia's article: https://en.wikipedia.org/wiki/Date_and_time_notation_in_Japan

> The most commonly used [date format](https://en.wikipedia.org/wiki/Date_format) in [Japan](https://en.wikipedia.org/wiki/Japan) is "year month day (weekday)",

So, again, I think this is more evidence that perhaps it is appropriate for the weekday to be listed when showing dates in Japanese.

Assuming the person who filed the 2018 bug was a localization vendor, I'd feel more comfortable verifying with the localization team. I'm just trying to do some due diligence in avoiding a regression on a bug that we were assigned in the past. In my opinion, your PR description did not provide sufficient evidence to justify removing something that a localization expert asked us to implement in the first place.

jeffkl (approved):

src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/DateTimeConverter.cs
donnie-msft: You're asking for positive confirmation from the loc team that we can use the default culture behavior, and I think that's totally valid.

Since the bug's justification was that "I agree that it's missing for JPN". Well, that's not justification, that's an observation.
Everything I understand about culture settings is that they are specifically to be adapted to each region. Just because US cares about day of week in their long date format, does not mean JPN readers would expect that. In fact, there's other cultures which do not, and so I feel it's very weird to have hard-coded this in our repo.

From NuGet's perspective, I see no business reason to force day of week into this field. In fact, I think it's probably largely ignored most of the time. Unless I'm carefully monitoring and only want packages published on "Tuesdays" because that's the release date I trust more than "Fridays", I can't think of why it's necessary.

I'll let you know once I hear back from the loc team on their view.

donnie-msft:
I'm moving back to Draft until I hear back on a question for the loc team:
https://github.com/NuGet/NuGet.Client/pull/5396#discussion_r1317611791

donnie-msft:
I've asked world ready experts to confirm either via PR review or offline discussion. I'll post here once I know more.

donnie-msft:
I have filed a question with the Loc Team. The first responder agrees with my assessment, but they are checking with peers to validate...

donnie-msft:
@zivkan I finally have an answer for you. Today, I received confirmation from someone with authority in our Loc Team. They checked with peers and confirmed that my assessment is correct, and we should not be hard-coding Japanese's date format for any reason. Find the [link to that response](https://github.com/NuGet/Client.Engineering/issues/2283#issuecomment-1725790179) in the issue.

zivkan (approved):

PR #5397, opened 2023-09-03
Correctly identify transitive package references

### LINKED ISSUE #10368: When a ProjectReference and PackageReference in different TFMs have the same identity/version, Restore throws "An item with the same key has already been added. [...]"
## Details about Problem

Full repro details at https://github.com/dagood/repro/tree/repro-restore-ref-same-key. I used the `mcr.microsoft.com/dotnet/sdk:5.0` Docker image to minimally repro, but this also repros the same on my Windows machine.

```
$ dotnet --info
.NET SDK (reflecting any global.json):
 Version:   5.0.100
 Commit:    5044b93829

Runtime Environment:
 OS Name:     debian
 OS Version:  10
 OS Platform: Linux
 RID:         debian.10-x64
 Base Path:   /usr/share/dotnet/sdk/5.0.100/

Host (useful for support):
  Version: 5.0.0
  Commit:  cf258a14b7

.NET SDKs installed:
  5.0.100 [/usr/share/dotnet/sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 5.0.0 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 5.0.0 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
```

System.Security.Cryptography.Cng.csproj

```xml
  <PropertyGroup>
    <TargetFramework>net5.0</TargetFramework>
    <Version>5.0.0-rc.2.20475.5</Version>
  </PropertyGroup>
</Project>
```

System.Security.Cryptography.Pkcs.csproj

```xml
  <PropertyGroup>
    <TargetFrameworks>net5.0;netcoreapp3.1</TargetFrameworks>
  </PropertyGroup>
  <ItemGroup Condition="'$(TargetFramework)' == 'net5.0'">
    <ProjectReference Include="../System.Security.Cryptography.Cng/System.Security.Cryptography.Cng.csproj" />
  </ItemGroup>
  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
    <PackageReference Include="System.Security.Cryptography.Cng" Version="5.0.0-rc.2.20475.5" />
  </ItemGroup>
</Project>
```

As far as I know I haven't seen this work before.

We hit it in source-build while trying to build dotnet/runtime: https://github.com/dotnet/source-build/pull/1845#discussion_r522994234

This might not be supported--it doesn't make a *whole* lot of sense to me for a project to intentionally do this--but the error message should be improved at least! 😄 It took a while to figure out what the message was complaining about, work around it in source-build, and write this small repro.

## Detailed repro steps so we can see the same problem

1. Clone https://github.com/dagood/repro/tree/repro-restore-ref-same-key
   (`git clone -b repro-restore-ref-same-key https://github.com/dagood/repro && cd repro`)

2. Run `dotnet build ./System.Security.Cryptography.Pkcs/System.Security.Cryptography.Pkcs.csproj /bl`

3. See this error:
    `/usr/share/dotnet/sdk/5.0.100/NuGet.targets(131,5): error : An item with the same key has already been added. Key: (System.Security.Cryptography.Cng, 5.0.0-rc.2.20475.5) [/work/System.Security.Cryptograp
hy.Pkcs/System.Security.Cryptography.Pkcs.csproj]`

4. Open `msbuild.binlog` to see full stack trace:

```
System.ArgumentException: An item with the same key has already been added. Key: (System.Security.Cryptography.Cng, 5.0.0-rc.2.20475.5)
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey](List`1 source, Func`2 keySelector, IEqualityComparer`1 comparer)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey](IEnumerable`1 source, Func`2 keySelector, IEqualityComparer`1 comparer)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey](IEnumerable`1 source, Func`2 keySelector)
   at NuGet.Commands.LockFileBuilder.CreateLockFile(LockFile previousLockFile, PackageSpec project, IEnumerable`1 targetGraphs, IReadOnlyList`1 localRepositories, RemoteWalkContext context)
   at NuGet.Commands.RestoreCommand.ExecuteAsync(CancellationToken token)
   at NuGet.Commands.RestoreRunner.ExecuteAsync(RestoreSummaryRequest summaryRequest, CancellationToken token)
   at NuGet.Commands.RestoreRunner.ExecuteAndCommitAsync(RestoreSummaryRequest summaryRequest, CancellationToken token)
   at NuGet.Commands.RestoreRunner.CompleteTaskAsync(List`1 restoreTasks)
   at NuGet.Commands.RestoreRunner.RunAsync(IEnumerable`1 restoreRequests, RestoreArgs restoreContext, CancellationToken token)
   at NuGet.Commands.RestoreRunner.RunAsync(RestoreArgs restoreContext, CancellationToken token)
   at NuGet.Build.Tasks.BuildTasksUtility.RestoreAsync(DependencyGraphSpec dependencyGraphSpec, Boolean interactive, Boolean recursive, Boolean noCache, Boolean ignoreFailedSources, Boolean disableParallel, Boolean force, Boolean forceEvaluate, Boolean hideWarningsAndErrors, Boolean restorePC, Boolean cleanupAssetsForUnsupportedProjects, ILogger log, CancellationToken cancellationToken)
   at NuGet.Build.Tasks.RestoreTask.ExecuteAsync(ILogger log)
```

### Verbose Logs

Binlog: [repro-restore-ref-same-key.zip](https://github.com/NuGet/Home/files/5661841/repro-restore-ref-same-key.zip)

ViktorHofer:
## Bug

Fixes https://github.com/NuGet/Home/issues/10368

Regression? Last working version: Never worked before

## Description

When a transitive package dependency has the same identity as a direct project dependency that is differenced by a different inner build (TFM), the dependency is incorrectly resolved as a project.

The resolver logic invoked by the dependency walker only knows about which projects are referenced but not by which TFM. To correctly isolate inner builds from each other and avoid incorrectly promoting a project dependency, pass the nearest project dependencies to the dependency resolving logic.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A


src/NuGet.Core/NuGet.DependencyResolver.Core/ResolverUtility.cs
```diff
@@ -24,14 +24,15 @@ public static class ResolverUtility
             NuGetFramework framework,
             string runtimeIdentifier,
             RemoteWalkContext context,
-            CancellationToken cancellationToken)
+            CancellationToken cancellationToken,
+            LibraryDependency[] projectDependencies = null)
```
zivkan: I'm really not a fan of optional types in public APIs. I prefer a new overload, which would also prevent a breaking ABI change. A new overload would also keep the convention of keeping the CancellationToken as the last parameter.

zivkan (commented):
I'm still trying to wrap my head around the original issue and the fix, to ensure I can't foresee unintended consequences from side effects. But if I understand correctly, say I have Contoso.Utilities as a transitive package, if I clone the repo and add Contoso.Utilities.csproj to my solution, then the build will use that project's dll? I can make changes to Contoso.Utilities and test with my app, without neeing to publish packages or changing dependency versions?

If so, that's very cool, and brings us closer to what we had with project.json back in the dnx/.NET Core 1.0 preview days.

ViktorHofer:
> I'm still trying to wrap my head around the original issue and the fix, to ensure I can't foresee unintended consequences from side effects. But if I understand correctly, say I have Contoso.Utilities as a transitive package, if I clone the repo and add Contoso.Utilities.csproj to my solution, then the build will use that project's dll? I can make changes to Contoso.Utilities and test with my app, without neeing to publish packages or changing dependency versions?

Here's my example project as a zip: [tfmnugetoverlap.zip](https://github.com/NuGet/NuGet.Client/files/12539570/tfmnugetoverlap.zip). Do a `dotnet restore tfmnugetoverlap\main\main.csproj` to repro the restore failure.

> I'm really not a fan of optional types in public APIs. I prefer a new overload, which would also prevent a breaking ABI change. A new overload would also keep the convention of keeping the CancellationToken as the last parameter.

Agreed. Since I opened the PR I also explored another solution that would do the following instead:

The general problem is that the `PackageSpecReferenceDependencyProvider` which gets added to the restore `context` isn't aware of target frameworks / inner builds: https://github.com/NuGet/NuGet.Client/blob/48fe41a4d8aa997095fde969ff1608789602a046/src/NuGet.Core/NuGet.ProjectModel/PackageSpecReferenceDependencyProvider.cs#L22-L26

Because of that, [the individual TFM restore operations](https://github.com/NuGet/NuGet.Client/blob/48fe41a4d8aa997095fde969ff1608789602a046/src/NuGet.Core/NuGet.Commands/RestoreCommand/ProjectRestoreCommand.cs#L57-L67) consider all project dependencies, even the ones that conditionally are excluded for the current framework. I experimented with updating `PackageSpecReferenceDependencyProvider` to encode which projects apply to which framework. The challenging part is calculating the TFM aware dependency tree though. NuGet / msbuild creates a dgspec.json file which lists all the dependencies but the TFM selection doesn't happen until NuGet actually resolves the dependency. So building that dependency tree upfront might cause unintended side-effects which I'm not too excited to make given that this code base is unfamiliar to me.

So all in all, I guess the current approach (passing the actual dependencies in to the resolver logic vs updating the `PackageSpecReferenceDependencyProvider` to be TFM aware and only return the matching dependencies) is fine.

src/NuGet.Core/NuGet.DependencyResolver.Core/ResolverUtility.cs
ViktorHofer: addressed

nkolev92:
Thanks for the contribution @ViktorHofer! I know this must have been a lot of work.

> The general problem is that the PackageSpecReferenceDependencyProvider which gets added to the restore context isn't aware of target frameworks / inner builds:

Can you elaborate on this?
It has methods that take a framework, so it can do the proper compat selection. 
https://github.com/NuGet/NuGet.Client/blob/40648c3cb3de37cf62b194accdc8e7117b36245f/src/NuGet.Core/NuGet.ProjectModel/PackageSpecReferenceDependencyProvider.cs#L80

For example, I recently did work there: https://github.com/NuGet/NuGet.Client/commit/df7d9954a8ac7bc1891f18d6e4936866e2d20250. 

The packagespecreferenceprovider should be called in the context of calling a dependency provider.
When the package spec reference provider is asked for certain dependencies it is asked with a framework: https://github.com/NuGet/NuGet.Client/blob/40648c3cb3de37cf62b194accdc8e7117b36245f/src/NuGet.Core/NuGet.DependencyResolver.Core/ResolverUtility.cs#L161C38-L161C59. 

You are probably onto something, but with the reasoning given here, I can't confirm it's the right thing yet.

> When a transitive package dependency has the same identity as a direct project dependency that is differenced by a different inner build (TFM), the dependency is incorrectly resolved as a project.

What does same identity mean for you here? Same ID? Same ID/Version?

nkolev92:
fwiw, I started writing a test so I can debug the behavior.
Here's the code.

```cs
[Fact] public async Task ExecuteAsync_WithConditionalProjectAndPackageReferences() {     // Arrange     using var context = new SourceCacheContext();     using var pathContext = new SimpleTestPathContext();     // Add nuget.org for the extra package.     pathContext.Settings.AddSource("nuget.org", "https://api.nuget.org/v3/index.json");      var mainProject = "main";     var mainProjectPath = Path.Combine(pathContext.SolutionRoot, mainProject);      var childProject = "System.Numerics.Vectors";     var childProjectPath = Path.Combine(pathContext.SolutionRoot, childProject);      var mainProjectJson = @"     {       ""version"": ""1.0.0"",       ""frameworks"": {         ""netstandard2.0"": {             ""imports"": [                         ""net461"",                         ""net462"",                         ""net47"",                         ""net471"",                         ""net472"",                         ""net48"",                         ""net481""                       ],             ""assetTargetFallback"": true,             ""dependencies"": {                 ""System.Memory"": {                               ""target"": ""Package"",                               ""version"": ""[4.5.5, )""                             },                   ""NETStandard.Library"": {                               ""suppressParent"": ""All"",                               ""target"": ""Package"",                               ""version"": ""[2.0.3, )"",                               ""autoReferenced"": true                             }             }         },         ""net8.0"": {             ""imports"": [                         ""net461"",                         ""net462"",                         ""net47"",                         ""net471"",                         ""net472"",                         ""net48"",                         ""net481""                       ],             ""assetTargetFallback"": true,             ""dependencies"": {             }         }       }     }";     PackageSpec systemNumericsVectorPackageSpec = ProjectTestHelpers.GetPackageSpec(childProject, pathContext.SolutionRoot, "net8.0",         useAssetTargetFallback: true, "net461\",\"net462\",\"net47\",\"net471\",\"net472\",\"net48\",\"net481");     PackageSpec mainPackageSpec = ProjectTestHelpers.GetPackageSpecWithProjectNameAndSpec(mainProject, pathContext.SolutionRoot, mainProjectJson);     var settings = Settings.LoadDefaultSettings(pathContext.SolutionRoot);     mainPackageSpec.RestoreMetadata.ConfigFilePaths = settings.GetConfigFilePaths();     mainPackageSpec.RestoreMetadata.Sources = SettingsUtility.GetEnabledSources(settings).ToList();     mainPackageSpec.RestoreMetadata.FallbackFolders = SettingsUtility.GetFallbackPackageFolders(settings).ToList();     mainPackageSpec.RestoreMetadata.PackagesPath = SettingsUtility.GetGlobalPackagesFolder(settings);      mainPackageSpec.RestoreMetadata.TargetFrameworks.Single(e => e.FrameworkName.Equals(NuGetFramework.Parse("net8.0")))         .ProjectReferences.Add(new ProjectRestoreReference()     {         ProjectUniqueName = systemNumericsVectorPackageSpec.RestoreMetadata.ProjectUniqueName,         ProjectPath = systemNumericsVectorPackageSpec.RestoreMetadata.ProjectPath,     });      var logger = new TestLogger();     var request = ProjectTestHelpers.CreateRestoreRequest(pathContext, logger, mainPackageSpec, systemNumericsVectorPackageSpec);      var restoreCommand = new RestoreCommand(request);     RestoreResult result = await restoreCommand.ExecuteAsync();     await result.CommitAsync(logger, CancellationToken.None);      result.Success.Should().BeTrue(because: logger.ShowMessages()); }
```

I used test\NuGet.Core.Tests\NuGet.Commands.Test\RestoreCommandTests\RestoreCommandTests.cs as the class

nkolev92:
I took a stab fixing the PackageSpecReferenceDependencyProvider. It basically gets asked for a library, but it ignores the version. If the version *matches*, then I think the project should be preferred over the package. 
See equivalent for packages: https://github.com/NuGet/NuGet.Client/blob/83912bcf8b650091655a9c0ea726c0adecc553fe/src/NuGet.Core/NuGet.Commands/RestoreCommand/SourceRepositoryDependencyProvider.cs#L289-L299

I'm not really sure if direct refs are the ones that should be checked when including a project in the graph. I think if you have an extra layer of dependencies, then you'd likely have issues, like for example, if System.Numerics vector was a transitive project reference.

My changes are in https://github.com/NuGet/NuGet.Client/tree/dev-nkolev92-10368, here's what happens when I run `dotnet list package`

```cs
C:\Code\Temp\tfmnugetoverlap [main]> dotnet list .\main\main.csproj package --include-transitive
Project 'main' has the following package references
   [netstandard2.0]:
   Top-level Package             Requested   Resolved
   > NETStandard.Library   (A)   [2.0.3, )   2.0.3
   > System.Memory               4.5.5       4.5.5

   Transitive Package                            Resolved
   > Microsoft.NETCore.Platforms                 1.1.0
   > System.Buffers                              4.5.1
   > System.Numerics.Vectors                     4.4.0
   > System.Runtime.CompilerServices.Unsafe      4.5.3

   [net8.0]:
   Transitive Package             Resolved
   > System.Numerics.Vectors      1.0.0

(A) : Auto-referenced package.
```

Now the trick is that this is a *very tricky* change. I'm not really sure it's the right thing just yet.

I'll let the CI run all the tests and revisit after at again when I have more time.

ViktorHofer:
I'm on holiday now and will respond when I'm back. Thanks for taking a look so far.

ViktorHofer:
Met with @nkolev92 offline and talked about different solutions to this problem. I like @nkolev92's proposal which avoids adding new public API and meets our requirements. We identified that there's still a problem in regards to choosing P2P over PackageReferences but we believe this is a super corner case that people shouldn't hit today.

Closing this PR as he will open a new one with the test and the fix.

PR #5398, opened 2023-09-04
Stop warning NU1905 when NuGetAudit doesn't have vulnerability data to check

### LINKED ISSUE #12875: NuGetAudit should not warn when no vulnerability data is available
### NuGet Product(s) Affected

MSBuild.exe, dotnet.exe

### Current Behavior

With the initial previews of NuGetAudit, when none of the package sources have vulnerability data, and the project explicitly enables NuGetAudit (as opposed to enabled by default), then NuGet raises the NU1905 warning.

### Desired Behavior

Don't warn.  It's confusing that audit runs, and warns about packages whether the feature is explicitly enabled, or left for default, but the "no vulnerability" warning is 

### Additional Context

Earlier draft of the PR https://github.com/NuGet/NuGet.Client/pull/5398 proposed a separate property to enable/disable the warning, but there was no support, and multiple oppositions.

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12875

Regression? No

## Description

Don't warn when no package sources contain vulnerability information.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
@@ -409,14 +409,12 @@ namespace NuGet.PackageManagement.VisualStudio
             string enableAudit = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAudit);
             string auditLevel = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditLevel);
             string auditMode = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditMode);
-            string auditRequired = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditRequired);
             RestoreAuditProperties auditProperties = !string.IsNullOrEmpty(enableAudit) || !string.IsNullOrEmpty(auditLevel)
                 ? new RestoreAuditProperties()
                 {
                     EnableAudit = enableAudit,
                     AuditLevel = auditLevel,
                     AuditMode = auditMode,
-                    AuditRequired = auditRequired,
                 }
                 : null;
 
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
@@ -409,12 +409,14 @@ namespace NuGet.PackageManagement.VisualStudio
             string enableAudit = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAudit);
             string auditLevel = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditLevel);
             string auditMode = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditMode);
+            string auditRequired = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditRequired);
             RestoreAuditProperties auditProperties = !string.IsNullOrEmpty(enableAudit) || !string.IsNullOrEmpty(auditLevel)
                 ? new RestoreAuditProperties()
                 {
                     EnableAudit = enableAudit,
                     AuditLevel = auditLevel,
                     AuditMode = auditMode,
+                    AuditRequired = auditRequired,
                 }
                 : null;
 
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
@@ -30,6 +30,7 @@ namespace NuGet.Commands.Restore.Utility
 
         internal PackageVulnerabilitySeverity MinSeverity { get; }
         internal NuGetAuditMode AuditMode { get; }
+        internal bool AuditRequired { get; }
         internal List<string>? DirectPackagesWithAdvisory { get; private set; }
         internal List<string>? TransitivePackagesWithAdvisory { get; private set; }
         internal int Sev0DirectMatches { get; private set; }
@@ -64,6 +65,7 @@ namespace NuGet.Commands.Restore.Utility
 
             MinSeverity = ParseAuditLevel();
             AuditMode = ParseAuditMode();
+            AuditRequired = ParseRequiredValue(_restoreAuditProperties?.AuditRequired, _projectFullPath, _logger);
         }
 
         public async Task CheckPackageVulnerabilitiesAsync(CancellationToken cancellationToken)
@@ -80,7 +82,7 @@ namespace NuGet.Commands.Restore.Utility
 
             if (allVulnerabilityData is null || !AnyVulnerabilityDataFound(allVulnerabilityData.KnownVulnerabilities))
             {
-                if (_auditEnabled == EnabledValue.ExplicitOptIn)
+                if (AuditRequired)
                 {
                     RestoreLogMessage restoreLogMessage = RestoreLogMessage.CreateWarning(NuGetLogCode.NU1905, Strings.Warning_NoVulnerabilityData);
                     restoreLogMessage.ProjectPath = _projectFullPath;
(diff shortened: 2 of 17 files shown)


src/NuGet.Core/NuGet.ProjectModel/RestoreAuditProperties.cs
```diff
@@ -27,13 +27,19 @@ public class RestoreAuditProperties : IEquatable<RestoreAuditProperties>
         /// <value>direct, all</value>
         public string? AuditMode { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether NuGetAudit should warn when no vulnerability data is available.
+        /// </summary>
+        public string? AuditRequired { get; set; }
```
zivkan: I'm not very happy with this name given the behaviour. Setting it to `true`, when the name is `Required`, makes it feel like it should raise an error, not a warning. However, I also don't like the name `WarnWhenNoNuGetAuditData` or `WarnNuGetAuditMissingData`, and I haven't been able to think of any better name.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -80,7 +82,7 @@ public async Task CheckPackageVulnerabilitiesAsync(CancellationToken cancellatio
 
             if (allVulnerabilityData is null || !AnyVulnerabilityDataFound(allVulnerabilityData.KnownVulnerabilities))
             {
-                if (_auditEnabled == EnabledValue.ExplicitOptIn)
+                if (AuditRequired)
                 {
                     RestoreLogMessage restoreLogMessage = RestoreLogMessage.CreateWarning(NuGetLogCode.NU1905, Strings.Warning_NoVulnerabilityData);
```
zivkan: I'm wondering if NU1905 should be raised not only when no package source contains vulnerability data, but also if `NuGetAudit` if `false`.

If the point of the warning is warn when Audit is not run, then it feels like yes it should warn.

On the other hand, it's hard to imagine scenarios where `NuGetAuditRequired` is set to true, but `NuGetAudit` is accidentally set to false. Whether customers use VS project properties window, or hand edit the project or Directory.Build.props file, it seems trivial to set both properties together, or even set `<NuGetAuditRequired>$(NuGetAudit)</NuGetAuditRequired>` in case they want to enable audit for some projects, but not others, while also enforcing that audit runs on the projects where it's enabled.

I think that silently ignoring `NuGetAuditRequired` when `NuGetAudit` is false is more similar to existing features that are configured by MSBuild. So, I think that's how I'm leaning, it's also fewer code changes, but I'm not certain it's the best customer experience.

jeffkl: I think if `NuGetAudit` is disabled, then I should receive no warnings and the whole feature should be disabled.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -440,6 +442,20 @@ public static EnabledValue ParseEnableValue(string? value, string projectFullPat
             return EnabledValue.Invalid;
         }
 
+        public static bool ParseRequiredValue(string? value, string projectFullPath, ILogger logger)
```
jeffkl: I would name this method `ParseNuGetAuditRequired()`, at first I thought it was a generic method to parse a property that itself is required.  But then I read the body and realized its for parsing the value of the `NuGetAuditRequired` property.

src/NuGet.Core/NuGet.ProjectModel/RestoreAuditProperties.cs
jeffkl: Maybe `IsAuditRequired`?  At least the XML doc comment explains what it does.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
```diff
@@ -906,6 +906,7 @@ public static RestoreAuditProperties GetRestoreAuditProperties(IMSBuildItem spec
             string enableAudit = specItem.GetProperty("NuGetAudit");
             string auditLevel = specItem.GetProperty("NuGetAuditLevel");
             string auditMode = specItem.GetProperty("NuGetAuditMode");
+            string auditRequired = specItem.GetProperty("NuGetAuditRequired");
```
nkolev92: This probably won't work with NuGet.targets

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
```diff
@@ -409,12 +409,14 @@ private async Task<PackageSpec> GetPackageSpecAsync(ISettings settings)
             string enableAudit = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAudit);
```
nkolev92: What would our docs say? Use `NuGetAudit` or use `NuGetAuditRequired`? 

I'm not a huge fan of having the extra property to just enable audit. 
Since we're not really shipping audit by default very soon, IMO, it's ok if NuGetAudit is the only configuration knob.

Do we expect that people will ever need to "enable" NuGet Audit themselves? Since it's enabled by default.

nkolev92 (commented):
This is a tough one. I get the problem you're trying to solve, but I'm not sure the solution is worth it.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
zivkan: > I'm not a huge fan of having the extra property to just enable audit.

There are customers who use alternative vulnerability scanners, so will probably want to disable NuGet's built-in scanning. For example within Microsoft, we already have Component Governance, who has more vulnerability sources than just GitHub's Advisory Database.

So, I see `<NuGetAudit>false` as a better alternative than `<NoWarn>NU1901;NU1902;NU1903;NU1904;$(NuWarn)`.

> Since we're not really shipping audit by default very soon

yes we are, in both .NET 8 SDK and VS17.8 it's on by default.

> Do we expect that people will ever need to "enable" NuGet Audit themselves? Since it's enabled by default.

Ok, now I don't understand your previous "we're not really shipping audit by default very soon" statement

> What would our docs say? Use NuGetAudit or use NuGetAuditRequired?

If you don't want to use NuGet to scan restored packages for known vulnerabilities, set the `NuGetAudit` property to `false`.

If you want a warning that NuGet Audit was unable to run, due to no package sources providing a list of packages with known vulnerabilities, then set the `NuGetAuditRequired` property to `true`. If you want restore to fail, also add `NU1905` to `WarningsAsErrors`.

nkolev92: > yes we are, in both .NET 8 SDK and VS17.8 it's on by default.

That was a typo on my end. 

> If you want a warning that NuGet Audit was unable to run, due to no package sources providing a list of packages with known vulnerabilities, then set the NuGetAuditRequired property to true. If you want restore to fail, also add NU1905 to WarningsAsErrors.

How many people would want to do that? Have the failure? I understand the "miss" in the scenario, wanting audit explicitly, but not getting audit cause sources were not configured, but is that scenario important enough that we want to introduce what I feel like an imperfect solution in V1?

zivkan: well, then I think we should remove NU1905 from the current code when audit is explicitly opt-in.

While some people, like all of Microsoft internal builds, have alternate scanning, I thought it would be useful for customers who don't, and want to use NuGet Audit as part of their security process, would want a way to ensure that it runs, so they don't assume that no warning == scanned and no vulnerability found.

aortiz-msft: FWIW I think the extra complexity of having two properties isn't worth it.

src/NuGet.Core/NuGet.Common/Errors/NuGetLogCode.cs
```diff
@@ -369,6 +369,7 @@ public enum NuGetLogCode
         /// <summary>
         /// NuGetAudit enabled, but no vulnerability data found
         /// </summary>
+        /// <remarks>Was added in 6.7 client SDK, but then removed from use in 6.8</remarks>
         NU1905 = 1905,
```
zivkan: I'm undecided if I should delete this code, or leave it in. I still think the feature to warn when Audit doesn't run is useful, which customers can `WarningsAsErrors` in order to break their build if CI was unable to scan for vulnerable packages, in the cases where customers want to do that.

jeffkl (dismissed):

src/NuGet.Core/NuGet.Common/Errors/NuGetLogCode.cs
nkolev92: I don't mind if it gets deleted. You can reintroduce the code if need at a later point. 

Your call though.

nkolev92 (dismissed):

nkolev92 (approved):

PR #5403, opened 2023-09-06
OptProf test project should target latest SDK TFM

zivkan:
## Bug

Fixes: CI

Regression? kind of, but not in NuGet. VS internal previews changed something that broke us.

## Description

Changed the test project that OptProf training test use to use the most appropriate TargetFramework for the version of the .NET SDK currently installed.

The .NET SDK can build and run any version of .NET that is equal to or less than the SDK version. For example, .NET 7 SDK can target net7.0 and below, and .NET 8 SDK can target net8.0 and below. However, the SDK only has reference assemblies and other runtime files for the SDK's version. When a project is targeting a framework that is lower than the version of the SDK, the SDK will use `<PackageDownload` to download the required files. 

Sometimes these packages get new versions published, and so a new version of the .NET SDK will want to use the new package version. However, everything releases publicly on the same day, but the new version of the .NET SDK that uses the new packages needs to be inserted into VS days early, in order to build and test VS that will use the new package version.

That's where we're at today. Our test targets net7.0, and the .NET SDK in VS 17.8 is .NET 8, and there's going to be new versions of the .NET 7 runtime packages published when everything goes GA. Therefore, our test that uses nuget.org as the only package source fails.

We have at least one alternative: Change `build\test.optprof.vsconfig` to install the .NET 7 SDK and runtime. But eventually .NET 7 will go out of support. VSSetup still seems to have .NET Core 3.1 and .NET 5, so I guess it's not likely that .NET 7 SDK will be removed anytime soon. However, I feel like allowing VSSetup to install the recommended/default version of the .NET SDK, and automatically targeting that version for our test project is a better solution that will never need maintainance, as long as the .NET SDK keeps populating and updating this MSBuild property.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception: manually ran OptProf pipeline to verify it works
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/test/NuGet.Tests.Apex/NuGet.OptProf/Assets/PackageReferenceSdk/PackageReferenceSdk/PackageReferenceSdk.csproj
+++ b/test/NuGet.Tests.Apex/NuGet.OptProf/Assets/PackageReferenceSdk/PackageReferenceSdk/PackageReferenceSdk.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net7.0</TargetFramework>
+    <TargetFramework>net$(BundledNETCoreAppTargetFrameworkVersion)</TargetFramework>
     <NuGetAudit>true</NuGetAudit>
   </PropertyGroup>
 


test/NuGet.Tests.Apex/NuGet.OptProf/Assets/PackageReferenceSdk/PackageReferenceSdk/PackageReferenceSdk.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net7.0</TargetFramework>
+    <TargetFramework>net$(BundledNETCoreAppTargetFrameworkVersion)</TargetFramework>
```
nkolev92, line 4: This property is new to me! :D

nkolev92 (approved):

martinrrm (approved):

donnie-msft (approved):
Great explanatory PR description! What you wrote sounds good. I'm not familiar with the property, and don't know how you figured out this property would do what you described.

test/NuGet.Tests.Apex/NuGet.OptProf/Assets/PackageReferenceSdk/PackageReferenceSdk/PackageReferenceSdk.csproj
zivkan: I ran `dotnet msbuild -pp:pp.txt`, then opened `pp.txt` in a text editor.

I first searched for `>net8.0<`, and I found one property, but it was in a targets file (so not available in the csproj), plus the property name started with an underscore, which is MSBuild convention for "do not use, it's meant to be internal, but MSBuild doesn't have scopes".

Next I searched for `>8.0<`, and I found `NETCoreAppMaximumVersion` and `BundledNETCoreAppTargetFrameworkVersion`. Interestingly I can't find them in github.com/dotnet/sdk, so I guess it's generated by the build, rather than commit into their source.

In any case, I mentioned it to them, and they said they don't recommend doing this. But in any case OptProf broke because of a change in VS's included SDK, so if another future change breaks it again, maybe we're no worse off than changing the vsconfig file to install the "correct" SDK and runtime for the version of .NET our test project targets. We could add a comment to the csproj saying update the vsconfig when the TargetFramework changes. Maybe that will be less fragile?  For now I'm willing to give this not-recommended way a try.

PR #5404, opened 2023-09-06
Vulnerabilities filter in PM UI

martinrrm:
## Bug

Fixes:

## Description
This PR adds a new checkbox that filters vulnerabilities. This also adds a Loading Indicator for vulnerabilities

### Solution Level

vulnerabilities-filter-solution-level-new

### Project Level
vulnerabilities-filter-project-level-new

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A Did manual testing, probably going to add them to vendors manual testing

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/GroupTypeToExpanderConverter.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/GroupTypeToExpanderConverter.cs
@@ -1,0 +1,27 @@
+using System;
+using System.Diagnostics;
+using System.Globalization;
+using System.Windows.Data;
+
+namespace NuGet.PackageManagement.UI
+{
+    internal class GroupTypeToExpanderConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            CollectionViewGroup grp = (CollectionViewGroup)value;
+            if (grp.Name.ToString().Equals("TopLevel", StringComparison.OrdinalIgnoreCase) || grp.Name.ToString().Equals("Transitive", StringComparison.OrdinalIgnoreCase))
+            {
+                return true;
+            }
+            return false;
+        }
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            // no op
+            Debug.Fail("Not Implemented");
+            return null;
+        }
+    }
+}
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
@@ -1032,4 +1032,7 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
   <data name="Text_PackageMappingsAutoCreate" xml:space="preserve">
     <value>A package source mapping will be created.</value>
   </data>
+  <data name="Checkbox_Show_Vulnerable_Only" xml:space="preserve">
+    <value>Show vulnerable only</value>
+  </data>
 </root>
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
@@ -1033,6 +1033,9 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
     <value>A package source mapping will be created.</value>
   </data>
   <data name="Checkbox_Show_Vulnerable_Only" xml:space="preserve">
-    <value>Show vulnerable only</value>
+    <value>Show only vulnerable</value>
+  </data>
+  <data name="Vulnerabilities_Loading" xml:space="preserve">
+    <value>Loading vulnerable packages...</value>
   </data>
 </root>
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml
@@ -23,6 +23,9 @@
       <nuget:PackageLevelToGroupNameConverter
       x:Key="PackageLevelToGroupNameConverter" />
 
+      <nuget:GroupTypeToExpanderConverter
+       x:Key="GroupTypeToExpanderConverter" />
+
       <DataTemplate
       DataType="{x:Type nuget:PackageDependencyMetadata}">
         <TextBlock
@@ -306,11 +309,22 @@
         </Expander>
       </ControlTemplate>
 
+      <ControlTemplate
+        x:Key="GroupedListItemNoExpanderControlTemplate"
+        TargetType="{x:Type GroupItem}">
+        <ItemsPresenter />
+      </ControlTemplate>
+
       <Style
         x:Key="GroupedListContainerStyle"
         TargetType="{x:Type GroupItem}">
         <Setter Property="Template" Value="{StaticResource GroupedListItemControlTemplate}" />
         <Setter Property="AutomationProperties.Name" Value="{StaticResource GroupHeaderMultiBinding}" />
+        <Style.Triggers>
+          <DataTrigger Binding="{Binding Converter={StaticResource GroupTypeToExpanderConverter}}" Value="False">
+            <Setter Property="Template" Value="{StaticResource GroupedListItemNoExpanderControlTemplate}" />
+          </DataTrigger>
+        </Style.Triggers>
       </Style>
     </ResourceDictionary>
   </UserControl.Resources>
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml
@@ -23,8 +23,8 @@
       <nuget:PackageLevelToGroupNameConverter
       x:Key="PackageLevelToGroupNameConverter" />
 
-      <nuget:GroupTypeToExpanderConverter
-       x:Key="GroupTypeToExpanderConverter" />
+      <nuget:PackageLevelGroupToBooleanConverter
+       x:Key="PackageLevelGroupToBooleanConverter" />
 
       <DataTemplate
       DataType="{x:Type nuget:PackageDependencyMetadata}">
@@ -321,7 +321,9 @@
         <Setter Property="Template" Value="{StaticResource GroupedListItemControlTemplate}" />
         <Setter Property="AutomationProperties.Name" Value="{StaticResource GroupHeaderMultiBinding}" />
         <Style.Triggers>
-          <DataTrigger Binding="{Binding Converter={StaticResource GroupTypeToExpanderConverter}}" Value="False">
+          <!--This data trigger is when a new group is added, if the group is not of type PackageLevel
+            it won't add a expander header and show it flat in the list -->
+          <DataTrigger Binding="{Binding Converter={StaticResource PackageLevelGroupToBooleanConverter}}" Value="False">
             <Setter Property="Template" Value="{StaticResource GroupedListItemNoExpanderControlTemplate}" />
           </DataTrigger>
         </Style.Triggers>
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml
@@ -322,7 +322,7 @@
         <Setter Property="AutomationProperties.Name" Value="{StaticResource GroupHeaderMultiBinding}" />
         <Style.Triggers>
           <!--This data trigger is when a new group is added, if the group is not of type PackageLevel
-            it won't add a expander header and show it flat in the list -->
+            it won't add a expander header and will instead show it flat in the list -->
           <DataTrigger Binding="{Binding Converter={StaticResource PackageLevelGroupToBooleanConverter}}" Value="False">
             <Setter Property="Template" Value="{StaticResource GroupedListItemNoExpanderControlTemplate}" />
           </DataTrigger>
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.ComponentModel;
+using System.Globalization;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
@@ -117,7 +118,7 @@ namespace NuGet.PackageManagement.UI
         {
             if (FilterByVulnerabilities)
             {
-                return item is not null && (item.IsPackageVulnerable == FilterByVulnerabilities);
+                return item is not null && item.IsPackageVulnerable;
             }
 
             return true;
@@ -233,7 +234,7 @@ namespace NuGet.PackageManagement.UI
             _logger = logger;
             _initialSearchResultTask = searchResultTask;
             _loadingStatusIndicator.Reset(loadingMessage);
-            _loadingVulnerabilitiesStatusIndicator.Reset("Loading Vulnerabilities Data...");
+            _loadingVulnerabilitiesStatusIndicator.Reset(string.Format(CultureInfo.CurrentCulture, Resx.Resources.Vulnerabilities_Loading));
             _loadingVulnerabilitiesStatusIndicator.Status = LoadingStatus.Loading;
             _loadingStatusBar.Visibility = Visibility.Hidden;
             _loadingStatusBar.Reset(loadingMessage, loader.IsMultiSource);
@@ -849,15 +850,6 @@ namespace NuGet.PackageManagement.UI
         private void Expander_ExpansionStateToggled(object sender, RoutedEventArgs e)
         {
             GroupExpansionChanged?.Invoke(sender, e);
-        }
-
-        protected void OnPropertyChanged(string propertyName)
-        {
-            if (PropertyChanged != null)
-            {
-                PropertyChangedEventArgs e = new PropertyChangedEventArgs(propertyName);
-                PropertyChanged(this, e);
-            }
         }
     }
 }
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
@@ -58,7 +58,7 @@ namespace NuGet.PackageManagement.UI
 
         private const string LogEntrySource = "NuGet Package Manager";
 
-        private bool FilterByVulnerabilities = false;
+        private bool _filterByVulnerabilities = false;
 
         // The count of packages that are selected
         private int _selectedCount;
@@ -86,31 +86,63 @@ namespace NuGet.PackageManagement.UI
 
             BindingOperations.EnableCollectionSynchronization(Items, _list.ItemsLock);
 
-            ItemsView = new CollectionViewSource() { Source = Items };
-            ListCollectionView itemsView = ItemsView.View as ListCollectionView;
+            ItemsView = new CollectionViewSource() { Source = Items }.View;
+            ICollectionViewLiveShaping itemsView = (ICollectionViewLiveShaping)ItemsView;
             itemsView.IsLiveFiltering = true;
             itemsView.IsLiveGrouping = true;
             itemsView.LiveFilteringProperties.Add(nameof(PackageItemViewModel.IsPackageVulnerable));
             itemsView.LiveGroupingProperties.Add(nameof(PackageItemViewModel.PackageLevel));
-            itemsView.Filter = item =>
+            ItemsView.Filter = item =>
             {
-                if (item is not null && item.Equals(_loadingVulnerabilitiesStatusIndicator))
-                {
-                    return FilterByVulnerabilities && !(_loadingVulnerabilitiesStatusIndicator.Status == LoadingStatus.NoItemsFound && VulnerablePackagesCount > 0);
-                }
-                PackageItemViewModel vitem = item as PackageItemViewModel;
-                return IsItemInFilter(vitem);
+                return IsItemAcecpted(item);
             };
 
-            DataContext = ItemsListCollection;
+            DataContext = itemsView;
             CheckBoxesEnabled = false;
 
             _loadingStatusIndicator.PropertyChanged += LoadingStatusIndicator_PropertyChanged;
         }
 
+
+        private bool IsItemAcecpted(object item)
+        {
+            if (item is LoadingStatusIndicator)
+            {
+                return IsLoadingIndicatorAccepted(item);
+            }
+
+            if (item is PackageItemViewModel vm)
+            {
+                return IsVulnerablePackageAccepted(vm);
+            }
+
+            return false;
+        }
+
+        private bool IsLoadingIndicatorAccepted(object item)
+        {
+            if (item is not null && item.Equals(_loadingStatusIndicator))
+            {
+                return !_filterByVulnerabilities;
+            }
+
+            if (item is not null && item.Equals(_loadingVulnerabilitiesStatusIndicator))
+            {
+                return _filterByVulnerabilities && !(_loadingVulnerabilitiesStatusIndicator.Status == LoadingStatus.NoItemsFound && VulnerablePackagesCount > 0);
+            }
+
+            return false;
+        }
+
+        private bool IsVulnerablePackageAccepted(PackageItemViewModel item)
+        {
+            return !_filterByVulnerabilities
+                || (_filterByVulnerabilities && item is PackageItemViewModel vm && vm.IsPackageVulnerable);
+        }
+
         internal bool IsItemInFilter(PackageItemViewModel item)
         {
-            if (FilterByVulnerabilities)
+            if (_filterByVulnerabilities)
             {
                 return item is not null && item.IsPackageVulnerable;
             }
@@ -148,14 +180,7 @@ namespace NuGet.PackageManagement.UI
 
         public ObservableCollection<object> Items { get; } = new ObservableCollection<object>();
 
-        public CollectionViewSource ItemsView { get; private set; }
-        public ListCollectionView ItemsListCollection
-        {
-            get
-            {
-                return ItemsView?.View as ListCollectionView;
-            }
-        }
+        public ICollectionView ItemsView { get; private set; }
 
         /// <summary>
         /// Count of Items (excluding Loading indicator) that are currently shown after applying any UI filtering.
@@ -185,7 +210,7 @@ namespace NuGet.PackageManagement.UI
         {
             get
             {
-                var group = ItemsListCollection.Groups.Where(g => (g as CollectionViewGroup).Name.ToString().Equals(PackageLevel.TopLevel.ToString(), StringComparison.OrdinalIgnoreCase)).FirstOrDefault();
+                var group = ItemsView.Groups.Where(g => (g as CollectionViewGroup).Name.ToString().Equals(PackageLevel.TopLevel.ToString(), StringComparison.OrdinalIgnoreCase)).FirstOrDefault();
                 return group is not null ? (group as CollectionViewGroup).ItemCount : 0;
             }
         }
@@ -194,7 +219,7 @@ namespace NuGet.PackageManagement.UI
         {
             get
             {
-                var group = ItemsListCollection.Groups.Where(g => (g as CollectionViewGroup).Name.ToString().Equals(PackageLevel.Transitive.ToString(), StringComparison.OrdinalIgnoreCase)).FirstOrDefault();
+                var group = ItemsView.Groups.Where(g => (g as CollectionViewGroup).Name.ToString().Equals(PackageLevel.Transitive.ToString(), StringComparison.OrdinalIgnoreCase)).FirstOrDefault();
                 return group is not null ? (group as CollectionViewGroup).ItemCount : 0;
             }
         }
@@ -804,19 +829,19 @@ namespace NuGet.PackageManagement.UI
 
         internal void ClearPackageLevelGrouping()
         {
-            ItemsListCollection.GroupDescriptions.Clear();
+            ItemsView.GroupDescriptions.Clear();
         }
 
         internal void AddVulnerabilitiesFiltering()
         {
-            FilterByVulnerabilities = true;
-            ItemsListCollection.Refresh();
+            _filterByVulnerabilities = true;
+            ItemsView.Refresh();
         }
 
         internal void RemoveVulnerabilitiesFiltering()
         {
-            FilterByVulnerabilities = false;
-            ItemsListCollection.Refresh();
+            _filterByVulnerabilities = false;
+            ItemsView.Refresh();
         }
 
         internal void RemoveVulnerabilitiesIndicator()
@@ -831,21 +856,21 @@ namespace NuGet.PackageManagement.UI
 
         internal void AddPackageLevelGrouping()
         {
-            if (ItemsListCollection.GroupDescriptions.Count > 0)
+            if (ItemsView.GroupDescriptions.Count > 0)
             {
                 return;
             }
 
-            ItemsListCollection.Refresh();
+            ItemsView.Refresh();
             if (Items
                     .OfType<PackageItemViewModel>()
                     .Any(p => p.PackageLevel == PackageLevel.Transitive))
             {
-                ItemsListCollection.GroupDescriptions.Add(new PropertyGroupDescription(nameof(PackageItemViewModel.PackageLevel)));
+                ItemsView.GroupDescriptions.Add(new PropertyGroupDescription(nameof(PackageItemViewModel.PackageLevel)));
             }
             else
             {
-                ItemsListCollection.GroupDescriptions.Add(new PropertyGroupDescription(nameof(LoadingStatusIndicator)));
+                ItemsView.GroupDescriptions.Add(new PropertyGroupDescription(nameof(LoadingStatusIndicator)));
             }
         }
 
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
@@ -36,7 +36,6 @@ namespace NuGet.PackageManagement.UI
         private readonly LoadingStatusIndicator _loadingVulnerabilitiesStatusIndicator = new LoadingStatusIndicator();
         private ScrollViewer _scrollViewer;
         private static TimeSpan PollingDelay = TimeSpan.FromMilliseconds(100);
-        public event PropertyChangedEventHandler PropertyChanged;
 
         public event SelectionChangedEventHandler SelectionChanged;
         public event RoutedEventHandler GroupExpansionChanged;
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
@@ -32,8 +32,10 @@ namespace NuGet.PackageManagement.UI
     public partial class InfiniteScrollList : UserControl
     {
         private readonly LoadingStatusIndicator _loadingStatusIndicator = new LoadingStatusIndicator();
+        private readonly LoadingStatusIndicator _loadingVulnerabilitiesStatusIndicator = new LoadingStatusIndicator();
         private ScrollViewer _scrollViewer;
         private static TimeSpan PollingDelay = TimeSpan.FromMilliseconds(100);
+        public event PropertyChangedEventHandler PropertyChanged;
 
         public event SelectionChangedEventHandler SelectionChanged;
         public event RoutedEventHandler GroupExpansionChanged;
@@ -84,14 +86,28 @@ namespace NuGet.PackageManagement.UI
 
             BindingOperations.EnableCollectionSynchronization(Items, _list.ItemsLock);
 
-            ItemsView = new CollectionViewSource() { Source = Items }.View;
-            ItemsView.Filter = item =>
+            ItemsView = new CollectionViewSource() { Source = Items };
+            ListCollectionView itemsView = ItemsView.View as ListCollectionView;
+            itemsView.IsLiveFiltering = true;
+            itemsView.IsLiveGrouping = true;
+            itemsView.LiveFilteringProperties.Add($"{nameof(PackageItemViewModel.IsPackageVulnerable)}");
+            itemsView.LiveGroupingProperties.Add($"{nameof(PackageItemViewModel.PackageLevel)}");
+            itemsView.Filter = item =>
             {
+                if (item is not null && item.Equals(_loadingVulnerabilitiesStatusIndicator))
+                {
+                    if (FilterByVulnerabilities)
+                    {
+                        return true;
+                    }
+                    return false;
+                }
+
                 PackageItemViewModel vitem = item as PackageItemViewModel;
                 return IsItemInFilter(vitem);
             };
 
-            DataContext = ItemsView;
+            DataContext = ItemsListCollection;
             CheckBoxesEnabled = false;
 
             _loadingStatusIndicator.PropertyChanged += LoadingStatusIndicator_PropertyChanged;
@@ -137,7 +153,14 @@ namespace NuGet.PackageManagement.UI
 
         public ObservableCollection<object> Items { get; } = new ObservableCollection<object>();
 
-        public ICollectionView ItemsView { get; private set; }
+        public CollectionViewSource ItemsView { get; private set; }
+        public ListCollectionView ItemsListCollection
+        {
+            get
+            {
+                return ItemsView?.View as ListCollectionView;
+            }
+        }
 
         /// <summary>
         /// Count of Items (excluding Loading indicator) that are currently shown after applying any UI filtering.
@@ -155,6 +178,8 @@ namespace NuGet.PackageManagement.UI
         /// </summary>
         public IEnumerable<PackageItemViewModel> PackageItems => Items.OfType<PackageItemViewModel>().ToArray();
 
+        public int VulnerablePackagesCount => Items.OfType<PackageItemViewModel>().Where(i => i.IsPackageVulnerable).Count();
+
         public PackageItemViewModel SelectedPackageItem => _list.SelectedItem as PackageItemViewModel;
 
         public int SelectedIndex => _list.SelectedIndex;
@@ -165,7 +190,8 @@ namespace NuGet.PackageManagement.UI
         {
             get
             {
-                return FilterByVulnerabilities is true ? TopLevelVulnerablePackagesCount : TopLevelAllPackagesCount;
+                var group = ItemsListCollection.Groups.Where(g => (g as CollectionViewGroup).Name.ToString().Equals("TopLevel", StringComparison.OrdinalIgnoreCase)).FirstOrDefault();
+                return group is not null ? (group as CollectionViewGroup).ItemCount : 0;
             }
         }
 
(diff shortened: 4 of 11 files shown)


src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
```diff
@@ -1032,4 +1032,7 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
   <data name="Text_PackageMappingsAutoCreate" xml:space="preserve">
     <value>A package source mapping will be created.</value>
   </data>
+  <data name="Checkbox_Show_Vulnerable_Only" xml:space="preserve">
+    <value>Show vulnerable only</value>
```
nkolev92: Have we validated this with @JonDouglas or @albarry4?

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -143,9 +161,29 @@ private int FilteredItemsCount
 
         public Guid? OperationId => _loader?.State.OperationId;
 
-        public int TopLevelPackageCount { get; private set; }
+        public int TopLevelPackageCount
+        {
+            get
+            {
+                return FilterByVulnerabilities is true ? TopLevelVulnerablePackagesCount : TopLevelAllPackagesCount;
```
nkolev92: Do we need separate properties? If we're applied the filter, why do we need the separate counts?

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
```diff
@@ -217,6 +217,7 @@ internal PackageSourceMoniker SelectedSource
 
         public INuGetExperimentationService NuGetExperimentationService { get; private set; }
 
+        public bool ShowVulnerabilities => _topPanel.CheckBoxVulnerabilities.IsChecked == true;
```
nkolev92: Is anything in `_topPanel.CheckBoxVulnerabilities.IsChecked` nullable? I'm guessing you don't need `== true`.

nkolev92 (commented):
I just have some questions. 

Btw, I think it'd be great to have designs in place for changes like this.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -83,12 +85,28 @@ internal InfiniteScrollList(Lazy<JoinableTaskFactory> joinableTaskFactory)
...
             CheckBoxesEnabled = false;
 
             _loadingStatusIndicator.PropertyChanged += LoadingStatusIndicator_PropertyChanged;
         }
 
+        internal bool IsItemInFilter(PackageItemViewModel item)
+        {
+            if (FilterByVulnerabilities)
+            {
+                return item is not null && (item.IsPackageVulnerable == FilterByVulnerabilities);
```
donnie-msft:
```suggestion
                return item is not null && item.IsPackageVulnerable;
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
donnie-msft:
```suggestion
                return FilterByVulnerabilities ? TopLevelVulnerablePackagesCount : TopLevelAllPackagesCount;
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
```diff
@@ -934,6 +941,12 @@ internal async Task SearchPackagesAndRefreshUpdateCountAsync(string searchText,
                 {
                     await RefreshInstalledAndUpdatesTabsAsync();
                 }
+                else
+                {
+                    // If warning icon is still visible, show the vulnerabilities filter.
+                    _topPanel.CheckBoxVulnerabilities.IsEnabled = _topPanel._warningIcon.IsVisible ? true : false;
```
donnie-msft: I challenge whether this is useful behavior. Disabling the filter just because there aren't items that meet the criteria may not be as beneficial as it sounds. It _could_ be confusing if customers want to leave it on and do multiple searches (or change other filters).

My vote is we leave the CheckBox enabled all the time. Then this bug isn't relevant: https://github.com/NuGet/Home/issues/12868

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -743,6 +783,18 @@ internal void ClearPackageLevelGrouping()
             ItemsView.GroupDescriptions.Clear();
         }
 
+        internal void AddVulnerabilitiesFiltering()
+        {
+            FilterByVulnerabilities = true;
+            ItemsView.Refresh();
```
donnie-msft, line 823: Did you see the filter become stale? I've got to play with this myself, but I recall only a particular scenario where this is necessary. The alternative is adding Live filtering to a list collection view:

`lcv.IsLiveFiltering = true;`
` lcv.LiveFilteringProperties.Add(your property)`

donnie-msft (commented):
This is pretty cool! 
Check a few of the booleans, I think there's some unnecessary `==true` and etc.
Also, think we should re-evaluate if disabling the filter is useful.

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
JonDouglas: Yep, this is fine as in IMO. Here's some alternatives if needed.

"Show vulnerable"

"Show vulnerable packages"

etc.

donnie-msft: "Show only vulnerable" might read better

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
martinrrm: This counts are for the collapsers/separators of TopLevel and Transitive's, sadly I'm not able to access the group lengths 😢. With these counters I'm updating how many packages are in each group depending on the Checkbox status

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
martinrrm: The problem with leaving it enabled is that users can use it when they are no vulnerabilities information.

Let say a user checks it and see no vulnerabilities, the user won't know that vulnerabilities data is still being fetch. The user, thinking that there are vulnerabilities, unchecks the checkbox before the data is available, in this scenario they will never know that he has vulnerabilities he can address in the PM UI.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
martinrrm: Yeah the filter became stale, I'll try `IsLiveFiltering` property, but probably is just doing a refresh too

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
donnie-msft: I agree that's a problem, but I disagree that the solution is disabling the checkbox. The solution would be to ensure a loading indicator remains active until vulnerability data is retrieved. 
https://github.com/NuGet/Home/issues/10811

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
donnie-msft: > Yeah the filter became stale, I'll try IsLiveFiltering property, but probably is just doing a refresh too

No, live filtering properties are subscribed to the `CollectionChanged` event. https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Data/ListCollectionView.cs#L1522

`Refresh`, however, re-creates the view. We should favor collection change over Refresh as much as possible.
https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Data/CollectionView.cs#L305

The only time I believe `Refresh` should be used in a case like this, is when the filter criteria itself is changing. In this case, the criteria is static, so I believe it's unnecessary.

Also, I think I shared this with you before, but I have a sample project I use to play with ListView behavior: 
https://github.com/donnie-msft/Prototypes/blob/add4f58ee9823648833b789c641df3255af4c653/BackgroundPrototype/BackgroundPrototype/ViewModels/PackageListViewModel.cs#L65
(let me know if I need to re-auth you to my private repo)

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
aortiz-msft: +1 to Donnie's suggestion :)

martinrrm:
@nkolev92 @donnie-msft @aortiz-msft Updated this PR, sorry for the delay 😞. I updated the PR description with new gifs, the checkbox is enabled always, and I added a vulnerabilities loader.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
```diff
@@ -893,6 +895,9 @@ internal async Task SearchPackagesAndRefreshUpdateCountAsync(string searchText,
             {
                 _packageList.ClearPackageLevelGrouping();
 
+                // Hide vulnerabilities checkbox if not in the isntalled tab
+                _topPanel.CheckBoxVulnerabilities.Visibility = ActiveFilter == ItemFilter.Installed ? Visibility.Visible : Visibility.Collapsed;
+
```
jebriede: This PR adds the checkbox for the vulnerabilities filter to the PackageManagerTopPanel. The PackageManagerTopPanel also owns the tabs and knows when they've changed. So, the PackageManagerTopPanel can and probably should be in charge of deciding when to show or hide its own controls. With this line, we are making it the responsibility of the PackageManagerControl to tell the PackageManagerTopPanel if it can hide or show its checkbox based on the state of the PackageManagerTopPanel's active tab.

It would make more sense to encapsulate the showing / hiding of the vulnerable packages filter checkbox in the PackageManagerTopPanel instead.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
```diff
@@ -1244,6 +1249,17 @@ private void Filter_SelectionChanged(object sender, FilterChangedEventArgs e)
...
+                if (ActiveFilter is not ItemFilter.Installed)
+                {
+                    _topPanel.CheckBoxVulnerabilities.Visibility = Visibility.Collapsed;
+                    _topPanel.CheckBoxVulnerabilities.IsChecked = false;
+                }
+                else
+                {
+                    _topPanel.CheckBoxVulnerabilities.Visibility = Visibility.Visible;
+                }
+
```
jebriede: I left another comment about having the TopPanel manage the visibility of its own checkboxes based on its own tab states.

Another comment is to de-duplicate the show/hide logic and reference it once. However, if we move the responsibility of managing the visibility of the checkbox into the TopPanel, we will be removing it from here, thereby addressing this comment.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -82,13 +86,43 @@ internal InfiniteScrollList(Lazy<JoinableTaskFactory> joinableTaskFactory)
...
+            itemsView.LiveGroupingProperties.Add($"{nameof(PackageItemViewModel.PackageLevel)}");
+            itemsView.Filter = item =>
+            {
+                if (item is not null && item.Equals(_loadingVulnerabilitiesStatusIndicator))
+                {
+                    if (FilterByVulnerabilities)
+                    {
+                        return true;
+                    }
+                    return false;
```
jebriede: More concise.

```suggestion
                return FilterByVulnerabilities;
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -82,13 +86,43 @@ internal InfiniteScrollList(Lazy<JoinableTaskFactory> joinableTaskFactory)
...
+                }
+
+                PackageItemViewModel vitem = item as PackageItemViewModel;
```
jebriede: Implicit casts suggest that it's okay for this not to be a PackageItemViewModel in which case vitem will be null. Is that what we want?

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -82,13 +86,43 @@ internal InfiniteScrollList(Lazy<JoinableTaskFactory> joinableTaskFactory)
...
+                return IsItemInFilter(vitem);
```
jebriede: More concise, avoids variable declaration that is only used once.

```suggestion
                return IsItemInFilter(item as PackageItemViewModel);
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
nkolev92: I'm guessing you meant to remove the `== true` check?

nkolev92 (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/GroupTypeToExpanderConverter.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+using System.Diagnostics;
+using System.Globalization;
+using System.Windows.Data;
+
+namespace NuGet.PackageManagement.UI
+{
+    internal class GroupTypeToExpanderConverter : IValueConverter
```
jebriede: The name of the converter should reflect what the converter is converting from and to. Can we rename this class to reflect that it's returning a boolean value?

src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/GroupTypeToExpanderConverter.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            CollectionViewGroup grp = (CollectionViewGroup)value;
+            if (grp.Name.ToString().Equals("TopLevel", StringComparison.OrdinalIgnoreCase) || grp.Name.ToString().Equals("Transitive", StringComparison.OrdinalIgnoreCase))
```
jebriede: We should match on the PackageLevel enum name since that's what's used when the group headers are created. That way, if we refactor or change the enum in the future, this code will update as well to still match.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml
```diff
@@ -306,11 +309,22 @@
...
       <Style
         x:Key="GroupedListContainerStyle"
         TargetType="{x:Type GroupItem}">
         <Setter Property="Template" Value="{StaticResource GroupedListItemControlTemplate}" />
         <Setter Property="AutomationProperties.Name" Value="{StaticResource GroupHeaderMultiBinding}" />
+        <Style.Triggers>
+          <DataTrigger Binding="{Binding Converter={StaticResource GroupTypeToExpanderConverter}}" Value="False">
+            <Setter Property="Template" Value="{StaticResource GroupedListItemNoExpanderControlTemplate}" />
+          </DataTrigger>
+        </Style.Triggers>
```
jebriede: Can we add a comment here that explains this data trigger is only supposed to trigger for when a new group is added for the vulnerabilities loading indicator, in which case it is supposed to not add the group header and just show the loading indicator in the flat list? Thanks!

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -82,13 +86,43 @@ internal InfiniteScrollList(Lazy<JoinableTaskFactory> joinableTaskFactory)
...
             BindingOperations.EnableCollectionSynchronization(Items, _list.ItemsLock);
 
-            ItemsView = new CollectionViewSource() { Source = Items }.View;
-            DataContext = ItemsView;
+            ItemsView = new CollectionViewSource() { Source = Items };
+            ListCollectionView itemsView = ItemsView.View as ListCollectionView;
+            itemsView.IsLiveFiltering = true;
+            itemsView.IsLiveGrouping = true;
+            itemsView.LiveFilteringProperties.Add($"{nameof(PackageItemViewModel.IsPackageVulnerable)}");
+            itemsView.LiveGroupingProperties.Add($"{nameof(PackageItemViewModel.PackageLevel)}");
```
jebriede: We can remove the string interpolation i.e. $""

```suggestion
            itemsView.LiveFilteringProperties.Add(nameof(PackageItemViewModel.IsPackageVulnerable));
            itemsView.LiveGroupingProperties.Add(nameof(PackageItemViewModel.PackageLevel));
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -137,15 +178,31 @@ private int FilteredItemsCount
...
         public int SelectedIndex => _list.SelectedIndex;
 
         public Guid? OperationId => _loader?.State.OperationId;
 
-        public int TopLevelPackageCount { get; private set; }
+        public int TopLevelPackageCount
+        {
+            get
+            {
+                var group = ItemsListCollection.Groups.Where(g => (g as CollectionViewGroup).Name.ToString().Equals("TopLevel", StringComparison.OrdinalIgnoreCase)).FirstOrDefault();
```
jebriede: Match on the enum value rather than a string literal so if the enum values change, this still works.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -137,15 +178,31 @@ private int FilteredItemsCount
...
+                return group is not null ? (group as CollectionViewGroup).ItemCount : 0;
+            }
+        }
 
-        public int TransitivePackageCount { get; private set; }
+        public int TransitivePackageCount
+        {
+            get
+            {
+                var group = ItemsListCollection.Groups.Where(g => (g as CollectionViewGroup).Name.ToString().Equals("Transitive", StringComparison.OrdinalIgnoreCase)).FirstOrDefault();
```
jebriede: Match on the enum value rather than a string literal so if the enum values change, this still works.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -82,13 +86,43 @@ internal InfiniteScrollList(Lazy<JoinableTaskFactory> joinableTaskFactory)
...
+                    if (FilterByVulnerabilities)
+                    {
+                        return true;
+                    }
+                    return false;
+                }
+
+                PackageItemViewModel vitem = item as PackageItemViewModel;
+                return IsItemInFilter(vitem);
+            };
```
jebriede: Rewrite this to delegate to two methods, one for filtering on the loading indicator, and one for the vulnerable packages. That way, each filter method does one thing and one thing well, rather than mixing the logic of checking for the vulnerability loading indicator and the vulnerable package into a single method.

Disclaimer: Please test the boolean logic. I did not test this, but this is the structure I recommend, and I believe the logic to be equivalent to what was in the PR.

```suggestion
            itemsView.Filter = item =>
            {
                return FilterLoadingIndicator(item)
                    && FilterVulnerablePackage(item);
            };
```

Then add two methods:

```
        private bool FilterLoadingIndicator(object item)
        {
            if (item is not null && item.Equals(_loadingVulnerabilitiesStatusIndicator))
            {
                return FilterByVulnerabilities;
            }

            return true;
        }

        private bool FilterVulnerablePackage(object item)
        {
            return !FilterByVulnerabilities
                || (FilterByVulnerabilities && item is PackageItemViewModel vm && vm.IsPackageVulnerable);
        }
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -82,13 +86,43 @@ internal InfiniteScrollList(Lazy<JoinableTaskFactory> joinableTaskFactory)
...
+
+            DataContext = ItemsListCollection;
             CheckBoxesEnabled = false;
 
             _loadingStatusIndicator.PropertyChanged += LoadingStatusIndicator_PropertyChanged;
         }
 
+        internal bool IsItemInFilter(PackageItemViewModel item)
```
jebriede: There is currently no reason for this method to be internal. Also, the name IsItemInFilter is a bit generic for the specific purpose of this method. However, in another comment I recommend a refactor that would eliminate the need for this method altogether.

jebriede (commented):
A few general concerns:

1) The functional design of this feature has us adding a checkbox for the packages that are vulnerable, but if we add another filter later, it could lead to checkbox overload in the UI. We should consider adding a filtering mechanism instead and a UI that would support said mechanism. Then adding a new filter should be as simple as defining a predicate.

2) The TopPanel control should expose a property for the vulnerabilities filter and the PackageManagerControl should read that property. Currently, we are responding to an event and setting boolean variables in response to the checkbox being toggled which is a lot of extra code and requires us to manage the state of the checkbox again. I'd recommend a design where the TopPanel binds to a property that is then exposed as a DependencyProperty.

I left a few other specific comments on suggested changes.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
martinrrm: This wasn't working for all scenarios and it was hard to understand where the bug was

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
martinrrm: _topPanel.CheckBoxVulnerabilities can be null 😢 I need to leave this

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerTopPanel.xaml.cs
```diff
@@ -297,6 +311,18 @@ private void TabsPackageManagement_SelectionChanged(object sender, SelectionChan
 
             //Store the tag for calculating the ItemFilter without having to access the UI Thread.
             Filter = GetItemFilter(selectedTabItem);
+            if(CheckBoxVulnerabilities is not null) // UI Element can be null
```
jeffkl:
```suggestion
            if (CheckBoxVulnerabilities is not null) // UI Element can be null 
```

jeffkl (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -82,13 +86,38 @@ internal InfiniteScrollList(Lazy<JoinableTaskFactory> joinableTaskFactory)
...
+            ListCollectionView itemsView = ItemsView.View as ListCollectionView;
+            itemsView.IsLiveFiltering = true;
+            itemsView.IsLiveGrouping = true;
+            itemsView.LiveFilteringProperties.Add(nameof(PackageItemViewModel.IsPackageVulnerable));
+            itemsView.LiveGroupingProperties.Add(nameof(PackageItemViewModel.PackageLevel));
+            itemsView.Filter = item =>
+            {
+                if (item is not null && item.Equals(_loadingVulnerabilitiesStatusIndicator))
+                {
+                    return FilterByVulnerabilities && !(_loadingVulnerabilitiesStatusIndicator.Status == LoadingStatus.NoItemsFound && VulnerablePackagesCount > 0);
```
martinrrm: This is a hotfix for https://github.com/NuGet/Client.Engineering/issues/2491. The error here is that changing tabs doesn't necessarily trigger a vulnerability fetch to the web, instead we do a local fetch that is another code area that updates the PackageItem IsVulnerable property, this triggers the filter and we can hide the loading indicator.

Created a follow up issue https://github.com/NuGet/Home/issues/12906

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -291,10 +286,14 @@ private async Task RepopulatePackageListAsync(PackageItemViewModel selectedPacka
                 if (!Items.Contains(_loadingStatusIndicator))
                 {
                     Items.Add(_loadingStatusIndicator);
-                    Items.Add(_loadingVulnerabilitiesStatusIndicator);
```
martinrrm: Moved this to not depend on the other loadingIndicator

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -354,14 +362,9 @@ private async Task RepopulatePackageListAsync(PackageItemViewModel selectedPacka
                     {
                         _loadingStatusIndicator.Status = LoadingStatus.NoItemsFound;
                     }
-                    else if (VulnerablePackagesCount == 0)
```
martinrrm: Moved this to not depend on the other loadingIndicator

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml
```diff
@@ -306,11 +309,24 @@
...
+      </ControlTemplate>
+
       <Style
         x:Key="GroupedListContainerStyle"
         TargetType="{x:Type GroupItem}">
         <Setter Property="Template" Value="{StaticResource GroupedListItemControlTemplate}" />
         <Setter Property="AutomationProperties.Name" Value="{StaticResource GroupHeaderMultiBinding}" />
+        <Style.Triggers>
+          <!--This data trigger is when a new group is added, if the group is not of type PackageLevel
+            it won't add a expander header and show it flat in the list -->
```
jebriede: As it's currently written, it is saying that it won't show it in a flat list.

```suggestion
            it won't add a expander header and will instead show it flat in the list -->
```

kartheekp-ms:
> vulnerabilities-filter-project-level-new

I think the `Show only vulnerable` option should be checked by default when users click on in the vulnerabilities InfoBar. What do you think?

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerTopPanel.xaml.cs
```diff
@@ -188,6 +188,8 @@ public void UpdateCountOnUpdatesTab(int count)
 
         public CheckBox CheckboxPrerelease => _checkboxPrerelease;
 
+        public CheckBox CheckBoxVulnerabilities => _checkboxVulnerabilities;
```
jebriede, line 191: We should not be exposing the control directly as a public property. I know this is done for the other checkboxes, but it's the wrong thing to do. We should instead be defining properties on this TopPanel control. Those properties would be [Dependency Properties](https://learn.microsoft.com/en-us/dotnet/desktop/wpf/properties/dependency-properties-overview?view=netdesktop-7.0) and would be accessible in XAML by a parent control or in code.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
jebriede: The variable name ShowVulnerabilities makes me think this determines if vulnerability data should be shown for packages or if vulnerable packages should be shown. It's actually a filter that filters out any packages that are not vulnerable. Can we rename this property so it isn't confused with showing vulnerability data or not?

Some name like:
IsVulnerableFilteringApplied
ShouldFilterOnVulnerablePackages
or some other name that makes it more clear that this determines filtering the list on vulnerable packages.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -291,10 +286,14 @@ private async Task RepopulatePackageListAsync(PackageItemViewModel selectedPacka
...
                 {
                     Items.Add(_loadingStatusIndicator);
-                    Items.Add(_loadingVulnerabilitiesStatusIndicator);
                     addedLoadingIndicator = true;
                 }
 
+                if (!Items.Contains(_loadingVulnerabilitiesStatusIndicator))
+                {
+                    Items.Add(_loadingVulnerabilitiesStatusIndicator);
+                }
```
jebriede, line 305: Will this new loading indicator throw off the list counts? There is code for the other loading indicator that helps determine when the list is empty if the loading indicator has been added and the list is 1, but we don't have the same mechanism for this.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -56,6 +58,8 @@ public partial class InfiniteScrollList : UserControl
 
         private const string LogEntrySource = "NuGet Package Manager";
 
+        private bool FilterByVulnerabilities = false;
```
jebriede: Violates naming for private instance variables

```suggestion
        private bool _filterByVulnerabilities = false;
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
jebriede: This was marked as resolved but not addressed. Please leave comment or fix. Thanks!

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
martinrrm: Not the list counts for Telemetry and counters, we usually use this property that only counts package items https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs#L138.


About the other count for the loading indicator, that is being easily avoided, I'm always removing the vulnerabilities indicator before the loading indicator, so the count stays the same.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerTopPanel.xaml.cs
martinrrm: Thanks for this! This looks way better than we are currently doing, I'm not going to address this in this PR due to time constraints, but we should definitely look into doing this change to some of our code.

martinrrm:
> > vulnerabilities-filter-project-level-new
> >       
> >         [
> >           
> >         
> >             vulnerabilities-filter-project-level-new
> >           ](https://user-images.githubusercontent.com/43253759/270048440-9c7304fe-1c83-488a-8024-30e328089e88.gif)
> >         
> >         
> >           
> >             
> >               
> >             
> >             
> >               
> >               
> >             
> >           
> >           [
> >             
> >               
> >             
> >           ](https://user-images.githubusercontent.com/43253759/270048440-9c7304fe-1c83-488a-8024-30e328089e88.gif)
> 
> I think the `Show only vulnerable` option should be checked by default when users click on in the vulnerabilities InfoBar. What do you think?

@kartheekp-ms Thats the next step, I'll create a task to follow this, unfortunately the current way we open the PM UI cannot handle this, but I have some ideas on what we can do.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -86,31 +86,63 @@ internal InfiniteScrollList(Lazy<JoinableTaskFactory> joinableTaskFactory)
...
-            itemsView.Filter = item =>
+            ItemsView.Filter = item =>
             {
-                if (item is not null && item.Equals(_loadingVulnerabilitiesStatusIndicator))
-                {
-                    return FilterByVulnerabilities && !(_loadingVulnerabilitiesStatusIndicator.Status == LoadingStatus.NoItemsFound && VulnerablePackagesCount > 0);
-                }
-                PackageItemViewModel vitem = item as PackageItemViewModel;
-                return IsItemInFilter(vitem);
+                return IsItemAcecpted(item);
```
jebriede:
```suggestion
                return IsItemAccepted(item);
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -86,31 +86,63 @@ internal InfiniteScrollList(Lazy<JoinableTaskFactory> joinableTaskFactory)
...
 
-            DataContext = ItemsListCollection;
+            DataContext = itemsView;
             CheckBoxesEnabled = false;
 
             _loadingStatusIndicator.PropertyChanged += LoadingStatusIndicator_PropertyChanged;
         }
 
+
+        private bool IsItemAcecpted(object item)
```
jebriede:
```suggestion
        private bool IsItemAccepted(object item)
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
martinrrm: @jebriede I did something a little bit different but use most of it, doing `FilterLoadingIndicator(item) && FilterVulnerablePackage(item)` was causing some weird behavior, I decided to rely on if's instead

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -86,31 +86,63 @@ internal InfiniteScrollList(Lazy<JoinableTaskFactory> joinableTaskFactory)
...
+                return IsLoadingIndicatorAccepted(item);
+            }
+
+            if (item is PackageItemViewModel vm)
+            {
+                return IsVulnerablePackageAccepted(vm);
+            }
+
+            return false;
+        }
```
jebriede: I don't think this method is needed at all if we handle casting in each method.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -86,31 +86,63 @@ internal InfiniteScrollList(Lazy<JoinableTaskFactory> joinableTaskFactory)
...
+                return _filterByVulnerabilities && !(_loadingVulnerabilitiesStatusIndicator.Status == LoadingStatus.NoItemsFound && VulnerablePackagesCount > 0);
+            }
+
+            return false;
+        }
+
+        private bool IsVulnerablePackageAccepted(PackageItemViewModel item)
+        {
+            return !_filterByVulnerabilities
+                || (_filterByVulnerabilities && item is PackageItemViewModel vm && vm.IsPackageVulnerable);
```
jebriede: You're casting item to PackageItemViewModel but the method signature already tells us this is a PackageItemViewModel. Personally, I'd opt for casting within this method, but if you're going to have the outer method casting before we enter this IsVulnerablePackageAccepted method, we don't need to cast again in here.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -86,31 +86,63 @@ internal InfiniteScrollList(Lazy<JoinableTaskFactory> joinableTaskFactory)
...
+
+            if (item is PackageItemViewModel vm)
+            {
+                return IsVulnerablePackageAccepted(vm);
+            }
+
+            return false;
+        }
+
+        private bool IsLoadingIndicatorAccepted(object item)
```
jebriede: There's inconsistencies here. This method accept a parameter of type object, but the caller ensures the object is of type LoadingStatusIndicator. Then we check again if this is not null. This is inconsistent with how  IsVulnerablePackageAccepted is handling casting. Let's decide on one way and stick with it for all filter methods.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
```diff
@@ -217,7 +217,7 @@ internal PackageSourceMoniker SelectedSource
 
         public INuGetExperimentationService NuGetExperimentationService { get; private set; }
 
-        public bool ShowVulnerabilities => _topPanel.CheckBoxVulnerabilities.IsChecked == true;
+        public bool IsVulnerableFilteringApplied => _topPanel.CheckBoxVulnerabilities.IsChecked == true;
```
jebriede: I believe I saw a comment from you earlier that said that CheckBoxVulnerabilities can be null. If that's the case, we can't dereference it to check IsChecked without guarding against null.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -137,15 +192,31 @@ private int FilteredItemsCount
         /// </summary>
         public IEnumerable<PackageItemViewModel> PackageItems => Items.OfType<PackageItemViewModel>().ToArray();
 
+        public int VulnerablePackagesCount => Items.OfType<PackageItemViewModel>().Where(i => i.IsPackageVulnerable).Count();
```
jebriede: This is public but it's not used anywhere else but in this class.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -83,12 +87,63 @@ internal InfiniteScrollList(Lazy<JoinableTaskFactory> joinableTaskFactory)
...
+
+        internal bool IsItemInFilter(PackageItemViewModel item)
+        {
+            if (_filterByVulnerabilities)
+            {
+                return item is not null && item.IsPackageVulnerable;
+            }
+
+            return true;
+        }
```
jebriede: Unused? Remove if no references.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -743,15 +826,46 @@ internal void ClearPackageLevelGrouping()
...
+        internal void RemoveVulnerabilitiesFiltering()
+        {
+            _filterByVulnerabilities = false;
+            ItemsView.Refresh();
+        }
+
+        internal void RemoveVulnerabilitiesIndicator()
+        {
+            Items.Remove(_loadingVulnerabilitiesStatusIndicator);
+        }
```
jebriede: Unused.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -743,15 +826,46 @@ internal void ClearPackageLevelGrouping()
...
+
+        internal void UpdateNoVulnerabilitiesFoundIndicator()
+        {
+            _loadingVulnerabilitiesStatusIndicator.Status = LoadingStatus.NoItemsFound;
+        }
```
jebriede: Unused.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -743,15 +826,46 @@ internal void ClearPackageLevelGrouping()
...
+            if (Items
                     .OfType<PackageItemViewModel>()
                     .Any(p => p.PackageLevel == PackageLevel.Transitive))
             {
                 ItemsView.GroupDescriptions.Add(new PropertyGroupDescription(nameof(PackageItemViewModel.PackageLevel)));
             }
+            else
+            {
+                ItemsView.GroupDescriptions.Add(new PropertyGroupDescription(nameof(LoadingStatusIndicator)));
+            }
```
jebriede: I don't think this is needed. We talked offline about it.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -743,15 +826,46 @@ internal void ClearPackageLevelGrouping()
...
+            _loadingVulnerabilitiesStatusIndicator.Status = LoadingStatus.NoItemsFound;
+        }
+
         internal void AddPackageLevelGrouping()
         {
+            if (ItemsView.GroupDescriptions.Count > 0)
+            {
+                return;
+            }
+
```
jebriede: Please check if this is needed. I don't believe we'll be in a situation where we call this method and have groups already.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
```diff
@@ -527,6 +528,7 @@ private void ApplySettings(UserSettings settings, ISettings nugetSettings)
             _detailModel.Options.RemoveDependencies = settings.RemoveDependencies;
             _detailModel.Options.ForceRemove = settings.ForceRemove;
             _topPanel.CheckboxPrerelease.IsChecked = settings.IncludePrerelease;
+            _topPanel.CheckBoxVulnerabilities.IsChecked = false;
```
jebriede: The defaults should be set by the TopPanel that manages this checkbox and not by the PackageManagerControl. See previous feedback about managing the checkbox state in the TopPanel.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
martinrrm: Removing this is not needed

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
martinrrm: Removing this line, the default of the checkbox is false so no need for this

jebriede (approved):

PR #5406, opened 2023-09-07
Protocol to convert VulnerabilityInfoResource with System.Text.Json

### LINKED ISSUE #12855: NuGetAudit: read vulnerability files with System.Text.Json
### NuGet Product(s) Affected

MSBuild.exe, dotnet.exe

### Current Behavior

The initial implementation of NuGetAudit was done quickly, to meet shipping dates. This meant using Newtonsoft.Json to read the vulnerability files, since we don't yet use System.Text.Json anywhere in NuGet.Protocol.

### Desired Behavior

Here's the results of a quick benchmark I put together:

|        Method |              Runtime |       Mean |    Error |   StdDev |     Gen0 |    Gen1 |  Allocated |
|-------------- |--------------------- |-----------:|---------:|---------:|---------:|--------:|-----------:|
|  NJ  |             .NET 8.0 | 1,819.6 us |  5.19 us |  4.86 us |  68.3594 | 37.1094 |  707.38 KB |
| STJ|             .NET 8.0 |   954.7 us | 18.84 us | 23.82 us |  15.6250 |  4.8828 |  168.33 KB |
|  NJ | .NET Framework 4.7.2 | 3,653.8 us | 19.91 us | 16.62 us | 183.5938 | 66.4063 | 1146.55 KB |
| STJ | .NET Framework 4.7.2 | 2,126.5 us |  8.92 us |  7.91 us |  66.4063 | 19.5313 |  419.91 KB |

Restore is optimized to read the file once per restore, no matter how many projects in the solution. So, a once-off 1ms saving isn't going to be noticeable to anyone. However, perhaps in a real restore, when other work is happening on other threads, reducing GC pauses will have a more significant impact.

### Additional Context

Things to check:

* Does MSBuild already ship System.Text.Json.dll, or do we have to include it in our VSIX for MSBuild CLI scenarios?
* Risk of VS perf test regression if nothing else loads System.Text.Json.dll in a measured scenario. It's not feasible to port ALL Newtonsoft.Json usage to System.Text.Json in one PR, so from NuGet's perspective it will be an additional assembly loaded. However, there is a very good chance that some VS platform component, like settings, accounts, etc, already uses System.Text.Json, so the assembly will probably be loaded already.
* Any other compliance work?

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12855
 
Regression? no

## Description

See linked issue for benchmarks of Newtonsoft.Json vs System.Text.Json benchmarks for the vulnerability data. Basically, on .NET Framework, there are about 1/3rd of the allocations, and on .NET 8 it's more like 1/4. Technically it's a little faster too, but the wall clock different is meaningless, because if you take a trace of any NuGet action, JSON deserialization doesn't use up any significant time.

This is the first time we're using System.Text.Json in NuGet, so some infrastructure had to be set up to make everything work properly. I didn't verify, but I expect that all our ILMerged assemblies have bloated in size because of this. If we ever migrate all other Newtonsoft.Json usage, then we can remove it and reclaim its size.

|Component|Changes|
|--|--|
|NuGet.exe ILMerge|No changes, it automatically ILMerges *.dll in the bin directory|
|NuGet.MSSign.exe|Updated ilmerge.props for the new dll, plus its new dependencies|
|Pack ILMerge|no changes. I can't think of any scenario where pack will use NuGet.Protocol, so I don't think it's needed. Internestingly ilmerge didn't complain about missing assemblies, so perhaps it also includes bin's *.dll, like nuget.exe does|
|VSIX|It's already in vsixignore, as it's a dependency of other packages we already use. System.Text.Json is part of VS's public assemblies, so no need for it in our VSIX for VS experiences. MSBuild has its own copy, so no need for msbuild CLI restore either|

I had to add `SuppressTfmSupportBuildWarnings` to the common project properties, because a lot of system .NET v7 packages [have props/targets that error out on TFMs that are "out of support"](https://github.com/dotnet/runtime/blob/1aecc383f38b58acb8e256eae19b217b7e951b6d/eng/packaging.targets#L209-L214), but the package still provides binaries for, and our test projects are still targeting .NET 5 and .NET Core 3.1.

Created a sharable `JsonSerializerOptions` in `JsonExtensions`, similar to how we have a Newtonsoft.Json serializer, so future protocol migrations to System.Text.Json might not need to create its own. New converters will need to be added, but by using the shared options, it will only be a one time effort per converter.

Only one `JsonConverter` was needed for VulnerabilityInfoResource, `VersionRangeConverter`, named `VersionRangeStjConverter` due to the name clash with the Newtonsoft.Json converter (both extend classes, not interfaces, so we can't easily have one that implements both). Added tests for the new converter, but otherwise else existing tests to cover NuGetAudt's functionality.

Finally, the changes to VulnerabilityInfoResourceV3 to use System.Text.Json vs Newtonsoft.Json was minimal, almost trivial.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -189,6 +189,7 @@
       <_allowBuildFromSourcePackage Include="System.Security.Cryptography.Pkcs" />
       <_allowBuildFromSourcePackage Include="System.Security.Cryptography.ProtectedData" />
       <_allowBuildFromSourcePackage Include="System.Security.Cryptography.Xml" />
+      <_allowBuildFromSourcePackage Include="System.Text.Json" />
 
       <_sourceBuildUnexpectedPackage Include="@(PackageReference)" Exclude="@(_allowBuildFromSourcePackage)" />
     </ItemGroup>
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -4,6 +4,7 @@
 
     <MicrosoftCodeAnalysisPublicApiAnalyzersVersion Condition="'$(MicrosoftCodeAnalysisPublicApiAnalyzersVersion)' == ''">3.3.4</MicrosoftCodeAnalysisPublicApiAnalyzersVersion>
     <NewtonsoftJsonPackageVersion Condition="'$(NewtonsoftJsonPackageVersion)' == ''">13.0.3</NewtonsoftJsonPackageVersion>
+    <SystemTextJsonVersion Condition="'$(SystemTextJsonVersion)' == ''">7.0.3</SystemTextJsonVersion>
     <SystemPackagesVersion Condition="'$(SystemPackagesVersion)' == ''">4.3.0</SystemPackagesVersion>
     <VSFrameworkVersion Condition="'$(VSFrameworkVersion)' == ''">17.4.33103.184</VSFrameworkVersion>
     <VSServicesVersion Condition="'$(VSServicesVersion)' == ''">16.153.0</VSServicesVersion>
@@ -99,6 +100,7 @@
     <PackageVersion Include="System.Security.Cryptography.ProtectedData" Version="4.4.0" />
     <PackageVersion Include="System.Security.Cryptography.Xml" Version="$(SystemSecurityCryptographyXmlVersion)" />
     <PackageVersion Include="System.Text.Encoding.Extensions" Version="$(SystemPackagesVersion)" />
+    <PackageVersion Include="System.Text.Json" Version="$(SystemTextJsonVersion)" />
     <PackageVersion Include="System.Threading" Version="$(SystemPackagesVersion)" />
     <PackageVersion Include="System.Threading.Tasks" Version="$(SystemPackagesVersion)" />
     <!--
--- a/build/common.project.props
+++ b/build/common.project.props
@@ -48,6 +48,7 @@
     <NoWarn>$(NoWarn);NU5105;MSB3277;NETSDK1138</NoWarn>
     <!-- additional warnings new in .NET 6 that we need to disable when building with source-build -->
     <NoWarn Condition="'$(DotNetBuildFromSource)' == 'true'">$(NoWarn);CS1998;CA1416;CS0618;CS1574</NoWarn>
+    <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings>
   </PropertyGroup>
 
   <!-- Defaults -->
--- a/eng/SourceBuildPrebuiltBaseline.xml
+++ b/eng/SourceBuildPrebuiltBaseline.xml
@@ -95,4 +95,8 @@
     <UsagePattern IdentityGlob="System.Threading.Tasks/*" />
     <UsagePattern IdentityGlob="System.Threading.Tasks.Dataflow/*" />
   </IgnorePatterns>
+  <Usages>
+    <Usage Id="Microsoft.Bcl.AsyncInterfaces" Version="7.0.0" />
+    <Usage Id="System.Threading.Tasks.Extensions" Version="4.5.4" />
+  </Usages>
 </UsageData>
--- a/eng/SourceBuildPrebuiltBaseline.xml
+++ b/eng/SourceBuildPrebuiltBaseline.xml
@@ -25,6 +25,7 @@
     <!--
       Packages owned by .NET - exclusion of all versions.
     -->
+    <UsagePattern IdentityGlob="Microsoft.Bcl.AsyncInterfaces/*" />
     <UsagePattern IdentityGlob="Microsoft.Build/*" />
     <UsagePattern IdentityGlob="Microsoft.Build.Framework/*" />
     <UsagePattern IdentityGlob="Microsoft.Build.Tasks.Core/*" />
@@ -94,9 +95,6 @@
     <UsagePattern IdentityGlob="System.Threading/*" />
     <UsagePattern IdentityGlob="System.Threading.Tasks/*" />
     <UsagePattern IdentityGlob="System.Threading.Tasks.Dataflow/*" />
+    <UsagePattern IdentityGlob="System.Threading.Tasks.Extensions/*" />
   </IgnorePatterns>
-  <Usages>
-    <Usage Id="Microsoft.Bcl.AsyncInterfaces" Version="7.0.0" />
-    <Usage Id="System.Threading.Tasks.Extensions" Version="4.5.4" />
-  </Usages>
 </UsageData>
--- a/src/NuGet.Core/NuGet.Protocol/Converters/VersionRangeStjConverter.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Converters/VersionRangeStjConverter.cs
@@ -1,0 +1,39 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+#nullable enable
+
+using System;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using NuGet.Versioning;
+
+namespace NuGet.Protocol.Converters
+{
+    internal class VersionRangeStjConverter : JsonConverter<VersionRange>
+    {
+        public override VersionRange Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            var stringValue = reader.GetString();
+            if (stringValue == null)
+            {
+                // This is actually impossible to get to, because JsonSerializer won't call into the converter when the JSON is null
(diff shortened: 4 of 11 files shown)


test/NuGet.Core.Tests/NuGet.Protocol.Tests/Converters/VersionRangeStjConverterTests.cs
```diff
@@ -0,0 +1,62 @@
...
+
+using System;
+using System.Text.Json;
+using FluentAssertions;
+using NuGet.Versioning;
+using Xunit;
+
+namespace NuGet.Protocol.Tests.Converters
+{
+    public class VersionRangeStjConverterTests
```
donnie-msft, line 14: Would it be worth testing that calling `Write` and then `Read` results in the same input value?
Also, the same input to `VersionRange.Parse` and `VersionRangeStjConverter` results in the same output?

donnie-msft (dismissed):

src/NuGet.Core/NuGet.Protocol/Converters/VersionRangeStjConverter.cs
```diff
@@ -0,0 +1,39 @@
...
+#nullable enable
+
+using System;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using NuGet.Versioning;
+
+namespace NuGet.Protocol.Converters
+{
+    internal class VersionRangeStjConverter : JsonConverter<VersionRange>
```
jeffkl, line 16: nit: XML doc comments to quickly explain what this class is for

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
```diff
@@ -46,14 +45,10 @@ public async Task<IReadOnlyList<V3VulnerabilityIndexEntry>> GetVulnerabilityFile
...
-                    result =>
+                    async result =>
                     {
-                        using (var textReader = new StreamReader(result.Stream))
-                        using (var jsonReader = new JsonTextReader(textReader))
-                        {
-                            var parsed = JsonExtensions.JsonObjectSerializer.Deserialize<IReadOnlyList<V3VulnerabilityIndexEntry>>(jsonReader);
-                            return Task.FromResult(parsed);
-                        }
+                        var parsed = await JsonSerializer.DeserializeAsync<IReadOnlyList<V3VulnerabilityIndexEntry>>(result.Stream, JsonExtensions.JsonSerializerOptions);
```
jeffkl, line 50: What's our policy on using `return await`?  Or is this just for ease of debugging?  Just curious, as my preference is `return await`.

jeffkl (dismissed):
I'm excited migrating to `System.Text.Json` wasn't a ton of code churn at least.

src/NuGet.Core/NuGet.Protocol/Resources/VulnerabilityInfoResourceV3.cs
zivkan: Personally, I hate `return await`, because I want to optimize code for debuggability. There's no runtime performance penalty.  Same reason I hate calling a method inside another method's parameter list. Variables are free.

But I don't remember anything in our repo coding guidelines about it one way or another.

eng/SourceBuildPrebuiltBaseline.xml
```diff
@@ -95,4 +95,8 @@
     <UsagePattern IdentityGlob="System.Threading.Tasks/*" />
     <UsagePattern IdentityGlob="System.Threading.Tasks.Dataflow/*" />
   </IgnorePatterns>
+  <Usages>
+    <Usage Id="Microsoft.Bcl.AsyncInterfaces" Version="7.0.0" />
+    <Usage Id="System.Threading.Tasks.Extensions" Version="4.5.4" />
+  </Usages>
```
zivkan: @NikolaMilosavljevic @MichaelSimons

I copied the automatically generated baseline, like the build script told me to when run locally.  Should I change it to use the `UsagePattern` items like all the above, given these are both .NET owned packages?

Directory.Packages.props
```diff
@@ -4,6 +4,7 @@
 
     <MicrosoftCodeAnalysisPublicApiAnalyzersVersion Condition="'$(MicrosoftCodeAnalysisPublicApiAnalyzersVersion)' == ''">3.3.4</MicrosoftCodeAnalysisPublicApiAnalyzersVersion>
     <NewtonsoftJsonPackageVersion Condition="'$(NewtonsoftJsonPackageVersion)' == ''">13.0.3</NewtonsoftJsonPackageVersion>
+    <SystemTextJsonVersion Condition="'$(SystemTextJsonVersion)' == ''">7.0.3</SystemTextJsonVersion>
```
zivkan, line 7: @NikolaMilosavljevic @MichaelSimons

Also, is this variable name going to work for Source Build?

eng/SourceBuildPrebuiltBaseline.xml
zivkan: ~This worked in WSL, but is failing on CI? 😭~

no, I was looking at a previous build.

jeffkl (dismissed):

Directory.Packages.props
```diff
@@ -4,6 +4,7 @@
```
nkolev92, line 4: > I can't think of any scenario where pack will use NuGet.Protocol, so I don't think it's needed. Internestingly ilmerge didn't complain about missing assemblies, so perhaps it also includes bin's *.dll, like nuget.exe does

Regarding the pack  ilmerge, pack reads the assets file, so I'm guessing we'll eventually want it there too, but I'm guessing it's fine for now.

build/common.project.props
```diff
@@ -48,6 +48,7 @@
     <NoWarn>$(NoWarn);NU5105;MSB3277;NETSDK1138</NoWarn>
     <!-- additional warnings new in .NET 6 that we need to disable when building with source-build -->
     <NoWarn Condition="'$(DotNetBuildFromSource)' == 'true'">$(NoWarn);CS1998;CA1416;CS0618;CS1574</NoWarn>
+    <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings>
```
nkolev92, line 51: Should this be in the test projects only?

src/NuGet.Core/NuGet.Protocol/Converters/VersionRangeStjConverter.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        public override VersionRange Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            var stringValue = reader.GetString();
+            if (stringValue == null)
+            {
+                // This is actually impossible to get to, because JsonSerializer won't call into the converter when the JSON is null
+                throw new JsonException("Value for version range cannot be null");
```
nkolev92: InvalidOperationException instead in that case?

Directory.Packages.props
MichaelSimons: Yes this appears to use the correct package naming convention for version properties.

eng/SourceBuildPrebuiltBaseline.xml
MichaelSimons: These should be added in the section above using the wildcard syntax.

```
    <UsagePattern IdentityGlob="Microsoft.Bcl.AsyncInterfaces/*" />
    <UsagePattern IdentityGlob="System.Threading.Tasks.Extensions/*" />
```

nkolev92 (dismissed):

Directory.Packages.props
zivkan: I just checked, and it's already merging `*.dll`: https://github.com/NuGet/NuGet.Client/blob/40648c3cb3de37cf62b194accdc8e7117b36245f/src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.csproj#L101

It doesn't need to though. .NET (Core) runtime already ships the DLL (and in fact, it won't be int he publish output, since I conditioned the package on netfx and netstandard). On .NET Framework, msbuild already ships the dll, so if it wasn't found in the merged assembly, the assembly loader should still find it in msbuild's directory.

build/common.project.props
zivkan: No, all the product assemblies that have different APIs for package signing target net5.0 (well, netcoreapp5.0. I really want to change that one day).

nkolev92 (approved):

PR #5409, opened 2023-09-07
Add event tracing to NuGet.Common and write events for the NuGet-based MSBuid project SDK resolver

### LINKED ISSUE #11445: [Feature]: Add more logging to NuGetSdkResolver
### NuGet Product(s) Involved

MSBuild.exe, dotnet.exe

### The Elevator Pitch

I'm adding more logging to the MSBuild APIs that call SDK resolvers which will also allow our SDK resolver to add more logging.

Dependent on https://github.com/dotnet/msbuild/issues/7136

_No response_

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Home/issues/11445

## Description

This adds a new `EventSource` to NuGet.Common and writes events for the NuGet-based MSBuild project SDK resolver.

I also implemented the following [Keywords](https://learn.microsoft.com/dotnet/api/system.diagnostics.tracing.eventattribute.keywords):
| Keyword | Purpose |
|---|---|
| Common | Any event coming from a component in the `NuGet.Common` namespace |
| Configuration | Any event coming from a component in the `NuGet.Configuration` namespace |
| Logging | Any event that is logging a message |
| Performance | Any event related to measuring performance |
| SdkResolver | Any event coming from a component in the NuGet-based MSBuild project SDK resolver |

This will allow users to filter out potential components that aren't necessary for a particular trace.

Users will be able to enable the tracing with tools like [PerfView](https://github.com/microsoft/perfview/blob/main/documentation/Downloading.md):

```
PerfView /OnlyProviders=*Microsoft-NuGet run dotnet restore
```
The resulting Event Trace Log (.etl) will be useful when diagnosing issues with it.  The above command also includes traces from MSBuild and there are additional sources like Visual Studio which include more information.

Here is a sample trace log you can view in tools like PerfView: 
[PerfViewData.etl.zip](https://github.com/NuGet/NuGet.Client/files/12590906/PerfViewData.etl.zip)

image
image
image

Its also possible to create these traces with [Visual Studio](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/eventsource-collect-and-view-traces#visual-studio) and [`dotnet trace`](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/eventsource-collect-and-view-traces#dotnet-trace)

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


jeffkl:
FYI @ladipro and @lifengl

src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/NuGetSdkResolverEventSource.cs
```diff
@@ -0,0 +1,190 @@
...
+using System.Diagnostics.Tracing;
+using NuGet.Common;
+
+namespace Microsoft.Build.NuGetSdkResolver
+{
+    /// <summary>
+    /// Represents an <see cref="EventSource" /> for the NuGet-based MSBuild project SDK resolver.
+    /// </summary>
+    [EventSource(Name = "Microsoft-Build-NuGetSdkResolver")]
+    internal sealed class NuGetSdkResolverEventSource : EventSource
```
zivkan: I think it was someone from the VS Perf & Reliability team who once told me that Windows has a max number of ETW providers that can run at a time, so they very much discourage having fine-grained providers, as they've had to investigate customer issues where it caused problems.

Therefore, I propose that instead we create an ETW provider in NuGet.Common, that we can re-use throughout all of `src/NuGet.Core/*` projects (well, except for NuGet.Framework and NuGet.Versioning, but I don't see those assemblies emitting ETW events as useful).

src/NuGet.Core/NuGet.Configuration/Properties/AssemblyInfo.cs
```diff
@@ -15,3 +15,4 @@
...
 [assembly: InternalsVisibleTo("Test.Utility, PublicKey=002400000480000094000000060200000024000052534131000400000100010007d1fa57c4aed9f0a32e84aa0faefd0de9e8fd6aec8f87fb03766c834c99921eb23be79ad9d5dcc1dd9ad236132102900b723cf980957fc4e177108fc607774f29e8320e92ea05ece4e821c0a5efe8f1645c4c0c93c1ab99285d622caa652c1dfad63d745d6f2de5f17e5eaf0fc4963d261c8a12436518206dc093344d5ad293")]
 [assembly: InternalsVisibleTo("NuGet.Commands.Test, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
+[assembly: InternalsVisibleTo("Microsoft.Build.NuGetSdkResolver, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
```
zivkan: Why is this needed?

If the NuGet SDK resolver needs APIs from NuGet.Configuration, then as much as I hate excessive public APIs, it seems to be that it should be a public API. Using `InternalsVisibleTo` for reasons other than testing feels like an anti-pattern, and makes it far more likely that the target assembly will be able to accidentally take on dependencies on APIs in Configuration that really should be internal only.

src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/NuGetSdkResolver.cs
```diff
@@ -51,6 +60,9 @@ public NuGetSdkResolver()
         internal NuGetSdkResolver(IGlobalJsonReader globalJsonReader)
         {
             _globalJsonReader = globalJsonReader;
+
+            _globalJsonReader.FileReadStart += (_, args) => NuGetSdkResolverEventSource.Instance.GlobalJsonReadStart(args.FullPath, args.ProjectFullPath, args.SolutionFullPath);
+            _globalJsonReader.FileReadStop += (_, args) => NuGetSdkResolverEventSource.Instance.GlobalJsonReadStop(args.FullPath, args.ProjectFullPath, args.SolutionFullPath);
```
zivkan: If we need the events, then these lambda functions should be converted to named methods, and then unsubscribed from the events in a Dispose method.

src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/GlobalJsonReader.cs
```diff
@@ -216,14 +214,43 @@ internal static bool TryGetPathOfFileAbove(string file, DirectoryInfo startingDi
...
+        /// <param name="solutionFullPath">The full path to solution file that cause the file to be read.</param>
+        private void OnFileReadStart(string filePath, string projectFullPath, string solutionFullPath)
         {
-            EventHandler<string> fileReadEventHandler = FileRead;
+            EventHandler<GlobalJsonFileReadEventArgs> fileReadStartEventHandler = FileReadStart;
 
-            fileReadEventHandler?.Invoke(this, filePath);
+            fileReadStartEventHandler?.Invoke(
+                this,
+                new GlobalJsonFileReadEventArgs
```
zivkan: Granted we don't expect json files to be read super often, especially because of caching. However, adding an event, just so an event listener replays the event through EventSource. How about this code just calls EventSource directly instead?

Again, we don't expect this to be called very often, so this will have negligible impact, but this code is allocating a `GlobalJsonFileReadEventArgs` instance, and `OnFileReadStop` allocates a second instance, that needs to be GC'd afterwards. Even if there is nothing listening to the EventSource's events. If this class stops using events, and invokes the EventSource directly, it doesn't need to allocate these event arg instances.

src/NuGet.Core/NuGet.Configuration/Properties/AssemblyInfo.cs
nkolev92: Because of `SettingsLoadContext.FileRead`. IMO, if your proposal for declaring a provider in NuGet.Common is accepted, then this can go away too, since SettingsLoadContext.FileRead can report an event itself directly.

src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/NuGetSdkResolverEventSource.cs
jeffkl: > I think it was someone from the VS Perf & Reliability team who once told me that Windows has a max number of ETW providers that can run at a time, so they very much discourage having fine-grained providers, as they've had to investigate customer issues where it caused problems.

Do you happen to know where we can verify this?  As far as I know, you can specify the exact providers you want which would essentially limit them per session.  But I'm interested to know where you heard this so we can decide what to do.  My preference is to have this new provider but I'm open to suggestion.

src/NuGet.Core/NuGet.Configuration/Properties/AssemblyInfo.cs
jeffkl: Yeah alternatively I could make `SettingsLoadingContext.FileRead` public, I don't think it would really have any ramifications as its just an event that's fired and not something an end user would directly call.  We just could never change the signature...

src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/NuGetSdkResolver.cs
jeffkl: The SDK resolver is never disposed of and is expected to be cleaned up along with the parent process when it exits.  But I could move these calls to the `GlobalJsonReader` itself, I was just trying to separate concerns and not sprinkle event tracing all over the place.  But that would keep me from having to introduce the new `GlobalJsonFileReadEventArgs` type.

src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/GlobalJsonReader.cs
jeffkl: Yeah my default inclination was to keep the owner of tracing to the single component, `NuGetSdkResolver`.  But you're right, the `GlobalJsonReader` is only used in this assembly and could probably just call the tracing events directly.

src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/NuGetSdkResolver.cs
zivkan: Unsubscribing from events is best practise. We've had [memory leaks in the past](https://github.com/NuGet/Home/issues/8471), and may well again, so people in the team should be calling what looks like bugs. If this class has special justification why it shouldn't follow best practises, it should be documented with comments. Not everyone is familar with SDK resolvers, and their lifecycle, and once this PR is merged, they won't have the context of this PR's comments. The PR description didn't explain this object lifecycle, but even if it did, once the PR is merged that context gets lost.

However, unless following the well known pattern will cause problems, I'd still suggest implementing the best practise, so that if static code analysis tools are run, they don't get tripped up and force a person to then investigate and find the comments.

jeffkl:
@zivkan and @nkolev92 so if I update this and move the event provider to NuGet.Common, make it public, update the event names, then fire the events where needed, you'll approve?

nkolev92:
I like that idea.
 
@zivkan has some experience with the events and he did a lot of work on the VS side with, so he's probably the best person to review it.

zivkan:
Yes, I'd approve without hesitation.  [I've been thinking about how we can monitor NuGet better since mid to late 2020](https://github.com/NuGet/Client.Engineering/pull/579), but there's always higher priority work. But adding an EventSource so we can start using it more is very much needed. It probably would have been nice to add some instrumentation into NuGetAudit to get detailed perf & diagnostic data in case someone ends up reporting a difficult to reproduce bug.

The only thing I haven't thought about is naming conventions for events. If we're going to have all NuGet assemblies using one provider, we need event names to include the assembly name, or feature area, so we don't end up with dumb event names like `NuGet/ReadFile/Start` and we don't know if it's reading a config file, or http cache file, or global.json, or something else. While we can put the filename being read as an event property, if I'm trying to investigate slow or failing nuget.config load errors, I don't want global.json and http cache reads spamming the event list, so I think it's better to have more events. From memory, ETW has a max of 32k events allowed per provider. We're not going to reach that limit.

Since the `EventSource` will unfortunately need to be in NuGet.Common's public API, my advice is to have a public `EventSource` instance, which we will call `.Write(` on. Similar to what I already added in our VS code: https://github.com/NuGet/NuGet.Client/blob/11a62174beb847bc2a59ffbb08b70eaa84781d25/src/NuGet.Clients/NuGet.VisualStudio.Common/Etw/NuGetETW.cs#L16

While this PR's first iteration uses a class that extends `EventSource`,  and adds methods to make emiting the events nicer, since we need to make our common provider part of our client SDK's public API, I don't think we should need to change public APIs every time someone adds/removes/modifies some EventSource event. Not only will it reduce PR workload, but it also makes NuGet much less fragile to mismatched dll versions. While we could argue that anyone using the CLient SDK should use the same package version for all the packages, I'd rather not introduce this as potential failure.

src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/NuGetSdkResolverEventSource.cs
jeffkl: I have moved this to `NuGet.Common`

jeffkl:
This has now been moved to `NuGet.Common`.

src/NuGet.Core/NuGet.Common/NuGetEventSource.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        /// <summary>
+        /// Represents a class for declaring event keywords. Each keyword must be a flag (2^N) for use in a bitwise operation.
+        /// </summary>
+        public static class Keywords
+        {
+            /// <summary>
+            /// The event keyword for tracing events related to NuGet's common functionality from the NuGet.Common namespace.
+            /// </summary>
+            public const EventKeywords Common = (EventKeywords)1;
```
nkolev92, line 30: So guessing this in response to @zivkan's suggestion that we have differentiators for each assembly?

nkolev92 (dismissed):
LGTM. 

The only thing I could think of is whether the EventKeywords need to be defined in common instead maybe in the assemblies themselves.

src/NuGet.Core/NuGet.Common/NuGetEventSource.cs
jeffkl: This doesn't quite "differentiate" the events but it does let you filter out events you might not care about depending on the scenario.  I have no idea if we'll ever use them but I can imagine if we have a ton of events then we might ask a customer for only NuGetSdkResolver events for example.

jeffkl:
> The only thing I could think of is whether the EventKeywords need to be defined in common instead maybe in the assemblies themselves.

Keywords have to be a flags so I was worried that if they weren't declared in one place then we could end up with collisions.  I can move them but their just constants that we'll never change so I wasn't as worried about them being part of the public API

src/NuGet.Core/NuGet.Configuration/Settings/SettingsFile.cs
```diff
@@ -108,15 +114,50 @@ public SettingsFile(string directoryPath, string fileName, bool isMachineWide, b
...
+                ConfigurationSettingsFileRead,
+                new EventSourceOptions
+                {
+                    ActivityOptions = EventActivityOptions.Detachable,
+                    Keywords = NuGetEventSource.Keywords.Configuration,
+                    Opcode = EventOpcode.Start,
+                },
+                new
+                {
+                    ConfigFilePath,
```
zivkan: anonymous types are classes, and therefore allocated on the heap.  Should we use `EventSource.IsEnabled(EventLevel, EventKeywords)` to avoid the allocation when nothing is listening to the event?

Same question applies to all the other usage of anonymous types for EventSource events, but this one, reading nuget.config files, is the one that will run most frequently.

Since `EventSource.Write` with a `data` overload all use generics, it means that if you define a struct for the data, then it should be possible to write the EventSource event with zero allocations. But it requires a bit more boilerplate, defining a struct for the data.

src/NuGet.Core/NuGet.Configuration/Settings/SettingsFile.cs
```diff
@@ -108,15 +114,50 @@ public SettingsFile(string directoryPath, string fileName, bool isMachineWide, b
...
+                    ConfigurationSettingsFileRead,
+                    new EventSourceOptions
+                    {
+                        ActivityOptions = EventActivityOptions.Detachable,
+                        Keywords = NuGetEventSource.Keywords.Configuration,
+                        Opcode = EventOpcode.Stop,
+                    },
+                    new
+                    {
+                        ConfigFilePath,
```
zivkan: The anonymous type used for the stop event's data appears to be the same as the start event's data. If we keep this as a class, that gets a heap allocation, we should at least re-use the same instance.

zivkan (dismissed):

src/NuGet.Core/NuGet.Configuration/Settings/SettingsFile.cs
jeffkl: Thanks, I've changed the pattern to use a private class for event methods and a `record struct` for event data.  I think the code looks much better now.

zivkan (dismissed):

nkolev92 (dismissed):

src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/GlobalJsonReader.cs
```diff
@@ -238,38 +240,79 @@ private void OnFileRead(string filePath)
...
             {
-                return ParseMSBuildSdkVersionsFromJson(json);
+                var eventOptions = new EventSourceOptions
+                {
+                    ActivityOptions = EventActivityOptions.Detachable,
+                    Keywords = NuGetEventSource.Keywords.SdkResolver | NuGetEventSource.Keywords.Performance,
+                    Opcode = EventOpcode.Start
+                };
+
+                NuGetEventSource.Instance.Write(EventNameGlobalJsonRead, eventOptions, new GlobalJsonReadEventData(globalJsonPath, sdkResolverContext.ProjectFilePath, sdkResolverContext.SolutionFilePath));
```
ladipro, line 299: nit: Please consider wrapping the body of these methods in `if (NuGetEventSource.Instance.IsEnabled()`.

The struct initialization here is probably negligible but `Write<T>()` will be JITted for each different event data even if tracing is not enabled and the effect will likely be measurable in some scenarios (e.g. running the resolver in MSBuild.exe where we currently can't use a native image).

jeffkl: Thanks, I assumed that `Write<T>()` would just check `IsEnabled()` first?  But you're saying the actual call to `Write<T>()` will potentially cause a perf issue?

ladipro: Yes, the method internally checks `IsEnabled()` but it needs to be JITted first, and since it's generic and instantiated with a different value type in each of these call sites, it will be JITted several times. Not a huge deal .. but easily avoidable.

ladipro: Thank you for making the change!

nkolev92 (approved):

zivkan (approved):

ladipro (approved):

PR #5417, opened 2023-09-21
Add nullable annotations to NuGet.LibraryModel

### LINKED ISSUE #12889: Add nullable annotations to NuGet.LibraryModel
[NuGet.LibraryModel](https://www.nuget.org/packages/NuGet.LibraryModel) is a package that is part of the [NuGet Client SDK](https://learn.microsoft.com/en-us/nuget/reference/nuget-client-sdk). We are adding null annotations, so projects using [nullable reference types](https://learn.microsoft.com/dotnet/csharp/nullable-references) will benefit from compiler null checking, leading to improved software reliability (fewer NullReferenceExceptions).

Generally when annotating nulls we try to avoid making any breaking API or ABI changes. However, the following changes have been made in order to allow the compiler to provide everyone with the most useful nullable warnings.

Projects that use this package and enable nullable checks might find that upgrading to a version of the package that now has nullable annotations will generate warnings that previously did not generate warnings. This is a natural consequence of annotating the assembly where it previously lacked annotations.

In addition, the following breaking changes were implemented:
 
## Using the `required` modifier on some class properties

The following classes are affected: `Library`, `LibraryDependency`, `LibraryIdentity`, `LibraryRange`

[C#11 introduced the `required` modifier](https://learn.microsoft.com/dotnet/csharp/language-reference/keywords/required). It means that projects creating instances of these type will be required to use the [object initializer syntax](https://learn.microsoft.com/dotnet/csharp/programming-guide/classes-and-structs/object-and-collection-initializers), and if any property marked as `required` is not assigned a value, the compiler will generate an error. We also provide constructors with all the required properties as constructor parameters. If you are not creating instances of these types, you should be unaffected, even if your project uses lower versions of the C# language.

While my own testing was not exhaustive, it appears that projects targeting lower C# language versions will still provide correct required property compiler checks as long as the compiler supports C#11. Therefore, if your project builds with the latest .NET SDK or Visual Studio (including MSBuild for command line builds), you should be able to use either the object initializer syntax, or a suitable constructor.

However, if you're using an older version of Visual Studio, MSBuild, or the .NET SDK (for example, using a *global.json* file to pin to .NET 6 SDK), you will get a compiler error saying the constructor is obsolete when using the default constructor, or any other constructor that does not have a parameter for every required property. In this case, you should use a different constructor overload that the compiler does not consider obsolete.

zivkan:
Note: this will not be merged until the dev branch targets V6.9

## Bug

Fixes: https://github.com/NuGet/Home/issues/12889

Regression? No

## Description

Enable nullable in NuGet.LibraryModel, and its test project.

This PR uses the `required` keyword for the first time in our code. This is a C#11 feature, and the compiler needs attributes that are only available in .NET 7's BCL. Therefore, I had to copy these attributes from the .NET runtime repo into this repo. For more info on the impact on customers, please see the linked issue. I wrote a long description in there. Note that this does create an API breaking change, but there's little reason for other repos to create instances of these classes, except maybe for tests.

Since I needed to add a new file in `build\Shared`, I noticed that a bunch of other files in that directory were not listed in the solution's virtual directory, so I added all of them, not just the one new file I added.

Otherwise, this was the easiest project to enable nullable on so far, but the project is mostly models, not business logic, and has a lot less code than other projects. From memory, the nullable checks didn't find any bugs, and other than the `required` keyword, I don't remember any other breaking changes.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


build/Shared/IsExternalInit.cs
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-#if NETFRAMEWORK || NETSTANDARD
+#if !NET5_0_OR_GREATER
```
zivkan, line 4: One of the test projects targeting netcoreapp3.1 started failing once I added the `RequiredModiferAttributes` file, hence I needed to fix this, as this `IsExernalInit` attribute was added in .NET 5.

src/NuGet.Core/NuGet.LibraryModel/DownloadDependency.cs
```diff
@@ -35,8 +35,9 @@ public sealed class DownloadDependency : IEquatable<DownloadDependency>, ICompar
             };
         }
 
-        public int CompareTo(DownloadDependency other)
+        public int CompareTo(DownloadDependency? other)
         {
+            if (other == null) return 1;
```
zivkan, line 40: this code would previously throw if `other` was null. But, the interface for `IComparable<T>` shows an expectation of handling nulls, and the docs explain that nulls should be sorted after non null values.  We can change this to throw an ArgumentNullException if we want, but to avoid a compiler warning/error, we have to keep the `?` in the method signature, since `IComparable<T>` does.

src/NuGet.Core/NuGet.LibraryModel/FrameworkDependencyFlagsUtils.cs
```diff
@@ -52,13 +52,13 @@ public static string GetFlagString(FrameworkDependencyFlags flags)
...
         /// </summary>
-        public static FrameworkDependencyFlags GetFlags(string flags)
+        public static FrameworkDependencyFlags GetFlags(string? flags)
         {
             var result = FrameworkDependencyFlags.None;
 
             if (!string.IsNullOrEmpty(flags))
             {
-                var splitFlags = flags.Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
+                var splitFlags = flags!.Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
```
zivkan, line 61: Two lines up, `!string.ISNullOrEmpty` prevents `flags` being null here. In case you didn't review my other PRs enabling nullable in other projects, only the .NET 5 BCL has annotated `string.IsNullOrEmpty`, hence on .NET Standard and .NET Framework (and theoretically .NET Core 3.1 and earlier, for our tests), the compiler doesn't know, hence we have to tell it with the `!`

src/NuGet.Core/NuGet.LibraryModel/Library.cs
```diff
@@ -31,18 +45,25 @@ public class Library
 
         public override string ToString()
         {
-            if (Identity == null)
+            if (Identity is null)
             {
-                return LibraryRange?.ToString();
+                if (LibraryRange is null)
+                {
+                    return typeof(Library).FullName!;
```
zivkan, line 52: This is a change in behaviour. `ToString()` shouldn't return null, hence I changed the code to never return the null. However, both `Identity` and `LibraryRange` are required, so this shouldn't be a possible codepath anyway.

src/NuGet.Core/NuGet.LibraryModel/Library.cs
```diff
@@ -31,18 +45,25 @@ public class Library
...
             return Identity + " (" + LibraryRange + ")";
         }
 
         private class LibraryIdentityComparer : IEqualityComparer<Library>
         {
-            public bool Equals(Library x, Library y)
+            public bool Equals(Library? x, Library? y)
             {
+                if (x is null && y is null) return true;
+                if (x is null || y is null) return false;
```
zivkan, line 65: This is also a change in behaviour. Previously it would throw a null reference exception. But I think this is better.

src/NuGet.Core/NuGet.LibraryModel/LibraryIdentity.cs
```diff
@@ -86,8 +88,10 @@ public override int GetHashCode()
             };
         }
 
-        public int CompareTo(LibraryIdentity other)
+        public int CompareTo(LibraryIdentity? other)
         {
+            if (other is null) return 1;
```
zivkan, line 93: another example of code that previously threw NullReferenceException, violating the `IComparable<T>` contract.

src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
```diff
@@ -39,18 +40,25 @@ public class LibraryDependency : IEquatable<LibraryDependency>
...
-        public VersionRange VersionOverride { get; set; }
+        public VersionRange? VersionOverride { get; set; }
 
         public LibraryDependency()
         {
             NoWarn = new List<NuGetLogCode>();
         }
 
+        [SetsRequiredMembers]
+        public LibraryDependency(LibraryRange libraryRange) : this()
```
jeffkl, line 61: Should we add XML doc comments indicating that specifying a null `libraryRange` will throw an `ArgumentNullException`?  Or is that overkill here?

src/NuGet.Core/NuGet.LibraryModel/LibraryExtensions.cs
```diff
@@ -16,9 +16,9 @@ public static bool IsEclipsedBy(this LibraryRange library, LibraryRange other)
                 && string.Equals(library.Name, other.Name, StringComparison.OrdinalIgnoreCase);
         }
 
-        public static T GetItem<T>(this Library library, string key)
+        public static T? GetItem<T>(this Library library, string key)
         {
-            object value;
+            object? value;
```
jeffkl, line 21: I like this syntax better personally if you want to update this class to use it:

```c#
return library.Items.TryGetValue(key, out object? value) ? value as T : default(T);
```

src/NuGet.Core/NuGet.LibraryModel/PublicAPI.Shipped.txt
```diff
@@ -1,76 +1,76 @@
 #nullable enable
 NuGet.LibraryModel.CentralPackageVersion
-~NuGet.LibraryModel.CentralPackageVersion.CentralPackageVersion(string name, NuGet.Versioning.VersionRange versionRange) -> void
-~NuGet.LibraryModel.CentralPackageVersion.Equals(NuGet.LibraryModel.CentralPackageVersion other) -> bool
-~NuGet.LibraryModel.CentralPackageVersion.Name.get -> string
-~NuGet.LibraryModel.CentralPackageVersion.VersionRange.get -> NuGet.Versioning.VersionRange
+NuGet.LibraryModel.CentralPackageVersion.CentralPackageVersion(string! name, NuGet.Versioning.VersionRange! versionRange) -> void
```
jeffkl, line 3: Just curious, what's our policy on this?  I wonder if these should be moved to the unshipped category instead since they're not technically shipped?  I guess this was automatically done by the analyzer?  What's your opinion?

jeffkl (dismissed):

src/NuGet.Core/NuGet.LibraryModel/Library.cs
kartheekp-ms: If `Identity` and `LibraryRange` can't be null (because they are required), I am curious to know why we need this if block. Can we remove this if block?

zivkan: With all the nullable annotation PRs, I generally try to minimize changes, except to keep the compiler happy. So, I intentionally did not refactor old code, in order to avoid the PR doing too many things, and to save time (other projects took many, many hours to annotate and fix compiler errors).

Technically, marking properties as `required` will only break existing code that tries to compile, referencing the new assembly with the `required` attribute. The binaries are still ABI compatible, and therefore any binary that already creates instances of this class without setting a value, will have null values for these properties, despite it being marked as `required`. But as mentioned several times elsewhere, the likelyhood of anything other that NuGet.Packaging.dll creating instances of these classes is really low, so it's probably safe to change.

I can change it if the team prefers, but the above is an explanation of why I didn't do so earlier.

src/NuGet.Core/NuGet.LibraryModel/LibraryExtensions.cs
zivkan: Firstly, `value as T` is a pet peeve of mine. 😁 Also see [this](https://github.com/NuGet/NuGet.Client/pull/5309). Unfortunately, this method allows nulls to be returned (otherwise `default(T)` would result in a compiler error, now that this project has nullable enabled), so `value as T` won't trigger the nullable compiler error. However, if the value has the wrong type, it will silently be returned as null, and the app will have the wrong behaviour, which will be difficult to diagnose. Using `(T)value` instead will throw an InvalidCastException, which will make it very clear what went wrong, and the line of code and stack trace, making it very easy to diagnose.

Secondly, as mentioned in another comment a few seconds ago, in order to save time (other projects took many hours to fix compiler errors), I didn't spend time refactoring.

I can make these small changes if preferred, but that's the reason I didn't before creating the PR.

src/NuGet.Core/NuGet.LibraryModel/PublicAPI.Shipped.txt
zivkan: These API have already shipped. This `CentralPackageVersion` constructor already existed. The method signature didn't change, except for being annotated. Since that specific method already also would throw ArgumentNullExceptions when a null is passed in, there's little chance that anyone will experience a breaking change when they upgrade to the version of this package with annotations. Other APIs which don't throw ArgumentNullExceptions have higher change to cause compile time errors for people upgrading, but only if their own projects enable nullable checks. If their projects don't enable nullable, then our package being annotated will have no impact on them.  Adding the `required` modifier is the most significant breaking change, but those properties already existed, and the PublicApiAnalizer didn't want to decorate the method signature in these PublicAPI files.

And null annotations are a compile-time only concept, there's no ABI changes, so binaries that were compiled against older versions of our package will still reference these methods when our dll is replaced with the newer version.

IMO, the goal of the Unshipped file is to signal that it's no risk of breaking anyone if APIs listed in that file are changed before the next release. However, as explained, that's not true for these APIs. The annotations are more "superficial", rather than "structural".

src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
zivkan: I'm not seeing the text within any exception docs, including in other XMLDoc that already exists, in VS 2022 Int Preview, public preview, or GA versions 😕 
image

jeffkl: Interesting, I expected the tooltip of the Exception to show the text but yeah I can't figure out how the text is displayed in the IDE.  It must only be for docs? 🤷

jeffkl (approved):

nkolev92 (approved):

PR #5420, opened 2023-09-22
Prefer updating only the conditioned packages when updating in VS and dotnet add package

nkolev92:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2237

Covers large parts of https://github.com/NuGet/Home/issues/4681 - due to the urgency of this fix, I'll validate all scenarios in that issues have been addressed afterwards and I'll have any follow up fixes if needed. 

## Description

Note that this started as a larger change, so I had written up a spec on this, however as I was implementing it and adding tests I discovered the accidental bug introduced in 4.4! You can read more in depth in https://github.com/NuGet/NuGet.Client/pull/5420, but it should not be necessary to review this PR. 

For a conditionally installed PackageReference, NuGet does not update conditional reference correctly consistenly. Currently all updates of conditional packages lead to attempting to install the package to all frameworks, thus often causing an NU1504 warning. The change here is for the update gesture to update a package only in the frameworks it's installed in.

Note that this behavior appears to be a regression of <https://github.com/NuGet/NuGet.Client/commit/f9569accef0c1746ecc2d88499112c0f8dce46de>.
I verified that `1.1.14` of .NET SDK does not have the problem at hand. It is extremely likely that Visual Studio did not have the same issues for updating, but very likely had issues where no-op was constantly incorrect after a package installation.

The preview step boils down to a call to `PackageSpecOperation.AddOrUpdate(..)`.
In <https://github.com/NuGet/NuGet.Client/commit/f9569accef0c1746ecc2d88499112c0f8dce46de>, the AddOrUpdate call was changed to *always* add to all framework. That was done because the AddOrUpdate without frameworks did not update the correct list for PackageReference based projects.
PackageReference based projects *always* use the TargetFrameworkInformation object dependency list.
The correct fix is to update the original AddOrUpdate without frameworks method to account for PackageReference based projects.

- The AddPackageReferenceCommandRunner changes affect `dotnet add package`.
- The NuGetPackageManager changes affect PM UI & PMC. 

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added - Added a bunch, but I want more. Sounds like Apex tests may be fixed soon, so I may add some there, or maybe just resort to some manual tests. https://github.com/NuGet/Client.Engineering/issues/2487
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


jeffkl (approved):

kartheekp-ms:
I have reviewed the changes and about to click approve but noticed that PR has been merged.  The proposed changes look good👍

PR #5426, opened 2023-09-27
Add Apex test -- Search/Install/Uninstall transitive package from PMUI

CiciLi1:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2493

## Description

Add apex test for SDK project to Search/Install/Uninstall transitive package from PM UI to verify that  package's transitive package can be Search/installed/uninstalled successfully and exists at the appropriate level.

## PR Checklist

- [X] PR has a linked issue.
- [X] Described changes

- **Tests**
  - [X] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [X] N/A 

- **Documentation**
  
  - **OR**
  - [X] N/A


test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetUITestCase.cs
```diff
@@ -714,5 +714,91 @@ public void UninstallDeprecatedPackageFromUI(ProjectTemplate projectTemplate, st
...
+            solutionService.CreateEmptySolution();
+            var project = solutionService.AddProject(ProjectLanguage.CSharp, ProjectTemplate.NetCoreClassLib, "TestProject");
+            VisualStudio.ClearOutputWindow();
+            solutionService.SaveAll();
+
+            // Act
+            CommonUtility.OpenNuGetPackageManagerWithDte(VisualStudio, XunitLogger);
+            var nugetTestService = GetNuGetTestService();
+            var uiwindow = nugetTestService.GetUIWindowfromProject(project);
+            uiwindow.InstallPackageFromUI("Moq", "4.20.69");
```
zivkan: Moq version >= 4.20.0 is blocked by internal compliance tooling. Please use a different package. Ideally one that is owned by Microsoft (lower risk of the packge owner making changes that cause it to get blocked), and has little chance of having a known vulnerability reported in the future. This will reduce the maintenance efforts required in the future. If the test generated its own test package to install, there'd be almost zero risk of compliance tooling causing maintenance efforts.

CiciLi1: > Moq version >= 4.20.0 is blocked by internal compliance tooling. Please use a different package. Ideally one that is owned by Microsoft (lower risk of the packge owner making changes that cause it to get blocked), and has little chance of having a known vulnerability reported in the future. This will reduce the maintenance efforts required in the future. If the test generated its own test package to install, there'd be almost zero risk of compliance tooling causing maintenance efforts.

Ok, thanks for the reminder, I'll change the package.

CiciLi1:
Hi @zivkan, I have change package "Moq" to "log4net" and these cases have passed in corresponding build, could you please help to review and approve this PR again, thanks!

jeffkl:
Team Review: Migrate this to MSTest

nkolev92:
Some more context for @CiciLi1. 
We migrated our apex tests from XUnit to MSTest for reliability purposes. 
See example: https://github.com/NuGet/NuGet.Client/commit/6a096d0504851db695bef483801a992d65262c64. 

Please rebase the PR and attempt to use the new infrastructure. I don't expect you'd run into too many issues.

cc @heng-liu

CiciLi1:
> Some more context for @CiciLi1. We migrated our apex tests from XUnit to MSTest for reliability purposes. See example: [6a096d0](https://github.com/NuGet/NuGet.Client/commit/6a096d0504851db695bef483801a992d65262c64).
> 
> Please rebase the PR and attempt to use the new infrastructure. I don't expect you'd run into too many issues.
> 
> cc @heng-liu

Hi @nkolev92 & @heng-liu, I have updated cases to use the new infrastructure, can you please help to review it, thanks!

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetUITestCase.cs
```diff
@@ -735,5 +735,96 @@ public void UninstallDeprecatedPackageFromUI(ProjectTemplate projectTemplate, st
...
+            solutionService.CreateEmptySolution();
+            var project = solutionService.AddProject(ProjectLanguage.CSharp, ProjectTemplate.NetCoreClassLib, "TestProject");
+            VisualStudio.ClearOutputWindow();
+            solutionService.SaveAll();
+
+            // Act
+            CommonUtility.OpenNuGetPackageManagerWithDte(VisualStudio, Logger);
+            var nugetTestService = GetNuGetTestService();
+            var uiwindow = nugetTestService.GetUIWindowfromProject(project);
+            uiwindow.InstallPackageFromUI("log4net", "2.0.15");
```
nkolev92: I'm guessing this package is coming from nuget.org?

I think @zivkan's feedback is to generate a package in a local feed and use that instead of using live packages.

v-luzh: But the aim of these three test cases is testing the transitive package. Creating a package in a local feed means the normal package, correct? We thought Andy just means the "Moq" is blocked by internal compliance tooling, so we changed it to another.

zivkan: I've provided feedback in the past that I don't believe tests should use nuget.org, but I feel like this suggestion often gets ignored (across the whole team, not just regarding Apex tests), so I've kind of given up explicitly asking for that.  But I do wish we'd stop using nuget.org packages so that we minimize issues when new vulnerabilities are disclosed, new package versions published (perhaps no longer compatible with the test project TFM, and the test was installing the latest version), and other causes for test failures or compliance issues.

Many of our restore tests use generated packages, including some with dependencies:
https://github.com/NuGet/NuGet.Client/blob/5ce332fb834feea1f6c974d17456c8c9d5504e09/test/NuGet.Clients.FuncTests/NuGet.CommandLine.FuncTest/Commands/RestoreCommandTests.cs#L655-L662

Then use the helper method to create the packages in the package source directory: https://github.com/NuGet/NuGet.Client/blob/5ce332fb834feea1f6c974d17456c8c9d5504e09/test/NuGet.Clients.FuncTests/NuGet.CommandLine.FuncTest/Commands/RestoreCommandTests.cs#L668-L672

However, `SimpleTestPathContext` assumes the tests are running within the NuGet.Client repo (it searches for the repo root, then creates subdirectories under `{repo root}\.test\`). But Apex tests are compiled on one machine and run on a different machine without the repo. Therefore, some changes need to be made to the test helper methods, and `solutionService.CreateEmptySolution();` needs to be investigated to be created in the solution subdirectory that `SimpleTestPathContext` creates, rather than whatever location it currently uses (temp directory?).  Anyway, all this is one-time work, and then we'll have helper methods that allow us to generate packages for Apex tests easily.

We still need to avoid using names that match packages with security vulnerabilities (so don't call the test package `moq` or `jquery`, etc), as the compliance tool doesn't know the difference. But using obviously test package names is unlikely to

CiciLi1: > I've provided feedback in the past that I don't believe tests should use nuget.org, but I feel like this suggestion often gets ignored (across the whole team, not just regarding Apex tests), so I've kind of given up explicitly asking for that. But I do wish we'd stop using nuget.org packages so that we minimize issues when new vulnerabilities are disclosed, new package versions published (perhaps no longer compatible with the test project TFM, and the test was installing the latest version), and other causes for test failures or compliance issues.
> 
> Many of our restore tests use generated packages, including some with dependencies:
> 
> https://github.com/NuGet/NuGet.Client/blob/5ce332fb834feea1f6c974d17456c8c9d5504e09/test/NuGet.Clients.FuncTests/NuGet.CommandLine.FuncTest/Commands/RestoreCommandTests.cs#L655-L662
> 
> Then use the helper method to create the packages in the package source directory:
> 
> https://github.com/NuGet/NuGet.Client/blob/5ce332fb834feea1f6c974d17456c8c9d5504e09/test/NuGet.Clients.FuncTests/NuGet.CommandLine.FuncTest/Commands/RestoreCommandTests.cs#L668-L672
> 
> However, `SimpleTestPathContext` assumes the tests are running within the NuGet.Client repo (it searches for the repo root, then creates subdirectories under `{repo root}\.test\`). But Apex tests are compiled on one machine and run on a different machine without the repo. Therefore, some changes need to be made to the test helper methods, and `solutionService.CreateEmptySolution();` needs to be investigated to be created in the solution subdirectory that `SimpleTestPathContext` creates, rather than whatever location it currently uses (temp directory?). Anyway, all this is one-time work, and then we'll have helper methods that allow us to generate packages for Apex tests easily.
> 
> We still need to avoid using names that match packages with security vulnerabilities (so don't call the test package `moq` or `jquery`, etc), as the compliance tool doesn't know the difference. But using obviously test package names is unlikely to

Hi @zivkan & @nkolev92, thanks for your suggestion, I have change the package to local package in cases, could you please help to review it, thanks a lot!

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetUITestCase.cs
```diff
@@ -735,5 +735,99 @@ public void UninstallDeprecatedPackageFromUI(ProjectTemplate projectTemplate, st
...
+            CommonUtility.OpenNuGetPackageManagerWithDte(VisualStudio, Logger);
+            var uiwindow = nugetTestService.GetUIWindowfromProject(project);
+            uiwindow.InstallPackageFromUI(TestPackageName, TestPackageVersionV1);
+            solutionService.Build();
+
+            // Assert
+            CommonUtility.AssertPackageReferenceExists(VisualStudio, project, TestPackageName, TestPackageVersionV1, Logger);
+            uiwindow.AssertPackageNotTransitive();
+
+            // Act (Search the transitive package since it will not show at the top of package list by default)
```
nkolev92: The goal of this test is to assert that the installed tab has both transitive and direct packages correct? 

If that's the case, then I don't think is a true test, since it searches. I think that's a different scenario (transitve packages are searchable). 

I think that the verification should be that there are 2 items in the list, top being top level and bottom being a transitive one.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetUITestCase.cs
```diff
@@ -735,5 +735,99 @@ public void UninstallDeprecatedPackageFromUI(ProjectTemplate projectTemplate, st
...
+            CommonUtility.OpenNuGetPackageManagerWithDte(VisualStudio, Logger);
+            var uiwindow = nugetTestService.GetUIWindowfromProject(project);
+            uiwindow.InstallPackageFromUI(TestPackageName, TestPackageVersionV1);
+            uiwindow.InstallPackageFromUI(transitivePackageName, TestPackageVersionV1);
+            uiwindow.SearchPackageFromUI(transitivePackageName);
+            solutionService.Build();
+
+            // Assert
+            VisualStudio.AssertNoErrors();
+            uiwindow.AssertPackageNotTransitive();
```
nkolev92: Similar feedback as earlier. I think adding a search changes the test case.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetUITestCase.cs
```diff
@@ -735,5 +735,99 @@ public void UninstallDeprecatedPackageFromUI(ProjectTemplate projectTemplate, st
...
+
+            NuGetApexTestService nugetTestService = GetNuGetTestService();
+            var solutionService = VisualStudio.Get<SolutionService>();
+            solutionService.CreateEmptySolution("TestSolution", _pathContext.SolutionRoot);
+            var project = solutionService.AddProject(ProjectLanguage.CSharp, ProjectTemplate.NetCoreConsoleApp, "TestProject");
+            VisualStudio.ClearOutputWindow();
+            solutionService.SaveAll();
+
+            // Act
+            CommonUtility.OpenNuGetPackageManagerWithDte(VisualStudio, Logger);
```
nkolev92, line 789: Act should normally be a single line. 

In this case, `Act` should be installation of the transitive package.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetUITestCase.cs
```diff
@@ -735,5 +735,99 @@ public void UninstallDeprecatedPackageFromUI(ProjectTemplate projectTemplate, st
...
+            var uiwindow = nugetTestService.GetUIWindowfromProject(project);
+            uiwindow.InstallPackageFromUI(TestPackageName, TestPackageVersionV1);
+            uiwindow.InstallPackageFromUI(transitivePackageName, TestPackageVersionV1);
```
nkolev92, line 799: Before we do this, we should assert the preconditions. We should assert that this package exists in the list as a transitive package. 
As the test case is written, if you don't add any dependencies among the created packages, it would still pass.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetUITestCase.cs
```diff
@@ -735,5 +735,99 @@ public void UninstallDeprecatedPackageFromUI(ProjectTemplate projectTemplate, st
...
+
+            // Assert
+            VisualStudio.AssertNoErrors();
+            uiwindow.AssertPackageNotTransitive();
+            CommonUtility.AssertPackageReferenceExists(VisualStudio, project, transitivePackageName, TestPackageVersionV1, Logger);
+        }
+
+        [TestMethod]
+        [Timeout(DefaultTimeout)]
+        public async Task UninstallTransitivePackageFromUI()
```
nkolev92: I'm not sure this does anything for us. 

1. Install package with transitive dep
2. Install that transitive dep as top level

The 2nd package being a transitive dependency at some point doesn't really affect how uninstall works, at least not in a .NET SDK based projects

nkolev92 (changes requested):
Thanks for updating the test cases @CiciLi1!
Apologies for the slow turnaround.

I've left some feedback.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetUITestCase.cs
v-luzh: Yes, you are correct. We should not test searching transitive package when trying to test transitive package installation. We have updated the case by deleting the searching steps and adding an assert to verify the top being top-level and bottom being a transitive for the case "InstallTransitivePackageFromUI()". And we keep the searching transitive package scenario and create a new test case as " SearchTransitivePackageInInstalledTabFromUI()".

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetUITestCase.cs
v-luzh: OK, thanks for your reminder. We will keep the key test point of every test case as a single Act line.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetUITestCase.cs
v-luzh: OK, thanks for your reminder. We will keep the key test point of every test case as a single Act line.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetUITestCase.cs
v-luzh: We have updated the case by adding an assert to verify the top being top-level and bottom being a transitive before installing a transitive package.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetUITestCase.cs
v-luzh: No, we are talking about different scenarios. Sorry about the confusion. In this case, firstly we installed a top-level package which having a dependency package. Then we installed the transitive package (that will change it into a top-level package). The purpose of this case is testing uninstallation of the top-level package (which was a transitive package before) will make it back to transitive package from top-level package.

v-luzh:
Hi @nkolev92, we have updated the test cases according to your latest suggestion. Please help to review them again. Thanks very much.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/Apex/NuGetUIProjectTestExtension.cs
```diff
@@ -62,6 +62,23 @@ public void AssertInstalledPackageNotDeprecated()
             DeprecatedPackageResult.Should().BeFalse();
         }
 
+        public void AssertPackageTransitive(string testPackageName, string transitivePackageName)
```
nkolev92: naming: This is assert more than just the transitive package.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/Apex/NuGetUIProjectTestExtension.cs
```diff
@@ -62,6 +62,23 @@ public void AssertInstalledPackageNotDeprecated()
...
+        {
+            var transitivePackageResult = _uiproject.VerifyTopLevelAndTransitivePackageOnInstalledTab(testPackageName, transitivePackageName);
+            transitivePackageResult.Should().BeTrue();
+        }
+
+        public void AssertPackageNotTransitive(string testPackageName, string transitivePackageName)
```
nkolev92: This method asserts 2 things, so if it fails, it's not obvious which one failed. 

The best approach would be to get the list inside the main test method, and then assert that the first value is top level and the 2nd is transitive. That would give you a better error message in case of a failure.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetUITestCase.cs
```diff
@@ -735,5 +735,106 @@ public void UninstallDeprecatedPackageFromUI(ProjectTemplate projectTemplate, st
...
+
+            // Act
+            uiwindow.InstallPackageFromUI(transitivePackageName, TestPackageVersionV1);
+            solutionService.Build();
+
+            // Assert
+            VisualStudio.AssertNoErrors();
+            CommonUtility.AssertPackageReferenceExists(VisualStudio, project, transitivePackageName, TestPackageVersionV1, Logger);
+            uiwindow.AssertPackageNotTransitive(TestPackageName, transitivePackageName);
+        }
```
nkolev92, line 806: This test looks great now, thanks for addressing the feedback.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetUITestCase.cs
nkolev92: Ah, I get it now, thanks for the explanation. 

Naming wise, I'm not really sure the current test name tells the whole story. Maybe we need to workshop that a little bit.

nkolev92 (commented):
Thanks for addressing the feedback! Looks a lot better!

There's some guidelines in here https://github.com/NuGet/NuGet.Client/blob/dev/docs/coding-guidelines.md#unit-test-structure, that should be helpful with some of the feedback.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/Apex/NuGetUIProjectTestExtension.cs
CiciLi1: We updated the naming when separating this method according to the next comment.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/Apex/NuGetUIProjectTestExtension.cs
CiciLi1: Thanks for the suggestion, we have split this method by asserting that the first value is the top-level value and the second value is the pass-through value.

CiciLi1:
Hi @nkolev92, we have updated test cases according to your latest suggestion. could you please help to review them. Thanks a lot.😊

test/NuGet.Tests.Apex/NuGet.PackageManagement.UI.TestContract/ApexTestUIProject.cs
```diff
@@ -95,6 +95,12 @@ public bool VerifyDeprecatedPackageOnTopOfInstalledTab()
             return result?.IsPackageDeprecated == true;
         }
 
+        public List<PackageItemViewModel> PackageItemOnInstalledTab()
```
nkolev92: naming: PackageItemsOnInstalledTab or even better, GetPackageItemsOnInstalledTab

test/NuGet.Tests.Apex/NuGet.Tests.Apex/Apex/NuGetUIProjectTestExtension.cs
```diff
@@ -62,6 +64,33 @@ public void AssertInstalledPackageNotDeprecated()
...
+        public void AssertTopLevelPackageExistsInInstalledTab(string packageName)
+        {
+            var topPackage = _uiproject.PackageItemOnInstalledTab().FirstOrDefault();
+            bool topPackageResult = topPackage.PackageLevel.Equals(PackageLevel.TopLevel) && topPackage.Id.Equals(packageName);
+            topPackageResult.Should().BeTrue();
+        }
+
+        public void AssertTransitivePackageExistsInInstalledTab(string transitivePackageName)
+        {
+            var bottomPackage = _uiproject.PackageItemOnInstalledTab().Last();
```
nkolev92: This method is making a lot of assumptions that are not communicated in the name. 
It assumes that the last package is transitive.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/Apex/NuGetUIProjectTestExtension.cs
```diff
@@ -62,6 +64,33 @@ public void AssertInstalledPackageNotDeprecated()
...
+            bool bottomPackageResult = bottomPackage.PackageLevel.Equals(PackageLevel.Transitive) && bottomPackage.Id.Equals(transitivePackageName);
+            bottomPackageResult.Should().BeTrue();
```
nkolev92: This doesn't really help us improve the error message, as it'll say, "expected true got false".

The best approach is:
```cs
bottomPackage.PackageLevel.Should().Be(PackageLevel.Transitive);
bottomPackage.Id.Should().Be(transitivePackageName)
```

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetUITestCase.cs
```diff
@@ -735,5 +735,115 @@ public void UninstallDeprecatedPackageFromUI(ProjectTemplate projectTemplate, st
...
+            CommonUtility.AssertPackageReferenceExists(VisualStudio, project, TestPackageName, TestPackageVersionV1, Logger);
+            uiwindow.AssertTopLevelPackageExistsInInstalledTab(TestPackageName);
+            uiwindow.AssertTransitivePackageExistsInInstalledTab(transitivePackageName);
+
+            // Act
+            uiwindow.SearchPackageFromUI(transitivePackageName);
+
+            // Assert
+            VisualStudio.AssertNoErrors();
+            uiwindow.AssertSearchedTransitivePackageExistsInInstalledTab(transitivePackageName);
```
nkolev92: It'd replace this with the individual statements: 

```cs
var bottomPackage = _uiproject.PackageItemOnInstalledTab().Last();
bottomPackage.PackageLevel.Should().Be(PackageLevel.Transitive);
bottomPackage.Id.Should().Be(transitivePackageName)
```

Then we have the fewest amount of questions about what's being assert. 

If you'd like you can also have a method that asserts name/package type. 

Like for example:

```cs
static void AssertPackageNameAndType( Package(not sure the type) package, string id, PackageLevel packageLevel)
{
    package.PackageLevel.Should().Be(packageLevel);
    package.Id.Should().Be(id);
}
```

The idea behind this is to reduce the number of methods that do a lot of things, and thus need really long names.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/Apex/NuGetUIProjectTestExtension.cs
```diff
@@ -62,6 +64,33 @@ public void AssertInstalledPackageNotDeprecated()
             DeprecatedPackageResult.Should().BeFalse();
         }
 
+        public void AssertTopLevelPackageExistsInInstalledTab(string packageName)
```
nkolev92: I think ideally all these methods are removed and we just have the code in the methods themselves.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/Apex/NuGetUIProjectTestExtension.cs
```diff
@@ -62,6 +64,33 @@ public void AssertInstalledPackageNotDeprecated()
...
+        {
+            var topPackage = _uiproject.PackageItemOnInstalledTab().FirstOrDefault();
+            bool topPackageResult = topPackage.PackageLevel.Equals(PackageLevel.TopLevel) && topPackage.Id.Equals(packageName);
+            topPackageResult.Should().BeTrue();
```
nkolev92: Similar feedback, this would lead to an error message saying got false expected true, which is not helpful

test/NuGet.Tests.Apex/NuGet.Tests.Apex/Apex/NuGetUIProjectTestExtension.cs
```diff
@@ -62,6 +64,33 @@ public void AssertInstalledPackageNotDeprecated()
...
+            var bottomPackage = _uiproject.PackageItemOnInstalledTab().Last();
+            bool bottomPackageResult = bottomPackage.PackageLevel.Equals(PackageLevel.Transitive) && bottomPackage.Id.Equals(transitivePackageName);
+            bottomPackageResult.Should().BeTrue();
+        }
+
+        public void AssertTransitivePackageNotExistsInInstalledTab(string transitivePackageName)
+        {
+            var bottomPackage = _uiproject.PackageItemOnInstalledTab().Last();
+            bool bottomPackageResult = bottomPackage.PackageLevel.Equals(PackageLevel.Transitive) && bottomPackage.Id.Equals(transitivePackageName);
+            bottomPackageResult.Should().BeFalse();
```
nkolev92: Similar feedback, this would lead to an error message saying got true expected false, which is not helpful

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetUITestCase.cs
```diff
@@ -735,5 +735,115 @@ public void UninstallDeprecatedPackageFromUI(ProjectTemplate projectTemplate, st
...
+
+            // Assert
+            VisualStudio.AssertNoErrors();
+            CommonUtility.AssertPackageReferenceExists(VisualStudio, project, transitivePackageName, TestPackageVersionV1, Logger);
+            uiwindow.AssertTransitivePackageNotExistsInInstalledTab(transitivePackageName);
+        }
+
+        [TestMethod]
+        [Timeout(DefaultTimeout)]
+        public async Task UninstallTopLevelToTransitivePackageFromUI()
```
nkolev92: I mentioned earlier that we need a better name to explain the test case, since this does a lot. 

Uninstall_WithMultiplePackagesThatDependOnEachOther_PackageGoesFromDirectToTransitive


this is wordy, but it's important that test name describe the test case.

v-luzh:
Hi  @nkolev92, thanks for the new suggestions. We have updated the test cases base on your latest suggestions and the corresponding private build is Pass. Please help to review them again. Thanks a lot.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/Apex/NuGetUIProjectTestExtension.cs
```diff
@@ -62,6 +63,12 @@ public void AssertInstalledPackageNotDeprecated()
             DeprecatedPackageResult.Should().BeFalse();
         }
 
+        public void AssertPackageNameAndType(string packageId, PackageLevel packageLevel)
+        {
+            var package = _uiproject.GetPackageItemOnInstalledTabById(packageId);
+            package.PackageLevel.Should().Be(packageLevel);
```
nkolev92, line 75: If the above method returns null, this would throw a null ref.

test/NuGet.Tests.Apex/NuGet.PackageManagement.UI.TestContract/ApexTestUIProject.cs
```diff
@@ -95,6 +95,11 @@ public bool VerifyDeprecatedPackageOnTopOfInstalledTab()
             return result?.IsPackageDeprecated == true;
         }
 
+        public PackageItemViewModel GetPackageItemOnInstalledTabById(string packageId)
+        {
+            return _packageManagerControl.PackageList.PackageItems.Where(x => x.Id == packageId).FirstOrDefault();
```
nkolev92: Probably best to have this method return the package list so you can account for nulls.

nkolev92 (commented):
Some null safety comments, the rest looks great. 

Thank you for being patient with reviews and addressing all the comments.

test/NuGet.Tests.Apex/NuGet.PackageManagement.UI.TestContract/ApexTestUIProject.cs
CiciLi1: Done

test/NuGet.Tests.Apex/NuGet.Tests.Apex/Apex/NuGetUIProjectTestExtension.cs
CiciLi1: Hi @nkolev92 , thanks for your review and suggestion, we added two helper validations for null refs to this Assert method, the first to avoid an empty package list on the "Installed" tab, and the second to avoid a package list that doesn't contain the currently validated package. Please help to review them again. Thanks a lot.

nkolev92 (approved):

PR #5427, opened 2023-09-27
Fix "Sequence contains no matching element" when listing outdated packages and no latest version available

### LINKED ISSUE #12256: "error: Sequence contains no matching element" when listing outdated packages
### NuGet Product Used

dotnet.exe

### Product Version

7.0.100, 6.0.300, 5.0.408, 3.1.425 (all I tested)

### Worked before?

no

### Impact

### Repro Steps & Context

1. `dotnet new console -o repro`
2. `dotnet add repro package AngleSharp --version 1.0.0-alpha-844` (note: this version was unlisted, the highest still-published version is 0.17.1)
3. (optional) `dotnet add repro package AngleSharp.Css --version 0.16.0` (note: highest released version is 0.16.4)
4. `dotnet list repro package --outdated --highest-minor` (or `--highest-patch`; the equivalent `dotnet nuget` command also fails)

What I expected:

I would expect the command to complete successfully. At a minimum it should still list the other package that's outdated (AngleSharp.Css 0.16.0 -> 0.16.4) if step 3 was run. It would also be nice to have a warning [like the one on nuget.org](https://www.nuget.org/packages/AngleSharp/1.0.0-alpha-844) about the AngleSharp version being unlisted, with a mention of the highest published version.

What I got:

```
The following sources were used:
   https://api.nuget.org/v3/index.json

error: Sequence contains no matching element


Usage: NuGet.CommandLine.XPlat.dll package list [arguments] [options]
[...]
```

---

This made it more difficult to identify packages that needed I to update (note that `dotnet list repro package --outdated` also does not list AngleSharp) on a project with several dependencies. I hope it can be fixed to save others the same difficulty. Thanks for considering this issue.

### Verbose Logs

```shell
The following sources were used:
   https://api.nuget.org/v3/index.json

  CACHE https://api.nuget.org/v3/registration5-gz-semver2/anglesharp/index.json
  CACHE https://api.nuget.org/v3/registration5-gz-semver2/anglesharp/page/0.2.0/0.13.0-alpha-744.json
  CACHE https://api.nuget.org/v3/registration5-gz-semver2/anglesharp/page/0.13.0-alpha-745/0.16.1-alpha-167.json
  CACHE https://api.nuget.org/v3/registration5-gz-semver2/anglesharp/page/0.16.1-alpha-168/1.0.0-alpha-844.json
  CACHE https://api.nuget.org/v3/registration5-gz-semver2/anglesharp.css/index.json
  CACHE https://api.nuget.org/v3/registration5-gz-semver2/anglesharp/index.json
  CACHE https://api.nuget.org/v3/registration5-gz-semver2/anglesharp/page/0.2.0/0.13.0-alpha-744.json
  CACHE https://api.nuget.org/v3/registration5-gz-semver2/anglesharp/page/0.13.0-alpha-745/0.16.1-alpha-167.json
  CACHE https://api.nuget.org/v3/registration5-gz-semver2/anglesharp/page/0.16.1-alpha-168/1.0.0-alpha-844.json
  CACHE https://api.nuget.org/v3/registration5-gz-semver2/anglesharp.css/index.json
error: Sequence contains no matching element
trace: System.AggregateException: One or more errors occurred. (Sequence contains no matching element)
trace:  ---> System.InvalidOperationException: Sequence contains no matching element
trace:    at System.Linq.ThrowHelper.ThrowNoMatchException()
trace:    at System.Linq.Enumerable.First[TSource](IEnumerable`1 source, Func`2 predicate)
trace:    at NuGet.CommandLine.XPlat.ListPackageCommandRunner.GetVersionsFromDictAsync(IEnumerable`1 packages, Dictionary`2 packagesVersionsDict, ListPackageArgs listPackageArgs)
trace:    at NuGet.CommandLine.XPlat.ListPackageCommandRunner.AddLatestVersionsAsync(IEnumerable`1 packages, ListPackageArgs listPackageArgs)
trace:    at NuGet.CommandLine.XPlat.ListPackageCommandRunner.ExecuteCommandAsync(ListPackageArgs listPackageArgs)
trace:    at NuGet.CommandLine.XPlat.ListPackageCommand.<>c__DisplayClass0_1.<<Register>b__1>d.MoveNext()
trace:    --- End of inner exception stack trace ---
trace:    at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
trace:    at Microsoft.Extensions.CommandLineUtils.CommandLineApplication.Execute(String[] args)
trace:    at NuGet.CommandLine.XPlat.Program.MainInternal(String[] args, CommandOutputLogger log)


Usage: NuGet.CommandLine.XPlat.dll package list [arguments] [options]

Arguments:
  <PROJECT | SOLUTION>  A path to a project, solution file or directory.

Options:
  -h|--help               Show help information
  --force-english-output  Forces the application to run using an invariant, English-based culture.
  --framework             Specifies the target framework for which the packages will be listed.
  --deprecated            Displays only the packages marked deprecated by the authors.
  --outdated              Displays only the packages that need updates with the latest version from the sources.
  --vulnerable            Displays only the packages flagged as vulnerable.
  --include-transitive    Includes transitive packages too in the result.
  --include-prerelease    Considers prerelease versions when looking for latest. Works only with `--outdated`.
  --highest-patch         Considers only the versions with matching minor and major. Works only with `--outdated`.
  --highest-minor         Considers only the versions with matching major. Works only with `--outdated`.
  --source                Sources to lookup for latest versions. Works only with `--outdated`.
  --config                A path to a config file to specify sources. Works only with `--outdated`.
  --interactive           Allow the command to block and require manual action for operations like authentication.
  -v|--verbosity          Set the verbosity level of the command. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
```

martinrrm:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12256

## Description

This PR fixes a null reference when users try to list outdated packages. This happens when the user adds the options `--highest-minor` or `--highest-patch` to the command `dotnet list package --outdated` and there is no version available with these constraints.

Also, this was breaking the command and not listing the other packages that had a latest version available.

For example, if the user has installed a prerelease version and use the command `dotnet list package --outdated --highest-minor` we won't find a version available, not and since the user is not using `--include-prerelease` the installed version won't be listed as an available version. 

#### CSPROJ
image

#### Command output 
image

## Fix
With this fix I'm just handling the null reference, that there is no `LatestPackageMetadata` and `UpdateLevel = UpdateLevel.NoUpdate` this will create an entry on the output table saying that there is no available version in the source.

### Command output (using the same csproj from previous example)
image

#### Example with multiple table entries
image

### Other considerations
Another approach I tried to fix this was to not add the outdated package to the list and show the regular message that there are no updates available and not show the table, But given that there is already a test for version not found and does create the table I decided to do it like this.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/ListPackageCommandRunner.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/ListPackageCommandRunner.cs
@@ -414,65 +414,67 @@ namespace NuGet.CommandLine.XPlat
             {
                 foreach (var topLevelPackage in frameworkPackages.TopLevelPackages)
                 {
-                    var matchingPackage = packageMetadata.Where(p => p.Key.Equals(topLevelPackage.Name, StringComparison.OrdinalIgnoreCase)).FirstOrDefault();
-
-                    // Get latest metadata *only* if this is a report requiring "outdated" metadata
-                    if (listPackageArgs.ReportType == ReportType.Outdated && matchingPackage.Value.Count > 0)
+                    if (packageMetadata.TryGetValue(topLevelPackage.Name, out List<IPackageSearchMetadata> matchingPackage))
                     {
-                        var latestVersion = matchingPackage.Value.Where(newVersion => MeetsConstraints(newVersion.Identity.Version, topLevelPackage, listPackageArgs)).Max(i => i.Identity.Version);
-
-                        if (latestVersion is not null)
+                        // Get latest metadata *only* if this is a report requiring "outdated" metadata
+                        if (listPackageArgs.ReportType == ReportType.Outdated && matchingPackage.Count > 0)
                         {
-                            topLevelPackage.LatestPackageMetadata = matchingPackage.Value.First(p => p.Identity.Version == latestVersion);
-                            topLevelPackage.UpdateLevel = GetUpdateLevel(topLevelPackage.ResolvedPackageMetadata.Identity.Version, topLevelPackage.LatestPackageMetadata.Identity.Version);
+                            var latestVersion = matchingPackage.Where(newVersion => MeetsConstraints(newVersion.Identity.Version, topLevelPackage, listPackageArgs)).Max(i => i.Identity.Version);
+
+                            if (latestVersion is not null)
+                            {
+                                topLevelPackage.LatestPackageMetadata = matchingPackage.First(p => p.Identity.Version == latestVersion);
+                                topLevelPackage.UpdateLevel = GetUpdateLevel(topLevelPackage.ResolvedPackageMetadata.Identity.Version, topLevelPackage.LatestPackageMetadata.Identity.Version);
+                            }
+                            else // no latest version available with the given constraints
+                            {
+                                topLevelPackage.LatestPackageMetadata = null;
+                                topLevelPackage.UpdateLevel = UpdateLevel.NoUpdate;
+                            }
                         }
-                        else // no latest version available with the given constraints
+
+                        var matchingPackagesWithDeprecationMetadata = await Task.WhenAll(
+                            matchingPackage.Select(async v => new { SearchMetadata = v, DeprecationMetadata = await v.GetDeprecationMetadataAsync() }));
+
+                        // Update resolved version with additional metadata information returned by the server.
+                        var resolvedVersionFromServer = matchingPackagesWithDeprecationMetadata
+                            .Where(v => v.SearchMetadata.Identity.Version == topLevelPackage.ResolvedPackageMetadata.Identity.Version &&
+                                    (v.DeprecationMetadata != null || v.SearchMetadata?.Vulnerabilities != null))
+                            .FirstOrDefault();
+
+                        if (resolvedVersionFromServer != null)
                         {
-                            topLevelPackage.LatestPackageMetadata = null;
-                            topLevelPackage.UpdateLevel = UpdateLevel.NoUpdate;
+                            topLevelPackage.ResolvedPackageMetadata = resolvedVersionFromServer.SearchMetadata;
                         }
-                    }
-
-                    var matchingPackagesWithDeprecationMetadata = await Task.WhenAll(
-                        matchingPackage.Value.Select(async v => new { SearchMetadata = v, DeprecationMetadata = await v.GetDeprecationMetadataAsync() }));
-
-                    // Update resolved version with additional metadata information returned by the server.
-                    var resolvedVersionFromServer = matchingPackagesWithDeprecationMetadata
-                        .Where(v => v.SearchMetadata.Identity.Version == topLevelPackage.ResolvedPackageMetadata.Identity.Version &&
-                                (v.DeprecationMetadata != null || v.SearchMetadata?.Vulnerabilities != null))
-                        .FirstOrDefault();
-
-                    if (resolvedVersionFromServer != null)
-                    {
-                        topLevelPackage.ResolvedPackageMetadata = resolvedVersionFromServer.SearchMetadata;
                     }
                 }
 
                 foreach (var transitivePackage in frameworkPackages.TransitivePackages)
                 {
-                    var matchingPackage = packageMetadata.Where(p => p.Key.Equals(transitivePackage.Name, StringComparison.OrdinalIgnoreCase)).First();
-
-                    // Get latest metadata *only* if this is a report requiring "outdated" metadata
-                    if (listPackageArgs.ReportType == ReportType.Outdated && matchingPackage.Value.Count > 0)
+                    if (packageMetadata.TryGetValue(transitivePackage.Name, out List<IPackageSearchMetadata> matchingPackage))
                     {
-                        var latestVersion = matchingPackage.Value.Where(newVersion => MeetsConstraints(newVersion.Identity.Version, transitivePackage, listPackageArgs)).Max(i => i.Identity.Version);
+                        // Get latest metadata *only* if this is a report requiring "outdated" metadata
+                        if (listPackageArgs.ReportType == ReportType.Outdated && matchingPackage.Count > 0)
+                        {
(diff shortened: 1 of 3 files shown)


zivkan (dismissed):
The title of this PR is misleading. "sequence does not contain any matching element" (`InvalidOperationException`?) is not the same as a `NullReferenceException`.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/ListPackageCommandRunner.cs
```diff
@@ -414,15 +414,23 @@ private void PopulateSourceRepositoryCache(ListPackageArgs listPackageArgs)
             {
                 foreach (var topLevelPackage in frameworkPackages.TopLevelPackages)
                 {
-                    var matchingPackage = packageMetadata.Where(p => p.Key.Equals(topLevelPackage.Name, StringComparison.OrdinalIgnoreCase)).First();
+                    var matchingPackage = packageMetadata.Where(p => p.Key.Equals(topLevelPackage.Name, StringComparison.OrdinalIgnoreCase)).FirstOrDefault();
```
dtivel: Off topic for this PR, but I wonder if [this dictionary can simply be created with this comparer](https://github.com/NuGet/NuGet.Client/pull/5427/files#diff-60e044f4a652f6a16cc3c89aedc2cbcf13361c672c237a95268a3e95905340e5R278), like so:

```C#
var packageMetadataById = new Dictionary<string, List<IPackageSearchMetadata>>(capacity: allPackages.Count);
```

Then, you could simply do a dictionary lookup:

```
if (packageMetadata.TryGetValue(topLevelPackage.Name, out List<string>? versions))
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/ListPackageCommandRunner.cs
```diff
@@ -414,15 +414,23 @@ private void PopulateSourceRepositoryCache(ListPackageArgs listPackageArgs)
...
 
                     // Get latest metadata *only* if this is a report requiring "outdated" metadata
                     if (listPackageArgs.ReportType == ReportType.Outdated && matchingPackage.Value.Count > 0)
```
dtivel: There seems to be a bug here.  The `FirstOrDefault()` call will return the first matching element or `default(T)`.  Here, `T` is `System.Collections.Generic.KeyValuePair<string, List<string>>`, which is a struct.  If there is no matching element, `FirstOrDefault()` will return `default(KeyValuePair<string, List<string>>)` and the `Key` and `Value` properties for this object will both return `null`.  That means the subsequent call to `matchingPackage.Value.Count` would throw `NullReferenceException`.

To see what I mean, try running this code on https://try.dot.net:

```C#
using System;
using System.Collections.Generic;
using System.Linq;

public class Program
{
  public static void Main()
  {
    Dictionary<string, List<string>> dictionary = new Dictionary<string, List<string>>();
    KeyValuePair<string, List<string>> keyValuePair = dictionary.FirstOrDefault();

    Console.WriteLine($"Key == {keyValuePair.Key ?? "(null)"}");
    Console.WriteLine($"Value == {(keyValuePair.Value is null ? "(null)" : keyValuePair.Value.Count.ToString())}");
  }
}
```

jeffkl: Yeah let's make sure there's a unit test for this scenario

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/ListPackageCommandRunner.cs
martinrrm: I discovered that this dictionary is initialized that way in https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/ListPackageCommandRunner.cs#L278. I pushed a new commit; do you think it addresses your suggestion?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/ListPackageCommandRunner.cs
martinrrm: I've not been able to test this, do you have any suggestions on how can I test it? 

Since this function is not public and static, I cannot create a unit test and it's been hard to create a more complex test with mocked data.

donnie-msft: > I've not been able to test this, do you have any suggestions on how can I test it?

The goal I understand is to write a test such that `matchingPackage` becomes `null`. Normally, that'd be as simple as creating a mocked packageMetadata which will have 0 results, and therefore your code will return `null`. However, I'm not sure with all these test "helper" contexts exactly how that works. Maybe you already know how metadata is being pulled from the test, but if not, I'd start there.

martinrrm: Added the test!

jeffkl (approved):

zivkan (approved):

Saibamen:
I'm still having this issue:
When I wanted to uninstall packages:

image

Windows 10, latest updates
Visual Studio 17.8.5

nkolev92:
@Saibamen That's a different scenario, a different command. 
Can you please file a new issue for that.

PR #5434, opened 2023-09-28
Added .NET 8 constants to NuGet.Frameworks

### LINKED ISSUE #12910: Add .NET 8 constants to NuGet.Frameworks
### NuGet Product(s) Involved

NuGet SDK

### The Elevator Pitch

Add the new .NET 8 framework literals and update the parsing logic in NuGet.Frameworks ahead of the .NET 8 release

_No response_

advay26:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12910

## Description

This change adds the new .NET 8 framework literals and updates the parsing logic in NuGet.Frameworks ahead of the .NET 8 release. It also adds tests covering the changes.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Core/NuGet.Frameworks/FrameworkConstants.cs
```diff
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
```
advay26, line 5: There was an alert saying this using directive wasn't necessary, so I removed it as part of this change

nkolev92 (dismissed):

nkolev92:
PR naming: this makes it seems like .NET 8 is not support, but it's really about creating a NuGetFramework constant for net8.0

kartheekp-ms:
I think we should add tests [here](https://github.com/NuGet/NuGet.Client/blob/dev/test/NuGet.Core.Tests/NuGet.Frameworks.Test/NuGetFrameworkTests.cs) to ensure that `NuGetFramework.Parse("net8.0")` succeeds. I think that is the API which you would like to invoke, correct?

src/NuGet.Core/NuGet.Frameworks/NuGetFrameworkFactory.cs
```diff
@@ -650,6 +650,12 @@ private static bool TryParseCommonFramework(string frameworkString, [NotNullWhen
                 case "net70":
                     framework = FrameworkConstants.CommonFrameworks.Net70;
                     break;
+                case "netcoreapp8.0":
+                case "netcoreapp80":
+                case "net8.0":
+                case "net80":
+                    framework = FrameworkConstants.CommonFrameworks.Net80;
+                    break;
```
dtivel, line 658: When I do something that has been done before --- like adding a new TFM ---, I review how others did it in the past.  Not to copy, but to see if there's something I may have missed.

[This](https://github.com/NuGet/NuGet.Client/commit/820bfe9d5e5e4e7bf0b6310c285db7b07aff3087#diff-9608c8cf5369cc492f2ed8d56c58057607817e6a6e1eb776854f24effc46e4b1R434) looks like a good place to add a test for `net8.0`.  Bonus points for adding `net7.0` since its PR overlooked it too.

Even without this change, restore works for `net8.0`.  The gist of this change is that we now recognize `net8.0` (and related TFM's) and return a static instance.  A good rule of thumb is that if it's worth making a product change, it's probably also worth adding a test.

Don't forget to update the PR description to indicate that tests were added.

dtivel (changes requested):

src/NuGet.Core/NuGet.Frameworks/NuGetFrameworkFactory.cs
advay26: That makes sense. I've now added tests covering the changes, and updated the PR description too

dtivel (approved):

jeffkl (approved):

PR #5440, opened 2023-09-29
Add option to enable floating versions in CPM

### LINKED ISSUE #10432: [DCR]: Allow floating versions with Central Package Management (CPM)
## Details about Problem

We use CPVM to manage our dependencies. Unfortunately, the last update presented us with a big problem. With the introduction of NU1011, it is no longer possible for us to run a resolution strategy that selects the highest possible version. This is especially relevant for our in-house dependencies, as we distribute regular patches and this can no longer be done automatically.

## Suggested things

Our suggestion is to implement a switch that deactivates the validation of NU1011. We understood the problem that led to the implementation of NU1011, but can rule out the problem for us.

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Home/issues/10432

## Description

By default, central package management (CPM) does not allow users to specify floating version (i.e. `1.*`).  CPM is considered an enterprise-level feature and since floating versions can introduce non-deterministic restores we want users to not be able to get situations where restores in different environments could not be identical.

However, the community has left a good amount of feedback that such a feature should be allowed with more details available in https://github.com/NuGet/Home/issues/10432.  This change allows a user to enable floating versions by setting an MSBuild property which remains off by default.

The only place the MSBuild property will be documented is at https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1011 where users will be encouraged to use non floating versions and educated on the potential issues they could encounter.

## PR Checklist

- [x] PR has a linked issue.
- [ ] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [x] Documentation PR or issue filled: https://github.com/NuGet/Home/issues/12919
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.SolutionRestoreManager/VsSolutionRestoreService.cs
+++ b/src/NuGet.Clients/NuGet.SolutionRestoreManager/VsSolutionRestoreService.cs
@@ -348,6 +348,7 @@ namespace NuGet.SolutionRestoreManager
                     CacheFilePath = NoOpRestoreUtilities.GetProjectCacheFilePath(cacheRoot: outputPath),
                     RestoreLockProperties = VSNominationUtilities.GetRestoreLockProperties(TargetFrameworks),
                     CentralPackageVersionsEnabled = cpvmEnabled,
+                    CentralPackageFloatingVersionsEnabled = VSNominationUtilities.IsCentralPackageFloatingVersionsEnabled(TargetFrameworks),
                     CentralPackageVersionOverrideDisabled = VSNominationUtilities.IsCentralPackageVersionOverrideDisabled(TargetFrameworks),
                     CentralPackageTransitivePinningEnabled = VSNominationUtilities.IsCentralPackageTransitivePinningEnabled(TargetFrameworks),
                     RestoreAuditProperties = VSNominationUtilities.GetRestoreAuditProperties(TargetFrameworks),
--- a/src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
+++ b/src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
@@ -808,7 +808,7 @@ namespace NuGet.Build.Tasks.Console
 
             ProjectStyle? projectStyleOrNull = BuildTasksUtility.GetProjectRestoreStyleFromProjectProperty(project.GetProperty("RestoreProjectStyle"));
 
-            (bool isCentralPackageManagementEnabled, bool isCentralPackageVersionOverrideDisabled, bool isCentralPackageTransitivePinningEnabled, bool isCentralPackageFloatingVersionsEnabled) = GetCentralPackageManagementSettings(project, projectStyleOrNull);
+            (bool isCentralPackageManagementEnabled, bool isCentralPackageVersionOverrideDisabled, bool isCentralPackageTransitivePinningEnabled, bool isCentralPackageFloatingVersionsEnabled) = MSBuildRestoreUtility.GetCentralPackageManagementSettings(project, projectStyleOrNull);
 
             RestoreAuditProperties auditProperties = MSBuildRestoreUtility.GetRestoreAuditProperties(project);
 
@@ -1024,22 +1024,6 @@ namespace NuGet.Build.Tasks.Console
 
                 return null;
             }
-        }
-
-        /// <summary>
-        /// Determines the current settings for central package management for the specified project.
-        /// </summary>
-        /// <param name="project">The <see cref="IMSBuildProject" /> to get the central package management settings for.</param>
-        /// <param name="projectStyle">The <see cref="ProjectStyle?" /> of the specified project.  Specify <see langword="null" /> when the project does not define a restore style.</param>
-        /// <returns>A <see cref="Tuple{T1, T2, T3, T4}" /> containing values indicating whether or not central package management is enabled, if the ability to override a package version is disabled, if the ability to transitively pin a package version is enabled, and if floating versions are enabled.</returns>
-        internal static (bool IsEnabled, bool IsVersionOverrideDisabled, bool IsCentralPackageTransitivePinningEnabled, bool IsFloatingVersionsEnabled) GetCentralPackageManagementSettings(IMSBuildProject project, ProjectStyle? projectStyle)
-        {
-            if (!projectStyle.HasValue || (projectStyle.Value == ProjectStyle.PackageReference))
-            {
-                return (project.IsPropertyTrue("_CentralPackageVersionsEnabled"), project.IsPropertyFalse("CentralPackageVersionOverrideEnabled"), project.IsPropertyTrue("CentralPackageTransitivePinningEnabled"), project.IsPropertyTrue("CentralPackageFloatingVersionsEnabled"));
-            }
-
-            return (false, false, false, false);
         }
 
         /// <summary>
--- a/src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
+++ b/src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
@@ -808,7 +808,7 @@ namespace NuGet.Build.Tasks.Console
 
             ProjectStyle? projectStyleOrNull = BuildTasksUtility.GetProjectRestoreStyleFromProjectProperty(project.GetProperty("RestoreProjectStyle"));
 
-            (bool isCentralPackageManagementEnabled, bool isCentralPackageVersionOverrideDisabled, bool isCentralPackageTransitivePinningEnabled) = GetCentralPackageManagementSettings(project, projectStyleOrNull);
+            (bool isCentralPackageManagementEnabled, bool isCentralPackageVersionOverrideDisabled, bool isCentralPackageTransitivePinningEnabled, bool isCentralPackageFloatingVersionsEnabled) = GetCentralPackageManagementSettings(project, projectStyleOrNull);
 
             RestoreAuditProperties auditProperties = MSBuildRestoreUtility.GetRestoreAuditProperties(project);
 
@@ -854,6 +854,7 @@ namespace NuGet.Build.Tasks.Console
                     SkipContentFileWrite = IsLegacyProject(project),
                     ValidateRuntimeAssets = project.IsPropertyTrue("ValidateRuntimeIdentifierCompatibility"),
                     CentralPackageVersionsEnabled = isCentralPackageManagementEnabled && projectStyle == ProjectStyle.PackageReference,
+                    CentralPackageFloatingVersionsEnabled = isCentralPackageFloatingVersionsEnabled,
                     CentralPackageVersionOverrideDisabled = isCentralPackageVersionOverrideDisabled,
                     CentralPackageTransitivePinningEnabled = isCentralPackageTransitivePinningEnabled,
                     RestoreAuditProperties = auditProperties
@@ -1030,15 +1031,15 @@ namespace NuGet.Build.Tasks.Console
         /// </summary>
         /// <param name="project">The <see cref="IMSBuildProject" /> to get the central package management settings for.</param>
         /// <param name="projectStyle">The <see cref="ProjectStyle?" /> of the specified project.  Specify <see langword="null" /> when the project does not define a restore style.</param>
-        /// <returns>A <see cref="Tuple{T1, T2}" /> containing values indicating whether or not central package management is enabled and if the ability to override a package version is disabled.</returns>
-        internal static (bool IsEnabled, bool IsVersionOverrideDisabled, bool IsCentralPackageTransitivePinningEnabled) GetCentralPackageManagementSettings(IMSBuildProject project, ProjectStyle? projectStyle)
+        /// <returns>A <see cref="Tuple{T1, T2, T3, T4}" /> containing values indicating whether or not central package management is enabled, if the ability to override a package version is disabled, if the ability to transitively pin a package version is enabled, and if floating versions are enabled.</returns>
+        internal static (bool IsEnabled, bool IsVersionOverrideDisabled, bool IsCentralPackageTransitivePinningEnabled, bool IsFloatingVersionsEnabled) GetCentralPackageManagementSettings(IMSBuildProject project, ProjectStyle? projectStyle)
         {
             if (!projectStyle.HasValue || (projectStyle.Value == ProjectStyle.PackageReference))
             {
-                return (project.IsPropertyTrue("_CentralPackageVersionsEnabled"), project.IsPropertyFalse("CentralPackageVersionOverrideEnabled"), project.IsPropertyTrue("CentralPackageTransitivePinningEnabled"));
+                return (project.IsPropertyTrue("_CentralPackageVersionsEnabled"), project.IsPropertyFalse("CentralPackageVersionOverrideEnabled"), project.IsPropertyTrue("CentralPackageTransitivePinningEnabled"), project.IsPropertyTrue("CentralPackageFloatingVersionsEnabled"));
             }
 
-            return (false, false, false);
+            return (false, false, false, false);
         }
 
         /// <summary>
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
@@ -599,12 +599,17 @@ namespace NuGet.Commands
                 await _logger.LogAsync(RestoreLogMessage.CreateError(NuGetLogCode.NU1010, string.Format(CultureInfo.CurrentCulture, Strings.Error_CentralPackageVersions_MissingPackageVersion, string.Join(";", packageReferencedDependenciesWithoutCentralVersionDefined.Select(d => d.Name)))));
                 return false;
             }
-            var floatingVersionDependencies = _request.Project.TargetFrameworks.SelectMany(tfm => tfm.CentralPackageVersions.Values).Where(cpv => cpv.VersionRange.IsFloating);
-            if (floatingVersionDependencies.Any())
+
+            if (!restoreRequest.Project.RestoreMetadata.CentralPackageFloatingVersionsEnabled)
             {
-                await _logger.LogAsync(RestoreLogMessage.CreateError(NuGetLogCode.NU1011, Strings.Error_CentralPackageVersions_FloatingVersionsAreNotAllowed));
-                return false;
+                var floatingVersionDependencies = _request.Project.TargetFrameworks.SelectMany(tfm => tfm.CentralPackageVersions.Values).Where(cpv => cpv.VersionRange.IsFloating);
+                if (floatingVersionDependencies.Any())
+                {
+                    await _logger.LogAsync(RestoreLogMessage.CreateError(NuGetLogCode.NU1011, Strings.Error_CentralPackageVersions_FloatingVersionsAreNotAllowed));
+                    return false;
+                }
             }
+
             return true;
         }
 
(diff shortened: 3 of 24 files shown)


martinrrm (commented):
Should we also enable floating versions in the PM UI?

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -599,12 +599,17 @@ private async Task<bool> AreCentralVersionRequirementsSatisfiedAsync(RestoreRequ
...
-            if (floatingVersionDependencies.Any())
+
+            if (!restoreRequest.Project.RestoreMetadata.CentralPackageFloatingVersionsEnabled)
             {
-                await _logger.LogAsync(RestoreLogMessage.CreateError(NuGetLogCode.NU1011, Strings.Error_CentralPackageVersions_FloatingVersionsAreNotAllowed));
-                return false;
+                var floatingVersionDependencies = _request.Project.TargetFrameworks.SelectMany(tfm => tfm.CentralPackageVersions.Values).Where(cpv => cpv.VersionRange.IsFloating);
+                if (floatingVersionDependencies.Any())
+                {
+                    await _logger.LogAsync(RestoreLogMessage.CreateError(NuGetLogCode.NU1011, Strings.Error_CentralPackageVersions_FloatingVersionsAreNotAllowed));
```
martinrrm, line 608: Should we update the error message to say that this can be enabled now?

src/NuGet.Core/NuGet.Build.Tasks.Console/Program.cs
```diff
@@ -7,7 +7,9 @@
 using System.Globalization;
 using System.IO;
 using System.Linq;
+#if NETFRAMEWORK
```
martinrrm: Can you explain me what is this change doing?

jeffkl: This namespace is only needed for .NET Framework so when you build for .NET Core in VS this line has a red squiggly so this just cleans that up

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
jeffkl: I explicitly don't want the error message to indicate how users can enable it.  I'd much rather users have to go to the documentation and read about the implications of floating versions and hopefully steer them away from it.

jeffkl:
> Should we also enable floating versions in the PM UI?

How much work would it be for PM UI to read the `CentralPackageFloatingVersionsEnabled` property and then allow floating versions?

martinrrm:
> > Should we also enable floating versions in the PM UI?
> 
> How much work would it be for PM UI to read the `CentralPackageFloatingVersionsEnabled` property and then allow floating versions?

@jeffkl Probably a day for work and another day for testing, we are already reading the CPM enabled property and that decides if the combobox is going to be editable or not. Floating Versions in PM UI is only a project level feature, since this is more of a "Solution" level feature is it worth enabling it?

jeffkl:
> > > Should we also enable floating versions in the PM UI?
> > 
> > 
> > How much work would it be for PM UI to read the `CentralPackageFloatingVersionsEnabled` property and then allow floating versions?
> 
> @jeffkl Probably a day for work and another day for testing, we are already reading the CPM enabled property and that decides if the combobox is going to be editable or not. Floating Versions in PM UI is only a project level feature, since this is more of a "Solution" level feature is it worth enabling it?

@martinrrm I guess we might as well read the new property and allow the user to specify a floating version?  If its only 1 day of work I think its worth it.  I would also be interested in adding telemetry so we know how many users are enabling it.

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
```diff
@@ -1030,15 +1031,15 @@ private ICollection<ProjectWithInnerNodes> LoadProjects(IEnumerable<ProjectGraph
...
+        /// <returns>A <see cref="Tuple{T1, T2, T3, T4}" /> containing values indicating whether or not central package management is enabled, if the ability to override a package version is disabled, if the ability to transitively pin a package version is enabled, and if floating versions are enabled.</returns>
+        internal static (bool IsEnabled, bool IsVersionOverrideDisabled, bool IsCentralPackageTransitivePinningEnabled, bool IsFloatingVersionsEnabled) GetCentralPackageManagementSettings(IMSBuildProject project, ProjectStyle? projectStyle)
         {
             if (!projectStyle.HasValue || (projectStyle.Value == ProjectStyle.PackageReference))
             {
-                return (project.IsPropertyTrue("_CentralPackageVersionsEnabled"), project.IsPropertyFalse("CentralPackageVersionOverrideEnabled"), project.IsPropertyTrue("CentralPackageTransitivePinningEnabled"));
+                return (project.IsPropertyTrue("_CentralPackageVersionsEnabled"), project.IsPropertyFalse("CentralPackageVersionOverrideEnabled"), project.IsPropertyTrue("CentralPackageTransitivePinningEnabled"), project.IsPropertyTrue("CentralPackageFloatingVersionsEnabled"));
```
zivkan: These changes are almost identical to the changes in `MSBuildRestoreUtility.cs`. Can this method be deleted, and call a shared one instead?

When I was working on NuGetAudit, I believe it was you who suggested more or less the same thing, so the way that NuGetAudit's properties are read for static vs "normal" restore, use a shared method in some kind of `MSBuildUtilities` class.

test/NuGet.Core.Tests/NuGet.Commands.Test/MSBuildRestoreUtilityTests.cs
```diff
@@ -3829,6 +3829,113 @@ public void MSBuildRestoreUtility_GetPackageSpec_CPVM_VersionOverrideCanBeDisabl
...
         }
 
+        [Theory]
+        [InlineData(null, false)]
+        [InlineData("", false)]
+        [InlineData("                     ", false)]
+        [InlineData("false", false)]
+        [InlineData("invalid", false)]
+        [InlineData("true", true)]
+        [InlineData("           true    ", true)]
```
zivkan: would it be more in the spirit of unit testing to make `MSBuildRestoreUtility.IsPropertyTrue` internal, then have a unit test validating whitespace is trimmed correctly, so that tests titled "CPVM floating versions can be enabled", so we don't need so many duplicate tests for a test whose named resposibility is not validating `IsPropertyTrue` edge cases?

test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/RestoreCommandTests.cs
```diff
@@ -2031,14 +2045,21 @@ public async Task RestoreCommand_CentralVersion_ErrorWhenFloatingCentralVersions
...
-                Assert.False(result.Success);
-                Assert.Equal(1, logger.ErrorMessages.Count);
-                logger.ErrorMessages.TryDequeue(out var errorMessage);
-                Assert.True(errorMessage.Contains("Centrally defined floating package versions are not allowed."));
-                var messagesForNU1011 = result.LockFile.LogMessages.Where(m => m.Code == NuGetLogCode.NU1011);
-                Assert.Equal(1, messagesForNU1011.Count());
-
+                if (enabled)
+                {
+                    // Assert
```
zivkan: nitpick: IMO the assert comment can move to before the `if (enabled)` line, and the second assert comment can be removed. I think the point of the Arrange, Act, Assert pattern is to break tests into these three broad sections, rather than to be very precise about exactly when the first Assert statement is executing.

zivkan (dismissed):
You probably already know, but just in case, we'll need to make dotnet/project-system change to have the new property flow through nominations.

> CPM is considered an enterprise-level feature

I don't agree, and I wish this would stop being repeated. Given the amount of customer feedback we've gotten, I think it's clear that CPM is broadly interesting, not just for "enterprise" developers. Ever since I first heard about the idea I thought it would be useful even for personal projects with more than 1 project in the repo, so I really don't understand why certain people keep repeating that it's "an enterprise-level feature".

jeffkl:
> You probably already know, but just in case, we'll need to make dotnet/project-system change to have the new property flow through nominations.

Pul request created for the property here: https://github.com/dotnet/project-system/pull/9295

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
jeffkl: Yeah good idea, I've had to add something to the public API for this but still a good change.

test/NuGet.Core.Tests/NuGet.Commands.Test/MSBuildRestoreUtilityTests.cs
jeffkl: Good idea, done!

test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/RestoreCommandTests.cs
```diff
@@ -1984,12 +1984,22 @@ public async Task RestoreCommand_DowngradeIsNotErrorWhen_DowngradedByCentralDire
             }
         }
 
-        [Fact]
-        public async Task RestoreCommand_CentralVersion_ErrorWhenFloatingCentralVersions()
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+
```
donnie-msft: nit: whitespace

test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/RestoreCommandTests.cs
```diff
@@ -1984,12 +1984,22 @@ public async Task RestoreCommand_DowngradeIsNotErrorWhen_DowngradedByCentralDire
...
+        public async Task RestoreCommand_CentralVersion_ErrorWhenFloatingCentralVersionsAreDisabled(bool enabled)
```
donnie-msft: Trying to be more specific with the name. Also, I'm in favor of dropping `RestoreCommand_` from any of these test names.

```suggestion
        public async Task RestoreCommand_CentralPackageFloatingVersions_FloatingVersionsArePresent_SucceedsWhenEnabledAndErrorsWhenDisabled(bool enabled)
```

donnie-msft (dismissed):

src/NuGet.Core/NuGet.ProjectModel/ProjectRestoreMetadata.cs
```diff
@@ -198,6 +204,7 @@ public bool Equals(ProjectRestoreMetadata other)
                    EqualityUtility.EqualsWithNullCheck(ProjectWideWarningProperties, other.ProjectWideWarningProperties) &&
                    EqualityUtility.EqualsWithNullCheck(RestoreLockProperties, other.RestoreLockProperties) &&
                    EqualityUtility.EqualsWithNullCheck(CentralPackageVersionsEnabled, other.CentralPackageVersionsEnabled) &&
+                   EqualityUtility.EqualsWithNullCheck(CentralPackageFloatingVersionsEnabled, other.CentralPackageFloatingVersionsEnabled) &&
```
nkolev92, line 207: can you update the cloning and equality tests?

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.cs
```diff
@@ -935,6 +936,10 @@ private static void ReadMSBuildMetadata(JsonTextReader jsonReader, PackageSpec p
                         centralPackageVersionsManagementEnabled = ReadNextTokenAsBoolOrFalse(jsonReader, packageSpec.FilePath);
                         break;
 
+                    case "centralPackageFloatingVersionsEnabled":
```
nkolev92, line 939: can we update the json reader tests? There shopuld be some round trip ones.

src/NuGet.Clients/NuGet.SolutionRestoreManager/VsSolutionRestoreService.cs
```diff
@@ -348,6 +348,7 @@ internal static PackageSpec ToPackageSpec(ProjectNames projectNames, IEnumerable
                     CacheFilePath = NoOpRestoreUtilities.GetProjectCacheFilePath(cacheRoot: outputPath),
                     RestoreLockProperties = VSNominationUtilities.GetRestoreLockProperties(TargetFrameworks),
                     CentralPackageVersionsEnabled = cpvmEnabled,
+                    CentralPackageFloatingVersionsEnabled = VSNominationUtilities.IsCentralPackageFloatingVersionsEnabled(TargetFrameworks),
```
nkolev92, line 351: Can we add tests for this as well?

nkolev92 (commented):
Looks good to me. I think this change is touching a lot of different parts of the code, so I think we should make sure we add the good test coverage.

test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/RestoreCommandTests.cs
jeffkl: I'm going to leave this as-is for now since its an existing test that I'm editing.  I want to be able to see history of the test in case it starts failing intermittently after this change.  All of the test names should be updated later in another PR if we get to it.

test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/RestoreCommandTests.cs
jeffkl: Fixed

src/NuGet.Core/NuGet.ProjectModel/ProjectRestoreMetadata.cs
jeffkl: Done, and I also determined that there were no tests for `CentralPackageVersionOverrideDisabled` or `CentralPackageTransitivePinningEnabled` so I added them.

src/NuGet.Clients/NuGet.SolutionRestoreManager/VsSolutionRestoreService.cs
jeffkl: I don't see any tests for `VSNominationUtilities`.  Should I create one with only tests for the CPM methods?

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.cs
jeffkl: Done

src/NuGet.Clients/NuGet.SolutionRestoreManager/VsSolutionRestoreService.cs
jeffkl: Done

test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/RestoreCommandTests.cs
donnie-msft: A follow-up PR to address the whole thing is generally fine, but I don't follow your comment about seeing test case history. You edited the test name in this PR, so how is history being affected by editing it again before merging?

test/NuGet.Core.Tests/NuGet.ProjectModel.Test/PackageSpecTests.cs
```diff
@@ -437,6 +437,82 @@ public void ProjectRestoreMetadataCloneChangeSourcesTest()
...
+
+            // Assert
+            Assert.NotEqual(originalProjectRestoreMetadata, happyClone);
+            Assert.Equal(originalProjectRestoreMetadata.CentralPackageVersionOverrideDisabled, !happyClone.CentralPackageVersionOverrideDisabled);
+        }
+
+        [Fact]
+        public void ProjectRestoreMetadataCloneChangeCentralPackageTransitivePinningEnabledTest()
+        {
+            // Set up
```
donnie-msft: Just curious why the latest iteration uses a different word that basically means what our coding guidelines already suggest?

```suggestion
            // Arrange
```

test/NuGet.Clients.Tests/NuGet.SolutionRestoreManager.Test/VsSolutionRestoreServiceTests.cs
```diff
@@ -2183,6 +2183,49 @@ public void ToPackageSpec_TransitiveDependencyPinning_CanBeEnabled(string Centra
...
         }
 
+        [Theory]
+        [InlineData(null, false)]
+        [InlineData("", false)]
+        [InlineData(" ", false)]
+        [InlineData("invalid", false)]
+        [InlineData("false", false)]
+        [InlineData("true", true)]
+        [InlineData("           true    ", true)]
```
donnie-msft, line 2193: I tested a few other MSBuild properties with padded spaces, and as I expected, got build errors. Do we really want to accept spaces in this particular property? Is it just to try to be friendly to typos, or is there a specific reason?

nkolev92 (dismissed):

test/NuGet.Clients.Tests/NuGet.SolutionRestoreManager.Test/VsSolutionRestoreServiceTests.cs
jeffkl: I'm not sure if we should support it or not.  I just know that our code trims the value and a bunch of other tests check for leading/trailing whitespace so I copy/pasted it.  This could probably be cleaned up later as tech debt.

test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/RestoreCommandTests.cs
jeffkl: I didn't realize I renamed the test, thanks for pointing it out.  I think I'll rename it back.

test/NuGet.Core.Tests/NuGet.ProjectModel.Test/PackageSpecTests.cs
jeffkl: This is just copy/pasted from other tests, I'll update them all

nkolev92 (approved):

test/NuGet.Core.Tests/NuGet.ProjectModel.Test/PackageSpecTests.cs
nkolev92: > Just curious why the latest iteration uses a different word that basically means what our coding guidelines already suggest?

The rest of that class uses `//Set up`. 

fwiw, the comments in general are option, you don't need to have `arrange` or anything like that. 
https://github.com/NuGet/NuGet.Client/blob/dev/docs/coding-guidelines.md#unit-test-structure

>The contents of every unit test should be split into three distinct stages, *optionally* separated by these comments:



Another note is that the coding style within a class is acceptable as well: https://github.com/NuGet/NuGet.Client/blob/dev/docs/coding-guidelines.md#c-coding-style

> When editing files, keep new code and changes consistent with the style in the files. For new files, it should conform to the style for that component

I don't mean that we shouldn't use `Arrange` because the rest of the classes set up since this is minor, but simply it's not something our guidelines are too strict about.

donnie-msft: My feedback was not blocking. I saw a bunch of new tests being added with a pattern I didn't recognize. I do see now there are 4 instances above. Rather than double the tests using a pattern from 6+ years ago, IMO it's negligible effort to just use the current pattern.

test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/RestoreCommandTests.cs
donnie-msft: Looks like you put the original test name back. Ok, makes sense.

donnie-msft (approved):
Consider filing a techdebt issue for the spaces around the msbuild property values.

kzu:
Could we know what milestone this will be included in @jeffkl @nkolev92? Thanks!

PR #5442, opened 2023-10-01
Fix tests failing when run on systems with non-English language/locale settings

### LINKED ISSUE #12820: Test:  use product resource string instead of hard-coded en-US string
It should be easy to clone the repo, build and test, irrespective of the current user's locale.  There's at least [one report](https://github.com/NuGet/NuGet.Client/pull/5002#issuecomment-1679707317) that this workflow was not possible without first changing the locale to en-US.  While the report lacks detail, I suspect that the problem is that some tests use hard-coded en-US strings ([example](https://github.com/NuGet/NuGet.Client/blob/4d815bf71aa1ce5fe8d9d9da6573303d7093c2f1/test/NuGet.Clients.FuncTests/NuGet.CommandLine.FuncTest/Commands/InstallCommandTests.cs#L32)) instead of using the product's resource string.

dotnokato:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12820

## Description

Fixed tests that were failing when run on systems with non-English language/locale settings

- tests that relied on English language
  - updated asserts to use product resource string instead of hardcoded en-US strings where viable
  - for tests asserting exception message text, set the `Thread.CurrentThread.CurrentUICulture = CultureInfo.InvariantCulture;`
- tests failing due to expected values being English-localized (e.g. decimal separator = `.`)
  - converted the expected values from string to decimal and back to (localized) string
  - set `Thread.CurrentThread.CurrentCulture = CultureInfo.InvariantCulture;` where the above was not viable
- updated test `PerformUpToDateCheck_WhenNonBuildIntegratedProjectIsAParentOfADirtySpec_ReturnsAListWithoutNonBuildIntegratedProjects` to use `TestDirectory` instead of creating test files in `C:\` drive
- renamed test `ToReadableTimeFormat_Localized_AssumesEnglishLocale_ContainsTimeNumber_Succeeds` to `ToReadableTimeFormat_Localized_AssumesCurrentCulture_ContainsTimeNumber_Succeeds` so that the name corresponds to the behaviour of method under test (after verifying the way it is being used)

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
+++ b/test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Threading;
@@ -2290,6 +2291,9 @@ namespace NuGet.PackageManagement.VisualStudio.Test
         [Fact]
         public async Task TestPackageManager_CancellationTokenPassed()
         {
+            // Get exception messages in English
+            Thread.CurrentThread.CurrentUICulture = CultureInfo.InvariantCulture;
+
             using (var testDirectory = TestDirectory.Create())
             using (var testSolutionManager = new TestSolutionManager())
             {
--- a/test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
+++ b/test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Threading;
@@ -2289,11 +2288,9 @@ namespace NuGet.PackageManagement.VisualStudio.Test
         }
 
         [Fact]
+        [UseCulture("en-US")] // We are asserting exception messages in English
         public async Task TestPackageManager_CancellationTokenPassed()
         {
-            // Get exception messages in English
-            Thread.CurrentThread.CurrentUICulture = CultureInfo.InvariantCulture;
-
             using (var testDirectory = TestDirectory.Create())
             using (var testSolutionManager = new TestSolutionManager())
             {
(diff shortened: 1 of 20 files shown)


test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
```diff
@@ -2290,6 +2291,9 @@ public async Task TestPackageManager_DowngradePackageFor_BottomtProject_Success(
         [Fact]
         public async Task TestPackageManager_CancellationTokenPassed()
         {
+            // Get exception messages in English
+            Thread.CurrentThread.CurrentUICulture = CultureInfo.InvariantCulture;
```
jeffkl: Would it make sense to set this globally for all of our tests so that in the future we don't cause this to happen again?

dotnokato: I think it would. The question is how to do it assembly/codebase wise?
Or do we want at least to set this on a test class level instead of individual tests level?

zivkan: To be honest, making the threads run in invariant culture doesn't make me proud of our code quality.

Something I particularly don't like is that there's no cleanup afterwards to set the culture back to the original value. If there are other tests that explicitly want to test some other culture (I think Fernando added one or two within the last 2 years, but before that I'd say NuGet has none), then this could introduce flakiness depending on whether or not the other test re-uses the same thread. We also run tests in parllel, and this method is async, which increases risk that another test could re-use the same thread.

My memory is that xunit has its own set of threads to run tests on, which means that anywhere NuGet does `.ConfigureAwait(false)` (which I think we only have two or three examples of, in NuGet.Protocol, but we might end up doing more in the future), or anywhere `Task.Run()` is used, code might end up running on the thread pool, which is still using the default UI culture.

There's always a case of "don't let perfect be the enemy of good", so we don't have to change the test code to work correctly when using the default system culture. Maybe this "set current thread UI culture to invariant" is good enough.  But how difficult is it to "properly fix" tests instead?  I've been meaning to try myself, but I don't want to change my system locale on a machine I really use, and creating a new VM, installing all the dev tools, and cloning the repo has been too much effort for me so far.

dotnokato:
I've pushed updated version that uses `UseCulture` attribute similar to https://github.com/xunit/samples.xunit/blob/main/UseCulture/UseCultureAttribute.cs. This way, the thread culture is only changed for the time of the test. Then it is being reset to the original value.

jeffkl (approved):

PR #5453, opened 2023-10-03
Respect interactive option when evaluation projects in static graph-based restore

### LINKED ISSUE #12907: Static graph-based restore does not respect Interactive option when loading projects
### NuGet Product Used

MSBuild.exe

### Product Version

Any

### Worked before?

No

### Impact

### Repro Steps & Context

Static graph-based restore evaluates MSBuild projects in-process but there previously was not an option for these evaluations to respect being interactive/non-interactive.  I fixed this by updating the proper APIs in MSBuild to allow setting this option: https://github.com/dotnet/msbuild/pull/8533

Static graph-based restore will need to compile against MSBuild 17.7+ to see this API and set it based on the user's option.

### Verbose Logs

_No response_

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12907

## Description

Previous versions of MSBuild did not accept an Interactive flag when evaluating projects.  This can lead to the user not being prompted for credentials in static graph-based restore even if the user has specified the `/interactive` command-line argument.  

This pull request contains two commits:
* Updated to MSBuild 17.7.2 so that the new API is available
* Pass the interactive option to the project load settings during project evaluation in static graph-based restore

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception - There is no abstraction currently to test this effectively
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


nkolev92 (approved):

zivkan (approved):
Honestly, I'm confused how this works. Does NuGet.Build.Tasks.Console.exe capture stdin when our MSBuild task launches it as a child process?  I was assuming no, and that there'd need to be some IPC, but I don't see any code changes for that. 🤷   Well, I'm assuming you tested and it worked, but it's entirely possible that child processes do automatically capture stdin. It's just not intuitive, because if a console app launches two child processes, which one gets stdin?  Anyway, I'm rambling and getting out of scope of this PR.

jeffkl:
> Honestly, I'm confused how this works. Does NuGet.Build.Tasks.Console.exe capture stdin when our MSBuild task launches it as a child process? I was assuming no, and that there'd need to be some IPC, but I don't see any code changes for that. 🤷 Well, I'm assuming you tested and it worked, but it's entirely possible that child processes do automatically capture stdin. It's just not intuitive, because if a console app launches two child processes, which one gets stdin? Anyway, I'm rambling and getting out of scope of this PR.

This flag only ends up being passed to the NuGet-based MSBuild project SDK resolver which does not prompt during authentication.  It either emits a message with device flow or pops up a dialog.  Prompting the user for credentials via STDIN would not work at all in this scenario.

PR #5459, opened 2023-10-13
Transitive dependencies in GPF are automatically source mapped by actions in PM UI 

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2421

## Description

Creating source mappings from PM UI actions now considers the GPF for transitive dependencies.

Following the [Tables of Behaviors in the spec](https://github.com/NuGet/Home/blob/dev/proposed/2023/package-source-mapping-actions-PMUI.md#table-behavior-for-packages-not-installed-to-the-gpf), when transitive packages exists in the GPF:
  1. If it's an enabled source for this solution, map to that source.
  2. If it's not an enabled source in this solution, fail with an error indicating the package/source in the message. 
       - The error is thrown from PM UI in this PR. Eventually, when implementing https://github.com/NuGet/Client.Engineering/issues/1166, we will be able to move this error to Restore.

Made a UI adjustment to the margins of the Preview Changes dialog so that multiple sources have spacing between them (ie, it appears as if a blank line separates each `Creating source mappings to` block in the list).

#### Top-level & Transitive packages not mapped (multiple sources found in GPF)
Multiple sources based on GPF state:
image

#### Transitive not mapped & in GPF to enabled source
Even just missing a mapping for 1 transitive (top level already mapped) still gets a mapping created:
image

#### Transitive not mapped & not in GPF
After clearing `System.IO` from my GPF, now it pulled it from selected source `vside`
(Notice the output window indicates downloading the package unlike the previous screenshot)
image

#### Transitive not mapped & in GPF but the source is not enabled
An error is thrown from PM UI since the GPF is pointing to `vside`, which is a source I have temporarily disabled here:
image

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added.
  
There wasn't a test utility to write `.nupkg.metadata`, so I added a method signature to `SimpleTestPackageUtility` that accepts a source. It was only writing `null`, before.

  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
@@ -3,10 +3,15 @@
 
 #nullable enable
 
+using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using NuGet.Configuration;
 using NuGet.PackageManagement.VisualStudio;
+using NuGet.Packaging;
+using NuGet.Packaging.Core;
+using NuGet.Protocol.Core.Types;
 
 namespace NuGet.PackageManagement.UI
 {
@@ -16,14 +21,14 @@ namespace NuGet.PackageManagement.UI
         /// Determines if new package source mappings should be written to the <paramref name="sourceMappingProvider"/> which is returned as a count.
         /// </summary>
         /// <param name="userAction"></param>
-        /// <param name="addedPackageIds"></param>
+        /// <param name="sourceMappingPreviewResult"></param>
         /// <param name="sourceMappingProvider"></param>
         /// <param name="existingPackageSourceMappingSourceItems"></param>
         /// <param name="countCreatedTopLevelSourceMappings">For Top-Level packages: <see langword="null" /> if not applicable; 0 if none needed to be added; > 0 is the count of new package source mappings added.</param>
         /// <param name="countCreatedTransitiveSourceMappings">For Transitive packages: <see langword="null" /> if not applicable; 0 if none needed to be added; > 0 is the count of new package source mappings added.</param>
-        internal static void ConfigureNewPackageSourceMapping(
+        internal static void ConfigureNewPackageSourceMappings(
             UserAction? userAction,
-            IReadOnlyList<string>? addedPackageIds,
+            PreviewResult? sourceMappingPreviewResult,
             PackageSourceMappingProvider sourceMappingProvider,
             IReadOnlyList<PackageSourceMappingSourceItem> existingPackageSourceMappingSourceItems,
             out int? countCreatedTopLevelSourceMappings,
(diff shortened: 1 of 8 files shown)


JonDouglas:
Great error message, but how can we empower the user to rid themselves of this error? There should be something at the end helping the user take the right action such as adding a new PSM with that package and a new or the existing source?

donnie-msft:
> Great error message, but how can we empower the user to rid themselves of this error? There should be something at the end helping the user take the right action such as adding a new PSM with that package and a new or the existing source?

Sure. What do you think of this first draft?

> The package `packageC` is available in the Global packages folder, but the source it came from `https://api.nuget.org/v3/index.json` is not one of this solution's configured sources. Consider source mapping this package explicitly to one of the enabled package sources in this solution. Alternatively, you can configure a separate Global packages folder for this solution, or clear this package from the current Global packages folder. See [How to manage the global packages, cache, temp folders in NuGet | Microsoft Learn](https://learn.microsoft.com/en-us/nuget/consume-packages/managing-the-global-packages-and-cache-folders)

donnie-msft:
Hi @JonDouglas - thanks for the suggestions! For this PR, I've decided to leave it without the action, so that it can be addressed via a new NU error code (and hyperlink). Tracking that in https://github.com/NuGet/Home/issues/12953

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -16,14 +21,14 @@ internal static class PackageSourceMappingUtility
...
         /// </summary>
         /// <param name="userAction"></param>
-        /// <param name="addedPackageIds"></param>
+        /// <param name="sourceMappingPreviewResult"></param>
         /// <param name="sourceMappingProvider"></param>
         /// <param name="existingPackageSourceMappingSourceItems"></param>
         /// <param name="countCreatedTopLevelSourceMappings">For Top-Level packages: <see langword="null" /> if not applicable; 0 if none needed to be added; > 0 is the count of new package source mappings added.</param>
         /// <param name="countCreatedTransitiveSourceMappings">For Transitive packages: <see langword="null" /> if not applicable; 0 if none needed to be added; > 0 is the count of new package source mappings added.</param>
-        internal static void ConfigureNewPackageSourceMapping(
+        internal static void ConfigureNewPackageSourceMappings(
```
jebriede, line 29: This method would benefit from unit tests that cover and validate its different code branches. The unit tests would also help in documenting what the expected output will be based on the various inputs, especially as the method's complexity increases.

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -32,73 +37,108 @@ internal static class PackageSourceMappingUtility
...
+                }
 
-            CreateAndSavePackageSourceMappings(
-                userAction.SelectedSourceName,
-                addedPackageIdsWithoutExistingMappings,
-                sourceMappingProvider,
-                existingPackageSourceMappingSourceItems);
+                return nupkgMetadata.Source;
+            }
+            catch
```
jebriede, line 118: To prevent this code from swallowing all exceptions thrown, we should catch specific exceptions we are expecting. However, we may not need to catch any in this code. If I'm reading this correctly, the call [here](https://github.com/NuGet/NuGet.Client/blob/7252122110f1e898516f0d8471a57d92f59b7f25/src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs#L109) calls into [this](https://github.com/NuGet/NuGet.Client/blob/9d6b1c6980445a84ee9fc0f114c79eb6bc32a8ba/src/NuGet.Core/NuGet.Packaging/NupkgMetadata/NupkgMetadataFileFormat.cs#L22-L25) which in turn calls [this](https://github.com/NuGet/NuGet.Client/blob/9d6b1c6980445a84ee9fc0f114c79eb6bc32a8ba/src/NuGet.Core/NuGet.Packaging/NupkgMetadata/NupkgMetadataFileFormat.cs#L29) which handles the UnauthorizedAccessException and IOException for us, returning the default. If what I'm seeing is correct, we don't need to handle those exceptions and unless we're expecting exceptions from the call to resolver.GetNupkgMetadataPath, we can remove the try/catch altogether since we're using a path that calls into a SafeRead.

Otherwise, we should catch and handle specific exceptions.

jebriede (approved):
Looks good! Left a comment for a method growing in complexity to have unit tests added. I'm approving with comments in case you want to create a new issue to add tests in a follow-up PR, but if you add them for this PR, I'll review again. Thanks!

Nigusu-Allehu (approved):

nkolev92 (approved):

donnie-msft:
>  I'm approving with comments in case you want to create a new issue to add tests in a follow-up PR,

I attempted to write an explicit test for `ConfigureNewPackageSourceMappings`, but it was going to take >1 hour and it felt like refactoring the Configuration classes would be a better first step. Since the configuration types can't be mocked, then a very bulky Setup and assert logic would be necessary to test which mappings the provider updated. Therefore, I suggest my TechDebt issue be addressed as step1 and then adding more these tests as step2.

https://github.com/NuGet/Home/issues/12825

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
donnie-msft: I can understand how a method called "SafeRead" sounds like it should be "safely" handling exceptions, but it isn't actually doing that. It's more of a retry mechanism. 

I disagree that handling specific exceptions is a good idea. There are many possible types thrown here. For example, the `FileStream` constructor alone has about a dozen possible exception types it can throw. 

image

That's in addition to bad data in the file itself: https://github.com/NuGet/NuGet.Client/blob/84cc5cc305e0c2cd528292c1230c1f80bb0d23d9/src/NuGet.Core/NuGet.Packaging/NupkgMetadata/NupkgMetadataFileFormat.cs#L49. 

How to handle the exception? In this PR, I treat it as we couldn't check the GPF. That's technically true. Whether we want a message sent to the user can be discussed (error messagebox, output window, etc). I suggest if we want this code to log something, we start with a basic fault telemetry (Filed https://github.com/NuGet/Client.Engineering/issues/2559). I don't expect this will have a high hit rate (possibly nearly zero). In fact, we didn't care about those faults before, but it is something worth reconsidering now that I've introduced another codepath for it. If we identify patterns, we can go from there developing an action plan for customers.

PR #5461, opened 2023-10-17
Create backup pipeline to save build artifacts

zivkan:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2479

Regression? No

## Description

copy build artifacts to a "drop" after a successful official build. We can use this as a backup, in case the normal process fails.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/eng/pipelines/backup_build_artifacts.yml
+++ b/eng/pipelines/backup_build_artifacts.yml
@@ -1,0 +1,102 @@
+# No CI trigger rely on Pipeline trigger only
+trigger: none
+
+resources:
+  pipelines:
+    - pipeline: officialbuild
+      source: NuGet.Client-Official
+      trigger: true
+
+jobs:
+  - job: Backup_Build_Artifacts
+    displayName: "Backup Build Artifacts"
+    timeoutInMinutes: 10
+    pool:
+      name: VSEngSS-MicroBuild2022-1ES
+
+    steps:
+      - checkout: none
+
+      - pwsh: |
+          $buildNumber = "$(resources.pipeline.officialbuild.runName)"
+          Write-Host "Setting build number to $buildNumber"
+          Write-Host "##vso[build.updatebuildnumber]$buildNumber"
+        displayName: "Set build number"
+
+      - pwsh: "Get-ChildItem Env: | Sort-Object Name | Format-Table -Wrap -AutoSize"
+        displayName: 'Display environment variables'
+
+      - task: DownloadPipelineArtifact@2
+        displayName: 'Download product NonRTM'
+        continueOnError: true # in case build didn't publish this artifact
+        inputs:
+          buildType: specific
+          project: $(resources.pipeline.officialbuild.projectID)
+          definition: $(resources.pipeline.officialbuild.pipelineID)
+          artifactName: 'VS15'
+          targetPath: $(Build.ArtifactStagingDirectory)/backup/VS15
+
+      - task: DownloadPipelineArtifact@2
+        displayName: 'Download product RTM'
+        continueOnError: true # in case build didn't publish this artifact
+        inputs:
+          buildType: specific
+          project: $(resources.pipeline.officialbuild.projectID)
+          definition: $(resources.pipeline.officialbuild.pipelineID)
+          artifactName: 'VS15-RTM'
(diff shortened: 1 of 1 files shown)


eng/pipelines/backup_build_artifacts.yml
```diff
@@ -0,0 +1,102 @@
...
+      - task: artifactDropTask@0
+        displayName: "Upload Drop"
+        inputs:
+          dropServiceURI: 'https://devdiv.artifacts.visualstudio.com'
+          buildNumber: 'Backup/$(System.TeamProject)/NuGet.Client/$(resources.pipeline.officialbuild.pipelineName)/$(resources.pipeline.officialbuild.runName)'
+          sourcePath: "$(Build.ArtifactStagingDirectory)/backup"
+          toLowerCase: false
+          usePat: true
+          dropMetadataContainerName: "DropMetadata-Backup"
+          retentionDays: 186
```
zivkan, line 102: This is the only thing I'm not sure about. Should we retain all our official builds for unreasonably long periods, as we did on the fileshare, or should we retain these builds for a limited period of time by default, and have our release pipeline retain this drop "permanently", but only for builds that we actually ship?

On the microbuild list there are emails saying that storage costs are non-trivial, and they're asking people to stop publishing unnecessary files (primarily symbols). So, I don't want to do the "wrong" thing and retain all official builds forever. We really only need builds that ship to customers.  But, we use these backups when our normal process fails, which once was due to unexpected expiration of a drop needed for the product publishing.

donnie-msft: There should be organization-level decisions made on how to store shipped assets.
You've set this to about 6 months, currently. After 6 months, we'll lose the assets, and I'd think that'd be a problem. 
Do we need to ask someone in the division about this?

nkolev92: Isn't that 6 months for binaries that did not ship?

Binaries we ship are retained forever.

zivkan: Right, as I wrote in my comment, one option is to:

> have our release pipeline retain this drop "permanently", but only for builds that we actually ship

The VS insertion in our release pipeline already retains the `Product/NuGet/` drop permanently. We could add another step in the release pipeline to retain this `Backup/NuGet/` drop permanently as well, but let builds that were never inserted be deleted after 6 months. Since the release pipeline isn't yaml, I couldn't modify it as part of this PR.

There are also questions if we should retain all builds that were inserted, even those that were overwritten by another insertion before it ships to customers?  Or should we have a manual step in the release pipeline that is triggered as part of our release process, just like publishing packages to nuget.org?

If anyone can think of a good reason to keep all our official builds permanently, we can do that as well, but as I wrote in my comment, the org is trying to reduce unnecessary costs. So builds that are never even inserted, they seem pointless to retain "forever" (until manually deleted)

eng/pipelines/backup_build_artifacts.yml
```diff
@@ -0,0 +1,102 @@
...
+        continueOnError: true # in case build didn't publish this artifact
+        inputs:
+          buildType: specific
+          project: $(resources.pipeline.officialbuild.projectID)
+          definition: $(resources.pipeline.officialbuild.pipelineID)
+          artifactName: 'VS15-RTM'
+          targetPath: $(Build.ArtifactStagingDirectory)/backup/VS15-RTM
+
+      - task: DownloadPipelineArtifact@2
+        displayName: 'Download nupkgs'
```
nkolev92: The rest of the steps have a nonRTM and RTM. 
To be consistent you can rename this to nonRTM

nkolev92 (dismissed):
LGTM. 
I have one naming suggestion, but I'm not too concerned.

eng/pipelines/backup_build_artifacts.yml
zivkan: good catch!

nkolev92 (approved):

jeffkl (approved):

donnie-msft (approved):

PR #5467, opened 2023-10-19
Specify protocol V3 when creating PackageSource in GetSymbolPackageUpdateResource

### LINKED ISSUE #12951: Symbols aren't getting pushed when using -s sourceUrl
### NuGet Product Used

NuGet.exe

### Product Version

dev branch

### Worked before?

_No response_

### Impact

### Repro Steps & Context

1. Build or aquire any valid with symbols `A.nupkg` and its corresponding `A.snupkg`
2. Specify a source which is not within a Nuget.Config
3. run: `nuget push A.nupkg -Source <source> -ApiKey <apikey>`

Expected Results:
- The `nupkg` and `snupkg` are pushed
- Authentication is requested from the user
- An error is thrown stating that the resource could not be accessed without authentication

Actual:
1. The `nupkg` is pushed

### Verbose Logs

_No response_

Zaneo:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12951

## Description
In `GetSymbolPackageUpdateResource` if the requested `source` cannot be found in the current `sourceProvider` a new `PackageSource` is created. Then using V3 requests the `SymbolPackageUpdateResourceV3` is attempted to be retreived. This will obviously fail on a V2 package source.

This changes the creation of `PackageSource` to use `ProtocolV3`.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Commands/Utility/CommandRunnerUtility.cs
+++ b/src/NuGet.Core/NuGet.Commands/Utility/CommandRunnerUtility.cs
@@ -106,7 +106,11 @@ namespace NuGet.Commands
 
             if (packageSource == null)
             {
-                packageSource = new PackageSource(source);
+                // Since SymbolPackageUpdateResourceV3 will be requested this should be a v3 PackageSource
+                packageSource = new PackageSource(source)
+                {
+                    ProtocolVersion = 3
+                };
             }
 
             var sourceRepositoryProvider = new CachingSourceProvider(sourceProvider);
--- a/src/NuGet.Core/NuGet.Commands/Utility/CommandRunnerUtility.cs
+++ b/src/NuGet.Core/NuGet.Commands/Utility/CommandRunnerUtility.cs
@@ -109,7 +109,7 @@ namespace NuGet.Commands
                 packageSource = new PackageSource(source);
                 
                 // If it ends with "index.json" then treat it as a V3 Protocol PackageSource
-                if (packageSource.Source.EndsWith("index.json", StringComparison.OrdinalIgnoreCase)
+                if (packageSource.Source.EndsWith("index.json", StringComparison.OrdinalIgnoreCase))
                 {
                     packageSource.ProtocolVersion = 3;
                 }
--- a/src/NuGet.Core/NuGet.Commands/Utility/CommandRunnerUtility.cs
+++ b/src/NuGet.Core/NuGet.Commands/Utility/CommandRunnerUtility.cs
@@ -106,11 +106,13 @@ namespace NuGet.Commands
 
             if (packageSource == null)
             {
-                // Since SymbolPackageUpdateResourceV3 will be requested this should be a v3 PackageSource
-                packageSource = new PackageSource(source)
+                packageSource = new PackageSource(source);
+                
+                // If it ends with "index.json" then treat it as a V3 Protocol PackageSource
+                if (packageSource.Source.EndsWith("index.json", StringComparison.OrdinalIgnoreCase)
                 {
-                    ProtocolVersion = 3
-                };
+                    packageSource.ProtocolVersion = 3;
+                }
             }
 
             var sourceRepositoryProvider = new CachingSourceProvider(sourceProvider);


Zaneo:
@dotnet-policy-service agree

src/NuGet.Core/NuGet.Commands/Utility/CommandRunnerUtility.cs
```diff
@@ -106,7 +106,11 @@ public static async Task<SymbolPackageUpdateResourceV3> GetSymbolPackageUpdateRe
 
             if (packageSource == null)
             {
-                packageSource = new PackageSource(source);
+                // Since SymbolPackageUpdateResourceV3 will be requested this should be a v3 PackageSource
+                packageSource = new PackageSource(source)
+                {
+                    ProtocolVersion = 3
```
zivkan: This doesn't feel safe to me. I started CI to see if any tests fail, but NuGet's tests aren't exhaustive, so even if all tests pass, I still won't be fully confident.

Specifically, if the customer is pushing to a V2 package source, this change will force NuGet into thinking it's a V3 package source, and then this line: https://github.com/NuGet/NuGet.Client/blob/c6e9542fcc9300361f03d5bf7fbab1732d4d19bf/src/NuGet.Core/NuGet.Commands/Utility/CommandRunnerUtility.cs#L119

instead of returning null, because v2 package sources don't support this resource type, NuGet will throw an exception.

Therefore, my gut feeling is that this will break pushing packages to any V2 package source. We'll see if any tests fail.

Zaneo: It seemed to be returning null and not throwing an exception, that would have been preferable.

I'm ammenable to making `GetSymbolPackageUpdateResource` throw an exception if `packageSource = new PackageSource(source);` results in a V2 source.

Zaneo: image

Zaneo: I cannot see anything in `PackageSource` constructor that sets `ProtocolVersion` based on the `source`.

Zaneo: image

zivkan: 15 func tests are failing:

- CommandLine.FuncTest.Commands.PushCommandTest.PushCommand_Server_SkipDuplicate_IsSpecified_PushProceeds
- NuGet.CommandLine.FuncTest.Commands.PushCommandTest.PushCommand_Server_SkipDuplicate_NotSpecified_PushHalts
- NuGet.CommandLine.Test.NuGetPushCommandTest.PushCommand_LogsServerWarningsWhenPresent(firstServerWarning: "First of two server warning messages", secondServerWarning: "Second of two server warning messages")
- NuGet.CommandLine.Test.NuGetPushCommandTest.PushCommand_PushToServer_GetCredentialFromPlugin
- NuGet.CommandLine.Test.NuGetPushCommandTest.PushCommand_PushToServer_DoesNotDeadLockWhenSTDOutLarge
- NuGet.CommandLine.Test.NuGetPushCommandTest.PushCommand_PushToServerBasicAuth
- NuGet.CommandLine.Test.NuGetPushCommandTest.PushCommand_APIV2Package_Endpoint
- NuGet.CommandLine.Test.NuGetPushCommandTest.PushCommand_PushToServerIntegratedWindowsAuthenticationDisableBuffering
- NuGet.CommandLine.Test.NuGetPushCommandTest.PushCommand_PushToServer
- NuGet.CommandLine.Test.NuGetPushCommandTest.PushCommand_PushTimeoutErrorMessage
- NuGet.CommandLine.Test.NuGetPushCommandTest.PushCommand_PushToServerIntegratedWindowsAuthentication
- NuGet.CommandLine.Test.NuGetPushCommandTest.PushCommand_PushToServer_ApiKeyAsThirdArgument
- NuGet.CommandLine.Test.NuGetPushCommandTest.PushCommand_PushToServerFollowRedirection
- NuGet.CommandLine.Test.NuGetPushCommandTest.PushCommand_LogsServerWarningsWhenPresent(firstServerWarning: "Single server warning message", secondServerWarning: null)
- NuGet.CommandLine.FuncTest.Commands.PushCommandTest.PushCommand_AllowsTimeoutToBeSpecifiedHigherThan100Seconds

At least two of the tests failed with timeout errors, so might just be flakey, but a bunch of them failed with HTTP 404 errors, trying to get the service index, as I suspected.

NuGet.CommandLine is nuget.exe, and is ilmerged, so only builds on Windows. It's not clear from your screenshots if you're on Windows or another OS, so it might not be easy for you to run these tests. `NuGet.CommandLine.Tests` needs a nuget.exe built, and on Mac the tests run though mono-project.org. We don't run NuGet.CommandLine.Tests on Linux. I suspect NuGet.CommandLine.FuncTests probably also need a nuget.exe binary, but I'm less familiar with that test project.

Interestingly, none of the tests in `dotnet.integration.tests` failed, which tells me we don't have any dotnet CLI tests pushing to a mock V2 server in our tests. In fact, it looks like we don't have any push tests at all 😞 It should be possibly to copy one of the nuget.exe tests and port it to use the same test infrastructure as other dotnet cli tests. But this is non-trivial for a first time contributor.

Zaneo: I'm building on Windows, and ran the UnitTests provide in the solution file but did not encounter any errors there. It might be a bit difficult as you say for me to take on porting all the tests though I think it would be good to have them.

I've changed the code to only set the `ProtocolVersion` to V3 if it ends in `index.json`, following the logic I found in: https://github.com/NuGet/NuGet.Client/blob/5ce332fb834feea1f6c974d17456c8c9d5504e09/src/NuGet.Clients/NuGet.VisualStudio.Common/Telemetry/TelemetryUtility.cs

Zaneo:
I would like to add the following warning but I have two concerns:

1) If people have a workflow where they build symbol packages, then push to a V2 source, then a V3 source. The first push will generate a warning which they may not care about.

2) I think this change requiressy least two PRs? The strings in String.resx are compiled, so the builds would fail until I first merge a PR with the new string. Is the prefered order to merge the new string first in a PR. Or to do three PRs? (1) To merge the warning with hard coded text. (2) To merge the warning message to Strings.resx and (3) To change the hardcoded warning to use the String.resx entry.

https://github.com/NuGet/NuGet.Client/blob/55f4def65efe4a2b6dbf0aee8d699e7f29cc008c/src/NuGet.Core/NuGet.Commands/Strings.resx#L1038-L1041

(https://github.com/NuGet/NuGet.Client/blob/55f4def65efe4a2b6dbf0aee8d699e7f29cc008c/src/NuGet.Core/NuGet.Commands/Strings.resx#L1038-L1041)

Zaneo:
I would like to add the following warning but I have two concerns:

1) If people have a workflow where they build symbol packages, then push to a V2 source, then a V3 source. The first push will generate a warning which they may not care about.

2) I think this change requires least two PRs? The strings in String.resx are compiled, so the builds would fail until I first merge a PR with the new string. Is the prefered order to merge the new string first in a PR. Or to do three PRs? (1) To merge the warning with hard coded text. (2) To merge the warning message to Strings.resx and (3) To change the hardcoded warning to use the String.resx entry.

https://github.com/NuGet/NuGet.Client/blob/55f4def65efe4a2b6dbf0aee8d699e7f29cc008c/src/NuGet.Core/NuGet.Commands/Strings.resx#L1038-L1041

https://github.com/Zaneo/NuGet.Client/blob/55f4def65efe4a2b6dbf0aee8d699e7f29cc008c/src/NuGet.Core/NuGet.Commands/CommandRunners/PushRunner.cs#L54-L75

PR #5468, opened 2023-10-20
Create NU1110 and show in Error List & Output Window during PM UI Install

### LINKED ISSUE #12953: Create NU Error Code for Package Source Mapping & GPF conflicts
### NuGet Product(s) Affected

### Current Behavior

In https://github.com/NuGet/NuGet.Client/pull/5459#issuecomment-1770664856, the suggestion is to give actions for the customer to address their Global Packages Folder (GPF) conflicting with their solution's package source mapping configuration. 

That is, the customer is installing a package, and the GPF contains a transitive dependency for that package which indicates a source that's not an enabled source for the solution.

The current behavior introduced in https://github.com/NuGet/NuGet.Client/pull/5459 is an error message shown in the error list that states the above. However, explicit advice isn't given.

### Desired Behavior

Create a new NU Error Code that can be used as a hyperlink from the VS Error List to point to documentation with more information about resolving the issue.

### Additional Context

_No response_

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12953

## Description

- Introduce NU1110 (see Docs draft) for the error message introduced in https://github.com/NuGet/NuGet.Client/pull/5459. 
  - Having the NU code allows the customer to jump to the docs page explaining the issue and solutions.
- Failing preview results will still calculate the added/updated/removed packages so that Telemetry will still reflect the appropriate values.
  - To achieve this, a `PreviewResult` now stores a `NuGetOperationStatus`. 
  - If any GPF analysis fails, then that will be stored in the results. The action will fail if any preview failed.

#### Log NU1110 to Error List
The error list will show this error only when Intellisense is selected (ie, it's not a Build error).
image

#### Log NU1110 to Output window
The output window will always receive this error, regardless of whether Intellisense is selected.
image

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [x] Documentation PR or issue filled https://github.com/NuGet/docs.microsoft.com-nuget/pull/3169
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
@@ -6,6 +6,7 @@
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
+using NuGet.Common;
 
 namespace NuGet.PackageManagement.UI
 {
@@ -21,6 +22,8 @@ namespace NuGet.PackageManagement.UI
 
         public string Name { get; }
 
+        public NuGetOperationStatus NuGetOperationStatus { get; } = NuGetOperationStatus.Succeeded;
+
         public PreviewResult(
             string projectName,
             IEnumerable<AccessiblePackageIdentity> added,
@@ -33,13 +36,14 @@ namespace NuGet.PackageManagement.UI
             Updated = updated;
         }
 
-        public PreviewResult(Dictionary<string, SortedSet<string>>? newSourceMappings)
+        public PreviewResult(Dictionary<string, SortedSet<string>>? newSourceMappings, NuGetOperationStatus nuGetOperationStatus)
         {
             Name = Resources.Label_Solution;
             NewSourceMappings = newSourceMappings?.ToImmutableDictionary();
             Added = Enumerable.Empty<AccessiblePackageIdentity>();
             Deleted = Enumerable.Empty<AccessiblePackageIdentity>();
             Updated = Enumerable.Empty<UpdatePreviewResult>();
+            NuGetOperationStatus = nuGetOperationStatus;
         }
     }
 }
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
@@ -22,7 +22,7 @@ namespace NuGet.PackageManagement.UI
 
         public string Name { get; }
 
-        public NuGetOperationStatus NuGetOperationStatus { get; } = NuGetOperationStatus.Succeeded;
+        public NuGetOperationStatus NuGetOperationStatus { get; }
 
         public PreviewResult(
             string projectName,
@@ -34,6 +34,7 @@ namespace NuGet.PackageManagement.UI
             Added = added;
             Deleted = deleted;
             Updated = updated;
+            NuGetOperationStatus = NuGetOperationStatus.Succeeded;
         }
 
         public PreviewResult(Dictionary<string, SortedSet<string>>? newSourceMappings, NuGetOperationStatus nuGetOperationStatus)
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
@@ -380,7 +380,16 @@ namespace NuGet.PackageManagement.UI
                     TelemetryServiceUtility.StartOrResumeTimer();
 
                     IReadOnlyList<ProjectAction> actions = await resolveActionsAsync(projectManagerService);
-                    IReadOnlyList<PreviewResult> results = await GetPreviewResultsAsync(projectManagerService, actions, userAction, uiService, cancellationToken);
+                    IReadOnlyList<PreviewResult> results;
+                    try
+                    {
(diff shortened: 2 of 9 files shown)


src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/UIActionEngine.cs
```diff
@@ -380,7 +380,16 @@ private static string GenerateUpgradeReport(string? projectName, string backupPa
...
                     IReadOnlyList<ProjectAction> actions = await resolveActionsAsync(projectManagerService);
-                    IReadOnlyList<PreviewResult> results = await GetPreviewResultsAsync(projectManagerService, actions, userAction, uiService, cancellationToken);
+                    IReadOnlyList<PreviewResult> results;
+                    try
+                    {
+                        results = await GetPreviewResultsAsync(projectManagerService, actions, userAction, uiService, cancellationToken);
+                    }
+                    catch
+                    {
+                        status = NuGetOperationStatus.Failed;
```
donnie-msft, line 387: Looking for alternative suggestions. The goal is to bail out of the action if any of the `NU1110`'s occur.

src/NuGet.Clients/NuGet.PackageManagement.UI/Utility/PackageSourceMappingUtility.cs
```diff
@@ -251,6 +266,11 @@ internal static class PackageSourceMappingUtility
                     newSourceMappings.Add(sourceNameToMap, new SortedSet<string>(new List<string>(capacity: added.Count) { addedPackage.Id }));
                 }
             }
+
+            if (firstLogError != null)
+            {
+                throw new ApplicationException(firstLogError.Message);
```
donnie-msft: Throwing here whatever the first error was which serves as
1. A way to bail out of the package installation.
2. When debugging, the `NU1110` will be readily available in the stack.

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
```diff
@@ -21,6 +22,8 @@ public class PreviewResult
 
         public string Name { get; }
 
+        public NuGetOperationStatus NuGetOperationStatus { get; } = NuGetOperationStatus.Succeeded;
```
martinrrm: I prefer having the default scenario to be false, but this is a personal preference, no need to change it. Other than that everything looks good.

martinrrm (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
donnie-msft: Good point. The reason I didn't do that is the current implicit status before this PR is always a successful state. Since I only introduced 1 non-success scenario, I'm letting it default to success for now.

nkolev92: Is there any reason why we can't make the constructor set it?

src/NuGet.Core/NuGet.Common/Errors/NuGetLogCode.cs
```diff
@@ -177,6 +177,11 @@ public enum NuGetLogCode
         /// </summary>
         NU1109 = 1109,
 
+        /// <summary>
+        /// The package `packageId` is available in the Global packages folder, but the source it came from `package source URI` is not one of this solution's configured sources.
```
nkolev92:
```suggestion
        /// The package `packageId` is available in the Global packages folder, but the source it came from `package source URI` is not one of the configured sources.
```
It's just configured sources. Technically the configuration can be project based even though there's caveats so best to just not call out either.

src/NuGet.Core/NuGet.Common/Errors/NuGetLogCode.cs
```diff
@@ -177,6 +177,11 @@ public enum NuGetLogCode
...
+        /// </summary>
+        NU1110 = 1110,
```
nkolev92, line 183: I think this error code makes most sense in this category. 
NU1100 includes resolver errors which includes missing packages, missing versions etc. In a way, this is a missing package.

src/NuGet.Core/NuGet.Common/Errors/NuGetLogCode.cs
donnie-msft: Gotcha! I've updated the string.

src/NuGet.Core/NuGet.Common/Errors/NuGetLogCode.cs
donnie-msft: Great👍...

src/NuGet.Clients/NuGet.PackageManagement.UI/Actions/PreviewResult.cs
donnie-msft: > Is there any reason why we can't make the constructor set it?

Either way works. I've switched it.

donnie-msft:
Thanks for the feedback! I believe this PR is reaching its time to be merged soon, so I've opened the docs PR.
Please review the accompanying Docs PR next, thanks! https://github.com/NuGet/docs.microsoft.com-nuget/pull/3169

martinrrm (approved):

jeffkl (approved):

nkolev92 (approved):

PR #5469, opened 2023-10-23
Set NuGetAudit property defaults in MSBuild

### LINKED ISSUE #12960: Set NuGetAudit defaults in MSBuild
### NuGet Product(s) Affected

### Current Behavior

An MSBuild evaluation of the project that does not explicitly set `NuGetAudit`, `NuGetAuditLevel` or `NuGetAuditMode` will not have any values set for these properties, and therefore dotnet/project-system doesn't know what the current value is

### Desired Behavior

As per the linked issue, set defaults as part of project evaluation, so the VS project properties page knows what value to display as the current value.

### Additional Context

https://github.com/dotnet/project-system/issues/9246

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12960

Regression? No

## Description

In order to make Visual Studio's project properties window more user friendly, we need to set these properties default values in project evaluation. See https://github.com/dotnet/project-system/issues/9246 for more info.

Since this breaks NuGetAudit's telemetry for telling the difference between "explicit enable" and "implicit enable", it was changed to just "enabled" or "disabled". This is a breaking change, so telemetry queries need to be updated, but it reduces long term confusion for people looking at telemetry, as they no longer need to understand what implicit/explicit really mean.

Added skips for 2 VS E2E tests that are failing because project system's nomination isn't passing through the `NuGetAuditMode` property. I already have a PR to fix that, but we have to wait until it's merged into the dotnet/project-system repo, and then dotnet/project-system is inserted into VS, before these tests can work. Fortunately project-system inserts pretty much every day, so it shouldn't take long, just depends on how quick the PR gets merged. We can either merge this PR with the skips and I'll need to revert the skips afterwards, or we can wait for project-system to merge and insert, then I can update this PR. 

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Build.Tasks/NuGet.props
+++ b/src/NuGet.Core/NuGet.Build.Tasks/NuGet.props
@@ -38,4 +38,14 @@ Copyright (c) .NET Foundation. All rights reserved.
     <CentralPackageVersionsFileImported>true</CentralPackageVersionsFileImported>
   </PropertyGroup>
 
+  <!--
+    Setup default NuGet property values.
+    Visual Studio's project property page requires defaults to be set to inform customers what the default values are.
+  -->
+  <PropertyGroup>
+    <NuGetAudit Condition=" '$(NuGetAudit)' == '' ">true</NuGetAudit>
+    <NuGetAuditLevel Condition=" '$(NuGetAuditLevel)' == '' ">low</NuGetAuditLevel>
+    <NuGetAuditMode Condition=" '$(NuGetAuditMode)' == '' ">direct</NuGetAuditMode>
+  </PropertyGroup>
+
 </Project>
--- a/src/NuGet.Core/NuGet.Build.Tasks/NuGet.props
+++ b/src/NuGet.Core/NuGet.Build.Tasks/NuGet.props
@@ -38,14 +38,4 @@ Copyright (c) .NET Foundation. All rights reserved.
     <CentralPackageVersionsFileImported>true</CentralPackageVersionsFileImported>
   </PropertyGroup>
 
-  <!--
-    Setup default NuGet property values.
-    Visual Studio's project property page requires defaults to be set to inform customers what the default values are.
-  -->
-  <PropertyGroup>
-    <NuGetAudit Condition=" '$(NuGetAudit)' == '' ">true</NuGetAudit>
-    <NuGetAuditLevel Condition=" '$(NuGetAuditLevel)' == '' ">low</NuGetAuditLevel>
-    <NuGetAuditMode Condition=" '$(NuGetAuditMode)' == '' ">direct</NuGetAuditMode>
-  </PropertyGroup>
-
 </Project>
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
@@ -304,14 +304,14 @@ namespace NuGet.Commands
                     });
                 }
 
-                AuditUtility.EnabledValue enableAudit = AuditUtility.ParseEnableValue(
+                bool auditEnabled = AuditUtility.ParseEnableValue(
                     _request.Project.RestoreMetadata?.RestoreAuditProperties?.EnableAudit,
                     _request.Project.FilePath,
                     _logger);
-                telemetry.TelemetryEvent[AuditEnabled] = AuditUtility.GetString(enableAudit);
-                if (enableAudit != AuditUtility.EnabledValue.ExplicitOptOut)
+                telemetry.TelemetryEvent[AuditEnabled] = auditEnabled ? "enabled" : "disabled";
+                if (auditEnabled)
                 {
-                    await PerformAuditAsync(enableAudit, graphs, telemetry, token);
+                    await PerformAuditAsync(graphs, telemetry, token);
                 }
 
                 telemetry.StartIntervalMeasure();
@@ -477,11 +477,10 @@ namespace NuGet.Commands
             }
         }
 
-        private async Task PerformAuditAsync(AuditUtility.EnabledValue enableAudit, IEnumerable<RestoreTargetGraph> graphs, TelemetryActivity telemetry, CancellationToken token)
+        private async Task PerformAuditAsync(IEnumerable<RestoreTargetGraph> graphs, TelemetryActivity telemetry, CancellationToken token)
         {
             telemetry.StartIntervalMeasure();
             var audit = new AuditUtility(
-                enableAudit,
                 _request.Project.RestoreMetadata.RestoreAuditProperties,
                 _request.Project.FilePath,
                 graphs,
--- a/test/EndToEnd/tests/NetCoreProjectTest.ps1
+++ b/test/EndToEnd/tests/NetCoreProjectTest.ps1
@@ -139,6 +139,8 @@ function Test-NetCoreConsoleAppRebuildDoesNotDeleteCacheFile {
 }
 
 function Test-NetCoreVSandMSBuildNoOp {
+    [SkipTest('https://github.com/NuGet/Home/issues/13003')]
+    param ()
 
     # Arrange
     $project = New-NetCoreConsoleApp ConsoleApp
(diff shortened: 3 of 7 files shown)


src/NuGet.Core/NuGet.Build.Tasks/NuGet.props
```diff
@@ -38,4 +38,14 @@ Copyright (c) .NET Foundation. All rights reserved.
     <CentralPackageVersionsFileImported>true</CentralPackageVersionsFileImported>
   </PropertyGroup>
 
+  <!--
+    Setup default NuGet property values.
+    Visual Studio's project property page requires defaults to be set to inform customers what the default values are.
+  -->
+  <PropertyGroup>
+    <NuGetAudit Condition=" '$(NuGetAudit)' == '' ">true</NuGetAudit>
+    <NuGetAuditLevel Condition=" '$(NuGetAuditLevel)' == '' ">low</NuGetAuditLevel>
```
zivkan: I was chatting with @drewnoakes about this a month ago, and I agree that this is a little confusing without a description/instructions.  Should we take this opportunity to also change the property name to `NuGetAuditMinSeverity`? This will make it more intuitive.  If `NuGetAuditLevel` is set, we'll use that value, for backwards compatibility.

@JonDouglas what do you think?

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -304,14 +304,14 @@ public async Task<RestoreResult> ExecuteAsync(CancellationToken token)
...
                 }
 
-                AuditUtility.EnabledValue enableAudit = AuditUtility.ParseEnableValue(
+                bool auditEnabled = AuditUtility.ParseEnableValue(
                     _request.Project.RestoreMetadata?.RestoreAuditProperties?.EnableAudit,
                     _request.Project.FilePath,
                     _logger);
-                telemetry.TelemetryEvent[AuditEnabled] = AuditUtility.GetString(enableAudit);
-                if (enableAudit != AuditUtility.EnabledValue.ExplicitOptOut)
+                telemetry.TelemetryEvent[AuditEnabled] = auditEnabled ? "enabled" : "disabled";
```
zivkan, line 311: I discussed this telemetry change with @jebriede, as we don't believe that knowing "implicitly enabled" vs "explicitly enabled" is sufficiently useful.  @jebriede also gave me feedback that he thinks introducing this breaking change in the telemetry will be better off in the long term, rather than reporting "ExplicitOptIn" and "ExplicitOptOut" as the only two values we end up reporting.

If explicit vs implicit opt in is useful, we can add it, by setting a second MSBuild property (with underscore property name). In order to implement this, we'd need to change the VS Nomination code (and therefore also create a pull request on dotnet/project-system), and `LegacyPackageReferenceProject`. I'm not entirely sure if we'd need to also change the DGSpec reader/writer.

We only report telemetry in VS, so the additional MSBuild property wouldn't be needed for CLI scenarios, however, we don't have any other VS specific MSBuild properties. And if the VS scenario does require changes to the DGSpec, then the CLI restore, VS no-op restore scenerio will break. So, if we'd want to keep everything consistent, we'd also need to change static and non-static graph restore. I think the DGSpec changes will make nuget.exe restore current, but there's a chance that nuget.exe might need a change too.

So, as we can see, there's a high cost to keeping this information. Therefore, we need a good justification for why it would be useful if we want to keep it.

src/NuGet.Core/NuGet.Build.Tasks/NuGet.props
JonDouglas: I don't think we change the name until it is GA and people give us their feedback. I do think the name proposed here could be helpful, but I also don't think we have any data to justify another property name today. We can re-visit that conversation later in my opinion. Docs and property page tooltips should be able to help this in the time being.

src/NuGet.Core/NuGet.Build.Tasks/NuGet.props
```diff
@@ -38,4 +38,14 @@ Copyright (c) .NET Foundation. All rights reserved.
     <CentralPackageVersionsFileImported>true</CentralPackageVersionsFileImported>
   </PropertyGroup>
 
+  <!--
+    Setup default NuGet property values.
+    Visual Studio's project property page requires defaults to be set to inform customers what the default values are.
+  -->
+  <PropertyGroup>
```
nkolev92: I'm guessing these defaults in NuGet.props instead of NuGet.targets is just "the right thing to do". 
Any other considerations? Anything related to ordering?

nkolev92 (dismissed):

src/NuGet.Core/NuGet.Build.Tasks/NuGet.props
zivkan: Honestly, I didn't think too hard about it when writing it.  But, moving the properties to the target file will allow customers to use their own `'$(Property)' == ''` condition in project files, in case that helps them.  Do you (or anyone else reading this comment) have a suggestion?

jeffkl (dismissed):

src/NuGet.Core/NuGet.Build.Tasks/NuGet.props
nkolev92: The way I see it, the empty check is more appropriate for the SDK changes (including nuget) rather than the project file. 

If you just set that property to something, it'd be respected so isn't that all that matters? 

Can you think of scenarios where someone would want to do that?
I can't really come up with anything right now.

jebriede (dismissed):

src/NuGet.Core/NuGet.Build.Tasks/NuGet.props
zivkan: I talked to Drew, and [project-system uses MSBuild's `DefaultValueSourceLocation.AfterContext`](https://github.com/dotnet/project-system/blob/3d36410406c3803df33ca7c0fec40d67de9c048a/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/PageItemRuleProvider.cs#L33) to detect when a project changes a property from its default value, so that the project properties page can show that it's a non-default value.

So, I pushed an update moving the properties to NuGet.targets

nkolev92 (dismissed):

test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/Utility/AuditUtilityTests.cs
```diff
@@ -32,34 +32,31 @@ public class AuditUtilityTests
...
+    [InlineData("invalid", true, true)]
+    public void ParseEnableValue_WithValue_ReturnsExpected(string input, bool expected, bool expectLog = false)
     {
         // Arrange
-        AuditUtility.EnabledValue expectedValue = (AuditUtility.EnabledValue)Enum.Parse(typeof(AuditUtility.EnabledValue), expected);
         string projectPath = "my.csproj";
-        TestLogger? logger = expectedValue == AuditUtility.EnabledValue.Invalid
-            ? new TestLogger()
-            : null;
+        TestLogger? logger = expectLog ? new TestLogger() : null;
```
donnie-msft: If the test cares about a scenario where 1 message is logged, shouldn't the test go ahead and assert that nothing is logged in the other scenarios?
I feel like the test code to determine whether to pass in a logger is really working against the test coverage, and I'm not seeing the benefit to that.

test/EndToEnd/tests/NetCoreProjectTest.ps1
```diff
@@ -163,6 +165,8 @@ function Test-NetCoreVSandMSBuildNoOp {
 }
 
 function Test-NetCoreTargetFrameworksVSandMSBuildNoOp {
+    [SkipTest('https://github.com/NuGet/Home/issues/13003')]
```
zivkan, line 168: I updated the PR description, but this is failing because I forgot to update dotnet/project-system to pass `NuGetAuditMode` in nomination:

* https://github.com/NuGet/Home/issues/13003

We can either:

1. Merge this PR with the skip and then I'll need to undo the skip once project-system merged my PR and inserts into VS
2. Wait until project-system merges and inserts the fix, then I need to update this PR to remove these skips.

nkolev92: personally, I'm fine with merging the skip.
Let's just get it done :)

nkolev92 (approved):

jeffkl (approved):

PR #5471, opened 2023-10-23
Stop using HttpClientHandler.PreAuthenticate

### LINKED ISSUE #12961: HTTP 401 after some time in VS
### NuGet Product Used

### Product Version

NuGet 6.8

### Worked before?

NuGet 6.6

### Impact

### Repro Steps & Context

Use a package source that needs credentials, and use a credential manager to provide those credentials. If VS is open for long enough (longer than token validity that the cred provider gave to NuGet), then all HTTP requests will result in HTTP 401, because HttpClient will keep using the old credential, and never refresh to the newer token that the cred provider gave.

The root cause is that NuGet started using `HttpClientHandler.PreAuthenticate`, but there's a bug in the .NET runtime:

* https://github.com/dotnet/runtime/issues/93340

Note, that the Azure Artifacts cred provider currently gives a token with a long valid duration, so it's unlikely most customers will experience this, with Azure DevOps. However, if customers are using other credential providers, that return short lived tokens, the probability of encountering this problem is greatly increased.

### Verbose Logs

_No response_

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12961

Regression? Yes Last working version: 6.6

## Description

The .NET runtime has a bug when using `HttpClientHandler.PreAuthenticate` that prevents it from getting new credentials when a time-limited credential expires, and the server starts requiring a new credential. NuGet's credential provider can provide the new credential, but `HttpClient` won't use it as long as `PreAuthenticate` is `true`.

* https://github.com/dotnet/runtime/issues/93340

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception: We didn't have a test for the previous behaviour. Since we should re-introduce the functionality, I skipped the existing PreAuthenticate test, rather than deleting it, so it's easy to reactive once we have a new solution.
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpHandlerResourceV3Provider.cs
+++ b/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpHandlerResourceV3Provider.cs
@@ -54,7 +54,6 @@ namespace NuGet.Protocol
             {
                 Proxy = proxy,
                 AutomaticDecompression = (DecompressionMethods.GZip | DecompressionMethods.Deflate),
-                PreAuthenticate = true
             };
 
 #if IS_DESKTOP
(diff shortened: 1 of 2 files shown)


nkolev92 (approved):

jeffkl (approved):
😢

PR #5473, opened 2023-11-02
Run Component Governance task after restore (5.11.x)

zivkan:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2555

Regression? no

## Description

CG must run on all jobs in the pipeline, so when it's not run explicitly, it runs automatically as the last task in the job. Hence, I removed the condition so it's explicit on both the "RTM" and "nonRTM" jobs.  I also moved it to after restore, so that it does not detect the packages that our pipeline creates, since those are out of scope for what CG is intended for.

Finally, I noticed that we had an old "add apex feed" task, which is no longer relevant, so I deleted it at the same time.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/eng/pipelines/templates/Build_and_UnitTest.yml
+++ b/eng/pipelines/templates/Build_and_UnitTest.yml
@@ -85,13 +85,6 @@ steps:
       }
   condition: " and(succeeded(), eq(variables['IsOfficialBuild'], 'true'))"
 
-- task: NuGetCommand@2
-  displayName: "Add Apex Feed Source"
-  inputs:
-    command: "custom"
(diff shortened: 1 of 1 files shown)


nkolev92 (approved):
Do we need these changes in other servicing branches as well?

We need it in dev as well that way future branches don't have the same issues.

nkolev92:
@zivkan 

Seems like that wasn't enough. https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=8636807&view=results. Func test on windows are still having the same issue.

PR #5494, opened 2023-11-06
Control embedding Restore files in binlog

### LINKED ISSUE #12957: RestoreTask: Control whether to embed files in the binlog
### NuGet Product(s) Involved

### The Elevator Pitch

Need some properties on RestoreTask to control whether to embed files in the binlog, and possibly filter out .dgspec.json files.

We've found that the embedded .nuget.dgspec.json files are huge, and our binlog is 1 GB in size. .dgspec.json files total 3.5 GB for us uncompressed. Would be nice to be able to control whether to include .dgspec files here:

https://github.com/NuGet/NuGet.Client/blob/5ce332fb834feea1f6c974d17456c8c9d5504e09/src/NuGet.Core/NuGet.Build.Tasks/RestoreTask.cs#L216

Not sure how fine grained to make it. Just one property on the task to embed dgspec files, default to true? (No effect if no binlog present). Or 4 properties to control each of the four types of files? Or a main property to control all files?

Embedding files is useful and a good default, but for large builds embedding dgspec file specifically renders the binlog almost unusable.

_No response_

KirillOsenkov:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12957

## Description

This changes allows users to specify what level of files to embed in the MSBuild binary log.

| Value | Files to embed |
|---|---|
| `0` or `false` | Do not embed any files
| `2` | Embed everything (dgspec, assets file, g.props, and g.targets) |
| Any other value | Embed assets file, g.props, and g.targets |

Users can specify what get embedded via the MSBuild property `RestoreEmbedFilesInBinlog`:

For example, to disable the functionality:
```
/p:RestoreEmbedFilesInBinlog=false
```

Or to opt into embedding everything:
```
/p:RestoreEmbedFilesInBinlog=2
```

Any other value embeds just the default (assets file, g.props, and g.targets)

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Build.Tasks/RestoreTask.cs
+++ b/src/NuGet.Core/NuGet.Build.Tasks/RestoreTask.cs
@@ -89,16 +89,12 @@ namespace NuGet.Build.Tasks
         public ITaskItem[] EmbedInBinlog { get; set; }
 
         /// <summary>
-        /// Controls whether to embed files produced by the Restore task in the binlog.
-        /// Defaults to true.
+        /// Gets or sets a value indicating whether to embed files produced by restore in the MSBuild binary logger.
+        /// 0 = Nothing
+        /// 1 = Assets file, g.props, and g.targets
+        /// 2 = dgspec, assets file, g.props, and g.targets
         /// </summary>
-        public bool EmbedRestoreFilesInBinlog { get; set; } = true;
-
-        /// <summary>
-        /// Controls whether to embed .nuget.dgspec.json files in the binlog.
-        /// Defaults to true.
-        /// </summary>
-        public bool EmbedDGSpecInBinlog { get; set; } = true;
+        public string EmbedFilesInBinlog { get; set; }
 
         public override bool Execute()
         {
@@ -200,8 +196,10 @@ namespace NuGet.Build.Tasks
         /// <returns>If the MSBuildBinaryLoggerEnabled environment variable is set, returns the paths to NuGet files to embed in the binlog, otherwise returns <see cref="Array.Empty{T}" />.</returns>
         private ITaskItem[] GetFilesToEmbedInBinlog(DependencyGraphSpec dependencyGraphSpec)
         {
-            if (!string.Equals(Environment.GetEnvironmentVariable("MSBUILDBINARYLOGGERENABLED"), bool.TrueString, StringComparison.OrdinalIgnoreCase)
-                || !EmbedRestoreFilesInBinlog)
+            // Determines what the user wants embedded in the binary log where 0 or false disables embedding anything, 2 embeds everything, and 1 or true embeds just the assets file, g.props, and g.targets.
+            int embedInBinlogSelection = BuildTasksUtility.GetFilesToEmbedInBinlogValue(EmbedFilesInBinlog);
+
+            if (embedInBinlogSelection == 0)
             {
                 return Array.Empty<ITaskItem>();
             }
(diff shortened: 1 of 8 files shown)


KirillOsenkov:
I have tested this by building this PR version of NuGet and inserting into a boostrap version of MSBuild that I also built myself.

I've verified that the two new properties work as expected when passing via command-line: `/p:EmbedDGSpecInBinlog=false`.

However I've found that setting `EmbedDGSpecInBinlog` to false in `Directory.Build.props` doesn't work when restoring a solution, because it doesn't get imported during Restore. Here are the files imported during Restore of the .sln.metaproj:
image

I did find that it works as expected when I create a `Directory.Solution.props` at the root of the repo:
```xml
<Project>
    <PropertyGroup>
        <EmbedDGSpecInBinlog>false</EmbedDGSpecInBinlog>
    </PropertyGroup>
</Project>
```

It then gets imported during Restore:
image

It also does work as expected when restoring a specific .csproj, `Directory.Build.props` does get imported in that case.

Another thing I've learned is that even if you set `EmbedRestoreFilesInBinlog` to false, `.nuget.g.props` and `.nuget.g.targets` files will be embedded anyway by virtue of all projects being imported automatically being embedded. We also still embed `project.assets.json` files by virtue of this logic:

https://github.com/dotnet/sdk/pull/16840/files

I've verified that `EmbedProjectAssetsFile` still works fine to control that embedding.

So technically to turn off embedding `project.assets.json` you need to set both `EmbedRestoreFilesInBinlog` as well as `EmbedProjectAssetsFile`, because they are now embedded from two separate places. I think this is fine for now, and we can later perhaps just delete the logic from https://github.com/dotnet/sdk/pull/16840.

However this PR works fine for the main goal: to be able to turn off .dgspec files specifically and keep other files, which is the behavior I want for those large binlogs.

KirillOsenkov:
Another minor unexpected thing is that setting properties in `Directory.Solution.props` will only affect the solution restore. Setting `EmbedProjectAssetsFile` in `Directory.Solution.props` will have no effect as it won't be imported by individual project builds.

nkolev92 (dismissed):

KirillOsenkov:
@jeffkl this continues to be super painful for us, my proposal is to merge something in to get us unblocked for the next release, and if you like, we can still find the proper nice solution later. I'm worried we'll miss the next train and I'll have to live with 900 MB binlogs for another several months.

jeffkl:
@NuGet/nuget-client I have pushed a commit to update this pull request based on a conversation I had with the author.  Please review.

src/NuGet.Core/NuGet.Build.Tasks/RestoreTask.cs
```diff
@@ -88,6 +88,14 @@ public class RestoreTask : Microsoft.Build.Utilities.Task, ICancelableTask, IDis
         [Output]
         public ITaskItem[] EmbedInBinlog { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether to embed files produced by restore in the MSBuild binary logger.
+        /// 0 = Nothing
+        /// 1 = Assets file, g.props, and g.targets
+        /// 2 = dgspec, assets file, g.props, and g.targets
+        /// </summary>
+        public string EmbedFilesInBinlog { get; set; }
```
nkolev92, line 97: did you consider an enum?

donnie-msft (dismissed):
I'm a little uncertain about "N/A" for documentation since there's a table in this PR description. Would customers benefit from updating one or more of the following?
- https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-command-line-reference?view=vs-2022#switches-for-loggers
- https://learn.microsoft.com/en-us/nuget/consume-packages/package-restore#restore-by-using-msbuild
- https://learn.microsoft.com/en-us/nuget/reference/msbuild-targets#restore-target

jeffkl:
> e of the following?

I think this feature is only useful for us when a customer is reporting a problem and we need the dgspec.  We would have them restore again with `/p:RestoreEmbedFilesInBinlog=2` and then the binlog would contain more files.  The dgspec isn't really useful for users and only for us.  The assets file is already being [embedded by the .NET SDK](https://github.com/dotnet/sdk/pull/16840/files) and the `.g.props` and `g.targets` are also being embedded automatically if the exist since they are imported.  I'm going to send a follow-up PR to MSBuild to turn that off for restore because the files that get embedded are what are already on disk and not what NuGet generates.  

But my opinion is that we don't need to document it at this time since its really only on when you enable the binary logger and these files have already been embedded by default for 2+ years.

src/NuGet.Core/NuGet.Build.Tasks/RestoreTask.cs
jeffkl: I could parse the value as an enum but I wanted to just support true/false plus one more value.  So I settled on:

0/false = off
1/true = default
2 = everything

I could still parse the value into an enum later instead of an integer if you prefer. But I didn't want users to have to specify values like `None`, `Default`, `All`.

nkolev92: GIven that this is a debug thing, we can always change it later. I'm not worried about it.

nkolev92 (approved):

KirillOsenkov (commented):
:+1:

KirillOsenkov:
OMG this makes me so happy! I can’t wait to consume this! Thanks!

PR #5498, opened 2023-11-08
Don't log task inputs and outputs when binary logger is enabled

### LINKED ISSUE #10696: GetReferenceNearestTargetFrameworkTask should not log params when LogTaskInputs is set
When MSBuild sets LogTaskInputs, there's no need to specifically log task inputs or outputs as MSBuild will take care of that.

See here:
https://github.com/NuGet/NuGet.Client/blob/58fd22bfc941a6d5cc9404f41a583efb4dfeeaf0/src/NuGet.Core/NuGet.Build.Tasks/GetReferenceNearestTargetFrameworkTask.cs#L64

See https://github.com/dotnet/msbuild/issues/6305 for guidance. Otherwise there's duplication, allocations, log bloat:
image

KirillOsenkov:
Stop logging task inputs and outputs now that the binary logger makes getting this information much easier.

This results in significant savings to binlog size and potentially faster restore as well (and less memory allocations).

Fixes https://github.com/NuGet/Home/issues/10696

See related:
 * https://github.com/dotnet/msbuild/issues/6305
 * https://github.com/dotnet/msbuild/pull/6803

## Bug

Fixes:

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception - Only removed test assertions
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


KirillOsenkov:
@jeffkl @nkolev92

jeffkl:
@nkolev92 do you think these logging statements are useful anymore at all?  Or can we just rely on the binary logger to have this information and remove the logging calls to these messages all together?

nkolev92:
I think those messages may have lived out their usefulness now that binary logs are so common.

jeffkl:
> I think those messages may have lived out their usefulness now that binary logs are so common.

Okay thanks, I agree.  I've removed all task input/output logging

jeffkl (dismissed):

jeffkl (approved):

test/NuGet.Core.Tests/NuGet.Build.Tasks.Test/WarnForInvalidProjectsTaskTests.cs
```diff
@@ -131,7 +128,6 @@ public void WarnForInvalidProjectsTask_AllValidItemsVerifyNoWarnings()
 
             testLogger.Warnings.Should().Be(0);
             testLogger.Errors.Should().Be(0);
-            testLogger.DebugMessages.Count.Should().Be(2);
```
nkolev92, line 134: Should these be 0 instead of removing them completely?

nkolev92 (approved):

KirillOsenkov (commented):
Looks great to me!

zivkan (approved):

test/NuGet.Core.Tests/NuGet.Build.Tasks.Test/WarnForInvalidProjectsTaskTests.cs
jeffkl: I don't think it matters unless we really care to ensure that a debug message isn't accidentally added.  The code diff clearly shows that I removed the existing ones so it seems odd to me to check that they were removed.  Do you feel strongly that we should update the test instead of remove this assertion?

nkolev92: it's one of those that if I was writing the code originally, I would've changed it to 0. 

I don't really have a strong reason to add it back though. I'm ok with it as is.

PR #5499, opened 2023-11-08
Fix conditional package updating in update all scenarios, add more tests for conditional package operations

### LINKED ISSUE #4681: PM UI multi targeting experience is incomplete - support for updating and uninstalling conditional package versions
If you have a package reference in multiple conditional item groups, then PM UI does not update all occurrences of the package.

Repro Steps- 
1. Start with - 
```
  <PropertyGroup>
    <TargetFrameworks>netstandard1.0;netstandard1.3;net45;uap10.0;portable-win81+wpa81;win8;wpa81;wp8</TargetFrameworks>
  </PropertyGroup>
</Project>
```

2. Install NUnit v3.4.0 - 

```
  <PropertyGroup>
    <TargetFrameworks>netstandard1.0;netstandard1.3;net45;uap10.0;portable-win81+wpa81;win8;wpa81;wp8</TargetFrameworks>
  </PropertyGroup>
  <ItemGroup Condition="'$(TargetFramework)' == 'win8'">
    <PackageReference Include="NUnit">
      <Version>3.4.0</Version>
    </PackageReference>
  </ItemGroup>
  <ItemGroup Condition="'$(TargetFramework)' == 'wp8'">
    <PackageReference Include="NUnit">
      <Version>3.4.0</Version>
    </PackageReference>
  </ItemGroup>
  <ItemGroup Condition="'$(TargetFramework)' == 'uap10.0'">
    <PackageReference Include="NUnit">
      <Version>3.4.0</Version>
    </PackageReference>
  </ItemGroup>
  <ItemGroup Condition="'$(TargetFramework)' == 'wpa81'">
    <PackageReference Include="NUnit">
      <Version>3.4.0</Version>
    </PackageReference>
  </ItemGroup>
  <ItemGroup Condition="'$(TargetFramework)' == 'net45'">
    <PackageReference Include="NUnit">
      <Version>3.4.0</Version>
    </PackageReference>
  </ItemGroup>
  <ItemGroup Condition="'$(TargetFramework)' == 'portable-win81+wpa81'">
    <PackageReference Include="NUnit">
      <Version>3.4.0</Version>
    </PackageReference>
  </ItemGroup>
</Project>
```

3. Update NUnit to v3.6.0 through the PM UI - 

```
  <PropertyGroup>
    <TargetFrameworks>netstandard1.0;netstandard1.3;net45;uap10.0;portable-win81+wpa81;win8;wpa81;wp8</TargetFrameworks>
  </PropertyGroup>
  <ItemGroup Condition="'$(TargetFramework)' == 'win8'">
    <PackageReference Include="NUnit">
      <Version>3.4.0</Version>
    </PackageReference>
  </ItemGroup>
  <ItemGroup Condition="'$(TargetFramework)' == 'wp8'">
    <PackageReference Include="NUnit">
      <Version>3.4.0</Version>
    </PackageReference>
  </ItemGroup>
  <ItemGroup Condition="'$(TargetFramework)' == 'uap10.0'">
    <PackageReference Include="NUnit">
      <Version>3.4.0</Version>
    </PackageReference>
  </ItemGroup>
  <ItemGroup Condition="'$(TargetFramework)' == 'wpa81'">
    <PackageReference Include="NUnit">
      <Version>3.4.0</Version>
    </PackageReference>
  </ItemGroup>
  <ItemGroup Condition="'$(TargetFramework)' == 'net45'">
    <PackageReference Include="NUnit">
      <Version>3.4.0</Version>
    </PackageReference>
  </ItemGroup>
  <ItemGroup Condition="'$(TargetFramework)' == 'portable-win81+wpa81'">
    <PackageReference Include="NUnit">
      <Version>3.4.0</Version>
    </PackageReference>
  </ItemGroup>
  <ItemGroup>
    <PackageReference Include="NUnit" Version="3.6.0" />
  </ItemGroup>
</Project>
```

After the third step, we add an unconditional reference to NUnit v3.6.0 without removing existing conditional package references. This leads to a customer bug - vsfeedback/812089

Related: https://github.com/NuGet/Home/issues/5521

nkolev92:
## Bug

Fixes: https://github.com/NuGet/Home/issues/4681

## Description

Conditional package updating was previously fixed in https://github.com/NuGet/NuGet.Client/commit/fed7892af6c079015431389d7cf7a7f630e22bbd. 

This PR fixes the `update all` scenario, ensuring all packages are updated instead of just 1. 
This PR also improves the tests cases for the scenarios.

Some technical details to help with the review of this PR: 

- BuildIntegratedProjectAction is always the result for a project action on a PackageReference project, even if there are multiple packages being updated. Note that this is not something we're adding in this PR. 
- For each package being updated as part of an action, the compatible frameworks should be consider with that package id in mind. That allows, Newtonsoft.Json to be updated for net472, but NuGet.Common to be updated for net5.0.
- BuildIntegratedProjectAction types are not expected to be used outside of the NuGet.PackageManagement, and given how overexposed are APIs are, I'm adding these as internal. This is probably not  the best design, but it avoids adding future work for us. An alternative would be to complete break the type.
- Added nullable annotation for BuildIntegratedProjectAction 

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added, will also add manual tests.
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.PackageManagement/BuildIntegration/BuildIntegratedProjectAction.cs
+++ b/src/NuGet.Core/NuGet.PackageManagement/BuildIntegration/BuildIntegratedProjectAction.cs
@@ -1,10 +1,11 @@
+#nullable enable
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using NuGet.Commands;
-using NuGet.Frameworks;
 using NuGet.Packaging.Core;
 using NuGet.ProjectManagement;
 using NuGet.ProjectModel;
@@ -38,13 +39,18 @@ namespace NuGet.PackageManagement
         /// <summary>
         /// Original user actions.
         /// </summary>
+        [Obsolete("The internal ActionAndContextList property should be used.")]
         public IReadOnlyList<NuGetProjectAction> OriginalActions { get; }
 
         /// <summary>
         /// The context necessary for installing a package.
         /// </summary>
+        [Obsolete("The internal ActionAndContextList property should be used.")]
         public BuildIntegratedInstallationContext InstallationContext { get; }
 
+        internal IReadOnlyList<(NuGetProjectAction, BuildIntegratedInstallationContext)> ActionAndContextList { get; }
+
+        [Obsolete("This type is not expected to be created externally.")]
         public BuildIntegratedProjectAction(NuGetProject project,
             PackageIdentity packageIdentity,
             NuGetProjectActionType nuGetProjectActionType,
@@ -57,6 +63,7 @@ namespace NuGet.PackageManagement
         {
         }
 
+        [Obsolete("This type is not expected to be created externally.")]
         public BuildIntegratedProjectAction(NuGetProject project,
             PackageIdentity packageIdentity,
             NuGetProjectActionType nuGetProjectActionType,
@@ -65,9 +72,14 @@ namespace NuGet.PackageManagement
             IReadOnlyList<SourceRepository> sources,
             IReadOnlyList<NuGetProjectAction> originalActions,
             BuildIntegratedInstallationContext installationContext,
-            VersionRange versionRange)
+            VersionRange? versionRange)
             : base(packageIdentity, nuGetProjectActionType, project, sourceRepository: null, versionRange)
         {
+            if (project == null)
+            {
+                throw new ArgumentNullException(nameof(project));
+            }
+
             if (packageIdentity == null)
             {
                 throw new ArgumentNullException(nameof(packageIdentity));
@@ -90,12 +102,12 @@ namespace NuGet.PackageManagement
 
             if (originalActions == null)
             {
-                throw new ArgumentNullException(nameof(sources));
+                throw new ArgumentNullException(nameof(originalActions));
             }
 
             if (installationContext == null)
             {
-                throw new ArgumentNullException(nameof(sources));
+                throw new ArgumentNullException(nameof(installationContext));
             }
 
             OriginalLockFile = originalLockFile;
@@ -104,6 +116,63 @@ namespace NuGet.PackageManagement
             Sources = sources;
             OriginalActions = originalActions;
             InstallationContext = installationContext;
+            ActionAndContextList = originalActions.Select(e => (e, installationContext)).ToList();
+        }
+
+        internal BuildIntegratedProjectAction(NuGetProject project,
+            PackageIdentity packageIdentity,
+            NuGetProjectActionType nuGetProjectActionType,
+            LockFile originalLockFile,
+            RestoreResultPair restoreResultPair,
+            IReadOnlyList<SourceRepository> sources,
+            IReadOnlyList<(NuGetProjectAction, BuildIntegratedInstallationContext)> originalActionsAndInstallationContexts,
+            VersionRange versionRange)
+            : base(packageIdentity, nuGetProjectActionType, project, sourceRepository: null, versionRange)
+        {
+            if (project == null)
+            {
+                throw new ArgumentNullException(nameof(project));
+            }
+
+            if (packageIdentity == null)
+            {
+                throw new ArgumentNullException(nameof(packageIdentity));
+            }
+
+            if (originalLockFile == null)
+            {
+                throw new ArgumentNullException(nameof(originalLockFile));
+            }
+
+            if (restoreResultPair == null)
+            {
+                throw new ArgumentNullException(nameof(restoreResultPair));
+            }
+
+            if (sources == null)
+            {
+                throw new ArgumentNullException(nameof(sources));
+            }
+
+            if (originalActionsAndInstallationContexts == null)
+            {
+                throw new ArgumentNullException(nameof(originalActionsAndInstallationContexts));
+            }
+
+            if (originalActionsAndInstallationContexts.Count < 1)
+            {
+                throw new ArgumentException("Must contain at least 1 element.", nameof(originalActionsAndInstallationContexts));
+            }
+
+            OriginalLockFile = originalLockFile;
+            RestoreResult = restoreResultPair.Result;
+            RestoreResultPair = restoreResultPair;
+            Sources = sources;
+            ActionAndContextList = originalActionsAndInstallationContexts;
+#pragma warning disable CS0618 // Type or member is obsolete
+            OriginalActions = originalActionsAndInstallationContexts.Select(e => e.Item1).ToList();
+            InstallationContext = originalActionsAndInstallationContexts[0].Item2;
+#pragma warning restore CS0618 // Type or member is obsolete
         }
 
         public IReadOnlyList<NuGetProjectAction> GetProjectActions()
@@ -117,12 +186,12 @@ namespace NuGet.PackageManagement
 
                 foreach (var package in removed)
                 {
-                    actions.Add(NuGetProjectAction.CreateUninstallProjectAction(package, Project));
+                    actions.Add(CreateUninstallProjectAction(package, Project));
                 }
 
                 foreach (var package in added)
                 {
-                    actions.Add(NuGetProjectAction.CreateInstallProjectAction(package, sourceRepository: null, project: Project));
+                    actions.Add(CreateInstallProjectAction(package, sourceRepository: null, project: Project));
                 }
             }
 
--- a/src/NuGet.Core/NuGet.PackageManagement/BuildIntegration/BuildIntegratedProjectAction.cs
+++ b/src/NuGet.Core/NuGet.PackageManagement/BuildIntegration/BuildIntegratedProjectAction.cs
@@ -1,6 +1,7 @@
-#nullable enable
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+#nullable enable
 
 using System;
 using System.Collections.Generic;
--- a/src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
+++ b/src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
@@ -3115,8 +3115,8 @@ namespace NuGet.PackageManagement
                 if (nuGetProjectActions.Length == 1 &&
                     firstAction.NuGetProjectActionType == NuGetProjectActionType.Install &&
                     !restoreResult.Result.Success &&
-                    successfulFrameworks.Any() &&
-                    unsuccessfulFrameworks.Any() &&
+                    successfulFrameworks.Count > 0 &&
+                    unsuccessfulFrameworks.Count > 0 &&
                     !PackageSpecOperations.HasPackage(originalPackageSpec, firstAction.PackageIdentity.Id))
                 {
                     updatedPackageSpec = originalPackageSpec.Clone();
--- a/src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
+++ b/src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
@@ -24,6 +24,7 @@ using NuGet.ProjectModel;
 using NuGet.Protocol;
 using NuGet.Protocol.Core.Types;
 using NuGet.Resolver;
+using NuGet.Shared;
 using NuGet.Versioning;
 
 namespace NuGet.PackageManagement
@@ -833,10 +834,7 @@ namespace NuGet.PackageManagement
 
             if (packageIdentities.Count == 0 && packageId == null)
             {
-                // Update-Package  all
-
-                //TODO: need to consider whether Update ALL simply does nothing for Build Integrated projects
-
+                // Update All
                 var lowLevelActions = new List<NuGetProjectAction>();
 
                 foreach (var installedPackage in projectInstalledPackageReferences)
@@ -3063,8 +3061,7 @@ namespace NuGet.PackageManagement
 
             foreach (var buildIntegratedProject in buildIntegratedProjects)
             {
-                var nuGetProjectActions = nugetProjectActionsLookup[buildIntegratedProject.MSBuildProjectPath];
-                var nuGetProjectActionsList = nuGetProjectActions;
+                NuGetProjectAction[] nuGetProjectActions = nugetProjectActionsLookup[buildIntegratedProject.MSBuildProjectPath];
                 var updatedPackageSpec = updatedNugetPackageSpecLookup[buildIntegratedProject.MSBuildProjectPath];
                 var originalPackageSpec = originalNugetPackageSpecLookup[buildIntegratedProject.MSBuildProjectPath];
                 var originalLockFile = lockFileLookup[buildIntegratedProject.MSBuildProjectPath];
@@ -3090,20 +3087,36 @@ namespace NuGet.PackageManagement
                     .Distinct()
                     .ToList();
 
-                var successfulFrameworks = allFrameworks
-                    .Except(unsuccessfulFrameworks)
-                    .ToList();
+                var originalFrameworks = updatedPackageSpec
+                    .TargetFrameworks
+                    .ToDictionary(x => x.FrameworkName, x => x.TargetAlias);
 
-                var firstAction = nuGetProjectActionsList[0];
+                List<(NuGetProjectAction, BuildIntegratedInstallationContext)> projectActionsAndInstallationContexts = new(nuGetProjectActions.Length);
+
+                foreach (var action in nuGetProjectActions)
+                {
+                    PackageSpec referencePackageSpec = null;
+                    referencePackageSpec = action.NuGetProjectActionType switch
+                    {
+                        NuGetProjectActionType.Install or NuGetProjectActionType.Update => restoreResult.Result.LockFile.PackageSpec,
+                        NuGetProjectActionType.Uninstall => originalPackageSpec,
+                        _ => throw new InvalidOperationException("Unknown NuGetProjectActionType"),
+                    };
+                    BuildIntegratedInstallationContext installationContext = CreateInstallationContextForPackageId(action.PackageIdentity.Id, referencePackageSpec, unsuccessfulFrameworks, originalFrameworks);
+                    projectActionsAndInstallationContexts.Add((action, installationContext));
+                }
+
+                var firstInstallationAndProjectContext = projectActionsAndInstallationContexts[0];
+                var firstAction = firstInstallationAndProjectContext.Item1;
+                var successfulFrameworks = firstInstallationAndProjectContext.Item2.SuccessfulFrameworks.AsList();
 
                 // If the restore failed and this was a single package install, try to install the package to a subset of
                 // the target frameworks.
-                if (nuGetProjectActionsList.Length == 1 &&
+                if (nuGetProjectActions.Length == 1 &&
                     firstAction.NuGetProjectActionType == NuGetProjectActionType.Install &&
                     !restoreResult.Result.Success &&
                     successfulFrameworks.Any() &&
                     unsuccessfulFrameworks.Any() &&
-                    // Exclude upgrades, for now we take the simplest case.
                     !PackageSpecOperations.HasPackage(originalPackageSpec, firstAction.PackageIdentity.Id))
                 {
                     updatedPackageSpec = originalPackageSpec.Clone();
@@ -3134,16 +3147,6 @@ namespace NuGet.PackageManagement
                     await MSBuildRestoreUtility.ReplayWarningsAndErrorsAsync(restoreResult.Result.LockFile?.LogMessages, logger);
                 }
 
-                // Build the installation context
-                var originalFrameworks = updatedPackageSpec
-                    .TargetFrameworks
-                    .ToDictionary(x => x.FrameworkName, x => x.TargetAlias);
-
-                var installationContext = new BuildIntegratedInstallationContext(
-                    successfulFrameworks,
-                    unsuccessfulFrameworks,
-                    originalFrameworks);
-
                 InstallationCompatibility.EnsurePackageCompatibility(
                     buildIntegratedProject,
                     pathContext,
@@ -3166,8 +3169,7 @@ namespace NuGet.PackageManagement
                     originalLockFile,
                     restoreResult,
                     sources.ToList(),
-                    nuGetProjectActionsList,
-                    installationContext,
+                    projectActionsAndInstallationContexts,
                     versionRange);
 
                 result.Add(new ResolvedAction(buildIntegratedProject, nugetProjectAction));
@@ -3187,6 +3189,35 @@ namespace NuGet.PackageManagement
             TelemetryActivity.EmitTelemetryEvent(actionTelemetryEvent);
 
             return result;
+        }
+
+        /// <summary>
+        /// Build a context for package given package id.
+        /// The "successful" frameworks are the ones that contain the package and are not part of the failed list.
+        /// The "unsuccessful" frameworks are the ones that never had the package or are part of the unsuccessful list.
+        /// </summary>>
+        internal static BuildIntegratedInstallationContext CreateInstallationContextForPackageId(string packageIdentityId, PackageSpec packageSpec, List<NuGetFramework> unsuccessfulFrameworks, Dictionary<NuGetFramework, string> originalFrameworks)
+        {
+            var frameworksWithResultingPackage = packageSpec
+                                .TargetFrameworks
+                                .Where(e => e.Dependencies.Any(a => a.Name == packageIdentityId))
+                                .Select(e => e.FrameworkName)
+                                .Distinct();
+
+            var frameworksWithoutResultingPackage = packageSpec
+                .TargetFrameworks
+                .Where(e => !e.Dependencies.Any(a => a.Name == packageIdentityId))
+                .Select(e => e.FrameworkName)
+                .Distinct();
+
+            var successfulFrameworksWithPackage = frameworksWithResultingPackage
+                .Except(unsuccessfulFrameworks)
+                .ToList();
+
+            return new BuildIntegratedInstallationContext(
+                successfulFrameworksWithPackage,
+                unsuccessfulFrameworks.Union(frameworksWithoutResultingPackage).Distinct(),
+                originalFrameworks);
         }
 
         /// <summary>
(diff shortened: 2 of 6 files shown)


src/NuGet.Core/NuGet.PackageManagement/BuildIntegration/BuildIntegratedProjectAction.cs
```diff
@@ -1,10 +1,11 @@
+#nullable enable
```
zivkan: I think the normal convention is that the license header stays at the top of the file, so this compiler directive should probably be moved down.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
```diff
@@ -3090,20 +3087,36 @@ var projectUniqueNamesForBuildIntToUpdate
                     .Distinct()
                     .ToList();
 
-                var successfulFrameworks = allFrameworks
-                    .Except(unsuccessfulFrameworks)
-                    .ToList();
+                var originalFrameworks = updatedPackageSpec
+                    .TargetFrameworks
+                    .ToDictionary(x => x.FrameworkName, x => x.TargetAlias);
```
zivkan, line 3092: This looks like it'll cause problems when TFMs as aliases gets implemented. Can framework name and target alias be swapped, so the alias is the key, not the value?

Looks like `BuildIntegratedInstallationContext` already exists, not created by this PR, and already uses NuGetFramework as the dictionary key, so my request is out of scope for this PR. 😭

zivkan (dismissed):

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
nkolev92: yeah it's one of many things that'll need tweaked :)

src/NuGet.Core/NuGet.PackageManagement/BuildIntegration/BuildIntegratedProjectAction.cs
nkolev92: Fixed, ready for rereview.

src/NuGet.Core/NuGet.PackageManagement/BuildIntegration/BuildIntegratedProjectAction.cs
```diff
@@ -57,6 +64,7 @@ public class BuildIntegratedProjectAction : NuGetProjectAction
         {
         }
 
+        [Obsolete("This type is not expected to be created externally.")]
```
kartheekp-ms, line 67: Just to clarify, this [constructor](https://learn.microsoft.com/dotnet/api/system.obsoleteattribute.-ctor?view=net-7.0#system-obsoleteattribute-ctor(system-string)) triggers a warning instead of an error. If we want to raise an error, then we need to use a different constructor.
```cs
[Obsolete("This type is not expected to be created externally.", error: true)]
```

src/NuGet.Core/NuGet.PackageManagement/BuildIntegration/BuildIntegratedProjectAction.cs
```diff
@@ -104,6 +117,63 @@ public class BuildIntegratedProjectAction : NuGetProjectAction
...
+            }
+
+            if (originalActionsAndInstallationContexts == null)
+            {
+                throw new ArgumentNullException(nameof(originalActionsAndInstallationContexts));
+            }
+
+            if (originalActionsAndInstallationContexts.Count < 1)
+            {
+                throw new ArgumentException("Must contain at least 1 element.", nameof(originalActionsAndInstallationContexts));
```
kartheekp-ms, line 165: I think we should localize this exception message.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
```diff
@@ -3090,20 +3087,36 @@ var projectUniqueNamesForBuildIntToUpdate
...
+                List<(NuGetProjectAction, BuildIntegratedInstallationContext)> projectActionsAndInstallationContexts = new(nuGetProjectActions.Length);
+
+                foreach (var action in nuGetProjectActions)
+                {
+                    PackageSpec referencePackageSpec = null;
+                    referencePackageSpec = action.NuGetProjectActionType switch
+                    {
+                        NuGetProjectActionType.Install or NuGetProjectActionType.Update => restoreResult.Result.LockFile.PackageSpec,
+                        NuGetProjectActionType.Uninstall => originalPackageSpec,
+                        _ => throw new InvalidOperationException("Unknown NuGetProjectActionType"),
```
kartheekp-ms, line 3103: Should we localize this exception message?

src/NuGet.Core/NuGet.PackageManagement/BuildIntegration/BuildIntegratedProjectAction.cs
zivkan: that would make it a breaking change. I mean, warnings are too if customers use treat warnings as errors, but at least in that case they're shotting themselves in the foot. There have been customers asking us questions about the NuGet.PackageManagement package before, so we know that some people are using it to install packages in their own apps.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
zivkan: IMO, this one won't benefit. The switch statement already exhausts every defined `NuGetProjectActionType` enum value, but because it's valid, from a compiler's point of view, to do `(NuGetProjectActionType)1234`, the compiler forces you to still have a default value, just in case. But that would be a programming error, one that we wouldn't expect a customer to ever see.

src/NuGet.Core/NuGet.PackageManagement/BuildIntegration/BuildIntegratedProjectAction.cs
nkolev92: This is an internal constructor. 
If this happens, it's our mistake.

src/NuGet.Core/NuGet.PackageManagement/BuildIntegration/BuildIntegratedProjectAction.cs
```diff
@@ -104,6 +117,63 @@ public class BuildIntegratedProjectAction : NuGetProjectAction
...
+
+            OriginalLockFile = originalLockFile;
+            RestoreResult = restoreResultPair.Result;
+            RestoreResultPair = restoreResultPair;
+            Sources = sources;
+            ActionAndContextList = originalActionsAndInstallationContexts;
+#pragma warning disable CS0618 // Type or member is obsolete
+            OriginalActions = originalActionsAndInstallationContexts.Select(e => e.Item1).ToList();
+            InstallationContext = originalActionsAndInstallationContexts[0].Item2;
+#pragma warning restore CS0618 // Type or member is obsolete
```
kartheekp-ms, line 176: I'm really curious to understand why we should access the obsolete members here. We already assigned the new property `ActionAndContextList` with the correct value in this constructor, which isn't obsolete. And have we done the same in the constructor that is marked as obsolete in this PR.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
nkolev92: Yep, these are exceptions that basically tell us that we haven't implemented all the scenarios correctly.

src/NuGet.Core/NuGet.PackageManagement/BuildIntegration/BuildIntegratedProjectAction.cs
nkolev92: The constructor is new, there's a new property, but that  doesn't mean that the old contract should be broken. 
Normally, if this wasn't a public API, I'd remove the old properties, but that's not possible without a breaking change currently. 
So until then, we make sure that the type is correctly populated.

src/NuGet.Core/NuGet.PackageManagement/BuildIntegration/BuildIntegratedProjectAction.cs
nkolev92: I agree with @zivkan. 

The error option is a really big hammer. We haven't really gotten a lot of feedback on our deprecations to change our approach imo.

src/NuGet.Core/NuGet.PackageManagement/BuildIntegration/BuildIntegratedProjectAction.cs
kartheekp-ms: The two properties `OriginalActions` and `InstallationContext` are currently unused in the product code. If a caller uses the new constructor, we'll assign a value to the newly introduced property in this PR, `ActionAndContextList`, which the product code references. However, if the caller uses the old constructor, we'll generate the `ActionAndContextList` property using `OriginalActions` and `InstallationContext`. This new property will then be used throughout the rest of the product code. Therefore, I'm uncertain about the necessity of assigning values to the old or obsolete properties in the new constructor.

src/NuGet.Core/NuGet.PackageManagement/BuildIntegration/BuildIntegratedProjectAction.cs
```diff
@@ -65,9 +73,14 @@ public class BuildIntegratedProjectAction : NuGetProjectAction
...
             IReadOnlyList<NuGetProjectAction> originalActions,
             BuildIntegratedInstallationContext installationContext,
-            VersionRange versionRange)
+            VersionRange? versionRange)
             : base(packageIdentity, nuGetProjectActionType, project, sourceRepository: null, versionRange)
         {
+            if (project == null)
+            {
+                throw new ArgumentNullException(nameof(project));
+            }
```
kartheekp-ms, line 82: How about modifying the obsolete constructor to call the new one? This approach would allow us to centralize the null checks in one place.

test/NuGet.Core.Tests/NuGet.PackageManagement.Test/NuGetPackageManagerUtilityTests.cs
```diff
@@ -0,0 +1,153 @@
...
+                }";
+            Dictionary<NuGetFramework, string> originalFrameworks = new()
+            {
+                { FrameworkConstants.CommonFrameworks.Net472, "net472" },
+                { FrameworkConstants.CommonFrameworks.Net50, "net50" }
+            };
+
+            var originalPackageSpec = ProjectTestHelpers.GetPackageSpecWithProjectNameAndSpec("project", @"C:\", referenceSpec);
+
+            var buildIntegrationInstallationContext = NuGetPackageManager.CreateInstallationContextForPackageId(packageIdentityId: "a", originalPackageSpec, unsuccessfulFrameworks: new() { FrameworkConstants.CommonFrameworks.Net472 }, originalFrameworks);
```
kartheekp-ms, line 120: Why did we choose net472 as the unsuccessful framework for this test?

test/NuGet.Core.Tests/NuGet.PackageManagement.Test/NuGetPackageManagerUtilityTests.cs
```diff
@@ -0,0 +1,153 @@
...
+                }";
+            Dictionary<NuGetFramework, string> originalFrameworks = new()
+            {
+                { FrameworkConstants.CommonFrameworks.Net472, "net472" },
+                { FrameworkConstants.CommonFrameworks.Net50, "net50" }
+            };
+
+            var originalPackageSpec = ProjectTestHelpers.GetPackageSpecWithProjectNameAndSpec("project", @"C:\", referenceSpec);
+
+            var buildIntegrationInstallationContext = NuGetPackageManager.CreateInstallationContextForPackageId(packageIdentityId: "a", originalPackageSpec, unsuccessfulFrameworks: new(), originalFrameworks);
```
kartheekp-ms, line 43: Adding a blank line after action would clarify that line 41 is an action followed by assertions in line 43. This comment applies to multiple tests in this class.

test/NuGet.Core.Tests/NuGet.PackageManagement.Test/PackageReferenceBasedNuGetPackageManagerTests.cs
```diff
@@ -225,6 +256,102 @@ public async Task PreviewUpdatePackagesAsync_WithMultiTargettedPackageReferenceP
...
+                b100,
+                b200);
+
+            // Main Act
+            var result = (await nuGetPackageManager.PreviewUpdatePackagesAsync(
+                new List<NuGetProject> { buildIntegratedProject },
+                new ResolutionContext(DependencyBehavior.Lowest, false, false, VersionConstraints.None),
+                new TestNuGetProjectContext(),
+                sourceRepositoryProvider.GetRepositories(),
+                sourceRepositoryProvider.GetRepositories(),
```
kartheekp-ms, line 304: Is it required to pass the same value for both the parameters?

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
```diff
@@ -3090,20 +3087,36 @@ var projectUniqueNamesForBuildIntToUpdate
...
+                var successfulFrameworks = firstInstallationAndProjectContext.Item2.SuccessfulFrameworks.AsList();
 
                 // If the restore failed and this was a single package install, try to install the package to a subset of
                 // the target frameworks.
-                if (nuGetProjectActionsList.Length == 1 &&
+                if (nuGetProjectActions.Length == 1 &&
                     firstAction.NuGetProjectActionType == NuGetProjectActionType.Install &&
                     !restoreResult.Result.Success &&
                     successfulFrameworks.Any() &&
                     unsuccessfulFrameworks.Any() &&
```
kartheekp-ms: Out of scope for this PR but accessing `.Count` is better in terms of performance than `.Any()`.

```cs
                    successfulFrameworks.Count > 0 &&
                    unsuccessfulFrameworks.Count > 0 &&
```

src/NuGet.Core/NuGet.PackageManagement/BuildIntegration/BuildIntegratedProjectAction.cs
nkolev92: We don't know who/how people use our code unfortunately.
If people say we're not worried about that, I can remove it, but I was erring on the cautious side.

src/NuGet.Core/NuGet.PackageManagement/BuildIntegration/BuildIntegratedProjectAction.cs
nkolev92: The null checks are different across the constructors. We'd end up with an additional constructor.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
```diff
@@ -3090,20 +3087,36 @@ var projectUniqueNamesForBuildIntToUpdate
...
+                        _ => throw new InvalidOperationException("Unknown NuGetProjectActionType"),
+                    };
+                    BuildIntegratedInstallationContext installationContext = CreateInstallationContextForPackageId(action.PackageIdentity.Id, referencePackageSpec, unsuccessfulFrameworks, originalFrameworks);
+                    projectActionsAndInstallationContexts.Add((action, installationContext));
+                }
 
-                var firstAction = nuGetProjectActionsList[0];
+                var firstInstallationAndProjectContext = projectActionsAndInstallationContexts[0];
+                var firstAction = firstInstallationAndProjectContext.Item1;
+                var successfulFrameworks = firstInstallationAndProjectContext.Item2.SuccessfulFrameworks.AsList();
```
kartheekp-ms, line 3111: If there are multiple items in the collection `projectActionsAndInstallationContexts`, I am curious to learn why we only access first item in the collection.

test/NuGet.Core.Tests/NuGet.PackageManagement.Test/NuGetPackageManagerUtilityTests.cs
nkolev92: It's part of the test name. 
A package is installed conditionally, but the update to that conditional package failed, hence net472.

test/NuGet.Core.Tests/NuGet.PackageManagement.Test/PackageReferenceBasedNuGetPackageManagerTests.cs
nkolev92: They are combined with an uniqueness check so not really relevant. 

I think this part of the code is not the most interesting part of the changes though. It's fairly standard test setup code.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
nkolev92: I'm not changing that code.

The way this was originally implemented was that a single PR update event would lead to a single BuildIntegratedProjectAction, which could mean more than 1 write to the project file through update all and similar commands. 
 
 BuildIntegratedProjectAction extends NuGetProjectAction which requires a single action, so that's why the first one was used.
Changing that is difficult without breaking changes.

nkolev92:
@kartheekp-ms 
Since you're adding comments a considerable time apart, can you please let me know when you're done adding comments so I can address the feedback in a single commit?

Note that I've marked comments where my next commit will resolve the feedback as `Resolved` already.

kartheekp-ms:
I am also wondering if we need a new (or update existing) apex test to assert the new behavior.

kartheekp-ms:
> @kartheekp-ms Since you're adding comments a considerable time apart, can you please let me know when you're done adding comments so I can address the feedback in a single commit?
> 
> Note that I've marked comments where my next commit will resolve the feedback as `Resolved` already.

I think I am done for now. Please feel free to push new commit(s).

May be out of scope for this PR. If I click on "Update" a package version in the "Installed" tab, then the versions in the csproj are not updated correctly when there are multiple target frameworks. I can open a separate bug for this scenario.

nkolev92:
> I am also wondering if we need a new (or update existing) apex test to assert the new behavior.

There's no multi targeted templates, so that makes Apex tests extremely difficult to add. .NET templates in previews also require constant source updates right now as packages are not always published to nuget.org, or a source that we control. Created an issue for that https://github.com/NuGet/Client.Engineering/issues/2589.
I'll ask the vendors to add manual tests once the whole thing is complete.

nkolev92:
> May be out of scope for this PR. If I click on "Update" a package version in the "Installed" tab, then the versions in the csproj are not updated correctly when there are multiple target frameworks. I can open a separate bug for this scenario.

With my changes? 
It'd be great if you can give me a setup. 

If you want to consider it out scope, you can just create an issue and assign it to me, and I can validate that afterwards.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
```diff
@@ -3189,6 +3191,35 @@ var projectUniqueNamesForBuildIntToUpdate
...
+                .Select(e => e.FrameworkName)
+                .Distinct();
+
+            var successfulFrameworksWithPackage = frameworksWithResultingPackage
+                .Except(unsuccessfulFrameworks)
+                .ToList();
+
+            return new BuildIntegratedInstallationContext(
+                successfulFrameworksWithPackage,
+                unsuccessfulFrameworks.Union(frameworksWithoutResultingPackage).Distinct(),
```
donnie-msft, line 3219: I find it easier to debug separate statements, and I don't expect it's much compilation benefit to inline here.

donnie-msft (approved):

kartheekp-ms:
> With my changes?
> It'd be great if you can give me a setup.

Yes Nikolche. Here are the repro steps.
 
- Clone this PR branch locally
- Build the repo
- Launch the experimental instance
- Create a new project
- Edit the csproj manually
```xml

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFrameworks>net472;net8.0</TargetFrameworks>  
    <Nullable>enable</Nullable>
  </PropertyGroup>
    <ItemGroup Condition="'$(TargetFramework)' == 'net472'">
	    <PackageReference Include="Newtonsoft.Json">
		    <Version>13.0.1</Version>
	    </PackageReference>
    </ItemGroup>
    <ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
	    <PackageReference Include="Newtonsoft.Json">
		    <Version>13.0.2</Version>
	    </PackageReference>
    </ItemGroup>
</Project>
```
- Open NuGet Package Manager UI for the project
- Navigate to "Installed" tab
- Click on the update icon with text 13.0.3
image
- As you can see in the below screenshot, package version changed only for "net472" target framework.
image

> If you want to consider it out scope, you can just create an issue and assign it to me, and I can validate that afterwards.

Please take a look at the repro steps and let me know your feedback so that we can plan accordingly.

nkolev92:
@kartheekp-ms 
Turns out project-system is having issues with where the condition is: 

For example, I tested a similar scenario as yours, but the conditions are on the package reference item instead, and that does work!

repro


I'll create an issue for follow up. That change may not be on the nuget side.

nkolev92:
Actually, nevermind you have 2 frameworks in your test and I have 3. 
The problem is that we don't consider the fact that all frameworks have a conditional reference. 

I'll create a follow up for that.
That *may* be something we can detect, but I'm not sure we can have a perfect solution.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
nkolev92: This is fairly easy to debug if you put a breakpoint outside of the method where you'd get the already created installation context. 
It's not really for a compilation benefit, I just thought this was readable, and there was no need to add extra lines.

kartheekp-ms:
> I'll create a follow up for that.
> That may be something we can detect, but I'm not sure we can have a perfect solution.

Has the follow-up issue been created?

nkolev92:
No, I was waiting for approval on this first. 

I created it now https://github.com/NuGet/Home/issues/13034

src/NuGet.Core/NuGet.PackageManagement/BuildIntegration/BuildIntegratedProjectAction.cs
kartheekp-ms: Can we extract all the common null checks into another method and invoke that method from the constructor?

Here's a suggested implementation of the `ValidateArguments` method:

```cs
private void ValidateArguments(NuGetProject project, PackageIdentity packageIdentity, LockFile originalLockFile, RestoreResultPair restoreResultPair, IReadOnlyList<SourceRepository> sources, object actions, BuildIntegratedInstallationContext? installationContext = null)
{
    if (project == null)
        throw new ArgumentNullException(nameof(project));
    if (originalLockFile == null)
        throw new ArgumentNullException(nameof(originalLockFile));
    if (restoreResultPair == null)
        throw new ArgumentNullException(nameof(restoreResultPair));
    if (sources == null)
        throw new ArgumentNullException(nameof(sources));
}
```

We can add null checks that are specific to a particular constructor within that constructor itself, after validating the common arguments.

We can remove the following null check from the constructors in this class because the [base constructor takes care of that](https://github.com/NuGet/NuGet.Client/blob/d85c6ea4588c5b4f2b4a4d7bf8073837fb308f58/src/NuGet.Core/NuGet.PackageManagement/NuGetProjectAction.cs#L49-L52):

```cs
if (packageIdentity == null)
    throw new ArgumentNullException(nameof(packageIdentity));
```

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
kartheekp-ms: Do you think there is any technical debt in this code path that requires a closer look at the API? If yes, feel free to create a new technical debt issue.

> The way this was originally implemented was that a single PR update event would lead to a single BuildIntegratedProjectAction, which could mean more than 1 write to the project file through update all and similar commands.

Will there be any consistency issues if there are multiple writes to the same project file?

test/NuGet.Core.Tests/NuGet.PackageManagement.Test/NuGetPackageManagerUtilityTests.cs
kartheekp-ms: Thanks for the clarification; I now understand. Essentially, the test is to verify that if an update fails for the `net472` framework, then the collection of unsuccessful frameworks should include both `net5`, since it lacks dependencies on package 'a', and 'net472', because the update failed for some reason. Is my understanding correct?

test/NuGet.Core.Tests/NuGet.PackageManagement.Test/PackageReferenceBasedNuGetPackageManagerTests.cs
kartheekp-ms: I agree with your feedback. Maybe we could have passed an empty list (Enumerable.Empty<SourceRepository>) because there are no secondary sources in this test.

kartheekp-ms (approved):
There are some unresolved comments in this pull request, primarily the ones we discussed earlier. Feel free to review them, but I don't want to block this PR until those comments are addressed.

test/NuGet.Core.Tests/NuGet.PackageManagement.Test/NuGetPackageManagerUtilityTests.cs
nkolev92: yes

test/NuGet.Core.Tests/NuGet.PackageManagement.Test/PackageReferenceBasedNuGetPackageManagerTests.cs
nkolev92: Nearly all of our tests do the same thing, so I don't think it warrants a change. This is what the PM UI would pass in for PR anyways.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
nkolev92: > Do you think there is any technical debt in this code path that requires a closer look at the API? If yes, feel free to create a new technical debt issue.

Maybe?

1 action meaning more packages has permeated throughout the APIs and I'm not really sure if that's necessarily the best design, but it's also not a frequent code path, but the public APIs just make things nearly impossible to change.
I'm gonna look into the whole tech debt thing, and I'll consider this then.

src/NuGet.Core/NuGet.PackageManagement/BuildIntegration/BuildIntegratedProjectAction.cs
nkolev92: I can clean that up in a future PR.

PR #5500, opened 2023-11-09
Include path to GPF package on `InvalidDataException` from `nupkg.metadata` files

### LINKED ISSUE #12997: Bad data in `nupkg.metadata` should include package id or path in error message
### NuGet Product Used

dotnet.exe, MSBuild.exe, NuGet.exe, Visual Studio Package Management UI, Visual Studio Package Manager Console, NuGet SDK

### Product Version

N/A

### Worked before?

_No response_

### Impact

### Repro Steps & Context

When performing a restore, the Global Packages Folder (GPF) is commonly used to find packages and read their included `nupkg.metadata` file. When doing so, if that file is empty or contains invalid metadata, this will cause restore to fail.

The output from restore does not indicate which package it found with invalid metadata. 
A message states: `Found invalid data while decoding.`
Enabling `SHOW_STACK_TRACE` shows the origination is from reading `nupkg.metadata`, but not which file caused it.
Detailed/Diagnostic verbosity does not help either.

image

### Verbose Logs

_No response_

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12997

## Description

When any `InvalidDataException` is thrown, rethrow the exception with the message set as the `path` to the `nupkg.metadata` file. By including the path in the output, it should be easier for customers to self-address the issue, or for the team to more quickly understand the output from restore when customers report this problem to us.

_Screenshots show behavior **before** (left side) / **after** (right side). 
NuGet Version: 6.7.0.127 was used as the testing control / before behavior_

#### Show Stack Trace OFF
image

#### Show Stack Trace ON
image

#### Show Stack Trace OFF

image

#### Show Stack Trace ON

image

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Packaging/NupkgMetadata/NupkgMetadataFileFormat.cs
+++ b/src/NuGet.Core/NuGet.Packaging/NupkgMetadata/NupkgMetadataFileFormat.cs
@@ -57,6 +57,11 @@ namespace NuGet.Packaging
                     Strings.Error_LoadingHashFile,
                     path, ex.Message));
 
+                if (!string.IsNullOrEmpty(path) && ex is InvalidDataException)
+                {
+                    throw new InvalidDataException(message: path, innerException: ex);
+                }
+
                 throw;
             }
         }
--- a/test/NuGet.Core.Tests/NuGet.Packaging.Test/NupkgMetadataFileFormatTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Packaging.Test/NupkgMetadataFileFormatTests.cs
@@ -107,6 +107,8 @@ namespace NuGet.Packaging.Test
             // Assert
             Assert.Equal(1, logger.Messages.Count);
             Assert.Contains(path, exception.Message);
+            Assert.NotNull(exception.InnerException);
+            Assert.IsType<InvalidDataException>(exception.InnerException);
         }
 
         [Fact]
@@ -126,6 +128,8 @@ namespace NuGet.Packaging.Test
             // Assert
             Assert.Equal(1, logger.Messages.Count);
             Assert.Contains(path, exception.Message);
+            Assert.NotNull(exception.InnerException);
+            Assert.IsType<InvalidDataException>(exception.InnerException);
         }
 
         [Fact]
--- a/test/NuGet.Core.Tests/NuGet.Packaging.Test/NupkgMetadataFileFormatTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Packaging.Test/NupkgMetadataFileFormatTests.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.IO;
+using System.Linq;
 using Moq;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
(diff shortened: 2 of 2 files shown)


nkolev92 (approved):

src/NuGet.Core/NuGet.Packaging/NupkgMetadata/NupkgMetadataFileFormat.cs
```diff
@@ -57,6 +57,11 @@ public static NupkgMetadataFile Read(TextReader reader, ILogger log, string path
                     Strings.Error_LoadingHashFile,
                     path, ex.Message));
 
+                if (!string.IsNullOrEmpty(path) && ex is InvalidDataException)
+                {
+                    throw new InvalidDataException(message: path, innerException: ex);
```
zivkan, line 62: Why only when the exception is InvalidDataException?  What if a different process had the file locked, preventing sharing with other processes? Or if some crazy process delete the file after NuGet has performed the `if File.Exists(path)` check, but before NuGet actually reads the file?  Isn't the filename still interesting to customers?

zivkan, line 62: I don't think that having the exception message be just the filename without anything else is the best experience. Consider ArgumentNullException in the BCL. The exception message isn't just the argument name. Similarly, `FileNotFoundException`'s exception message is "Could not find file 'c:\does not exist.txt'.", not just the filename.

```suggestion
                    throw new InvalidDataException(message: $"Error reading 'path'.", innerException: ex);
```

However, since that message gets surfaced to customers, it probably needs to go into strings.resx.

donnie-msft: Are you saying catch `Exception` and rethrow all of them as inner exceptions with a message? I don't really have time this sprint to test every scenario, like what will the output be from your example, and whether the logger already shows something or we should rethrow exceptions or not. Therefore, if that's necessary, I'll probably close this PR and let someone plan the work to encompass all scenarios.

donnie-msft: I see. I felt that a path and "invalid data" as an exception was pretty straight forward. I can introduce some localized strings.
Your example states there was an error reading the file. Is that intuitive to mean that the file's data was corrupt/not expected, or the contents were empty? 
There are many scenarios to the invalid data exception, so I assume that's why .NET didn't bake a sentence into this exception like they did your examples.

zivkan: Yes, I'd just catch all exceptions. Regardless of what the root cause (file locked, invalid json content, file does not exist), I believe it's useful for customers to know what file had an issue, so it's not intuitive to me why we'd only report filename on specific exception types. Also, this code isn't trying to recover from an exception, it' just immediately throwing a new exception with additional information (retaining the original exception as an inner exception), so I don't think the general guideline of "catch specific exceptions" is very convincing here.

Also consider that `NupkgMetadataFileFormat.Read` is a public API that customers might use in their own apps. While we care mostly about our own tools experience, I think it would be a mistake to ignore the NuGet Client SDK experience.

nkolev92: Is the invalid data exception always the one thrown at line 49?

donnie-msft: > Is the invalid data exception always the one thrown at line 49?

Yes. `throw new InvalidDataException();`

donnie-msft:
Closing since there are concerns with this being too specific, and possibly causing SDK issues.

PR #5502, opened 2023-11-10
Test Fix: PackageRestoreManager uses `ConcurrentBag` for restored packages

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2442

## Description

- Locally running this test, I experienced a different test failure than CI. I'm not certain if this fixes the original issue (that _restore failures_  was less than the expectation). Locally, I found that the restored packages were sometimes empty (found <null> for restoredPackages[0]):

  - image

- To fix that, I've tried using a `ConcurrentBag`, and after over 100 runs with no failures, I believe this is worth doing.

- If this doesn't fix the restore failures assertion on CI, then I'll file a new issue. 
  - I introduced `ITestOutputHelper` to output the invocation of those delegate methods, so hopefully that will help understand the issue better.
  - image

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/test/NuGet.Core.Tests/NuGet.PackageManagement.Test/PackageRestoreManagerTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.PackageManagement.Test/PackageRestoreManagerTests.cs
@@ -15,18 +15,26 @@ using NuGet.Test.Utility;
 using NuGet.Versioning;
 using Test.Utility;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace NuGet.Test
 {
     public class PackageRestoreManagerTests
     {
         private readonly List<PackageIdentity> Packages = new List<PackageIdentity>
-            {
-                new PackageIdentity("jQuery", new NuGetVersion("1.4.4")),
-                new PackageIdentity("jQuery", new NuGetVersion("1.6.4")),
-                new PackageIdentity("jQuery.Validation", new NuGetVersion("1.13.1")),
-                new PackageIdentity("jQuery.UI.Combined", new NuGetVersion("1.11.2"))
-            };
+        {
+            new PackageIdentity("jQuery", new NuGetVersion("1.4.4")),
+            new PackageIdentity("jQuery", new NuGetVersion("1.6.4")),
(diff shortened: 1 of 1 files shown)


test/NuGet.Core.Tests/NuGet.PackageManagement.Test/PackageRestoreManagerTests.cs
```diff
@@ -353,7 +368,7 @@ public async Task Test_PackageRestoreFailure_WithRaisedEvents()
                 nuGetPackageManager.PackageExistsInPackagesFolder(testPackage2).Should().BeFalse();
 
                 // Act
-                await packageRestoreManager.RestoreMissingPackagesInSolutionAsync(testSolutionManager.SolutionDirectory,
+                PackageRestoreResult result = await packageRestoreManager.RestoreMissingPackagesInSolutionAsync(testSolutionManager.SolutionDirectory,
```
donnie-msft, line 371: This isn't required, but I saw it returned something and prefer seeing that. I could use a discard. Other opinions?

zivkan: I never investigated why it's different in some projects, but some projects cause a compiler warning (or maybe it's just an info/underline?) when variables are not used.

But if you add a `result.Should().NotBeNull()` then it'll be used in the test, and available during debugging for inspection.

zivkan (approved):

jeffkl (approved):

test/NuGet.Core.Tests/NuGet.PackageManagement.Test/PackageRestoreManagerTests.cs
donnie-msft: Let me merge this and see how it helps the flakiness, and I'll do that in another PR.

PR #5505, opened 2023-11-11
[Draft] Add Owner and Contributors to PM UI Packages List

### LINKED ISSUE #12501: Swap authors for owners in Visual Studio Package Manager UI
### NuGet Product(s) Involved

### The Elevator Pitch

Currently in the Visual Studio Package Manager UI, we show the `authors` field which is a comma delimited list that is commonly used to help a developer know who "authored" the package.

This concept is a bit outdated as NuGet.org provides a `owners` field in the search service that provides an **up-to-date maintainer list that can be used as a primary and if a server doesn't provide it, can fallback on `authors`**. This is not to be confused with the local [`owners` nuspec field](https://learn.microsoft.com/en-us/nuget/reference/nuspec#owners).

https://learn.microsoft.com/en-us/nuget/api/search-query-service-resource#search-result

Some further details as pointed out by @zivkan might be that we would only display owners for sources that support it and are fetched from the remote and ignore some nuspec metadata in the global packages folder. 

This will help reduce confusion of a field that is commonly used to impersonate notable entities or have changed in ownership in the years since it was published where the original author is still listed.

MicrosoftTeams-image (3)

Related: https://github.com/NuGet/Home/issues/10260

_No response_

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12501

## Description

In the Packages List, a distinction is made between package owner and author:

- **"by:"** refers to the **owner**
  - Always shown when available
  - I added the colon (:) to for parity with nuget.org
  - The prefix reserved indicator will still be shown to the left, just like it is on nuget.org. Note that prefix reservation is not required for an owner to be specified in the metadata. (ex: [NuGet Gallery | GraphQL.NewtonsoftJson 7.6.1](https://www.nuget.org/packages/GraphQL.NewtonsoftJson))

- **"contributors:"** refers to the **authors**
  - Only shown if owner is not available
  - Another option would be "authors:" but I wonder if is still misleading, since at a glance, it may seem that the author is the responsible party for owning the package. This confusion is exactly what we're trying to address, so I think a new term makes sense. GitHub uses the term "contributors" so I chose that.

The **Tooltip** when hovering over a package item **will show both** pieces of metadata (when available).
That is, after the package ID, both the owner "by: <owners>", and the authors, "contributors: <authors>" are shown, in that order. Owners are always first.

### Example of owners now shown in PM UI:
image

### Example of owners shown as "by: <owners> from nuget.org:
image

### Example of authors shown when owners are not available on the feed in PM UI:
image

## PR Checklist

- [ ] PR has a linked issue.
- [ ] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
@@ -301,6 +301,7 @@ namespace NuGet.PackageManagement.UI
                     Version = metadata.Identity.Version,
                     IconUrl = metadata.IconUrl,
                     Owner = metadata.Owners,
+                    IsTrustedOwner = !string.IsNullOrWhiteSpace(metadata.Owners),
                     Author = metadata.Authors,
                     DownloadCount = metadata.DownloadCount,
                     Summary = metadata.Summary,
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
@@ -300,6 +300,7 @@ namespace NuGet.PackageManagement.UI
                     Id = metadata.Identity.Id,
                     Version = metadata.Identity.Version,
                     IconUrl = metadata.IconUrl,
+                    Owner = metadata.Owners,
                     Author = metadata.Authors,
                     DownloadCount = metadata.DownloadCount,
                     Summary = metadata.Summary,
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.Designer.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.Designer.cs
@@ -1854,7 +1854,7 @@ namespace NuGet.PackageManagement.UI {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to contributors: {0}.
+        ///   Looks up a localized string similar to by {0}.
         /// </summary>
         public static string Text_ByAuthor {
             get {
@@ -1863,7 +1863,7 @@ namespace NuGet.PackageManagement.UI {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to by: {0}.
+        ///   Looks up a localized string similar to by {0}.
         /// </summary>
         public static string Text_ByOwner {
             get {
@@ -2300,6 +2300,15 @@ namespace NuGet.PackageManagement.UI {
         public static string ToolTip_TransitiveDependencyVersion {
             get {
                 return ResourceManager.GetString("ToolTip_TransitiveDependencyVersion", resourceCulture);
+            }
+        }
+        
+        /// <summary>
+        ///   Looks up a localized string similar to Owner according to nuget.org.
+        /// </summary>
+        public static string ToolTip_TrustedOwnerIndicator {
+            get {
+                return ResourceManager.GetString("ToolTip_TrustedOwnerIndicator", resourceCulture);
             }
         }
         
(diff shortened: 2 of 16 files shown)


donnie-msft:
@JonDouglas & @albarry4 - please take a look at my proposal before I open the draft for code review.
Jon, I think this addresses #12501 ? The idea is, if owners are available, show them in the list directly. But if owner isn't available, still show the author. I've taken some liberties here with wording, so thought I'd get your input as well as Allie's before I move forward in this direction. Thanks!

JonDouglas:
The current experience seems to combine both owners and authors? We should only do one and have the fallback as discussed in the issue. The proposal would be only showing nuget.org owners here, not adding a string saying "contributors" with the authors concatenated too. It is a good thought though, just a bit confusing to see names there twice. 

Can we surface nuget.org owners information in the details panel too? Right now, it still shows authors.

Also consider using a NuGet.org icon before or after these owners list so people can clearly distinguish these packages have owners on the central registry. This context will help them realize the difference of remote owners vs. local metadata authors.

Lastly, there's no need to introduce a "contributors" string here. Let's just keep the current experience of "by" and appropriately use the context.

donnie-msft:
@JonDouglas - thanks for reviewing. I need your answer to the last question about non-nuget.org package sources implementing Owners before I can decide on a path forward.

> The current experience seems to combine both owners and authors? We should only do one and have the fallback as discussed in the issue. 
> The proposal would be only showing nuget.org owners here, not adding a string saying "contributors" with the authors concatenated too. It is a good thought though, just a bit confusing to see names there twice.

Nothing is _concatenated_ in this **packages list**. The Tooltip is the only place it is showing both "by: <owners>" and "contributors: <authors>" (when available) for completeness. If you think the tooltip should show less information like the list does, I can remove it from the tooltip.

> Can we surface nuget.org owners information in the details panel too? Right now, it still shows authors.

Sure, in a different PR. Do you happen to have a design spec or other issue for that? #12501 seems to be written about the packages list.

> Also consider using a NuGet.org icon before or after these owners list so people can clearly distinguish these packages have owners on the central registry. This context will help them realize the difference of remote owners vs. local metadata authors.
> Lastly, there's no need to introduce a "contributors" string here. Let's just keep the current experience of "by" and appropriately use the context.

If "by" means different things, that to me still seems like it's easy to confuse which metadata I'm looking at.
The nuget icon might help with that, but I haven't seen any conclusion to zivkan 's comment here https://github.com/NuGet/Home/issues/12501#issuecomment-1478074091

JonDouglas:
> @JonDouglas - thanks for reviewing. I need your answer to the last question about non-nuget.org package sources implementing Owners before I can decide on a path forward.
> 
> > The current experience seems to combine both owners and authors? We should only do one and have the fallback as discussed in the issue.
> > The proposal would be only showing nuget.org owners here, not adding a string saying "contributors" with the authors concatenated too. It is a good thought though, just a bit confusing to see names there twice.
> 
> Nothing is _concatenated_ in this **packages list**. The Tooltip is the only place it is showing both "by: " and "contributors: " (when available) for completeness. If you think the tooltip should show less information like the list does, I can remove it from the tooltip.
> 
> > Can we surface nuget.org owners information in the details panel too? Right now, it still shows authors.
> 
> Sure, in a different PR. Do you happen to have a design spec or other issue for that? #12501 seems to be written about the packages list.
> 
> > Also consider using a NuGet.org icon before or after these owners list so people can clearly distinguish these packages have owners on the central registry. This context will help them realize the difference of remote owners vs. local metadata authors.
> > Lastly, there's no need to introduce a "contributors" string here. Let's just keep the current experience of "by" and appropriately use the context.
> 
> If "by" means different things, that to me still seems like it's easy to confuse which metadata I'm looking at. The nuget icon might help with that, but I haven't seen any conclusion to zivkan 's comment here [NuGet/Home#12501 (comment)](https://github.com/NuGet/Home/issues/12501#issuecomment-1478074091)

Yes, please remove it from the tooltip. Otherwise it will be very confusing.

#12501 should apply to this concept in general. Not just packages list.

For non-nuget sources, we keep the current behavior.

PR #5507, opened 2023-11-17
Fix missing transitive dependency tree nodes in some multi-targeting projects

### LINKED ISSUE #6832: Transitive dependency items not attached for all target frameworks when multi-targeting
Following the merge of PR #6712 (which fixed several bugs) we see some package and project reference nodes not having their children populated for multi-targeting projects. That PR changed the representation of the target we store on these items, which is now no longer understood by the code that attaches these children.

We are no longer parsing the `TargetFramework` dimension to produce a sanitised version for external consumption. Internally this works well enough as an opaque identifier, however it breaks when we communicate that information externally, in this case to the code in NuGet.Client that populates descendents of package and project references.

The fix is likely to capture whatever property values we need information from the `ConfiguredProject` and pass that throughout the system in our `TargetFramework` class. There are some rough edges here that will make this challenging and require careful thought.

drewnoakes:
The "target name" used in the project model can take different forms. For example, both `net8.0` and `.NETFramework,Version=v4.8.1` are valid values. The dependencies tree uses the "target alias" for its target nodes, in multi-target projects. This is because two different aliases can resolve to the same TFM, for example, yet attached items may differ.

This change digs the alias out of the lock file and uses that value throughout. It also renames a property and fixes the documentation on it so as to avoid confusion.

image

## Bug

Fixes https://github.com/dotnet/project-system/issues/6832

Regression? Last working version: I think we broke this in 16.9.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/AssetsFileDependenciesTreeSearchProvider.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/AssetsFileDependenciesTreeSearchProvider.cs
@@ -70,7 +70,7 @@ namespace NuGet.VisualStudio.SolutionExplorer
 
             foreach ((_, AssetsFileTarget target) in snapshot.DataByTarget)
             {
-                ConfiguredProject? configuredProject = await FindConfiguredProjectAsync(target.TargetFrameworkMoniker);
+                ConfiguredProject? configuredProject = await FindConfiguredProjectAsync(target.TargetAlias);
 
                 if (configuredProject == null)
                 {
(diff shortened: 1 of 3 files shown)


nkolev92 (approved):

drewnoakes:
Thanks @nkolev92. Is that test failure a concern? I can't see how it would be related. Will one of your team merge this and my other PR? I am unable to because of the test failures.

donnie-msft (approved):

donnie-msft:
> Thanks @nkolev92. Is that test failure a concern? I can't see how it would be related. Will one of your team merge this and my other PR? I am unable to because of the test failures.

Likely flaky test... it's rerunning now.

PR #5509, opened 2023-11-17
Add three apex tests for verify commonly used commands in PMC

CiciLi1:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2591

## Description

We want to add below three APEX test cases for verify common commands in PMC:

1. Verify that the package name and latest package version are displayed correctly in PMC after running the command `"find-package -ExactMatch"`.

2. Verify that the latest package versions is show correctly in PMC after running the command `“get-package -update”`.

3. Verify that the project name, project type and project full path are show correctly in PMC after running the command `“get-project”`.

## PR Checklist

- [X] PR has a linked issue.
- [X] Described changes

- **Tests**
  - [X] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [X] N/A 

- **Documentation**
  
  - **OR**
  - [X] N/A

### DIFF
--- a/test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetConsoleTestCase.cs
+++ b/test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetConsoleTestCase.cs
@@ -964,7 +964,7 @@ namespace NuGet.Tests.Apex
                 Assert.IsTrue(nugetConsole.IsMessageFoundInPMC("Install.ps1"), "The Install.ps1 script in TestProject was not executed when the Entityframework.sqlservercompact package was installed.");
             }
         }
-        
+
         [DataTestMethod]
         [DynamicData(nameof(GetNetCoreTemplates), DynamicDataSourceType.Method)]
         [Timeout(DefaultTimeout)]
@@ -1052,7 +1052,7 @@ namespace NuGet.Tests.Apex
                 PMCText.Should().Contain(testContext.Project.FullPath);
             }
         }
-        
+
         public static IEnumerable<object[]> GetNetCoreTemplates()
         {
             yield return new object[] { ProjectTemplate.NetCoreConsoleApp };
--- a/test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetConsoleTestCase.cs
+++ b/test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetConsoleTestCase.cs
@@ -991,8 +991,9 @@ namespace NuGet.Tests.Apex
                     nugetConsole.Execute($"find-package {PackageName} -ExactMatch");
 
                     // Assert
-                    Assert.IsTrue(nugetConsole.IsMessageFoundInPMC(PackageName), $"The package name {PackageName} doesn't show correctly in PMC.");
-                    Assert.IsTrue(nugetConsole.IsMessageFoundInPMC(v100), $"The package version {v100} doesn't show correctly in PMC.");
+                    string PMCText = nugetConsole.GetText();
+                    PMCText.Should().Contain(PackageName);
+                    PMCText.Should().Contain(v100);
                 }
             }
         }
@@ -1027,7 +1028,8 @@ namespace NuGet.Tests.Apex
                     nugetConsole.Execute("get-package -update");
 
                     // Assert
-                    Assert.IsTrue(nugetConsole.IsMessageFoundInPMC(v200), $"The latest package version {v200} doesn't show correctly in PMC.");
+                    string PMCText = nugetConsole.GetText();
+                    PMCText.Should().Contain(v200);
                 }
             }
         }
@@ -1048,9 +1050,10 @@ namespace NuGet.Tests.Apex
                 nugetConsole.Execute("Get-Project");
 
                 // Assert
-                Assert.IsTrue(nugetConsole.IsMessageFoundInPMC(testContext.Project.Name), "Fail to find the ProjectName in PMC");
-                Assert.IsTrue(nugetConsole.IsMessageFoundInPMC("C#"), "Fail to find the project type in PMC");
-                Assert.IsTrue(nugetConsole.IsMessageFoundInPMC(testContext.Project.FullPath), "Fail to find the project's FullName in PMC");
+                string PMCText = nugetConsole.GetText();
+                PMCText.Should().Contain(testContext.Project.Name);
+                PMCText.Should().Contain("C#");
+                PMCText.Should().Contain(testContext.Project.FullPath);
             }
         }
 
--- a/test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetConsoleTestCase.cs
+++ b/test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetConsoleTestCase.cs
@@ -966,9 +966,8 @@ namespace NuGet.Tests.Apex
         }
 
         [DataTestMethod]
-        [DynamicData(nameof(GetNetCoreTemplates), DynamicDataSourceType.Method)]
         [Timeout(DefaultTimeout)]
-        public async Task VerifyCmdFindPackageExactMatchInPMC(ProjectTemplate projectTemplate)
+        public async Task VerifyCmdFindPackageExactMatchInPMC()
         {
             EnsureVisualStudioHost();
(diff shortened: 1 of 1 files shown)


test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetConsoleTestCase.cs
```diff
@@ -969,6 +969,91 @@ public void VerifyInitScriptsExecution(ProjectTemplate projectTemplate)
...
+                using (var testContext = new ApexTestContext(VisualStudio, projectTemplate, Logger, addNetStandardFeeds: true, simpleTestPathContext: simpleTestPathContext))
+                {
+                    SolutionService solutionService = VisualStudio.Get<SolutionService>();
+                    var nugetConsole = GetConsole(testContext.Project);
+
+                    // Act
+                    nugetConsole.Execute($"find-package {PackageName} -ExactMatch");
+
+                    // Assert
+                    Assert.IsTrue(nugetConsole.IsMessageFoundInPMC(PackageName), $"The package name {PackageName} doesn't show correctly in PMC.");
```
nkolev92: You might be better off calling `nugetConsole.GetText()` and then you can assert the result in 1 call.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetConsoleTestCase.cs
```diff
@@ -969,6 +969,91 @@ public void VerifyInitScriptsExecution(ProjectTemplate projectTemplate)
...
+                    nugetConsole.InstallPackageFromPMC(packageName, v100);
+                    testContext.SolutionService.Build();
+                    testContext.NuGetApexTestService.WaitForAutoRestore();
+                    nugetConsole.Clear();
+
+                    // Act
+                    nugetConsole.Execute("get-package -update");
+
+                    // Assert
+                    Assert.IsTrue(nugetConsole.IsMessageFoundInPMC(v200), $"The latest package version {v200} doesn't show correctly in PMC.");
```
nkolev92: Same here. Run GetText and then use similar as in https://github.com/NuGet/NuGet.Client/pull/5426/files#diff-d40c827f8805190cb124af37c978b3ac4d843091e7d43a5500182596cf476640R75. 

Use something like. 
```cs
string PMCText = nugetConsole.GetText();
PMCText.Should().Contain("whatever the expected message is");
```

If this test fails, the then fail message will contain actionable information + you don't need to write a custom message. 

In general you want to see the `expected` and `actual` in the error message for test cases.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetConsoleTestCase.cs
```diff
@@ -969,6 +969,91 @@ public void VerifyInitScriptsExecution(ProjectTemplate projectTemplate)
             }
         }
 
+        [DataTestMethod]
+        [DynamicData(nameof(GetNetCoreTemplates), DynamicDataSourceType.Method)]
```
nkolev92: Do the template matter in any of these tests?

nkolev92 (commented):
The tests look great, but I do have some small feedback about the asserts. 

Thanks for using local feeds in every test case! That keeps our tests more stable.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetConsoleTestCase.cs
donnie-msft: 👍

donnie-msft (commented):
I would imagine several project templates could be applied to each of these, right?
Curious how you are choosing which templates to use for these tests? Are there certain commands that fail more often with particular project templates?

For example, `VerifyCmdFindPackageExactMatchInPMC` is using `GetNetCoreTemplates`, while `VerifyCmdGetProjectInPMC` is using `GetPackagesConfigTemplates`.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetConsoleTestCase.cs
CiciLi1: Thanks for your suggestion, I have updated the assert, please help to review them again. Thanks a lot.

CiciLi1:
> I would imagine several project templates could be applied to each of these, right? Curious how you are choosing which templates to use for these tests? Are there certain commands that fail more often with particular project templates?
> 
> For example, `VerifyCmdFindPackageExactMatchInPMC` is using `GetNetCoreTemplates`, while `VerifyCmdGetProjectInPMC` is using `GetPackagesConfigTemplates`.

Hi @donnie-msft, There was a GitHub issue https://github.com/NuGet/Home/issues/7798 when running the command `“get-project”` using `GetNetCoreTemplates`, so we used `GetPackagesConfigTemplates` in the `VerifyCmdGetProjectInPMC` test case.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetConsoleTestCase.cs
CiciLi1: Hi @nkolev92, we are using two different templates since the SDK project having an GitHub issue https://github.com/NuGet/Home/issues/7798 when running the `"get -project"` command, we are using `"GetNetCoreTemplates"` in the other two test cases to cover more templates in the test cases, would you like us to also add `"GetPackagesConfigTemplates"` in other two cases (`VerifyCmdFindPackageExactMatchInPMC` and `VerifyCmdGetPackageUpdateInPMC`?

nkolev92: No, I was actually thinking we only need 1 test case. 

The command being run here is not really project specific, so I don't see why we need to get more than 1 template.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetConsoleTestCase.cs
```diff
@@ -969,6 +969,94 @@ public void VerifyInitScriptsExecution(ProjectTemplate projectTemplate)
...
+                using (var testContext = new ApexTestContext(VisualStudio, projectTemplate, Logger, addNetStandardFeeds: true, simpleTestPathContext: simpleTestPathContext))
+                {
+                    SolutionService solutionService = VisualStudio.Get<SolutionService>();
+                    var nugetConsole = GetConsole(testContext.Project);
+
+                    // Act
+                    nugetConsole.Execute($"find-package {PackageName} -ExactMatch");
+
+                    // Assert
+                    string PMCText = nugetConsole.GetText();
```
nkolev92, line 989: Excellent!

nkolev92 (commented):
Thanks for making the assert changes. 

I think we don't think to test different templates with `Find-Package`. 
Let me know what you think.

CiciLi1:
> Thanks for making the assert changes.
> 
> I think we don't think to test different templates with `Find-Package`. Let me know what you think.

We agree with you!

test/NuGet.Tests.Apex/NuGet.Tests.Apex/NuGetEndToEndTests/NuGetConsoleTestCase.cs
CiciLi1: Hi @nkolev92, we consider that if we test three commands in one case, the others will not run in the test case if the first command fails to execute.

nkolev92: I'm not suggesting we test all 3 commands in one case, but rather to remove the project template in this particular test case. 

We don't need `[DynamicData(nameof(GetNetCoreTemplates), DynamicDataSourceType.Method)]`
We only need 1 project of any type.

CiciLi1: OK, we got it, thanks for your reply, I have removed the project template. Could you please help to review them again. Thanks a lot.

nkolev92 (approved):

donnie-msft (approved):

PR #5511, opened 2023-11-17
Target net9.0 for .NET source build

### LINKED ISSUE #13018: NuGet.Client repo needs to support net9.0 TFM
The [NuGet.Client repo](https://github.com/NuGet/NuGet.Client/) currently targets `net8.0`. For .NET source build, we're updating to use the .NET 9 SDK which requires that NuGet target `net9.0`.

These are the affected sources that need to be updated:

https://github.com/NuGet/NuGet.Client/blob/39ea8e8081541fb1eb4702a5546e12f81fcad34d/build/common.project.props#L15

https://github.com/NuGet/NuGet.Client/blob/39ea8e8081541fb1eb4702a5546e12f81fcad34d/build/common.targets#L9

https://github.com/NuGet/NuGet.Client/blob/39ea8e8081541fb1eb4702a5546e12f81fcad34d/src/NuGet.Core/NuGet.Configuration/NuGet.Configuration.csproj#L29

https://github.com/NuGet/NuGet.Client/blob/39ea8e8081541fb1eb4702a5546e12f81fcad34d/src/NuGet.Core/NuGet.PackageManagement/NuGet.PackageManagement.csproj#L35

mthalman:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13018

## Description

For .NET 9, source build is now dependent on a .NET 9 build of the SDK. This causes the build of the NuGet.Client repo to fail because it's still targeting net8.0. These changes update the TFM references to include net9.0.
 
## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Configuration/NuGet.Configuration.csproj
+++ b/src/NuGet.Core/NuGet.Configuration/NuGet.Configuration.csproj
@@ -26,7 +26,7 @@
     <Reference Include="System.Xml.Linq" />
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(TargetFramework)' == '$(NetStandardVersion)' or '$(TargetFramework)' == 'net7.0' or '$(TargetFramework)' == 'net8.0' ">
+  <ItemGroup Condition=" '$(TargetFramework)' == '$(NetStandardVersion)' or '$(TargetFramework)' == 'net7.0' or '$(TargetFramework)' == 'net8.0' or '$(TargetFramework)' == 'net9.0' ">
     <PackageReference Include="System.Security.Cryptography.ProtectedData" />
   </ItemGroup>
 
--- a/build/common.project.props
+++ b/build/common.project.props
@@ -12,7 +12,7 @@
     <NETFXTargetFrameworkVersion>v4.7.2</NETFXTargetFrameworkVersion>
     <NETFXTargetFramework>net472</NETFXTargetFramework>
     <NETCoreTargetFramework>netcoreapp3.1</NETCoreTargetFramework>
-    <NETCoreTargetFramework Condition="'$(DotNetBuildFromSource)' == 'true'">net8.0</NETCoreTargetFramework>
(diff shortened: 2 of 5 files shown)


src/NuGet.Core/NuGet.Configuration/NuGet.Configuration.csproj
```diff
@@ -26,7 +26,7 @@
     <Reference Include="System.Xml.Linq" />
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(TargetFramework)' == '$(NetStandardVersion)' or '$(TargetFramework)' == 'net7.0' or '$(TargetFramework)' == 'net8.0' ">
+  <ItemGroup Condition=" '$(TargetFramework)' == '$(NetStandardVersion)' or '$(TargetFramework)' == 'net7.0' or '$(TargetFramework)' == 'net8.0' or '$(TargetFramework)' == 'net9.0' ">
```
nkolev92, line 29: We don't really build for all t3 here do we? 

Maybe replacement is in order instead of just adding an extra condition?

I don't feel super strongly about fixing though, cause this is more on us as NuGet.Client rather than something you'd need to figure out.

nkolev92 (dismissed):

donnie-msft (approved):

jeffkl (approved):

src/NuGet.Core/NuGet.Configuration/NuGet.Configuration.csproj
ViktorHofer: Why not just use `"'$(TargetFrameworkIdentifier)' == '.NETCoreApp' or '$(TargetFrameworkIdentifier)' == '.NETStandard'"` here and in all the other places instead?

PR #5514, opened 2023-11-17
Disable TLS certificate validation when `disableTLSCertificateValidation` is set in the config file for a source.

Nigusu-Allehu:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13023

## Description

Disable TLS certificate validation for the HTTP client handler when `disableTLSCertificateValidation` is set in the config file.
- By default, the built in certificate validation is used, however, this PR allows to ignoring TLS certificate problems completely.
## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Core/NuGet.Protocol/HttpSource/HttpHandlerResourceV3Provider.cs
```diff
@@ -56,6 +58,11 @@ private HttpHandlerResourceV3 CreateResource(PackageSource packageSource)
                 AutomaticDecompression = (DecompressionMethods.GZip | DecompressionMethods.Deflate),
             };
 
+            if (packageSource.DisableTLSCertificateValidation)
+            {
+                clientHandler.ServerCertificateCustomValidationCallback = (HttpRequestMessage message, X509Certificate2 cert, X509Chain chain, SslPolicyErrors errors) => true;
```
donnie-msft: This feels broader than just disabling TLS validation. Is it safe?

Perhaps at least check the enum. If I understand the test:
`=> errors == SslPolicyErrors.RemoteCertificateChainErrors`

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpHandlerResourceV3ProviderTests.cs
```diff
@@ -124,5 +125,45 @@ static IEnumerable<DelegatingHandler> GetDelegatingHandlers(HttpMessageHandler h
...
+            result.Item1.Should().BeTrue();
+            HttpHandlerResourceV3 resource = (HttpHandlerResourceV3)result.Item2;
+            resource.Should().NotBeNull();
+
+            HttpClientHandler clientHandler = resource.ClientHandler;
+
+            if (disableCertificateValidation)
+            {
+                clientHandler.ServerCertificateCustomValidationCallback.Should().NotBeNull();
+                var callbackResult = clientHandler.ServerCertificateCustomValidationCallback.Invoke(null, null, null, SslPolicyErrors.RemoteCertificateChainErrors);
```
donnie-msft: This test is specifying a `SslPolicyErrors`, which makes me wonder if the implementation should check for that?

Nigusu-Allehu: I was testing the client handler to ignore the errors. Since for the package source we have set `DisableTLSCertificateValidation` to `true`, it should be ignored and the test checks for that.

donnie-msft: Could pass in all the values instead of just one:

```suggestion
                var callbackResult = clientHandler.ServerCertificateCustomValidationCallback.Invoke(null, null, null, 
                    SslPolicyErrors.RemoteCertificateNotAvailable
                        & SslPolicyErrors.RemoteCertificateNameMismatch
                        & SslPolicyErrors.RemoteCertificateChainErrors);
```

src/NuGet.Core/NuGet.Protocol/HttpSource/HttpHandlerResourceV3Provider.cs
donnie-msft: I see that there's [a proposal ](https://github.com/NuGet/Home/pull/13024)that calls this out and dismisses the concern about it being a global disabling of all certificate errors.
However, I see that the [original proposal PR ](https://github.com/NuGet/Home/pull/12542) did not get reviewed by @dtivel, and I'd be very interested to hear his perspective on it.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpHandlerResourceV3ProviderTests.cs
Nigusu-Allehu: Nice one. Fixed

src/NuGet.Core/NuGet.Protocol/HttpSource/HttpHandlerResourceV3Provider.cs
Nigusu-Allehu: That is a good point. However, I am not sure what you mean by globally. If you meant, disabling TLS globally across all sources, that would be a huge security risk definitely. I think the goal is to disable TLS certification completely, however, per source level.  Please let me know if the code doesn't do that. I have updated the doc to mention that this is done per source level.

kartheekp-ms: As I have mentioned in https://github.com/NuGet/Home/issues/4387#issuecomment-1186028492 comment, I think we should leverage [HttpClientHandler.DangerousAcceptAnyServerCertificateValidator](https://docs.microsoft.com/dotnet/api/system.net.http.httpclienthandler.dangerousacceptanyservercertificatevalidator?view=net-6.0#remarks) Property as suggested in the docs.

[Remarks](https://learn.microsoft.com/dotnet/api/system.net.http.httpclienthandler.dangerousacceptanyservercertificatevalidator?view=net-6.0#remarks) suggest that, 

Particularly in test scenarios, a common pattern use [HttpClient](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=net-6.0) to connect to a server with a certificate that shouldn't be validated, such as a self-signed certificate. You commonly do this with [HttpClientHandler](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler?view=net-6.0) by setting the [ServerCertificateCustomValidationCallback](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.servercertificatecustomvalidationcallback?view=net-6.0) property to a delegate that always returns True; this indicates that the certificate has passed validation. However, not all implementations support this callback, and some throw [PlatformNotSupportedException](https://learn.microsoft.com/en-us/dotnet/api/system.platformnotsupportedexception?view=net-6.0).

The [DangerousAcceptAnyServerCertificateValidator](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.dangerousacceptanyservercertificatevalidator?view=net-6.0) property addresses this limitation. The delegate returned by the [DangerousAcceptAnyServerCertificateValidator](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.dangerousacceptanyservercertificatevalidator?view=net-6.0) property can be assigned to the [ServerCertificateCustomValidationCallback](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.servercertificatecustomvalidationcallback?view=net-6.0) property

kartheekp-ms (changes requested):
IMHO, we need to improve the test coverage for this change. For example, we should ensure that the restore succeeds when `disableTLSCertificateValidate` is set to true, and the mock server implementation includes a self-signed certificate in the chain. I am not an expert on what constitutes good test coverage for this change. Perhaps I am overcomplicating the issue regarding test coverage.

src/NuGet.Core/NuGet.Protocol/HttpSource/HttpHandlerResourceV3Provider.cs
dtivel: Earlier iterations of @JonDouglas's spec did have a global option for disabling TLS certificate verification for all package sources.  Based on feedback (like [this](https://github.com/NuGet/Home/pull/12542#discussion_r1178157422)), the proposal for a global option was [moved](https://github.com/NuGet/Home/pull/12542/commits/ac74ed18f1e82a5cdc9d9604fa2341ca702e14f2) to a ["redacted" section](https://github.com/NuGet/Home/blob/dev/proposed/2023/InsecureConnectionsDisableCertificateValidation.md#redacted-functional-experiences-based-on-community-feedback), meaning you can still see the rejected global option proposal.  The final version of the spec does not support an option for global disablement.  You could have the same net effect by disabling verification on every package source, but a single global option does not exist.

As far as [this feedback](https://github.com/NuGet/NuGet.Client/pull/5514/files#r1397951051) goes, the spec was clear that all verification should be disable-able on a per package source basis.  While the verification _could_ have had more granular knobs (e.g.:  respect some verification failures but ignore others), the spec landed (with community approval) on ignoring all verification failures with one setting on a per package source basis.

+1 to @kartheekp-ms's [excellent suggestion](https://github.com/NuGet/NuGet.Client/pull/5514/files#r1398023860).

Nigusu-Allehu: `HttpClientHandler.DangerousAcceptAnyServerCertificateValidator` property is not available in netstandard2.0. After my discussion with Kartheek today, it was suggested that we could conditionally use `HttpClientHandler.DangerousAcceptAnyServerCertificateValidator` property based on the target framework. For netstandard2.0, we could use this code https://github.com/dotnet/runtime/blob/7532f65c9d8de91600ab5f36000295b0334e75d4/src/libraries/System.Net.Http/src/System/Net/Http/HttpClientHandler.cs#L363 by copying the DangerousAcceptAnyServerCertificateValidator property implementation.

zivkan: Unless I'm missing something, the code you linked is just a complex way of returning `true`, except it doesn't increase the binary size when tree-shaking is used. NuGet isn't a core library where we care about tree-shaking optimization, so I don't see the benefit of copying that code.

src/NuGet.Core/NuGet.Protocol/HttpSource/HttpHandlerResourceV3Provider.cs
```diff
@@ -56,6 +58,11 @@ private HttpHandlerResourceV3 CreateResource(PackageSource packageSource)
                 AutomaticDecompression = (DecompressionMethods.GZip | DecompressionMethods.Deflate),
             };
 
+            if (packageSource.DisableTLSCertificateValidation)
```
kartheekp-ms, line 72: I think we should raise a warning if customers set this property, stating that this setting is not a secure option in a production environment.

Nigusu-Allehu: Added a warning message.

nkolev92: Given that this is a design change, I think this should be covered in the spec as well. 

I'm wondering if adding a warning for something customers explicitly disabled is a bit too loud. 

The workflow here is:
- You get an HTTP deprecation warning.
- You're told about deprecation options, you set the disable tls flag.
- Restore again, get a warning about the tls flag that you just set as a response to the deprecation warning. 

My suggestion would be to not add a warning here. 

Some considerations about warnings being raised in this codepath: 
- How often is the warning raised per source?
- Does every project get the warning or just one?

test/TestUtilities/Test.Utility/TestServer/TcpListenerServer.cs
```diff
@@ -53,6 +66,57 @@ public async Task<T> ExecuteAsync<T>(Func<string, Task<T>> action)
...
+            {
+                var request = new CertificateRequest("cn=test", rsa, HashAlgorithmName.SHA256, RSASignaturePadding.Pkcs1);
+
+                var start = DateTime.UtcNow;
+                var end = DateTime.UtcNow.AddYears(1);
+
+                var cert = request.CreateSelfSigned(start, end);
+
+                var certBytes = cert.Export(X509ContentType.Pfx, "password");
+
```
jebriede: nit: can we remove some blank lines?

test/TestUtilities/Test.Utility/TestServer/TcpListenerServer.cs
```diff
@@ -53,6 +66,57 @@ public async Task<T> ExecuteAsync<T>(Func<string, Task<T>> action)
...
+        }
+
+        private async Task StartInvalidTlsCertificateServer(TcpListener tcpListener, CancellationToken token)
+        {
+            while (!token.IsCancellationRequested)
+            {
+                using (var client = await Task.Run(tcpListener.AcceptTcpClientAsync, token))
+                using (var sslStream = new SslStream(client.GetStream(), false))
+                {
+
```
jebriede: nit: remove blank line.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpHandlerResourceV3ProviderTests.cs
```diff
@@ -124,5 +127,122 @@ static IEnumerable<DelegatingHandler> GetDelegatingHandlers(HttpMessageHandler h
...
+            HttpHandlerResourceV3Provider target = new(proxyCache.Object);
+
+            var result = await target.TryCreate(sourceRepository, CancellationToken.None);
+
+            HttpHandlerResourceV3 resource = (HttpHandlerResourceV3)result.Item2;
+
+            HttpClientHandler clientHandler = resource.ClientHandler;
+
+            var client = new HttpClient(clientHandler);
+
```
jebriede: nit: Can we remove some of the blank lines of code? Typically, I like to separate code with a blank line if it helps readability but if there's a blank line between every line, I think it defeats that purpose and just makes it so there's less code on the screen at once.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpHandlerResourceV3ProviderTests.cs
```diff
@@ -124,5 +127,122 @@ static IEnumerable<DelegatingHandler> GetDelegatingHandlers(HttpMessageHandler h
...
```
jebriede: Nit: Same here, as below, can we remove some blank lines?

test/TestUtilities/Test.Utility/TestServer/TcpListenerServer.cs
```diff
@@ -41,6 +49,11 @@ public async Task<T> ExecuteAsync<T>(Func<string, Task<T>> action)
                     var serverTask = startServer(tcpListener, serverCts.Token);
                     var address = $"http://localhost:{port}/";
 
+                    if (Mode == TestServerMode.InvalidTLSCertificate)
+                    {
+                        address = $"https://localhost:{port}/";
+                    }
```
jebriede: This might be easier to read if expressed as an if/else rather than an assignment then an if with an override of that assignment.

Either as below or using a ternary operator for the assignment:

```suggestion
                    if (Mode == TestServerMode.InvalidTLSCertificate)
                    {
                        address = $"https://localhost:{port}/";
                    }
                    else
                    {
                        var address = $"http://localhost:{port}/";
                    }
```

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpHandlerResourceV3ProviderTests.cs
```diff
@@ -124,5 +127,122 @@ static IEnumerable<DelegatingHandler> GetDelegatingHandlers(HttpMessageHandler h
...
+            Mock<IProxyCache> proxyCache = new();
+            proxyCache.Setup(pc => pc.GetProxy(It.IsAny<Uri>())).Returns((IWebProxy)null);
+
+            PackageSource packageSource = new(_testPackageSourceURL, "source")
+            {
+                DisableTLSCertificateValidation = disableCertificateValidation
+            };
+
+            SourceRepository sourceRepository = new(packageSource, Array.Empty<INuGetResourceProvider>());
+
```
jebriede: nit: Can we remove extra blank lines?

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpHandlerResourceV3ProviderTests.cs
```diff
@@ -124,5 +127,122 @@ static IEnumerable<DelegatingHandler> GetDelegatingHandlers(HttpMessageHandler h
...
+                var callbackResult = clientHandler.ServerCertificateCustomValidationCallback.Invoke(null, null, null, SslPolicyErrors.RemoteCertificateChainErrors
+                    & SslPolicyErrors.RemoteCertificateNameMismatch
+                    & SslPolicyErrors.RemoteCertificateNotAvailable
+                    & SslPolicyErrors.None);
+                callbackResult.Should().BeTrue();
+            }
+            else
+            {
+                clientHandler.ServerCertificateCustomValidationCallback.Should().BeNull();
+            }
```
jebriede: Conditionals in a test typically indicate that we're testing different inputs or state and they should be different tests or a data driven test with different inputs. Please consider breaking this test into multiple tests.

src/NuGet.Core/NuGet.Protocol/Strings.resx
```diff
@@ -530,4 +530,4 @@ The "s" should be localized to the abbreviation for seconds.</comment>
 {1} - number of vulerabilitiy files reported
 {2} - max count allowed</comment>
   </data>
-</root>
+</root>
```
donnie-msft: Revert file

nkolev92: No real changes here, please revert this.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpHandlerResourceV3ProviderTests.cs
```diff
@@ -124,5 +128,119 @@ static IEnumerable<DelegatingHandler> GetDelegatingHandlers(HttpMessageHandler h
...
+            result.Item1.Should().BeTrue();
+            HttpHandlerResourceV3 resource = (HttpHandlerResourceV3)result.Item2;
+            resource.Should().NotBeNull();
+            HttpClientHandler clientHandler = resource.ClientHandler;
+
+            clientHandler.ServerCertificateCustomValidationCallback.Should().BeNull();
+        }
+
+        [Fact]
+        public async Task TryCreate_WhenCertificateValidationIsNotDisabled_ClientHandlerThrowsAnException()
```
donnie-msft: nit: Is there a more precise name?

```suggestion
        public async Task TryCreate_InvalidCertificateWithValidationEnabled_ClientHandlerThrowsAnException()
```

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpHandlerResourceV3ProviderTests.cs
```diff
@@ -124,5 +128,119 @@ static IEnumerable<DelegatingHandler> GetDelegatingHandlers(HttpMessageHandler h
...
+            await server.ExecuteAsync(async uri =>
+            {
+                // Act & Assert
+                var exception = await Assert.ThrowsAsync<HttpRequestException>(async () => await client.GetAsync(uri));
+                return 0;
+            });
+        }
+
+        [Fact]
+        public async Task TryCreate_WhenCertificateValidationIsDisabled_ClientHandlerDoesNotThrowAnException()
```
donnie-msft: nit: Is there a more precise name?

```suggestion
        public async Task TryCreate_InvalidCertificateWithValidationDisabled_ClientHandlerDoesNotThrowAnException()
```

test/TestUtilities/Test.Utility/TestServer/TcpListenerServer.cs
```diff
@@ -53,6 +70,52 @@ public async Task<T> ExecuteAsync<T>(Func<string, Task<T>> action)
...
+            while (!token.IsCancellationRequested)
+            {
+                using (var client = await Task.Run(tcpListener.AcceptTcpClientAsync, token))
+                using (var sslStream = new SslStream(client.GetStream(), false))
+                {
+                    sslStream.AuthenticateAsServer(_tlsCertificate, clientCertificateRequired: false, SslProtocols.Tls12, checkCertificateRevocation: true);
+                    using (var reader = new StreamReader(sslStream, Encoding.ASCII, false, 1))
+                    using (var writer = new StreamWriter(sslStream, Encoding.ASCII, 1, false))
+                    {
+                        while (!string.IsNullOrEmpty(reader.ReadLine()))
```
donnie-msft, line 101: Instead of looping, would reading the entire stream at once work?

```suggestion
                        reader.ReadToEnd();
```
https://learn.microsoft.com/en-us/dotnet/api/system.io.streamreader.readtoend?view=netframework-4.7.2

test/TestUtilities/Test.Utility/TestServer/TcpListenerServer.cs
```diff
@@ -53,6 +70,52 @@ public async Task<T> ExecuteAsync<T>(Func<string, Task<T>> action)
...
+
+        private async Task StartInvalidTlsCertificateServer(TcpListener tcpListener, CancellationToken token)
+        {
+            while (!token.IsCancellationRequested)
+            {
+                using (var client = await Task.Run(tcpListener.AcceptTcpClientAsync, token))
+                using (var sslStream = new SslStream(client.GetStream(), false))
+                {
+                    sslStream.AuthenticateAsServer(_tlsCertificate, clientCertificateRequired: false, SslProtocols.Tls12, checkCertificateRevocation: true);
+                    using (var reader = new StreamReader(sslStream, Encoding.ASCII, false, 1))
```
donnie-msft: Is the reader necessary? It looks like it's unused as the following lines just generate the content.

test/TestUtilities/Test.Utility/TestServer/TcpListenerServer.cs
```diff
@@ -53,6 +70,52 @@ public async Task<T> ExecuteAsync<T>(Func<string, Task<T>> action)
...
+                    using (var writer = new StreamWriter(sslStream, Encoding.ASCII, 1, false))
```
donnie-msft: Is this a buffer size of 1 character? It seems it will default to 128 characters. Might as well use that value for clarity.

https://learn.microsoft.com/en-us/dotnet/api/system.io.streamreader.-ctor?view=netframework-4.7.2#system-io-streamreader-ctor(system-io-stream-system-text-encoding-system-boolean-system-int32)

donnie-msft (commented):
Looks like you've learned a bit about certificates, and I have too by reviewing your work :)
Just a few nits..

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpHandlerResourceV3ProviderTests.cs
```diff
@@ -124,5 +128,119 @@ static IEnumerable<DelegatingHandler> GetDelegatingHandlers(HttpMessageHandler h
                 }
             }
         }
+
+        [Fact]
+        public async Task TryCreate_WhenCertificateValidationIsDisabled_HandlerShouldNotBeNull()
```
nkolev92: You're validating more than just null here, you're calling it: ServerCertificateCustomValidationCallback.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpHandlerResourceV3ProviderTests.cs
```diff
@@ -124,5 +128,119 @@ static IEnumerable<DelegatingHandler> GetDelegatingHandlers(HttpMessageHandler h
...
+            var result = await target.TryCreate(sourceRepository, CancellationToken.None);
+
+            // Assert
+            result.Item1.Should().BeTrue();
+            HttpHandlerResourceV3 resource = (HttpHandlerResourceV3)result.Item2;
+            resource.Should().NotBeNull();
+            HttpClientHandler clientHandler = resource.ClientHandler;
+
+            clientHandler.ServerCertificateCustomValidationCallback.Should().NotBeNull();
+            var callbackResult = clientHandler.ServerCertificateCustomValidationCallback.Invoke(null, null, null, SslPolicyErrors.RemoteCertificateChainErrors
```
nkolev92, line 171: This is another Act & Assert, so ideally this is a new method.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpHandlerResourceV3ProviderTests.cs
nkolev92: The method being tested here is not TryCreate, but it's rather the call using the handler.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpHandlerResourceV3ProviderTests.cs
nkolev92: same as the previous method.

test/TestUtilities/Test.Utility/TestServer/TcpListenerServer.cs
Nigusu-Allehu: Yes, it is not used. However, we want the server to wait for a request message from the client before sending a response. That's why it here. Otherwise, the server could send a response before the client sends a request.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpHandlerResourceV3ProviderTests.cs
Nigusu-Allehu: renamed

test/TestUtilities/Test.Utility/TestServer/TcpListenerServer.cs
donnie-msft: No longer applicable based on other comments.

donnie-msft (approved):

kartheekp-ms:
Team Triage: Please consider adding some dotnet integration tests https://github.com/NuGet/NuGet.Client/blob/dev/test/TestUtilities/Test.Utility/FileSystemBackedV3MockServer.cs

src/NuGet.Core/NuGet.Protocol/HttpSource/HttpHandlerResourceV3Provider.cs
kartheekp-ms: Make sense Andy. So, @Nigusu-Allehu my understanding is that for `netstandard2.0` we will just create a simple delegate that returns `true` similar to what you have right now. We rely on `HttpClientHandler.DangerousAcceptAnyServerCertificateValidator` for other target frameworks because it is recommended as per docs (I linked in my first comment on this thread)

Nigusu-Allehu:
@kartheekp-ms @zivkan , I added some integration tests that are focused on dotnet restore.

test/TestUtilities/Test.Utility/TestServer/TcpListenerServer.cs
```diff
@@ -53,6 +70,52 @@ public async Task<T> ExecuteAsync<T>(Func<string, Task<T>> action)
                 CancellationToken.None);
         }
 
+        public TcpListenerServer()
+        {
+            _tlsCertificate = GenerateSelfSignedCertificate();
```
zivkan: Let's generate the self signed cert only if `TestServerMode.InvalidTLSCertificate` is specified.

test/TestUtilities/Test.Utility/FileSystemBackedTcpListener.cs
```diff
@@ -0,0 +1,213 @@
...
+using System.Security.Cryptography.X509Certificates;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using NuGet.Test.Server;
+
+namespace Test.Utility
+{
+    public class FileSystemBackedTcpListener
```
zivkan: This class name makes it sound like a general purpose package source for any test that wants to serve nupkgs from the filesystem, like what FileSystemBackedV3MockServer is. However, this class really is a limited nuget server that uses a self-signed TLS certificate. So, I think a name like `SelfSignedCertificateMockServer` might be more appropriate.

You've already gone to the effort of making this all work, but I think you could have saved time if you wrote the test to use `dotnet package search`, and you could have hard-coded an empty search result, since what you're trying to test is an HTTP response being received, rather than a TLS error. Alternatively, a restore where the package returns a 404, so the test passes if restore fails with a "package not found" error.  I think doing either of these would have reduced the number of lines of code in this class, since it's not intended to be a generic, reusable and extensible package server.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetRestoreTests.cs
```diff
@@ -2673,6 +2675,62 @@ public async Task DotnetRestore_WithMultiTargettingProject_WhenTargetFrameworkIs
             allTargets.Should().Contain(condition);
         }
 
+        [PlatformFact(Platform.Windows)]
+        public async Task DotnetRestore_withTLSCertificateValidationDisabled_DoesnotThrowException()
```
zivkan: opinionated/nitpick:  This file is already nearly 3,000 lines long, so when a customer reports a bug in restore, it's hard to find if the scenario that the customer is reporting is already tested. This test's focus isn't really about restore, but about testing http communication and if the nuget.config setting applied correctly. So, perhaps we could create a new file whose name make it clear it's about integration tests of our http stack, and the test's implementation just happens to use restore.

src/NuGet.Core/NuGet.Protocol/HttpSource/HttpHandlerResourceV3Provider.cs
```diff
@@ -56,6 +58,18 @@ private HttpHandlerResourceV3 CreateResource(PackageSource packageSource)
                 AutomaticDecompression = (DecompressionMethods.GZip | DecompressionMethods.Deflate),
             };
 
+#if NETSTANDARD2_0
+            if (packageSource.DisableTLSCertificateValidation)
+            {
+                clientHandler.ServerCertificateCustomValidationCallback = (HttpRequestMessage message, X509Certificate2 cert, X509Chain chain, SslPolicyErrors errors) => true;
```
kartheekp-ms: How about declaring the delegate as a static member? For example, https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/src/System/Net/Http/HttpClientHandler.cs#L365

test/TestUtilities/Test.Utility/SelfSignedCertificateMockServer.cs
```diff
@@ -0,0 +1,213 @@
...
+
+        public SelfSignedCertificateMockServer(string packageDirectory)
+        {
+            _packageDirectory = packageDirectory;
+            _certificate = GenerateSelfSignedCertificate();
+        }
+
+        public async Task StartServer()
+        {
+
```
kartheekp-ms: I guess we can remove this blank line.

test/TestUtilities/Test.Utility/SelfSignedCertificateMockServer.cs
```diff
@@ -0,0 +1,213 @@
...
+        private bool _runServer = true;
+        public string URI;
+
+        public SelfSignedCertificateMockServer(string packageDirectory)
+        {
+            _packageDirectory = packageDirectory;
+            _certificate = GenerateSelfSignedCertificate();
+        }
+
+        public async Task StartServer()
```
kartheekp-ms: How about `StartServerAsync` for the method name?

test/TestUtilities/Test.Utility/SelfSignedCertificateMockServer.cs
```diff
@@ -0,0 +1,213 @@
...
+            _runServer = false;
+            _tcpListener.Stop();
+        }
+
+        private async Task HandleClient(TcpClient client)
+        {
+            using (client)
+            using (var sslStream = new SslStream(client.GetStream(), false))
+            {
+                await sslStream.AuthenticateAsServerAsync(_certificate, clientCertificateRequired: false, enabledSslProtocols: SslProtocols.Tls12, checkCertificateRevocation: true);
```
kartheekp-ms: I removed `enabledSslProtocols: SslProtocols.Tls12` argument because there is another overload that doesn't need this parameter and all the 4 new dotnet integration tests passed locally. I guess we can remove passing this parameter. The reason is, if Tls12 protocol gets deprecated in the future, we have to make a code change to use another protocol.

test/TestUtilities/Test.Utility/TestServer/TcpListenerServer.cs
```diff
@@ -53,6 +71,50 @@ public async Task<T> ExecuteAsync<T>(Func<string, Task<T>> action)
...
+        }
+
+        private async Task StartInvalidTlsCertificateServer(TcpListener tcpListener, CancellationToken token)
+        {
+            while (!token.IsCancellationRequested)
+            {
+                using (var client = await Task.Run(tcpListener.AcceptTcpClientAsync, token))
+                using (var sslStream = new SslStream(client.GetStream(), false))
+                {
+                    sslStream.AuthenticateAsServer(_tlsCertificate, clientCertificateRequired: false, SslProtocols.Tls12, checkCertificateRevocation: true);
```
kartheekp-ms: Can we just remove `SslProtocols.Tls12` argument as discussed in https://github.com/NuGet/NuGet.Client/pull/5514#discussion_r1512016642 ?

test/TestUtilities/Test.Utility/TestServer/TcpListenerServer.cs
```diff
@@ -53,6 +70,50 @@ public async Task<T> ExecuteAsync<T>(Func<string, Task<T>> action)
                 CancellationToken.None);
         }
 
+        public TcpListenerServer()
+        { }
+
+        private static X509Certificate2 GenerateSelfSignedCertificate()
```
kartheekp-ms, line 76: `GenerateSelfSignedCertificate` method has same logic in this class and also `SelfSignedCertificateMockServer`. If possible, please consider removing the code duplication in a follow-up PR.

kartheekp-ms (approved):

PR #5515, opened 2023-11-20
Upgrade System.Commandline version

### LINKED ISSUE #13032: Upgrade System.Commandline version
Upgrade the package version to match to the one being used in the sdk.

martinrrm:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13032

## Description

We are in the progress of migrating our current packages to System.Commandline, currently the version in the project is not the latest available version, sdk team is using a greater version https://github.com/search?q=repo%3Adotnet%2Fsdk%20SystemCommandLineVersion&type=code. This version has breaking changes that affect some of our current tooling.

This change will also help to release the feature `dotnet package search` in .NET 8
 - https://github.com/NuGet/Home/issues/13031

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -4,6 +4,7 @@
     <clear />
     <add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" />
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
+    <add key="dotnet-libraries" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries/nuget/v3/index.json" />
     <add key="myget-legacy@Local" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/myget-legacy%40Local/nuget/v3/index.json" />
     <add key="nuget-build" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/nuget-build/nuget/v3/index.json" />
     <add key="vside" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/msft_consumption/nuget/v3/index.json" />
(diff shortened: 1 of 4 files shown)


NuGet.Config
```diff
@@ -4,6 +4,7 @@
     <clear />
     <add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" />
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
+    <add key="dotnet-libraries" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries/nuget/v3/index.json" />
```
donnie-msft, line 7: It's typically fine to do this with source mapping, but wondering if you have any context on why this feed isn't an upstream of `dotnet-public`? I did a quick wiki search as I've never heard of `dotnet-libraries` and found nothing about it.

martinrrm: Is the source that the sdk is using for this package. Latest available version in `dotnet-public` is not the one the sdk team is using. When Nigusu started working on `dotnet package search` and created the PR on the sdk side https://github.com/dotnet/sdk/pull/36289 I realized that we have different versions and should probably work with the one in sdk so we don't work with different versions. 

This new latest version has breaking changes so probably start working with these changes now.

This is also something was commented when Fernando worked on this https://github.com/NuGet/NuGet.Client/pull/4957#discussion_r1050017436

donnie-msft: Totally makes sense that you want to match their package version. I just wonder about the reason they have `dotnet-libraries` separate from the other feeds. Was thinking we could ask them about this to understand the purpose behind the separation, but I found 2 repos with `nuget.configs` that have both feeds side-by-side as well. Perhaps that's enough reason for us to do the same.

donnie-msft (approved):

Nigusu-Allehu (approved):

jeffkl (approved):

PR #5516, opened 2023-11-22
Small refactor of PackageSourceProvider

zivkan:
## Bug

Towards/prerequisite for https://github.com/NuGet/Home/issues/12698

Regression? No

## Description

While starting work on `<auditSources>`, I found `PackageSourceProvider.cs` difficult to understand and reuse. This refactor helped me understand how the default configuration in particular works, and I think can be merged stand-alone, before the rest of the `auditSources` work.

Things changed:

* Added comment for `GetPackageSourceBySource` to explain behaviour when two sources have different keys/names, but the same URL. Restore dedupes them, but commands like `dotnet nuget list source` does not. Added test for this before the refactor, to validate my refactor didn't change behaviour.
* Changed `GetPackageSourceByName` and `GetPackageSourceBySource` to get all package sources the same way that `LoadPackageSources()` does, so that there's only once place that "machine-wide default sources" need to be implemented/checked/used.
* minor, theoretical perf improvements. Things like using `List<T>` rather than `IEnumerable<T>`, so that `foreach` doesn't need an allocation. Also create lists with correct capacity to minimize risk of resizing.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
```diff
@@ -103,20 +107,20 @@ private static IEnumerable<PackageSource> LoadConfigurationDefaultSources(IEnume
...
-            return LoadPackageSourceLookup(byName: true, settings);
-        }
+                packageSources = new(capacity: packageSourceLookup.Count);
+                packageSources.AddRange(packageSourceLookup.Values
+                    .OrderBy(psi => psi.Index).
+                    Select(psi => psi.PackageSource));
+            }
+            else
+            {
+                packageSources = new List<PackageSource>();
```
kartheekp-ms, line 124: Can we use `Enumerable.Empty<PackageSource>` here?

EDIT: Using `Enumerable.Empty<PackageSource>` may not be possible here because the return type is a List. Maybe we could use `new List<PackageSource>(0)`. Found a similar pattern used in the repository.

https://github.com/NuGet/NuGet.Client/blob/229ddddf43a8b4d9a4a796d6b36736f87b8f0e16/src/NuGet.Core/NuGet.ProjectModel/JsonUtility.cs#L98

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
```diff
@@ -56,45 +55,50 @@ public class PackageSourceProvider : IPackageSourceProvider
...
+            if (Common.RuntimeEnvironmentHelper.IsMono)
             {
-                return Enumerable.Empty<PackageSource>();
+                return Array.Empty<PackageSource>();
             }
 #endif
             var packageSourceLookup = new Dictionary<string, IndexedPackageSource>(StringComparer.OrdinalIgnoreCase);
             var packageIndex = 0;
 
             foreach (var packageSource in configurationDefaultSources)
```
kartheekp-ms, line 74: I understand this is beyond the scope of this PR. However, if you're addressing other feedback within this PR, please consider incorporating ``.NoAllocEnumerate()``. This addition will help avoid memory allocations if the underlying collection is a list.

```suggestion
            foreach (var packageSource in configurationDefaultSources.NoAllocEnumerate())
```

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
```diff
@@ -56,45 +55,50 @@ public class PackageSourceProvider : IPackageSourceProvider
...
             {
-                packageIndex = AddOrUpdateIndexedSource(packageSourceLookup, packageIndex, packageSource, packageSource.Name);
+                packageIndex = AddOrUpdateIndexedSource(packageSourceLookup, packageIndex, packageSource);
```
kartheekp-ms: The `AddOrUpdateIndexedSource` method updates the `packageIndex` and returns the new value. Could we consider changing the parameter to `ref`? This suggestion might be beyond the scope of this PR, but the current usage of `packageIndex` is a bit confusing without navigating into the `AddOrUpdateIndexedSource` method.

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
```diff
@@ -56,45 +55,50 @@ public class PackageSourceProvider : IPackageSourceProvider
...
+            defaultSources.AddRange(packageSourceLookup.Values
                 .OrderBy(source => source.Index)
-                .Select(source => source.PackageSource);
+                .Select(source => source.PackageSource));
+
+            return defaultSources.AsReadOnly();
         }
 
-        private static Dictionary<string, IndexedPackageSource> LoadPackageSourceLookup(bool byName, ISettings settings)
+        private static List<PackageSource> GetPackageSourceFromSettings(ISettings settings)
```
kartheekp-ms, line 87: Is there a specific reason why we didn't return `IReadOnlyList` from this method, similar to the `LoadConfigurationDefaultSources` method?

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
zivkan: The default constructor already creates a list with zero capacity.

.NET Framework: https://referencesource.microsoft.com/#mscorlib/system/collections/generic/list.cs,49

.NET (Core): https://github.com/dotnet/runtime/blob/cbc501ca196371572c38f8d12a66969864d99c08/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/List.cs#L33-L40

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
zivkan: We need to add sources from `NuGetDefaults.config` to the list (call the method `AddDefaultPackageSources`) after getting this list from the `ISettings` object.

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
kartheekp-ms: Thank you for addressing this feedback in https://github.com/NuGet/NuGet.Client/pull/5516/commits/0719477b182093e376462bb2d8983c336fda6a39 commit.

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
kartheekp-ms: I was under the impression that default constructor creates an array of with some default capacity. Thank you for the clarification.

nkolev92 (approved):

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
```diff
@@ -296,50 +301,31 @@ private static PackageSourceCredential ReadCredentialFromEnvironment(string sour
                 validAuthenticationTypesText: match.Groups["authTypes"].Value);
         }
 
-        private PackageSource GetPackageSource(string key, Dictionary<string, IndexedPackageSource> sourcesLookup)
+        public PackageSource GetPackageSourceByName(string name)
```
kartheekp-ms, line 304: There are no unit tests for this public method. I acknowledge that this issue wasn't introduced in this PR. I am comfortable with either creating a follow-up issue or adding a simple unit test in this PR, similar to how a unit test was added for the `GetPackageSourceBySource` method.

test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs
```diff
@@ -2271,6 +2271,34 @@ public void PackageSourcesChanged_EventRunsSubscriptions(bool subscribeToEvent)
...
         }
 
+        [Fact]
+        public void GetPackageSourceBySource_TwoSourcesWithSameUrl_ReturnsFirstSource()
+        {
+            // Arrange
+            using TestDirectory testDirectory = TestDirectory.Create();
+
+            const string sourceUrl = "https://contoso.test/nuget/index.json";
+            const string contents = $@"<configuration>
```
kartheekp-ms, line 2281: I think we can leverage the `CreateNuGetConfigContent` method to create NuGet.Config file content by simply passing the package sources. In my opinion, the said method has room for improvement, which is beyond the scope of this PR. For instance, an improvement could be to create the nuget.config string just once (cache it), and then invoke the `Replace` method as needed. This is effective because the `Replace` method creates a new string anyway.

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
```diff
@@ -358,7 +344,17 @@ public PackageSource GetPackageSourceBySource(string source)
...
+
+            foreach (var packageSource in packageSources)
+            {
+                if (string.Equals(source, packageSource.Source, StringComparison.OrdinalIgnoreCase))
+                {
+                    return packageSource;
+                }
+            }
+
+            return null;
```
kartheekp-ms, line 357: We could simplify this logic using LINQ. Is the original implementation chosen to avoid memory allocations? For instance, the code can be condensed to:

```cs
return packageSources.FirstOrDefault(ps => string.Equals(source, ps.Source, StringComparison.OrdinalIgnoreCase));
```

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
```diff
@@ -859,6 +855,11 @@ public string ActivePackageSourceName
         /// <param name="source"></param>
         public void SaveActivePackageSource(PackageSource source)
         {
+            if (source is null)
+            {
+                throw new ArgumentNullException(nameof(source));
```
kartheekp-ms, line 860: In this PR, we implemented `throw new ArgumentException(Resources.Argument_Cannot_Be_Null_Or_Empty, nameof(source));` as exemplified in https://github.com/NuGet/NuGet.Client/pull/5516/files#diff-93482a8631efd30f5df833baa903bcf342fb99a2e93cf9a3dea6a672b44d3a61R308. 

I assume the only difference is that the exception message will be localized when obtained from the resources file. Although I am comfortable with the current implementation, feel free to consider this suggestion if you are addressing other feedback on this PR.

test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs
donnie-msft: There's also a `SettingsTestUtils.CreateConfigurationFile`

test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs
```diff
@@ -2271,6 +2271,34 @@ public void PackageSourcesChanged_EventRunsSubscriptions(bool subscribeToEvent)
...
+  <packageSources>
+    <add key=""s1"" value=""{sourceUrl}"" />
```
donnie-msft, line 2283: Thoughts on adding a `clear`? I think it would just be as a reminder for the reader that we 1. expect it as best practice and 2. the test doesn't anticipate any other configs to apply. 

```suggestion
    </clear>
    <add key=""s1"" value=""{sourceUrl}"" />
```

donnie-msft (approved):

test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs
zivkan: I had a quick look at `CreateNuGetConfigContent`, but it appears to be that using it makes the test significantly less readable. If you disagree, can you please point me to a test that's using it in a clear way?

IMO `SettingsTestUtils.CreateConfigurationFile` is functionally equivalent to `File.WriteAllText` (and seems to be the same method signature), but causes Large Object Heap allocations, if the config content is too large. So, I think that `SettingsTestUtils.CreateConfigurationFile` would be better off deleted, and all tests currently using it should migrate to the BCL API.

test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs
zivkan: The file is loaded using an API that loads only the single file. It doesn't do the "normal" directory traversal and multi-file accumulation. It would have been nice if the API name made that clear (even better if it was possible to parse a `Settings` instance from memory, rather than needing a file on disk), but I'm just working what already exists.

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
zivkan: The BCL already localizes exception messages.

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
zivkan: Calling `FirstOrDefault` will cause one allocation (well, not directly, but the implementation will call `IEnumerable<T>.GetEnumerator()` which will allocate). Also, creating the lambda function will allocate. It's "only" two allocations, it's not O(n) allocations, but still the no-LINQ implementation is zero allocations.

test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs
kartheekp-ms: > I had a quick look at CreateNuGetConfigContent, but it appears to be that using it makes the test significantly less readable. If you disagree, can you please point me to a test that's using it in a clear way?

There are a couple of places where the CreateNuGetConfigContent method is used, as seen here: https://github.com/search?q=repo%3ANuGet%2FNuGet.Client%20CreateNuGetConfigContent%20&type=code. I am also okay with the current approach.

kartheekp-ms (approved):
A couple of my comments in this PR remain unresolved. Feel free to take appropriate action, which could mean creating a follow-up issue, addressing it, or taking no action.

https://github.com/NuGet/NuGet.Client/pull/5516#discussion_r1406690544
https://github.com/NuGet/NuGet.Client/pull/5516#discussion_r1402858295

PR #5520, opened 2023-11-29
Hide vulnerabilities filter for Solution Browse tab

### LINKED ISSUE #12974: Vulnerabilities filter shows incorrectly on “Brower” tab when the default tab was “Browse” instead of “Installed” for the previous opening of solution PM UI
### NuGet Product Used

### Product Version

NuGet Client Dev\6.9.0.22

### Worked before?

_No response_

### Impact

#### Repro Steps:       

1. Create a C# Class Library (.NET Core 8.0) project.   
2. Right-click the solution in Solution Explorer and select "Manage NuGet Packages for Solution…" menu.   
3. Click the "Browse" tab and close the solution-level PM UI.  
4. Right-click the solution in Solution Explorer and select "Manage NuGet Packages for Solution…" menu to open the solution-level PM UI again.   

#### Expected:        

Vulnerabilities filter should not show on “Brower” tab. 

#### Actual:        

Vulnerabilities filter shows incorrectly on “Brower” tab as below: 
Unbenannt

#### Note:    

1.The repro rate is 100%.  
2.This is not a regression since it is a new feature— Vulnerabilities filter.  

### Verbose Logs

_No response_

martinrrm:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12974

## Description

PM UI remembers the last tab you had open before closing it. If you open Solution PM UI, change to Browse tab and close it. Next time you open Solution PM UI it will be in the Browse tab, current implementation of the filter is not considering this.

I changed the Filter to be hidden by default and change according to the current tab. The funcitons that handles the visibility of the filter is `TabsPackageManagement_SelectionChanged` which is called when the UI is created and when we change tab.

When creating the UI `CheckBoxVulnerabilities`, the filter, is null so we cannot change the visibility here, so we rely on `TabsPackageManagement_SelectionChanged` being called again to update the visibility. Fortunately (or unfortunately) when the Tab is not Browse, `TabsPackageManagement_SelectionChanged` is called twice because first we create the UI with the tab selected to Browse and then switch the selection to the actual Tab.
 
vulnerabilities-filter-12974

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A  UI

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerTopPanel.xaml
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerTopPanel.xaml
@@ -244,7 +244,7 @@
           VerticalContentAlignment="Center"
           Checked="_checkboxVulnerabilities_Checked"
           Unchecked="_checkboxVulnerabilities_Unchecked"
-          Visibility="Visible"
+          Visibility="Hidden"
           IsChecked="False">
           <TextBlock Text="{x:Static nuget:Resources.Checkbox_Show_Vulnerable_Only}" />
         </CheckBox>
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
@@ -135,7 +135,6 @@ namespace NuGet.PackageManagement.UI
             if (_topPanel.IsSolution)
             {
                 _topPanel.CreateAndAddConsolidateTab();
-                _topPanel.CheckBoxVulnerabilities.Visibility = Visibility.Hidden;
             }
 
             _settingsKey = await GetSettingsKeyAsync(CancellationToken.None);
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
@@ -135,6 +135,7 @@ namespace NuGet.PackageManagement.UI
             if (_topPanel.IsSolution)
             {
                 _topPanel.CreateAndAddConsolidateTab();
+                _topPanel.CheckBoxVulnerabilities.Visibility = Visibility.Hidden;
             }
 
             _settingsKey = await GetSettingsKeyAsync(CancellationToken.None);


Nigusu-Allehu (approved):

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerTopPanel.xaml
```diff
@@ -244,7 +244,7 @@
           VerticalContentAlignment="Center"
           Checked="_checkboxVulnerabilities_Checked"
           Unchecked="_checkboxVulnerabilities_Unchecked"
-          Visibility="Visible"
+          Visibility="Hidden"
```
donnie-msft, line 247: I would expect `Collapsed` unless there's a penalty on redrawing the controls on resizing.
Did you test a small PM UI window and see a problem with toggling between Collapsed/Visible?

kartheekp-ms, line 247: If I understand https://github.com/NuGet/Home/issues/12974#issuecomment-1785966512 correctly, I think the suggestion is to hide the vulnerabilities filter for the project and solution level browse tab but not to hide it completely only on the solution level browse tab. 

[Gif file](https://private-user-images.githubusercontent.com/43253759/286721662-60a9a9b1-0d06-4f06-9f8f-3e6ca6489f52.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTEiLCJleHAiOjE3MDEyOTU5OTcsIm5iZiI6MTcwMTI5NTY5NywicGF0aCI6Ii80MzI1Mzc1OS8yODY3MjE2NjItNjBhOWE5YjEtMGQwNi00ZjA2LTlmOGYtM2U2Y2E2NDg5ZjUyLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFJV05KWUFYNENTVkVINTNBJTJGMjAyMzExMjklMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjMxMTI5VDIyMDgxN1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWI5NDI3MWQ1ZDAwNzMxMWIxMTljZjE3ODlhZDRlMGNjZWE1YmJlMjlmZjA5OTZkZDJiNmRhZjljOTBjYzFlZDMmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.D7SfDPxxvuEW10iTSVXHMkkaN-DgVG8D3St_Ej_G1Bk) mentioned in the PR description shows only project level UI but doesn't show how this change would impact solution level UI.

Please let me know if I missed anything.

martinrrm: Playing with this I found something, if the Visibility is `Collapsed` then the UI doesn't reserve space for the checkbox and if you change tab the search control is going to change size. What do you think about it? I like more Hidden

Hidden
vulnerabilities-filter-12974-project-hidden

Collapse, the gif also shows the behavior of the solution tab
vulnerabilities-filter-12974-solution

martinrrm: The UI component that contains the checkbox `PackageManagerTopPanel` is shared between the Project and Solution level so the change is going to affect both.

Here is a gif of how it will look in the Solution level
vulnerabilities-filter-12974-solution

donnie-msft: Yes, that's the difference between Collapsed/Hidden. However, I don't see any reason why the Search textbox should shrink in your example as your screen has plenty of blank space in the Top Panel. That probably means the containing layout control is too small, so it's making the vulnerability checkbox consume some of its space.

I think we need to fix this to use the proper width, and only resize controls when necessary (based on the PM UI's width).
It's reasonable to merge this then fix that separately since 1. the hidden behavior is already released before this PR and 2. the collapsed behavior will need to be tested well (different resolutions, DPI scaling, etc).

donnie-msft: Martin, yes the control is shared, but that doesn't mean we can't have different behavior in Solution vs. Project level. Eg, we only have Consolidate tab in one, etc.

Kartheek, I think the title of the issue 12974 is misleading. The repro happened on Solution level, but I believe the change should be for either Solution or Project level PM UI.

donnie-msft (approved):
I'm approving based on need to not show the filter on Browse tab. I left a comment that we should follow-up and stop shrinking the Search box as we have no real reasoning for that behavior.

martinrrm:
@donnie-msft I created the follow up issue, I'll work on that next sprint https://github.com/NuGet/Home/issues/13090

PR #5521, opened 2023-11-29
Stop enumerating MSBuild item metadata names in static graph-based restore

### LINKED ISSUE #13049: Static graph-based restore should not enumerate every item's metadata
### NuGet Product Used

MSBuild.exe, dotnet.exe

### Product Version

All

### Worked before?

_No response_

### Impact

### Repro Steps & Context

Static graph-based restore uses an implementation of IMSBuildItem which is supposed to allow access to the names of all of the metadata items associated with a particular MSBuild item.  However, the names are not needed during static graph-based restore.

https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildProjectItemInstance.cs#L30

I only found one usage in Pack which seems like its not causing any slowness issues.

### Verbose Logs

Image

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13049

## Description

The `MSBuildItemBase` object is just an implementation of `IMSBuildItem` which is supposed to implement the `Properties` property.  However, this code path does not need access to the list of item metadata so it's causing unnecessary allocation.  I also added a `Debug.Fail()` to prevent usage of it since its required to implement the member.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception - No need to test this since its not actually used but the constructor is
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildItemBase.cs
+++ b/src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildItemBase.cs
@@ -1,7 +1,9 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using NuGet.Commands;
 
 namespace NuGet.Build.Tasks.Console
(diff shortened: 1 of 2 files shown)


src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildItemBase.cs
```diff
@@ -15,7 +17,15 @@ internal abstract class MSBuildItemBase : IMSBuildItem
         public virtual string Identity { get; }
 
         /// <inheritdoc cref="IMSBuildItem.Properties" />
-        public virtual IReadOnlyList<string> Properties { get; }
+        public IReadOnlyList<string> Properties
+        {
+            get
+            {
+                Debug.Fail("This property should not be accessed.  Calculating the names of the item metadata allocates a new list and can cause performance issues.  Use GetProperty(string) instead and check if the value is null or whitespace.");
```
nkolev92: I feel like Debug checks like that are not super helpful, since they rely on people debugging restore, and it's not something that's done that frequently.

Can we throw and mark the method as Obsolete?

jeffkl: What I kind of expected is that unit tests or restores with locally built bits would fail if any future code path accessed this property.  I'll mark it as obsolete and throw instead.  This class is internal so at least we know no one is already using it.

martinrrm (approved):

nkolev92 (approved):

PR #5523, opened 2023-11-29
switch dictionary reference to no longer use Add to avoid concurrent …

### LINKED ISSUE #13067: Getting "An item with the same key has already been added" error when restoring.
### NuGet Product Used

Other/NA

### Product Version

17.8.1

### Worked before?

17.7.6

### Impact

None

### Repro Steps & Context

Unable to reproduce. Addressing [Feedback Ticket](https://developercommunity.visualstudio.com/t/Error-reading-git-repository-information/10525656). 

### Verbose Logs

_No response_

jgonz120:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13067
Feedback Ticket: https://developercommunity.visualstudio.com/t/Error-reading-git-repository-information/10525656

## Description

When checking if a feature is enabled the NuGetFeatureFlagService was using the ".Add" function. We received DevCom feedback showing that a user was getting an error because the key already existed in the dictionary. 
## PR Checklist

- [ ] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.VisualStudio.Common/Services/NuGetFeatureFlagService.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Common/Services/NuGetFeatureFlagService.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.Threading.Tasks;
 using Microsoft.Internal.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.Services.Common;
 using Microsoft.VisualStudio.Shell;
 using NuGet.Common;
 
@@ -44,8 +45,11 @@ namespace NuGet.VisualStudio
             {
                 var featureFlagService = await _ivsFeatureFlags.GetValueAsync();
                 await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
-                featureEnabled = featureFlagService.IsFeatureEnabled(featureFlag.Name, defaultValue: featureFlag.DefaultState);
-                _featureFlagCache[featureFlag.Name] = featureEnabled;
+                if (!_featureFlagCache.TryGetValue(featureFlag.Name, out featureEnabled))
+                {
+                    featureEnabled = featureFlagService.IsFeatureEnabled(featureFlag.Name, defaultValue: featureFlag.DefaultState);
+                    _featureFlagCache.TryAdd(featureFlag.Name, featureEnabled);
+                }
             }
             return !isFeatureForcedDisabled && (isFeatureForcedEnabled || featureEnabled);
         }
--- a/src/NuGet.Clients/NuGet.VisualStudio.Common/Services/NuGetFeatureFlagService.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Common/Services/NuGetFeatureFlagService.cs
@@ -45,7 +45,7 @@ namespace NuGet.VisualStudio
                 var featureFlagService = await _ivsFeatureFlags.GetValueAsync();
                 await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
                 featureEnabled = featureFlagService.IsFeatureEnabled(featureFlag.Name, defaultValue: featureFlag.DefaultState);
-                _featureFlagCache.Add(featureFlag.Name, featureEnabled);
+                _featureFlagCache[featureFlag.Name] = featureEnabled;
             }
             return !isFeatureForcedDisabled && (isFeatureForcedEnabled || featureEnabled);
         }
--- a/src/NuGet.Clients/NuGet.VisualStudio.Common/Services/NuGetFeatureFlagService.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Common/Services/NuGetFeatureFlagService.cs
@@ -6,7 +6,6 @@ using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.Threading.Tasks;
 using Microsoft.Internal.VisualStudio.Shell.Interop;
-using Microsoft.VisualStudio.Services.Common;
 using Microsoft.VisualStudio.Shell;
 using NuGet.Common;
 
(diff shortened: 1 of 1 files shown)


src/NuGet.Clients/NuGet.VisualStudio.Common/Services/NuGetFeatureFlagService.cs
```diff
@@ -45,7 +45,7 @@ public async Task<bool> IsFeatureEnabledAsync(NuGetFeatureFlagConstants featureF
                 var featureFlagService = await _ivsFeatureFlags.GetValueAsync();
                 await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
                 featureEnabled = featureFlagService.IsFeatureEnabled(featureFlag.Name, defaultValue: featureFlag.DefaultState);
-                _featureFlagCache.Add(featureFlag.Name, featureEnabled);
+                _featureFlagCache[featureFlag.Name] = featureEnabled;
```
kartheekp-ms: Should we use [`ConcurrentDictionary`](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2?view=net-8.0) instead of `Dictionary` to avoid concurrency issues?

kartheekp-ms:
> Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1921835

As per the current process we create a tracking issue in NuGet/Home and mention it PR description so that the bug fix can be included in the release notes.

src/NuGet.Clients/NuGet.VisualStudio.Common/Services/NuGetFeatureFlagService.cs
jgonz120: I was discussing with @nkolev92 and it's really not clear why we would fall into this scenario. So this might be an edge case we don't get very often. I looked at the ConcurrentDictionary but thought the overhead might not be worth it.

donnie-msft: My guess is the discussion wasn't about whether this can happen - because clearly it does from your feedback ticket - but rather, it won't matter if multiple calls result in updating the cache as it should always be the same value (the service won't read settings multiple times). Is that right?

Instead of allowing the cache to update again, what about another checking `if` the key exists immediately after switching to the main thread? No reason to call the `featureFlagService` again if the reasoning I gave above is right.

jeffkl (dismissed):

src/NuGet.Clients/NuGet.VisualStudio.Common/Services/NuGetFeatureFlagService.cs
jgonz120: So adding another TryGetValue? 

            if (!_featureFlagCache.TryGetValue(featureFlag.Name, out bool featureEnabled))
            {
                var featureFlagService = await _ivsFeatureFlags.GetValueAsync();
                await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
                if (!_featureFlagCache.TryGetValue(featureFlag.Name, out featureEnabled))
                {
                    featureEnabled = featureFlagService.IsFeatureEnabled(featureFlag.Name, defaultValue: featureFlag.DefaultState);
                    _featureFlagCache[featureFlag.Name] = featureEnabled;
                }
            }

donnie-msft: Yes, and could go back to using `Add`. I believe if the check is done on the main thread, the concurrency issue wouldn't happen? It is an alternative that may alleviate the question of whether a `ConcurrentDictionary` is necessary.
If I'm right, then switching to the main thread is in effect entering into a lock (since there's only 1 main thread).

jgonz120: Yea that's correct, which is part of what we discussed during the PR call today, there shouldn't be two items updating the cache at the same time. I think this is a decent compromise though. We could also just do TryAdd 🤷

nkolev92 (dismissed):

src/NuGet.Clients/NuGet.VisualStudio.Common/Services/NuGetFeatureFlagService.cs
jgonz120: I did a little bit of both, updated it to check again if the feature flag is in the cache, and use try add

src/NuGet.Clients/NuGet.VisualStudio.Common/Services/NuGetFeatureFlagService.cs
```diff
@@ -6,6 +6,7 @@
 using System.ComponentModel.Composition;
 using System.Threading.Tasks;
 using Microsoft.Internal.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.Services.Common;
```
nkolev92: Do we need this?

jgonz120: that's where TryAdd comes from apparently 🤔

nkolev92: Oh yeah, TryAdd isn't available on regular dict. 
Given that you're making the call on the UI thread, you can probably just call `[]`. 

I don't care too much, but it seems weird to bring in namespaces for something basic like that.

jeffkl (approved):

build/common.project.props
```diff
@@ -82,7 +82,7 @@
     <Features>strict</Features>
     <!-- Same as SDK default, but without CandidateAssemblyFiles in front, which would search in Content items -->
     <AssemblySearchPaths>{HintPathFromItem};{TargetFrameworkDirectory};{RawFileName}</AssemblySearchPaths>
-    <LangVersion>11</LangVersion>
+    <LangVersion>12</LangVersion>
```
zivkan: This change, and others that GitHub are now showing, are not new in this PR.  I don't know why GitHub thinks it's in the diff, but this is an example of why I never merge dev to my feature branches, and always rebase my feature branches (and therefore force push) instead.

Please fix the branch, so I can see what your changes are.

jgonz120: I thought I did rebase form the IDE but I guess not? anyway fixed

zivkan (approved):

PR #5524, opened 2023-11-30
Filter out transitive packages with no origins when deciding if to show vulnerabilities warning on Installed tab

### LINKED ISSUE #13068: Vulnerability indicator shows up on dependent project if parent project has transitive vulnerabilities
---
**Issue moved from NuGet/Client.Engineering#2600**
- Please respond to @jebriede.

---

_From @jebriede on Wednesday, November 29, 2023 9:07:58 PM_

### NuGet Product Used

### Product Version

Version 17.9.0 Preview 2.0 [34328.292.main]

### Worked before?

N/A, new feature

### Impact

### Repro Steps & Context

1. Create a solution with 2 projects, ProjectA and ProjectB
2. In ProjectA, install a top-level package that depends on one or more transitive package that is vulnerable. For example, Microsoft.AspNet.WebApi.Client version 5.2.9 which depends on Newtonsoft.Json 10.0.1 which is vulnerable
3. Add a project reference from ProjectB -> ProjectA. That is, ProjectB depends on ProjectA.
4. Open the NuGet Package Manager for Project B.

**Expected:**
The Installed Tab should not have a warning indicator by it in ProjectB, since it has no packages installed.

**Actual**
The Installed Tab has a warning indicator that 3 vulnerable packages are installed (presumably because the dependent project, ProjectA, has 3 transitive vulnerable packages), even though ProjectB has no installed packages:

image

Note: This only seems to happen when there's a project dependency between ProjectB which depends on ProjectA. If there are no project dependencies, the issue does not repro. Also, this only repros if ProjectA has transitive package vulnerabilities. If it has top-level package vulnerabilities only, the issue does not repro.

### Verbose Logs

_No response_

jebriede:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13068

## Description
Vulnerabilities for Transitive Dependencies fetched Installed and Transitive Dependencies when loading or refreshing the PM UI so it can show a warning indicator if a project has transitive vulnerabilities. The issue is that if a project depends on another project and the parent project has any package dependencies, the dependent project considers all those package dependencies to be transitive dependencies, and the way to differentiate them between transitive packages that are because of a project dependency vs. transitive packages that are because of a top-level package that is installed to that project is by checking if it has any transitive origins.

This fix calculates transitive origins for transitive packages and filters out any transitive packages that have no transitive origins, which means they come from a dependent project, so the vulnerability indicator only shows if the project brings in the transitive packages with vulnerabilities, not when the project it depends on does.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception
  - **OR**
  - [ ] N/A 

Manually tested with a solution that contains two projects. ProjectA has transitive vulnerabilities, and ProjectB depends on ProjectA. ProjectB has no packages installed. Opened PM UI for ProjectB to verify that the warning indicator does not show up next to the Installed tab. Verified ProjectA does have the indicator.

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
@@ -1010,9 +1010,9 @@ namespace NuGet.PackageManagement.UI
             }
             else
             {
-                IInstalledAndTransitivePackages installedAndTransitivePackages = await PackageCollection.GetInstalledAndTransitivePackagesAsync(loadContext.ServiceBroker, loadContext.Projects, includeTransitiveOrigins: false, token);
+                IInstalledAndTransitivePackages installedAndTransitivePackages = await PackageCollection.GetInstalledAndTransitivePackagesAsync(loadContext.ServiceBroker, loadContext.Projects, includeTransitiveOrigins: true, token);
                 installedPackageCollection = PackageCollection.FromPackageReferences(installedAndTransitivePackages.InstalledPackages);
-                PackageCollection transitivePackageCollection = PackageCollection.FromPackageReferences(installedAndTransitivePackages.TransitivePackages);
+                PackageCollection transitivePackageCollection = PackageCollection.FromPackageReferences(installedAndTransitivePackages.TransitivePackages.Where(p => p.TransitiveOrigins.Any()));
                 IEnumerable<PackageVulnerabilityMetadataContextInfo>[] transitivePackageVulnerabilities = await Task.WhenAll(transitivePackageCollection.Select(p => _packageVulnerabilityService.GetVulnerabilityInfoAsync(p, token)));
 
                 foreach (IEnumerable<PackageVulnerabilityMetadataContextInfo> vulnerabilityInfo in transitivePackageVulnerabilities)


src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
```diff
@@ -1010,9 +1010,9 @@ private async Task<(int, int)> GetInstalledVulnerableAndDeprecatedPackagesCountA
             }
             else
             {
-                IInstalledAndTransitivePackages installedAndTransitivePackages = await PackageCollection.GetInstalledAndTransitivePackagesAsync(loadContext.ServiceBroker, loadContext.Projects, includeTransitiveOrigins: false, token);
+                IInstalledAndTransitivePackages installedAndTransitivePackages = await PackageCollection.GetInstalledAndTransitivePackagesAsync(loadContext.ServiceBroker, loadContext.Projects, includeTransitiveOrigins: true, token);
                 installedPackageCollection = PackageCollection.FromPackageReferences(installedAndTransitivePackages.InstalledPackages);
-                PackageCollection transitivePackageCollection = PackageCollection.FromPackageReferences(installedAndTransitivePackages.TransitivePackages);
+                PackageCollection transitivePackageCollection = PackageCollection.FromPackageReferences(installedAndTransitivePackages.TransitivePackages.Where(p => p.TransitiveOrigins.Any()));
```
nkolev92, line 1015: I'm not a fan of this whole origins situation :(

If I have 10 projects, 9 class libs and 1 runnable project, I'd probably focus on the vulnerabilities being reported by the top project, the runnable project, since that's what gets deployed. 

The way it is today, you'll go to the PM UI, and get 3 transitive vulnerabilities, while restore might report 5.

jebriede: I'm not sure how I can take action on this comment. Can you please expand on what concerns you have with this fix, specifically?

nkolev92: My suggestion is we should be showing all transitive dependencies and not hide the fact that certain transitive dependencies are vulnerable, even if they're coming through a project reference.

Normally, my suggestion might be considered out of scope, but I think in the context of vulnerabilities, this design decision is worth reconsidering.

jebriede: Thanks for the feedback on transitive dependencies! We synced offline about the current design which is to not show packages that were installed across project dependencies. Could you please create a DCR for this suggestion that the project-level PM UI can include packages across project dependencies in dependent projects?

As we talked about, this bug makes the count of vulnerable packages in the warning indicator be out of sync with what we show in the Installed tab which can be confusing. A project can have 0 installed packages but still show it has 3 vulnerabilities if the project it depends on has 3 vulnerabilities. This fix addresses this scenario and puts the warning indicator in line with the current design of transitive dependencies in the PM UI. Thanks for your help reviewing this PR!

nkolev92: https://github.com/NuGet/Home/issues/13060

nkolev92 (approved):

zivkan:
This PR should have a public issue in NuGet/Home, since this affects customers, so we can link customers to this issue, and also it'll appear in the next version of NuGet's release notes.

I believe that it's likely the root cause for this feedback: https://developercommunity.visualstudio.com/t/NuGet-Reports-Vulnerable-Packages-But-Do/10532062

donnie-msft (approved):

PR #5525, opened 2023-11-30
Add three more options to `dotnet package search`

Nigusu-Allehu:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12978

## Description

This PR adds three new options to `dotnet package search`
- `--format`, `--verbosity` , `--configfile`
- In addition. it modifies the `--exact-match` option to list all version of a package rather than the latest version.
Spec: https://github.com/NuGet/Home/pull/12970

### Screenshots
- `dotnet package search nuget --verbosity minimal`
image
- `dotnet package search nuget --verbosity normal`
image
- `dotnet package search nuget --verbosity detailed`
image
- `dotnet package search nuget --verbosity minimal --format json`
image

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/DetailedVerbosityPackage.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/DetailedVerbosityPackage.cs
@@ -8,7 +8,7 @@ using NuGet.Protocol.Core.Types;
 
 namespace NuGet.CommandLine.XPlat
 {
-    internal class PackageSearchResultPackageDetailed : PackageSearchResultPackageNormal
+    internal class DetailedVerbosityPackage : NormalVerbosityPackage
     {
         [JsonPropertyName("description")]
         public string Description { get; set; }
@@ -22,11 +22,11 @@ namespace NuGet.CommandLine.XPlat
         [JsonPropertyName("projectUrl")]
         public Uri ProjectUrl { get; set; }
 
-        public PackageSearchResultPackageDetailed() : base()
+        public DetailedVerbosityPackage() : base()
         {
         }
 
-        public PackageSearchResultPackageDetailed(IPackageSearchMetadata packageSearchMetadata, string deprecation) : base(packageSearchMetadata)
+        public DetailedVerbosityPackage(IPackageSearchMetadata packageSearchMetadata, string deprecation) : base(packageSearchMetadata)
         {
             Description = packageSearchMetadata.Description;
 
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/IPackageSearchResultRenderer.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/IPackageSearchResultRenderer.cs
@@ -28,11 +28,17 @@ namespace NuGet.CommandLine.XPlat
         /// <param name="source">The source</param>
         /// <param name="error">The error message to be rendered</param>
         /// <returns></returns>
-        void Add(PackageSource source, string error);
+        void Add(PackageSource source, PackageSearchProblem packageSearchProblem);
 
         /// <summary>
         /// Finishes the rendering operation.
         /// </summary>
         void Finish();
+
+        /// <summary>
+        ///  Renders a problem that prevented the search from happening at all.
+        /// </summary>
+        /// <param name="packageSearchProblem"></param>
+        void RenderProblem(PackageSearchProblem packageSearchProblem);
     }
 }
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchArgs.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchArgs.cs
@@ -19,16 +19,20 @@ namespace NuGet.CommandLine.XPlat
         public bool Interactive { get; set; }
         public ILoggerWithColor Logger { get; set; }
         public string SearchTerm { get; set; }
+        public PackageSearchVerbosity Verbosity { get; set; } = PackageSearchVerbosity.Normal;
+        public bool JsonFormat { get; set; } = false;
 
-        public PackageSearchArgs(string skip, string take)
+        public PackageSearchArgs(string skip, string take, string format, string verbosity)
         {
-            Skip = VerifyInt(skip, DefaultSkip);
-            Take = VerifyInt(take, DefaultTake);
+            Skip = VerifyInt(skip, DefaultSkip, "--skip");
+            Take = VerifyInt(take, DefaultTake, "--take");
+            JsonFormat = VerifyFormat(format);
+            Verbosity = VerifyVerbosity(verbosity);
         }
 
         public PackageSearchArgs() { }
 
-        public int VerifyInt(string number, int defaultValue)
+        public int VerifyInt(string number, int defaultValue, string option)
         {
             if (string.IsNullOrEmpty(number))
             {
@@ -40,7 +44,53 @@ namespace NuGet.CommandLine.XPlat
                 return verifiedNumber;
             }
 
-            throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_invalid_number, number));
+            throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_invalidOptionValue, number, option));
+        }
+
+        public bool VerifyFormat(string format)
+        {
+            if (!string.IsNullOrEmpty(format))
+            {
+                if (string.Equals(format, "json", StringComparison.CurrentCultureIgnoreCase))
+                {
+                    return true;
+                }
+                else if (string.Equals(format, "table", StringComparison.CurrentCultureIgnoreCase))
+                {
+                    return false;
+                }
+                else
+                {
+                    throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_invalidOptionValue, format, "--format"));
+                }
+            }
+
+            return false;
+        }
+
+        private PackageSearchVerbosity VerifyVerbosity(string verbosity)
+        {
+            if (verbosity != null)
+            {
+                if (string.Equals(verbosity, nameof(PackageSearchVerbosity.Detailed), StringComparison.CurrentCultureIgnoreCase))
+                {
+                    return PackageSearchVerbosity.Detailed;
+                }
+                else if (string.Equals(verbosity, nameof(PackageSearchVerbosity.Normal), StringComparison.CurrentCultureIgnoreCase))
+                {
+                    return PackageSearchVerbosity.Normal;
+                }
+                else if (string.Equals(verbosity, nameof(PackageSearchVerbosity.Minimal), StringComparison.CurrentCultureIgnoreCase))
+                {
+                    return PackageSearchVerbosity.Minimal;
+                }
+                else
+                {
+                    throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_invalidOptionValue, verbosity, "--verbosity"));
+                }
+            }
+
+            return PackageSearchVerbosity.Normal;
         }
     }
 }
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchArgs.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchArgs.cs
@@ -44,7 +44,7 @@ namespace NuGet.CommandLine.XPlat
                 return verifiedNumber;
             }
 
-            throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_invalidOptionValue, number, option));
+            throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_InvalidOptionValue, number, option));
         }
 
         private PackageSearchFormat GetFormatFromOption(string format)
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchArgs.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchArgs.cs
@@ -20,13 +20,13 @@ namespace NuGet.CommandLine.XPlat
         public ILoggerWithColor Logger { get; set; }
         public string SearchTerm { get; set; }
         public PackageSearchVerbosity Verbosity { get; set; } = PackageSearchVerbosity.Normal;
-        public bool JsonFormat { get; set; } = false;
+        public PackageSearchFormat Format { get; set; } = PackageSearchFormat.Table;
 
         public PackageSearchArgs(string skip, string take, string format, string verbosity)
         {
             Skip = VerifyInt(skip, DefaultSkip, "--skip");
             Take = VerifyInt(take, DefaultTake, "--take");
-            JsonFormat = IsJsonFormat(format);
+            Format = GetFormatFromOption(format);
             Verbosity = GetVerbosityFromOption(verbosity);
         }
 
@@ -47,31 +47,28 @@ namespace NuGet.CommandLine.XPlat
             throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_invalidOptionValue, number, option));
         }
 
-        private bool IsJsonFormat(string format)
+        private PackageSearchFormat GetFormatFromOption(string format)
         {
-            if (!string.IsNullOrEmpty(format) && string.Equals(format, "json", StringComparison.CurrentCultureIgnoreCase))
+            PackageSearchFormat packageSearchFormat = PackageSearchFormat.Table;
+
+            if (!string.IsNullOrEmpty(format) && !Enum.TryParse(format, ignoreCase: true, out packageSearchFormat))
             {
-                return true;
+                packageSearchFormat = PackageSearchFormat.Table;
             }
 
-            return false;
+            return packageSearchFormat;
         }
 
         private PackageSearchVerbosity GetVerbosityFromOption(string verbosity)
         {
-            if (verbosity != null)
+            PackageSearchVerbosity packageSearchVerbosity = PackageSearchVerbosity.Normal;
+
+            if (!string.IsNullOrEmpty(verbosity) && !Enum.TryParse(verbosity, ignoreCase: true, out packageSearchVerbosity))
             {
-                if (string.Equals(verbosity, nameof(PackageSearchVerbosity.Detailed), StringComparison.CurrentCultureIgnoreCase))
-                {
-                    return PackageSearchVerbosity.Detailed;
-                }
-                else if (string.Equals(verbosity, nameof(PackageSearchVerbosity.Minimal), StringComparison.CurrentCultureIgnoreCase))
-                {
-                    return PackageSearchVerbosity.Minimal;
-                }
+                packageSearchVerbosity = PackageSearchVerbosity.Normal;
             }
 
-            return PackageSearchVerbosity.Normal;
+            return packageSearchVerbosity;
         }
     }
 }
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchArgs.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchArgs.cs
@@ -26,13 +26,13 @@ namespace NuGet.CommandLine.XPlat
         {
             Skip = VerifyInt(skip, DefaultSkip, "--skip");
             Take = VerifyInt(take, DefaultTake, "--take");
-            JsonFormat = VerifyFormat(format);
-            Verbosity = VerifyVerbosity(verbosity);
+            JsonFormat = IsJsonFormat(format);
+            Verbosity = GetVerbosityFromOption(verbosity);
         }
 
         public PackageSearchArgs() { }
 
-        public int VerifyInt(string number, int defaultValue, string option)
+        private int VerifyInt(string number, int defaultValue, string option)
         {
             if (string.IsNullOrEmpty(number))
             {
@@ -47,28 +47,17 @@ namespace NuGet.CommandLine.XPlat
             throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_invalidOptionValue, number, option));
         }
 
-        public bool VerifyFormat(string format)
+        private bool IsJsonFormat(string format)
         {
-            if (!string.IsNullOrEmpty(format))
+            if (!string.IsNullOrEmpty(format) && string.Equals(format, "json", StringComparison.CurrentCultureIgnoreCase))
             {
-                if (string.Equals(format, "json", StringComparison.CurrentCultureIgnoreCase))
-                {
-                    return true;
-                }
-                else if (string.Equals(format, "table", StringComparison.CurrentCultureIgnoreCase))
-                {
-                    return false;
-                }
-                else
-                {
-                    throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_invalidOptionValue, format, "--format"));
-                }
+                return true;
             }
 
             return false;
         }
 
-        private PackageSearchVerbosity VerifyVerbosity(string verbosity)
+        private PackageSearchVerbosity GetVerbosityFromOption(string verbosity)
         {
             if (verbosity != null)
             {
@@ -76,17 +65,9 @@ namespace NuGet.CommandLine.XPlat
                 {
                     return PackageSearchVerbosity.Detailed;
                 }
-                else if (string.Equals(verbosity, nameof(PackageSearchVerbosity.Normal), StringComparison.CurrentCultureIgnoreCase))
-                {
-                    return PackageSearchVerbosity.Normal;
-                }
                 else if (string.Equals(verbosity, nameof(PackageSearchVerbosity.Minimal), StringComparison.CurrentCultureIgnoreCase))
                 {
                     return PackageSearchVerbosity.Minimal;
-                }
-                else
-                {
-                    throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_invalidOptionValue, verbosity, "--verbosity"));
                 }
             }
 
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchCommand.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchCommand.cs
@@ -20,7 +20,7 @@ namespace NuGet.CommandLine.XPlat
             Register(rootCommand, getLogger, SetupSettingsAndRunSearchAsync);
         }
 
-        public static void Register(CliCommand rootCommand, Func<ILoggerWithColor> getLogger, Func<PackageSearchArgs, CancellationToken, Task<int>> setupSettingsAndRunSearchAsync)
+        public static void Register(CliCommand rootCommand, Func<ILoggerWithColor> getLogger, Func<PackageSearchArgs, string, CancellationToken, Task<int>> setupSettingsAndRunSearchAsync)
         {
             var searchCommand = new CliCommand("search", Strings.pkgSearch_Description);
 
@@ -66,6 +66,24 @@ namespace NuGet.CommandLine.XPlat
                 Arity = ArgumentArity.ExactlyOne
             };
 
+            var format = new CliOption<string>("--format")
+            {
+                Description = Strings.pkgSearch_FormatDescription,
+                Arity = ArgumentArity.ExactlyOne
+            };
+
+            var verbosity = new CliOption<string>("--verbosity")
+            {
+                Description = Strings.pkgSearch_VerbosityDescription,
+                Arity = ArgumentArity.ExactlyOne
+            };
+
+            var configFile = new CliOption<string>("--configfile")
+            {
+                Description = Strings.pkgSearch_ConfigFileDescription,
+                Arity = ArgumentArity.ExactlyOne
+            };
+
             var help = new HelpOption()
             {
                 Arity = ArgumentArity.Zero
@@ -78,6 +96,9 @@ namespace NuGet.CommandLine.XPlat
             searchCommand.Options.Add(interactive);
             searchCommand.Options.Add(take);
             searchCommand.Options.Add(skip);
+            searchCommand.Options.Add(format);
+            searchCommand.Options.Add(verbosity);
+            searchCommand.Options.Add(configFile);
             searchCommand.Options.Add(help);
 
             searchCommand.SetAction(async (parserResult, cancelationToken) =>
@@ -86,7 +107,7 @@ namespace NuGet.CommandLine.XPlat
 
                 try
                 {
-                    var packageSearchArgs = new PackageSearchArgs(parserResult.GetValue(skip), parserResult.GetValue(take))
+                    var packageSearchArgs = new PackageSearchArgs(parserResult.GetValue(skip), parserResult.GetValue(take), parserResult.GetValue(format), parserResult.GetValue(verbosity))
                     {
                         Sources = parserResult.GetValue(sources),
                         SearchTerm = parserResult.GetValue(searchTerm),
@@ -96,7 +117,7 @@ namespace NuGet.CommandLine.XPlat
                         Logger = logger,
                     };
 
-                    return await setupSettingsAndRunSearchAsync(packageSearchArgs, cancelationToken);
+                    return await setupSettingsAndRunSearchAsync(packageSearchArgs, parserResult.GetValue(configFile), cancelationToken);
                 }
                 catch (ArgumentException ex)
                 {
@@ -108,13 +129,13 @@ namespace NuGet.CommandLine.XPlat
             rootCommand.Subcommands.Add(searchCommand);
         }
 
-        public static async Task<int> SetupSettingsAndRunSearchAsync(PackageSearchArgs packageSearchArgs, CancellationToken cancellationToken)
+        public static async Task<int> SetupSettingsAndRunSearchAsync(PackageSearchArgs packageSearchArgs, string configFile, CancellationToken cancellationToken)
         {
             DefaultCredentialServiceUtility.SetupDefaultCredentialService(packageSearchArgs.Logger, !packageSearchArgs.Interactive);
 
             ISettings settings = Settings.LoadDefaultSettings(
                 Directory.GetCurrentDirectory(),
-                configFileName: null,
+                configFileName: configFile,
                 machineWideSettings: new XPlatMachineWideSetting());
             PackageSourceProvider sourceProvider = new PackageSourceProvider(settings);
 
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchCommand.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchCommand.cs
@@ -80,7 +80,7 @@ namespace NuGet.CommandLine.XPlat
 
             var configFile = new CliOption<string>("--configfile")
             {
-                Description = Strings.pkgSearch_ConfigFileDescription,
+                Description = Strings.Option_ConfigFile,
                 Arity = ArgumentArity.ExactlyOne
             };
 
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchMainOutput.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchMainOutput.cs
@@ -1,0 +1,27 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Text.Json.Serialization;
+
+namespace NuGet.CommandLine.XPlat
+{
+    internal class PackageSearchMainOutput
+    {
+        [JsonPropertyName("version")]
+        public int Version { get; set; }
+
+        [JsonPropertyName("problems")]
+        public List<PackageSearchProblem> Problems { get; set; }
+
+        [JsonPropertyName("searchResult")]
+        public List<PackageSearchResult> SearchResult { get; set; }
+
+        public PackageSearchMainOutput()
+        {
+            Version = 1;
+            Problems = new List<PackageSearchProblem>();
+            SearchResult = new List<PackageSearchResult>();
+        }
+    }
+}
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchProblem.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchProblem.cs
@@ -1,0 +1,32 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Text.Json.Serialization;
+
+namespace NuGet.CommandLine.XPlat
+{
+    internal class PackageSearchProblem
+    {
+        [JsonPropertyName("text")]
+        public string Text { get; private set; }
+
+        [JsonPropertyName("problemType")]
+        [JsonConverter(typeof(JsonStringEnumConverter))]
+        public PackageSearchProblemType ProblemType { get; }
+
+        private PackageSearchProblem()
+        { }
+
+        internal PackageSearchProblem(PackageSearchProblemType packageSearchProblemType, string text)
+        {
+            ProblemType = packageSearchProblemType;
+            Text = text;
+        }
+
+        public PackageSearchProblem(string text, PackageSearchProblemType problemType)
+        {
+            Text = text;
+            ProblemType = problemType;
+        }
+    }
+}
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResult.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResult.cs
@@ -1,0 +1,32 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Newtonsoft.Json;
+using NuGet.CommandLine.XPlat.Commands.PackageSearch;
+
+namespace NuGet.CommandLine.XPlat
+{
+    /// <summary>
+    /// Represents the result of a package search for a specific source.
+    /// </summary>
+    internal class PackageSearchResult
+    {
+        [JsonProperty("sourceName")]
+        public string SourceName { get; set; }
+
+        [JsonProperty("packages")]
+        public List<IPackageSearchResultPackage> Packages { get; set; }
+
+        public PackageSearchResult(string source)
+        {
+            SourceName = source;
+            Packages = new List<IPackageSearchResultPackage>();
+        }
+
+        public void AddPackage(IPackageSearchResultPackage package)
+        {
+            Packages.Add(package);
+        }
+    }
+}
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResult.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResult.cs
@@ -2,7 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
-using Newtonsoft.Json;
+using System.Text.Json.Serialization;
 using NuGet.CommandLine.XPlat.Commands.PackageSearch;
 
 namespace NuGet.CommandLine.XPlat
@@ -12,10 +12,10 @@ namespace NuGet.CommandLine.XPlat
     /// </summary>
     internal class PackageSearchResult
     {
-        [JsonProperty("sourceName")]
+        [JsonPropertyName("sourceName")]
         public string SourceName { get; set; }
 
-        [JsonProperty("packages")]
+        [JsonPropertyName("packages")]
         public List<IPackageSearchResultPackage> Packages { get; set; }
 
         public PackageSearchResult(string source)
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResult.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResult.cs
@@ -15,6 +15,9 @@ namespace NuGet.CommandLine.XPlat
         [JsonPropertyName("sourceName")]
         public string SourceName { get; set; }
 
+        [JsonPropertyName("errors")]
+        public List<string> Errors { get; set; }
+
         [JsonPropertyName("packages")]
         public List<IPackageSearchResultPackage> Packages { get; set; }
 
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResult.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResult.cs
@@ -15,21 +15,16 @@ namespace NuGet.CommandLine.XPlat
         [JsonPropertyName("sourceName")]
         public string SourceName { get; set; }
 
-        [JsonPropertyName("errors")]
-        public List<string> Errors { get; set; }
+        [JsonPropertyName("problems")]
+        public List<PackageSearchProblem> Problems { get; set; }
 
         [JsonPropertyName("packages")]
-        public List<IPackageSearchResultPackage> Packages { get; set; }
+        public List<ISearchResultPackage> Packages { get; set; }
 
         public PackageSearchResult(string source)
         {
             SourceName = source;
-            Packages = new List<IPackageSearchResultPackage>();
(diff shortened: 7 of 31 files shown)


ghost:

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchArgs.cs
```diff
@@ -40,7 +44,53 @@ public int VerifyInt(string number, int defaultValue)
                 return verifiedNumber;
             }
 
-            throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_invalid_number, number));
+            throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_invalidOptionValue, number, option));
+        }
+
+        public bool VerifyFormat(string format)
```
kartheekp-ms: IMO, this method can be private. Please let me know if I am missing something.

kartheekp-ms: I think the name of this method is a bit confusing because its functionality is to check if the format is JSON or not. I propose using `IsJsonFormat` as a slightly better choice.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchArgs.cs
```diff
@@ -40,7 +44,53 @@ public int VerifyInt(string number, int defaultValue)
...
+                {
+                    return true;
+                }
+                else if (string.Equals(format, "table", StringComparison.CurrentCultureIgnoreCase))
+                {
+                    return false;
+                }
+                else
+                {
+                    throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_invalidOptionValue, format, "--format"));
```
kartheekp-ms: Instead of throwing an exception, can we fall back to the default option and notify the user about the decision?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchArgs.cs
```diff
@@ -40,7 +44,53 @@ public int VerifyInt(string number, int defaultValue)
...
+                {
+                    return PackageSearchVerbosity.Normal;
+                }
+                else if (string.Equals(verbosity, nameof(PackageSearchVerbosity.Minimal), StringComparison.CurrentCultureIgnoreCase))
+                {
+                    return PackageSearchVerbosity.Minimal;
+                }
+                else
+                {
+                    throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_invalidOptionValue, verbosity, "--verbosity"));
```
kartheekp-ms: Instead of throwing an exception, can we fall back to the default option and notify the user about the decision?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchArgs.cs
```diff
@@ -40,7 +44,53 @@ public int VerifyInt(string number, int defaultValue)
...
+                else
+                {
+                    throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_invalidOptionValue, format, "--format"));
+                }
+            }
+
+            return false;
+        }
+
+        private PackageSearchVerbosity VerifyVerbosity(string verbosity)
```
kartheekp-ms: How about remaining `VerifyVerbosity` to `GetVerbosityFromOption` to better reflect its functionality. The new name clearly indicates that the method retrieves a verbosity level based on an option provided.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchCommand.cs
```diff
@@ -66,6 +66,24 @@ public static void Register(CliCommand rootCommand, Func<ILoggerWithColor> getLo
...
 
+            var format = new CliOption<string>("--format")
+            {
+                Description = Strings.pkgSearch_FormatDescription,
+                Arity = ArgumentArity.ExactlyOne
+            };
+
+            var verbosity = new CliOption<string>("--verbosity")
+            {
+                Description = Strings.pkgSearch_VerbosityDescription,
```
kartheekp-ms, line 77: How about leveraging existing string resource that other commands use for verbosity? for example https://github.com/NuGet/NuGet.Client/blob/4ea7c85930a70c19f69293b572136e46040ff745/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Signing/VerifyCommand.cs#L48

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchCommand.cs
```diff
@@ -66,6 +66,24 @@ public static void Register(CliCommand rootCommand, Func<ILoggerWithColor> getLo
...
+                Arity = ArgumentArity.ExactlyOne
+            };
+
+            var configFile = new CliOption<string>("--configfile")
+            {
+                Description = Strings.pkgSearch_ConfigFileDescription,
```
kartheekp-ms: How about leveraging existing string resource that other commands use for config file option? for example

https://github.com/NuGet/NuGet.Client/blob/4ea7c85930a70c19f69293b572136e46040ff745/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Signing/VerifyCommand.cs#L43

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResult.cs
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Newtonsoft.Json;
```
kartheekp-ms: How about using System.Text.Json? If it increases the scope of this PR, then I am okay with creating a follow-up issue. I assume, in general, we would like to move to STJ because it is part of the .NET SDK, but I could be wrong here :)"

kartheekp-ms (commented):
I haven't reviewed all the changes proposed but left few comments to begin with.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchArgs.cs
Nigusu-Allehu: I checked how the other commands are implemented, you are right, they also fall back to the default option. I have fixed it. They do not print a message to notify the user though. Should we do that here? if so, what type should it be? A verbose log maybe?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchCommand.cs
Nigusu-Allehu: I could not use it for this command because there is a difference between the options the two messages provide. For `Verbosity_Description`  Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic]

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResult.cs
Nigusu-Allehu: fixed

ghost:

loic-sharma:
The screenshot for `--format json` has non-JSON output: some stars `*` followed by `Search Result`. Is that intentional or should that be removed? This non-JSON output makes it hard to build tooling around the JSON output.

Nigusu-Allehu:
> The screenshot for `--format json` has non-JSON output: some stars `*` followed by `Search Result`. Is that intentional or should that be removed? This non-JSON output makes it hard to build tooling around the JSON output.

 The asterisks * and `Search Result` you see in the `--format json` screenshot are intentionally placed. The command is designed to query multiple sources by interacting with their APIs, and occasionally, these sources might return errors. It's crucial to capture these issues for troubleshooting, but the process shouldn't stop there. That's why we log the errors first and then use the stars and `Search Result` as a marker. This indicates the beginning of the actual JSON output of the search results, helping you differentiate between the two and ensuring your tooling gets clean data to work with.

loic-sharma:
Could we put errors in the JSON itself? Hypothetical example:

```json
[
  {
    "sourceName": "nuget.org",
    "error": "Zomg search is on 🔥",
    "packages": []
  }
]
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchArgs.cs
```diff
@@ -40,7 +44,34 @@ public int VerifyInt(string number, int defaultValue)
                 return verifiedNumber;
             }
 
-            throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_invalid_number, number));
+            throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_invalidOptionValue, number, option));
+        }
+
+        private bool IsJsonFormat(string format)
```
nkolev92: Would it be easier to use an enum?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchArgs.cs
```diff
@@ -40,7 +44,34 @@ public int VerifyInt(string number, int defaultValue)
...
+            }
+
+            return false;
+        }
+
+        private PackageSearchVerbosity GetVerbosityFromOption(string verbosity)
+        {
+            if (verbosity != null)
+            {
+                if (string.Equals(verbosity, nameof(PackageSearchVerbosity.Detailed), StringComparison.CurrentCultureIgnoreCase))
```
nkolev92: same here, would it be easier to use an enum?

There's probably examples with other options: https://github.com/NuGet/NuGet.Client/blob/58f5aad3fc1684c20798ab28200aa62fe0512df9/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/ListPackageCommand.cs#L174-L182

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResult.cs
```diff
@@ -0,0 +1,32 @@
...
+        [JsonPropertyName("packages")]
+        public List<IPackageSearchResultPackage> Packages { get; set; }
+
+        public PackageSearchResult(string source)
+        {
+            SourceName = source;
+            Packages = new List<IPackageSearchResultPackage>();
+        }
+
+        public void AddPackage(IPackageSearchResultPackage package)
```
nkolev92: If the property is public, do we need the method?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultJsonPrinter.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using NuGet.CommandLine.XPlat.Commands.PackageSearch;
+using NuGet.Configuration;
+using NuGet.Protocol;
+using NuGet.Protocol.Core.Types;
+
+namespace NuGet.CommandLine.XPlat
+{
+    internal class PackageSearchResultJsonPrinter : IPackageSearchResultRenderer
```
nkolev92: nit naming: The interface is `Renderer`, the implementation should probably also be `Renderer`.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultJsonPrinter.cs
```diff
@@ -0,0 +1,94 @@
...
+                    packageSearchResult.AddPackage(new PackageSearchResultPackageNormal(metadata));
+                }
+            }
+
+            _packageSearchResults.Add(packageSearchResult);
+        }
+
+        public void Add(PackageSource source, string error)
+        {
+            _logger.LogMinimal(SourceSeparator);
```
nkolev92: Ideally you don't take a logger in the renderer.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultJsonPrinter.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        public void Finish()
+        {
+            var options = new JsonSerializerOptions
+            {
+                WriteIndented = true,
+                Converters = { new JsonPolymorphicConverter<IPackageSearchResultPackage>() }
+            };
+            var json = JsonSerializer.Serialize(_packageSearchResults, options);
+            _logger.LogMinimal(SourceSeparator);
```
nkolev92: Why do we need tihs?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultPackageDetailed.cs
```diff
@@ -0,0 +1,42 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using System.Text.Json.Serialization;
+using NuGet.Protocol.Core.Types;
+
+namespace NuGet.CommandLine.XPlat
+{
+    internal class PackageSearchResultPackageDetailed : PackageSearchResultPackageNormal
```
nkolev92: There's double package in the name here. 

You can simplify these to 
DetailedPackageSearchResult and NormalPackageSearchResult

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultPackageMinimal.cs
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Text.Json.Serialization;
+using NuGet.CommandLine.XPlat.Commands.PackageSearch;
+using NuGet.Protocol.Core.Types;
+
+namespace NuGet.CommandLine.XPlat
+{
+    internal class PackageSearchResultPackageMinimal : IPackageSearchResultPackage
```
nkolev92: same naming consideration.

martinrrm:
@Nigusu-Allehu Can you check this https://github.com/NuGet/NuGet.Client/pull/5549/files#diff-25501523e030ffb0568e75b67f873e8a3670575ed63d9168c2a72e730c95538dR94-R224 file from my PR? I think this is an improvement from what we have in the branch.

These lines are changing `exitCodeValue = parseResult.InvokeAsync(tokenSource.Token).GetAwaiter().GetResult(); `, adding a help when it throws and better logging of error. Please let me know if you have questions.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchArgs.cs
kartheekp-ms: I am wondering if we should add a warning in this situation. The reason is that customers execute this command in default verbosity mode the majority of the time. If we display it as a warning, customers can take immediate action on it.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchArgs.cs
```diff
@@ -40,7 +44,34 @@ public int VerifyInt(string number, int defaultValue)
...
+        }
+
+        private bool IsJsonFormat(string format)
+        {
+            if (!string.IsNullOrEmpty(format) && string.Equals(format, "json", StringComparison.CurrentCultureIgnoreCase))
+            {
+                return true;
+            }
+
+            return false;
```
jebriede: This should be equivalent and more concise. Be sure to test it.

```suggestion
            return !string.IsNullOrEmpty(format) && string.Equals(format, "json", StringComparison.CurrentCultureIgnoreCase)
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultTablePrinter.cs
```diff
@@ -52,7 +106,7 @@ public void Start()
         /// </summary>
         /// <param name="results">An enumerable of package search metadata to be processed and added to the table.</param>
         /// <param name="table">The table where the results will be added as rows.</param>
-        private static void PopulateTableWithResults(IEnumerable<IPackageSearchMetadata> results, Table table)
+        private async void PopulateTableWithResults(IEnumerable<IPackageSearchMetadata> results, Table table)
```
jebriede: Methods marked async should have the suffix Async:

```suggestion
        private async void PopulateTableWithResultsAsync(IEnumerable<IPackageSearchMetadata> results, Table table)
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -842,9 +842,9 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
   <data name="pkgSearch_termDescription" xml:space="preserve">
     <value>Search term to filter package names, descriptions, and tags. Used as a literal value. Example: `dotnet package search some.package`. See also `--exact-match`.</value>
   </data>
-  <data name="Error_invalid_number" xml:space="preserve">
-    <value>{0} is not a valid integer.</value>
-    <comment>0 - The invalid number.</comment>
+  <data name="Error_invalidOptionValue" xml:space="preserve">
```
jebriede: Other entries in here have an upper case letter after the underscore. Should we be consistent with those?

```suggestion
  <data name="Error_InvalidOptionValue" xml:space="preserve">
```

test/NuGet.Core.Tests/NuGet.CommandLine.Xplat.Tests/PackageSearchCommandTests.cs
```diff
@@ -162,6 +162,85 @@ public void Register_withInvalidSkipOption_ShowsErrorMessage()
...
+
+            // Assert
+            if (format == "json")
+            {
+                Assert.True(CapturedArgs.JsonFormat);
+            }
+            else
+            {
+                Assert.False(CapturedArgs.JsonFormat);
+            }
```
jebriede: Tests should not have conditional branches. This typically means we are writing more than one test. A better practice is to break it into multiple tests. You can also do this with InlineData, passing another argument that is the expected value, add a parameter to the test method, and assert the expected result.

test/NuGet.Core.Tests/NuGet.CommandLine.Xplat.Tests/PackageSearchCommandTests.cs
```diff
@@ -162,6 +162,85 @@ public void Register_withInvalidSkipOption_ShowsErrorMessage()
...
+                Assert.Equal(PackageSearchVerbosity.Minimal, CapturedArgs.Verbosity);
+            }
+            else if (verbosity == "detailed")
+            {
+                Assert.Equal(PackageSearchVerbosity.Detailed, CapturedArgs.Verbosity);
+            }
+            else
+            {
+                Assert.Equal(PackageSearchVerbosity.Normal, CapturedArgs.Verbosity);
+            }
```
jebriede: Same as above, avoid conditional branches in tests and opt for passing expected value in the test data.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultTablePrinter.cs
```diff
@@ -14,18 +16,70 @@ internal class PackageSearchResultTablePrinter : IPackageSearchResultRenderer
...
+        private readonly string[] _minimalVerbosityTableHeaderForExactMatch = { "Package ID", "Version" };
+        private readonly int[] _minimalColumnsToHighlight = { 0 };
+
+        private readonly string[] _normalVerbosityTableHeader = { "Package ID", "Latest Version", "Owners", "Total Downloads" };
+        private readonly string[] _normalVerbosityTableHeaderForExactMatch = { "Package ID", "Version", "Owners", "Total Downloads" };
+        private readonly int[] _normalColumnsToHighlight = { 0, 2 };
+
+        private readonly string[] _detailedVerbosityTableHeader = { "Package ID", "Latest Version", "Owners", "Total Downloads", "Vulnerable", "Deprecation", "Project URL", "Description" };
+        private readonly string[] _detailedVerbosityTableHeaderForExactMatch = { "Package ID", "Version", "Owners", "Total Downloads", "Vulnerable", "Deprecation", "Project URL", "Description" };
+        private readonly int[] _detailedColumnsToHighlight = { 0, 2, 6, 7 };
```
kartheekp-ms: Due to different data being displayed to the customer for various verbosity levels, The `PackageSearchResultTablePrinter` class in its current form is handling too many responsibilities - it decides the table format based on verbosity, creates the table, populates it, and prints it.

One way to separate these responsibilities is by defining separate classes for each verbosity level and creating a factory method to instantiate the appropriate object based on the verbosity level.

1. Define Strategy interface
```cs
interface ITableFormatStrategy
{
    Table CreateTable(IEnumerable<IPackageSearchMetadata> results, bool exactMatch);
}
```

2. Implement strategies for each verbosity Level

```cs
class MinimalTableFormatStrategy : ITableFormatStrategy
{
    public Table CreateTable(IEnumerable<IPackageSearchMetadata> results, bool exactMatch)
    {
        // Implement logic for creating a minimal verbosity table
    }
}

class NormalTableFormatStrategy : ITableFormatStrategy
{
    public Table CreateTable(IEnumerable<IPackageSearchMetadata> results, bool exactMatch)
    {
        // Implement logic for creating a normal verbosity table
    }
}

class DetailedTableFormatStrategy : ITableFormatStrategy
{
    public Table CreateTable(IEnumerable<IPackageSearchMetadata> results, bool exactMatch)
    {
        // Implement logic for creating a detailed verbosity table
    }
}
```

4. Create factory to get appropriate strategy

```cs
class TableFormatStrategyFactory
{
    public static ITableFormatStrategy GetStrategy(PackageSearchVerbosity verbosity)
    {
        switch (verbosity)
        {
            case PackageSearchVerbosity.Minimal:
                return new MinimalTableFormatStrategy();
            case PackageSearchVerbosity.Normal:
                return new NormalTableFormatStrategy();
            case PackageSearchVerbosity.Detailed:
                return new DetailedTableFormatStrategy();
            default:
                throw new ArgumentException("Invalid verbosity level");
        }
    }
}
```

6. Refactor PackageSearchResultTablePrinter

```cs
class PackageSearchResultTablePrinter : IPackageSearchResultRenderer
{
    // Existing fields...

    public PackageSearchResultTablePrinter(string searchTerm, ILoggerWithColor loggerWithColor, PackageSearchVerbosity verbosity, bool exactMatch)
    {
        // Constructor implementation...
    }

    public void Add(PackageSource source, IEnumerable<IPackageSearchMetadata> completedSearch)
    {
        _loggerWithColor.LogMinimal(SourceSeparator);
        _loggerWithColor.LogMinimal($"Source: {source.Name} ({source.SourceUri})");

        ITableFormatStrategy strategy = TableFormatStrategyFactory.GetStrategy(_verbosity);
        Table table = strategy.CreateTable(completedSearch, _exactMatch);
        table.PrintResult(_searchTerm, _loggerWithColor);
    }

    public void Add(PackageSource source, string error)
    {
        // Error handling implementation...
    }

    // Implement other methods...
}
```

test/NuGet.Core.Tests/NuGet.CommandLine.Xplat.Tests/PackageSearchCommandTests.cs
jebriede: As an example, see how Register_WithOptions_SetsExpectedValues has expected values in the InlineData.

test/NuGet.Core.Tests/NuGet.CommandLine.Xplat.Tests/PackageSearchResultJsonPrinterTests.cs
```diff
@@ -0,0 +1,119 @@
...
+      }},
+      {{
+        ""total downloads"": 123456,
+        ""owners"": null,
+        ""id"": ""NuGet.Versioning"",
+        ""latestVersion"": ""4.3.0""
+      }}
+    ]
+  }}
+]";
```
jebriede: This is part of the Arrange step and you could consider extracting this to a constant in this class and referring to it in the test to make the test a little clearer to understand.

test/NuGet.Core.Tests/NuGet.CommandLine.Xplat.Tests/PackageSearchResultJsonPrinterTests.cs
```diff
@@ -0,0 +1,119 @@
...
+                mockLoggerWithColor.Verify(x => x.LogMinimal(expectedTwoJson), Times.Exactly(1));
+            }
+            else if (numberOfPackages == 0)
+            {
+                mockLoggerWithColor.Verify(x => x.LogMinimal(expectedZeroJson), Times.Exactly(1));
+            }
+            else
+            {
+                mockLoggerWithColor.Verify(x => x.LogMinimal(expectedOneJson), Times.Exactly(1));
+            }
```
jebriede: Avoid conditionals in the test.

test/NuGet.Core.Tests/NuGet.CommandLine.Xplat.Tests/PackageSearchRunnerTests.cs
```diff
@@ -80,11 +90,20 @@ public async Task PackageSearchRunner_SearchAPIWithVariousSkipTakePrereleaseOpti
                 Assert.Contains(expected, ColoredMessage.Select(tuple => tuple.Item1));
             }
 
+            foreach (var notExpected in notExpectedValues)
+            {
+                Assert.DoesNotContain(notExpected, ColoredMessage.Select(tuple => tuple.Item1));
+            }
```
jebriede, line 96: Loops in a test method should be avoided as they increase complexity of the test, making them harder to read and might require tests themselves. Instead, opt to use test frameworks that take care of the complexity without having to write loops or conditionals. You can use Fluent Insertions here to test that the list is exactly what you expect it to be:
`List.Should().BeEquivalentTo()` can be used if the list is supposed to be exactly equal to your expected results
`List.Should().ContainInOrder()` can be used if you have other items in the list but you know you expect certain items to appear in the result and in a particular order.

test/NuGet.Core.Tests/NuGet.CommandLine.Xplat.Tests/PackageSearchRunnerTests.cs
```diff
@@ -96,19 +115,272 @@ public async Task PackageSearchRunner_ExactMatchOptionEnabled_OnePackageTableOut
...
+            // Assert
+            foreach (var expected in expectedValues)
+            {
+                Assert.Contains(expected, ColoredMessage.Select(tuple => tuple.Item1));
+            }
+
+            foreach (var notExpected in notExpectedValues)
+            {
+                Assert.DoesNotContain(notExpected, ColoredMessage.Select(tuple => tuple.Item1));
+            }
```
jebriede, line 167: See my comment above about avoiding loops in test methods and a suggestion for how to avoid it.

test/NuGet.Core.Tests/NuGet.CommandLine.Xplat.Tests/PackageSearchRunnerTests.cs
```diff
@@ -96,19 +115,272 @@ public async Task PackageSearchRunner_ExactMatchOptionEnabled_OnePackageTableOut
...
+            await PackageSearchRunner.RunAsync(
+                sourceProvider: sourceProvider,
+                packageSearchArgs,
+                cancellationToken: System.Threading.CancellationToken.None);
+
+            // Assert
+            foreach (var expected in expectedValues)
+            {
+                Assert.Contains(expected, ColoredMessage.Select(tuple => tuple.Item1));
+            }
```
jebriede, line 239: See my comment above about avoiding loops in test methods and a suggestion for how to avoid it.

Nigusu-Allehu:
> Could we put errors in the JSON itself? Hypothetical example:
> 
> ```json
> [
>   {
>     "sourceName": "nuget.org",
>     "error": "Zomg search is on 🔥",
>     "packages": []
>   }
> ]
> ```

Thanks for the suggestion. I agree, this will make it easier for user to digest the output. Updated the spec accordingly : https://github.com/NuGet/Home/pull/13133

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultTablePrinter.cs
aortiz-msft: @Nigusu-Allehu  - Feel free to address this comment in a follow-up PR. cc @kartheekp-ms

test/NuGet.Core.Tests/NuGet.CommandLine.Xplat.Tests/PackageSearchRunnerTests.cs
aortiz-msft: I don't think those would help here, and expanding the loop in-line doesn't seem to really help either. 

@Nigusu-Allehu - Let's have a conversation on these and address as needed in a follow-up PR.

test/NuGet.Core.Tests/NuGet.CommandLine.Xplat.Tests/PackageSearchRunnerTests.cs
aortiz-msft: Same.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchArgs.cs
Nigusu-Allehu: fixed

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchArgs.cs
Nigusu-Allehu: fixed

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultPackageDetailed.cs
Nigusu-Allehu: renamed

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResult.cs
Nigusu-Allehu: fixed

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultJsonPrinter.cs
Nigusu-Allehu: fixed

src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs
```diff
@@ -89,15 +90,19 @@ public static int MainInternal(string[] args, CommandOutputLogger log)
                 tokenSource.CancelAfter(TimeSpan.FromMinutes(DotnetPackageSearchTimeOut));
                 int exitCodeValue = 0;
 
+                CliConfiguration config = new(rootCommand);
+                ParseResult parseResult = rootCommand.Parse(args, config);
+
                 try
```
aortiz-msft, line 96: extra empty line above.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultTablePrinter.cs
kartheekp-ms: I agree with Arturo to address this comment in a follow-up PR. 

Nigusu, please feel free to create a tracking issue, add a link here and resolve the conversation.

src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs
```diff
@@ -89,15 +90,19 @@ public static int MainInternal(string[] args, CommandOutputLogger log)
...
+                    exitCodeValue = parseResult.Invoke();
                 }
                 catch (Exception ex)
                 {
-                    log.LogError(ex.Message);
+                    LogException(ex, log);
+                    var tokenList = parseResult.Tokens.TakeWhile(token => token.Type == CliTokenType.Argument || token.Type == CliTokenType.Command || token.Type == CliTokenType.Directive).Select(t => t.Value).ToList();
+                    tokenList.Add("-h");
+                    rootCommand.Parse(tokenList).Invoke();
                     exitCodeValue = 1;
```
aortiz-msft, line 106: >exitCodeValue = 1; [](http://example.com/codeflow?start=20&length=18)

Does returning failure continue to make sense now that we are falling back to known values?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultJsonRenderer.cs
```diff
@@ -0,0 +1,97 @@
...
+        private ILoggerWithColor _logger;
+        private PackageSearchVerbosity _verbosity;
+        private PackageSearchMainOutput _packageSearchMainOutput;
+
+        public PackageSearchResultJsonRenderer(ILoggerWithColor loggerWithColor, PackageSearchVerbosity verbosity)
+        {
+            _logger = loggerWithColor;
+            _verbosity = verbosity;
+        }
+        public async void Add(PackageSource source, IEnumerable<IPackageSearchMetadata> completedSearch)
```
aortiz-msft, line 26: NIT: missing empty line above.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultTableRenderer.cs
```diff
@@ -0,0 +1,190 @@
...
+            _loggerWithColor.LogMinimal(SourceSeparator);
+            _loggerWithColor.LogMinimal($"Source: {source.Name} ({source.SourceUri})");
+
+            Table table;
+
+            if (_verbosity == PackageSearchVerbosity.Minimal)
+            {
+                if (_exactMatch)
+                {
+                    table = new Table(_minimalColumnsToHighlight, _minimalVerbosityTableHeaderForExactMatch);
```
aortiz-msft, line 53: (Can be addressed afterwards) Consider using the ternary operator ? instead of if statements for assignments. Example:  headerType = _exactMatch ? minimalForExactMatch : minimalTableHeader;

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultTableRenderer.cs
```diff
@@ -0,0 +1,190 @@
...
+                    table.AddRow(packageId, version);
+                }
+                else if (verbosity == PackageSearchVerbosity.Detailed)
+                {
+                    PackageDeprecationMetadata packageDeprecationMetadata = await result.GetDeprecationMetadataAsync();
+                    string vulnerable = "N/A";
+                    string projectUri = "N/A";
+                    string deprecation = "N/A";
+
+                    if (result.Vulnerabilities != null && result.Vulnerabilities.Any())
```
aortiz-msft, line 146: >result.Vulnerabilities.Any() [](http://example.com/codeflow?start=58&length=28)

can we simplify by doing result.Vulnerabilities?.Any()  ?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultTableRenderer.cs
```diff
@@ -0,0 +1,190 @@
...
+                    {
+                        vulnerable = "True";
+                    }
+
+                    if (result.ProjectUrl != null)
```
aortiz-msft, line 151: >!= [](http://example.com/codeflow?start=42&length=2)

can we use "is not null"?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/IPackageSearchResultRenderer.cs
```diff
@@ -28,11 +28,17 @@ internal interface IPackageSearchResultRenderer
...
         /// <summary>
         /// Finishes the rendering operation.
         /// </summary>
         void Finish();
+
+        /// <summary>
+        ///  Renders a problem that prevented the search from happening at all.
+        /// </summary>
+        /// <param name="packageSearchProblem"></param>
+        void RenderProblem(PackageSearchProblem packageSearchProblem);
```
nkolev92, line 42: Why render problem separately from finish?

aortiz-msft (approved):
:shipit:

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchArgs.cs
```diff
@@ -40,7 +44,31 @@ public int VerifyInt(string number, int defaultValue)
...
+            throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_InvalidOptionValue, number, option));
+        }
+
+        private PackageSearchFormat GetFormatFromOption(string format)
+        {
+            PackageSearchFormat packageSearchFormat = PackageSearchFormat.Table;
+
+            if (!string.IsNullOrEmpty(format) && !Enum.TryParse(format, ignoreCase: true, out packageSearchFormat))
+            {
+                packageSearchFormat = PackageSearchFormat.Table;
```
nkolev92, line 56: Enum.TryParse is null safe. 

You could simplify this method to only a few lines: 

```cs
if (!Enum.TryParse(format, ignoreCase: true, out PackageSearchFormat packageSearchFormat))
{
    packageSearchFormat = PackageSearchFormat.Table;
}

return packageSearchFormat;
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchArgs.cs
```diff
@@ -40,7 +44,31 @@ public int VerifyInt(string number, int defaultValue)
...
+            }
+
+            return packageSearchFormat;
+        }
+
+        private PackageSearchVerbosity GetVerbosityFromOption(string verbosity)
```
nkolev92, line 62: same as above, the method could be simplified.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchMainOutput.cs
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Text.Json.Serialization;
+
+namespace NuGet.CommandLine.XPlat
+{
+    internal class PackageSearchMainOutput
```
nkolev92, line 9: naming: `PackageSearchCommandResult` or `SearchCommandResult`? 

Matching the Result naming. 

This is just minor feedback though that can wait or be dropped. We're not committing to a public API, so we can easily change it.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchRunner.cs
```diff
@@ -29,19 +29,41 @@ internal static class PackageSearchRunner
...
+                packageSearchResultRenderer.Finish();
+
                 return 1;
             }
 
             WarnForHTTPSources(listEndpoints, packageSearchArgs.Logger);
 
+            if (listEndpoints == null || listEndpoints.Count == 0)
+            {
+                packageSearchResultRenderer.Start();
```
nkolev92, line 60: Can the start be called at 1 place only?

jebriede (approved):
Great work! Approved. Please follow up on the comments in a separate PR as discussed offline. You can create GitHub issues to track the feedback that can be addressed later. Thanks!

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/DetailedVerbosityPackage.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        public DetailedVerbosityPackage() : base()
+        {
+        }
+
+        public DetailedVerbosityPackage(IPackageSearchMetadata packageSearchMetadata, string deprecation) : base(packageSearchMetadata)
+        {
+            Description = packageSearchMetadata.Description;
+
+            if (packageSearchMetadata.Vulnerabilities != null && packageSearchMetadata.Vulnerabilities.Any())
```
kartheekp-ms, line 33: if `packageSearchMetadata.Vulnerabilities` is of type `List` then `packageSearchMetadata.Vulnerabilities.Count > 0` would be better for performance.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchProblem.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        [JsonPropertyName("text")]
+        public string Text { get; private set; }
+
+        [JsonPropertyName("problemType")]
+        [JsonConverter(typeof(JsonStringEnumConverter))]
+        public PackageSearchProblemType ProblemType { get; }
+
+        private PackageSearchProblem()
+        { }
```
kartheekp-ms, line 18: I am curious to understand the advantage of private constructor? Given that there are other constructors for this class, as far as I know compiler will not create a default public constructor also.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchMainOutput.cs
kartheekp-ms: I have similar feedback. If we ever decide to improve the naming as suggested above, I suggest taking another look at the other types introduced in this PR because the word "Package" has been used in many types.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultTableRenderer.cs
kartheekp-ms: If `result.Vulnerabilities` is of type `List` then please consider `result.Vulnerabilities?.Count > 0`

kartheekp-ms:
I see that the PR has been merged now. Given that the changes have been merged into a feature branch, I have restored it, assuming that we need that branch for creating a PR into the dev branch.

I am also in the middle of reviewing this PR. I will submit my comments once I am done with the review, so that @Nigusu-Allehu can consider fixing them in the follow-up PR.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultJsonRenderer.cs
```diff
@@ -0,0 +1,97 @@
...
+    internal class PackageSearchResultJsonRenderer : IPackageSearchResultRenderer
+    {
+        private ILoggerWithColor _logger;
+        private PackageSearchVerbosity _verbosity;
+        private PackageSearchMainOutput _packageSearchMainOutput;
+
+        public PackageSearchResultJsonRenderer(ILoggerWithColor loggerWithColor, PackageSearchVerbosity verbosity)
+        {
+            _logger = loggerWithColor;
+            _verbosity = verbosity;
```
kartheekp-ms, line 24: I suggest that we obtain either the `MinimalVerbosityPackage` or the `DetailedVerbosityPackage` type based on the value of the `verbosity parameter`. This approach would eliminate the need for the `_verbosity` instance variable and also the if conditional statement loop in the `Add` method https://github.com/NuGet/NuGet.Client/pull/5525/files#diff-6664b04f18b8d014b79266899d6b9ed1a173a22fa2fc6489b3a86e8706dbc00dR32-R44.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultJsonRenderer.cs
```diff
@@ -0,0 +1,97 @@
...
+        public override T Read(ref Utf8JsonReader reader, Type typeToSerialize, JsonSerializerOptions options)
+        {
+            throw new NotImplementedException();
+        }
+
+        public override void Write(Utf8JsonWriter writer, T value, JsonSerializerOptions options)
+        {
+            Type type = value.GetType();
+
+            JsonSerializer.Serialize(writer, value, type, new JsonSerializerOptions());
```
kartheekp-ms, line 94: how about using the `options` argument value instead of `new JsonSerializerOptions()`?

kartheekp-ms (commented):
I was in the middle of reviewing when this PR has been merged.

@Nigusu-Allehu - I am also okay with addressing the comments if applicable in the follow-up PRs. FYI, there are some unresolved comments in https://github.com/dotnet/sdk/pull/36289 PR.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultTablePrinter.cs
Nigusu-Allehu: Fixed here : https://github.com/NuGet/NuGet.Client/pull/5602

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultJsonPrinter.cs
Nigusu-Allehu: In the situation we have multiple sources, we want to log the search results of the source that returns first immediately. That will make sure that a slow source among the sources will not slow down the whole search operation. As a result, the renderer takes a logger so that it can log the search result from a source immediately. 

We could have the renderer return a search result string it constructed at the end and do the logging in the `PackageSearchRunner` class. However, that would mean we have to wait for all the sources to return.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/IPackageSearchResultRenderer.cs
Nigusu-Allehu: There is two problem sections: in the general section, in the source section.

For example, in this JSON output, there is two problems properties. One of them is found as part of the general properties. And the second one is found as part of the "source Name1" properties. If there are any problems that makes the searching operation not-proceed, it will be added in the first problems section. However, if there are problems that originate from a specific source, that will be added to the problems section that is found under the specific source. For example, let's say the search service for `source Name1` is not available, that problem will be added in the problems section for `source Name1`. This `RenderProblem` will add problems that are originate from a specific source. the method `Finish` on the other hand is called once iterating through all the sources is over. As a result, it cannot add per source specific problems to the JSON efficiently.


      {
        "version": 1,
        "problems": [
          {
            "level": "error",
            "text": "An error occurred!"
          }
        ],
        "searchResult": [
          {
            "source": "source Name1",
            "problems": [
              {
                "level": "error",
                "text": "An error occurred!"
              }
            ],
            "packages": [
              {
                "Owners": "",
                "totalDownloads": "",
                "latestVersion": "",
                "id": ""
              }
            ]
          } ]
          }
        ]
      }

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultJsonPrinter.cs
nkolev92: This is the json printer. 
Do you want to log with the regular logger there?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/IPackageSearchResultRenderer.cs
nkolev92: I see. 
So we have 2 classes of problems. I think the name should still be add problem, and not render immediately though.

src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs
Nigusu-Allehu: We reach here after encountering an error as a result we make sure that is communicated to the caller by setting exitCodeValue = 1. The code above prints out a help output in addition to returning an exit code value of 1.

PR #5528, opened 2023-12-04
Add option to enable floating versions in CPM

### LINKED ISSUE #10432: [DCR]: Allow floating versions with Central Package Management (CPM)

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Home/issues/10432

## Description

By default, central package management (CPM) does not allow users to specify floating version (i.e. `1.*`).  CPM is considered an enterprise-level feature and since floating versions can introduce non-deterministic restores we want users to not be able to get situations where restores in different environments could not be identical.

However, the community has left a good amount of feedback that such a feature should be allowed with more details available in https://github.com/NuGet/Home/issues/10432.  This change allows a user to enable floating versions by setting an MSBuild property which remains off by default.

The only place the MSBuild property will be documented is at https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1011 where users will be encouraged to use non floating versions and educated on the potential issues they could encounter.

Original pull request: https://github.com/NuGet/NuGet.Client/pull/5440
This pull request is a revert of a revert plus a change to set the property so that it won't fail DDRITs this time.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
@@ -459,6 +459,7 @@ namespace NuGet.PackageManagement.VisualStudio
                         MSBuildStringUtility.IsTrue(GetPropertySafe(_vsProjectAdapter.BuildProperties, ProjectBuildProperties.RestoreLockedMode))),
                     CentralPackageVersionsEnabled = isCpvmEnabled,
                     CentralPackageVersionOverrideDisabled = GetPropertySafe(_vsProjectAdapter.BuildProperties, ProjectBuildProperties.CentralPackageVersionOverrideEnabled).EqualsFalse(),
+                    CentralPackageFloatingVersionsEnabled = MSBuildStringUtility.IsTrue(GetPropertySafe(_vsProjectAdapter.BuildProperties, ProjectBuildProperties.CentralPackageFloatingVersionsEnabled)),
                     CentralPackageTransitivePinningEnabled = MSBuildStringUtility.IsTrue(GetPropertySafe(_vsProjectAdapter.BuildProperties, ProjectBuildProperties.CentralPackageTransitivePinningEnabled)),
                     RestoreAuditProperties = auditProperties,
                 }
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
@@ -459,7 +459,7 @@ namespace NuGet.PackageManagement.VisualStudio
                         MSBuildStringUtility.IsTrue(GetPropertySafe(_vsProjectAdapter.BuildProperties, ProjectBuildProperties.RestoreLockedMode))),
                     CentralPackageVersionsEnabled = isCpvmEnabled,
                     CentralPackageVersionOverrideDisabled = GetPropertySafe(_vsProjectAdapter.BuildProperties, ProjectBuildProperties.CentralPackageVersionOverrideEnabled).EqualsFalse(),
-                    CentralPackageFloatingVersionsEnabled = MSBuildStringUtility.IsTrue(GetPropertySafe(_vsProjectAdapter.BuildProperties, ProjectBuildProperties.CentralPackageFloatingVersionsEnabled)),
+                    CentralPackageFloatingVersionsEnabled = MSBuildStringUtility.IsTrue(_vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.CentralPackageFloatingVersionsEnabled)),
                     CentralPackageTransitivePinningEnabled = MSBuildStringUtility.IsTrue(GetPropertySafe(_vsProjectAdapter.BuildProperties, ProjectBuildProperties.CentralPackageTransitivePinningEnabled)),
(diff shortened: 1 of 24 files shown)


nkolev92 (dismissed):
I just reviewed: https://github.com/NuGet/NuGet.Client/pull/5528/commits/6809e899c66a80dbc718fac074d22aedb54f9a0d :D

donnie-msft (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
```diff
@@ -459,6 +459,7 @@ private async Task<PackageSpec> GetPackageSpecAsync(ISettings settings)
                         MSBuildStringUtility.IsTrue(GetPropertySafe(_vsProjectAdapter.BuildProperties, ProjectBuildProperties.RestoreLockedMode))),
                     CentralPackageVersionsEnabled = isCpvmEnabled,
                     CentralPackageVersionOverrideDisabled = GetPropertySafe(_vsProjectAdapter.BuildProperties, ProjectBuildProperties.CentralPackageVersionOverrideEnabled).EqualsFalse(),
+                    CentralPackageFloatingVersionsEnabled = MSBuildStringUtility.IsTrue(GetPropertySafe(_vsProjectAdapter.BuildProperties, ProjectBuildProperties.CentralPackageFloatingVersionsEnabled)),
```
zivkan: `GetPropertySafe` uses `GetPropertyValueWithDteFallback`, which should not be used for any new properties.

Please update this so that it calls `projectBuildProperties.GetPropertyValue(string)` instead (the method without the DTE fallbacl)

zivkan (changes requested):
`GetPropertyValueWithDteFallback` should not be used by any new MSBuild properties. As a team, we need to find a way to stop people blindly copy/pasting "old code" that uses the DTE fallback, so we can slowly migrate towards a no-DTE future.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
jeffkl: Thanks, done!

nkolev92 (approved):

zivkan (approved):
If you have any ideas on how we can reduce the risk of someone using `GetPropertySafe` or `GetPropertyWithDteFallback` the next time anyone adds a new property, then the whole team will benefit.

Unfortunately I don't have any ideas, other then hoping that someone knowledgeable notices during pull requests, which is at high risk of human error.

jeffkl:
> If you have any ideas on how we can reduce the risk of someone using `GetPropertySafe` or `GetPropertyWithDteFallback` the next time anyone adds a new property, then the whole team will benefit.
> 
> Unfortunately I don't have any ideas, other then hoping that someone knowledgeable notices during pull requests, which is at high risk of human error.

Could we mark the implementation as obsolete and add an exception for existing calls?  We could put it in an `#if DEBUG` so it would only fail our local build if someone tried to use it going forward.

zivkan:
> Could we mark the implementation as obsolete and add an exception for existing calls? We could put it in an #if DEBUG so it would only fail our local build if someone tried to use it going forward.

I don't see how that would work. Consider we already had a block of:

```csharp
#pragma disable ObsoleteWarning
var featureThing = GetPropertySafe("FeatureThing");
var featureStuff = GetPropertySafe("FeatureStuff");
var featureWhatever = GetPropertySafe("FeatureWhatever");
#pragma enable ObsoleteWarning
```

Now someone wants to add a property `featureAnotherOne`. I think they're likely to just copy paste one of the existing lines, where the obsolete warning has already been disabled.

Especially if you consider that the legacy PR project class has dozens of lines getting properties. Unless we have the obsolete warning suppression/re-enable every few lines, then someone modifying code right in the middle of the block is unlikely to even see the obsolete warning being suppressed.

On the other hand, maybe doing it anyway at least has a chance that someone will notice that `GetPropertySafe` and `GetPropertyValueWithDteFallback` are marked obsolete. Especially if we also make sure XMLDoc says to use the other method, so mouse hover intellisense is also a way this information can be discovered.

Previously I was letting perfect be the enemy of good. Apologies to @kartheekp-ms who brought up the same idea earlier, and I similarly said at the time I didn't think it would work. It's still an improvement, so better than doing nothing.

I'm happy to do this in another PR if you like.

naiksujit:
Hi @jeffkl :  Can we use this with projects that are compiled with netSDK6?

jeffkl:
As long as you're building with the latest .NET SDK or Visual Studio, you can still compile applications and libraries for .NET 6 and use this functionality.  If you're using an older version of .NET SDK or Visual Studio, this functionality will not exist and can not be used.

csi-lund:
Thanks for the work re-enabling this. come across this now and I think having the optional is helpful for trusted packages and when we want to accept the risk.

PR #5529, opened 2023-12-04
Migrate from using Newtson.Json to System.Text.Json for parsing the assets file.

### LINKED ISSUE #12715: Stop using JObject in assets file reading to reduce allocations
https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1817651

jgonz120:
## Bug

## Description

Currently we use the NewtsonJson's JObject implementation for parsing the projectAssets.json. This implementation isn't memory efficient so we decided to switch to using System.Text.Json (STJ). 

Implementing custom converters in STJ resulted in the JSON file being read at that start, leading to a LOH allocation. To avoid this I implemented a solution based off of [this guide](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/use-utf8jsonreader#read-from-a-stream-using-utf8jsonreader). This will allow us parse through the file without ever having to load the full file at once. 

| Method                          | Runtime              | InputFile    | Mean         | StdDev       | Median       | Ratio | Gen0       | Gen1      | Gen2      | Allocated    | Alloc Ratio |
|-------------------------------- |--------------------- |------------- |-------------:|-------------:|-------------:|------:|-----------:|----------:|----------:|-------------:|------------:|
| 'LockFileFormat read StreamSTJ' | .NET 5.0             | 10KB.json    |     593.2 us |     20.45 us |     593.7 us |  0.77 |     1.9531 |         - |         - |     42.64 KB |        0.21 |
| 'LockFileFormat read NJ'        | .NET 5.0             | 10KB.json    |     771.6 us |     22.54 us |     773.3 us |  1.00 |    11.7188 |    2.9297 |         - |    199.64 KB |        1.00 |
|                                 |                      |              |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET Core 3.1        | 10KB.json    |     629.0 us |      8.61 us |     627.9 us |  0.78 |     1.9531 |         - |         - |     42.33 KB |        0.21 |
| 'LockFileFormat read NJ'        | .NET Core 3.1        | 10KB.json    |     804.7 us |     12.28 us |     804.5 us |  1.00 |    11.7188 |    2.9297 |         - |    198.78 KB |        1.00 |
|                                 |                      |              |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET Framework 4.7.2 | 10KB.json    |     313.1 us |      5.21 us |     312.1 us |  0.70 |     7.8125 |    0.4883 |         - |     50.97 KB |        0.23 |
| 'LockFileFormat read NJ'        | .NET Framework 4.7.2 | 10KB.json    |     447.2 us |      8.28 us |     448.3 us |  1.00 |    35.1563 |   10.2539 |         - |    218.54 KB |        1.00 |
|                                 |                      |              |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET 5.0             | 786KB.json   |   7,437.1 us |     21.04 us |   7,433.8 us |  0.63 |   179.6875 |   85.9375 |         - |   2985.32 KB |        0.41 |
| 'LockFileFormat read NJ'        | .NET 5.0             | 786KB.json   |  11,899.3 us |     62.16 us |  11,887.7 us |  1.00 |   437.5000 |  218.7500 |         - |      7357 KB |        1.00 |
|                                 |                      |              |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET Core 3.1        | 786KB.json   |   7,882.0 us |     59.31 us |   7,888.3 us |  0.60 |   171.8750 |   78.1250 |         - |   2970.14 KB |        0.41 |
| 'LockFileFormat read NJ'        | .NET Core 3.1        | 786KB.json   |  13,205.2 us |    187.11 us |  13,169.7 us |  1.00 |   437.5000 |  218.7500 |         - |    7317.3 KB |        1.00 |
|                                 |                      |              |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET Framework 4.7.2 | 786KB.json   |  10,399.7 us |    124.57 us |  10,409.3 us |  0.60 |   515.6250 |  250.0000 |         - |   3182.19 KB |        0.41 |
| 'LockFileFormat read NJ'        | .NET Framework 4.7.2 | 786KB.json   |  17,258.3 us |    366.31 us |  17,192.9 us |  1.00 |  1250.0000 |  625.0000 |   93.7500 |   7695.95 KB |        1.00 |
|                                 |                      |              |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET 5.0             | 1308KB.json  |  10,769.8 us |     51.83 us |  10,786.7 us |  0.64 |   265.6250 |  125.0000 |         - |   4478.94 KB |        0.47 |
| 'LockFileFormat read NJ'        | .NET 5.0             | 1308KB.json  |  16,737.7 us |    143.36 us |  16,713.1 us |  1.00 |   562.5000 |  281.2500 |         - |   9556.33 KB |        1.00 |
|                                 |                      |              |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET Core 3.1        | 1308KB.json  |  11,831.7 us |     46.65 us |  11,834.1 us |  0.66 |   265.6250 |  125.0000 |         - |   4465.27 KB |        0.47 |
| 'LockFileFormat read NJ'        | .NET Core 3.1        | 1308KB.json  |  17,954.0 us |    268.64 us |  18,024.8 us |  1.00 |   562.5000 |  281.2500 |         - |   9517.85 KB |        1.00 |
|                                 |                      |              |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET Framework 4.7.2 | 1308KB.json  |  15,375.0 us |    130.66 us |  15,337.3 us |  0.62 |   781.2500 |  390.6250 |         - |   4832.49 KB |        0.48 |
| 'LockFileFormat read NJ'        | .NET Framework 4.7.2 | 1308KB.json  |  24,673.5 us |    246.52 us |  24,662.8 us |  1.00 |  1750.0000 |  781.2500 |  218.7500 |  10052.99 KB |        1.00 |
|                                 |                      |              |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET 5.0             | 2756KB.json  |  29,047.0 us |    125.66 us |  29,041.5 us |  0.51 |   718.7500 |  343.7500 |         - |  11956.48 KB |        0.44 |
| 'LockFileFormat read NJ'        | .NET 5.0             | 2756KB.json  |  57,121.3 us |    369.50 us |  57,184.0 us |  1.00 |  1888.8889 | 1000.0000 |  222.2222 |  27482.74 KB |        1.00 |
|                                 |                      |              |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET Core 3.1        | 2756KB.json  |  31,527.0 us |    251.44 us |  31,531.1 us |  0.56 |   687.5000 |  312.5000 |         - |  11908.92 KB |        0.44 |
| 'LockFileFormat read NJ'        | .NET Core 3.1        | 2756KB.json  |  56,282.2 us |    569.83 us |  56,214.5 us |  1.00 |  1555.5556 |  777.7778 |  111.1111 |  27358.63 KB |        1.00 |
|                                 |                      |              |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET Framework 4.7.2 | 2756KB.json  |  47,430.5 us |    404.78 us |  47,439.5 us |  0.56 |  2181.8182 |  909.0909 |  272.7273 |  12596.34 KB |        0.44 |
| 'LockFileFormat read NJ'        | .NET Framework 4.7.2 | 2756KB.json  |  85,334.2 us |  1,111.29 us |  84,782.2 us |  1.00 |  5142.8571 | 2285.7143 |  857.1429 |  28617.17 KB |        1.00 |
|                                 |                      |              |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET 5.0             | 11527KB.json | 201,285.2 us |  3,929.03 us | 202,010.5 us |  0.59 |  4333.3333 | 2666.6667 | 1000.0000 |  63626.85 KB |        0.47 |
| 'LockFileFormat read NJ'        | .NET 5.0             | 11527KB.json | 364,149.1 us | 34,805.34 us | 352,503.5 us |  1.00 |  9000.0000 | 5000.0000 | 1000.0000 | 136493.16 KB |        1.00 |
|                                 |                      |              |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET Core 3.1        | 11527KB.json | 236,958.2 us |  4,460.41 us | 235,785.6 us |  0.63 |  4333.3333 | 2666.6667 | 1000.0000 |  63415.04 KB |        0.47 |
| 'LockFileFormat read NJ'        | .NET Core 3.1        | 11527KB.json | 375,601.1 us | 10,888.11 us | 373,736.1 us |  1.00 |  9000.0000 | 5000.0000 | 1000.0000 | 135976.18 KB |        1.00 |
|                                 |                      |              |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET Framework 4.7.2 | 11527KB.json | 295,284.9 us |  7,029.31 us | 292,850.2 us |  0.71 | 11500.0000 | 4500.0000 | 1500.0000 |  66077.99 KB |        0.47 |
| 'LockFileFormat read NJ'        | .NET Framework 4.7.2 | 11527KB.json | 415,378.6 us |  7,298.57 us | 415,826.5 us |  1.00 | 24000.0000 | 9000.0000 | 2000.0000 | 140948.86 KB |        1.00 |
## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A


zivkan:
FYI @davkean

davkean:
Exciting, I will take a look tomorrow.

src/NuGet.Core/NuGet.ProjectModel/ListObjectConvertor.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Collections.Generic;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace NuGet.ProjectModel
+{
+    /// <summary>
+    /// A <see cref="JsonConverter{T}"/> to allow System.Text.Json to read/write <see cref="List{T}"/> where the list is setup as an object
+    /// </summary>
+    internal class ListObjectConvertor<T> : JsonConverter<IList<T>>
```
zivkan, line 14: This converter no longer appears to be used, so should be deleted.

src/NuGet.Core/NuGet.ProjectModel/FileFormatException.cs
```diff
@@ -101,32 +111,50 @@ internal static FileFormatException Create(string message, int line, int column,
...
+            return new FileFormatException(message, exception)
+                .WithFilePath(path)
+                .WithLineInfo(exception);
+        }
 
+        internal static FileFormatException Create(System.Text.Json.JsonException exception, string path)
+        {
             string message;
-            if (jex == null)
+            if (exception.Data.Contains(SurfaceMessage))
```
davkean, line 131: Comment for this? The _why_ is missing.

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.cs
```diff
@@ -49,1757 +37,46 @@ public static PackageSpec GetPackageSpec(string json, string name, string packag
...
-                }
-                catch (Exception ex)
-                {
-                    throw FileFormatException.Create(ex, frameworkLine, frameworkColumn, packageSpec.FilePath);
-                }
-            });
         }
 
-        private static void ReadImports(PackageSpec packageSpec, JsonTextReader jsonReader, TargetFrameworkInformation targetFrameworkInformation)
+        [Obsolete("This method is obsolete and will be removed in a future release.")]
```
davkean, line 60: Love it, stops the bleeding.

src/NuGet.Core/NuGet.ProjectModel/ListObjectConvertor.cs
```diff
@@ -0,0 +1,60 @@
...
+            //We use JsonObjects for the arrays so we advance to the first property in the object which is the name/ver of the first library
+            reader.Read();
+
+            if (reader.TokenType == JsonTokenType.EndObject)
+            {
+                return new List<T>(0);
+            }
+
+            var objectConverter = (JsonConverter<T>)options.GetConverter(typeof(T));
+            var listObjects = new List<T>();
```
davkean, line 37: What's the common amount of elements for this? 

This is currently using the default which is 4, if the common size diffs from this, either we will resize too many times causing allocations, or make this collection too large. What's the chances that the BCL picked the same default as what your common amount of elements are?

Also, its a low amount, then I would consider something like editor's `FrugalList<T>` which is smaller than `List<T>`.

src/NuGet.Core/NuGet.ProjectModel/ListObjectConvertor.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public override void Write(Utf8JsonWriter writer, IList<T> value, JsonSerializerOptions options)
+        {
+            if (value is null)
+            {
+                writer.WriteNullValue();
+            }
+            else
+            {
+                writer.WriteStringValue("hi");
```
davkean, line 56: I'm guessing this isn't used? Maybe throw/assert from this?

src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileFormat.cs
```diff
@@ -86,17 +90,29 @@ public LockFile Read(Stream stream, string path)
 
         public LockFile Read(Stream stream, ILogger log, string path)
         {
-            using (var textReader = new StreamReader(stream))
+            var useNj = EnvironmentVariableWrapper.Instance.GetEnvironmentVariable("NUGET_EXPERIMENTAL_USE_NJ_FOR_FILE_PARSING");
```
davkean, line 93: Can we cache this? I suspect this will show up on radar.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamAssetsLogMessageConverter.cs
```diff
@@ -0,0 +1,131 @@
...
+                else
+                {
+                    reader.TrySkip();
+                }
+            }
+            if (isValid)
+            {
+                var assetLogMessage = new AssetsLogMessage(level, code, message)
+                {
+                    TargetGraphs = targetGraphs ?? new List<string>(0),
```
davkean, line 117: Array.Empty<string>(), I think as this is typed as IReadOnlyList<string>.

src/NuGet.Core/NuGet.ProjectModel/ListObjectConvertor.cs
```diff
@@ -0,0 +1,60 @@
...
+            {
+                return new List<T>(0);
+
+            }
+            //We use JsonObjects for the arrays so we advance to the first property in the object which is the name/ver of the first library
+            reader.Read();
+
+            if (reader.TokenType == JsonTokenType.EndObject)
+            {
+                return new List<T>(0);
```
davkean, line 33: This is returning IList<T>, so this could be Array.Empty<T>();

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamAssetsLogMessageConverter.cs
```diff
@@ -0,0 +1,131 @@
...
+            string libraryId = default;
+            IReadOnlyList<string> targetGraphs = null;
+
+            while (reader.Read() && reader.TokenType == JsonTokenType.PropertyName)
+            {
+                if (!isValid)
+                {
+                    reader.TrySkip();
+                }
+                if (reader.ValueTextEquals(Utf8Level))
```
davkean, line 56: Can pattern matching via a switch expression make this a little nicer?

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileItemConverter.cs
```diff
@@ -0,0 +1,51 @@
...
+                lockFileItem = new LockFileItem(lockItemPath);
+            }
+
+            reader.Read();
+            if (reader.TokenType == JsonTokenType.StartObject)
+            {
+                while (reader.Read() && reader.TokenType == JsonTokenType.PropertyName)
+                {
+                    var propertyName = reader.GetString();
+                    lockFileItem.Properties[propertyName] = reader.ReadNextTokenAsString();
```
davkean, line 44: Can these items have empty properties? if so, I'd handled the empty path by passing the lock file item a shared empty readonly-dictionary on creation.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetConverter.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            if (reader.TokenType != JsonTokenType.PropertyName)
+            {
+                throw new JsonException("Expected PropertyName, found " + reader.TokenType);
+            }
+
+            var lockFileTarget = new LockFileTarget();
+            //We want to read the property name right away
+            var propertyName = reader.GetString();
+            var parts = propertyName.Split(JsonUtility.PathSplitChars, 2);
```
davkean, line 24: Simple pattern: can we avoid the extra array: https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/24181/Garbage-Collection?anchor=%E2%9D%8C-**avoid**-using-%60string.split%60-on-simple-patterns?

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetConverter.cs
```diff
@@ -0,0 +1,37 @@
...
+        public override LockFileTarget Read(ref Utf8JsonStreamReader reader)
+        {
+            if (reader.TokenType != JsonTokenType.PropertyName)
+            {
+                throw new JsonException("Expected PropertyName, found " + reader.TokenType);
+            }
+
+            var lockFileTarget = new LockFileTarget();
+            //We want to read the property name right away
+            var propertyName = reader.GetString();
```
davkean, line 23: Can we do the same trick for target frameworks as we are doing for the common property names where we compare them as UTF8? We're very likely parsing the same target frameworks over and over again.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetLibraryConverter.cs
```diff
@@ -0,0 +1,203 @@
...
+        private static readonly byte[] Utf8RuntimeTargets = Encoding.UTF8.GetBytes("runtimeTargets");
+        private static readonly byte[] Utf8Tools = Encoding.UTF8.GetBytes("tools");
+        private static readonly byte[] Utf8Embed = Encoding.UTF8.GetBytes("embed");
+        private static readonly byte[] Utf8FrameworkReferences = Encoding.UTF8.GetBytes("frameworkReferences");
+
+        private IList<PackageDependency> ReadPackageDependencyList(ref Utf8JsonStreamReader reader)
+        {
+            if (reader.TokenType != JsonTokenType.StartObject)
+            {
+                return new List<PackageDependency>(0);
```
davkean, line 37: `Array.Empty<PackageDependency>()`

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetLibraryConverter.cs
```diff
@@ -0,0 +1,203 @@
...
+            }
+
+            var packageDependencies = new List<PackageDependency>();
```
davkean, line 40: Better default capacity based on most common value? This is defaulting to 4.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetLibraryConverter.cs
```diff
@@ -0,0 +1,203 @@
...
+            while (reader.Read() && reader.TokenType == JsonTokenType.PropertyName)
+            {
+                string propertyName = reader.GetString();
+                string versionString = reader.ReadNextTokenAsString();
```
davkean, line 44: I imagine version strings and ranges are the same throughout the assets file, can we dedupe them (via `Dictionary<Memory<T>, VersionRange>` or something) so we're only parsing the same bytes once per file?

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetLibraryConverter.cs
```diff
@@ -0,0 +1,203 @@
...
+            {
+                throw new JsonException("Expected PropertyName, found " + reader.TokenType);
+            }
+
+            var lockFileTargetLibrary = new LockFileTargetLibrary();
+
+            //We want to read the property name right away
+            var propertyName = reader.GetString();
+#pragma warning disable CA1307 // Specify StringComparison
+            int slashIndex = propertyName.IndexOf('/');
```
davkean, line 65: Curious why you chose to suppress this rather than just specifying the comparison?

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetLibraryConverter.cs
```diff
@@ -0,0 +1,203 @@
...
+#pragma warning restore CA1307 // Specify StringComparison
+            if (slashIndex == -1)
+            {
+                lockFileTargetLibrary.Name = propertyName;
+            }
+            else
+            {
+                lockFileTargetLibrary.Name = propertyName.Substring(0, slashIndex);
+                lockFileTargetLibrary.Version = NuGetVersion.Parse(propertyName.Substring(slashIndex + 1));
```
davkean, line 74: Ditto on the version number as above.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetLibraryConverter.cs
```diff
@@ -0,0 +1,203 @@
...
+        }
+
+        public override LockFileTargetLibrary Read(ref Utf8JsonStreamReader reader)
+        {
+            if (reader.TokenType != JsonTokenType.PropertyName)
+            {
+                throw new JsonException("Expected PropertyName, found " + reader.TokenType);
+            }
+
+            var lockFileTargetLibrary = new LockFileTargetLibrary();
```
davkean, line 60: I image these are commonly referenced throughout an assets file, can we dedupe these to avoid parsing the same data over and over again?

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonReaderExtensions.cs
```diff
@@ -0,0 +1,233 @@
...
+                    strings.Add(value);
+                }
+            }
+
+            return strings;
+        }
+
+
+        internal static IList<string> ReadStringArrayAsIList(this ref Utf8JsonReader reader, IList<string> strings = null)
+        {
```
davkean, line 126: Can we handle empty and return Array.Empty? Lots of the callers of this pass in a new empty List<string>.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,262 @@
...
+                GetMoreBytesFromStream();
+            }
+            return wasSkipped;
+        }
+
+        internal IList<T> ReadObjectAsList<T>(Utf8JsonStreamReaderConverter<T> streamReaderConverter)
+        {
+            if (TokenType != JsonTokenType.StartObject)
+            {
+                return new List<T>(0);
```
davkean, line 90: Array.Empty<T>();

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,262 @@
...
+
+            }
+            //We use JsonObjects for the arrays so we advance to the first property in the object which is the name/ver of the first library
+            Read();
+
+            if (TokenType == JsonTokenType.EndObject)
+            {
+                return new List<T>(0);
```
davkean, line 98: Ditto, Array.Empty<T>();

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,262 @@
...
+
+        internal List<string> ReadStringArrayAsList(List<string> strings = null)
+        {
+            if (TokenType == JsonTokenType.StartArray)
+            {
+                while (Read() && TokenType != JsonTokenType.EndArray)
+                {
+                    string value = _reader.ReadTokenAsString();
+
+                    strings = strings ?? new List<string>();
```
davkean, line 189: I would love better default capacity on every single List<string> creation based on most likely end result to reduce allocations due to resizes and reduce memory usage.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,262 @@
...
+                    default:
+                        var invalidCastException = new InvalidCastException();
+                        throw new JsonException(invalidCastException.Message, invalidCastException);
+                }
+            }
+
+            return null;
+        }
+
+        internal List<string> ReadStringArrayAsList(List<string> strings = null)
```
davkean, line 181: This looks very similar ReadStringArrayAsList above, unify into the one method?

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,262 @@
...
+
+        internal Utf8JsonStreamReader(Stream stream)
+        {
+            if (stream is null)
+            {
+                throw new ArgumentNullException(nameof(stream));
+            }
+
+            _stream = stream;
+            var firstThreeBytes = new byte[3];
```
davkean, line 30: Any reason this can't be `_buffer`?

src/NuGet.Core/NuGet.ProjectModel/JsonUtility.cs
```diff
@@ -13,6 +13,8 @@ namespace NuGet.ProjectModel
 {
     internal static class JsonUtility
     {
+        private static ReadOnlySpan<byte> Utf8Bom => new byte[] { 0xEF, 0xBB, 0xBF };
```
zivkan, line 16: This doesn't seem to be used, so should be removed.

src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileFormat.cs
zivkan: also, there should be an internal method overload that takes in a `IEnvironmentVariableWrapper`, so we can write tests that run both the NJ and the STJ versions, to make sure they have the same results.  This will be important when we make future changes to the assets file schema, but still support this environment variable to allow using NJ.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamAssetsLogMessageConverter.cs
```diff
@@ -0,0 +1,131 @@
...
+        private static readonly byte[] Utf8Code = Encoding.UTF8.GetBytes(LogMessageProperties.CODE);
+        private static readonly byte[] Utf8WarningLevel = Encoding.UTF8.GetBytes(LogMessageProperties.WARNING_LEVEL);
+        private static readonly byte[] Utf8FilePath = Encoding.UTF8.GetBytes(LogMessageProperties.FILE_PATH);
+        private static readonly byte[] Utf8StartLineNumber = Encoding.UTF8.GetBytes(LogMessageProperties.START_LINE_NUMBER);
+        private static readonly byte[] Utf8StartColumnNumber = Encoding.UTF8.GetBytes(LogMessageProperties.START_COLUMN_NUMBER);
+        private static readonly byte[] Utf8EndLineNumber = Encoding.UTF8.GetBytes(LogMessageProperties.END_LINE_NUMBER);
+        private static readonly byte[] Utf8EndColumnNumber = Encoding.UTF8.GetBytes(LogMessageProperties.END_COLUMN_NUMBER);
+        private static readonly byte[] Utf8Message = Encoding.UTF8.GetBytes(LogMessageProperties.MESSAGE);
+        private static readonly byte[] Utf8LibraryId = Encoding.UTF8.GetBytes(LogMessageProperties.LIBRARY_ID);
+        private static readonly byte[] Utf8TargetGraphs = Encoding.UTF8.GetBytes(LogMessageProperties.TARGET_GRAPHS);
```
zivkan, line 27: I think using `PropertyName` suffix, rather than `Utf8` prefix, will make this easier to understand. Down in the `Read` method, it will be `if (reader.ValueTextEquals(LevelPropertyName))` rather than `if (reader.ValueTextEquals(Utf8Level))`

I can't remember where I first saw this pattern, but another option is to make a private static class called `PropertyNames`. For example:

```csharp
internal class UtfJsonStreamAssetsLogMessageConverter : ...
{
  public override AssetsLogMessage Read(ref Utf8JsonStreamReader reader)
  { 
    // stuff
    if (reader.ValueTextEquals(PropertyName.Level)
   // other stuff
  }

  private static class PropertyName
  {
    public static readonly byte[] Level = Encoding.UTF8.GetBytes(LogMessageProperties.LEVEL);
    // other properties
  }
}
```

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,262 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Text.Json;
+
+namespace NuGet.ProjectModel
+{
+    internal ref struct Utf8JsonStreamReader
```
zivkan, line 13: My gut reaction when I saw this is that it looks like you're copying parts of `Utf8JsonReader` from the .NET runtime repo. But after reading the PR description, and spending more time reading the PR, I see now that there are other reasons. However, once this code is merged, the PR description will no longer be quick to locate. Therefore, I think it's important to comment this class to explain why it exists, why we're not using STJ's `Utf8JsonReader` or `JsonSerializer` directly.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,262 @@
...
+        {
+            if (stream is null)
+            {
+                throw new ArgumentNullException(nameof(stream));
+            }
+
+            _stream = stream;
+            var firstThreeBytes = new byte[3];
+            _stream.Read(firstThreeBytes, 0, 3);
+            _buffer = new byte[1024];
```
zivkan, line 32: Since most filesystems now use block sizes 4k or larger, I think using a smaller buffer size may be inefficient, especially since the vast majority of NuGet's assets files will be much, much larger than 1k (or even 4k).  I think STJ's JsonSerializer starts with a default buffer size of 8k, so maybe we should learn from their expertise and use that as the initial buffer size too.  Even if this class ends up getting reused for other NuGet json files that are not as large as the assets file, I still think that using 4k or 8k initial buffers would be worthwhile.

zivkan, line 32: Did you avoid using an array pool intentionally?  I have mixed feelings about whether benchmark code should count the pooled arrays as part of memory usage (BenchmarkDotNet won't count it towards memory usage if the array pool is used). But for production code, I think it's really important to use an array pool

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,262 @@
...
+            var offset = 0;
+            if (!_utf8Bom.SequenceEqual(firstThreeBytes))
+            {
+                firstThreeBytes.CopyTo(_buffer, 0);
+                offset = 3;
+            }
+            _stream.Read(_buffer, offset, _buffer.Length - offset);
```
zivkan, line 39: The 3 byte array, and 3 byte file read feels like a waste, but that's a gut reaction without an obvious solution.

Is there any way to have `Stream` read to a `Span<byte>` on .NET Framework?  If so, then we can use `Span<byte> firstThreeBytes = stackalloc byte[3]`, and avoid a temporary small array that is eligible for GC almost immediately.

Otherwise another option is to not use this 3 byte array, and just read the first 3 bytes into `_buffer`, and then use `if (_buffer.AsSpan(0, 3).SequenceEqals(_utf8Bom)`, then depending on the result, either fill the rest of the array, or ignore the BOM and overwrite the whole array.

However, I'm wondering if we can avoid needing to do the 3 byte read at all. If we always just read the first `_buffer.Length`, then check the first 3 bytes, if it's the BOM, then set the span start location to 3, otherwise use span start location to 0, and then when you create the Utf8JsonReader, create a Span based on the start location calculated, rather than always using the entire array.

Basically, something like this:

```csharp
byte[] _buffer =new byte[intiailSize];
int read = _stream.Read(_buffer, 0, _buffer.Length);
int start = _utf8Bom.SequenceEqual(_buffer.AsSpan(0, 3)) ? 3 : 0;
_reader = new Utf8JsonReader(_buffer.AsSpan(start, read-start), ...)
```

zivkan, line 39: The number of bytes successfully read is not checked, so I think this will have a bug on a file smaller than `_buffer.Length`

test/NuGet.Core.Tests/NuGet.ProjectModel.Test/DependencyTargetTests.cs
```diff
@@ -159,11 +159,47 @@ public void DependencyTarget_UnknownValueFails()
...
             // Assert
             Assert.NotNull(exception);
-            Assert.Equal("Invalid dependency target value 'blah'.", exception.Message);
+            Assert.Equal("Error reading 'project.json' : Invalid dependency target value 'blah'.", exception.Message);
             Assert.EndsWith("project.json", exception.Path);
-            Assert.Equal(5, exception.Line);
         }
 
+        //[Fact]
+        //public void DependencyTarget_UnknownValueFails()
```
zivkan, line 167: dead code should be deleted. If this was intended to be commented out only temporarily, then it needs to be uncommented.

test/NuGet.Core.Tests/NuGet.ProjectModel.Test/DependencyTargetTests.cs
```diff
@@ -196,11 +232,48 @@ public void DependencyTarget_NonWhiteListValueFails()
...
             Assert.NotNull(exception);
-            Assert.Equal("Invalid dependency target value 'winmd'.", exception.Message);
+            Assert.Equal("Error reading 'project.json' : Invalid dependency target value 'winmd'.", exception.Message);
             Assert.EndsWith("project.json", exception.Path);
-            Assert.Equal(5, exception.Line);
         }
 
+
+        //[Fact]
+        //public void DependencyTarget_NonWhiteListValueFails()
```
zivkan, line 241: see my previous comment about dead code

test/NuGet.Core.Tests/NuGet.ProjectModel.Test/DependencyTargetTests.cs
```diff
@@ -219,25 +292,51 @@ public void DependencyTarget_MultipleValuesFail()
...
             // Assert
             Assert.NotNull(exception);
-            Assert.Equal("Invalid dependency target value 'package,project'.", exception.Message);
+            Assert.Equal("Error reading 'project.json' : Invalid dependency target value 'package,project'.", exception.Message);
             Assert.EndsWith("project.json", exception.Path);
-            Assert.Equal(5, exception.Line);
         }
 
+        //[Fact]
+        //public void DependencyTarget_MultipleValuesFail()
```
zivkan, line 304: see my previous comment about dead code.

test/NuGet.Core.Tests/NuGet.ProjectModel.Test/NjJsonPackageSpecReaderTests.cs
```diff
@@ -20,7 +21,8 @@
 namespace NuGet.ProjectModel.Test
 {
     [UseCulture("")] // Fix tests failing on systems with non-English locales
-    public class JsonPackageSpecReaderTests
+    [Obsolete]
+    public class NjJsonPackageSpecReaderTests
```
zivkan, line 25: As I mentioned in another comment about environment variables, I think it would be better to have a generic `JsonPackageSpecReaderTests` , but modify all the tests to be `[Theory]`, have the method take a `bool useNewtonsoftJson`, then have a single test class test both the Newtonsoft.Json and System.Text.Json code paths. This will make it much easier to ensure that both produce the same outputs for the same inputs. This will be particularlyl important when there are future changes to the assets file schema, as it might be too easy to accidentally change only one of the two parsers, and then customers using the environment variable have a bug.

test/NuGet.Tests.Apex/NuGet.Tests.Apex/Utility/CommonUtility.cs
```diff
@@ -400,7 +400,9 @@ private static LockFile GetAssetsFileWithRetry(string path)
                     {
                         content = File.ReadAllText(path);
                         var format = new LockFileFormat();
+#pragma warning disable CS0612 // Type or member is obsolete
                         return format.Parse(content, path);
+#pragma warning restore CS0612 // Type or member is obsolete
```
zivkan, line 405: this code looks easy to convert to a stream based overload, so it doesn't need to use an obsolete overload. Might make our tests marginally faster too, not needing to load the entire file into a LOH string.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamAssetsLogMessageConverter.cs
```diff
@@ -0,0 +1,131 @@
...
+            int endLineNumber = default;
+            int endColNumber = default;
+            string libraryId = default;
+            IReadOnlyList<string> targetGraphs = null;
+
+            while (reader.Read() && reader.TokenType == JsonTokenType.PropertyName)
+            {
+                if (!isValid)
+                {
+                    reader.TrySkip();
```
zivkan, line 54: What does it mean if `TrySkip` returns false? Do we need to handle it?

Maybe this is a copy/port from the existing NJ code, but if it's not valid, why don't we immediately throw an exception? NuGet writes the assets file, we don't expect 3rd parties to create this file, so it's not obvious to me why would try to "gracefully" handle errors.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,262 @@
...
+        }
+
+        private void GetMoreBytesFromStream()
+        {
+            if (_reader.BytesConsumed < _buffer.Length)
+            {
+                ReadOnlySpan<byte> leftover = _buffer.AsSpan((int)_reader.BytesConsumed);
+                if (leftover.Length == _buffer.Length)
+                {
+                    Array.Resize(ref _buffer, _buffer.Length * 2);
```
zivkan, line 250: [The docs for `Array.Resize`](https://learn.microsoft.com/en-us/dotnet/api/system.array.resize?view=net-8.0#system-array-resize-1(-0()@-system-int32)) says:

> This method allocates a new array with the specified size, copies elements from the old array to the new one, and then replaces the old array with the new one.

If most of the bytes from the buffer are consumed already, there's no point having `Array.Resize` copy them.

```suggestion
                    _buffer = new byte[_buffer.Length * 2];
```


But there is also the array pool question from elsewhere.

zivkan (commented):
Regarding the benchmarks in the PR description, considering the current version of NuGet gets inserted into .NET 8 and VS, I don't think having benchmark results for .NET 5 and .NET Core 3.1 provide a lot of value. The results would be much more meaningful if you showed .NET Framework 4.8 (or 4.7.2), and .NET 8.

I didn't go through everything in detail, but I already have so many comments, some of which might result in significant changes, so I don't want to spend more time on this now. Hopefully my comments are useful, and not annoying!

davkean:
@jgonz120 This is great progress, so happy to see this work. Don't let the plethora of comments take away from this great work.

I super love to get my hands on trace of the benchmarks - before and after would be great, but only need .NET Framework, can you collect one for me using [PerfView](https://github.com/Microsoft/perfview/releases) via Collect -> Collect with these options:

image

This will help me better understand if those comments I've added are problems or not.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
zivkan: I just realized that `GetMoreBytesFromStream` would also have to account for these extra 3 bytes.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
donnie-msft: I like that idea. Since there are many, would you do that inline or would a static function be better to do this in a single place?

src/NuGet.Core/NuGet.ProjectModel/FileFormatException.cs
```diff
@@ -10,6 +10,8 @@ namespace NuGet.ProjectModel
 {
     public class FileFormatException : Exception
     {
+        static internal string SurfaceMessage = "SurfaceMessage";
```
donnie-msft, line 13: Can this be `const` or does it need to be accessible without an exception instance?
Otherwise, I'd suggest readonly:

```suggestion
        static internal readonly string SurfaceMessage = "SurfaceMessage";
```

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamProjectFileDependencyGroupConverter.cs
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Text.Json;
+
+namespace NuGet.ProjectModel
+{
+    /// <summary>
+    /// A <see cref="Utf8JsonStreamReaderConverter{T}"/> to allow read JSON into <see cref="ProjectFileDependencyGroup"/>
```
donnie-msft, line 10: nit
```suggestion
    /// A <see cref="Utf8JsonStreamReaderConverter{T}"/> to allow reading JSON into <see cref="ProjectFileDependencyGroup"/>
```

test/NuGet.Tests.Apex/NuGet.Tests.Apex/Utility/CommonUtility.cs
donnie-msft: 👀

donnie-msft (commented):
Thanks for taking on this work. I know it's been a goal for a while, and it looks like no small feat. 
There's a lot of great feedback, so I'll take another pass later. Not much additional jumped out at me. 
May the force be with you 🖖

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
nkolev92: @davkean This might be interesting for that: https://github.com/NuGet/Home/pull/12124/files#diff-0e899c815846f62b058281b85f22b98df554f592455dcd732e44c72205ab7cc8R246. 

I'm sure @zivkan would love to know what you think.

jgonz120:
I'm going to close this PR and break it out into smaller PRs.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetLibraryConverter.cs
jgonz120: I believe these may vary a bit they're specific to package they're dependent on, and each package will have a different version.

davkean: I just looked at an assets file from https://github.com/dotnet/project-system/tree/main/src/Microsoft.VisualStudio.ProjectSystem.Managed, and `6.0.0` shows up 162 times.

When we look at low memory dumps with NuGet assets in memory, I see millions of version numbers that are identical across a variety of data structures.

jgonz120: @davkean I'm trying to get better acquainted with Memory<T>. We have three scenarios where we can cache the values here. 
VersionRange, NuGetVersion, and NuGetFramework. 

For NuGetVersion, it looks like we're parsing it once we've already retrieved the string for the current token. So I was planning on using Dictionary<string, NuGetVersion>.

For NuGetFramework and VersionRange we could use the current ValueSpan/ValueSequence of the current token. So if we can leverage the readonlyspan<byte> we could do this check and get the corresponding object without allocating a string each time. I'm guessing that was the idea behind the suggestion of using Dictionary<Memory<T>, VersionRange>? Would that mean using Dictionary<Memory<char[]>, VersionRange>, then using an IEqualityComparer to compare the arrays are equal with SequenceEqual?

I found this [open issue](https://github.com/dotnet/runtime/issues/2010) that looks like it's trying to address this.

nkolev92: NuGetFramework already has some optimizations, https://github.com/NuGet/NuGet.Client/blob/fefd3eb3b77936a685260e26dad510b56ac7e66e/src/NuGet.Core/NuGet.Frameworks/NuGetFrameworkFactory.cs#L533. This caches well known frameworks. 

Nothing for version range/version though.

jgonz120: Okay, so framework might not be worth it unless we want to figure out how to avoid the extra string allocation.

PR #19, opened 2015-10-12
Fixes https://github.com/NuGet/Home/issues/1157

danliu:

PR #14, opened 2015-10-08
Implemented nuget.exe add and nuget.exe init commands

deepakaravindr:
Both nuget.exe add and nuget.exe init follows the new v3 model to only copy the nupkg, extract the nuspec and create a SHA512 of the nupkg file at the destination. With a provision to expand the packages using '-Expand' switch

nuget.exe add: https://github.com/NuGet/Home/issues/1421
Adds a given package to the specified source if it is a file source. http sources are not supported.

nuget.exe init: https://github.com/NuGet/Home/issues/1423
Adds all the packages from a given feed to the specified destination feed if it is a file source feed. http feeds are not supported.

Following are pending and/or not covered in this change.
1. nuget.exe init only takes a folder of nupkgs as the source. Should this be upgraded to support every v2 folder source and probably v3 folder sources as well? 
2. nuget.exe init could parallelize package additions.
3. Need more tests for the new '-Expand' switch. Also, need to verify and update the messages for the -Expand switch as needed.
4. Marker file for nuget.exe add and init: https://github.com/NuGet/Home/issues/1456
5. Concurrency for "multiple exes" on other platforms: https://github.com/NuGet/Home/issues/1455
6. With the latest design, Updating nuget clients may not be necessary anymore. We should close https://github.com/NuGet/Home/issues/1425 and https://github.com/NuGet/Home/issues/1424 as Won't Fix.
7. Update documentation on docs.nuget.org

@yishaigalatzer @MeniZalzman @emgarten

PR #13, opened 2015-10-08
implement clearcache command

xavierdecoster:
https://github.com/NuGet/Home/issues/1516

cc @emgarten @yishaigalatzer

PR #5530, opened 2023-12-05
Create class for reading Json files in chunks

jgonz120:
## Bug

Fixes: 

## Description

These classes and structs serve as the base for parsing json objects in memory chunks. I plan on deleting all unnecessary commented code in the final pr. 

### Utf8JsonStreamReader
This struct is a wrapper of the Utf8JsonReader. It maintains the array buffer which stores the current memory chunk being parsed. It wraps the read and write functions from Utf8JsonReader to determine if we've reached the end of the current stream and move the buffer down. 

### Utf8JsonStreamReaderConverter
This is used to facilitate parsing object arrays. STJ uses a struct for the data we are reading, which means we cannot use anonymous functions or delegates anymore. This class will be used more when I create the PR for parsing the asset files.
 
## PR Checklist

- [ ] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Core/NuGet.ProjectModel/Utf8JsonReaderExtensions.cs
```diff
@@ -0,0 +1,231 @@
...
+
+        //            strings.Add(value);
+        //        }
+        //    }
+
+        //    return strings;
+        //}
+
+
+        //internal static IList<string> ReadStringArrayAsIList(this ref Utf8JsonReader reader, IList<string> strings = null)
```
jgonz120: https://github.com/NuGet/NuGet.Client/pull/5529/files#r1414980091

jgonz120: > Can we handle empty and return Array.Empty? Lots of the callers of this pass in a new empty List. -@davkean 

This is currently commented out but I want to address this. The reason I added the option was because some of the classes we're deserializing create an empty list on instantiation, so I was trying to avoid creating multiple empty lists.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonReaderExtensions.cs
```diff
@@ -0,0 +1,231 @@
...
+        //    return (IReadOnlyList<string>)strings ?? Array.Empty<string>();
+        //}
+
+        internal static string ReadTokenAsString(this ref Utf8JsonReader reader)
+        {
+            switch (reader.TokenType)
+            {
+                case JsonTokenType.True:
+                case JsonTokenType.False:
+                    return reader.GetBoolean().ToString();
```
jeffkl: This could probably return `bool.TrueString` or `bool.FalseString` instead to avoid allocations

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,338 @@
...
+            {
+                throw new ArgumentNullException(nameof(stream));
+            }
+
+            _complete = false;
+            _stream = stream;
+            _buffer = ArrayPool<byte>.Shared.Rent(1024);
+            _stream.Read(_buffer, 0, 3);
+            var offset = 0;
+            if (!_utf8Bom.SequenceEqual(_buffer.AsSpan(0, 3)))
```
jeffkl: Do we need to worry about this?  We could just write the file without the BOM to avoid this.  Of course if one is already on disk and people upgrade NuGet, it could fail...

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,338 @@
...
+using System.Buffers;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Text.Json;
+
+namespace NuGet.ProjectModel
+{
+    /// <summary>
```
jgonz120, line 13: https://github.com/NuGet/NuGet.Client/pull/5529/files#r1415139352
> My gut reaction when I saw this is that it looks like you're copying parts of Utf8JsonReader from the .NET runtime repo. But after reading the PR description, and spending more time reading the PR, I see now that there are other reasons. However, once this code is merged, the PR description will no longer be quick to locate. Therefore, I think it's important to comment this class to explain why it exists, why we're not using STJ's Utf8JsonReader or JsonSerializer directly. - @zivkan 

I added a comment let me know what you think

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,338 @@
...
+        internal Utf8JsonStreamReader(Stream stream)
+        {
+            if (stream is null)
+            {
+                throw new ArgumentNullException(nameof(stream));
+            }
+
+            _complete = false;
+            _stream = stream;
+            _buffer = ArrayPool<byte>.Shared.Rent(1024);
```
jgonz120: https://github.com/NuGet/NuGet.Client/pull/5529/files#r1414995822
https://github.com/NuGet/NuGet.Client/pull/5529/files#r1415161019

> Any reason this can't be _buffer? -@davkean  

No reason! Updated.

jgonz120: https://github.com/NuGet/NuGet.Client/pull/5529/files#r1415144302
> Since most filesystems now use block sizes 4k or larger, I think using a smaller buffer size may be inefficient, especially since the vast majority of NuGet's assets files will be much, much larger than 1k (or even 4k). I think STJ's JsonSerializer starts with a default buffer size of 8k, so maybe we should learn from their expertise and use that as the initial buffer size too. Even if this class ends up getting reused for other NuGet json files that are not as large as the assets file, I still think that using 4k or 8k initial buffers would be worthwhile. - @zivkan 

When I was testing this I never found us actually changing the size of the buffer. The buffer is only changed when the individual token being read or skipped exceeds the buffer size. I'm happy to adjust  the buffer size if you think it's still necessary.

jgonz120: https://github.com/NuGet/NuGet.Client/pull/5529/files#r1415147613
> Did you avoid using an array pool intentionally? I have mixed feelings about whether benchmark code should count the pooled arrays as part of memory usage (BenchmarkDotNet won't count it towards memory usage if the array pool is used). But for production code, I think it's really important to use an array pool - @zivkan 

Nope! Just don't have that habit. I updated it to use the buffer and created an additional function so make sure we return it. Do you think there's a better way to implement the return?

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,338 @@
...
+            _stream.Read(_buffer, 0, 3);
+            var offset = 0;
+            if (!_utf8Bom.SequenceEqual(_buffer.AsSpan(0, 3)))
+            {
+                offset = 3;
+            }
+            _stream.Read(_buffer, offset, _buffer.Length - offset);
```
jgonz120: https://github.com/NuGet/NuGet.Client/pull/5529/files#r1415162412
> The number of bytes successfully read is not checked, so I think this will have a bug on a file smaller than _buffer.Length - @zivkan 

I added a unit test to see what would happen with a small JSON test and I didn't get any exceptions.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,338 @@
...
+                GetMoreBytesFromStream();
+            }
+            return wasSkipped;
+        }
+
+        //internal IList<T> ReadObjectAsList<T>(Utf8JsonStreamReaderConverter<T> streamReaderConverter)
+        //{
+        //    if (TokenType != JsonTokenType.StartObject)
+        //    {
+        //        return Array.Empty<T>();
```
jgonz120: https://github.com/NuGet/NuGet.Client/pull/5529/files#r1414981447
> Array.Empty(); - @davkean 

Done for both instances

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,338 @@
...
+                    strings = strings ?? new List<string>();
+
+                    strings.Add(value);
+                }
+            }
+
+            return strings;
+        }
+
+        internal List<string> ReadStringArrayAsList(List<string> strings = null)
```
jgonz120: https://github.com/NuGet/NuGet.Client/pull/5529/files#r1414992003
> This looks very similar ReadStringArrayAsList above, unify into the one method? - @davkean 

Yea, the struggle was when the class being deserialized is creating a new list on instantiation, some of them have an IList, others have List, and others have IReadonlyList.  

I added a comment to make sure I look into merging or changing the classes we were deserializing to make sure they have consistent list implementations.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,338 @@
...
+                throw new InvalidOperationException("Cannot read from completed Utf8JsonStreamReader");
+            }
+
+            if (TokenType == JsonTokenType.StartArray)
+            {
+                while (Read() && TokenType != JsonTokenType.EndArray)
+                {
+                    string value = _reader.ReadTokenAsString();
+
+                    strings = strings ?? new List<string>();
```
jgonz120: https://github.com/NuGet/NuGet.Client/pull/5529/files#r1414989994

> I would love better default capacity on every single List creation based on most likely end result to reduce allocations due to resizes and reduce memory usage. - @davkean 

> I like that idea. Since there are many, would you do that inline or would a static function be better to do this in a single place? - @donnie-msft 

> @davkean This might be interesting for that: https://github.com/NuGet/Home/pull/12124/files#diff-0e899c815846f62b058281b85f22b98df554f592455dcd732e44c72205ab7cc8R246.
I'm sure @zivkan would love to know what you think. - @nkolev92

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,338 @@
...
+
+        private void GetMoreBytesFromStream()
+        {
+            if (_reader.BytesConsumed < _buffer.Length)
+            {
+                ReadOnlySpan<byte> leftover = _buffer.AsSpan((int)_reader.BytesConsumed);
+                if (leftover.Length == _buffer.Length)
+                {
+                    ArrayPool<byte>.Shared.Return(_buffer);
+                    _buffer = ArrayPool<byte>.Shared.Rent(_buffer.Length * 2);
```
jgonz120: https://github.com/NuGet/NuGet.Client/pull/5529/files#r1415240627

> If most of the bytes from the buffer are consumed already, there's no point having Array.Resize copy them. - @zivkan 

done

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: that's the question, I added it as a precaution but I have no idea if it's actually needed.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: I like that solution @nkolev92 otherwise if feels like we're making a guess for what default to use.

nkolev92: yeah, I think the idea would be to make a good guess, but I'm not sure how much better that'd be.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonReaderExtensions.cs
```diff
@@ -0,0 +1,232 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Text.Json;
+
+namespace NuGet.ProjectModel
+{
+    internal static class Utf8JsonReaderExtensions
+    {
+        //private static readonly char[] DelimitedStringDelimiters = new char[] { ' ', ',' };
```
nkolev92: This is tricky.

I think the commented out code should be deleted. 
Right now, it's not self contained, it's making people think about a different PR. 

The idea behind splitting the PR into multiple parts is to allow them to be self contained. 

Note that the standards for feature branches are the same as non feature branches, as feature branches don't get reviewed when they get merged.

jgonz120: I can just remove it now then,

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,299 @@
...
+namespace NuGet.ProjectModel
+{
+    /// <summary>
+    /// This struct is used to read over a memeory stream in parts, in order to avoid reading the entire stream into memory.
+    /// It functions as a wrapper around <see cref="Utf8JsonStreamReader"/>, while maintaining a stream and a buffer to read from.
+    /// must call <see cref="CompleteRead"/> to return the buffer to the pool when completed.
+    /// </summary>
+    internal ref struct Utf8JsonStreamReader
+    {
+        private static readonly char[] DelimitedStringDelimiters = new char[] { ' ', ',' };
```
nkolev92: Where do we use space as a separator in the assets file?

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
nkolev92: Doesn't checking a single assets file we've written enough to confirm what the previous implementation is doing?

I think handling it in the reader is ok though, just curious.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,299 @@
...
+            while (!(wasRead = _reader.Read()) && !_reader.IsFinalBlock)
+            {
+                GetMoreBytesFromStream();
+            }
+            return wasRead;
+        }
+
+        internal bool TrySkip()
+        {
+            if (_complete)
```
nkolev92: Extra these 3 lines to a method and call that method everywhere. 

A lot less duplication :)

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
nkolev92: List implements the other 2, why do we need the other 2?

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: It doesn't look like any of the asset files I have on my computer contain a BOM. 

I rather keep the check, that way if a go to parse a different file we don't risk having it break because we don't check for BOM.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: I don't know if we do, but the [JsonTextReaderExtensions](https://github.com/NuGet/NuGet.Client/blob/1942c17866695ae1175ec3d8b42a6b41bcd6a9c9/src/NuGet.Core/NuGet.ProjectModel/JsonTextReaderExtensions.cs#L12) class has the delimiters defined this way. I was just copying it over.

nkolev92: I see. 
I was just concerned that we might end up splitting something incorrect, but the method is specifically reading a delimited string, so I think it's ok.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: I went through where it's used and we have two cases. ReadStringArrayAsList will be called and we set the property in the object we're creating with the results.

The instances where the objects instantiate the list object have their property set as an IList, so I can't pass in an IList to ReadStringArrayAsList and return a list, without forcing a cast. If we change it to return a List then all the places that are setting it to a list will need a cast. 

I updated it so most of the logic lives in the IList function and the List one just casts the results.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: The default is 0 then it uses for 4 so we could do 4 or 8 just to save the first couple of resizes?

nkolev92: I think 4 vs 0 has the same amount of allocations. 

tbh, I'd just table that for later.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonReaderExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Text.Json;
+
+namespace NuGet.ProjectModel
+{
+    internal static class Utf8JsonReaderExtensions
+    {
+        internal static string ReadTokenAsString(this ref Utf8JsonReader reader)
```
kartheekp-ms, line 11: I am curious about the use of the `ref` keyword for the reader parameter. My understanding is that reference types are already passed by reference, so I'm not sure of the benefit of using ref. Please correct me if my understanding is off.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
zivkan: My comment isn't about resizing, it's about general performance. I have a tendency to overthink things, and be wrong about performance, so this might be wrong, but here are my thoughts:

Consider what has to happen when a program (written in C) reads 1k from a file. The app makes a syscall, so the app then transitions to kernel space, assuming the file contents aren't in the OS filesystem cache, it needs to go to the filesystem driver, the filesystem says "this 1k location is part of this 4k filesystem block", then there's lower level stuff to read the 4k block from disk, then the OS will copy the 1k chunk into the app's buffer, throwing away (well, ignoring) the other 3k. The next time the app wants to read another 1k, the whole thing happens again, although this time the file contents is likely to be in the OS filesystem cache. But it still needs a userspace to kernel transition and back again.

If the app does reads in multiples of 4k, then there will be fewer userspace-kernel transitions, and if the app is careful to do all reads starting from a 4k block boundary (block aligned reads), there will be no partial blocks read either.

With our code, it's a bit more complex, because I don't know what buffering `Stream` does. Also, to make sure that reads are aligned, that start of file, 3 byte BOM, needs to be handled carefully.

Again, I could be way off the mark here, but with my theoretical computer science knowledge, it seems to me that using a 4k or 8k buffer size might help performance, and there's pretty much no downside (honestly, the block aligned reads is probably the most important micro-optimization for perf, but I just don't know how .NET `Stream`'s own buffering affects us).

Also, it looks like I was wrong about System.Text.Json's default buffer size. They seem to use 16k: https://github.com/dotnet/runtime/blob/2b571c505a444522892a272ea7cf6ff485f59da5/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/JsonSerializerOptions.cs#L24

It's a shame they didn't explain why it's that size, but perhaps they have experience we don't, so we can learn from them by using the same buffer size?  Now that we're using the array pool (the same array pool that System.Text.Json uses), then using the same buffer size will pretty much ensure that we'll share buffers with `JsonSerializer` (which NuGetAudit does for the vulnerability database) where as using a different buffer size might result at least one additional allocation.

zivkan: System.Text.Json does it via `IDisposable`: https://github.com/dotnet/runtime/blob/2b571c505a444522892a272ea7cf6ff485f59da5/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/ReadBufferState.cs#L14

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
zivkan: 0 only causes the `List<T>` object to be allocated, where as 4 causes both the `List<T>` object and the backing array to be allocated. This is more allocations than needed for lists that never get any values.

IMO the fundamental problem is that this helper function is used everywhere, for all sorts of parts of the assets file schema, that have very different defaults.  I don't think it's possible to have a sane default here. Instead, the default would need to be passed down from the deserialization code, so can be adjusted for each call's default.

I think we can also consider this a future improvement. Even using List's default, the benchmark shows a significant perf improvement.  We can then use a PerfView trace to see how often `List<T>.Resize` appears on the call stack, or even as a once-off debug, put a `Console.WriteLine(strings.Count)` to see how large the spread of different values are.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonReaderExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+                        return shortValue.ToString();
+                    }
+                    if (reader.TryGetInt32(out int intValue))
+                    {
+                        return intValue.ToString();
+                    }
+                    else if (reader.TryGetInt64(out long longValue))
+                    {
+                        return longValue.ToString();
+                    }
```
zivkan: Was this pattern copied from somewhere?  I don't understand the benefit. 

Consider a number too large for 32 bits. `TryGetInt16` is going to parse the first characters, up to what 16 bits allows, then fail, `TryGetInt32` will parse the same characters in the string then fail, and finally `TryGetInt64` will parse the whole string. The first characters in the string will be parsed multiple times. Granted, the parsing should be relatively fast (subtraction, addition, and multiplication), but still it's duplicated effort when the return type is the same, no matter how large the number is.

Secondly, since since a string is returned, and System.Text.Json already determined it's a number, can we use `reader.GetString()`, and avoid all the parsing and then string formatting CPU costs?  There is an advantage to using `TryGetInt64` and `GetDouble()`, since if the number is too large even for those data types, then we get input validation. Otherwise, parsing the string into an int or double, then converting it back into a string seems wasteful (especially for doubles, that data type is not cheap to format into a string, and I'm not sure about parsing either).

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,254 @@
...
+            {
+                throw new ArgumentNullException(nameof(stream));
+            }
+
+            _complete = false;
+            _stream = stream;
+            _buffer = buffer;
+            _stream.Read(_buffer, 0, 3);
+            var offset = 0;
+            if (!Encoding.UTF8.GetPreamble().AsSpan().SequenceEqual(_buffer.AsSpan(0, 3)))
```
zivkan: `GetPreamble` allocates a new array every time: https://github.com/dotnet/runtime/blob/2b571c505a444522892a272ea7cf6ff485f59da5/src/libraries/System.Private.CoreLib/src/System/Text/UTF8Encoding.cs#L866

I think it would be better to copy System.Text.Json's pattern of using a static, readonly array, so it doesn't need to allocate every time.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,254 @@
...
+
+                    strings.Add(value);
+                }
+            }
+            return strings;
+        }
+
+        internal List<string> ReadStringArrayAsList()
+        {
+            return (List<string>)ReadStringArrayAsIList();
```
zivkan: Casting is cheap, but not free. And `List<T>` implements `IList<T>`, so I don't see any harm in having only 1 method.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,254 @@
...
+                    default:
+                        var invalidCastException = new InvalidCastException();
+                        throw new JsonException(invalidCastException.Message, invalidCastException);
+                }
+            }
+
+            return null;
+        }
+
+        internal bool ReadNextTokenAsBoolOrFalse()
```
zivkan, line 157: This method makes me sad. We control both the reading and writing of the assets file, so I don't understand why we'd have code that would allow a property that we expect to be a boolean to actually be a number or an array or object or whatever, and then silently ignore the schema error.

Since the bigger PR was split, I can no longer "see the forest for the trees", making it hard to find if this method is being used for a legitimate reason 😞

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,254 @@
...
+        {
+            if (_reader.BytesConsumed < _buffer.Length)
+            {
+                ReadOnlySpan<byte> leftover = _buffer.AsSpan((int)_reader.BytesConsumed);
+                if (leftover.Length == _buffer.Length)
+                {
+                    ArrayPool<byte>.Shared.Return(_buffer);
+                    _buffer = ArrayPool<byte>.Shared.Rent(_buffer.Length * 2);
+                }
+                leftover.CopyTo(_buffer);
```
zivkan: You're reading from a buffer (leftover) that you're already returned to the pool. If another thread has already rented the buffer and changed the contents, then this is going to result in a VERY difficult bug to track down.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,254 @@
...
+                _stream.Read(_buffer, leftover.Length, _buffer.Length - leftover.Length);
+            }
+            else
+            {
+                _stream.Read(_buffer, 0, _buffer.Length);
+            }
+            _reader = new Utf8JsonReader(_buffer, isFinalBlock: _stream.Length == _stream.Position, _reader.CurrentState);
```
zivkan: passing in `_buffer`, rather than caputing the result of `_stream.Read` and then passing in the actually used span of the buffer will result in parsing errors.  I'm honestly surprised that tests didn't catch this as a problem, because logically I would have expected this to cause errors at the end of the file nearly 100% of the time 😕   Although it requires the code using this class to validate that `.Read()` returns false once the object being deserialized is complete, which maybe it doesn't. However, if the current production code is parsing via NJ's `JObject`, I'd expect that to catch errors when there's unexpected characters after the end of the first object.

In case it's not clear, I expect that our current assets file deserializer will throw an exception if the file contents is `{}hello`, and I'd expect a test that catches that error, will also catch the bug here where the number of bytes read from the stream is not used to limit the span for which `Utf8JsonReader` is requested to read from `_buffer`.

zivkan: `Length` can throw exceptions in certain implementations of `Stream`, where the length isn't known: https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.length?view=net-8.0#exceptions

So, while this shouldn't affect NuGet with the way we currently use it, this is not a generalized `Utf8JsonStreamReader`. In fact, since HTTP requests return the compressed response size, not the uncompressed, I think this means that this class, as currently implemented, will not be reusable for parsing JSON coming from (compressed) HTTP responses. Again, the way NuGet is implemented, we usually save responses to disk. But I'm not sure what happens when someone uses NuGet.Protocol directory, and sets `SourceCacheContext.DirectDownload = true`.

Anyway, I'd play it safe by using the same logic that System.Text.Json uses. I'm not sure if that's by checking that `_stream.Read` returns zero, or if it's safe to assume to when `_stream.Read` returns less than the number of bytes requested if that always means it's the end of the stream.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: Okay that makes sense, I think. I've updated it to match that number, though when using ArrayPool we may in theory get something larger right?

jgonz120: ref structs can't implement interfaces but apparently you can just add a public dispose method to the struct and it's fine

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonReaderExtensions.cs
jgonz120: [GetString()](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.getstring?view=net-8.0#exceptions) throws an exception for all types other than string, property name, and null. I can reduce it down to only TryGetInt64, since that will work for int16 and int32. 

Another option is to get the string directly from the ReadOnlySpan, I found what they were doing [here](https://github.com/dotnet/runtime/blob/f09a97176b0069962036596fed327c560228809e/src/libraries/System.Text.Json/src/System/Text/Json/Reader/JsonReaderHelper.Unescaping.cs#L209C30-L209C45). They use pointers which means adding the unsafe flag. For now I implemented it by just running ToArray() on the span.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: mkay, gonna resolve this for now then.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: It's used a bunch of times in the current implementation of [JsonPackageSpecReader](https://github.com/NuGet/NuGet.Client/blob/1942c17866695ae1175ec3d8b42a6b41bcd6a9c9/src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.cs#L387C110-L387C110). So least it was being used before, whether or not it's legit idk, but we can look at that during the next PR

nkolev92: My instinct immediately was whether it's something that was used during the project.json times, but based on where it's being used, I think it's always used as a bool, but I spent like 30s looking :D

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: Ah, AsSpan keeps a copy of the original buffer passed in?

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: I updated one of LockFIleFormatTests to have "hello" outside of the final close brace and it still passed. I also updated one of the tests in JsonSpecReaderTests and it also passes with "Hello" at the end of the JSON string. 

I tried it in dotnetfiddle because I couldn't find any documentation saying this was expected behavior. https://dotnetfiddle.net/snHuQm 

Anyway, that's weird. I'm not sure what you mean by capturing the results of _stream.Read, like keeping track of the number of bytes read from the stream and passing in only those specific bytes as a span to the Utf8JsonReader?

jgonz120: Okay, Updated the way I'm populating the Utf8JsonReader so now only the data that was read form the stream is passed in, avoiding passing in a larger buffer with a bunch of empty items. 

If I pass in a malformed JSON to Utf8JsonReader, with text after the final closing bracket, it doesn't fail until the end if you do an additional read after the final close.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonReaderExtensions.cs
zivkan: Today I'm feeling like while it would technically be faster, maybe the benefit isn't worth it. Especially since there seem to be complications about getting it technically correct, since we'd need to handle when `HasValueSequence` is both true and false, possibly in addition to `ValueIsEscaped`.  So, let's go ahead with the number parsing plus `ToString`.  If perf traces show this `ReadTokenAsString` method taking too much time, then we can consider an optimization.

Something else you could consider, create sub-methods:

```csharp
switch (reader.TokenType)
{
    case JsonTokenType.Number:
        return ReadNumberAsString(ref reader);
    case JsonTokenType.String:
        return ReadStringAsString(ref reader);
   // other cases
}
```

This way the call stack will tell us which token types are actually being hit and taking up the CPU time.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
zivkan: yes, that's why it's fast. It's just a "view" over an existing block of memory. It's not so clearly spelled out, but the docs do contain quite a lot of information in the remarks section: https://learn.microsoft.com/en-us/dotnet/api/system.span-1?view=net-8.0

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
zivkan: > I tried it in dotnetfiddle because I couldn't find any documentation saying this was expected behavior

I find it very surprising that Newtonsoft.Json doesn't throw an exception. FWIW, if you try System.Text.Json's `JsonSerializer.Deserialize` method, it does validate that the end of the root object is the end of the file, and will throw if there's any trailing data. I really, really thought Newtonsoft.Json did the same!

> I'm not sure what you mean by capturing the results of _stream.Read, like keeping track of the number of bytes read from the stream and passing in only those specific bytes as a span to the Utf8JsonReader?

You've already updated your PR, but yes, that's exactly what I mean. If you read `Stream.Read`'s docs: https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.read?view=net-8.0#system-io-stream-read(system-byte()-system-int32-system-int32)

>> An implementation is free to return fewer bytes than requested even if the end of the stream has not been reached.

So, there's no guarantee that the array is going to be filled, even when there is more data available in the stream. But in the case of reaching the end of the file, it's extremely unlikely that the last byte of the file will coincidentally match the last byte of the array.

Therefore, it's important to ensure that `Utf8JsonReader` isn't allowed to read past the end of the bytes that the `Stream` actually read. But it looks like you've fixed that now 👍

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,274 @@
...
+        internal int BufferSize()
+        {
+            ThrowExceptionIfDisposed();
+
+            return _buffer.Length;
+        }
+
+        internal bool ValueTextEquals(ReadOnlySpan<byte> utf8Text) => _reader.ValueTextEquals(utf8Text);
+
+        internal bool ValueTextEquals(string text) => _reader.ValueTextEquals(text);
```
zivkan: This string overload requires `Utf8JsonReader` to convert the string into UTF8 bytes, and will do a heap allocation if the string length is over some threshold.

Therefore, in the interests of ensuring that NuGet deserialization is always as optimal as possible, should we consider removing this overload, and force all our own code to use the `ReadOnlySpan<byte>` overload?

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonReaderExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+        internal static string ReadTokenAsString(this ref Utf8JsonReader reader)
+        {
+            switch (reader.TokenType)
+            {
+                case JsonTokenType.True:
+                    return bool.TrueString;
+                case JsonTokenType.False:
+                    return bool.FalseString;
+                case JsonTokenType.Number:
+                    var span = reader.HasValueSequence ? reader.ValueSequence.ToArray() : reader.ValueSpan;
```
zivkan:
```suggestion
                    // Utf8JsonStreamReader always uses a single array, not a sequence, so this `.ToArray` should never be called, but it's here for correctness.
                    var span = reader.HasValueSequence ? reader.ValueSequence.ToArray() : reader.ValueSpan;
```

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,274 @@
...
+        private ReadOnlySpan<byte> _utf8Bom = new byte[] { 0xEF, 0xBB, 0xBF };
+        private Utf8JsonReader _reader;
+        // The buffer is used to read from the stream in chunks.
+        private byte[] _buffer;
+        private bool _disposed;
+
+        internal Utf8JsonStreamReader(Stream stream) : this(stream, ArrayPool<byte>.Shared.Rent(BufferSizeDefault))
+        {
+
+        }
```
zivkan: nitpick: unnecessary empty line

```suggestion
        {
        }
```

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,274 @@
...
+
+            bool wasRead;
+            while (!(wasRead = _reader.Read()) && !_reader.IsFinalBlock)
+            {
+                GetMoreBytesFromStream();
+            }
+            return wasRead;
+        }
+
+        internal bool TrySkip()
```
zivkan: Since this stream reader is automatically updating the buffer when a read can't be completed, do we need this `TrySkip` method?

I think we only need a `Skip()` method that is either successful, or throws. The only scenarios I can think of that a skip fails is when there's a syntax error in the Json, or when `Skip` is called at an invalid time. For example, does `Utf8JsonReader` allow you to skip when the token type is `PropertyName`?  Anyway, `TrySkip` suggests that the caller can gracefully handle the failure, which I think is what this stream reader class is abstracting away.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,274 @@
...
+        internal string ReadNextTokenAsString()
+        {
+            ThrowExceptionIfDisposed();
+
+            if (Read())
+            {
+                return _reader.ReadTokenAsString();
+            }
+
+            return null;
```
zivkan, line 116: I only had a quick look over the tests, but I didn't notice many error scenarios being tested.  With JsonSerializer, if I try to deserialize `"string without trailing double quote`, I would expect an exception saying the JSON file has invalid syntax. However, the way this code reads, I think it'll return null instead.

STJ's `Read()` method returns false when the end of the buffer is reached (and there's a special case when the current buffer is the final block). However, this `Utf8JsonStreamReader` is more like `Stream`, where read only returns false at the end of the stream. So, we need to be careful when copying STJ parsing code, because the underlying assumptions about the reader are different.  Perhaps it's worthwhile to add XML doc to all the methods, to reinforce the ideas about expected behaviour?

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,274 @@
...
+        }
+
+        internal string GetCurrentBufferAsString()
```
zivkan:
```suggestion
        /// <Summary>For tests only, do not call in production code</Summary>
        internal string GetCurrentBufferAsString()
```

similarly, `BufferLength` is a leaky abstraction for production code.

Even though these methods are internal, it somehow doesn't feel right for these methods to be available to all the other production code in the same assembly.  How horrible would it be to remove both methods, and then use reflection in the test code to get `_buffer`? 😱

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,274 @@
...
+            _reader.Read();
+        }
+
+        private Stream Stream { get; set; }
+
+        internal bool IsFinalBlock => _reader.IsFinalBlock;
+
+        internal JsonTokenType TokenType => _reader.TokenType;
+
+        internal int BufferSize()
```
zivkan: naming:  properties should be nouns, methods should be verbs.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,274 @@
...
+{
+    /// <summary>
+    /// This struct is used to read over a memeory stream in parts, in order to avoid reading the entire stream into memory.
+    /// It functions as a wrapper around <see cref="Utf8JsonStreamReader"/>, while maintaining a stream and a buffer to read from.
+    /// </summary>
+    internal ref struct Utf8JsonStreamReader
+    {
+        private static readonly char[] DelimitedStringDelimiters = new char[] { ' ', ',' };
+        private const int BufferSizeDefault = 16 * 1024;
+        private ReadOnlySpan<byte> _utf8Bom = new byte[] { 0xEF, 0xBB, 0xBF };
```
zivkan:
```suggestion
        private static ReadOnlySpan<byte> Utf8Bom = new byte[] { 0xEF, 0xBB, 0xBF };
```

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,274 @@
...
+        private void GetMoreBytesFromStream()
+        {
+            int leftoverBytes = 0;
+            int bytesReadFromStream;
+
+            if (_reader.BytesConsumed < _buffer.Length)
+            {
+                // If the number of bytes consumed by the reader is less than the buffer size then we have leftover bytes that need to be shifted
+                var oldBuffer = _buffer;
+                ReadOnlySpan<byte> leftover = oldBuffer.AsSpan((int)_reader.BytesConsumed);
```
zivkan, line 215: I think my concern is just a performance optimization, not a correctness bug, so technically can probably be ignored. But consider if `Stream.Read` returns less than the number of requested bytes. For example, the buffer is 2k, but it only reads 1k of data. This will happen effectively 100% of the time at the end of the file. We don't need to shift the unfilled parts of the buffer.

So, I think that either we should store the index of the first unused part of the buffer in a field. Alternatively, we could loop in this method until the buffer is either full, or we reach the end of the stream (so we can set `isFinalBlock`). If we do the loop, there would need to be a minor change in the logic for when we resize. I think we'd only need to resize when `_reader.BytesConsumed == 0`.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,274 @@
...
+
+                strings = strings ?? new List<string>();
+
+                strings.Add(value);
+            }
+
+            return (IReadOnlyList<string>)strings ?? Array.Empty<string>();
+        }
+
+        // This function is called when Read() returns false
```
zivkan:
```suggestion
        // This function is called when Read() returns false and we're not already in the final block
```

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,274 @@
...
+                if (returnOldBuffer)
+                {
+                    ArrayPool<byte>.Shared.Return(oldBuffer);
+                }
+            }
+            else
+            {
+                bytesReadFromStream = Stream.Read(_buffer, 0, _buffer.Length);
+            }
+            _reader = new Utf8JsonReader(_buffer.AsSpan(0, leftoverBytes + bytesReadFromStream), isFinalBlock: leftoverBytes + bytesReadFromStream < _buffer.Length, _reader.CurrentState);
```
zivkan: I find this much easier to read. I'm really not a fan of expressions in method parameters, but I rarely see anyone else in the team complain about it, so take it with a grain of salt.

```suggestion
            var readerSpan = _buffer.AsSpan(0, leftoverBytes + bytesReadFromStream);
            var isFinalBlock = leftoverBytes + bytesReadFromStream < _buffer.Length
            _reader = new Utf8JsonReader(readerSpan, isFinalBlock, _reader.CurrentState);
```

However, as mentioned previously, [`Stream.Read`'s remarks says](https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.read?view=net-8.0#system-io-stream-read(system-byte()-system-int32-system-int32))

> An implementation is free to return fewer bytes than requested even if the end of the stream has not been reached.

Therefore, it's an unsafe assumption for `isFinalBlock` to be set when the buffer is not filled to capacity.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,274 @@
...
+            _buffer = buffer;
+            Stream.Read(_buffer, 0, 3);
+            var offset = 0;
+            if (!_utf8Bom.SequenceEqual(_buffer.AsSpan(0, 3)))
+            {
+                offset = 3;
+            }
+            var blocksRead = Stream.Read(_buffer, offset, _buffer.Length - offset);
+
+            _reader = new Utf8JsonReader(_buffer.AsSpan(0, blocksRead + offset), isFinalBlock: blocksRead + offset < _buffer.Length, state: new JsonReaderState(new JsonReaderOptions
```
zivkan: As mentioned elsewhere, Stream.Read doesn't promise that the buffer will be filled, only that at least 1 byte will be filled. Therefore for certain `Stream` implementations, the `isFinalBlock` calculation will be incorrect.

Also, consider using variables for the method parameters. I'm not a fan of expressions in method parameters, but I don't remember seeing other people in the team give similar PR feedback, so maybe I'm the only one who cares.  If I can't hover my mouse in the debugger to see the exact value being passed to the method (so `someObject.PropertyName` is fine), I'd like a variable, so I can inspect the value when debugging.

zivkan (commented):
sorry about the number of comments I'm adding. I hope they're not too annoying/opinionated.  I do have some comments in this round about possible correctness issues that I'd like to see addressed before I'd feel comfortable signing off on this PR.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: Switched to no longer use stream.length.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonReaderExtensions.cs
jgonz120: No longer relevant because I switched back to getting the nunber using TryInt64

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: Switched to Skip for the name, but I'm still using TrySkip in the function, it's the best way to determine if we're at the end of the buffer.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: Read throws an exception if the JSON it's trying to parse is malformed. I added some tests to for that scenario.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: This gives an error, CS8345: Field or auto-implemented property cannot be of type unless it is an instance member of a ref struct. 

Switched it to be a byte array.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: I couldn't figure out how to get reflection to work on a ref struct. I did switch the constructor to take in an ArrayPool<byte> and updated the tests to pass in mock array pool. Lemme know what you think of that solution.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: I like the idea of reading the stream until the buffer is full. That way we can figure out "IsFinalBlock" from the start

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,273 @@
...
+            {
+                AllowTrailingCommas = true,
+                CommentHandling = JsonCommentHandling.Skip,
+            });
+
+            ReadStreamIntoBuffer(iniialJsonReaderState);
+            _reader.Read();
+        }
+
+        private Stream Stream { get; set; }
```
zivkan: nitpick (sorry if I didn't notice this earlier):  private auto-properties don't really have a lot of value, and isn't part of common .NET conventions. Maybe the JIT compiler optimizes it to the same perf as a field, but I think it's clearer to just use a field rather than a private auto-prop.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,273 @@
...
+            }
+
+            ReadStreamIntoBuffer(_reader.CurrentState);
+        }
+
+        /// <summary>
+        /// Loops through the stream and reads it into the buffer until the buffer is full or the stream is empty
+        /// </summary>
+        /// <returns>True if the stream is empty</returns>
+        private void ReadStreamIntoBuffer(JsonReaderState jsonReaderState)
```
zivkan: xmldoc is not consistent with the method signature.

zivkan (approved):
thank you for making all the improvements!

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: This one is super annoying, since Utf8JsonStreamReader is disposable all of the fields it has must also be disposed in the Dispose() method. I don't want to dispose of the stream though because it's passed in and that feels like it could be an issue.

davkean: > Utf8JsonStreamReader is disposable all of the fields 

How did you come to this conclusion? It should only dispose fields it owns. The auto property versus field doesn't change this anyway as they both end up with a field of type "Stream".

zivkan: I think there's a built-in roslyn analyzer when a field that implements `IDisposable` isn't disposed anywhere in the class (maybe it even validates that it must be disposed from the `Dispose()` method, I'm not sure).

However, we can suppress that analyzer in this class, if that's correct.  Plus add XMLDoc to the class, maybe the constructor too, to make it clear that disposing this struct won't dispose the stream.

Also consider that classes like `StreamReader` have overloads for `bool leaveOpen`. I'm not saying we have to copy that pattern, but if you think it makes sense, then it's an option. But maybe we never want to auto-dispose the underlying stream for this type 🤷

jgonz120: I got a this compile error [CA2213](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2213) and didn't dig into it until just now, looks like something to suppress. The description makes it seem like this really shouldn't be triggering in this instance anyway.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,273 @@
...
+        /// <summary>
+        /// Loops through the stream and reads it into the buffer until the buffer is full or the stream is empty, creates the Utf8JsonReader. 
+        /// </summary>
+        private void ReadStreamIntoBuffer(JsonReaderState jsonReaderState)
+        {
+            int bytesRead;
+            do
+            {
+                var spaceLeftInBuffer = _buffer.Length - _bufferUsed;
+                bytesRead = _stream.Read(_buffer, _bufferUsed, spaceLeftInBuffer);
```
kartheekp-ms, line 247: How about using the [Stream.ReadAsync Method](https://learn.microsoft.com/dotnet/api/system.io.stream.readasync?view=net-8.0) instead of the blocking Read method? It would be great if we could use the overload that accepts a `CancellationToken`.

[Docs](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/use-utf8jsonreader) state that for an asynchronous version of example, see [.NET samples JSON project](https://github.com/dotnet/samples/blob/18e31a5f1abd4f347bf96bfdc3e40e2cfb36e319/core/json/Program.cs).

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,273 @@
...
+            _buffer = _bufferPool.Rent(bufferSize);
+            _disposed = false;
+            _stream = stream;
+            _stream.Read(_buffer, 0, 3);
+            if (!Utf8Bom.AsSpan().SequenceEqual(_buffer.AsSpan(0, 3)))
+            {
+                _bufferUsed = 3;
+            }
+
+            var iniialJsonReaderState = new JsonReaderState(new JsonReaderOptions
```
kartheekp-ms, line 56: typo `initialJsonReaderState`

```suggestion
            var initialJsonReaderState = new JsonReaderState(new JsonReaderOptions
```

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,273 @@
...
+        {
+            int bytesRead;
+            do
+            {
+                var spaceLeftInBuffer = _buffer.Length - _bufferUsed;
+                bytesRead = _stream.Read(_buffer, _bufferUsed, spaceLeftInBuffer);
+                _bufferUsed += bytesRead;
+            }
+            while (bytesRead != 0 && _bufferUsed != _buffer.Length);
+            _reader = new Utf8JsonReader(_buffer.AsSpan(0, _bufferUsed), isFinalBlock: bytesRead == 0, jsonReaderState);
```
kartheekp-ms, line 251: https://learn.microsoft.com/dotnet/standard/serialization/system-text-json/use-utf8jsonreader doc suggests to use `_reader.CurrentState`. If this suggestion is valid for our scenario, then we can remove the `jsonReaderState` parameter from the method definition.

```suggestion
            _reader = new Utf8JsonReader(_buffer.AsSpan(0, _bufferUsed), isFinalBlock: bytesRead == 0, _reader.CurrentState);
```

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,273 @@
...
+        private void ReadStreamIntoBuffer(JsonReaderState jsonReaderState)
+        {
+            int bytesRead;
+            do
+            {
+                var spaceLeftInBuffer = _buffer.Length - _bufferUsed;
+                bytesRead = _stream.Read(_buffer, _bufferUsed, spaceLeftInBuffer);
+                _bufferUsed += bytesRead;
+            }
+            while (bytesRead != 0 && _bufferUsed != _buffer.Length);
```
kartheekp-ms, line 250: I'm curious about the need for the do-while loop inside the `ReadStreamIntoBuffer` method, especially considering that `ReadStreamIntoBuffer` is already invoked within a while loop in the `Read` method. If the buffer is smaller than the stream's content, or if the buffer has more capacity than what's currently available in the stream, the first call to `_stream.Read` in `ReadStreamIntoBuffer` should be enough to fill the buffer up to its capacity or the stream's current availability. Is the do-while loop necessary in this case, or could it lead to redundant read operations?

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
kartheekp-ms, line 56: I am wondering if we should create a static variable to hold the default `JsonReaderOptions` settings.

test/NuGet.Core.Tests/NuGet.ProjectModel.Test/Utf8JsonStreamReaderTests.cs
```diff
@@ -0,0 +1,833 @@
...
+            var json = Encoding.UTF8.GetBytes(JsonWithOverflowObject);
+            using (var stream = new MemoryStream(json))
+            using (var reader = new Utf8JsonStreamReader(stream, 1024))
+            {
+                Assert.False(reader.IsFinalBlock);
+            }
+        }
+
+        [Fact]
+        public void Read_WhenReadingMalfornedJsonString_Throws()
```
kartheekp-ms, line 85: typo `Malformed`

```suggestion
        public void Read_WhenReadingMalformedJsonString_Throws()
```

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,273 @@
...
+            _reader.Read();
+        }
+
+        internal bool IsFinalBlock => _reader.IsFinalBlock;
+
+        internal JsonTokenType TokenType => _reader.TokenType;
+
+        internal bool ValueTextEquals(ReadOnlySpan<byte> utf8Text) => _reader.ValueTextEquals(utf8Text);
+
+        internal bool TryGetInt32(out int value) => _reader.TryGetInt32(out value);
```
kartheekp-ms, line 72: Visual Studio indicates that these two methods (ValueTextEquals, TryGetInt32) have no references, which I assume means there is no unit test coverage for them. Could we add some tests, please?

test/NuGet.Core.Tests/NuGet.ProjectModel.Test/Utf8JsonStreamReaderTests.cs
```diff
@@ -0,0 +1,833 @@
...
+                    Assert.Equal(JsonTokenType.StartObject, reader.TokenType);
+                    reader.Read();
+                    Assert.Equal(JsonTokenType.PropertyName, reader.TokenType);
+                    reader.Read();
+                }
+            });
+        }
+
+        [Fact]
+        public void Read_WhenReadingMalfornedJson_Throws()
```
kartheekp-ms, line 104: typo `Malformed`

test/NuGet.Core.Tests/NuGet.ProjectModel.Test/Utf8JsonStreamReaderTests.cs
```diff
@@ -0,0 +1,833 @@
...
+
+        [Fact]
+        public void Read_WhenReadingWithoutOverflow_Read()
+        {
+            var json = Encoding.UTF8.GetBytes(JsonWithoutOverflow);
+
+            using (var stream = new MemoryStream(json))
+            using (var reader = new Utf8JsonStreamReader(stream))
+            {
+                Assert.Equal(JsonTokenType.StartObject, reader.TokenType);
```
kartheekp-ms, line 182: I think we should add `Assert.True(reader.IsFinalBlock);` to ensure that the default buffer capacity is sufficient to read the entire stream content in a single `Read()` operation when the JSON payload is small compared to the default buffer capacity, which is 16 * 1024.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,273 @@
...
+
+        internal bool Read()
+        {
+            ThrowExceptionIfDisposed();
+
+            bool wasRead;
+            while (!(wasRead = _reader.Read()) && !_reader.IsFinalBlock)
+            {
+                GetMoreBytesFromStream();
+            }
```
kartheekp-ms, line 88: In some cases, which I currently don't understand, the STJ implementation raises an exception when `IsFinalBlock` is set to true. Therefore, it would be better to check this value first before invoking the underlying `Read` method. Another advantage is that reading `IsFinalBlock` is an O(1) operation.

```suggestion
            while (!_reader.IsFinalBlock && !(wasRead = _reader.Read()))
            {
                GetMoreBytesFromStream();
            }
```



https://github.com/dotnet/runtime/blob/3a5bea5d60ea04b897ac968a358ca99a1189d368/src/libraries/System.Text.Json/src/System/Text/Json/Reader/Utf8JsonReader.cs#L269-L289

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,273 @@
...
+
+        internal IList<string> ReadStringArrayAsIList(IList<string> strings = null)
+        {
+            if (TokenType == JsonTokenType.StartArray)
+            {
+                while (Read() && TokenType != JsonTokenType.EndArray)
+                {
+                    string value = _reader.ReadTokenAsString();
+
+                    strings = strings ?? new List<string>();
```
kartheekp-ms, line 127:
```suggestion
                    strings ??= new List<string>();
```
https://learn.microsoft.com/dotnet/fundamentals/code-analysis/style-rules/ide0054-ide0074

test/NuGet.Core.Tests/NuGet.ProjectModel.Test/Utf8JsonStreamReaderTests.cs
```diff
@@ -0,0 +1,833 @@
...
+                    reader.GetString());
+            }
+        }
+
+        [Fact]
+        public void Read_WhenReadingWithLargeTokenReadPastFinal()
```
kartheekp-ms, line 236: I'm curious to understand what `ReadPastFinal` means in the test name. Does it imply that invoking `Read` after the input stream has been read completely, will always return false? Is that understanding, correct? If so, perhaps we could update the test name to better reflect this expectation.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,273 @@
...
+using System.Linq;
+using System.Text.Json;
+
+namespace NuGet.ProjectModel
+{
+    /// <summary>
+    /// This struct is used to read over a memeory stream in parts, in order to avoid reading the entire stream into memory.
+    /// It functions as a wrapper around <see cref="Utf8JsonStreamReader"/>, while maintaining a stream and a buffer to read from.
+    /// </summary>
+    internal ref struct Utf8JsonStreamReader
```
kartheekp-ms, line 17: `Utf8JsonStreamReader` struct appears to have dual responsibilities violating single responsibility principle.

**Stream Wrapper:** The struct acts as a wrapper around a stream, specifically for the purpose of reading it in chunks rather than loading the entire stream into memory. 

**Utf8JsonReader Wrapper:** The struct also wraps the functionality of Utf8JsonReader. This includes functionalities like reading various data types (strings, integers, booleans) from JSON, handling different JSON token types, and managing the state of the JSON reader.

How about splitting it into 2 structs, one handles stream buffer, resizing the buffer etc., and another one wraps UTF8JsonReader methods such as GetString(), GetBoolean() etc.,

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,273 @@
...
+
+            if (Read())
+            {
+                return _reader.ReadTokenAsString();
+            }
+
+            return null;
+        }
+
+        internal IList<string> ReadStringArrayAsIList(IList<string> strings = null)
```
kartheekp-ms, line 119: I am wondering if this method should return `IReadOnlyList<string>` just like `ReadDelimitedString` method.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,273 @@
...
+#pragma warning disable CA2213 // Disposable fields should be disposed
+        private Stream _stream;
+#pragma warning restore CA2213 // Disposable fields should be disposed
+        // The buffer is used to read from the stream in chunks.
+        private byte[] _buffer;
+        private bool _disposed;
+        private ArrayPool<byte> _bufferPool;
+        private int _bufferUsed = 0;
+
+        internal Utf8JsonStreamReader(Stream stream, int bufferSize = BufferSizeDefault, ArrayPool<byte> arrayPool = null)
```
kartheekp-ms, line 34: The constructor throws following exception when `Array.Empty<byte>()` is passed to the memory stream. I am just letting you know incase if you would like to add a test.

```cs
 using (var stream = new MemoryStream(Array.Empty<byte>()))
 using (var reader = new Utf8JsonStreamReader(stream))
{
}
```

 `System.Text.Json.JsonReaderException : '0x00' is an invalid start of a value. LineNumber: 0 | BytePositionInLine: 0.`

kartheekp-ms (commented):
I began reviewing the https://github.com/NuGet/NuGet.Client/pull/5541, but I realized that I lack sufficient context regarding the STJ API introduced in this PR. Hence, I started reviewing this PR. During this review, I identified several areas for improvement and have added comments to the PR. I apologize for the delay in providing this feedback. Since this PR has already been merged into a feature branch, please consider creating a follow-up PR to address these comments.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
zivkan: I don't have links handy, but my understanding is that on .NET Framework at least, files without passing the async flag to the operating system, is that there's a non-trivial perf impact to reading the files with async APIs. Similarly, files opened (from the OS point of view) with the async flag, but then using blocking APIs has a non-trivial perf impact. Also, .NET Framework's async scheduling just isn't nearly as good as .NET 6+'s, so even when async file IO is done consistently, when the delays are low (OS has file cached in memory?) then there's still going to be a measurable perf impact.

If this suggestion, using Stream.ReadAsync, is being seriously considered, I'd recommend reaching out to the VS Perf team and checking if my believe about async file IO on .NET Framework is wrong.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: The function is taking in the jsonReaderState because in the constructor the state doesn't already exist.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: I put it in a loop because of Andy's comment here: https://github.com/NuGet/NuGet.Client/pull/5530#discussion_r1421302997.

There isn't any guarantee that the stream is actually read to the limit requested. So the loop is necessary to ensure we are reading until the buffer is full.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: I was avoiding creating unit tests for these functions since they are just wrapping the functions in the Utf8JsonReader.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
davkean: Async isn't about making things faster; its about yielding thread pool threads while its blocking, which lets us make sure we're efficiently using the CPU by having only CPU work on them.

NuGet during restore spends a lot of time blocking thread pools threads (see Thread Pool watson, its \#13, \#14, \#19, \#22 in 17.8), so async where possible should be the approach.

zivkan: Thanks @davkean good to know. It's a shame that most of the thread pool watson buckets are caused by .NET & Windows APIs where there are no async alternatives. But we should make time to investigate async file reads (where APIs exist).

test/NuGet.Core.Tests/NuGet.ProjectModel.Test/Utf8JsonStreamReaderTests.cs
jgonz120: yup! renamed, let me know if you prefer the new name, it's in the currently open PR.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: This would be a large change if we want to do this I would like to do it in a new PR after the last one is complete.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: If we check for the final block first then we won't all the read at all. The exception you're seeing there means that the Utf8JsonReader was told it has all the data but the current property is [none](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsontokentype?view=net-8.0#fields). None meaning that there is no JSON data in the reader. If we want to check for the scenario we can, I would think it should be in the constructor though not here.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: The objects that use ReadStringArrayAsIList are either expecting a List or IList, so this wouldn't be possible.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: I added a test to demonstrate this behavior but I don't think it's worthwhile adding validation for this.

PR #5531, opened 2023-12-05
Reduce allocations in calls to CreateGraphNode()

Erarndt:
## Bug

## Description

We're consistently seeing heavy allocations during a restore and this PR addresses some of the major contributors. The creation of tracking tasks for `CreateGraphNode()` contributes a surprising amount to the overall allocations during this time. Switching to `ValueTask` helps alleviate this. Additionally, the temporary lists created to track these tasks ends up creating a significant amount of garbage. I've introduced `LightweightList` that is a struct that can hold up to 4 items before allocating a list. This way we can avoid the allocation in the vast number of scenarios. I've also removed the usage of `Task.WhenAll()` in this path due to its allocations caused by defensive copying and enumerator boxing.

I created some local functions to encapsulate the logic, but they aren't behaviorally significant and I can inline them if that's preferred.

## PR Checklist

- [x] PR has a linked issue.
- [ ] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
+++ b/src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
@@ -28,7 +28,7 @@ namespace NuGet.DependencyResolver
         public async Task<GraphNode<RemoteResolveResult>> WalkAsync(LibraryRange library, NuGetFramework framework, string runtimeIdentifier, RuntimeGraph runtimeGraph, bool recursive)
         {
             var transitiveCentralPackageVersions = new TransitiveCentralPackageVersions();
-            var rootNode = await CreateGraphNode(
+            var rootNode = await CreateGraphNodeAsync(
                 libraryRange: library,
                 framework: framework,
                 runtimeName: runtimeIdentifier,
@@ -64,7 +64,7 @@ namespace NuGet.DependencyResolver
             return rootNode;
         }
 
-        private async Task<GraphNode<RemoteResolveResult>> CreateGraphNode(
+        private async ValueTask<GraphNode<RemoteResolveResult>> CreateGraphNodeAsync(
             LibraryRange libraryRange,
             NuGetFramework framework,
             string runtimeName,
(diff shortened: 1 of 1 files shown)


src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
jeffkl, line 1: Do you have measurements to show how much memory is saved with this change?

jeffkl (approved):

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
```diff
@@ -75,41 +75,10 @@ public async Task<GraphNode<RemoteResolveResult>> WalkAsync(LibraryRange library
...
-                        }
-                    }
-                    else
-                    {
-                        // Otherwise it's a dependency of this node
-                        runtimeDependencies ??= new HashSet<LibraryDependency>(LibraryDependencyNameComparer.OrdinalIgnoreCaseNameComparer);
-                        runtimeDependencies.Add(libraryDependency);
-                    }
-                }
+                EvaluateRuntimeDependencies(ref libraryRange, runtimeName, runtimeGraph, ref runtimeDependencies);
```
nkolev92, line 81: This is just a refactoring correct?

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
```diff
@@ -128,115 +97,148 @@ public async Task<GraphNode<RemoteResolveResult>> WalkAsync(LibraryRange library
             };
 
             Debug.Assert(node.Item != null, "FindLibraryCached should return an unresolved item instead of null");
+            MergeRuntimeDependencies(runtimeDependencies, node);
```
nkolev92, line 100: Same with this correct?

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
```diff
@@ -552,4 +554,100 @@ public int GetHashCode(LibraryDependency obj)
             }
         }
     }
+
+    internal struct LightweightList<T>
+    {
+        private const int Fields = 4;
```
nkolev92, line 560: Why 4? 

I'm curious, because I've run queries to get the avg number of dependencies for package son nuget.org and the answer is 3.1.

```kql
NiPackageManifests
| join kind=inner NiPackageVersions on Identity
| where IsListed
| extend parsedGroups = parse_json(DependencyGroups)
| mv-expand parsedGroups
| extend DependencyCount = iff(isempty(parsedGroups['Packages']), 0, array_length(parsedGroups['Packages']))
| summarize avg(DependencyCount)
```

nkolev92 (approved):
Something I've tried myself that may add some benefits: https://github.com/NuGet/NuGet.Client/tree/dev-nkolev92-valueTaskResolverUtility. 

My changes need to handle the breaking APIs ofc, but it's the results that matter.

I haven't gotten traces to understand how big of a different it would make.

nkolev92:
I see this PR says:

> Fixes: https://github.com/NuGet/Home/issues/12748

Do you think these are all the fixes we'll have for that issue?

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
Erarndt: Correct. I can revert this. It was helpful while investigating.

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
Erarndt: Same. Not intended to be a significant change.

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
nkolev92: I don't care too much.
Sometimes we like keeping history because it makes blame more valuable, but we've done a lot of refactoring in this area. 

As far as I'm concerned, it's ok as is.

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
Erarndt: It was more or less a guess. I chose 4 because it's the default size for `List<T>` and because the existing code initialized the list to 1. Looks like 4 is in the ballpark of being a good heuristic. I'd be curious if we can do better though. For instance, if the 90th percentile is 4.8 (or something), making this 5 would be better.

nkolev92: 50 - 2, 
70 - 3,
80 - 4
90 - 7
95 - 9

This ofc doesn't mean usage though, so I don't want to read too much into it. 
I think we can keep it and then watch the data and adjust if needed.

nkolev92: > I chose 4 because it's the default size for List<T>

The initial capacity when something gets added for the sleuths :D

Erarndt:
> Something I've tried myself that may add some benefits: https://github.com/NuGet/NuGet.Client/tree/dev-nkolev92-valueTaskResolverUtility.
> 
> My changes need to handle the breaking APIs ofc, but it's the results that matter.
> 
> I haven't gotten traces to understand how big of a different it would make.

Looking briefly through the changes, I believe you'll run into issues using a `ValueTask` in the `ResolverUtility`. I could be wrong but the current usage looks like the `ValueTask` can be awaited multiple times concurrently. That usage is problematic:
https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/


_In general, the following operations should never be performed on a ValueTask / ValueTask\<TResult\>:_

_Awaiting a ValueTask / ValueTask\<TResult\> multiple times. The underlying object may have been recycled already and be in use by another operation. In contrast, a Task / Task\<TResult\> will never transition from a complete to incomplete state, so you can await it as many times as you need to, and will always get the same answer every time.
Awaiting a ValueTask / ValueTask\<TResult\> concurrently. The underlying object expects to work with only a single callback from a single consumer at a time, and attempting to await it concurrently could easily introduce race conditions and subtle program errors. It’s also just a more specific case of the above bad operation: “awaiting a ValueTask / ValueTask\<TResult\> multiple times.” In contrast, Task / Task\<TResult\> do support any number of concurrent awaits.
Using .GetAwaiter().GetResult() when the operation hasn’t yet completed. The IValueTaskSource / IValueTaskSource\<TResult\> implementation need not support blocking until the operation completes, and likely doesn’t, so such an operation is inherently a race condition and is unlikely to behave the way the caller intends. In contrast, Task / Task<TResult> do enable this, blocking the caller until the task completes.__

Erarndt:
> I see this PR says:
> 
> > Fixes: [NuGet/Home#12748](https://github.com/NuGet/Home/issues/12748)
> 
> Do you think these are all the fixes we'll have for that issue?

I think these are the most impactful changes I'll have by a large margin. Based on some of the traces, I expect allocations in this path to be reduced by ~25%. I do expect there to be more, but the impact will be smaller unless we consider larger breaking changes. My thought was to create separate PRs for each of these items to make reviews easier. Happy to change that approach though.

nkolev92:
I think smaller PRs is preferable. 
In GH, if you reference an issue with a fixes statement in the PR body, it'll automatically close the issue. 

if this is not the only change, we can probably create smaller issues and turn the main issue into an epic.

craigktreasure (approved):
:shipit:

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
Erarndt: I do, but the majority of the motivation for these changes comes from traces we've gathered via telemetry. I can see where the problem areas are, but I can't do an exact before and after comparison. However, I did use OrchardCore to get an idea of the impact for before and after the proposed changes. The following traces were captured while doing a Rebuild of the OrchardCore solution. The top one is before and the bottom one is after.

image


Some things to note:

1. A large number of `Task` objects are created by calls to `CreateGraphNode` that are all but eliminated in the trace by switching to `ValueTask`
2.  The tracking `List` allocations are significantly reduced by usage of `LightweightList`, but we still see significant allocations there indicating that we're frequently seeing packages with more than 4 dependencies. Manually inspecting the packages used in OrchardCore confirm this.
3. There is a measurable allocation impact caused by usage of `Task.WhenAll` and `Task.WhenAny`. I only addressed the `Task.WhenAll` case in this PR. You can see the `Task.WhenAny` impact if I expanded the `FindLibraryCachedAsync` node.
4. We have some extreme user cases where the allocations are significantly worse than in this trace. For example, just the `Task` objects allocated in this path account for gigabytes worth of allocations, and there are gigabytes worth of `GraphNode` and `GraphEdge` objects. This code path must be getting hit an incredible amount of time.

PR #5532, opened 2023-12-06
Handle conditional uninstallation in PackageReference based projects

### LINKED ISSUE #11914: [Bug]: PM UI cannot uninstall packages in multitargeting projects
### NuGet Product Used

### Product Version

VS 2022 Enterprise, Version 17.2.5

### Worked before?

No, this does not work in VS2019

### Impact

### Repro Steps & Context

Create a new console project.

Edit the project file to look like the following:

```xml

	<PropertyGroup>
		<OutputType>Exe</OutputType>
		<TargetFrameworks>net5.0;net472</TargetFrameworks>
	</PropertyGroup>

	<ItemGroup Condition="$(TargetFramework) == 'net6.0'">
		<PackageReference Include="Text2Xml.Lib" Version="1.1.4" />
		<PackageReference Include="Newtonsoft.json" Version="13.0.1" />
		<PackageReference Include="Polly.Extensions.Http" Version="3.0.0" />
	</ItemGroup>

	<ItemGroup Condition="$(TargetFramework) == 'net472'">
		<PackageReference Include="Newtonsoft.json" Version="12.0.1" />
		<PackageReference Include="Polly.Extensions.Http" Version="2.0.1" />
	</ItemGroup>

</Project>
```

Right click on the project in Solution Explorer and select 'Manage NuGet Packages'

Uninstall package in 'Polly.Extensions.Http'.

### Expected
 
Polly.extensions.http package is uninstalled

### Actual

Package is not uninstalled. See screenshot below.

image

### Verbose Logs

```shell
Restoring packages for C:\Users\feaguila.REDMOND\source\repos\ConsoleAppMultitargeting2022\ConsoleAppMultitargeting2022\ConsoleAppMultitargeting2022.csproj...
System.ArgumentException: Package with id "Polly.Extensions.Http" does not exist.
   at Microsoft.VisualStudio.ProjectSystem.References.PackageReferencesService.<>c__DisplayClass3_0.<<RemoveAsync>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.VisualStudio.Threading.JoinableTask.<JoinAsync>d__76.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.VisualStudio.ProjectSystem.ProjectLockService.<ExecuteWithinLockAsync>d__129.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Microsoft.VisualStudio.ProjectSystem.ProjectLockService.<ExecuteWithinLockAsync>d__129.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NuGet.PackageManagement.VisualStudio.CpsPackageReferenceProject.<UninstallPackageAsync>d__17.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at NuGet.PackageManagement.NuGetPackageManager.<ExecuteBuildIntegratedProjectActionsAsync>d__87.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NuGet.PackageManagement.NuGetPackageManager.<ExecuteNuGetProjectActionsAsync>d__84.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NuGet.PackageManagement.NuGetPackageManager.<ExecuteNuGetProjectActionsAsync>d__83.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at NuGet.PackageManagement.NuGetPackageManager.<ExecuteNuGetProjectActionsAsync>d__82.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NuGet.PackageManagement.VisualStudio.NuGetProjectManagerService.<>c__DisplayClass18_0.<<ExecuteActionsAsync>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NuGet.PackageManagement.VisualStudio.NuGetProjectManagerService.<CatchAndRethrowExceptionAsync>d__29.MoveNext()
Time Elapsed: 00:00:00.1476767
========== Finished ==========
```

### LINKED ISSUE #13034: Conditional updating when *all*  packages are conditioned is broken
Take an example: 

```csproj

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFrameworks>net472;net8.0</TargetFrameworks>  
    <Nullable>enable</Nullable>
  </PropertyGroup>
    <ItemGroup Condition="'$(TargetFramework)' == 'net472'">
	    <PackageReference Include="Newtonsoft.Json">
		    <Version>13.0.1</Version>
	    </PackageReference>
    </ItemGroup>
    <ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
	    <PackageReference Include="Newtonsoft.Json">
		    <Version>13.0.2</Version>
	    </PackageReference>
    </ItemGroup>
</Project>
```

Updating in VS will only update the first one, since NuGet doesn't really know that the packages itself have been set without a condition. 

Note that, while scenario like the above are detectable, scenarios where the same version is used on the reference are likely not possible today.  FWIW, the 2nd iteration (same version, 2 items) is a lot less likely, so just fixing the consolidation change is probably more than enough.

nkolev92:
## Bug

Fixes: https://github.com/NuGet/Home/issues/11914
Fixes: https://github.com/NuGet/Home/issues/13034

## Description

- Handle conditional package uninstallation similarly to conditional package installation, by having a successful and unsuccessful list.
- At uninstallation time, use the conditional reader if the change does not apply to all frameworks.
- Handle conditional uninstallation/updating where the package is conditionally referenced in more than 1 framework.
The heuristic is that if a package is referenced with different versions but part of every framework, then the references have to be conditional. 
Note that there are some scenarios that are simply not fixable in VS with the current APIs, and that's when a package is referenced with the same version but conditionally in every framework. 

- Considerations: Change the installation/uninstallation context to use the alias only. I decided against this since Andy will be making that change, and while I can make guesses, it's probably best if those changes are self contained within the work in https://github.com/nuget/home/issues/5154. Andy can change to whatever BuildIntegratedInstallationContext needs to be (maybe create a new type for example).

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


zivkan (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/CpsPackageReferenceProject.cs
```diff
@@ -358,10 +358,48 @@ private async Task SetPackagePropertyValueAsync(IProjectProperties metadata, str
...
         {
-            var configuredProject = await _unconfiguredProject.GetSuggestedConfiguredProjectAsync();
-            await configuredProject?.Services.PackageReferences.RemoveAsync(packageIdentity.Id);
+            throw new InvalidOperationException("For CPSPackageReferenceProject, the uninstall method that can handle conditional targeting must be called instead.");
+        }
+
+        public override async Task<bool> UninstallPackageAsync(string packageId, BuildIntegratedInstallationContext installationContext, CancellationToken token)
+        {
+            token.ThrowIfCancellationRequested();
+            if (packageId == null) throw new ArgumentNullException(nameof(packageId));
```
jgonz120: empty packageId are okay?

nkolev92: No, I just didn't check for it. 

I can add that check too.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
```diff
@@ -3214,10 +3215,32 @@ internal static BuildIntegratedInstallationContext CreateInstallationContextForP
...
+
+            foreach (var framework in packageSpec.TargetFrameworks)
+            {
+                foreach (var dependency in framework.Dependencies)
+                {
+                    if (dependency.Name == packageIdentityId)
+                    {
+                        versions ??= new();
+                        versions.Add(dependency.LibraryRange.VersionRange);
+                        break;
```
kartheekp-ms, line 3239: I guess we could return from here if `versions.Count > 1`, in that way we can exit the loop when our condition is met. Finally, we could return false because the condition was never met.

nkolev92: We certainly could, but this isn't a hot path, and projects with 3 frameworks aren't that common, so I didn't bother optimizing.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/CpsPackageReferenceProject.cs
```diff
@@ -358,10 +358,48 @@ private async Task SetPackagePropertyValueAsync(IProjectProperties metadata, str
...
+            throw new InvalidOperationException("For CPSPackageReferenceProject, the uninstall method that can handle conditional targeting must be called instead.");
+        }
+
+        public override async Task<bool> UninstallPackageAsync(string packageId, BuildIntegratedInstallationContext installationContext, CancellationToken token)
+        {
+            token.ThrowIfCancellationRequested();
+            if (string.IsNullOrEmpty(packageId)) throw new ArgumentException(string.Format(Strings.Argument_Cannot_Be_Null_Or_Empty, nameof(packageId)));
+            if (installationContext == null) throw new ArgumentNullException(nameof(installationContext));
+
+            if (installationContext.SuccessfulFrameworks.Any() && installationContext.UnsuccessfulFrameworks.Any())
```
kartheekp-ms, line 372: I am curious why are we checking `installationContext.UnsuccessfulFrameworks.Any()` because the collection is not accessed inside the if block.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/CpsPackageReferenceProject.cs
```diff
@@ -358,10 +358,48 @@ private async Task SetPackagePropertyValueAsync(IProjectProperties metadata, str
...
+                    throw new InvalidOperationException(string.Format(
+                        CultureInfo.CurrentCulture,
+                        Strings.UnableToGetCPSPackageInstallationService,
+                        ProjectFullPath));
+                }
+
+                foreach (var framework in installationContext.SuccessfulFrameworks)
+                {
+                    string originalFramework;
+                    if (!installationContext.OriginalFrameworks.TryGetValue(framework, out originalFramework))
```
kartheekp-ms, line 390: My understanding is that the `installationContext.SuccessfulFrameworks` is a subset of `installationContext.OriginalFrameworks`. If this is accurate then I guess the dictionary will always contain the `framework` and we dont need to invoke `framework.GetShortFolderName()` method. Please let me know if I am missing something.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
```diff
@@ -223,11 +223,15 @@ private async Task AddProjectItemAsync(string filePath)
             container.AddFromFileCopy(fullPath);
         }
 
-        public override async Task<bool> UninstallPackageAsync(
+        public override Task<bool> UninstallPackageAsync(
             PackageIdentity packageIdentity, INuGetProjectContext _, CancellationToken token)
         {
-            await ProjectServices.References.RemovePackageReferenceAsync(packageIdentity.Id);
+            return UninstallPackageAsync(packageIdentity.Id);
```
kartheekp-ms, line 229: I'm not sure how this works from PM UI, but if the user's intent is to uninstall the package for a specific target framework, then it seems this method would uninstall the package from all target frameworks that depend on this package. 

EDIT: IIUC, PM UI doesn't support installing/uninstalling packages for a particular target framework. In this case, this comment is not applicable.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
```diff
@@ -3214,10 +3215,32 @@ internal static BuildIntegratedInstallationContext CreateInstallationContextForP
...
-                originalFrameworks);
+                originalFrameworks,
+                areAllPackagesConditional);
+        }
+
+        private static bool DoesPackageAppearWithDifferentVersions(string packageIdentityId, PackageSpec packageSpec)
+        {
+            HashSet<VersionRange> versions = default;
+
+            foreach (var framework in packageSpec.TargetFrameworks)
```
kartheekp-ms, line 3231: I am wondering if we should check for only `successfulFrameworksWithPackage` instead of all the target frameworks. The reason is we invoke conditional service only for successful frameworks [here](https://github.com/NuGet/NuGet.Client/pull/5532/files#diff-5fae96e8312674fe53149dbdb5a5f036aea82698a5aeff49617e8e7ae66d2685R387).

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
nkolev92: This is legacy anyways, which does not support mutlitargeting.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/CpsPackageReferenceProject.cs
nkolev92: We only want to do conditional uninstallation if there are both successful and unsuccessful frameworks. Otherwise, it's likely not conditional, it's everywhere.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/CpsPackageReferenceProject.cs
nkolev92: I was just being defensive here, similar to the rest of the conditional handling which already does this. 
I think @zivkan is gonna have to change these APIs when working on aliases, and can make them sctricter. For example, he'd probably only need to pass a string and not a framework as a condition.

src/NuGet.Core/NuGet.PackageManagement/NuGetPackageManager.cs
nkolev92: This is a before thing, if you look at the calls, it's only using the original package spec, not the resulting spec.

In general though, my focus is on having the correct heuristic even if we're not short circuiting as early. 
These code paths are not very hot, and as such there's no need for a heavy performance focus at the cost of readability.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/CpsPackageReferenceProject.cs
nkolev92: Digging deeper, probably being defensive here is unnecessary. 

Right now it's the only pending change, so if you'd like me to change it, let me know and I'd be happy to do it.

nkolev92:
@NuGet/nuget-client

It'd be great if I can get timely reviews for this, as we need this for P3, and with vacations and holidays there's not a lot of time left, even fi there's a lot of chronological time.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/CpsPackageReferenceProject.cs
kartheekp-ms: I am okay with you creating a follow-up PR to address this comment.

kartheekp-ms (approved):

PR #5533, opened 2023-12-07
Output a more debuggable message if a single value isn't specified

### LINKED ISSUE #8749: NuGet should handle duplicate nomination data better.
Currently, the package spec would get generated and then we'd throw during the restore operation itself, with a cryptic error: 
```
Error occurred while restoring NuGet packages: Sequence contains more than one matching element
```

Blocked on https://github.com/NuGet/Home/issues/7717

jasonmalinowski:
A number of places call GetSingleNonEvaluatedPropertyOrNull to get the single value that spans all TFMs -- if that's not distinct though the failure that would bubble up to the user wouldn't point to the underlying issue. This makes the failure easier.

## Bug

Fixes: https://github.com/NuGet/Home/issues/8749

Regression? No.

## Description
This adds a better exception message than simply calling SingleOrDefault(). The test changes mostly involve extracting out a helper function for other tests doing the same thing.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.SolutionRestoreManager/VSNominationUtilities.cs
+++ b/src/NuGet.Clients/NuGet.SolutionRestoreManager/VSNominationUtilities.cs
@@ -369,7 +369,23 @@ namespace NuGet.SolutionRestoreManager
             string propertyName,
             Func<string, TValue> valueFactory)
         {
-            return GetNonEvaluatedPropertyOrNull(values, propertyName, valueFactory).SingleOrDefault();
+            var distinctValues = GetNonEvaluatedPropertyOrNull(values, propertyName, valueFactory).ToList();
+
+            if (distinctValues.Count == 0)
+            {
+                return default(TValue);
+            }
+            else if (distinctValues.Count == 1)
+            {
+                return distinctValues[0];
+            }
+            else
+            {
+                distinctValues.Sort();
+                var distinctValueStrings = string.Join(", ", distinctValues);
+                var message = string.Format(CultureInfo.CurrentCulture, Resources.PropertyDoesNotHaveSingleValue, propertyName, distinctValueStrings);
+                throw new InvalidOperationException(message);
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/NuGet.Clients/NuGet.SolutionRestoreManager/VSNominationUtilities.cs
```diff
@@ -369,7 +369,21 @@ private static NuGetFramework GetToolFramework(IEnumerable targetFrameworks)
...
+                return default(TValue);
+            }
+            else if (distinctValues.Count == 1)
+            {
+                return distinctValues[0];
+            }
+            else
+            {
+                var distinctValueStrings = string.Join(", ", distinctValues);
+                throw new InvalidOperationException($"The property {propertyName} was expected to have a single value across all TFMs, but instead had the following values: {distinctValueStrings}");
```
nkolev92: This should probably be localized.

nkolev92 (commented):
The change looks great!

A few things: 

- Can you please create an issue for this?
- Can youn please add some tests: https://github.com/NuGet/NuGet.Client/blob/dev/test/NuGet.Clients.Tests/NuGet.SolutionRestoreManager.Test/VsSolutionRestoreServiceTests.cs has integration tests, but this is something that should be easily unit testable.

jasonmalinowski:
@nkolev92: looking to see if there was an existing bug, I see https://github.com/NuGet/Home/issues/8749 looks somewhat related -- want to call this as fixing that or have me file a new bug?

nkolev92:
> @nkolev92: looking to see if there was an existing bug, I see [NuGet/Home#8749](https://github.com/NuGet/Home/issues/8749) looks somewhat related -- want to call this as fixing that or have me file a new bug?

Nice catch, that's perfect.

jasonmalinowski:
@nkolev92 Exception message is now localized, and tests added. I broke out a test refactoring into a prior commit so it's a touch easier to review.

src/NuGet.Clients/NuGet.SolutionRestoreManager/VSNominationUtilities.cs
jasonmalinowski: Fixed.

nkolev92 (approved):

zivkan (approved):

PR #5534, opened 2023-12-07
Use Macros when writing the assets file to avoid updates due to user profile changes

nkolev92:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2178

## Description

Many properties within the PackageSpec, and by extension, the lock file, are paths.
These paths can often contain the user profile, the home path etc. Paths are user specific.

The assets file itself isn't designed to be shared across machines cleanly, but there are things that one can do to allow that to be a thing.
For example, specifying the same global packages folder, putting the code in the same location etc. 
If when writing the assets file, we convert paths like UserProfile to a macro, we can allow the assets file to be shared across machines. 

* Write `$(User)` where the user settings directory is a part of the path. 
* Extract `$(User)` macro at assets file read, effectively only change what's written on disk, but nothing changes for the rest of  code.
* At this point, this behavior is hidden behind an environment variable. We can decide whether to enable it for everyone at a later point.
* As of now, only the paths that have been found to cause issues for the devbox are being changed. This should minimize the overhead of the change.

* To enable the feature, the `NUGET_ENABLE_EXPERIMENTAL_MACROS` environment variable needs to be set to `true`.  

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.ProjectModel/MacroStringsUtility.cs
+++ b/src/NuGet.Core/NuGet.ProjectModel/MacroStringsUtility.cs
@@ -1,0 +1,72 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace NuGet.ProjectModel
+{
+    // Add JsonPackageReader and PackageSpecWriter tests.
+    internal static class MacroStringsUtility
+    {
+        internal const string NUGET_ENABLE_EXPERIMENTAL_MACROS = nameof(NUGET_ENABLE_EXPERIMENTAL_MACROS);
+        internal const string UserMacro = "$(User)";
+
+        /// <summary>
+        /// Applies macros in place to every string in the list.
+        /// Macros are applied only at the beginning of a string.
+        /// </summary>
+        /// <param name="list">The list of elements to apply the macro on.</param>
+        /// <param name="macroValue">The macro value that'll need to match from the strings.</param>
+        /// <param name="macroName">The macro that'll replace the value within the string.</param>
+        /// <param name="stringComparison">The comparer to use. Normally these strings are paths, so the comparer should be OS aware.</param>
+        internal static void ApplyMacros(IList<string> list, string macroValue, string macroName, StringComparison stringComparison)
+        {
+            if (list != null)
+            {
+                for (int i = 0; i < list.Count; i++)
+                {
+                    var result = ApplyMacro(list[i], macroValue, macroName, stringComparison);
+                    list[i] = result;
+                }
+            }
+        }
+
+        internal static void ExtractMacros(List<string> list, string macroValue, string macroName)
+        {
+            if (list != null)
+            {
+                for (int i = 0; i < list.Count; i++)
+                {
+                    var result = ExtractMacro(list[i], macroValue, macroName);
+                    list[i] = result;
+                }
+            }
+        }
+
+        /// <summary>
+        /// Applies macro to the string in the list.
+        /// </summary>
+        /// <param name="originalString">The string to apply the macro on.</param>
+        /// <param name="macroValue">The macro value that'll need to match from the string.</param>
+        /// <param name="macroName">The macro that'll replace the value within the string.</param>
+        /// <param name="stringComparison">The comparer to use. Normally these strings are paths, so the comparer should be OS aware.</param>
+        internal static string ApplyMacro(string originalString, string macroValue, string macroName, StringComparison stringComparison)
+        {
+            if (!string.IsNullOrEmpty(originalString) && !string.IsNullOrEmpty(macroValue) && originalString.StartsWith(macroValue, stringComparison))
+            {
+                return macroName + originalString.Substring(macroValue.Length);
+            }
+            return originalString;
+        }
+
+        internal static string ExtractMacro(string originalString, string macroValue, string macroName)
+        {
+            if (!string.IsNullOrEmpty(originalString) && !string.IsNullOrEmpty(macroName) && originalString.StartsWith(macroName, StringComparison.Ordinal))
+            {
+                return macroValue + originalString.Substring(macroName.Length);
+            }
+            return originalString;
+        }
+    }
+}
--- a/test/NuGet.Core.Tests/NuGet.ProjectModel.Test/JsonPackageSpecReaderTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.ProjectModel.Test/JsonPackageSpecReaderTests.cs
@@ -15,6 +15,7 @@ using NuGet.LibraryModel;
 using NuGet.Packaging.Core;
 using NuGet.RuntimeModel;
 using NuGet.Versioning;
+using Test.Utility;
 using Xunit;
 
 namespace NuGet.ProjectModel.Test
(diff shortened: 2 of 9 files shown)


src/NuGet.Core/NuGet.ProjectModel/MacroStringsUtility.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Collections.Generic;
+
+namespace NuGet.ProjectModel
+{
+    // TODO NK - Use OS Aware comparer.
+    // Hide behind environment variable, and require it to always be set for the time being.
+    // Add JsonPackageReader and PackageSpecWriter tests.
+    internal static class MacroStringsUtility
+    {
+        internal static string UserProfileMacro = "$(UserProfile)";
```
zivkan: `UserProfile` is an environment variable on Windows, but not Linux (no idea about Mac).  `HOME` is an environment variable on Linux, but not Windows, but it fewer characters. So, I like Home more than UserProfile.

nkolev92: Yeah, I think in order to cover all scenarios, I'll need to use both Home and UserSettingsDirectory. 

On linux/mac those are the same directory, but on windows they aren't, which is the primary scenario.

nkolev92: I think UserSettings is more helpful to being with though, so I think I'll with that.

src/NuGet.Core/NuGet.ProjectModel/MacroStringsUtility.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// </summary>
+        /// <param name="originalString">The string to apply the macro on.</param>
+        /// <param name="macroValue">The macro value that'll need to match from the string.</param>
+        /// <param name="macroName">The macro that'll replace the value within the string.</param>
+        /// <param name="stringComparison">The comparer to use. Normally these strings are paths, so the comparer should be OS aware.</param>
+        internal static string ApplyMacro(string originalString, string macroValue, string macroName, StringComparison stringComparison)
+        {
+            if (!string.IsNullOrEmpty(originalString) && !string.IsNullOrEmpty(macroValue) && originalString.StartsWith(macroValue, stringComparison))
+            {
+                return macroName + originalString.Substring(macroValue.Length);
```
zivkan, line 58: Has the perf of this been measured?  For example, on OrchardCore?  I hope it's immeasurable, but I guess it's better to validate than be surprised.

I'm not how how frequently paths are used in the assets file, but both here and in `ExtractMacro` there are 2 additional string allocations per call (that use the macro). NuGet will need to do these 2 additional allocations per path on write, and then again on read, so effectively 4 additional allocations per affected path on every CI restore + build operation.

My concern is that this PR's feature only benefits the first restore on a devbox. And in this case, the global packages folder is already filled, so it shouldn't be a long restore. But the perf impact is on every single restore, on every single machine, including on CI.

Given we're frequently asked to improve restore performance, I'm wondering if this feature is conflicting with our perf requirements.

Perhaps the question is, is the environment variable to enable the feature intended to be a temporary thing, while we test that it doesn't break anyone, or do we intend on keeping this environment variable forever (and keep the feature off by default forever), so that only devboxes can opt-in to the feature?  In this case, the PR checklist might be wrong, because no documentation PR is filed to update the environment variables list doc page.

test/NuGet.Core.Tests/NuGet.ProjectModel.Test/JsonPackageSpecReaderTests.cs
```diff
@@ -3638,6 +3639,111 @@ public void GetPackageSpec_WithRestoreAuditProperties_ReturnsRestoreAuditPropert
...
+        {
+            // Arrange
+            var json = @"{  
+                            ""restore"": {
+    ""projectUniqueName"": ""C:\\users\\me\\source\\code\\project.csproj"",
+    ""projectName"": ""project"",
+    ""projectPath"": ""C:\\users\\me\\source\\code\\project.csproj"",
+    ""projectJsonPath"": ""C:\\users\\me\\source\\code\\project.json"",
+    ""packagesPath"": ""$(User).nuget\\packages"",
+    ""outputPath"": ""C:\\users\\me\\source\\code\\obj"",
```
zivkan, line 3707: here, only `packagePath` uses `$(User)`.  Should all of the others as well?

they're not getting validated in the asserts, but given the name of the test method, it's not clear why these other properties are not using the macro, when it feels like they should be.

test/NuGet.Core.Tests/NuGet.ProjectModel.Test/PackageSpecWriterTests.cs
```diff
@@ -725,6 +727,59 @@ public void Write_RestoreAuditProperties_RoundTrips()
...
+        {
+            // Arrange
+            var json = @"{  
+                            ""restore"": {
+    ""projectUniqueName"": ""C:\\users\\me\\source\\code\\project.csproj"",
+    ""projectName"": ""project"",
+    ""projectPath"": ""C:\\users\\me\\source\\code\\project.csproj"",
+    ""projectJsonPath"": ""C:\\users\\me\\source\\code\\project.json"",
+    ""packagesPath"": ""$(User).nuget\\packages"",
+    ""outputPath"": ""C:\\users\\me\\source\\code\\obj"",
```
zivkan, line 741: same question as another comment I wrote. All these `c:\users\me` paths seem like they should be replaced with `$(User)`, like it is for `packagePaths`

nkolev92: In order to completely make the assets file able to float from machine to machine, we'd need to apply macro to *every* path, and that'd include project refs etc. So doing these without that is not going far enough. 

The reason why I did the nuget.config and fallback folders is because those are defaults, things you can't control. For example, the nuget.config will get created whether you want it or not. 

I chose to go the more minimal direction for the time being as I didn't want to pay the cost for now.

test/NuGet.Core.Tests/NuGet.ProjectModel.Test/JsonPackageSpecReaderTests.cs
nkolev92: Same as the other one.

Originally, I had all of these as well, but then I realized I needed to change projects refs and all of that as well, and that simply was more than the minimal change dev box needs given that the project is always in the same location. 

msbuild/project system caches depend on that too.

src/NuGet.Core/NuGet.ProjectModel/MacroStringsUtility.cs
nkolev92: I tried measuring, but no difference.

> I'm not how how frequently paths are used in the assets file, but both here and in ExtractMacro there are 2 additional string allocations per call (that use the macro). NuGet will need to do these 2 additional allocations per path on write, and then again on read, so effectively 4 additional allocations per affected path on every CI restore + build operation.

That's why I minimized the number of changes because it'll almost always be like 2-3 paths per project, which is not a crazy amount. 

The thing with devbox is that the priority is fast "get to code", so the cost of 6 allocations is better than the cost of project-system redoing their change because nuget touched the assets file with a change that doesn't really change anything for the project system (basically saying a new config was added).

test/NuGet.Core.Tests/NuGet.ProjectModel.Test/PackageSpecWriterTests.cs
zivkan: Makes sense. And yes, thinking about it now, these are the paths to the project, which won't change per-user, so it's logical that it doesn't need to use the macros 👍

src/NuGet.Core/NuGet.ProjectModel/MacroStringsUtility.cs
zivkan: > The thing with devbox is that the priority is fast "get to code"

I understand, but the point I was trying to make is that some optimizations to make "first open solution and starting coding" can be harmful to "this happens on every CI build, where there is no user interactively waiting for anything".

However, if the change doesn't impact perf in a measurable way, and especially if we keep a feature flag which we except never to be used on CI, then it could be fine.

zivkan (approved):

PR #3040, opened 2019-09-01
Use StringBuilder instead of multiple string concatenations

zivkan:
## Bug

Fixes: NuGet/Home#8540
Regression: No  
* Last working version:   
* How are we preventing it in future:   

## Fix

Details: Mostly by using StringBuilder, but in a few places other techniques were used to avoid string concatenation.

## Testing/Validation

Tests Added: No  
Reason for not adding tests:  No features added or changed. Relying on existing tests instead.
Validation:

PR #5536, opened 2023-12-08
Avoid need to sort sequences to compare equality.

### LINKED ISSUE #13247: Sorting sequences to compare equality should be improved
### NuGet Product Used

MSBuild.exe, NuGet.exe

### Product Version

All

### Worked before?

_No response_

### Impact

None

### Repro Steps & Context

Several places in our code are comparing sequences which requires sorting them.  This can add a measurable amount of overhead and we should investigate better ways to do this.

### Verbose Logs

_No response_

Erarndt:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13247

## Description

Improves performance and reduces allocations when checking lists for equality.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


build/Shared/EqualityUtility.cs
```diff
@@ -14,48 +14,72 @@ namespace NuGet.Shared
...
             {
                 return identityEquals;
             }
 
-            return self
-                .OrderBy(keySelector, orderComparer)
-                .SequenceEqual(other.OrderBy(keySelector, orderComparer), sequenceComparer);
+            int selfCount = 0;
+            equalityComparer ??= EqualityComparer<TKey>.Default;
+            var sourceItems = new Dictionary<TKey, int>(equalityComparer);
```
zivkan: The PR description is lacking details, but from what I can tell, instead of sorting two collections (via LINQ's `OrderBy`), this is adding keys from one collection to a dictionary, counting when a key is encountered multiple times, then kind of reversing the process for the second collection, to check if they're equal.

I agree that in theory, this sounds like it should be faster and fewer allocations. Sorting is generally an O(n log n) operation, whereas hash sets are in theory O(n). However, Big O notations is meaningless for small collections, it only tends us general trends as inputs tend towards very large numbers.

Additionally, this approach uses a single `Dictionary`, whereas I suspect that `OrderBy` uses `List`, and since `OrderBy` is called twice, that means there are fewer allocations. Furthermore, I suspect that Dictionary resizes increase capacity faster, resulting in fewer resizes.

However, adding a new item to a `List` is trivial, just inserting at `_backingArray[Count]`. When the backing array runs out of capacity, a new array is allocated and the existing list can be `Array.Copy`. Meanwhile, to insert into a Dictionary, the key's hash code needs to be obtained (possibly non-trivial calculation), the result is divided by the backing array length (to get the modulous). Dividing (and modulous) has historically been some of the slowest CPU instructions. And then hash collisions need to be handled. If the dictionary's capacity is exceeded, then effectively the entire process needs to repeat for a new backing array, since the index of each item is going to change. These costs might outweigh the O(n) vs O(n log n) theoretical benefit.

So, I'm afraid I'm not at all certain that this is actually faster in all scenarios. Do you have any benchmarks or perf traces to provide as evidence?

Sometimes what is theoretically superior results in worse performance [[1](https://github.com/NuGet/NuGet.Client/pull/5303#discussion_r1274806397), [2](https://www.zivkan.com/blog/nuget-version-performance-deep-dive/#nugetcopy5)]. It's why, to the best of my knowledge, good Quick Sort implementations often fall back to an N^2 sort for small arrays, because despite the theoretical Big O advantage of Quick Sort, in practice it's slower in some scenarios.

A challenge with this PR is it appears to have systematically replaced every call to `OrderedEquals`. A whole lot of the code changed is not relevant to restore, so those won't affect restore performance, or `RemoteDespendencyWalker`, which this issue is linked to. But even ones that are used in restore, there's a huge variety in ways it's being used.

In `SpecValidationUtility`, the key is a string, so a "primitive" type, where getting the hash code should be fast. Also, there will be 1 item in the collection for every inner-build a multi-targeting project performs. This is typically small, and even with MAUI apps isn't likely to be more than about 6.

On the other hand, the changes under `NuGet.ProjectModel` are frequently using a class as the key, so the class' `GetHashCode` implementation needs to be called, and there can be hundreds of entries, depending on how many packages are in each project's restore graph.

The linked issue says that `RemoteDependencyWalker` allocates a lot, but I wonder if systematically changing everything is being too blunt?

Another idea would be to change the `OrderedEquals` method signature from `IEnumerable<TSource>` to `ICollection<TSource>`, then instead of using LINQ's `OrderBy`. we can first compare the length of the two collections and early exist when they're different. When they're the same, we can pre-allocate a List of the correct length, so there will be no resizing while inserting.

Unfortunately NuGet HEAVILY uses `IEnumerable` all over the place, so types might not be castable to `ICollection` at the location that `OrderedEquals`, or now `ElementsEqual`, is called. But in the places that they are, we can guarantee a maximum of one single intermediate array.

If the goal is only to reduce allocations, not necessarily CPU complexity, then ensuring we allocate temporary arrays of the correct size, so they don't need to be resized, has a better theoretical result than using a `Dictionary`.

Erarndt: The primary goal is to reduce the allocations in this helper function. An improvement to CPU time would be a nice side benefit, but I would consider this a win as long as the allocations in this path are improved and CPU time gets no worse. Here's an OrchardCore trace I took that might help show the number of unexpected allocations that are happening in this simple helper:

image

A few things to note:

1. This is in the `ICollection<T>` overload for `OrderedEquals`, so this represents the best case scenario for opting out of work early. There are currently overloads for `IEnumerable<T>`, `ICollection<T>` and `IList<T>`. I can leave the current implementation of `OrderedEquals` in place and add the `ElementsEqual` overload, but I'd be curious what the justification for this would be.
2. OrchardCore is a large repo, but there are far worse cases that we've seen in the wild. I don't have their code, so I can't do a comparison.
3. Each call to `OrderBy` will create two arrays the size of the original collection. It caches the result of the delegate you pass in (`ComputeKeys`) so potentially expensive calculations are only done once and then creates another mapping array in the sort algorithm. This means that we create 4 arrays for each call to `OrderedEquals` when we can't opt out early. The proposed implementation allocates one dictionary.
4. Just invoking and using `OrderBy` creates an instance of `System.Linq.EnumerableSorter` and `System.Linq.OrderedEnumerable`

Erarndt: I spent some time digging into this more to see if I can streamline things further. I had the following assumptions when looking into this:

1. The collections that are compared will generally be small. Most of the time they'll be smaller than 20 items.
2. Both CPU usage and allocations are important to improve.
3. Calculating `HashCode()` for certain types can be unexpectedly expensive, so that needs to be taken into consideration as well.

I came up with an additional approach that is essentially a streamlined version of the existing approach. I ran some benchmarks using two different collection types of several different sizes (`List<string>` and `List<CustomStruct>` where the struct has several fields used in both the `Equals` and `HashCode` implementations).

| Method                 | Count | Mean         | Error       | StdDev      | Gen0    | Allocated |
|----------------------- |------ |-------------:|------------:|------------:|--------:|----------:|
| OrderedEquals          | 1     |     445.2 ns |     4.68 ns |     3.91 ns |       - |         - |
| NewOrderedEquals       | 1     |     411.9 ns |     3.10 ns |     2.90 ns |       - |         - |
| ElementsEqual          | 1     |     324.6 ns |     1.77 ns |     1.57 ns |       - |         - |
| OrderedEqualsStruct    | 1     |     458.4 ns |     2.26 ns |     2.11 ns |       - |         - |
| NewOrderedEqualsStruct | 1     |     493.8 ns |     3.33 ns |     2.60 ns |       - |         - |
| ElementsEqualStruct    | 1     |     672.2 ns |     3.69 ns |     3.45 ns |       - |         - |
| OrderedEquals          | 3     |   4,784.0 ns |    30.32 ns |    25.32 ns |  0.6256 |    3285 B |
| NewOrderedEquals       | 3     |   3,902.1 ns |    21.38 ns |    18.95 ns |  0.1373 |     721 B |
| ElementsEqual          | 3     |   4,406.1 ns |    35.92 ns |    31.84 ns |  0.2518 |    1322 B |
| OrderedEqualsStruct    | 3     |   7,219.3 ns |    28.97 ns |    24.20 ns |  1.0529 |    5528 B |
| NewOrderedEqualsStruct | 3     |   5,255.2 ns |    19.15 ns |    16.97 ns |  0.3204 |    1683 B |
| ElementsEqualStruct    | 3     |   9,992.9 ns |   135.85 ns |   127.07 ns |  0.3357 |    1803 B |
| OrderedEquals          | 5     |   7,263.9 ns |    29.85 ns |    24.93 ns |  0.7172 |    3766 B |
| NewOrderedEquals       | 5     |   7,140.5 ns |    45.87 ns |    40.66 ns |  0.1678 |     881 B |
| ElementsEqual          | 5     |   7,517.4 ns |    73.48 ns |    68.74 ns |  0.3967 |    2123 B |
| OrderedEqualsStruct    | 5     |  12,006.9 ns |    69.00 ns |    57.62 ns |  1.3885 |    7291 B |
| NewOrderedEqualsStruct | 5     |  12,565.4 ns |    64.77 ns |    57.42 ns |  0.4730 |    2484 B |
| ElementsEqualStruct    | 5     |  16,187.5 ns |    92.44 ns |    81.95 ns |  0.6104 |    3245 B |
| OrderedEquals          | 10    |  15,274.5 ns |   116.26 ns |    97.09 ns |  0.9460 |    4967 B |
| NewOrderedEquals       | 10    |  15,704.8 ns |   168.35 ns |   149.23 ns |  0.2441 |    1282 B |
| ElementsEqual          | 10    |  13,414.9 ns |    45.45 ns |    42.51 ns |  0.5493 |    2924 B |
| OrderedEqualsStruct    | 10    |  26,406.7 ns |   168.53 ns |   140.73 ns |  2.2278 |   11697 B |
| NewOrderedEqualsStruct | 10    |  28,651.1 ns |   105.64 ns |    93.64 ns |  0.8545 |    4487 B |
| ElementsEqualStruct    | 10    |  32,005.4 ns |   194.28 ns |   172.23 ns |  0.8545 |    4687 B |
| OrderedEquals          | 20    |  32,282.7 ns |   643.06 ns |   836.16 ns |  1.4038 |    7371 B |
| NewOrderedEquals       | 20    |  34,858.7 ns |   177.51 ns |   166.04 ns |  0.3662 |    2084 B |
| ElementsEqual          | 20    |  27,010.6 ns |   106.13 ns |    82.86 ns |  1.0071 |    5328 B |
| OrderedEqualsStruct    | 20    |  52,247.3 ns |   279.81 ns |   248.04 ns |  3.9063 |   20511 B |
| NewOrderedEqualsStruct | 20    |  67,056.5 ns |   306.45 ns |   255.90 ns |  1.5869 |    8493 B |
| ElementsEqualStruct    | 20    |  66,288.7 ns |   413.94 ns |   387.20 ns |  1.7090 |    9014 B |
| OrderedEquals          | 100   | 258,672.4 ns | 1,528.21 ns | 1,276.13 ns |  4.8828 |   26602 B |
| NewOrderedEquals       | 100   | 294,398.9 ns | 2,224.79 ns | 1,972.22 ns |  1.4648 |    8493 B |
| ElementsEqual          | 100   | 134,683.5 ns |   680.95 ns |   603.65 ns |  4.1504 |   22157 B |
| OrderedEqualsStruct    | 100   | 385,531.2 ns | 3,490.08 ns | 3,093.87 ns | 17.0898 |   91019 B |
| NewOrderedEqualsStruct | 100   | 478,474.9 ns | 1,804.18 ns | 1,599.36 ns |  7.3242 |   40551 B |
| ElementsEqualStruct    | 100   | 332,537.5 ns | 1,321.26 ns | 1,103.31 ns |  7.3242 |   39301 B |


The benchmarks indicate that there's a pretty clear winner in terms of allocations with `NewOrderedEquals` for either reference types, small structs, or small collections. It's only the combination of larger collections of larger structs where this approach starts to lose out. Additionally, `NewOrderedEquals` either edges out or is roughly in line with the existing implementation as far as CPU time goes. I'm leaning towards switching to this implementation, but I'm curious what you think.


Here's the implementation for `NewOrderedEquals`:

        internal static bool OrderedEquals<TSource>(this IList<TSource>? self, IList<TSource>? other, Comparison<TSource> orderComparer, IEqualityComparer<TSource>? sequenceComparer = null)
        {
            Debug.Assert(sequenceComparer != null || typeof(TSource) != typeof(string), "Argument " + "sequenceComparer" + " must be provided if " + "TSource" + " is a string.");

            if (TryIdentityEquals(self, other, out bool identityEquals))
            {
                return identityEquals;
            }

            if (self.Count != other.Count)
            {
                return false;
            }

            if (self.Count == 0)
            {
                return true;
            }

            if (sequenceComparer == null)
            {
                sequenceComparer = EqualityComparer<TSource>.Default;
            }

            if (self.Count == 1)
            {
                return sequenceComparer.Equals(self[0], other[0]);
            }

            TSource[] selfCopy = new TSource[self.Count];
            TSource[] otherCopy = new TSource[other.Count];
            self.CopyTo(selfCopy, 0);
            other.CopyTo(otherCopy, 0);

            Array.Sort(selfCopy, orderComparer);
            Array.Sort(otherCopy, orderComparer);

            for (int i = 0; i < selfCopy.Length; ++i)
            {
                if (!sequenceComparer.Equals(selfCopy[i], otherCopy[i]))
                {
                    return false;
                }
            }

            return true;
        }

zivkan: Thank you for doing the benchmark, trying different approaches. I think the results show what I was trying to explain in my previous, long message, that what might feel initially intuitive might end up having worse results than other approaches.

I agree, the new ordered equals seems better. Unfortunately a practical concern will be that NuGet frequently passes collections around as `IEnumerable<T>`, so the places you want to use this might not be not implicitly castable as `IList<T>`.  But I think the best option would be to take this `IList<T>` implementation, and then where perf matters, change the product code to stop using `IEnumerable<T>`, so we can use the `IList<T>` method. This might meaning some API breaking changes, but honestly I'm getting a little tired of being prevented to make improvements because of API design bugs, so I'm thinking we need to be a little more aggressive at changing this, and forcing our NuGet Client SDK customers to change where it causes breaks.

Another question is that maybe the `self.Count == 1` optimization is uncommon in NuGet's practical usage (I have absolutely no idea, I have no data for either side of the argument). So it if helps make the method more usable, changing the method signature to `ICollection<T>` instead might make it a little more reusable. The method could also try an `if (self.Count == 1 && self is IList<T> && other is IList<T>)` if that perf optimization is really beneficial, but the two typecasts of course have a CPU cost.  However, my gut feeling is that `ICollection<T>` will not allow much more usage than `IList<T>`. So, an `IList<T>` implementation, and migrating other product code away from `IEnumerable<T>` is probably the best option.

Erarndt: I'm slightly confused. The current `OrderedEquals` has three different overloads `IEnumerable<T>`, `ICollection<T>`, `IList<T>`. Are you suggesting that we attempt to consolidate into one? For, `IEnumerable<T>`, we can do something like this:

        internal static bool OrderedEquals<TSource>(this IEnumerable<TSource>? self, IEnumerable<TSource>? other, Comparison<TSource> orderComparer, IEqualityComparer<TSource>? sequenceComparer = null)
        {
            Debug.Assert(sequenceComparer != null || typeof(TSource) != typeof(string), "Argument " + "sequenceComparer" + " must be provided if " + "TSource" + " is a string.");

            bool identityEquals;
            if (TryIdentityEquals(self, other, out identityEquals))
            {
                return identityEquals;
            }

            List<TSource> selfCopy = new List<TSource>(self);

            int otherCount = 0;
            TSource[] otherCopy = new TSource[selfCopy.Count];
            foreach (TSource item in other)
            {
                if (otherCount >= selfCopy.Count)
                {
                    return false;
                }

                otherCopy[otherCount] = item;
                ++otherCount;
            }

            if (selfCopy.Count != otherCount)
            {
                return false;
            }

            if (selfCopy.Count == 0)
            {
                return true;
            }

            if (sequenceComparer == null)
            {
                sequenceComparer = EqualityComparer<TSource>.Default;
            }

            if (selfCopy.Count == 1)
            {
                return sequenceComparer.Equals(selfCopy[0], otherCopy[0]);
            }

            selfCopy.Sort(orderComparer);
            Array.Sort(otherCopy, orderComparer);

            for (int i = 0; i < selfCopy.Count; ++i)
            {
                if (!sequenceComparer.Equals(selfCopy[i], otherCopy[i]))
                {
                    return false;
                }
            }

            return true;
        }

- We want to avoid multiple enumeration, so we consume an `IEnumerable<T>` and use `List<T>` to handle the auto expansion
- The second `IEnumerable<T>` needs to be exactly the same size as the first one so we can optimistically allocate an array once instead of relying on allocating potentially many intermediate arrays.
- We check to ensure that we don't exceed the size of the array and then afterward to make sure they're the same size. We can short circuit if either of these checks fail.
- Otherwise, we can run the same logic in the other overloads by sorting and doing a pairwise comparison.

I think the `self.Count == 1` might be uncommon but it's a very cheap check to avoid several allocations. Unless we have evidence that it's actively hurting, I'd be reluctant to remove it.

build/Shared/EqualityUtility.cs
```diff
@@ -14,48 +14,72 @@ namespace NuGet.Shared
...
+        /// <param name="equalityComparer">An optional comparer for sequences</param>
+        internal static bool ElementsEqual<TSource, TKey>(this ICollection<TSource>? self, ICollection<TSource>? other, Func<TSource, TKey> keySelector, IEqualityComparer<TKey>? equalityComparer = null) where TKey : notnull
         {
-            Debug.Assert(orderComparer != null || typeof(TKey) != typeof(string), "Argument " + "orderComparer" + " must be provided if " + "TKey" + " is a string.");
-            Debug.Assert(sequenceComparer != null || typeof(TSource) != typeof(string), "Argument " + "sequenceComparer" + " must be provided if " + "TSource" + " is a string.");
+            Debug.Assert(equalityComparer != null || typeof(TKey) != typeof(string), "Argument " + nameof(equalityComparer) + " must be provided if " + nameof(TKey) + " is a string.");
 
-            bool identityEquals;
-            if (TryIdentityEquals(self, other, out identityEquals))
+            if (TryIdentityEquals(self, other, out bool identityEquals))
```
nkolev92, line 97: this is just an opinion, not an ask to make chnages, but for PRs that are heavy on changes like this one, some of the non-functional changes make it more challenging to review.

build/Shared/EqualityUtility.cs
nkolev92: Can we please add the description to the PR body? 

There's lots to read here :)

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
```diff
@@ -305,8 +305,8 @@ public bool Equals(RuntimeGraph? other)
                 return true;
             }
 
-            return Runtimes.OrderedEquals(other.Runtimes, pair => pair.Key, StringComparer.Ordinal)
-                && Supports.OrderedEquals(other.Supports, pair => pair.Key, StringComparer.Ordinal);
+            return Runtimes.ElementsEqual(other.Runtimes, pair => pair)
```
nkolev92: These are Dict/Sets, should we an equals utility for dict/sets? This would avoid the extra dic allocation.

nkolev92: I think that feedback applies to a few more types.

nkolev92 (commented):
It'd be great to add a PR description in the PR body + the numbers in the description.

I think we should be able to special case dict/sets as well.

build/Shared/EqualityUtility.cs
zivkan: > I'm slightly confused.

Me too 😆 

> The current `OrderedEquals` has three different overloads `IEnumerable<T>`, `ICollection<T>`, `IList<T>`.

This comment thread started inside the `IEnumerable<T>` overload, then you provided a benchmark that requires `IList<T>`. I didn't pay enough attention at the time, but now I see the implementation of the 3 overloads are the same algorith, with the differences being `count == 1` optimizations and being able to use `[]` indexing, rather than an enumerator, where appropriate.

> Are you suggesting that we attempt to consolidate into one?

In the long term that sounds great, but it is probably outside the scope of this PR.

For what it's worth, I loaded up `NuGet-src.slnf`, and VS is telling me there are 16 references to the `IEnumerable<T>` overload, 4 to the `ICollection<T>` overload, and 18 to the `IList<T>` overload. With nearly 40 calls to `OrderedEquals` across the 3 overloads, how confident are we that we can choose one algorithm that will be good for all scenarios? You provided microbenchmarks, but making assumptions on the types and collection size, which we don't know how realistic those assumptions are for NuGet's actual usage. On the other hand, maybe the call count in a typical restore is so low that it doesn't matter. Or maybe the call count is so high it matters a lot.  I'm just stressing myself out now 😆 

I don't want to make perfect the enemy of good, but I'd like some confidence that we're not unintentionally making things worse. I fear it, because I've wasted time having done it multiple times, despite feeling like my change was supposed to be theoretically better. While memory allocations might be the most important perf metric to Visual Studio, all the code in `src/NuGet.Core/` also runs on CI machines where wall clock time is far more important. When one implementation is faster, but another uses less memory, how do we choose which implementation to use, for the code that runs restore on both CI agents and also in VS?

To reiterate a previous point, [there was a previous PR](https://github.com/NuGet/NuGet.Client/pull/5303#discussion_r1274806397) that reduced allocations in a class that VS telemetry/Watson flagged as causing high allocations, but when I benchmarked the code it was slower, despite fewer memory allocations and GCs. With `OrderedEquals` being called with various different collection sizes and various different types (primitive types, classes, and structs), what if some scenarios become worse, while others are better? How do we know this PR is a net positive? I agree it theoretically sounds like it should, but so was the closed/abandoned PR, plus some other investigations I never brought to PR (and [sometimes ones I did bring to PR](https://github.com/NuGet/NuGet.Client/pull/3040))

Erarndt: > how confident are we that we can choose one algorithm that will be good for all scenarios?
The existing code used the same algorithm for each overload. If we have evidence that we can improve on this, I'm all for it. So far, I haven't seen anything that would indicate such a change would make a difference past the smaller optimizations that I included in the updates.
 
> I don't want to make perfect the enemy of good, but I'd like some confidence that we're not unintentionally making things worse. I fear it, because I've wasted time having done it multiple times, despite feeling like my change was supposed to be theoretically better.

I definitely don't want to unintentionally make things worse. You reference other PRs where this happens. Is there a set of tests that the team consistently runs against changes to verify that there aren't regressions? It sounds like these performance scenarios are rather sensitive, and there is a high likelihood of regressions. How are you validating performance for code changes?

> While memory allocations might be the most important perf metric to Visual Studio

I don't think this is true for VS. VS cares about the memory footprint and memory allocations that impact GC time, but CPU time is on equal footing. The CPU time in the benchmarks were reasonably close to the existing implementation (about the same or better depending on usage), but both proposed new implementations were significantly better on memory. I don't want my comment to be misinterpreted as "VS doesn't care about CPU time".

> all the code in `src/NuGet.Core/` also runs on CI machines where wall clock time is far more important.

This aligns with VS goals. Since no managed code can run during blocking GCs, the number of allocations directly impacts wall clock time. For a concrete example of this, I benchmarked the NuGet CLI doing a restore on OrchardCore:

image

The GC pause time accounts for 11.5% of total execution time which is ~14.5 seconds in this trace. During restore, there were 167 collections that were run. All of this matters for wall clock time. Note that we have significantly worse examples coming from customers in the wild.

> With `OrderedEquals` being called with various different collection sizes and various different types (primitive types, classes, and structs), what if some scenarios become worse, while others are better? How do we know this PR is a net positive?

Fair questions. We do the best we can with the information available. The better information we have about how this is used, the better we can tune the code. It's the same reasoning that applies in the sorting example you provide above. That example has more uncertainty since the framework has to operate on an open set of types which can have various performance characteristics. In this case, we have an internal utility method with significantly more control on how it's used. I'd propose benchmarking with scenarios that align with expected use cases and see where we end up. Are there cases you think that need to be considered that haven't been?

zivkan: tl;dr: I was being overly idealistic about choosing between the dict or list based methods, and forgetting a benchmarking pitfall.

Firstly, our product code uses `OrderedEquals` in around 40 places, with various different types. The benchmark tested two data types, strings and a fake struct that isn't actually in the product, and found different perf characteristics for the two different data types. But more on this below.

I was contacted offline by someone else, telling me my comments are disconnected from the data provided and that I'm slowing down this PR preventing other work from being worked on. So, I just want to reiterate the point that the benchmark data provided shows that the current LINQ based OrderedSort is fastest for structs of collection size 5 and 10, despite whatever GC pauses the benchmark experienced.

What I was wondering is if we should use the Memory Watson data to figure out which of the 40 calls into OrderedEquals are being flagged, and then target those call sites, rather than changing the implementation of all 40 calls.  If we're unlucky and NuGet's real world usage is most similar to structs of collection sizes 5 or 10, then the data provided suggests perf will be worse.

However, the benchmarking pitfall that I forgot about in my earlier reviews is that the benchmark only tests a single thread. It ignores practical perf of an entire app that would be doing other work on other threads at the same time. If the reduced allocations benefit the real world app performance, despite the single-threaded microbenchmark showing worse wall-clock time, then this wasn't provided in the data, and I don't remember it being brought up as a discussion point earlier in this thread. When I've got a ton of other work to do, and reviewing PRs is preventing me from my other work, in my opinion it's not trivial to remember this. When presented with benchmark results, it's too easy to just take the duration column as "fact", which again, shows the existing code fastest in some scenarios. So I think my comments were in relation to the information & data provided in this PR.

Anyway, now multi-threading has been brought to my attention. I wrote a microbenchmark about reading many files with and without async, and got unexpected results, so I was thinking for a long time to try to guess why it might behave like this. So, I didn't even benefit from someone pointing this out to me, I had to spend extra time figuring it out for myself, and only did so by accident while investigating something else. But I am now confident that this PR will make an overall perf improvement, despite benchmark results presented suggesting the opposite might be true.

> The CPU time in the benchmarks were reasonably close to the existing implementation (about the same or better depending on usage), but both proposed new implementations were significantly better on memory.

I wasn't clear in my previous comments, but I was thinking about whether we should choose the List based NewOrderedEquals, or the Dictionary based ElementsEqual. Unless I made mistakes in my copy-pastes, here's what I see comparing the perf of the two:

|collection size|strings|structs|
|--|--|--|
|1|+27%|-27%|
|3|-11%|+37%|
|5|-5%|-22%|
|10|+17%|-22%|
|20|+29%|+1%|
|100|+43%|+43%|

So, while you're right that they're "reasonably close" in CPU time in absolute numbers (micro-seconds), I think that 22%, 37%, and 43% differences are not "reasonably close" in percentage terms.

What's worse is that at a collection size of 3, one of the two is better for strings, whereas for structs the other is better, and there approx 30-40% relative difference between the two which could be very significant if called a large number of times. But then for a single data type, the better method also changes from 11% worse off to 43% better off for strings, or 22% worse off to 43% better off, as collection size changes. If we're swapping all 40 calls to a single implementation, which do we choose? Data suggests that some of those 40 will be better off with the List method and the others will be better off with the Dict method.

For example, if `OrderedEquals` is used for comparing an assets file TFM list, since we know the collection size will always be small (even MAUI projects I'd expect less than 10, but for non-MAUI projects we'd probably expect less than 4, most typically just 1 or 2), and the data type is basically a string, we know that the List implementation is better. But for the assets file packages list, or packages lock files, if that's ever compared using `OrderedEquals`, it's not particularly difficult for the number of packages to be over 100, so the Dict implementation is clearly better there.

Maybe the data is saying that we need both implementations, and a brief investigating into the 40 calls sites (or a subset that are actually being used frequently) should be done to choose which method each of those 40 should use.

And all of this is ignoring that the benchmark results showed that for structs, the existing LINQ based OrderedEquals was faster than either of the new methods, despite the increased allocations. The data shows that a collection size of 20, both the List and Dict methods are similar wall clock time, and the existing Linq method is about 20% faster. That's not "reasonably close" in percentage terms. However, the single threaded benchmark probably means that the benchmark was actually misleading about wall-clock time when used in a multi-threaded restore. but 20% is a lot to have to make up for.

Finally, a collection of 3 structs, the benchmark results show that the Dict approach 38% slower than the existing LINQ method, which is very big in percentage terms (again, only relevant if called many times, like in a loop). Whereas for the collection size of 100 structs, the List method is 24% slower than the LINQ method (I don't understand why this would be the case, more CPU time in copying memory?). So again, the benchmark data shows that if we make the wrong choice, that this PR could theoretically result in worse perf. Is the reduced allocations and perf benefit in multi-threaded scenarios worth the tradeoff of 20+% reduction in single threaded perf?   Data is insufficient to prove anything. I assume that I'm being pressured into assuming yes, memory reductions are always better, but IMO the data presented doesn't prove this.

This is why I was asking questions. The only thing the data is clear about is that both the List and Dict methods have a lot fewer allocations. But if you actually look closely at the benchmark results, I don't agree that there's anything clear about the data provided. As you can tell, I'm bitter about being told my comments are disconnected from the data provided. As I implied earlier, if all we care about is reducing allocations, then sure, this PR is a no-brainer. If we care about wall-clock performance (taking into account multi-treading and GC pauses), I don't think the data provided gives us an answer.

> Is there a set of tests that the team consistently runs against changes to verify that there aren't regressions?

Consistently, no. Not even regularly. We have these scripts, which only get used when someone really wants to investigate: https://github.com/NuGet/NuGet.Client/blob/dev/scripts/funcTests/runFuncTests.sh

Unfortunately we don't have anything like VS Perf team's, or the .NET runtime/library team's, automated perf tests. Various people in the NuGet team have tried implementing an automated perf pipeline several times. But we don't have the benefit of a team dedicated to nothing but perf, so we have to prioritize vs other work. It would be absolutely wonderful to have, especially since there are so many scenarios that could have a significant perf impact. Package source mapping or not. lock file or not. CPM or not. Some of these features have mutually exclusive code paths, so it's not possible to have a single test solution to exercise all perf code paths. There's also very considerable differences between restore in VS and the command line.

Anyway, I'm going on vacation, so won't be monitoring this PR any more. I hoped that by deeply analyzing the benchmark results that I would have proved to myself that the feedback I got was correct and that I was indeed being unreasonable with my questions, rather than just accepting the PR.  But since the List method is 20+% slower for a collection of 100 structs, while the Dict is 38% slower for 3 structs, I just can't convince myself that I am being unreasonable, unless of course none of these 40 calls are in a tight loop. If OrderedEquals is only being called a few times per project per restore, then even in a 1000+ solution these microsecond differences won't add up to much. But if each project can result in OrderedEquals being called more than 1000 times, then in a 1000+ project solution these small differences can result in multi-second delays.

src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeGraph.cs
Erarndt: Happy to add more overloads if we want the extra code. In addition to the cases you mentioned, There are a handful of calls using `IReadOnlyList<T>` that bind to the `IEnumerable<T>` overload which causes inefficient allocation of the intermediate collections.

nkolev92: These comparisons happen often enough where it's worth it.

Erarndt: Clarification question for you. The current implementation (here and in other places) compares keys using a specified string comparer. However, this could lead to unexpected issues if the dictionaries that are compared are constructed using different compareres. for example:

    var dictionaryA = new Dictionary<string, bool>(StringComparer.Ordinal);
    dictionaryA.Add("foo", true); // ContainsKey("FOO") returns false
    var dictionaryB = new Dictionary<string, bool>(StringComparer.OrdinalIgnoreCase);
    dictionaryB.Add("FOO", true); // ContainsKey("foo") returns true

If we use the existing implementation of `OrderedEquals`, the dictionaries are considered different. If we Enumerate the keys of one dictionary and try to do a lookup in the other dictionary, the order we do that in matters. Curious about what you think the expected behavior is for NuGet's scenarios.

nkolev92: I'd expect we're not trying to compare dictionaries/hashsets with different comparers. Those would be bugs. 

Should we compare the comparers in the equals method?

Erarndt: Sounds good. I was hoping to verify that those cases are unexpected.

As far as checking comparers, the exposed type is `IReadOnlyDictionary<TKey, TValue>`. I can check the underlying type to see if it has the comparer exposed. Looks like the underlying type is `Dictionary<TKey, TValue>`, so that should work.

build/Shared/EqualityUtility.cs
Erarndt: TL;DR: Benchmarking is difficult.

I appreciate your input and skepticism as it encourages me to dig deeper and back up changes I propose with data and sound reasoning. I made several errors when benchmarking that led to some confusion:

1. The random data was too random. What I mean by this is that the random strings I was generating were random between runs. Since the majority of the cost is in the sort, the results can be skewed based on what the data is.
2. The execution time of the tests was way too low. We're running into issues with the resolution of the hardware timer that's effectively giving us "random" results. For example, consider this code:

            const int stopwatchRuns = 100000;
            Stopwatch sw = new Stopwatch();
            long nanosecPerTick = (1000L * 1000L * 1000L) / Stopwatch.Frequency; // from https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch?view=net-8.0

            // warm up
            for (int i = 0; i < stopwatchRuns; ++i)
            {
                sw.Restart();
                sw.Stop();
            }

            List<long> ticks = new List<long>(stopwatchRuns);
            for (int runs = 0; runs < 10; ++runs)
            {
                for (int i = 0; i < stopwatchRuns; ++i)
                {
                    sw.Restart();
                    sw.Stop();
                    ticks.Add(sw.ElapsedTicks);
                }

                Console.WriteLine($"Run {runs + 1}:");
                Console.WriteLine($"Min {ticks.Min() * nanosecPerTick} ns");
                Console.WriteLine($"Max {ticks.Max() * nanosecPerTick} ns");
                Console.WriteLine($"Avg {ticks.Average() * nanosecPerTick} ns");
                Console.WriteLine();
                ticks.Clear();
            }

Output:

    Run 1:
    Min 0 ns
    Max 162800 ns
    Avg 27.237 ns

    Run 2:
    Min 0 ns
    Max 1900 ns
    Avg 23.666 ns

    Run 3:
    Min 0 ns
    Max 1200 ns
    Avg 24.719 ns

    Run 4:
    Min 0 ns
    Max 7000 ns
    Avg 24.905 ns

    Run 5:
    Min 0 ns
    Max 1300 ns
    Avg 24.593 ns

    Run 6:
    Min 0 ns
    Max 80700 ns
    Avg 25.877 ns

    Run 7:
    Min 0 ns
    Max 600 ns
    Avg 23.65 ns

    Run 8:
    Min 0 ns
    Max 10700 ns
    Avg 24.711 ns

    Run 9:
    Min 0 ns
    Max 1200 ns
    Avg 24.39 ns

    Run 10:
    Min 0 ns
    Max 1200 ns
    Avg 24.5 ns

There's a potential for a wide variance range just based on the timer resolution, and frankly, I was surprised to see just how large that variance can be (over 160,000 ns??). I did some digging and found out that the [recommendations from benchmark.net](https://github.com/dotnet/performance/blob/main/docs/microbenchmark-design-guidelines.md#iterationsetup) are that a single invocation of the benchmark code needs to run longer than 100ms or **"your benchmark will be entirely invalid**". They provide an example for testing `span.Slice()` where they demonstrate how to use the `OperationsPerInvoke` property of the `Benchmark` attribute to set up benchmarks. I adjusted the benchmarks to run each method 10000 times per invocation and got the following results:

| Method                 | Count | Mean        | Error     | StdDev    | Median      | Gen0   | Allocated |
|----------------------- |------ |------------:|----------:|----------:|------------:|-------:|----------:|
| OrderedEquals          | 3     |    448.6 ns |   0.79 ns |   0.70 ns |    448.5 ns | 0.0625 |     328 B |
| NewOrderedEquals       | 3     |    282.1 ns |   1.65 ns |   1.29 ns |    282.2 ns | 0.0090 |      48 B |
| ElementsEqual          | 3     |    374.6 ns |   2.68 ns |   2.24 ns |    374.0 ns | 0.0250 |     132 B |
| OrderedEqualsStruct    | 3     |    521.7 ns |   3.56 ns |   3.16 ns |    520.0 ns | 0.1047 |     553 B |
| NewOrderedEqualsStruct | 3     |    371.1 ns |   5.55 ns |   5.70 ns |    369.1 ns | 0.0273 |     144 B |
| ElementsEqualStruct    | 3     |    503.8 ns |   9.99 ns |  17.23 ns |    494.6 ns | 0.0344 |     180 B |
| OrderedEquals          | 5     |    635.5 ns |   3.42 ns |   3.20 ns |    635.4 ns | 0.0711 |     377 B |
| NewOrderedEquals       | 5     |    514.2 ns |   6.07 ns |   5.38 ns |    511.8 ns | 0.0117 |      64 B |
| ElementsEqual          | 5     |    625.2 ns |   1.59 ns |   1.24 ns |    625.1 ns | 0.0398 |     212 B |
| OrderedEqualsStruct    | 5     |    828.6 ns |   3.84 ns |   3.20 ns |    827.7 ns | 0.1375 |     729 B |
| NewOrderedEqualsStruct | 5     |    983.4 ns |   8.32 ns |   7.78 ns |    980.6 ns | 0.0422 |     224 B |
| ElementsEqualStruct    | 5     |    756.4 ns |   2.26 ns |   1.89 ns |    756.3 ns | 0.0609 |     324 B |
| OrderedEquals          | 10    |  1,405.5 ns |  27.70 ns |  41.46 ns |  1,429.2 ns | 0.0938 |     497 B |
| NewOrderedEquals       | 10    |  1,244.1 ns |  21.78 ns |  18.18 ns |  1,251.8 ns | 0.0188 |     104 B |
| ElementsEqual          | 10    |  1,277.5 ns |   9.18 ns |   8.59 ns |  1,280.2 ns | 0.0547 |     292 B |
| OrderedEqualsStruct    | 10    |  1,895.9 ns |  37.10 ns |  39.69 ns |  1,914.1 ns | 0.2219 |    1170 B |
| NewOrderedEqualsStruct | 10    |  2,470.2 ns |  28.50 ns |  26.66 ns |  2,479.5 ns | 0.0781 |     425 B |
| ElementsEqualStruct    | 10    |  1,496.3 ns |  29.00 ns |  24.22 ns |  1,494.9 ns | 0.0891 |     469 B |
| OrderedEquals          | 20    |  3,350.4 ns |  10.83 ns |   9.05 ns |  3,349.3 ns | 0.1400 |     737 B |
| NewOrderedEquals       | 20    |  2,704.4 ns |  50.01 ns |  46.78 ns |  2,722.8 ns | 0.0344 |     184 B |
| ElementsEqual          | 20    |  2,626.7 ns |  38.81 ns |  36.30 ns |  2,639.2 ns | 0.1000 |     533 B |
| OrderedEqualsStruct    | 20    |  4,217.5 ns |  28.02 ns |  24.84 ns |  4,220.5 ns | 0.3833 |    2051 B |
| NewOrderedEqualsStruct | 20    |  5,742.5 ns | 108.31 ns | 101.31 ns |  5,782.8 ns | 0.1556 |     825 B |
| ElementsEqualStruct    | 20    |  2,863.1 ns |  49.27 ns |  46.09 ns |  2,883.3 ns | 0.1719 |     901 B |
| OrderedEquals          | 50    |  9,590.1 ns | 142.18 ns | 132.99 ns |  9,633.7 ns | 0.2667 |    1458 B |
| NewOrderedEquals       | 50    |  8,784.3 ns | 103.62 ns |  96.93 ns |  8,805.7 ns | 0.0667 |     425 B |
| ElementsEqual          | 50    |  6,024.2 ns | 117.51 ns | 104.17 ns |  6,060.5 ns | 0.2333 |    1254 B |
| OrderedEqualsStruct    | 50    | 14,427.5 ns | 179.86 ns | 168.24 ns | 14,518.8 ns | 0.8750 |    4695 B |
| NewOrderedEqualsStruct | 50    | 17,393.8 ns | 166.90 ns | 156.12 ns | 17,451.3 ns | 0.3667 |    2027 B |
| ElementsEqualStruct    | 50    |  7,343.9 ns |  76.71 ns |  71.75 ns |  7,363.2 ns | 0.4125 |    2199 B |
| OrderedEquals          | 100   | 23,395.1 ns | 445.98 ns | 438.02 ns | 23,566.3 ns | 0.5000 |    2660 B |
| NewOrderedEquals       | 100   | 22,119.5 ns | 125.79 ns | 117.66 ns | 22,155.0 ns | 0.1333 |     825 B |
| ElementsEqual          | 100   | 12,663.6 ns |  77.72 ns |  72.70 ns | 12,690.1 ns | 0.4000 |    2216 B |
| OrderedEqualsStruct    | 100   | 32,510.7 ns | 634.46 ns | 968.88 ns | 33,194.0 ns | 1.7000 |    9102 B |
| NewOrderedEqualsStruct | 100   | 39,633.4 ns | 779.35 ns | 957.12 ns | 40,095.9 ns | 0.7000 |    4032 B |
| ElementsEqualStruct    | 100   | 14,768.8 ns | 265.71 ns | 248.55 ns | 14,882.7 ns | 0.7250 |    3930 B |

The first thing to point out is that the memory savings from the `Dictionary` variant don't save much at all when operating on the string collections. The internal `Entry` array is a struct that's sufficiently large to cancel out most of the benefit.

The major takeaway is now `NewOrderedEquals` is always ahead of the existing `OrderedEquals` in CPU time when operating on strings and is significantly lighter on memory use. However, the results are reversed when operating on structs. The difference is primarily caused by a slightly different heuristic used by `Array.Sort()`. If you dig through the implementation, you'll find this note:

        // This is the threshold where Introspective sort switches to Insertion sort.
        // Imperically, 16 seems to speed up most cases without slowing down others, at least for integers.
        // Large value types may benefit from a smaller number.
        internal const int IntrosortSizeThreshold = 16;

The merge sort implementation used by `OrderBy()` doesn't use this heuristic which is why we see the different behavior:

    internal abstract class EnumerableSorter<TElement>
    {
        internal abstract void ComputeKeys(TElement[] elements, int count);

        internal abstract int CompareKeys(int index1, int index2);

        internal int[] Sort(TElement[] elements, int count) {
            ComputeKeys(elements, count);
            int[] map = new int[count];
            for (int i = 0; i < count; i++) map[i] = i;
            QuickSort(map, 0, count - 1);
            return map;
        }

        void QuickSort(int[] map, int left, int right) {
            do {
                int i = left;
                int j = right;
                int x = map[i + ((j - i) >> 1)];
                do {
                    while (i < map.Length && CompareKeys(x, map[i]) > 0) i++;
                    while (j >= 0 && CompareKeys(x, map[j]) < 0) j--;
                    if (i > j) break;
                    if (i < j) {
                        int temp = map[i];
                        map[i] = map[j];
                        map[j] = temp;
                    }
                    i++;
                    j--;
                } while (i <= j);
                if (j - left <= right - i) {
                    if (left < j) QuickSort(map, left, j);
                    left = i;
                }
                else {
                    if (i < right) QuickSort(map, i, right);
                    right = j;
                }
            } while (left < right);
        }
    }

The good news is that the callers of `OrderedEquals` all operate on classes rather than structs, so the `Array.Sort()` heuristic helps in all cases.

3. By not returning the result, I allow for potential compiler and JIT optimizations that misrepresent how this is actually used.

One thing that I haven't properly addressed is how frequently these helper methods are called. I think you'd agree that that's hard to pin down exactly since different restores are going to exercise the code in different ways, but here's an indicator that there are real world restores where these methods are called very frequently.

image

This trace is from OrchardCore and allocations caused by the `IList<T>` overload of `OrderedEquals` accounts for 135MB of allocations (1.2% of all allocs). Doing some estimation based on the benchmark results (1 call allocates ~500B when the input lists are size 10) indicates that this could be called over 270,000 times. Definitely not an exact number and my estimation could be wildly off. It might be worth investigating if we need to be calling this as frequently as we are in addition to streamlining the calls. Other customer traces exhibit worse behavior here.

I hope this alleviates most (all?) concerns about switching the implementation. Curious what your thoughts are.

jeffkl (dismissed):

build/Shared/EqualityUtility.cs
```diff
@@ -112,14 +166,93 @@ internal static class EqualityUtility
...
+            {
+                sequenceComparer = EqualityComparer<TSource>.Default;
+            }
+
+            if (self.Count == 1)
+            {
+                return sequenceComparer.Equals(self[0], other[0]);
+            }
+
+            TSource[] selfCopy = new TSource[self.Count];
```
nkolev92, line 204: The array comparisons are equivalent in the Enumerable, IList, IReadOnlyList implementations.
Can we create a method to avoid accidental issues?

src/NuGet.Core/NuGet.ProjectModel/PackageSpec.cs
```diff
@@ -239,8 +239,8 @@ public bool Equals(PackageSpec other)
                    EqualityUtility.DictionaryEquals(PackInclude, other.PackInclude, (s, o) => StringComparer.Ordinal.Equals(s, o)) &&
                    EqualityUtility.EqualsWithNullCheck(PackOptions, other.PackOptions) &&
 #pragma warning restore CS0612 // Type or member is obsolete
-                   EqualityUtility.OrderedEquals(Dependencies, other.Dependencies, dep => dep.Name, StringComparer.OrdinalIgnoreCase) &&
-                   EqualityUtility.OrderedEquals(TargetFrameworks, other.TargetFrameworks, tfm => tfm.TargetAlias, StringComparer.OrdinalIgnoreCase) &&
+                   EqualityUtility.OrderedEquals(Dependencies, other.Dependencies, (a, b) => StringComparer.OrdinalIgnoreCase.Compare(a.Name, b.Name)) &&
```
kartheekp-ms, line 242: There has been an engaging conversation on this pull request regarding performance improvements. However, I'm curious about the following:

- Why are we using the EqualityUtility.OrderedEquals method in the first place?
- Do we know how many times the `OrderBy` method is invoked on a single object? Specifically, let’s examine the case where the Equals method is called on a single `PackageSpec` object with varying argument values for the `other` parameter. If this occurs, it suggests that we are sorting the same collection—namely, `this.Dependencies`—multiple times. Such redundancy could be avoided if the collection is sorted at the first place.
- Have we considered maintaining the collection in sorted order? Specifically, could we use a `SortedList` instead of a regular `List`? Doing so might eliminate the need to later invoke `OrderBy` or `Sort` methods. After examining a few callers of the `EqualityUtility.OrderedEquals` method, it seems that the collection is sorted based on a single key rather than projecting on different keys. The documentation states that [`SortedList` has an O(n) time complexity for inserting elements](https://learn.microsoft.com/dotnet/api/system.collections.generic.sortedlist-2?view=net-8.0). Perhaps one option is to continue using the List object but sort the data within the collection only once, possibly using the `List.Sort` method..
- I don’t see the value of `GetHashCode` being cached for objects that invoke this API (EqualityUtility.OrderedEquals). Maybe that would help to speed up the process.

jeffkl (dismissed):

jeffkl:
@Erarndt some tests are failing with this change https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=9173771&view=ms.vss-test-web.build-test-results-tab&runId=95788042&resultId=112702&paneView=debug

psychvc (commented):
Thank

jeffkl:
Team Review: Closing this PR for now

PR #5538, opened 2023-12-12
Skip NuGetAudit when project doesn't use any packages

### LINKED ISSUE #13073: NuGetAudit should not download vulnerabilities database when project does not use any packages
### NuGet Product(s) Affected

### Current Behavior

As the title says, NuGetAudit download the vulnerabilities database, even when there are no packages to check.

The "easiest" way to validate is to open a CLI shell, set the environment variable `http_proxy` to `http://localhost:12345/` (make sure there isn't a real HTTP proxy running on this port), then do `dotnet new console`. Note, if you ran a restore more recently than 30 minutes ago, you might need to run `dotnet nuget locals http-cache --clear`, to ensure that the cached vulnerabilities isn't used.

### Desired Behavior

The scenario described above should not fail. In other words, when there are no packages, NuGetAudit should "skip" itself, and avoid making HTTP requests.

### Additional Context

_No response_

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13073

Regression? no

## Description

Early exit out NuGet Audit (before downloading the vulnerability database) when the project doesn't use any packages.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
@@ -65,6 +65,12 @@ namespace NuGet.Commands.Restore.Utility
 
         public async Task CheckPackageVulnerabilitiesAsync(CancellationToken cancellationToken)
         {
+            // Performance: Early exit if restore graph does not contain any packages.
+            if (!HasPackages())
+            {
+                return;
+            }
+
             Stopwatch stopwatch = Stopwatch.StartNew();
             GetVulnerabilityInfoResult? allVulnerabilityData = await GetAllVulnerabilityDataAsync(cancellationToken);
             stopwatch.Stop();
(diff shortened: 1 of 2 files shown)


src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -86,6 +92,18 @@ public async Task CheckPackageVulnerabilitiesAsync(CancellationToken cancellatio
                 CheckPackageVulnerabilities(allVulnerabilityData.KnownVulnerabilities);
             }
 
+            bool HasPackages()
```
jeffkl, line 95: Do you think this will have bad performance for very large restores?  It kind of looks like it will return very quickly as soon as any package reference is found and small projects that don't reference any packages will return quickly as well.  Just wondering your opinion in case my instincts are off.

jeffkl (dismissed):

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
zivkan: If you're asking if this new method will cause new perf problems, no, I don't think so. The restore graph is already resolved, so this method is just looking up a property on a collection of items in memory. And as you mentioned, it early exists when a package is found (vs only project references).

If the project doesn't have any packages, only project references, it's doing the same work that AuditUtility already does elsewhere (when it actually checks each package for known vulnerabilities). If the project doesn't have any packages, it'll be faster, since it won't need to obtain the vulnerability database. If the project does use at least one package, then it is additional work, but it early exits when it can, and even if it has to check a whole lot of Project references before it finds a package reference, it's just stuff in memory already.

My gut feeling is that it shouldn't have a big impact, but only a performance trace will tell us how often this method (or AuditUtility more generally) appears on sampled CPU stacks.

nkolev92 (dismissed):

jeffkl (approved):

PR #5540, opened 2023-12-12
Enable central package management implicitly when Directory.Packages.props exists

### LINKED ISSUE #11834: [DCR]: Central Package Management - Respect .props file as a way to opt-in to the feature.
### NuGet Product(s) Affected

Other/NA

### Current Behavior

To opt-in to central package management, one must explicitly enable a MSBuild property today. This brings quite a bit of confusion to users given the original spec mentions that if a file exists, all projects will be opted-in to the feature.

https://github.com/NuGet/Home/wiki/Centrally-managing-NuGet-package-versions#opt-in-central-package-version-management

### Desired Behavior

If a .props file exists or is imported appropriately, projects should be automatically opted-in to central package management unless there is an explicit MSBuild property disabling the use of the feature.

### Additional Context

This is currently a papercut that brings confusion to everyday users. We have documented this for now, but would like to make this more intuitive in the future.

### Breaking Change
This would be considered a breaking change in the scenario where a user has a `Directory.Packages.props` for some reason and is not setting the `ManagePackageVersionsCentrally` property.  CPM would now be on and they would need to set `ManagePackageVersionsCentrally` to `false` explicitly if they don't want to use CPM.

```xml
<PropertyGroup>
  <ManagePackageVersionsCentrally>false</ManagePackageVersionsCentrally>
</PropertyGroup>
```

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Home/issues/11834

## Description

This change makes it so the presence of a `Directory.Packages.props` means that central package management is implicitly on.  I've had countless people ask over the last year why you have to have the file and explicitly set `ManagePackageVersionsCentrally` to `true`.  The logic now assumes if `Directory.Packages.props` exists then `ManagePackageVersionsCentrally` should be `true` unless the user has already set that property to a value, including if its set in `Directory.Packages.props`.  This will make it easier for customers to onboard.  

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Build.Tasks/NuGet.props
+++ b/src/NuGet.Core/NuGet.Build.Tasks/NuGet.props
@@ -9,33 +9,32 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
 Copyright (c) .NET Foundation. All rights reserved.
 ***********************************************************************************************
 -->
-
 <Project>
-
   <!--
-      Import 'Directory.Packages.props' which will contain centralized packages for all the projects and solutions under
-      the directory in which the file is present. This is similar to 'Directory.Build.props/targets' logic which is present
-      in the common props/targets which serve a similar purpose.
+    Determine the path to the 'Directory.Packages.props' file, if the user did not:
+      1. Set $(ManagePackageVersionsCentrally) to false
+      2. Set $(ImportDirectoryPackagesProps) to false
+      3. Already specify the path to a 'Directory.Packages.props' file via $(DirectoryPackagesPropsPath)
   -->
-
-  <PropertyGroup>
-    <ImportDirectoryPackagesProps Condition="'$(ImportDirectoryPackagesProps)' == ''">true</ImportDirectoryPackagesProps>
-  </PropertyGroup>
-
-  <!--
-      Determine the path to the 'Directory.Packages.props' file, if the user did not disable $(ImportDirectoryPackagesProps) and
-      they did not already specify an absolute path to use via $(DirectoryPackagesPropsPath)
-  -->
-  <PropertyGroup Condition="'$(ImportDirectoryPackagesProps)' == 'true' and '$(DirectoryPackagesPropsPath)' == ''">
+  <PropertyGroup Condition="'$(ManagePackageVersionsCentrally)' != 'false' And '$(ImportDirectoryPackagesProps)' != 'false' And '$(DirectoryPackagesPropsPath)' == ''">
     <_DirectoryPackagesPropsFile Condition="'$(_DirectoryPackagesPropsFile)' == ''">Directory.Packages.props</_DirectoryPackagesPropsFile>
     <_DirectoryPackagesPropsBasePath Condition="'$(_DirectoryPackagesPropsBasePath)' == ''">$([MSBuild]::GetDirectoryNameOfFileAbove('$(MSBuildProjectDirectory)', '$(_DirectoryPackagesPropsFile)'))</_DirectoryPackagesPropsBasePath>
     <DirectoryPackagesPropsPath Condition="'$(_DirectoryPackagesPropsBasePath)' != '' and '$(_DirectoryPackagesPropsFile)' != ''">$([MSBuild]::NormalizePath('$(_DirectoryPackagesPropsBasePath)', '$(_DirectoryPackagesPropsFile)'))</DirectoryPackagesPropsPath>
   </PropertyGroup>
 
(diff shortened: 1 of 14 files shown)


nkolev92 (approved):

src/NuGet.Core/NuGet.Build.Tasks/NuGet.props
```diff
@@ -9,33 +9,32 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
...
-      in the common props/targets which serve a similar purpose.
+    Determine the path to the 'Directory.Packages.props' file, if the user did not:
+      1. Set $(ManagePackageVersionsCentrally) to false
+      2. Set $(ImportDirectoryPackagesProps) to false
+      3. Already specify the path to a 'Directory.Packages.props' file via $(DirectoryPackagesPropsPath)
   -->
-
-  <PropertyGroup>
-    <ImportDirectoryPackagesProps Condition="'$(ImportDirectoryPackagesProps)' == ''">true</ImportDirectoryPackagesProps>
+  <PropertyGroup Condition="'$(ManagePackageVersionsCentrally)' != 'false' And '$(ImportDirectoryPackagesProps)' != 'false' And '$(DirectoryPackagesPropsPath)' == ''">
```
kartheekp-ms, line 19: I found it difficult to understand the XML syntax while reviewing this pull request, so I have pasted the formatted XML here in case it helps to quickly grasp the changes.

```xml
<PropertyGroup Condition="'$(ManagePackageVersionsCentrally)' != 'false' And '$(ImportDirectoryPackagesProps)' != 'false' And '$(DirectoryPackagesPropsPath)' == ''">
  <_DirectoryPackagesPropsFile Condition="'$(_DirectoryPackagesPropsFile)' == ''">
    Directory.Packages.props
  </_DirectoryPackagesPropsFile>
  <_DirectoryPackagesPropsBasePath Condition="'$(_DirectoryPackagesPropsBasePath)' == ''">
    $([MSBuild]::GetDirectoryNameOfFileAbove('$(MSBuildProjectDirectory)', '$(_DirectoryPackagesPropsFile)'))
  </_DirectoryPackagesPropsBasePath>
  <DirectoryPackagesPropsPath Condition="'$(_DirectoryPackagesPropsBasePath)' != '' and '$(_DirectoryPackagesPropsFile)' != ''">
    $([MSBuild]::NormalizePath('$(_DirectoryPackagesPropsBasePath)', '$(_DirectoryPackagesPropsFile)'))
  </DirectoryPackagesPropsPath>
</PropertyGroup>

```

src/NuGet.Core/NuGet.Build.Tasks/NuGet.props
```diff
@@ -9,33 +9,32 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
...
+  <PropertyGroup Condition="'$(ManagePackageVersionsCentrally)' != 'false' And '$(ImportDirectoryPackagesProps)' != 'false' And Exists('$(DirectoryPackagesPropsPath)')">
+    <ManagePackageVersionsCentrally Condition="'$(ManagePackageVersionsCentrally)' == ''">true</ManagePackageVersionsCentrally>
   </PropertyGroup>
 
-  <Import Project="$(DirectoryPackagesPropsPath)" Condition="'$(ImportDirectoryPackagesProps)' == 'true' and '$(DirectoryPackagesPropsPath)' != '' and Exists('$(DirectoryPackagesPropsPath)')"/>
+  <Import Project="$(DirectoryPackagesPropsPath)" Condition="'$(ManagePackageVersionsCentrally)' != 'false' And '$(ImportDirectoryPackagesProps)' != 'false' And Exists('$(DirectoryPackagesPropsPath)')" />
 
-  <PropertyGroup Condition="'$(ImportDirectoryPackagesProps)' == 'true' and '$(DirectoryPackagesPropsPath)' != '' and Exists('$(DirectoryPackagesPropsPath)')">
+  <PropertyGroup Condition="'$(ManagePackageVersionsCentrally)' != 'false' And '$(ImportDirectoryPackagesProps)' != 'false' And Exists('$(DirectoryPackagesPropsPath)')">
```
kartheekp-ms, line 37: How about just checking `<PropertyGroup Condition="'$(ManagePackageVersionsCentrally)' == 'true' >` because ManagePackageVersionsCentrally is only `true` only when `'$(ImportDirectoryPackagesProps)' != 'false' And Exists('$(DirectoryPackagesPropsPath)')`?

src/NuGet.Core/NuGet.Build.Tasks/NuGet.props
zivkan: @kartheekp-ms if you have github configured to use "unified" diffs, consider changing to "split":
image

zivkan (approved):

PR #5541, opened 2023-12-13
System.Text.Json Migration - Adding code to parse the PackageSpec using STJ

jgonz120:
## Bug

## Description
Second PR in this series, this one focuses on the PackageSpec, which is required to complete the LockFile conversion. 

### FileFormatException
Updated this file to allow us to create this exception without having to provide the line or column where the error occurred. Unfortunately, we do not have access to this information, it's an internal member of the JsonReaderState

### JsonPackageSpecReader
Has been split into two different files, one for NewtonsoftJson and another for SystemTextJson 

The logic for both of these version's is similar since we were iterating through the JSON using the JsonTextReader. 

## PR Checklist

- [X] PR has a linked issue.
- [X] Described changes

- **Tests**
  - [X] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [X] N/A


src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamPackageSpecReader.cs
```diff
@@ -0,0 +1,1873 @@
...
+using NuGet.LibraryModel;
+using NuGet.Packaging.Core;
+using NuGet.RuntimeModel;
+using NuGet.Versioning;
+
+namespace NuGet.ProjectModel
+{
+    internal static class JsonStreamPackageSpecReaderPropertyName
+    {
+        private static readonly char[] VersionSeparators = new[] { ';' };
```
jgonz120: > I think using PropertyName suffix, rather than Utf8 prefix, will make this easier to understand. Down in the Read method, it will be if (reader.ValueTextEquals(LevelPropertyName)) rather than if (reader.ValueTextEquals(Utf8Level))
> 
> I can't remember where I first saw this pattern, but another option is to make a private static class called PropertyNames. For example:
> 
> internal class UtfJsonStreamAssetsLogMessageConverter : ...
> {
>   public override AssetsLogMessage Read(ref Utf8JsonStreamReader reader)
>   { 
>     // stuff
>     if (reader.ValueTextEquals(PropertyName.Level)
>    // other stuff
>   }
> 
>   private static class PropertyName
>   {
>     public static readonly byte[] Level = Encoding.UTF8.GetBytes(LogMessageProperties.LEVEL);
>     // other properties
>   }
> }
- @zivkan https://github.com/NuGet/NuGet.Client/pull/5529#discussion_r1415105618

src/NuGet.Core/NuGet.ProjectModel/NjPackageSpecReader.cs
```diff
@@ -0,0 +1,1765 @@
...
+using NuGet.Configuration;
+using NuGet.Frameworks;
+using NuGet.LibraryModel;
+using NuGet.Packaging.Core;
+using NuGet.RuntimeModel;
+using NuGet.Versioning;
+
+namespace NuGet.ProjectModel
+{
+    [Obsolete]
```
nkolev92: Why not deprecate JsonPackageSpecReader directly?

jgonz120: One of the functions in that class uses the STJ implementation so we shouldn't be deprecating that right?

nkolev92: It's not that we don't have to deprecate it, but we just shouldn't use it. 

My thinking was if we leavel JsonPackageSpecReader the footprint of this PR is a lot smaller. 

Maybe make it a partial class that has all the methods that split between old & new implementation, but right now it's just a lot of changes.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamPackageSpecReader.cs
```diff
@@ -0,0 +1,1873 @@
...
+
+                        if (jsonReader.TokenType == JsonTokenType.String)
+                        {
+                            dependencyVersionValue = jsonReader.GetString();
+                        }
+                        else if (jsonReader.TokenType == JsonTokenType.StartObject)
+                        {
+                            while (jsonReader.Read() && jsonReader.TokenType == JsonTokenType.PropertyName)
+                            {
+                                IEnumerable<string> values = null;
```
nkolev92: you probably don't need to declare this variable. You just assign and use it within the condition.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamPackageSpecReader.cs
```diff
@@ -0,0 +1,1873 @@
...
```
nkolev92: Same as earlier.

nkolev92: There's a bunch of cleanup that the editor suggests that could make these changes shorter: 
image
.

nkolev92 (commented):
Mostly cleanup comments, nothing major. 

I do have a suggestion for reducing the number of changes, thus making it easier to review.

src/NuGet.Core/NuGet.ProjectModel/NjPackageSpecReader.cs
zivkan: [JsonPackageSpecReader is used in a bunch of places](https://github.com/search?q=repo%3ANuGet%2FNuGet.Client+JsonPackageSpecReader&type=code), including nuget.exe and pack (nuget.commands), so it's not clear to me what the suggestion is for those calls if the entire JsonPackageSpecReader class is marked as obsolete.

I think Nikolche's suggestion is to leave the Newtonsoft.Json spec reader code inside `JsonPackageSpecReader.cs`, and just mark all the methods as obsolete, rather than moving them into a new file, where git's diff doesn't tell us if any of the method bodies have changed.

jgonz120: Oh okay, yea I thought about that. I put the NJ code back in the original file.

jgonz120: And switched to partial class

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,1801 @@
...
+        {
+            var packageSpec = new PackageSpec();
+
+            List<CompatibilityProfile> compatibilityProfiles = null;
+            List<RuntimeDescription> runtimeDescriptions = null;
+            var wasPackOptionsSet = false;
+            var isMappingsNull = false;
+            string filePath = name == null ? null : Path.GetFullPath(packageSpecPath);
+
+            if (jsonReader.TokenType == JsonTokenType.StartObject)
```
zivkan, line 140: It appears that if the file has a semantic error, for example, it's a start array, or just some literal value, this code silently ignores the error. I'm not sure what the Newtonsoft.Json code does, but typically deserializers will throw an exception.

There are so many tests, it's hard to find if such a test already exists. Do you know if there are error condition tests?

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,1801 @@
...
+                {
+                    var propertyName = jsonReader.GetString();
+                    if (string.IsNullOrEmpty(propertyName))
+                    {
+                        throw FileFormatException.Create(
+                            "Unable to resolve dependency ''.",
+                            packageSpecPath);
+                    }
+
+                    if (jsonReader.Read())
```
zivkan, line 295: There is no `else` for this statement. My understanding is that `Read` only returns false at the end of the file, in which case it means this JSON file is incomplete. Traditionally, parsers would throw an exception at the point of syntax error (like missing value following a property name), but this code appears to return an incomplete value.

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,1801 @@
...
+                        var suppressParentFlagsValue = LibraryIncludeFlagUtils.DefaultSuppressParent;
+                        string dependencyVersionValue = null;
+
+                        if (jsonReader.TokenType == JsonTokenType.String)
+                        {
+                            dependencyVersionValue = jsonReader.GetString();
+                        }
+                        else if (jsonReader.TokenType == JsonTokenType.StartObject)
+                        {
+                            while (jsonReader.Read() && jsonReader.TokenType == JsonTokenType.PropertyName)
```
zivkan: Maybe this code is a direct copy of the Newtonsoft.Json code, but I think it would be nice to reduce nesting within this fuction, and have different objects parsed in different functions. Since optimized code loses line numbers in exceptions, this helps investigate errors. Shorter functions are also easier to read & understand.

A [local function](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/local-functions) can be used, to keep this dependency parsing hidden outside the `ReadCentralTransitiveDependencyGroup` method, if that's desired.

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,1801 @@
...
+
+        private static void ReadDependencies(
+            ref Utf8JsonStreamReader jsonReader,
+            IList<LibraryDependency> results,
+            string packageSpecPath,
+            bool isGacOrFrameworkReference)
+        {
+            if (jsonReader.Read() && jsonReader.TokenType == JsonTokenType.StartObject)
+            {
+                while (jsonReader.Read() && jsonReader.TokenType == JsonTokenType.PropertyName)
```
zivkan, line 414: Like other comments, I'm not sure the error handling here provides a good API experience. If `jsonReader.Read()` returns false, doesn't that mean the file has an error?  If  read succeeds, but the token type is wrong, that's deinately an error, but no error is being reported here.

zivkan (commented):
This looks good for the "happy case". But looking at this from a library point of view, I have concerns about the error handling. However, since NuGet should be the only component that writes this file (we don't expect 3rd party tools to write the file, then NuGet to read it), maybe error handling is less important, since it should be so unlikely. 

There's always a question of "doing it right" vs "good enough", but also the principal of least surprise. I'm worried about how difficult it might be to diagnose some problems if customers report bugs, and we don't realize that the json parsing code is silently ignoring JSON errors, rather than throwing exceptions and making it abundantly clear that something unexpected happened.

zivkan:
Also, can the title of this PR please be updated to something meaningful?  It confuses me every time I see it in my notifications list 😄

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.Utf8JsonStreamReader.cs
jgonz120: There wasn't a test for this specific case, I just tried one and it fails silently in both versions of this code. I'm happy to add exceptions and stop failing silently but that would be a different behavior from what we already do.

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.Utf8JsonStreamReader.cs
jgonz120: I split it across multiple functions. I can't use local functions with Ref Structs unfortunately.

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.Utf8JsonStreamReader.cs
jgonz120: So you're right that read should only return false at the end of the file. That said Read() will throw an error if we're at the end of the file at this point. The file needs to be structured to be able to call read on the reader and be at a StartObject. I added a test to confirm this. 

Utf8JsonReader throws an exception if we try to read and hit the end of file when it's several layers deep into an object already.

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.Utf8JsonStreamReader.cs
zivkan: ok, I don't want to add scope creep, so even though I feel like this behaviour is wrong, given that assets and dgspec files are only expected to be written by nuget, which will always output a valid file, it's unlikely that a customer will have an invalid file being read.

We haven't had any customer complains in the years since the Newtonsoft.Json code was changed in this way, so consider it out of scope for this STJ work.

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.Utf8JsonStreamReader.cs
zivkan: I see. So from a practical point of view, `Read()` is never expected to return false, except for potentially the last read of the root object's deserialization (but from another PR, you found that Newtosoft.Json doesn't even validate that end of an object is also the end of the file).

This makes all the `if (jsonReader.Read())` conditions unnecessary nesting.

I'm wondering from an API design point of view, if `Utf8JsonStreamReader`'s `Read()` method should be changed to return `void`. We can add a `EnsureEndOfFile()` if we want to check that there is no contents at the end of the root object.

On the other hand, I feel guilty when my comments feel like scope creep, so this probably really doesn't matter, as `Utf8JsonStreamReader` is internal, and nuget should write the files that are read, so we have very high confidence there will not be any errors in the json.

zivkan (approved):
It appears all my feedback is relevant to the Newtonsoft.Json code, so probably out of scope for this PR.

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.cs
```diff
@@ -74,12 +69,33 @@ public static PackageSpec GetPackageSpec(JObject rawPackageSpec, string name, st
...
         internal static PackageSpec GetPackageSpec(JsonTextReader jsonReader, string packageSpecPath)
         {
             return GetPackageSpec(jsonReader, name: null, packageSpecPath, snapshotValue: null);
         }
 
-        private static PackageSpec GetPackageSpec(JsonTextReader jsonReader, string name, string packageSpecPath, string snapshotValue)
+        internal static PackageSpec GetPackageSpec(Stream stream, string name, string packageSpecPath, string snapshotValue, IEnvironmentVariableReader environmentVariableReader)
+        {
+            var useNj = environmentVariableReader.GetEnvironmentVariable("NUGET_EXPERIMENTAL_USE_NJ_FOR_FILE_PARSING");
+            if (string.IsNullOrEmpty(useNj) || useNj.Equals("false", StringComparison.OrdinalIgnoreCase))
```
kartheekp-ms, line 81:
```suggestion
            if (string.IsNullOrEmpty(useNj) || useNj.Equals(bool.FalseString, StringComparison.OrdinalIgnoreCase))
```

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.cs
```diff
@@ -292,6 +310,7 @@ private static void ReadBuildOptions(JsonTextReader jsonReader, PackageSpec pack
             });
         }
 
+        [Obsolete]
```
kartheekp-ms, line 313: An [obsolete attribute](https://learn.microsoft.com/dotnet/api/system.obsoleteattribute.-ctor?view=net-8.0#system-obsoleteattribute-ctor(system-string)) allows to add a short explanation suggesting alternative workarounds. It could be great if we can add that information here. Are these methods marked as obsolete due to their usage in the Newtonsoft.Json API? If that's the case, perhaps we could add a message to inform the caller. In certain instances, the caller might be a customer of the NuGet SDK.

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.Utf8JsonStreamReader.cs
```diff
@@ -0,0 +1,1822 @@
...
+
+            List<CompatibilityProfile> compatibilityProfiles = null;
+            List<RuntimeDescription> runtimeDescriptions = null;
+            var wasPackOptionsSet = false;
+            var isMappingsNull = false;
+            string filePath = name == null ? null : Path.GetFullPath(packageSpecPath);
+
+            if (jsonReader.TokenType == JsonTokenType.StartObject)
+            {
+                while (jsonReader.Read() && jsonReader.TokenType == JsonTokenType.PropertyName)
```
kartheekp-ms, line 142: I've noticed that the `GetPackageSpec` method contains several if-else blocks within this while loop, which might affect the readability and maintainability of the code. To improve this, I'm considering a refactor that involves using smaller, focused classes, each class is responsible for handling a specific property in the package spec. This way, the `GetPackageSpec` method can simply delegate the handling of each property to these specialized classes. One approach I'm thinking now is to combine the Strategy Pattern with the Abstract Factory Pattern. Implementing these patterns would significantly streamline the `GetPackageSpec` method, making it more concise and simpler to understand. Each property handler could be encapsulated in its own class, and an abstract factory could be used to instantiate the appropriate handler based on the property being processed. This would not only enhance readability but also make the codebase more modular and easier to extend or modify in the future.

```cs
interface IJsonPropertyHandler
{
    void HandleProperty(ref Utf8JsonStreamReader jsonReader, PackageSpec packageSpec);
}
```
```cs
class AuthorsPropertyHandler : IJsonPropertyHandler
{
    public void HandleProperty(ref Utf8JsonStreamReader jsonReader, PackageSpec packageSpec)
    {
        // Implement logic for handling Authors property
    }
}

class DescriptionPropertyHandler : IJsonPropertyHandler
{
    public void HandleProperty(ref Utf8JsonStreamReader jsonReader, PackageSpec packageSpec)
    {
        // Implement logic for handling Description property
    }

    // ... other implementations for different properties
}
```
```cs
interface IJsonPropertyHandlerFactory
{
    IJsonPropertyHandler GetHandler(string propertyName);
}
```
```cs
class JsonPropertyHandlerFactory : IJsonPropertyHandlerFactory
{
    private readonly Dictionary<string, IJsonPropertyHandler> _handlers;

    public JsonPropertyHandlerFactory()
    {
        _handlers = new Dictionary<string, IJsonPropertyHandler>
        {
            { "Authors", new AuthorsPropertyHandler() },
            { "Description", new DescriptionPropertyHandler() },
            // ... other mappings
        };
    }

    public IJsonPropertyHandler GetHandler(string propertyName)
    {
        if (_handlers.TryGetValue(propertyName, out var handler))
        {
            return handler;
        }

        return null; // Or some default handler
    }
}
```
```cs
internal static PackageSpec GetPackageSpec(ref Utf8JsonStreamReader jsonReader, string name, string packageSpecPath, string snapshotValue)
{
    var packageSpec = new PackageSpec();
    var handlerFactory = new JsonPropertyHandlerFactory();

    if (jsonReader.TokenType == JsonTokenType.StartObject)
    {
        while (jsonReader.Read() && jsonReader.TokenType == JsonTokenType.PropertyName)
        {
            string propertyName = jsonReader.GetString();
            var handler = handlerFactory.GetHandler(propertyName);

            if (handler != null)
            {
                handler.HandleProperty(ref jsonReader, packageSpec);
            }
            else
            {
                jsonReader.Skip(); // Skip unknown properties
            }
        }
    }

    // ... finalizing packageSpec
    return packageSpec;
}
```
```

kartheekp-ms (commented):
Again, I apologize for providing feedback after the pull request has been merged. Given that it was merged into a feature branch, I would like you to consider these comments in your future PRs during the migration process.

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.cs
jgonz120: This function is private so we shouldn't have any customers calling it right? The public functions have a message in them already but I can update it if you think a specific call out to STJ would be better.

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.Utf8JsonStreamReader.cs
jgonz120: Do you think this type of pattern is necessary when many of these properties are just reading a string from the stream?

PR #5544, opened 2023-12-14
Refactor VMs to listen to Settings being changed and notify the View to update

### LINKED ISSUE #13044: PackageSourceMappingActionViewModel should handle interactions with model

jebriede:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13044

## Description
This refactor puts the responsibility of listening for changes to Settings on the ViewModel objects that care about when the settings are updated and moves it away from the PackageManagerControl View which previously was listening to changes on Settings so it could instruct the ViewModels when to send property change notifications to the View. By removing the extra coordination from the PackageManagerControl, we end up with ViewModels that more closely follow the MVVM pattern, encapsulate their interaction with Settings, and whose dependencies are better expressed by the arguments they take in the constructor.

This change represents the implementation of PR feedback I provided here: https://github.com/NuGet/NuGet.Client/pull/5166#discussion_r1198459461 and created draft PR to demonstrate the requested change: https://github.com/NuGet/NuGet.Client/pull/5173. There are opportunities for further improvement, but this refactor is limited to moving the responsibility of listening for changes to Settings from the PackageManagerControl to the ViewModels that need to react when Settings change.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 
As a refactor, no additional tests were added, but I updated existing tests.

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
@@ -9,6 +9,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using System.Windows.Media.Imaging;
 using Microsoft.ServiceHub.Framework;
+using NuGet.Configuration;
 using NuGet.PackageManagement.UI.ViewModels;
 using NuGet.PackageManagement.VisualStudio;
 using NuGet.Packaging.Core;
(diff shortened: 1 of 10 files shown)


nkolev92 (approved):

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
```diff
@@ -49,19 +50,23 @@ public abstract class DetailControlModel : INotifyPropertyChanged, IDisposable
...
             IEnumerable<IProjectContextInfo> projects,
-            INuGetUI uiController)
+            INuGetUI uiController,
+            ISettings settings)
         {
             _nugetProjects = projects;
             ServiceBroker = serviceBroker;
             _uiController = uiController;
+            _settings = settings ?? throw new ArgumentNullException(nameof(settings));
+            _settings.SettingsChanged += Settings_SettingsChanged;
```
zivkan, line 66: events must always be unsubscribed, or there will be memory leaks.

The event (`_settings.SettingsChanged`) will hold a reference to this instance of `DetailControlModel`, and therefore it will not be garbage collected until `_settings` is GC'd.  However, `VsSettings` is a singleton, and therefore none of the `DetailControlModels` will be GC'd during the process lifetime. For this reason, the event unsubscribe cannot happen in a `Dispose()` method, unless something explicitly calls dispose on the instance, since the automatic dispose on GC will never happen.

zivkan (changes requested):

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageSourceMappingActionViewModel.cs
```diff
@@ -125,9 +130,16 @@ public void SettingsChanged()
             RaisePropertyChanged(nameof(CanAutomaticallyCreateSourceMapping));
         }
 
-        public static PackageSourceMappingActionViewModel Create(INuGetUI uiController)
+        public void Dispose()
         {
-            return new PackageSourceMappingActionViewModel(uiController);
+            _settings.SettingsChanged -= Settings_SettingsChanged;
```
zivkan, line 135: Can you confirm that this `Dispose` is explicitly called somewhere, and it's not relying on the GC calling dispose? As per my previous comment, `_settings.SettingsChanged` will hold a reference to this object, hence the GC will never dispose this object until after `_settings` releases the instance (the event is unsubscribed).

I had a look through the PR, but I don't see any new `.Dispose()` calls,

PR #5551, opened 2023-12-20
Use string.Contains instead of IEnumerable.Contains

### LINKED ISSUE #13124: Use string.Contains instead of IEnumerable.Contains in VersionRange parsing

Erarndt:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13124

## Description

This code was mistakenly using the `Enumerable.Contains()` overload which ends up allocating an enumerator. Using `string.Contains()` avoids this allocation and is faster. Example benchmark over 100 invocations:

| Method              | Count | Mean      | Error     | StdDev    | Gen0   | Allocated |
|-------------------- |------ |----------:|----------:|----------:|-------:|----------:|
| Enumerable.Contains | 100   | 29.830 us | 0.0658 us | 0.0583 us | 0.3662 |    2003 B |
| string.Contains     | 100   |  5.516 us | 0.0105 us | 0.0082 us |      - |         - |

## PR Checklist

- [ ] PR has a linked issue.
- [ ] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Versioning/VersionRangeFactory.cs
+++ b/src/NuGet.Core/NuGet.Versioning/VersionRangeFactory.cs
@@ -204,7 +204,7 @@ namespace NuGet.Versioning
 #if NETCOREAPP2_1_OR_GREATER
                 if (allowFloating && minVersionString.Contains('*', StringComparison.Ordinal))
 #else
-                if (allowFloating && minVersionString.Contains('*'))
+                if (allowFloating && minVersionString.Contains("*"))
 #endif
                 {
                     // single floating version


src/NuGet.Core/NuGet.Versioning/VersionRangeFactory.cs
```diff
@@ -204,7 +204,7 @@ public static bool TryParse(string value, bool allowFloating, [NotNullWhen(true)
 #if NETCOREAPP2_1_OR_GREATER
                 if (allowFloating && minVersionString.Contains('*', StringComparison.Ordinal))
```
nkolev92, line 205: Does this have the same problem?

nkolev92:
@Erarndt I think we'd probably want more granular issues here, since there's some changes that aren't related to what the main issue is claiming.

src/NuGet.Core/NuGet.Versioning/VersionRangeFactory.cs
Erarndt: Additional overloads for `string.Contains` were added in 2.1+, so this call doesn't have the same issue.

Erarndt:
> @Erarndt I think we'd probably want more granular issues here, since there's some changes that aren't related to what the main issue is claiming.

Sure, I defer to you folks on how to organize these. I've found these while investigating the main issue, so I was unsure how you'd like to categorize them.

zivkan (approved):

PR #5553, opened 2023-12-20
Avoid some allocations in GetCompatibilityData

Erarndt:
## Bug

## Description

Calls to GetCompatibilityData were boxing enumerators and creating closure objects. This change avoids these allocations.

## PR Checklist

- [ ] PR has a linked issue.
- [ ] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/CompatibilityChecker.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/CompatibilityChecker.cs
@@ -455,20 +455,35 @@ namespace NuGet.Commands
         {
             // Use data from the current lock file if it exists.
             LockFileTargetLibrary targetLibrary = null;
-            var target = _lockFile.Targets.FirstOrDefault(t => Equals(t.TargetFramework, graph.Framework) && string.Equals(t.RuntimeIdentifier, graph.RuntimeIdentifier, StringComparison.Ordinal));
-            if (target != null)
+
+            for (int i = 0; i < _lockFile.Targets.Count; ++i)
             {
-                targetLibrary = target.Libraries
-                    .FirstOrDefault(t => t.Name.Equals(libraryId.Name, StringComparison.OrdinalIgnoreCase) && t.Version.Equals(libraryId.Version));
+                var target = _lockFile.Targets[i];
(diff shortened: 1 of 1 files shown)


src/NuGet.Core/NuGet.Commands/RestoreCommand/CompatibilityChecker.cs
```diff
@@ -455,20 +455,35 @@ private CompatibilityData GetCompatibilityData(RestoreTargetGraph graph, Library
...
+            for (int i = 0; i < _lockFile.Targets.Count; ++i)
             {
-                targetLibrary = target.Libraries
-                    .FirstOrDefault(t => t.Name.Equals(libraryId.Name, StringComparison.OrdinalIgnoreCase) && t.Version.Equals(libraryId.Version));
+                var target = _lockFile.Targets[i];
+                if (Equals(target.TargetFramework, graph.Framework) && string.Equals(target.RuntimeIdentifier, graph.RuntimeIdentifier, StringComparison.Ordinal))
+                {
+                    for (int j = 0; j < target.Libraries.Count; ++j)
+                    {
+                        var library = target.Libraries[j];
```
zivkan, line 466: The old code would iterate `_lockFile.Targets` and stop when the TFM & RID match, then will look though `target.Libraries`. But if it couldn't find the patching package/project name in libraries, it wouldn't resume searching though `_lockFile.Targets`.

While NuGet's data model doesn't prevent it, there should never be a case of a TFM & RID combo being duplicated in `_lockFile.Targets`.

Therefore, this PR's change, where it will resume iterating the rest of the targets, is doing unnecessary CPU work. Having said that, even for MAUI projects, the `_lockFile.Targets` list isn't huge. But this PR is removing an early exit optimization.

Erarndt: I could be missing something, but I believe this maintains the early exit. The `break` on line 474 will break out of the outermost loop on the first TMF & RID match.

zivkan: oops, yes, I missed that.

zivkan (approved):

nkolev92 (approved):
Looks great! 
It'll be great to get a matching issue created.

PR #5555, opened 2023-12-20
Unroll Linq in GetFlags

Erarndt:
## Bug

## Description

Avoids some allocations due to LINQ usage. Also creates a new list that is sorted in place rather than using `OrderBy`, which internally allocates a buffer for all the items, and then calling `ToList()` which allocates a list again.

## PR Checklist

- [ ] PR has a linked issue.
- [ ] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Packaging/NuspecReader.cs
+++ b/src/NuGet.Core/NuGet.Packaging/NuspecReader.cs
@@ -641,13 +641,23 @@ namespace NuGet.Packaging
 
             // PERF: Avoid Linq on hot paths
             var splitFlags = flags.Split(CommaArray, StringSplitOptions.RemoveEmptyEntries);
+
+#if NETSTANDARD2_0
+            var set = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+#elif NET472_OR_GREATER || NET5_0_OR_GREATER
             var set = new HashSet<string>(splitFlags.Length, StringComparer.OrdinalIgnoreCase);
-            for (int i = 0; i < splitFlags.Length; ++i)
+#endif
+            foreach (string flag in splitFlags)
             {
-                set.Add(splitFlags[i].Trim());
+                set.Add(flag.Trim());
             }
 
-            var result = set.ToList();
+            var result = new List<string>(set.Count);
+            foreach (var s in set)
+            {
+                result.Add(s);
+            }
+
             result.Sort(StringComparer.OrdinalIgnoreCase);
 
             return result;
--- a/src/NuGet.Core/NuGet.Packaging/NuspecReader.cs
+++ b/src/NuGet.Core/NuGet.Packaging/NuspecReader.cs
@@ -639,12 +639,18 @@ namespace NuGet.Packaging
                 return EmptyList;
             }
 
-            var set = new HashSet<string>(
-                flags.Split(CommaArray, StringSplitOptions.RemoveEmptyEntries)
-                    .Select(flag => flag.Trim()),
-                StringComparer.OrdinalIgnoreCase);
+            // PERF: Avoid Linq on hot paths
+            var splitFlags = flags.Split(CommaArray, StringSplitOptions.RemoveEmptyEntries);
+            var set = new HashSet<string>(splitFlags.Length, StringComparer.OrdinalIgnoreCase);
+            for (int i = 0; i < splitFlags.Length; ++i)
+            {
+                set.Add(splitFlags[i].Trim());
+            }
 
-            return set.OrderBy(s => s, StringComparer.OrdinalIgnoreCase).ToList();
+            var result = set.ToList();
+            result.Sort(StringComparer.OrdinalIgnoreCase);
+
+            return result;
         }
 
         private HashSet<PackageDependency> GetPackageDependencies(IEnumerable<XElement> nodes, bool useStrictVersionCheck)


src/NuGet.Core/NuGet.Packaging/NuspecReader.cs
```diff
@@ -639,12 +639,18 @@ private static List<string> GetFlags(string flags)
...
             }
 
-            var set = new HashSet<string>(
-                flags.Split(CommaArray, StringSplitOptions.RemoveEmptyEntries)
-                    .Select(flag => flag.Trim()),
-                StringComparer.OrdinalIgnoreCase);
+            // PERF: Avoid Linq on hot paths
+            var splitFlags = flags.Split(CommaArray, StringSplitOptions.RemoveEmptyEntries);
+            var set = new HashSet<string>(splitFlags.Length, StringComparer.OrdinalIgnoreCase);
+            for (int i = 0; i < splitFlags.Length; ++i)
```
davkean: This can just be a foreach as C# will unroll under the covers to this.

src/NuGet.Core/NuGet.Packaging/NuspecReader.cs
```diff
@@ -639,12 +639,18 @@ private static List<string> GetFlags(string flags)
...
+            {
+                set.Add(splitFlags[i].Trim());
+            }
 
-            return set.OrderBy(s => s, StringComparer.OrdinalIgnoreCase).ToList();
+            var result = set.ToList();
```
zivkan:
```suggestion
            var result = new List<string>(set.Count);
            result.AddRange(set);
```

If we're trying to micro-optimize allocations, we should supply capacity arguments to collection constructors, especially in easy scenarios like this one: https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/23126/Performance-Best-Practices?anchor=minimizing-memory-allocation

It might not have a huge impact on performance, because the vast majority of packages have fewer than 5 include/exclude flags, but a quick query of [NuGet Insights](https://github.com/NuGet/Insights) it appears there are 18,000 packages (fewer unique package ids, but then multiple versions of the ID are likely to have the same flags) with 5 or more flags. The package with the highest downloads isMicrosoft.CodeAnalysis.Workspaces.Common, and if you [check the nuspec](https://nuget.info/packages/Microsoft.CodeAnalysis.Workspaces.Common/4.8.0), there are 6 items in the list. So we can avoid 1 backing array resize, per dependency with more than 4 flags, by pre-allocating the correct capacity.

Erarndt: Using `ToList` on a type that is an `ICollection<T>` doesn't have the deficiency that you describe. Here's the code that's called:

    public static List<TSource> ToList<TSource>(this IEnumerable<TSource> source)
    {
        if (source == null)
        {
            throw Error.ArgumentNull("source");
        }

        return new List<TSource>(source);
    }


    public List(IEnumerable<T> collection)
    {
        if (collection == null)
        {
            ThrowHelper.ThrowArgumentNullException(ExceptionArgument.collection);
        }

        if (collection is ICollection<T> collection2)
        {
            int count = collection2.Count;
            if (count == 0)
            {
                _items = _emptyArray;
                return;
            }

            _items = new T[count];
            collection2.CopyTo(_items, 0);
            _size = count;
            return;
        }

        _size = 0;
        _items = _emptyArray;
        foreach (T item in collection)
        {
            Add(item);
        }
    }

Using `ToList` basically just calls the `List<T>` constructor which internally does a cast to see if it can determine the proper initialization size. The best practices you link to are valid and are accounted for in this case.

A hidden cost of the suggested change is that calling `AddRange` has some unexpected allocations:

    public void AddRange(IEnumerable<T> collection)
    {
        InsertRange(_size, collection);
    }

    public void InsertRange(int index, IEnumerable<T> collection)
    {
        if (collection == null)
        {
            ThrowHelper.ThrowArgumentNullException(ExceptionArgument.collection);
        }

        if ((uint)index > (uint)_size)
        {
            ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.index, ExceptionResource.ArgumentOutOfRange_Index);
        }

        if (collection is ICollection<T> collection2)
        {
            int count = collection2.Count;
            if (count > 0)
            {
                EnsureCapacity(_size + count);
                if (index < _size)
                {
                    Array.Copy(_items, index, _items, index + count, _size - index);
                }

                if (this == collection2)
                {
                    Array.Copy(_items, 0, _items, index, index);
                    Array.Copy(_items, index + count, _items, index * 2, _size - index);
                }
                else
                {
                    T[] array = new T[count];
                    collection2.CopyTo(array, 0);
                    array.CopyTo(_items, index);
                }

                _size += count;
            }
        }
        else
        {
            using IEnumerator<T> enumerator = collection.GetEnumerator();
            while (enumerator.MoveNext())
            {
                Insert(index++, enumerator.Current);
            }
        }

        _version++;
    }

In the case where we're passing in an `ICollection<T>` (our case), it'll ensure we have sufficient capacity (we would since we gave it enough in the constructor), but then allocates an intermediate array in the `else` block. This is a defensive step to ensure that some arbitrary implementation of `ICollection<T>.CopyTo` doesn't corrupt `List<T>` (https://stackoverflow.com/a/2123661). The code looks similar to what's used in the `List<T>` constructor, but the `CopyTo` in the constructor only copies to the final internal array. For our specific instance, the most efficient implementation would likely be this:

    var result = new List<string>(set.Count);
    foreach (var s in set)
    {
        result.Add(s);
    }

The only real advantage over `ToList` is that this avoids the cast to `ICollection<T>`. Note that this being true depends on how `CopyTo` is implemented. The `HashSet<T>` implementation enumerates the buckets and updates the array one at a time

    public void CopyTo(T[] array, int arrayIndex, int count)
    {
        if (array == null)
        {
            throw new ArgumentNullException("array");
        }

        if (arrayIndex < 0)
        {
            throw new ArgumentOutOfRangeException("arrayIndex", System.SR.GetString("ArgumentOutOfRange_NeedNonNegNum"));
        }

        if (count < 0)
        {
            throw new ArgumentOutOfRangeException("count", System.SR.GetString("ArgumentOutOfRange_NeedNonNegNum"));
        }

        if (arrayIndex > array.Length || count > array.Length - arrayIndex)
        {
            throw new ArgumentException(System.SR.GetString("Arg_ArrayPlusOffTooSmall"));
        }

        int num = 0;
        for (int i = 0; i < m_lastIndex; i++)
        {
            if (num >= count)
            {
                break;
            }

            if (m_slots[i].hashCode >= 0)
            {
                array[arrayIndex + num] = m_slots[i].value;
                num++;
            }
        }
    }

However, `List<T>.CopyTo()` defers to `Array.Copy` and would outperform us doing manual enumeration.

    public void CopyTo(T[] array, int arrayIndex)
    {
        Array.Copy(_items, 0, array, arrayIndex, _size);
    }

I think using `ToList()` or calling the `List<T>` constructor directly leads to less brittle code in general, but I could see an argument for the "streamlined" version to avoid the cast if we expect it to matter. Thoughts?

zivkan: I didn't know that `ToList` has this perf optimization. The times I debugged were probably in scenarios where `Where` or `Select` LINQ expressions were used, hence it can't cast to `ICollection` in those cases. It's great that LINQ has at least some simple perf optimizations.

zivkan (dismissed):

src/NuGet.Core/NuGet.Packaging/NuspecReader.cs
nkolev92: There are compile errors, so maybe this can be addressed in the same push @Erarndt

nkolev92 (commented):
There's compile errors. 

But the general change looks good to me. 

Please create a matching issue for these changes.

src/NuGet.Core/NuGet.Packaging/NuspecReader.cs
Erarndt: Updated

nkolev92 (approved):

jeffkl (approved):

jeffkl (approved):

Saibamen:
@Erarndt: Do you have benchmark results?

PR #5556, opened 2023-12-20
Switch from ConcurrentDictionary to Dictionary with lock to avoid rep…

Erarndt:
…eated closure allocations.

## Bug

## Description

Switches from usage of `ConcurrentDictionary` to `Dictionary` used within a lock. Empirically, it appears that this lock is unlikely to suffer from significant contention, and the allocation overhead from using `ConcurrentDictionary` ends up being significant.

## PR Checklist

- [ ] PR has a linked issue.
- [ ] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/SourceRepositoryDependencyProvider.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/SourceRepositoryDependencyProvider.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


src/NuGet.Core/NuGet.Commands/RestoreCommand/SourceRepositoryDependencyProvider.cs
```diff
@@ -237,6 +223,32 @@ internal static SemaphoreSlim GetThrottleSemaphoreSlim(IEnvironmentVariableReade
...
+            Task<LibraryIdentity> FindLibraryIdentityAsync(LibraryRange libraryRange, SourceCacheContext cacheContext)
+            {
+                Task<LibraryIdentity> result;
+                if (cacheContext.RefreshMemoryCache)
+                {
+                    lock (_libraryMatchCache)
+                    {
+                        result = FindLibraryCoreAsync(libraryRange, cacheContext, logger, cancellationToken);
+                        _libraryMatchCache[libraryRange] = result;
+                    }
```
zivkan: I think the `FindLibraryCoreAsync` call should be wrapped in a `Task.Run`, to ensure that it always yields quickly. While my own testing shows that for HTTP sources there's no perf impact to this PR (time differences are within run-to-run variance), the same is not true for file share package sources, where there is no async API to open files, or enumerate directories. Additionally, i'm not sure we use `FileStream.CopyAsync` either. As a consequence, all restores that need to use these dictionaries end up locking it for a long time, until the entire `FindLibraryCoreAsync` method completes, rather than yielding at first IO.

I tested with OrchardCore, doing a `git clean -xdfq` and `dotnet nuget locals all --clear` between ever restore. It's not a perfect comparison, because there are many other commits in the branch that aren't in the GA build I'm comparing against, but I'm reasonably confident that these perf regressions are due to this PR, not anything else already merged, but not yet shipped.

|Package source|GA restore|restore with this PR|diff
|--|--|--|--|
|HTTP|15s|17s|close enough|
|local directory|14s|14s|same|
|low latency file share|17s|24s|+41%|
|high latency file share (1)|16m 16s|30m 32s|+88%|

(1) "only" 160ms, but it goes to show how latency sensitive SMB is. Also, given the long duration, I only ran these tests once each. The other tests I ran multiple times, to reduce impact from operating system filesystem caches, MSBuild "servers" (node reuse), etc.

I don't know what percentage of our customers use HDDs, rather than SSD either. Before NuGet makes any HTTP request, it checks a filesystem based http cache, and uses the contents on disk if the file in http is less than 30 minutes old. Therefore, even with HTTP sources, there are scenarios where calls to `FileLibraryCoreAsync` will block on disk IO.

Erarndt: Good call on that. We want to just kick off the async work and not hold the lock in potential synchronous paths. I experimented with `Task.Yield()` inside of `FindLibraryCoreAsync`, but benchmarks indicate that it performs worse.

zivkan (approved):

nkolev92 (approved):

PR #5558, opened 2023-12-22
System.Text.Json Migration - Adding code to parse the Project.Assets.Json file using STJ. 

### LINKED ISSUE #12715: Stop using JObject in assets file reading to reduce allocations

jgonz120:
## Bug 

## Description
This PR update the LockFileFormat class to use STJ for parsing the assets file. There are some comments that were pointed out from https://github.com/NuGet/NuGet.Client/pull/5530 which I think should be done as a separate PR if we decide to move forward with them, so there is potentially one more PR to come. 

### Benchmark Results
| Method                          | Runtime              | InputFile    | Mean         | StdDev       | Ratio | Gen0       | Gen1      | Gen2      | Allocated    | Alloc Ratio |
|-------------------------------- |--------------------- |------------- |-------------:|-------------:|------:|-----------:|----------:|----------:|-------------:|------------:|
| 'LockFileFormat read StreamSTJ' | .NET 8.0             | 10KB.json    |   1,096.2 us |     15.93 us |  0.88 |     1.9531 |         - |         - |     57.64 KB |        0.29 |
| 'LockFileFormat read NJ'        | .NET 8.0             | 10KB.json    |   1,239.3 us |     17.15 us |  1.00 |    11.7188 |    3.9063 |         - |     199.3 KB |        1.00 |
|                                 |                      |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET Framework 4.7.2 | 10KB.json    |     380.8 us |      6.29 us |  0.65 |    10.2539 |    0.9766 |         - |     65.13 KB |        0.30 |
| 'LockFileFormat read NJ'        | .NET Framework 4.7.2 | 10KB.json    |     590.1 us |     11.59 us |  1.00 |    35.1563 |    9.7656 |         - |    218.74 KB |        1.00 |
|                                 |                      |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET 8.0             | 11527KB.json | 172,547.8 us |  6,243.82 us |  0.43 |  4500.0000 | 4000.0000 | 1000.0000 |  63580.02 KB |        0.47 |
| 'LockFileFormat read NJ'        | .NET 8.0             | 11527KB.json | 397,688.1 us |  8,228.42 us |  1.00 | 10000.0000 | 9000.0000 | 2000.0000 | 136010.63 KB |        1.00 |
|                                 |                      |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET Framework 4.7.2 | 11527KB.json | 293,957.7 us |  7,402.11 us |  0.64 | 11500.0000 | 4500.0000 | 1500.0000 |   66057.4 KB |        0.47 |
| 'LockFileFormat read NJ'        | .NET Framework 4.7.2 | 11527KB.json | 462,337.6 us | 10,347.79 us |  1.00 | 24000.0000 | 9000.0000 | 2000.0000 |  140931.8 KB |        1.00 |
|                                 |                      |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET 8.0             | 1308KB.json  |   6,410.1 us |     69.73 us |  0.39 |   273.4375 |  257.8125 |         - |   4473.46 KB |        0.47 |
| 'LockFileFormat read NJ'        | .NET 8.0             | 1308KB.json  |  16,530.2 us |    292.99 us |  1.00 |   562.5000 |  531.2500 |         - |   9534.15 KB |        1.00 |
|                                 |                      |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET Framework 4.7.2 | 1308KB.json  |  12,922.8 us |    147.81 us |  0.46 |   765.6250 |  375.0000 |         - |   4769.82 KB |        0.47 |
| 'LockFileFormat read NJ'        | .NET Framework 4.7.2 | 1308KB.json  |  28,322.2 us |    152.84 us |  1.00 |  1750.0000 |  781.2500 |  218.7500 |  10053.53 KB |        1.00 |
|                                 |                      |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET 8.0             | 2756KB.json  |  18,265.2 us |    126.52 us |  0.29 |   718.7500 |  656.2500 |         - |  11935.02 KB |        0.44 |
| 'LockFileFormat read NJ'        | .NET 8.0             | 2756KB.json  |  66,284.9 us |  3,021.89 us |  1.00 |  2000.0000 | 1666.6667 |  333.3333 |   27390.5 KB |        1.00 |
|                                 |                      |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET Framework 4.7.2 | 2756KB.json  |  46,018.3 us |    729.88 us |  0.50 |  2250.0000 |  916.6667 |  333.3333 |  12489.87 KB |        0.44 |
| 'LockFileFormat read NJ'        | .NET Framework 4.7.2 | 2756KB.json  |  92,955.5 us |  1,347.34 us |  1.00 |  5000.0000 | 2166.6667 |  833.3333 |  28609.74 KB |        1.00 |
|                                 |                      |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET 8.0             | 786KB.json   |   4,713.7 us |     30.15 us |  0.42 |   179.6875 |  164.0625 |         - |   2989.07 KB |        0.41 |
| 'LockFileFormat read NJ'        | .NET 8.0             | 786KB.json   |  11,352.3 us |     86.55 us |  1.00 |   437.5000 |  406.2500 |         - |   7343.26 KB |        1.00 |
|                                 |                      |              |              |              |       |            |           |           |              |             |
| 'LockFileFormat read StreamSTJ' | .NET Framework 4.7.2 | 786KB.json   |   8,629.2 us |     32.36 us |  0.45 |   500.0000 |  250.0000 |         - |   3156.93 KB |        0.41 |
| 'LockFileFormat read NJ'        | .NET Framework 4.7.2 | 786KB.json   |  18,987.8 us |    365.01 us |  1.00 |  1250.0000 |  625.0000 |   93.7500 |   7696.09 KB |        1.00 |
## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Core/NuGet.ProjectModel/LazyStringSplit.cs
```diff
@@ -0,0 +1,138 @@
...
+    ///     <see cref="StringSplitOptions.RemoveEmptyEntries"/>.
+    /// </summary>
+    /// <remarks>
+    ///     Unlike <see cref="string.Split(char[])"/> and overloads, <see cref="LazyStringSplit"/>
+    ///     does not allocate an array for the return, and allocates strings on demand during
+    ///     enumeration. A custom enumerator type is used so that the only allocations made are
+    ///     the substrings themselves. We also avoid the large internal arrays assigned by the
+    ///     methods on <see cref="string"/>.
+    /// </remarks>
+    internal readonly struct LazyStringSplit : IEnumerable<string>
```
nkolev92, line 24: Did we copy this from somewhere?

Can we add some tests for it?

src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileFormat.cs
```diff
@@ -88,21 +88,51 @@ public LockFile Read(Stream stream, string path)
 
         public LockFile Read(Stream stream, ILogger log, string path)
         {
-            using (var textReader = new StreamReader(stream))
+            return Read(stream, log, path, EnvironmentVariableWrapper.Instance);
+        }
+
+        internal LockFile Read(string filePath, IEnvironmentVariableReader environmentVariableReader)
```
nkolev92: Where do we use this?

We don't need to match every method from the previous implementation.

src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileFormat.cs
```diff
@@ -88,21 +88,51 @@ public LockFile Read(Stream stream, string path)
...
             {
-#pragma warning disable CS0612 // Type or member is obsolete
-                return Read(textReader, log, path);
-#pragma warning restore CS0612 // Type or member is obsolete
+                return Read(stream, NullLogger.Instance, filePath, environmentVariableReader);
             }
         }
 
-        [Obsolete]
+        internal LockFile Parse(string lockFileContent, string path, IEnvironmentVariableReader environmentVariableReader)
```
nkolev92: same here, where is this used?

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamAssetsLogMessageConverter.cs
```diff
@@ -0,0 +1,131 @@
...
+            int startLineNumber = default;
+            int startColNumber = default;
+            int endLineNumber = default;
+            int endColNumber = default;
+            string libraryId = default;
+            IReadOnlyList<string> targetGraphs = null;
+
+            while (reader.Read() && reader.TokenType == JsonTokenType.PropertyName)
+            {
+                if (!isValid)
```
nkolev92: I think we should finish reading the object here. 
The previous implementation would ignore unknown warnings and read the rest. This implementation would effectively fail since you don't really know  what reader.skip() does

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileConverter.cs
```diff
@@ -0,0 +1,120 @@
...
+                    reader.Read();
+                    lockFile.PackageFolders = reader.ReadObjectAsList<LockFileItem>(Utf8JsonReaderExtensions.LockFileItemConverter);
+                }
+                else if (reader.ValueTextEquals(ProjectPropertyName))
+                {
+                    reader.Read();
+                    lockFile.PackageSpec = JsonPackageSpecReader.GetPackageSpec(
+                        ref reader,
+                        name: null,
+                        packageSpecPath: null,
```
nkolev92, line 87: Do we need to passing down the name and the path here?

How is the name/path used? Is it used for raising better error messags?

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetLibraryConverter.cs
```diff
@@ -0,0 +1,203 @@
...
+        private static readonly byte[] RuntimeTargetsPropertyName = Encoding.UTF8.GetBytes("runtimeTargets");
+        private static readonly byte[] ToolsPropertyName = Encoding.UTF8.GetBytes("tools");
+        private static readonly byte[] EmbedPropertyName = Encoding.UTF8.GetBytes("embed");
+        private static readonly byte[] FrameworkReferencesPropertyName = Encoding.UTF8.GetBytes("frameworkReferences");
+
+        private IList<PackageDependency> ReadPackageDependencyList(ref Utf8JsonStreamReader reader)
+        {
+            if (reader.TokenType != JsonTokenType.StartObject)
+            {
+                return new List<PackageDependency>(0);
```
nkolev92:
```suggestion
                return new Array.Empty<PackageDependency>();
```

No allocation for that. 

Applies to any other places you may have this.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetLibraryConverter.cs
```diff
@@ -0,0 +1,203 @@
...
+        private static readonly byte[] NativePropertyName = Encoding.UTF8.GetBytes("native");
+        private static readonly byte[] BuildPropertyName = Encoding.UTF8.GetBytes("build");
+        private static readonly byte[] BuildMultiTargetingPropertyName = Encoding.UTF8.GetBytes("buildMultiTargeting");
+        private static readonly byte[] ContentFilesPropertyName = Encoding.UTF8.GetBytes("contentFiles");
+        private static readonly byte[] RuntimeTargetsPropertyName = Encoding.UTF8.GetBytes("runtimeTargets");
+        private static readonly byte[] ToolsPropertyName = Encoding.UTF8.GetBytes("tools");
+        private static readonly byte[] EmbedPropertyName = Encoding.UTF8.GetBytes("embed");
+        private static readonly byte[] FrameworkReferencesPropertyName = Encoding.UTF8.GetBytes("frameworkReferences");
+
+        private IList<PackageDependency> ReadPackageDependencyList(ref Utf8JsonStreamReader reader)
```
nkolev92: fwiw, for ordering purposes, it'd be good to have the "public" (I know the class is internal) and the really main method on top.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetLibraryConverter.cs
```diff
@@ -0,0 +1,203 @@
...
+                    reader.Read();
+                    if (reader.ReadObjectAsList<LockFileItem>(Utf8JsonReaderExtensions.LockFileItemConverter) is { Count: not 0 } nativeLibraries)
+                    {
+                        lockFileTargetLibrary.NativeLibraries = nativeLibraries;
+                    }
+                }
+                else if (reader.ValueTextEquals(BuildPropertyName))
+                {
+                    reader.Read();
+                    if (reader.ReadObjectAsList<LockFileItem>(Utf8JsonReaderExtensions.LockFileItemConverter) is { Count: not 0 } build)
```
nkolev92: Applies to all of the collections, are we better off setting Array.Empty instead of leaving it unset?
What happens if we leave it unset? Any potential issues with Array.Empty?

src/NuGet.Core/NuGet.ProjectModel/LazyStringSplit.cs
jgonz120: Original Comment: https://github.com/NuGet/NuGet.Client/pull/5529#discussion_r1414973751
Wiki Entry: https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/24181/Garbage-Collection?anchor=%E2%9D%8C-**avoid**-using-%60string.split%60-on-simple-patterns
Code Source: https://github.com/dotnet/project-system/blob/main/src/Microsoft.VisualStudio.ProjectSystem.Managed/Text/LazyStringSplit.cs
Tests: https://github.com/dotnet/project-system/blob/main/tests/Microsoft.VisualStudio.ProjectSystem.Managed.UnitTests/Text/LazyStringSplitTests.cs 

I did want to talk about this, should I copy the tests they implemented for this? I thought about implementing the suggestion from the wiki entry but this lazy string split seems more reusable.

src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileFormat.cs
jgonz120: It's used in the unit tests.

src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileFormat.cs
jgonz120: Same, unit tests

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamAssetsLogMessageConverter.cs
jgonz120: I added a test to verify this, if the log message has is invalid it continues to parse through the object until it hits the end. Skipping in this scenario is the same as a standard "read", it moves the reader to the value of the property, so on the next loop it moves onto the next property in the log. So if we have 2 logs, one valid and one invalid, the object will contain the one valid log.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileConverter.cs
jgonz120: When the name is passed in we use both the name and packageSpecPath to set the Name and FilePath propertieson the packageSpec. If the name isn't set we ignore the packageSpecPath. It's also used for error messages. If they aren't set we use the RestoreMetaData values.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetLibraryConverter.cs
jgonz120: I'll have to check through other instances of this but for the LockFileTargetLibrary the list is set in the getter for the property.

src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileFormat.cs
nkolev92: I'm going into nit mode, so feel free to tell me to bug off here :D 

Should the unit tests just call the methods with multiple parameters? The extra methods can be in the tests instead.
Having extra methods that are specific for the tests adds cognitive load to an already long class.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamAssetsLogMessageConverter.cs
nkolev92: thanks for adding a test.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetLibraryConverter.cs
nkolev92: It sets it if it's not there. 
If you already know you can have array empty, then I think you can just set that.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileConverter.cs
nkolev92: Does that mean we need to pass it down then?
I was checking for equivalence and it seemed like the old implementation was passing them but I could be wrong.

src/NuGet.Core/NuGet.ProjectModel/LazyStringSplit.cs
nkolev92: I think we can just add the tests.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileConverter.cs
jgonz120: oh no, I don't think we need to pass it down. The [old implementation](https://github.com/NuGet/NuGet.Client/blob/16939136c5722cc91435917e4fd6da2029a1c07d/src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileFormat.cs#L791C42-L791C56) passed null too. The path was being passed down to ReadCentralTransitiveDependencyGroup and only used for exception messages. I can update the classes to accept file path if you think it's worth passing it down.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetLibraryConverter.cs
jgonz120: Seems fine with me, updated.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileConverter.cs
nkolev92: Let's keep parity. 

Once we've confirmed and removed the nj implementation, we can do cleanup on the now internal implementation :D

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetLibraryConverter.cs
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        private IList<PackageDependency> ReadPackageDependencyList(ref Utf8JsonStreamReader reader)
+        {
+            if (reader.TokenType != JsonTokenType.StartObject)
+            {
+                return Array.Empty<PackageDependency>();
+            }
+
+            var packageDependencies = new List<PackageDependency>();
```
jgonz120: @nkolev92 / @zivkan  do you know of a good way to determine what the most common value is for the number of packageDependencies? 

[Original Comment](https://github.com/NuGet/NuGet.Client/pull/5529#discussion_r1414975442) 
> Better default capacity based on most common value? This is defaulting to 4. 
-@davkean

nkolev92: It's difficult to validate whatever you chose is a good value. I've attempted to choose one before and here's what I've considered:

- I added a logger in the code that printed the number of dependencies. I ran restore for a few different repos and looked at the results. Median varied between 2-4 for different repos, and average was between 3-5. 
- Look at the average number of dependencies for packages on nuget.org and got something like 3.1. 

See discussion: https://github.com/NuGet/NuGet.Client/pull/5531#discussion_r1417952158.
I think the defaults are ok, but I'm a lot less knowledgeable about this than others :)

src/NuGet.Core/NuGet.ProjectModel/StringExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    internal static class StringExtensions
+    {
+        internal static (string firstPart, string? secondPart) SplitInTwo(this string s, char separator)
+        {
+            if (string.IsNullOrEmpty(s))
+            {
+                return (s, null);
+            }
+#if NETCORE5_0
```
jeffkl: This will need to be NETCOREAPP5_OR_GREATOR

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetLibraryConverter.cs
jgonz120: 3.1 is close enough to 4 that I feel comfortable leaving it as is.

src/NuGet.Core/NuGet.ProjectModel/JsonUtility.cs
```diff
@@ -141,5 +170,36 @@ internal static JToken WriteString(string item)
         {
             return item != null ? new JValue(item) : JValue.CreateNull();
         }
+
+        internal static NuGetVersion ParseNugetVersion(string value)
```
jgonz120: https://github.com/NuGet/NuGet.Client/pull/5529#discussion_r1414976725

@davkean @nkolev92  continuing this conversation here. Would we want the cache of version/version range to live only during the parsing of the file?

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileItemConverter.cs
```diff
@@ -0,0 +1,51 @@
...
+            {
+                throw new JsonException("Expected PropertyName, found " + reader.TokenType);
+            }
+
+            //We want to read the property name right away
+            var lockItemPath = reader.GetString();
+            LockFileItem lockFileItem;
+            if (genericType == typeof(LockFileContentFile))
+            {
+                lockFileItem = new LockFileContentFile(lockItemPath);
```
davkean: I think this will result in a lot less branches during parsing if you make this a `Func<string, T> lockFileItemCreator` that gets passed into the constructor instead of branching based on the generic type.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetLibraryConverter.cs
```diff
@@ -0,0 +1,159 @@
...
+using System.Text;
+using System.Text.Json;
+using NuGet.Packaging.Core;
+using NuGet.Versioning;
+
+namespace NuGet.ProjectModel
+{
+    /// <summary>
+    /// A <see cref="Utf8JsonStreamReaderConverter{T}"/> to allow read JSON into <see cref="LockFileTargetLibrary"/>
+    /// </summary>
```
davkean, line 15: It would be super helpful if these convertors showed an example of what exactly they were parsing to be able to match that to the code.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetLibraryConverter.cs
```diff
@@ -0,0 +1,159 @@
...
+                throw new JsonException("Expected PropertyName, found " + reader.TokenType);
+            }
+
+            var lockFileTargetLibrary = new LockFileTargetLibrary();
+
+            //We want to read the property name right away
+            var propertyName = reader.GetString();
+            var (targetLibraryName, version) = propertyName.SplitInTwo('/');
+            lockFileTargetLibrary.Name = targetLibraryName;
+            lockFileTargetLibrary.Version = version is null ? null : JsonUtility.ParseNugetVersion(version);
```
davkean: Separate to this PR, but JsonUtility.ParseNuGetVersion should be a candidate for moving to parsing from a UTF8 string, or at least a ReadOnlySpan<char> to avoid the wasted allocation to split the version into a string, only to throw it away. This is about ~3% of allocations in the trace you shared.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetLibraryConverter.cs
davkean: I looked at the trace, 6% of allocations is this collection being resized, so it feels like I would favor a little higher.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetLibraryConverter.cs
```diff
@@ -0,0 +1,159 @@
...
+            {
+                return Array.Empty<PackageDependency>();
+            }
+
+            var packageDependencies = new List<PackageDependency>();
+            while (reader.Read() && reader.TokenType == JsonTokenType.PropertyName)
+            {
+                string propertyName = reader.GetString();
+                string versionString = reader.ReadNextTokenAsString();
+
```
davkean: 35% of allocations come from these values, I feel like sharing the strings for common dependencies would be helpful in reducing allocations, especially given the version number is immediately thrown away.

src/NuGet.Core/NuGet.ProjectModel/StringExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+        internal static (string firstPart, string? secondPart) SplitInTwo(this string s, char separator)
+        {
+            if (string.IsNullOrEmpty(s))
+            {
+                return (s, null);
+            }
+#if NETCOREAPP2_1_OR_GREATER
+            var index = s.IndexOf(separator, StringComparison.OrdinalIgnoreCase);
+#else
+            var index = s.IndexOf(separator.ToString(), StringComparison.OrdinalIgnoreCase);
```
davkean: Instead of this allocation, call [this overload](https://learn.microsoft.com/en-us/dotnet/api/system.string.indexof?view=net-8.0#system-string-indexof(system-char)), which defaults to 'Ordinal'. You only ever pass a slash to this method so the results will not change.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamProjectFileDependencyGroupConverter.cs
```diff
@@ -0,0 +1,28 @@
...
+        public ProjectFileDependencyGroup Read(ref Utf8JsonStreamReader reader)
+        {
+            if (reader.TokenType != JsonTokenType.PropertyName)
+            {
+                throw new JsonException("Expected PropertyName, found " + reader.TokenType);
+            }
+
+            var frameworkName = reader.GetString();
+            reader.Read();
+            var dependencies = reader.ReadStringArrayAsIList(new List<string>());
```
davkean: Is 'dependencies' ever empty? If so, figure out way to `Array.Empty<string>` instead of `new List<string>()` as the default.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetLibraryConverter.cs
```diff
@@ -0,0 +1,159 @@
...
+                }
+                else if (reader.ValueTextEquals(DependenciesPropertyName))
+                {
+                    reader.Read();
+                    lockFileTargetLibrary.Dependencies = ReadPackageDependencyList(ref reader);
+                }
+                else if (reader.ValueTextEquals(FrameworkAssembliesPropertyName))
+                {
+                    reader.Read();
+                    lockFileTargetLibrary.FrameworkAssemblies = reader.ReadStringArrayAsIList(new List<string>());
```
davkean: Is 'FrameworkAssemblies' ever empty? If so, figure out way to `Array.Empty<string>` instead of `new List<string>()` as the default.

src/NuGet.Core/NuGet.ProjectModel/JsonUtility.cs
nkolev92: I think defining a great lifetime for these caches would be challenging.

I'm comfortable if this change starts with the assets file read scope, but we should look into expanding it.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamProjectFileDependencyGroupConverter.cs
nkolev92: We should basically check all `new List` changes are considering whether Array.Empty is appropriate. 
https://github.com/NuGet/NuGet.Client/pull/5558#discussion_r1441142371.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetLibraryConverter.cs
jgonz120: Switched it to 10.

src/NuGet.Core/NuGet.ProjectModel/JsonUtility.cs
```diff
@@ -6,13 +6,19 @@
...
+using NuGet.Common;
 using NuGet.Packaging.Core;
 using NuGet.Versioning;
 
 namespace NuGet.ProjectModel
 {
     internal static class JsonUtility
     {
+        private static readonly Dictionary<string, NuGetVersion> NuGetVersionCache = new();
+        private static readonly Dictionary<string, VersionRange> VersionRangeCache = new();
```
zivkan: This will leak memory. Every version and version range ever parsed during the entire lifetime of the process will be retained.

jgonz120: Yea I'm not super stoked on this solution. I asked in [this comment](https://github.com/NuGet/NuGet.Client/pull/5558#discussion_r1448182312) if we should consider a specific lifetime for these dictionaries.

nkolev92: fwiw, my recommendation in that thread was to make the lifetime an assets file read, not a static inside the class.

jgonz120: yea I think that scope makes the most sense.

src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamLockFileTargetLibraryConverter.cs
nkolev92: > I looked at the trace, 6% of allocations is this collection being resized, so it feels like I would favor a little higher.

How do we decide how to trade off allocations unnecessarily vs resizing? Why is resizing worse than allocation extra unnecessarily?

nkolev92: fwiw, the above dependencies count calculation does not account for how many dependencies a project itself references. That's probably higher, but unfortunately we don't have data for that.

So wouldn't shock me if higher than that works better.

src/NuGet.Core/NuGet.ProjectModel/JsonUtility.cs
zivkan: The title of this PR is to migrate to System.Text.Json, and the existing Newtonsoft.Json code doesn't do caching, so can you bump caching work to a different PR?

I'm behind on my other priorities, but ideally I should start work on https://github.com/NuGet/Home/pull/12124 in the next sprint or two. For that work, I need this PR merged into the feature branch, and the feature branch merged into the dev branch.

If merging the assets file System.Text.Json work is delayed, then that will delay my ability to start on this other work. Therefore, I'd like to minimize scope creep that blocks other work.

jgonz120: I love that, so I went ahead and undid the caching stuff and saved a stash of it.

src/NuGet.Core/NuGet.ProjectModel/JsonUtility.cs
jgonz120: Undid this change per Andy's suggestion to reduce scope creep.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -104,6 +105,52 @@ internal void Skip()
             }
         }
 
+        internal IList<T> ReadObjectAsList<T>(IUtf8JsonStreamReaderConverter<T> streamReaderConverter)
+        {
+            if (TokenType != JsonTokenType.StartObject)
+            {
+                return Array.Empty<T>();
```
zivkan, line 112: If this was `if (TokenType == JsonTokenType.Null) { return Array.Empty<T>(); }`, I'd be ok with it.

But the way this code is written makes me think that json `{ "propertyName": 4 }` would return the empty array, rather than an exception saying the JSON is invalid.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -104,6 +105,52 @@ internal void Skip()
...
+            }
+            while (TokenType != JsonTokenType.EndObject);
+            return listObjects;
+        }
+
+        internal void ReadArrayOfObjects<T1, T2>(IList<T2> objectList, IUtf8JsonStreamReaderConverter<T1> streamReaderConverter) where T1 : T2
+        {
+            if (objectList is null)
+            {
+                return;
```
zivkan: From an API design point of view, I don't understand why it would be acceptable to call this method with a null `objectList`, and for this method to no-op in that scenario. If there's data to be read.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -104,6 +105,52 @@ internal void Skip()
...
+            }
+
+            if (Read() && TokenType == JsonTokenType.StartArray)
```
zivkan: if this expression returns false, I think it could hide a json schema exception. This violates the principal of failing early, so that errors are clearly understood.

For example, if there's an object `{ "something": 5, "another": "value" }`, where the `something` should be an array of objects, this code will silently ignore the error, and the `another` property would still be correctly read, assuming it's supposed to have a string value. The result would potentially be deserialized into an object with an invalid state (if `something` is not supposed to be nullablle/empty), rather than throwing an exception to report the json error.

test/NuGet.Core.Tests/NuGet.ProjectModel.Test/JsonPackageSpecReaderTests.cs
```diff
@@ -196,26 +196,26 @@ public void PackageSpecReader_ExplicitIncludesOverrideTypePlatform(IEnvironmentV
         }
 
         [Theory]
-        [MemberData(nameof(TestEnvironmentVariableReader), "{}")]
-        [MemberData(nameof(TestEnvironmentVariableReader), "{}")]
+        [MemberData(nameof(TestEnvironmentVariableReader), "{}", MemberType = typeof(LockFileParsingEnvironmentVariable))]
+        [MemberData(nameof(TestEnvironmentVariableReader), "{}", MemberType = typeof(LockFileParsingEnvironmentVariable))]
```
zivkan: Aren't these two identical/duplicate?

test/NuGet.Core.Tests/NuGet.ProjectModel.Test/LockFileParsingEnvironmentVariable.cs
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Test.Utility;
+
+namespace NuGet.ProjectModel
+{
+    public class LockFileParsingEnvironmentVariable
```
zivkan:
```suggestion
    public static class LockFileParsingEnvironmentVariable
```

All methods members are static, so the class can be declared as static too

src/NuGet.Core/NuGet.ProjectModel/JsonUtility.cs
```diff
@@ -51,6 +60,23 @@ internal static PackageDependency ReadPackageDependency(string property, JToken
                 versionStr == null ? null : VersionRange.Parse(versionStr));
         }
 
+        internal static bool UseNewstonSoftJsonForParsing(IEnvironmentVariableReader environmentVariableReader, bool bypassCache)
```
zivkan:
```suggestion
        internal static bool UseNewtonsoftJsonForParsing(IEnvironmentVariableReader environmentVariableReader, bool bypassCache)
```

zivkan (commented):
I have a few correctness concerns about methods in the shared/generic `Utf8JsonStreamReader` class. Even if specific converters are using them in a safe way, since it's in a utility class that could be called by any (internal) method, I think that raises the quality bar so that it should adhere to the principal of least surprise.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: Yea, I changed it to just return the list.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
jgonz120: Updated it to check throw an error if it's called without an array.

zivkan (approved):

jgonz120:
@zivkan @nkolev92 one thought, I currently have the default path for the restore to use STJ if the environment variable isn't set. Do we want to release this with that default or release it still using Newtonsoft by default asking internal folks to opt into it?

zivkan:
We don't expect 3rd party tools to write the asset file, and we have a lot of integration tests. So, I think if the tests pass using STJ, then it should be quite safe, but the env var to fall back to NJ is a sufficient safety barrier. So, I think that STJ by default is fine.

nkolev92 (approved):

nkolev92:
Great job getting this done!

You probably need to do a rebase before merging into dev.
If you look at the feature branch and see which dev commits are missing,https://github.com/NuGet/NuGet.Client/compare/dev-jgonz120-FeatureBranch-Stj-Migration...dev, this one https://github.com/NuGet/NuGet.Client/commit/95f6e0c8b7bd627862a08dcf60f43397f42723ca has a conflict with you changes, so it'll probably need some work to get it to work equivalently.

PR #5559, opened 2023-12-26
Fix non-SDK csproj with CPM and static graph restore

### LINKED ISSUE #13108: CPM doesn't work with non-SDK style projects and static graph restore
### NuGet Product Used

MSBuild.exe

### Product Version

internal preview

### Worked before?

public preview

### Impact

### Repro Steps & Context

Have a non-SDK style project that uses CPM and restore static graph evaluation. Restore behaves as if CPM is disabled.

This regressed in https://github.com/NuGet/NuGet.Client/pull/5528

### Verbose Logs

_No response_

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13108

Regression? Yes Last working version: 6.9.0.54

## Description

Removed the `&& projectStyle == ProjectStyle.PackageReference`, because it's being checked by the `if` statement 2 lines above, which also handles the `projectStyle == null`, because static restore can't figure out the project style.

I also added variables for each CPM property being parsed, because doing everything in a single expression makes it hard to inspect values in the debugger.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/test/NuGet.Core.FuncTests/Msbuild.Integration.Test/MsbuildRestoreTaskTests.cs
+++ b/test/NuGet.Core.FuncTests/Msbuild.Integration.Test/MsbuildRestoreTaskTests.cs
@@ -12,6 +12,7 @@ using NuGet.Commands;
 using NuGet.Common;
 using NuGet.Frameworks;
 using NuGet.Packaging;
+using NuGet.Packaging.Core;
 using NuGet.ProjectModel;
 using NuGet.Test.Utility;
 using NuGet.Versioning;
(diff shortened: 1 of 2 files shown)


test/NuGet.Core.FuncTests/Msbuild.Integration.Test/MsbuildRestoreTaskTests.cs
```diff
@@ -1321,5 +1322,57 @@ public void MsbuildRestore_WithLegacyCsproj_GlobalPackageReferencesAreProcessed(
...
+            // By default, this is not an error condition, just a warning, but it will be limited to stable versions.
+            // By using a SemVer prerelease version, we ensure that restore always fails if CPM is not enabled (NU1103).
+            var packageX = new SimpleTestPackageContext()
+            {
+                Id = "x",
+                Version = "1.0.0-rc.1"
+            };
+            await SimpleTestPackageUtility.CreateFolderFeedV3Async(pathContext.PackageSource, packageX);
+
+            var projectA = SimpleTestProjectContext.CreateLegacyPackageReference("a",
```
kartheekp-ms, line 1344: Did this test fail before the product changes introduced in this PR were implemented? The reason I'm asking is that, upon examining the enum definition for [ProjectStyle](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.ProjectModel/ProjectStyle.cs), I noticed there's no separate constant defined for legacy package reference. Therefore, I assume this project will have `ProjectStyle` set to `ProjectStyle.PackageReference`. In this case, the code should work because ProjectStyle is set to PackageReference. Please let me know if I've missed anything.

zivkan: Yes, I wrote the test before making the product change, to verify that the test was failing for static graph restore, but working for normal restore.

The MSBuild property `RestoreProjectStyle` is set by an MSBuild target `_GetRestoreProjectStyle` (which in turn is set by a task with an output parameter), which get run during "dgspec generation" during non-static graph restore. However, our static graph restore code doesn't run the `_GetRestoreProjectStyle` target. The .NET SDK explicitly sets `RestoreProjectStyle` in their props/targets, but for non-SDK style project our target usually sets it (so it can detect packages.config, PackageReference, or project.json). Therefore, our product code is currently getting a null for the value, hence why the `if` statement 2 lines up my change in the product code is allowing nulls for `projectStyle`.

I'm not 100% sure, but I think it's also related to why:
* https://github.com/NuGet/Home/issues/13109

kartheekp-ms (approved):

test/NuGet.Core.FuncTests/Msbuild.Integration.Test/MsbuildRestoreTaskTests.cs
kartheekp-ms: > However, our static graph restore code doesn't run the _GetRestoreProjectStyle target. 

I approved the PR. May the right fix is to ensure static graph restore runs that target.

jeffkl (approved):

test/NuGet.Core.FuncTests/Msbuild.Integration.Test/MsbuildRestoreTaskTests.cs
zivkan: I agree we should fix reading the `RestoreProjectStyle` property in static graph restore, but the product code was still, in my opinion, a bit non-sensical. The `if` block guarding the property reads is already checking the `projectStyle` parameter, and it doesn't make sense to me why one property should have a different condition to the other properties. So, I think the product change is a good one anyway.

PR #5561, opened 2023-12-27
Static graph restore should work with packages.config only solution

### LINKED ISSUE #13109: Static graph restore doesn't restore packages.config only solution
### NuGet Product Used

MSBuild.exe

### Product Version

all

### Worked before?

_No response_

### Impact

None

### Repro Steps & Context

Have a solution with only packages.config projects. `msbuild -t:restore -p:RestorePackagesConfig=true` will successfully restore the packages.config packages, but `msbuild -t:restore -pRestorePackagesConfig=true -p:RestoreUseStaticGraphEvaluation=true` will not.

[staticRestorePC.zip](https://github.com/NuGet/Home/files/13771902/staticRestorePC.zip)

~I think it's caused by https://github.com/NuGet/NuGet.Client/pull/3116~ edit: it was not

~Previously, MSBuild's static graph API could have evaluated the MSBuild script, but since that PR, a target needs to be executed to populate the `RestoreProjectStyle` MSBuild property.~

### Verbose Logs

_No response_

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13109

Regression? no

## Description

Static graph restore has an early exit that "normal" restore doesn't, in order to provide a different error message. However, the early exist code assumed there'd be at least 1 PackageReference project. While it probably doesn't make much sense to use static graph restore with a packages.config only solution, IMO it shouldn't have such a behaviour difference.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/test/NuGet.Core.FuncTests/Msbuild.Integration.Test/MsbuildRestoreTaskTests.cs
+++ b/test/NuGet.Core.FuncTests/Msbuild.Integration.Test/MsbuildRestoreTaskTests.cs
@@ -29,8 +29,10 @@ namespace Msbuild.Integration.Test
             _msbuildFixture = fixture;
         }
 
-        [PlatformFact(Platform.Windows)]
-        public async Task MsbuildRestore_PackagesConfigDependencyAsync()
+        [PlatformTheory(Platform.Windows)]
+        [InlineData(true)]
+        [InlineData(false)]
+        public async Task MsbuildRestore_PackagesConfigDependencyAsync(bool useStaticGraphRestore)
         {
             // Arrange
             using (var pathContext = new SimpleTestPathContext())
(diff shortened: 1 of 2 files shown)


test/NuGet.Core.FuncTests/Msbuild.Integration.Test/MsbuildRestoreTaskTests.cs
```diff
@@ -1173,8 +1178,8 @@ public async Task MsbuildRestore_PackagesConfigDependency_WithHttpSource_Warns()
                     packageX);
 
                 // Act
-                var result = _msbuildFixture.RunMsBuild(pathContext.WorkingDirectory, $"/t:restore {pathContext.SolutionRoot} /p:RestorePackagesConfig=true", ignoreExitCode: true);
-
+                string args = $"/t:restore {pathContext.SolutionRoot} /p:RestorePackagesConfig=true /p:RestoreUseStaticGraphEvaluation={useStaticGraphEvaluation}";
```
nkolev92, line 1181: Is this catching anything different from the other test? Feels redundant.

nkolev92 (approved):

jeffkl (approved):

test/NuGet.Core.FuncTests/Msbuild.Integration.Test/MsbuildRestoreTaskTests.cs
zivkan: To be honest, I just searched for packages.config tests and then made then run both with & without static graph restore. I didn't look at what the tests were actually doing.

PR #5563, opened 2023-12-28
when accessing a private feed, try last known valid credentials before invoking credential providers.

### LINKED ISSUE #11210: [Bug]: VS Credential Provider Incorrectly Setting Value of `isRetry`
### NuGet Product Used

### Product Version

Microsoft Visual Studio Professional 2019 Version 16.10.3 (NuGet Package Manager 5.10.0)

### Worked before?

_No response_

### Impact

## Problem

The [documentation](https://docs.microsoft.com/en-us/nuget/reference/extensibility/nuget-credential-providers-for-visual-studio#creating-a-nuget-credential-provider-for-visual-studio) for NuGet credential providers for Visual Studio (NCPVS) states that the value of the `isRetry` parameter should be set to true "if credentials were previously requested for this Uri, but the supplied credentials did not allow authorized access."

[Other](https://docs.microsoft.com/en-us/nuget/reference/extensibility/nuget-exe-credential-providers#input-parameters) Microsoft documentation (albeit not specifically for VS, but still for NuGet credential providers) states that "[IsRetry] indicates that this attempt is a retry of a previously failed attempt."

For context, we have an on-premise authenticated NuGet feed that is configured to disallow anonymous access. That is, each user must provide an Nuget-specific username & password when accessing that NuGet feed. Each time an operation is performed on the  feed from within the Visual Studio Package Manager UI (e.g. list all packages, install a package), we're presented with a 401 Unauthorized challenge. Then, if valid credentials are presented, the challenge flow proceeds successfully, but if not, we are met with a 403 Forbidden.

I can verify that my NCPVS is being called with `isRetry == false` for the first operation, and then `isRetry == true` for all subsequent operations. This is contrary to what I would expect -- I would expect `isRetry == false` for all of the subsequent requests because the NuGet-specific username and password remains the same, so there is no reason why the set of credentials should not be authorized. Indeed, after debugging my NCPVS, the reason for setting `isRetry == true` is not because the credentials were explicitly forbidden, but rather because the request to the NuGet feed returned a 401 challenge.

Setting `isRetry == true` only upon encountering a 403 makes sense because a NuGet feed that requires authentication should be forcing the NuGet client (whether that be VS, `NuGet.exe`, etc.) to authenticate with a 401 request each time we "hit" that protected NuGet feed. I believe that NuGet should only retry if the provided credentials were explicitly forbidden, not if we were simply presented with a 401 challenge.

As it stands right now, the existing implementation will cause our NCPVS to have its [GetCredentialAsync](https://docs.microsoft.com/en-us/nuget/reference/extensibility/nuget-credential-providers-for-visual-studio#getcredentialasync) method with `isRetry == true`. Becuase there's no other way of determining whether the reason for setting `isRetry == true` was because of a 401 Unauthorized, 403 Forbidden, or something else, we always assume the worst case (i.e. a 403 caused by stale credentials) and obtain new credentials from our internal service. This causes unnecessary delay in returning correct credentials from the credential provider, and strain on our internal services.

## Proposed Solution

The logic for determining what value to assign `isRetry` is at [`CredentialService.cs:93`](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Credentials/CredentialService.cs#L93). From inspection, I'm deducing that the `_retryCache`'s purpose is to record which credential providers have authenticated with which URLs. The `_retryCache` for a given credential provider and URL will be set to true if the NCPVS has attempted to authenticate with this URL before, irrespective of the [reason](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Credentials/CredentialService.cs#L77) (e.g. Unauthorized, Proxy, Forbidden).

I propose that the `isRetry` variable should only be set to true if the [`type`](https://gist.github.deshaw.com/leong/bfc381e551cbbadfc029071de05a92bc#file-credentialservice-cs-L77) argument to the `GetCredentialsAsync` method is `type == CredentialRequestType.Forbidden`. Indeed, in [NuGet's source code](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Configuration/Credential/CredentialRequestType.cs#L18-L21) for `CredentialRequestType.Forbidden`, it states that "A different set of credentials could solve this failure". Indeed, this is consistent with the documentation quoted in the first two paragraphs of this bug report. Thus, it makes sense only to set `isRetry` to `true` if the following conditions are met:

```csharp
diff --git a/src/NuGet.Core/NuGet.Credentials/CredentialService.cs b/src/NuGet.Core/NuGet.Credentials/CredentialService.cs
index 2524dd0a1..c6fe34461 100644
--- a/src/NuGet.Core/NuGet.Credentials/CredentialService.cs
+++ b/src/NuGet.Core/NuGet.Credentials/CredentialService.cs
@@ -90,7 +90,7 @@ public CredentialService(AsyncLazy<IEnumerable<ICredentialProvider>> providers,
 
                 var retryKey = RetryCacheKey(uri, type, provider);
-                var isRetry = _retryCache.ContainsKey(retryKey);
+                var isRetry = (type == CredentialRequestType.Forbidden) && _retryCache.ContainsKey(retryKey);
 
                 try
                 {
```

### Verbose Logs

_No response_

kartheekp-ms:
## Bug

Fixes: https://github.com/NuGet/Home/issues/11210

## Description

The [documentation](https://docs.microsoft.com/en-us/nuget/reference/extensibility/nuget-credential-providers-for-visual-studio#creating-a-nuget-credential-provider-for-visual-studio) for NuGet credential providers for Visual Studio specifies that the value of the isRetry parameter should be set to true `if credentials were previously requested for this URI, but the supplied credentials did not allow authorized access.` However, credential providers are initially invoked with `isRetry = false` for the first operation, and then with `isRetry = true` for all subsequent operations.

In the case of the `Azure Artifacts Credential Provider`, the `isRetry` value has a different meaning: If set to false or unset, cached credentials are returned to the caller. The caller must validate these returned credentials, and if found invalid, should call the credential provider again with -IsRetry set. If set to true, the credential provider will obtain new credentials instead of returning potentially invalid ones from the cache.

This behavior could impact the MSBuild server feature, as discussed in this [issue comment](https://github.com/microsoft/artifacts-credprovider/issues/347#issuecomment-1363093587).

With the changes proposed in this PR, the NuGet Client would first verify if the cached credentials are effective before invoking credential providers with `isRetry` set to true. The proposed changes will potentially reduce the number of prompts in Visual Studio and also decrease the frequency of cache invalidation in .NET CLI.

Andy clearly explained the old (current) behavior in [this issue comment](https://github.com/NuGet/Home/issues/11210#issuecomment-1132240645). I have attempted to describe the new behavior that will be implemented once this PR is merged, as follows:

```mermaid

sequenceDiagram
  autonumber
  actor Customer
  participant CredProvider
  participant CredentialService
  participant NuGet
  participant Server
  NuGet->>Server: no auth header
  Server->>NuGet: HTTP 401
  NuGet->> CredentialService: Check for cached credentials
  CredentialService->>NuGet: No cached credentials
  NuGet->> CredentialService: Acquire credentials from providers
  CredentialService->>CredProvider: isRetry: false
  CredProvider->>CredProvider: Check for cached token
  CredProvider->>Customer: Prompt for username/password
  Customer->>CredProvider: Enters CorrectAccount, correct password
  CredProvider->>NuGet: HTTP BASIC Combo1
NuGet->>Server: Auth: HTTP BASIC Combo1
  Server->>NuGet: HTTP 200
  NuGet->>Server: Auth: no auth header
  Server->>NuGet: HTTP 401
  NuGet->> CredentialService: Check for cached credentials
  CredentialService->>NuGet: Returns cached credentials
  NuGet->>Server: Auth: HTTP BASIC Combo1
  Server->>NuGet: HTTP 200
  NuGet->>Server: Auth: no auth header
  Server->>NuGet: HTTP 401
  NuGet->> CredentialService: Check for cached credentials
  CredentialService->>NuGet: Returns cached credentials
  NuGet->>Server: Auth: HTTP BASIC Combo1
  Server->>NuGet: HTTP 401 (cached tokens expired)
  NuGet->> CredentialService: Acquire credentials from providers
  CredentialService->>CredProvider: isRetry: true
  CredProvider->>Customer: Prompt for username/password
  Customer->>CredProvider: Enters CorrectAccount, correct password
  CredProvider->>NuGet: HTTP BASIC Combo2
  NuGet->>Server: Auth: HTTP BASIC Combo2
  Server->>NuGet: HTTP 200
```

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
 
- **Documentation**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpSourceAuthenticationHandler.cs
+++ b/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpSourceAuthenticationHandler.cs
@@ -158,8 +158,12 @@ namespace NuGet.Protocol
             }
         }
 
-        private async Task<ICredentials> AcquireCredentialsAsync(HttpStatusCode statusCode, Guid credentialsVersion, ILogger log,
-            bool areLastKnownGoodCredentialsTried, CancellationToken cancellationToken)
+        private async Task<ICredentials> AcquireCredentialsAsync(
+            HttpStatusCode statusCode,
+            Guid credentialsVersion,
+            ILogger log,
+            bool areLastKnownGoodCredentialsTried,
+            CancellationToken cancellationToken)
         {
             // Only one request may prompt and attempt to auth at a time
             await _httpClientLock.WaitAsync(cancellationToken);
@@ -206,6 +210,7 @@ namespace NuGet.Protocol
                 ICredentials promptCredentials = default;
                 if (!areLastKnownGoodCredentialsTried)
                 {
+                    // isProxy is false because the previous if statement allows only Unauthorized or Forbidden types, not Proxy.
                     _ = _credentialService.TryGetLastKnownGoodCredentialsFromCache(uri: _packageSource.SourceUri, isProxy: false, out promptCredentials);
                 }
 
--- a/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpSourceAuthenticationHandler.cs
+++ b/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpSourceAuthenticationHandler.cs
@@ -76,6 +76,7 @@ namespace NuGet.Protocol
             ICredentials promptCredentials = null;
 
             var configuration = request.GetOrCreateConfiguration();
+            bool areLastKnownGoodCredentialsTried = false;
 
             // Authorizing may take multiple attempts
             while (true)
@@ -127,7 +128,10 @@ namespace NuGet.Protocol
                         response.StatusCode,
                         beforeLockVersion,
                         configuration.Logger,
+                        areLastKnownGoodCredentialsTried,
                         cancellationToken);
+
+                    areLastKnownGoodCredentialsTried = true;
 
                     if (stopwatches != null)
                     {
@@ -154,7 +158,8 @@ namespace NuGet.Protocol
             }
         }
 
-        private async Task<ICredentials> AcquireCredentialsAsync(HttpStatusCode statusCode, Guid credentialsVersion, ILogger log, CancellationToken cancellationToken)
+        private async Task<ICredentials> AcquireCredentialsAsync(HttpStatusCode statusCode, Guid credentialsVersion, ILogger log,
+            bool areLastKnownGoodCredentialsTried, CancellationToken cancellationToken)
         {
             // Only one request may prompt and attempt to auth at a time
             await _httpClientLock.WaitAsync(cancellationToken);
@@ -198,7 +203,13 @@ namespace NuGet.Protocol
                         _packageSource.Source);
                 }
 
-                var promptCredentials = await PromptForCredentialsAsync(
+                ICredentials promptCredentials = default;
+                if (!areLastKnownGoodCredentialsTried)
+                {
+                    _ = _credentialService.TryGetLastKnownGoodCredentialsFromCache(uri: _packageSource.SourceUri, isProxy: false, out promptCredentials);
+                }
+
+                promptCredentials ??= await PromptForCredentialsAsync(
                     type,
                     message,
                     authState,
--- a/test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
+++ b/test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
@@ -137,7 +137,7 @@ namespace NuGet.Protocol.FuncTest
                     _ = await source.GetAsync(request, ProcessResponse, logger.Object, cancellationToken);
 
                     request = new HttpSourceCachedRequest(packageUrl, "2", httpSourceCacheContext);
-                    _ = await source.GetAsync(request, ProcessResponse, logger.Object, cancellationToken);                    
+                    _ = await source.GetAsync(request, ProcessResponse, logger.Object, cancellationToken);
                 }
                 finally
                 {
--- a/test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
+++ b/test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
@@ -145,7 +145,36 @@ namespace NuGet.Protocol.FuncTest
                 }
 
                 // Assert
+
+                // Each attempt to access a private feed initially receives a 401 Unauthorized response.
+                // Following the 401 response, NuGet attempts to acquire the necessary credentials.
+                // These credentials are then used for subsequent requests to the feed.
+                // Note: If 'HttpClientHandler.PreAuthenticate' is set to true, this behavior might differ as 
+                // credentials would be sent preemptively with the initial request.
+
+                // In this test, the expected number of requests is 4, but it turns out to be 5. This discrepancy occurs because 
+                // we initialize _credentials =  new HttpSourceCredentials() in the HttpSourceAuthenticationHandler constructor.
+                // The HttpHandler pipeline operates under the assumption that credentials have already been set, although they have not.
+                // As a result, NuGet initially sends 2 requests to the server without credentials when accessing the private feed for the first time
+                // and then invokes the credential service.
+
+                // Request flow while accessing index.json
+                // NuGet -> Server (No Credentials) - 1st request
+                // Server -> NuGet (401 Unauthorized)
+                // NuGet -> Server (Sends HttpClientHandler.Credentials) - 2nd request
+                // Server -> NuGet (401 Unauthorized because HttpClientHandler.Credentials returns a null value by default)
+                // NuGet -> Credential Service
+                // Credential Service -> NuGet (Returns credentials)
+                // NuGet -> Server (Sends credentials received from the credential service) - 3rd request
+                // Server -> NuGet (200 OK)
+
+                // Request flow while retrieving package information
+                // NuGet -> Server (No Credentials) - 4th request
+                // Server -> NuGet (401 Unauthorized)
+                // NuGet -> Server (Sends HttpClientHandler.Credentials) - 5th request
+                // Server -> NuGet (200 OK)
                 server.Requests.Should().HaveCount(5);
+                // This should have been 2, but is 3 because of the reason mentioned above.
                 server.Requests.Count(RequestWithoutAuthorizationHeader).Should().Be(3);
                 server.Requests.Any(r => r.Url!.OriginalString == serviceIndexUrl).Should().BeTrue();
 
@@ -157,6 +186,11 @@ namespace NuGet.Protocol.FuncTest
                 }
 
                 mockedCredentialService.Verify(x => x.GetCredentialsAsync(It.IsAny<Uri>(), It.IsAny<IWebProxy>(), It.IsAny<CredentialRequestType>(), It.IsAny<string>(), It.IsAny<CancellationToken>()), Times.Once);
+                // The method TryGetLastKnownGoodCredentialsFromCache is invoked only once, despite the client sending two requests:
+                // the first to access index.json and the second to retrieve package information.
+                // This occurs because when HttpClientHandler.Credentials is set (i.e., not null) while accessing the index.json.
+                // These credentials are automatically sent with requests to the private feed upon receiving 401 challenge.
+                // As a result, NuGet does not need to invoke the credential service again while retrieving the package information.
                 mockedCredentialService.Verify(x => x.TryGetLastKnownGoodCredentialsFromCache(It.IsAny<Uri>(), It.IsAny<bool>(), out It.Ref<ICredentials>.IsAny), Times.Once);
 
                 // ExecuteAsync returns Task<T>, so need to return something to give it a <T>.
--- a/test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
+++ b/test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
@@ -150,7 +150,7 @@ namespace NuGet.Protocol.FuncTest
                 // Following the 401 response, NuGet attempts to acquire the necessary credentials.
                 // These credentials are then used for subsequent requests to the feed.
                 // Note: If 'HttpClientHandler.PreAuthenticate' is set to true, this behavior might differ as 
-                // credentials would be sent preemptively with the initial request.
+                // credentials would be sent preemptively after the initial request.
 
                 // In this test, the expected number of requests is 4, but it turns out to be 5. This discrepancy occurs because 
                 // we initialize _credentials =  new HttpSourceCredentials() in the HttpSourceAuthenticationHandler constructor.
--- a/test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
+++ b/test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
@@ -155,10 +155,10 @@ namespace NuGet.Protocol.FuncTest
                 // In this test, 2 requests are sent from the test's perspective, and the previous paragraph's explanation will
                 // make you believe the server should see 4, but it turns out to be 5. This discrepancy occurs because 
                 // HttpSourceCredentials is only initialized with credentials if the PackageSource object has credentials set.
-                // In other words, if there are creds in nuget.config or the appropriate environment variable. However, if the
+                // In other words, if there are creds in NuGet.Config or the appropriate environment variable. However, if the
                 // package source uses a credential provider, the provider is not asked for credentials until after the first 401
                 // response. After the auth handler requests HttpClientHandler to make a second request, it will again make
-                // an unautheticated request, and then finally obtain the credentials and send an authenticated request from
+                // an unauthenticated request, and then finally obtain the credentials and send an authenticated request from
                 // what the server sees as the 3rd request.
 
                 // Request flow while accessing index.json
@@ -179,12 +179,18 @@ namespace NuGet.Protocol.FuncTest
                 server.Requests.Should().HaveCount(5);
                 // This should have been 2, but is 3 because of the reason mentioned above.
                 server.Requests.Count(RequestWithoutAuthorizationHeader).Should().Be(3);
-
-                server.CapturedCredentials.Should().HaveCount(2);
-                foreach (var creds in server.CapturedCredentials)
+                // Validate that the credentials sent with the 3rd request & 5th request are the ones received from the credential service.
+                foreach (var httpListenerRequest in server.Requests)
                 {
-                    Assert.Equal(expectedCredentials.UserName, creds.UserName);
-                    Assert.Equal(expectedCredentials.Password, creds.Password);
+                    string? authorization = httpListenerRequest.Headers["Authorization"];
+                    if (authorization != null)
+                    {
+                        var encodedCredentials = authorization.Substring("Basic ".Length).Trim();
+                        var credentials = Encoding.UTF8.GetString(Convert.FromBase64String(encodedCredentials));
+                        var creds = credentials.Split(':');
+                        Assert.Equal(expectedCredentials.UserName, creds[0]);
+                        Assert.Equal(expectedCredentials.Password, creds[1]);
+                    }
                 }
 
                 mockedCredentialService.Verify(x => x.GetCredentialsAsync(It.IsAny<Uri>(), It.IsAny<IWebProxy>(), It.IsAny<CredentialRequestType>(), It.IsAny<string>(), It.IsAny<CancellationToken>()), Times.Once);
@@ -239,7 +245,6 @@ namespace NuGet.Protocol.FuncTest
             private string? _baseUrl;
             private HttpListener? _httpListener;
             private List<HttpListenerRequest> _requests = new();
-            private List<NetworkCredential> _capturedCredentials { get; } = [];
             private ITestOutputHelper _output;
 
             public RequestCollectingServer(ITestOutputHelper output)
@@ -266,8 +271,6 @@ namespace NuGet.Protocol.FuncTest
                     return _requests;
                 }
             }
-
-            public IReadOnlyList<NetworkCredential> CapturedCredentials => _capturedCredentials;
 
             public void Start(int port)
             {
@@ -309,11 +312,6 @@ namespace NuGet.Protocol.FuncTest
                         else
                         {
                             context.Response.StatusCode = 200;
-
-                            var encodedCredentials = authorization.Substring("Basic ".Length).Trim();
-                            var credentials = Encoding.UTF8.GetString(Convert.FromBase64String(encodedCredentials));
-                            var split = credentials.Split(':');
-                            _capturedCredentials.Add(new NetworkCredential(userName: split[0], password: split[1]));
                         }
 
                         _output.WriteLine($"Got request for {context.Request.Url}. Auth: {authorization}");
--- a/test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
+++ b/test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
@@ -152,11 +152,14 @@ namespace NuGet.Protocol.FuncTest
                 // Note: If 'HttpClientHandler.PreAuthenticate' is set to true, this behavior might differ as 
                 // credentials would be sent preemptively after the initial request.
 
-                // In this test, the expected number of requests is 4, but it turns out to be 5. This discrepancy occurs because 
-                // we initialize _credentials =  new HttpSourceCredentials() in the HttpSourceAuthenticationHandler constructor.
-                // The HttpHandler pipeline operates under the assumption that credentials have already been set, although they have not.
-                // As a result, NuGet initially sends 2 requests to the server without credentials when accessing the private feed for the first time
-                // and then invokes the credential service.
+                // In this test, 2 requests are sent from the test's perspective, and the previous paragraph's explanation will
+                // make you believe the server should see 4, but it turns out to be 5. This discrepancy occurs because 
+                // HttpSourceCredentials is only initialized with credentials if the PackageSource object has credentials set.
+                // In other words, if there are creds in nuget.config or the appropriate environment variable. However, if the
+                // package source uses a credential provider, the provider is not asked for credentials until after the first 401
+                // response. After the auth handler requests HttpClientHandler to make a second request, it will again make
+                // an unautheticated request, and then finally obtain the credentials and send an authenticated request from
+                // what the server sees as the 3rd request.
 
                 // Request flow while accessing index.json
                 // NuGet -> Server (No Credentials) - 1st request
--- a/test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
+++ b/test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
@@ -179,7 +179,6 @@ namespace NuGet.Protocol.FuncTest
                 server.Requests.Should().HaveCount(5);
                 // This should have been 2, but is 3 because of the reason mentioned above.
                 server.Requests.Count(RequestWithoutAuthorizationHeader).Should().Be(3);
-                server.Requests.Any(r => r.Url!.OriginalString == serviceIndexUrl).Should().BeTrue();
 
                 server.CapturedCredentials.Should().HaveCount(2);
                 foreach (var creds in server.CapturedCredentials)
--- a/test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
+++ b/test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
@@ -98,11 +98,112 @@ namespace NuGet.Protocol.FuncTest
             }
         }
 
+        [Fact]
+        public async Task GetAsync_GetPackageServiceIndex_EveryRequestForCredentialsInvokesCacheFirstAndCredentialProvidersIfNeeded_SucceedsAsync()
+        {
+            // Arrange
+            var portReserver = new PortReserver();
+            await portReserver.ExecuteAsync(async (port, cancellationToken) =>
+            {
+                var mockedCredentialService = new Mock<ICredentialService>();
+                var expectedCredentials = new NetworkCredential("user", "password1");
+
+                var server = new RequestCollectingServer(_output);
+                server.Start(port);
+                string serviceIndexUrl = server.BaseUrl + "v3/index.json";
+                string packageUrl = server.BaseUrl + "v3/flatcontainer/packageid/index.json";
+
+                try
+                {
+                    var packageSource = new PackageSource(server.BaseUrl + "v3/index.json", "test");
+                    var repository = Repository.Factory.GetCoreV3(packageSource);
+
+                    SetupCredentialServiceMock(mockedCredentialService, expectedCredentials, packageSource);
+                    HttpHandlerResourceV3.CredentialService = new Lazy<ICredentialService>(() => mockedCredentialService.Object);
+
+                    var httpSourceResource = await repository.GetResourceAsync<HttpSourceResource>();
+                    var source = httpSourceResource.HttpSource;
+
+                    using var sourceCacheContext = new SourceCacheContext()
+                    {
+                        DirectDownload = true,
+                        NoCache = true,
+                    };
+                    Mock<ILogger> logger = new();
+                    HttpSourceCacheContext httpSourceCacheContext = HttpSourceCacheContext.Create(sourceCacheContext, isFirstAttempt: true);
+
+                    // Act                    
+                    var request = new HttpSourceCachedRequest(serviceIndexUrl, "1", httpSourceCacheContext);
+                    _ = await source.GetAsync(request, ProcessResponse, logger.Object, cancellationToken);
+
+                    request = new HttpSourceCachedRequest(packageUrl, "2", httpSourceCacheContext);
+                    _ = await source.GetAsync(request, ProcessResponse, logger.Object, cancellationToken);                    
+                }
+                finally
+                {
+                    server.Stop();
+                }
+
+                // Assert
+                server.Requests.Should().HaveCount(5);
+                server.Requests.Count(RequestWithoutAuthorizationHeader).Should().Be(3);
+                server.Requests.Any(r => r.Url!.OriginalString == serviceIndexUrl).Should().BeTrue();
+
+                server.CapturedCredentials.Should().HaveCount(2);
+                foreach (var creds in server.CapturedCredentials)
+                {
+                    Assert.Equal(expectedCredentials.UserName, creds.UserName);
+                    Assert.Equal(expectedCredentials.Password, creds.Password);
+                }
+
+                mockedCredentialService.Verify(x => x.GetCredentialsAsync(It.IsAny<Uri>(), It.IsAny<IWebProxy>(), It.IsAny<CredentialRequestType>(), It.IsAny<string>(), It.IsAny<CancellationToken>()), Times.Once);
+                mockedCredentialService.Verify(x => x.TryGetLastKnownGoodCredentialsFromCache(It.IsAny<Uri>(), It.IsAny<bool>(), out It.Ref<ICredentials>.IsAny), Times.Once);
+
+                // ExecuteAsync returns Task<T>, so need to return something to give it a <T>.
+                return (object?)null;
+            },
+            CancellationToken.None);
+
+            static void SetupCredentialServiceMock(Mock<ICredentialService> mockedCredentialService, NetworkCredential expectedCredentials, PackageSource packageSource)
+            {
+                NetworkCredential? cachedCredentials = default;
+                mockedCredentialService.SetupGet(x => x.HandlesDefaultCredentials).Returns(true);
+                // Setup GetCredentialsAsync mock
+                mockedCredentialService
+                    .Setup(x => x.GetCredentialsAsync(packageSource.SourceUri, It.IsAny<IWebProxy>(), CredentialRequestType.Unauthorized, It.IsAny<string>(), It.IsAny<CancellationToken>()))
+                    .ReturnsAsync(() =>
+                    {
+                        cachedCredentials = expectedCredentials;
+                        return cachedCredentials;
+                    });
+                // Setup TryGetLastKnownGoodCredentialsFromCache mock
+                mockedCredentialService
+                    .Setup(x => x.TryGetLastKnownGoodCredentialsFromCache(packageSource.SourceUri, It.IsAny<bool>(), out It.Ref<ICredentials>.IsAny))
+                    .Returns((Uri sourceUri, bool isProxyRequest, out ICredentials? outCredentials) =>
+                    {
+                        outCredentials = cachedCredentials;
+                        return outCredentials != null;
+                    });
+            }
+
+            static bool RequestWithoutAuthorizationHeader(HttpListenerRequest request)
+            {
+                string? value = request.Headers["Authorization"];
+                return string.IsNullOrEmpty(value);
+            }
+
+            static Task<string> ProcessResponse(HttpSourceResult result)
+            {
+                return Task.FromResult(string.Empty);
+            }
+        }
+
         private class RequestCollectingServer
         {
             private string? _baseUrl;
             private HttpListener? _httpListener;
             private List<HttpListenerRequest> _requests = new();
+            private List<NetworkCredential> _capturedCredentials { get; } = [];
             private ITestOutputHelper _output;
 
             public RequestCollectingServer(ITestOutputHelper output)
@@ -129,6 +230,8 @@ namespace NuGet.Protocol.FuncTest
                     return _requests;
                 }
             }
+
+            public IReadOnlyList<NetworkCredential> CapturedCredentials => _capturedCredentials;
 
             public void Start(int port)
             {
@@ -170,6 +273,11 @@ namespace NuGet.Protocol.FuncTest
                         else
                         {
                             context.Response.StatusCode = 200;
+
+                            var encodedCredentials = authorization.Substring("Basic ".Length).Trim();
+                            var credentials = Encoding.UTF8.GetString(Convert.FromBase64String(encodedCredentials));
+                            var split = credentials.Split(':');
+                            _capturedCredentials.Add(new NetworkCredential(userName: split[0], password: split[1]));
                         }
 
                         _output.WriteLine($"Got request for {context.Request.Url}. Auth: {authorization}");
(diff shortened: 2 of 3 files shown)


src/NuGet.Core/NuGet.Protocol/HttpSource/HttpSourceAuthenticationHandler.cs
```diff
@@ -198,7 +203,13 @@ private async Task<ICredentials> AcquireCredentialsAsync(HttpStatusCode statusCo
                         _packageSource.Source);
                 }
 
-                var promptCredentials = await PromptForCredentialsAsync(
+                ICredentials promptCredentials = default;
+                if (!areLastKnownGoodCredentialsTried)
+                {
+                    _ = _credentialService.TryGetLastKnownGoodCredentialsFromCache(uri: _packageSource.SourceUri, isProxy: false, out promptCredentials);
```
kartheekp-ms, line 214: I have hardcoded, `isProxy: false` because the if statement above sets the type to either `Unauthorized` or `Forbidden`. Hence it can't be [CredentialRequestType.Proxy](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Configuration/Credential/CredentialRequestType.cs#L8).

kartheekp-ms, line 214: The implementation of [CredentialService.TryGetLastKnownGoodCredentialsFromCache](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Credentials/CredentialService.cs#L160-L183) includes a for loop that checks whether the credentials for the given URL were successfully acquired previously. If true, it returns the cached credentials to the caller; otherwise, it returns false.

zivkan: This would be more useful as a comment, so that anyone reading the code after this PR is merged, can get the same context.

src/NuGet.Core/NuGet.Protocol/HttpSource/HttpSourceAuthenticationHandler.cs
```diff
@@ -154,7 +158,8 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
             }
         }
 
-        private async Task<ICredentials> AcquireCredentialsAsync(HttpStatusCode statusCode, Guid credentialsVersion, ILogger log, CancellationToken cancellationToken)
+        private async Task<ICredentials> AcquireCredentialsAsync(HttpStatusCode statusCode, Guid credentialsVersion, ILogger log,
+            bool areLastKnownGoodCredentialsTried, CancellationToken cancellationToken)
```
zivkan: nitpick: I don't remember if our coding guidelines say anything about this scenario, but I don't remember many cases of method signatures being manually line wrapped. Either one long line (I'm not a big fan, it's hard to read when it scrolls off the edge of my editor), or one parameter per line.

I don't think this type of "manual" line wrapping is bad, but it doesn't feel like it's a common pattern in our code.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpSource/HttpSourceAuthenticationHandlerTests.cs
```diff
@@ -312,6 +312,177 @@ public async Task SendAsync_WithWrongCredentials_StopsRetryingAfter3Times()
...
+            {
+                retryCount++;
+                var credentials = clientHandler.Credentials.GetCredential(packageSource.SourceUri, "basic");
+                return HandleRequest(retryCount, credentials, credentialsReturnedByAProvider);
+            });
+
+            // Act and Assert in a loop
+            for (int i = 0; i < 5; i++)
+            {
+                await ExecuteRequestAsync(packageSource, clientHandler, credentialService, innerHandler);
```
zivkan: I think this is a little "dangerous", because it's being tested a significantly different "way" than what the product code does.

In the product code, each `SourceRepository` will create a single instance of the `HttpSourceAuthenticationHandler` class, which will be re-used for every request. This means that the class's fields will be stateful.

However, by using this helper method that creates a new instance every iteration, it's treating the class as stateless, which I think is a significant risk given the second test is for an expiring credential scenario, and the class has a field named `_authStates`, which sounds relevant.

I think these tests should create a single instance of the auth handler, and then call `SendAsync` multiple times, instead. Honestly, I'm not a fan of this method's `SendAsync` method either, since it creates a new `HttpClient` instance every call. Personally, I'd feel more comfortable if tests that send more than 1 HTTP request created a single `HttpClient` instance (and therefore a single message handler instance), and then called `SendAsync`, or `GetAsync` multiple times on that instance.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpSource/HttpSourceAuthenticationHandlerTests.cs
```diff
@@ -312,6 +312,177 @@ public async Task SendAsync_WithWrongCredentials_StopsRetryingAfter3Times()
...
+
+            // In this test, although we are explicitly making 5 requests to the feed, 
+            // internally a total of 10 requests are made due to the authentication behavior described near 'retryCount' variable.
+            // For each request to the private feed, the first attempt results in a 401 Unauthorized response,
+            // followed by a second attempt with the acquired credentials. This effectively doubles the number of 
+            // actual HTTP requests sent during the test.
+            Assert.Equal(10, retryCount);
+
+            Mock.Get(credentialService).Verify(x => x.TryGetLastKnownGoodCredentialsFromCache(packageSource.SourceUri, It.IsAny<bool>(), out It.Ref<ICredentials>.IsAny), Times.Exactly(5));
+            Mock.Get(credentialService).Verify(x => x.GetCredentialsAsync(packageSource.SourceUri, It.IsAny<IWebProxy>(), CredentialRequestType.Unauthorized, It.IsAny<string>(), It.IsAny<CancellationToken>()), Times.Once);
```
zivkan: My comment is very opinionated, but I'm not a fan of using `Mock.Of<T>()` and `Mock.Get(obj)`. I prefer creating instances of `new Mock<T>()`, and then you don't need to do anything "special" to get get mock instance for setup or verification.

FWIW, using `Mock.Get(obj)` also has a small perf cost, which I think when done multiple times in a test, over thousands of tests in a test suite, might contribute to a measurable duration increase. Although I don't have data on that, but of course measuring actual impact on a test suite requires an effort to change all instances, which is too much effort to validate.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpSource/HttpSourceAuthenticationHandlerTests.cs
```diff
@@ -312,6 +312,177 @@ public async Task SendAsync_WithWrongCredentials_StopsRetryingAfter3Times()
...
+                var credentials = clientHandler.Credentials.GetCredential(packageSource.SourceUri, "basic");
+                var response = HandleRequest(retryCount, ref isUnauthorizedResponse, credentials, initialCredentials);
+
+                // Change credentials after 3 successful requests to the feed to mimic the scenario of expired credentials.
+                if (retryCount == 8)
+                {
+                    // Verify that response is 401 Unauthorized
+                    Assert.Equal(HttpStatusCode.Unauthorized, response.StatusCode);
+                    initialCredentials = new NetworkCredential("user", "password2");
+                    Mock.Get(credentialService).Setup(x => x.GetCredentialsAsync(packageSource.SourceUri, It.IsAny<IWebProxy>(), CredentialRequestType.Unauthorized, It.IsAny<string>(), It.IsAny<CancellationToken>())).ReturnsAsync(initialCredentials);
```
zivkan: Mocks can be setup to call a `Func`, so you wouldn't need to change the setup of the mock when input data changes. In fact, you already did this for the setup for `TryGetLastKnownGoodCredentialsFromCache`.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpSource/HttpSourceAuthenticationHandlerTests.cs
```diff
@@ -312,6 +312,177 @@ public async Task SendAsync_WithWrongCredentials_StopsRetryingAfter3Times()
...
+                }
+
+                return response;
+            });
+
+            // Act and Assert in a loop
+            for (int i = 0; i < 5; i++)
+            {
+                await ExecuteRequestAsync(packageSource, clientHandler, credentialService, innerHandler);
```
zivkan: Just thinking out loud, but I have three concerns with these tests. However, I'm worried I might be overly idealistic/perfectionist, so if you think that changing the tests will be non-trivial effort, perhaps get feedback for other people in the team to see if my concerns are shared.

Firstly, understanding how the credential change works (in the second test) is non-linear, making it harder to follow. Secondly, the act & assert code is intermingled. Thirdly, since this is about authentication, where in the product code there's stuff automatically handled by the `HttpClientHandler`. But these tests override the auth handler's `InnerHandler` to not use `HttpClientHandler`. Is there a risk that future changes will change product behaviour, but these tests give a false sense of security?

About the act & assert logic being non-linear, I'm wondering if it would be clearer to people reading the test if you execute 3 requests here, then explicitly change the credentials, then execute another 2 requests. That way changing the expected credentials isn't a side effect handled by two separate `if (retryCount == 8)` statements, but the act section of the test would be something like:
```cs
// Act
for (int i = 0; i < 3; i++)
{
   await ExecuteRequestAsync();
}

expectedCredentials = new NetworkCredentials("user", "password2");

for (int i = 0; i < 2; i++)
{
  await ExecuteRequestAsync();
}
```

It's more lines of code, but I think this makes it clear to readers what the test execution/logic flow is, and might help understand why makes 5 requests, rather than just 2.

Although I think the first test (that doesn't change creds) probably only "needs" 2 requests, and the test that does change creds probably only needs 1 request before the change and 1 after. but I'm defiantly being overly idealistic here. As long as the tests aren't too slow, it doesn't matter. Anyway, the point is that if there are two behaviours that want to be tested (one: creds get cached. two: changing creds will eventually use the new creds), then two tests that make 2 requests each is the minimum needed to verify that behaviour. Technically, the cred change can't be tested without the product code making a request without the old, cached creds, so one test could be used to validate both behaviours. On one hand different tests, with different names can help understand the focus of the test. On the other hand, if the cred change test fails when caching fails, then the benefit of well named, single scenario tests is lost. Again, I realize this paragraph is overly pedantic, it's just how my brain works. Please don't get too hung up on this paragraph, the other points in this comment are much more important.


Next, on the subjects of both intermingling of act & assert code, and also not using `HttpClientHandler` in the tests, I think porting the test to [use the infrastructure I created for testing PreAuthenticate](https://github.com/NuGet/NuGet.Client/blob/0adf7ac2d046bbc6d7e8db29ff82b3b2f8fc5f14/test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs#L101) will help. `RequestCollectingServer` will need to be updated to support `ExpectedCredentials`, so that it can be changed after one or more requests. It might also need a change to the `_requests` list, if additional information is needed, but you can get the request's `Authenticate` header to validate the username & password, which I think is all you need.

Anyway, what I want to raise awareness about is that whether we eventually [reimplement PreAuthenticate or just set creds in `HttpClinetHandler.DefaultHeaders`](https://github.com/NuGet/Home/issues/12962), by using `HttpClientHandler` in the test, then the test will be validating the real product behaviour. Currently the new tests in the first version of this PR are creating a `LamdaMessageHandler`, which is imitating some `HttpClientHandler` behaviour, but if it's not imitating the real handler how it actually works, then the tests might not catch regressions. The comments in the test code do explain to some extent, but code exists to make testing real HTTP requests exists. Even if the `LamdaMessageHandler` is imitating `HttpClientHandler` sufficiently accurately, developers reading the test who are not deeply knowledgeable on `HttpClient`, `HttpClientHandler`, and the `MessageHandler` pipeline, and also how NuGet's auth works, they might not fully appreciate what the HTTP requests the actual product sends to HTTP servers look like. With the `HttpListener`, it's possible to debug and inspect the `HttpRequestMessage`'s headers to see what's actually getting sent. Unfortunately it's at the cost of performance, but it's a trade off of perf vs risk/benefit. Even so, making HTTP requests over localhost should still be sub-millisecond, so it's hopefully not unreasonably slow.

But again, maybe I'm being overly idealistic. If you don't like my suggestions, it can be worthwhile getting feedback from others in the team.

src/NuGet.Core/NuGet.Protocol/HttpSource/HttpSourceAuthenticationHandler.cs
kartheekp-ms: Fixed in the latest commit.

src/NuGet.Core/NuGet.Protocol/HttpSource/HttpSourceAuthenticationHandler.cs
kartheekp-ms: I changed the method signature to have one parameter per line in the latest commit.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpSource/HttpSourceAuthenticationHandlerTests.cs
kartheekp-ms: In the latest commit, I modified code to invoke the HTTP requests on a single HttpClient instance.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpSource/HttpSourceAuthenticationHandlerTests.cs
kartheekp-ms: Fixed in the latest commit to use instance of `new Mock<T>()`

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpSource/HttpSourceAuthenticationHandlerTests.cs
kartheekp-ms: Fixed in the latest commit. Please review again and let me know your feedback.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpSource/HttpSourceAuthenticationHandlerTests.cs
kartheekp-ms: Thank you for the detailed feedback, Andy. I agree that comprehending this test is not straightforward, particularly with `retryCount == 8` checks, which can be challenging to follow.

In the latest commit, I've tried to address first part of the above feedback by incorporating your suggestion i.e. Initially, the test invokes 3 requests to the feed, then changes the credentials to simulate an expired token scenario. Following this, it makes 2 more requests to the feed, which will only succeed if the new credentials are correctly passed.

I will respond to the second part of the comment that is related to PreAuthenticate as soon as possible.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpSource/HttpSourceAuthenticationHandlerTests.cs
```diff
@@ -312,6 +312,170 @@ public async Task SendAsync_WithWrongCredentials_StopsRetryingAfter3Times()
...
+            Assert.Equal(10, retryCount);
+
+            credentialServiceMock.Verify(x => x.TryGetLastKnownGoodCredentialsFromCache(packageSource.SourceUri, It.IsAny<bool>(), out It.Ref<ICredentials>.IsAny), Times.Exactly(5));
+            credentialServiceMock.Verify(x => x.GetCredentialsAsync(packageSource.SourceUri, It.IsAny<IWebProxy>(), CredentialRequestType.Unauthorized, It.IsAny<string>(), It.IsAny<CancellationToken>()), Times.Once);
+
+            static HttpResponseMessage HandleRequest(int retryCount, NetworkCredential credentials, NetworkCredential credentialsReturnedByAProvider)
+            {
+                if (retryCount % 2 == 1)
+                {
+                    return new HttpResponseMessage(HttpStatusCode.Unauthorized);
```
zivkan, line 380: I don't know if it's obvious to people reading the test why every second request should respond with an unauthorized response. I remember the previous iteration of this PR had a lot of comments. I think some of those comments explained that HttpClient sends unauthenticated requests and then only sends creds after getting a 401 response. I think people who don't have this background knowledge may be confused why it's doing this, and therefore why the retryCount is double the number of requests sent.

My gut reaction is to stop discarding the `HttpRequestMessage` in the `LambdaMessageHandler`, and then parse the Authorization header. However, I need to keep reminding myself that `HttpClientHandler` does that, and this unit test bypasses `HttpClinetHandler`. So again, I feel like porting this unit test to a func test that uses a HttpListener would allow the test code to be clearer, but slower to execute. As I mentioned in yesterday's feedback, doing this would also allow removing the `== 8` check in the second test's `HandleRequest` method, as you'd be able to parse the actual header, showing the reader that this is how `HttpClient` really behaves.

anyway, my comment is just a nitpick based on an assumption that others reading this code won't understand why every second request should have an unauthorized response. If this background knowledge is more widespread than I assume, it's not a problem.  But a comment is a quick and easy mitigation.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpSource/HttpSourceAuthenticationHandlerTests.cs
```diff
@@ -312,6 +312,170 @@ public async Task SendAsync_WithWrongCredentials_StopsRetryingAfter3Times()
...
+                .ReturnsAsync(() => initialCredentials);
+
+            // Setup TryGetLastKnownGoodCredentialsFromCache mock
+            credentialServiceMock
+                .Setup(x => x.TryGetLastKnownGoodCredentialsFromCache(packageSource.SourceUri, It.IsAny<bool>(), out It.Ref<ICredentials>.IsAny))
+                .Returns((Uri sourceUri, bool isProxyRequest, out ICredentials outCredentials) =>
+                {
+                    outCredentials = retryCount == 1 ? null : initialCredentials;
+                    return outCredentials != null;
+                });
```
zivkan: I think if you create a variable `NetworkCredentials? cachedCredentials = null`, then `GetCredentialsAsync`'s callback can be changed to

```cs
() =>
{
  cachedCredentials = expectedCredentials;
  return cachedCredentials;
}
```

and the callback for `TryGetLastKnownGoodCredentialsFromCache` can be changed to:
```cs
{
  outCredentaisl = cachedCredentials;
  return outCdredentials != null;
}
```

and after the `expectedCredentials` are changed in the test below, you don't need to re-setup these two methods.

But more importantly, this also allows you to remove the confusing `retryCount == 8` check from the cred service setup.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpSource/HttpSourceAuthenticationHandlerTests.cs
```diff
@@ -312,6 +312,170 @@ public async Task SendAsync_WithWrongCredentials_StopsRetryingAfter3Times()
...
+            });
+
+            var handler = new HttpSourceAuthenticationHandler(packageSource, clientHandler, credentialServiceMock.Object)
+            {
+                InnerHandler = innerHandler
+            };
+
+            using var client = new HttpClient(handler);
+
+            // Send 3 requests to the feed which succeed only if initial credentials are used.
```
zivkan, line 445:
```suggestion
            // Act
            // Send 3 requests to the feed which succeed only if initial credentials are used.
```

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpSource/HttpSourceAuthenticationHandlerTests.cs
```diff
@@ -312,6 +312,170 @@ public async Task SendAsync_WithWrongCredentials_StopsRetryingAfter3Times()
...
+                .Returns((Uri sourceUri, bool isProxyRequest, out ICredentials outCredentials) =>
+                {
+                    outCredentials = retryCount == 8 ? initialCredentials : newCredentials;
+                    return true;
+                });
+
+            // Send 2 more requests to the feed which succeed only if new credentials are used.
+            await SendAsync(client, 2);
+
+            Assert.Equal(11, retryCount);
```
zivkan, line 461:
```suggestion
            // Assert
            Assert.Equal(11, retryCount);
```

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpSource/HttpSourceAuthenticationHandlerTests.cs
```diff
@@ -312,6 +312,170 @@ public async Task SendAsync_WithWrongCredentials_StopsRetryingAfter3Times()
...
+                    return new HttpResponseMessage(HttpStatusCode.OK);
+                }
+            }
+        }
+
+        private static async Task SendAsync(HttpClient client, int count)
+        {
+            for (int i = 0; i < count; i++)
+            {
+                await client.SendAsync(new HttpRequestMessage(HttpMethod.Get, "http://foo"), CancellationToken.None);
```
zivkan: major nitpick (as in I realize this comment is not very impactful): it seems kind of weird that the tests use package source URL `http://package.source.test/`, but this request uses `http://foo`. It's this `HttpClient.SendAsync` call that is most important to use a domain name ending in `.test` to ensure it can't accidentally make HTTP requests over the real internet (if the `HttpClient` was configured with `HttpClientHandler` in its middleware, which this method doesn't know is true or false, and if the domain name actually resolves to an IP address). Also, trying to put myself in the mindset of an inexperienced dev, maybe there's a question if the request URL being different to the package source URL is a potential problem?

Anyway, definitely not worth changing if there's no other changes made. But if you end up pushing other changes, maybe consider changing this too. Not a big deal at all.

zivkan (dismissed):
Thanks for the changes. I feel a lot better now that a single `HttpSourceAuthenticationHandler` instance is used for all requests within each test.

My remaining comments are basically just "I would have done things differently", but that doesn't mean you have to do things the same way I would have. Just something to consider.

zivkan (dismissed):

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpSource/HttpSourceAuthenticationHandlerTests.cs
kartheekp-ms: There was a CI failure. Hence, I had to push a change to fix it. I made the following couple of changes in the latest commit.

1. Added a string constant for "http://package.source.test/" URL and referred it in all the tests.
2. Replaced all the references to "http://foo" with package source url constant added in this PR.

All the test passed locally. Hence pushed the changes.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpSource/HttpSourceAuthenticationHandlerTests.cs
kartheekp-ms: > Anyway, what I want to raise awareness about is that whether we eventually [reimplement PreAuthenticate or just set creds in HttpClinetHandler.DefaultHeaders](https://github.com/NuGet/Home/issues/12962), by using HttpClientHandler in the test, then the test will be validating the real product behaviour.

While I was adding tests, I reviewed the ones you implemented for the [PreAuthenticate work](https://github.com/NuGet/NuGet.Client/pull/5157). At that time I was under the impression that mocking [CredentialService](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpSourceAuthenticationHandler.cs#L26) is challenging due to its depth in the call stack. However, after reconsidering it based on your PR feedback, I discovered that [HttpHandlerResourceV3.CredentialService](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpHandlerResourceV3.cs#L40) is a static property. At this moment, I'm uncertain if having CredentialService as a static property on the HttpHandlerResourceV3 class is the best design choice. I managed to create a functional test by using the infrastructure you set up for the PreAuthenticate work and mocking `HttpHandlerResourceV3.CredentialService` in the tests.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpSource/HttpSourceAuthenticationHandlerTests.cs
kartheekp-ms: > I remember the previous iteration of this PR had a lot of comments. I think some of those comments explained that HttpClient sends unauthenticated requests and then only sends creds after getting a 401 response.

Sorry for the confusion. The comments have been restored. I made a mistake during the code push, but it has now been addressed.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/HttpSource/HttpSourceAuthenticationHandlerTests.cs
kartheekp-ms: Thank you for the feedback. I've taken your suggestion into account and have pushed a commit accordingly.

test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
```diff
@@ -98,11 +98,146 @@ static Task<string> ProcessResponse(HttpSourceResult result)
...
+                    server.Stop();
+                }
+
+                // Assert
+
+                // Each attempt to access a private feed initially receives a 401 Unauthorized response.
+                // Following the 401 response, NuGet attempts to acquire the necessary credentials.
+                // These credentials are then used for subsequent requests to the feed.
+                // Note: If 'HttpClientHandler.PreAuthenticate' is set to true, this behavior might differ as 
+                // credentials would be sent preemptively with the initial request.
```
zivkan:
```suggestion
                // credentials would be sent preemptively after the initial request.
```

test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
```diff
@@ -98,11 +98,146 @@ static Task<string> ProcessResponse(HttpSourceResult result)
...
+
+                // In this test, the expected number of requests is 4, but it turns out to be 5. This discrepancy occurs because 
+                // we initialize _credentials =  new HttpSourceCredentials() in the HttpSourceAuthenticationHandler constructor.
+                // The HttpHandler pipeline operates under the assumption that credentials have already been set, although they have not.
+                // As a result, NuGet initially sends 2 requests to the server without credentials when accessing the private feed for the first time
+                // and then invokes the credential service.
```
zivkan:
```suggestion
                // In this test, 2 requests are sent from the test's perspective, and the previous paragraph's explanation will
                // make you believe the server should see 4, but it turns out to be 5. This discrepancy occurs because 
                // HttpSourceCredentials is only initialized with credentials if the PackageSource object has credentials set.
                // In other words, if there are creds in nuget.config or the appropriate environment variable. However, if the
                // package source uses a credential provider, the provider is not asked for credentials until after the first 401
                // response. After the auth handler requests HttpClientHandler to make a second request, it will again make
                // an unautheticated request, and then finally obtain the credentials and send an authenticated request from
                // what the server sees as the 3rd request.
```

I think it might be worthwhile for the comments to make it clearer that this is only an issue when using credential providers, not nuget.config (or env var) credentials. Also, maybe I just didn't read the comment carefully enough, but my initial read made me think there was a bug (because of the "operates under the assumption ... although they have not"), so maybe my suggestion reduces risk of misunderstanding, although my suggestion is even more lines of text 😞 

Also, I wonder if there's a reason we can't (some time in the future) rearchitect cred providers to happen within a class implementing `ICredentials` (for example, `HttpSourceCredentials`), so that when a cred provider is used, the server won't see a second pointless unauthenticated request. Unfortunately it would mean we don't have access to the `HttpRequest` or the failed `HttpResponse` objects. I don't know if the cred provider needs any info from either of these objects. Is it worthwhile to create a tracking issue to investigate this idea?  Unfortunately my gut feeling is that having access to the request and response objects are too useful, so I'm not sure this is really a feasible idea.

test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
```diff
@@ -170,6 +307,11 @@ public void HandleHttpRequest(IAsyncResult result)
                         else
                         {
                             context.Response.StatusCode = 200;
+
+                            var encodedCredentials = authorization.Substring("Basic ".Length).Trim();
+                            var credentials = Encoding.UTF8.GetString(Convert.FromBase64String(encodedCredentials));
+                            var split = credentials.Split(':');
+                            _capturedCredentials.Add(new NetworkCredential(userName: split[0], password: split[1]));
```
zivkan: Can the test use `Requests[i].Header["Authentication"]` instead or creating a second list?  This way tests that don't validate the credentials don't pay the cost of parsing the credential. Not that this minor perf cost in func tests really matters. I just feel it's weird that test infrastructure is spending effort on something that potentially only a subtest of tests using it will need.

If you think it's particularly useful to keep this captured credentials list, then I think that unauthenticated requests should insert `null`, so that `CapturedCredentials[i]` always corresponds to `Requests[i]`.

I'm thinking about it from an API design point of view, and what would make sense to other people who might want to write new tests and re-use the `CapturedCredentials` API.

test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
```diff
@@ -98,11 +98,146 @@ static Task<string> ProcessResponse(HttpSourceResult result)
             }
         }
 
+        [Fact]
+        public async Task GetAsync_GetPackageServiceIndex_EveryRequestForCredentialsInvokesCacheFirstAndCredentialProvidersIfNeeded_SucceedsAsync()
```
zivkan, line 102: since this PR is about expired credentials changing, I think it would be better for this test to validate that scenario. Doing so will naturally test the cached creds that this test is currently doing, so will only expand test coverage.

One issue with the way that the test is currently written, is that you're setting up the cred service provider in a separate method, which means that the captured variable is in a scope that the test itself can't modify: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/lambda-expressions#capture-of-outer-variables-and-variable-scope-in-lambda-expressions

If you "inline" that method, then it will only be an extra few lines to test the "creds have changed" scenario that this PR is all about.

test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
```diff
@@ -98,11 +98,146 @@ static Task<string> ProcessResponse(HttpSourceResult result)
...
+
+                // Request flow while retrieving package information
+                // NuGet -> Server (No Credentials) - 4th request
+                // Server -> NuGet (401 Unauthorized)
+                // NuGet -> Server (Sends HttpClientHandler.Credentials) - 5th request
+                // Server -> NuGet (200 OK)
+                server.Requests.Should().HaveCount(5);
+                // This should have been 2, but is 3 because of the reason mentioned above.
+                server.Requests.Count(RequestWithoutAuthorizationHeader).Should().Be(3);
+                server.Requests.Any(r => r.Url!.OriginalString == serviceIndexUrl).Should().BeTrue();
```
zivkan: I don't understand why this is validating that the url is the service index URL for at least 1 of the 5 requests, but then doesn't validate the URL for any other request.

Personally, I'd remove the line, and not assert any of the request URLs, because it doesn't feel relevant to me, for the details that the test is supposed to be validating against regressions (credentials to the package source are cached, and possibly get refreshed if expired).

test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
kartheekp-ms: I accepted your comment suggestion. It is much clearer than my version.

> Also, I wonder if there's a reason we can't (some time in the future) rearchitect cred providers to happen within a class implementing ICredentials (for example, HttpSourceCredentials), so that when a cred provider is used, the server won't see a second pointless unauthenticated request. Unfortunately it would mean we don't have access to the HttpRequest or the failed HttpResponse objects. I don't know if the cred provider needs any info from either of these objects. Is it worthwhile to create a tracking issue to investigate this idea? Unfortunately my gut feeling is that having access to the request and response objects are too useful, so I'm not sure this is really a feasible idea.



I am happy to create a tracking technical debt issue to rethink how credential providers have been implemented to address following problems.

1. Credential Service is a static variable in [HttpHandlerResourceV3.cs](https://github.com/NuGet/NuGet.Client/blob/15b4da06d4cb1a8e1bd2a0e710520b68ce0a6915/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpHandlerResourceV3.cs#L40) for injecting the credential providers. I guess there are better ways to inject the dependencies in terms of API design. We know this has impacted MSBuild server feature during .NET 7 RC1/2 timeframe.
2. Preferable avoid pointless request to the feed because of the way how `HttpSourceCredentials` has been implemented.

> I don't know if the cred provider needs any info from either of these objects i.e. HttpRequest  and HttpResponse.

In the current implementation, HttpSourceAuthenticationHandler sends the [`CredentialRequestType`](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Configuration/Credential/CredentialRequestType.cs) to the `CredentialService` which is derived from the [`HttpResponse.StatusCode`](https://github.com/NuGet/NuGet.Client/blob/15b4da06d4cb1a8e1bd2a0e710520b68ce0a6915/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpSourceAuthenticationHandler.cs#L184-L199). I am not sure if credential providers have different behavior based on the response status code. But looking at the implementation, it looks like they need `CredentialRequestType` only to check if the [request involves a proxy or not](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/VsCredentialProviderAdapter.cs#L55).

test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
kartheekp-ms: I removed this assert statement. Fixed in the latest commit.

test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
kartheekp-ms: > Can the test use Requests[i].Header["Authentication"] instead or creating a second list?

Yup. I removed the CapturedCredentials collection from the API. I followed your suggestion use Requests[i].Header["Authentication"] to obtain the credentials from the Requests collection. Test passed locally, fixed in the latest commit.

test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
kartheekp-ms: Yes, I had the same test scenario in mind. However, I realized the credential service is invoked only once because the HttpClientAuthenticationHandler sends the credentials to the server upon receiving a 401 challenge, once they are set via the [HttpSourceCredentials.Credentials](https://github.com/NuGet/NuGet.Client/blob/15b4da06d4cb1a8e1bd2a0e710520b68ce0a6915/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpSourceAuthenticationHandler.cs#L213) property. That's why in this test, the `CredentialService.TryLastKnownGoodCredential()` method is invoked only once https://github.com/NuGet/NuGet.Client/pull/5563/files#diff-12afadc4686902895930ca4ad9a177f6539fda8576adee33c07ca9a4698334bfR202 even though there are 2 requests sent to the feed with authorization header as explained in https://github.com/NuGet/NuGet.Client/pull/5563/files#diff-12afadc4686902895930ca4ad9a177f6539fda8576adee33c07ca9a4698334bfR177 comment.

I am unsure if there is any other scenario that invalidates the HttpClientHandler.Credentials, which would force the HttpSourceAuthenticationHandler to try the NuGet credential cache and then invoke Credential Providers.

zivkan: I only spent a few seconds reading this, and I have other things on my mind at the moment, so I can't fully wrap my head around the details of what you're saying about the credential service, sorry.

What I'm thinking is that I'm not aware of any test that validates when a credential expires, that new creds are obtained from the cred provider, and then the new creds are eventually sent to the server. So, a func test that does that sounds useful to me, and the details of the number of times `TryLastKnownGoodCredential` is of secondary concern, IMO.

In fact, I think the point of scenario is to ensure that `CredService.GetCredentailsAsync` is not called more than once per username/password combo. That's the goal of credential caching, right?  So, if the test validates the number of calls to `GetCredentialsAsync`, it doesn't really matter if `TryLastKnownGoodCredential` is used, or if `HttpClientHandler` gets it directly from `HttpSourceCredential`. As long as the cred service doesn't potentially create a blocking popup waiting for interactive login, then the scenario is satisfied, right?

Anyway, just throwing the idea out there.  Even if we don't add a "perfect" func test, we're still better off than before, so I shouldn't block this PR 😁

zivkan (approved):

test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/AuthenticationHandlerTests.cs
kartheekp-ms: > What I'm thinking is that I'm not aware of any test that validates when a credential expires, that new creds are obtained from the cred provider, and then the new creds are eventually sent to the server

As you are aware, I added a unit test to cover this scenario in this PR. I agree that there is no functional test covering this scenario. What I am trying to convey is that at this moment, I don't know of any scenario where HttpSourceCredentials would be invalid while the credentials we have in the cache will be valid. There are high chances that the credentials in the CredentialService's cache are the same as the HttpSourceCredentials. This means there could be an unnecessary 401 request to the server because if HttpSourceCredentials are invalid and server returns 401, we will then try the last known good credentials from the credential service. When the server returns a 401 again because the last known good credentials are also invalid, we will invoke the credential providers with isRetry = true, thereby invalidating the cache on the server side and prompting the user for credentials.

I have created issue [NuGet/Client.Engineering#2637](https://github.com/NuGet/Client.Engineering/issues/2637) to address the unresolved comments in this PR. I will work with you during the quality week.

**EDIT**: I suppose the reason why the `TryLastKnownGoodCredential` method is invoked only once during this functional test is that both the service index URL and the package information URL referred to in the test are running on the same port on the local machine. In contrast, in real-world scenarios, such as Azure DevOps, there are different URL patterns for various resources.

For example, when I installed a package from a private feed, I noticed that the authentication flow was triggered for the following URLs. What's more interesting is that even though HttpSourceCredentials had credentials for the private feed's 'index.json file', the authentication flow (credential service) was still activated for the other two URLs while accessing additional package-related information. This behavior suggests that if the URL differs, the authentication flow will be initiated. This PR could potentially reduce prompting because we first try the cached credentials before invoking credential providers.

https://org.pkgs.vs.com/org/_packaging/feed/nuget/v3/index.json -> feed URL
https://org.pkgs.vs.com/project/_packaging/resource1/nuget/v3/flat2/antlr4.codegenerator/index.json
https://org.pkgs.vs.com/project/_packaging/resource2/nuget/v3/flat2/antlr4.codegenerator/index.json

PR #5564, opened 2023-12-29
Update SettingsUtility.cs

WeihanLi:
## Description

update return value to List<PackageSource> since we return a list now

## PR Checklist

- [ ] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Configuration/PublicAPI.Shipped.txt
+++ b/src/NuGet.Core/NuGet.Configuration/PublicAPI.Shipped.txt
@@ -425,7 +425,7 @@ static NuGet.Configuration.PackageSourceMapping.GetPackageSourceMapping(NuGet.Co
 ~static NuGet.Configuration.SettingsUtility.GetConfigValue(NuGet.Configuration.ISettings settings, string key, bool decrypt = false, bool isPath = false) -> string
 ~static NuGet.Configuration.SettingsUtility.GetDecryptedValueForAddItem(NuGet.Configuration.ISettings settings, string section, string key, bool isPath = false) -> string
 ~static NuGet.Configuration.SettingsUtility.GetDefaultPushSource(NuGet.Configuration.ISettings settings) -> string
-~static NuGet.Configuration.SettingsUtility.GetEnabledSources(NuGet.Configuration.ISettings settings) -> System.Collections.Generic.IEnumerable<NuGet.Configuration.PackageSource>
+~static NuGet.Configuration.SettingsUtility.GetEnabledSources(NuGet.Configuration.ISettings settings) -> System.Collections.Generic.List<NuGet.Configuration.PackageSource>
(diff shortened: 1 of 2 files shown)


nkolev92:
@WeihanLi That's an API breaking change, please refer to https://github.com/NuGet/NuGet.Client/blob/dev/docs/nuget-sdk.md.

src/NuGet.Core/NuGet.Configuration/PublicAPI.Shipped.txt
```diff
@@ -425,7 +425,7 @@ static NuGet.Configuration.PackageSourceMapping.GetPackageSourceMapping(NuGet.Co
 ~static NuGet.Configuration.SettingsUtility.GetConfigValue(NuGet.Configuration.ISettings settings, string key, bool decrypt = false, bool isPath = false) -> string
 ~static NuGet.Configuration.SettingsUtility.GetDecryptedValueForAddItem(NuGet.Configuration.ISettings settings, string section, string key, bool isPath = false) -> string
 ~static NuGet.Configuration.SettingsUtility.GetDefaultPushSource(NuGet.Configuration.ISettings settings) -> string
-~static NuGet.Configuration.SettingsUtility.GetEnabledSources(NuGet.Configuration.ISettings settings) -> System.Collections.Generic.IEnumerable<NuGet.Configuration.PackageSource>
+~static NuGet.Configuration.SettingsUtility.GetEnabledSources(NuGet.Configuration.ISettings settings) -> System.Collections.Generic.List<NuGet.Configuration.PackageSource>
```
nkolev92, line 428: This is still a breaking change, it's just a documented one. 
Personally, I'm not sure it's worth introducing a breaking change here.
@NuGet/nuget-client thoughts?

nkolev92 (changes requested):

src/NuGet.Core/NuGet.Configuration/PublicAPI.Shipped.txt
WeihanLi: Yeah, it is.

I think that it may be safe to update since the `List` is compatible with `IEnumerable`, and it helps avoid duplicate `ToList()` in user code 

image


and it also helps with the new performance rule

https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1859



However, it is still a breaking change, so feel free to close it

nkolev92: There are methods that can be used to still get the perf benefit of a list, https://github.com/NuGet/NuGet.Client/blob/15b4da06d4cb1a8e1bd2a0e710520b68ce0a6915/build/Shared/SharedExtensions.cs#L19-L28. 

Unfortunately because a lot of our doubles as API, we have to consider the trade-offs.

PR #5566, opened 2024-01-03
Use file existence cache for no-op restores

### LINKED ISSUE #13058: NuGet restore cache check is no longer using file existence cache
### NuGet Product Used

Other/NA

### Product Version

Visual Studio 17.9

### Worked before?

not sure

### Impact

Other

### Repro Steps & Context

Check Visual Studio solution loading trace when cache is available,  or performance DDRITs can be used here. During NuGet restore runner time, NuGet took out many thread pool threads, and most of them will be blocked to check whether sha256 files exist (or basically, whether the version of NuGet package is still in the NuGet cache folder).  Because package references are added to the closure of project references, each file is often checked in multiple projects. With often hundreds of files per project, in a large solution, it introduces heavy amounts of system calls/IOs.

Here is the thread time of one trace:

image

### Verbose Logs

_No response_

nkolev92:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13058

## Description

For large solution restores, it is not uncommon that a package is a part of more than one project. In fact due to transitivity, that's the likely scenarios.
When doing no-op, we should cache the file existence checks across projects. 

This does 2 things: 

- The cache file reading itself should not be doing the work for figuring out if all the files are on disk. It mixes model representation and the meaning of the lock file. I deprecated the extra property. 
- Use the package file cache from the request that's shared across projects to check for file existence. 

Note that by using AsList, we avoid enumerator allocations given that the type is always going to be a list.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


martinrrm (dismissed):

src/NuGet.Core/NuGet.ProjectModel/CacheFile.cs
```diff
@@ -23,6 +23,7 @@ public class CacheFile : IEquatable<CacheFile>
         /// </summary>
         public IList<string> ExpectedPackageFilePaths { get; set; }
 
+        [Obsolete("File existence checks are a function of time not the cache file content.")]
```
zivkan, line 26: Should we change this from an auto-prop, to a property whose getting and setting method bodies throw `NotImplementedException` or `NotSupportedException`.

```cs
[Obsolete("This API is no longer supported")]
public bool HasAnyMissingPackageFiles
{
  get => throw new NotImplementedException("This API is no longer supported");
  set => throw new NotImplementedExeption("This API is no longer supported");
}
```

Imagine a 3rd party PkgX that uses NuGet.ProjectModel, and this `HasAnyMissingPackageFiles` property. If I create a project that uses PkgX, and also directly reference a newer version of NuGet.ProjectModel, then from my perspective PkgX will have a bug, because the behaviour is wrong, but I'll never see this obsolete compiler warning because I'm not referencing this API directly. At least a not implemented exception will prevent silent unexpected behaviour.

Although, changing the API to always throw, allows us to give an easy to understand error message, but functionally is effectively equivalent to deleting the public API. Whether we throw or delete, any code using the API will be broken and anyone using the package, directly or indirectly, will have no choice but to make changes. I can't decide if that's better or worse than another app potentially calling the API, but getting unexpected behaviour.

If we want to do the right thing from an API compatibility point of view, we can use a `Lazy`, or equivalent, to only hit the filesystem if someone actually uses this API. I guess it's a question of how much effort we want to go to for a public API that's unlikely to be used by anyone. On GitHub, I can only find this API in forks of NuGet.Client.

Anyway, my comment is more of "something to think about", rather than a request for a change.

nkolev92: I like your idea for throwing. I'll push a change for that and resolve the comment when I do that.

martinrrm (approved):

PR #5569, opened 2024-01-05
Make PMUI packages browse page to allow loading additional packages

### LINKED ISSUE #13063: PM UI Scrolling no longer loads additional packages.
### NuGet Product Used

### Product Version

17.8.34309.116

### Worked before?

17.7.7

### Impact

It’s more difficult to complete my work

### Repro Steps & Context

Reported from [DevCom Feedback](https://developercommunity.visualstudio.com/t/Problem-with-NuGet-Package-Manager-UI-af/10530864)
1. Open the PM UI. 
2. Scroll down to the bottom of the list.

Expected: More packages are loaded
Actual: The end of the list is reached and no new items are loaded. Sometimes the list will load one additional page and then stop loading more. 

### Verbose Logs

_No response_

Nigusu-Allehu:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13063

Regression? Last working version: https://github.com/NuGet/NuGet.Client/commit/0dd5a1ea536201af94725353e4bc711d7560b246

## Description

When trying to browse for packages it only returns the first page and won’t refresh when scrolling down to the bottom of the page. 
- This is because the refreshing the list of packages is not triggered.
- The reason it is not triggered is because `Items` contains `_loadingVulnerabilitiesStatusIndicator` which would make `Items.count` 1 more than it should be. 
- I added a fix that makes sure that more packages are loaded based on the number of packages present not Items.count.

#### Before

before

#### After
Aftrt

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
@@ -738,8 +738,7 @@ namespace NuGet.PackageManagement.UI
                 var first = _scrollViewer.VerticalOffset;
                 var last = _scrollViewer.ViewportHeight + first;
 
-                // Minus one to account for the loading indicator
-                if (_scrollViewer.ViewportHeight > 0 && last >= Items.Count - 1)
+                if (_scrollViewer.ViewportHeight > 0 && last >= PackageItems.Count())
                 {
                     NuGetUIThreadHelper.JoinableTaskFactory.RunAsync(() =>
                         LoadItemsAsync(selectedPackageItem: null, token: CancellationToken.None)
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
@@ -375,7 +375,6 @@ namespace NuGet.PackageManagement.UI
                     else
                     {
                         Items.Remove(_loadingStatusIndicator);
-                        Items.Remove(_loadingVulnerabilitiesStatusIndicator);
                     }
                 }
             }
@@ -734,11 +733,23 @@ namespace NuGet.PackageManagement.UI
 
         private void ScrollViewer_ScrollChanged(object sender, ScrollChangedEventArgs e)
         {
+            int packagesCount = Items.Count;
+
+            if (Items.Contains(_loadingStatusIndicator))
+            {
+                packagesCount--;
+            }
+
+            if (Items.Contains(_loadingVulnerabilitiesStatusIndicator))
+            {
+                packagesCount--;
+            }
+
             if (_loader?.State.LoadingStatus == LoadingStatus.Ready)
             {
                 var first = _scrollViewer.VerticalOffset;
                 var last = _scrollViewer.ViewportHeight + first;
-                if (_scrollViewer.ViewportHeight > 0 && last >= Items.Count)
+                if (_scrollViewer.ViewportHeight > 0 && last >= packagesCount)
                 {
                     NuGetUIThreadHelper.JoinableTaskFactory.RunAsync(() =>
                         LoadItemsAsync(selectedPackageItem: null, token: CancellationToken.None)
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
@@ -375,6 +375,7 @@ namespace NuGet.PackageManagement.UI
                     else
                     {
                         Items.Remove(_loadingStatusIndicator);
+                        Items.Remove(_loadingVulnerabilitiesStatusIndicator);
                     }
                 }
             }
@@ -738,7 +739,7 @@ namespace NuGet.PackageManagement.UI
                 var first = _scrollViewer.VerticalOffset;
                 var last = _scrollViewer.ViewportHeight + first;
 
-                if (_scrollViewer.ViewportHeight > 0 && last >= PackageItems.Count())
+                if (_scrollViewer.ViewportHeight > 0 && last >= Items.Count)
                 {
                     NuGetUIThreadHelper.JoinableTaskFactory.RunAsync(() =>
                         LoadItemsAsync(selectedPackageItem: null, token: CancellationToken.None)
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
@@ -737,7 +737,9 @@ namespace NuGet.PackageManagement.UI
             {
                 var first = _scrollViewer.VerticalOffset;
                 var last = _scrollViewer.ViewportHeight + first;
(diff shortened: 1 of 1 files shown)


nkolev92:
> Regression? Last working version: https://github.com/NuGet/NuGet.Client/commit/eb04000f97ade4add28a9e7447baaadee22ee863

Isn't that a test change? Or are you linking the last change *prior* to the change that broke it? Can we get a link to that instead?

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -737,7 +737,9 @@ private void ScrollViewer_ScrollChanged(object sender, ScrollChangedEventArgs e)
             {
                 var first = _scrollViewer.VerticalOffset;
                 var last = _scrollViewer.ViewportHeight + first;
-                if (_scrollViewer.ViewportHeight > 0 && last >= Items.Count)
+
+                // Minus one to account for the loading indicator
+                if (_scrollViewer.ViewportHeight > 0 && last >= Items.Count - 1)
```
nkolev92: Is this indicator always going to be there?

Can we change the code so that we're not using an indicator that's added to the item list potentially? Maybe a state that's independent from the list of items we're going to show to the user?

donnie-msft: > Can we change the code so that we're not using an indicator that's added to the item list potentially?

Tried that before and ended up having issues with scrolling (couldn't do item-by-item scrolling). I can try figuring another strategy that avoids that issue, but I'm not sure it's in scope for fixing this bug.

https://github.com/NuGet/Home/issues/10150

donnie-msft (dismissed):
This line hasn't been changed in 9 years. Therefore, it doesn't make sense to me how it could be the cause of a regression since 17.7(ish?). If you're sure it solves it, I think it's a resonable change; however, we should understand what actually introduced this bug.

nkolev92:
@donnie-msft 
The _loadingVulnerabilitiesStatusIndicator  hasn't been there for 9 years, it's new. 
You're thinking of _loadingStatusIndicator  which is different.

donnie-msft:
@nkolev92 the code in this PR hasn't been changed in 9 years: https://github.com/NuGet/NuGet.Client/blob/e25cba88147cfdbb361d0a6b2402717030b1d6ce/src/NuGet.Clients/PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs#L420

It looks like this event is being reused, but wasn't modified to account for 2 loading indicators.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
aortiz-msft: Coming back to this question - Is this indicator always going to be there? If yes, let's merge now and follow up on a better solution later. I'd like this fix considered for servicing.

Nigusu-Allehu: fixed by using `PackagetItems` which is independet of loading status indicators.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
```diff
@@ -737,7 +737,8 @@ private void ScrollViewer_ScrollChanged(object sender, ScrollChangedEventArgs e)
             {
                 var first = _scrollViewer.VerticalOffset;
                 var last = _scrollViewer.ViewportHeight + first;
-                if (_scrollViewer.ViewportHeight > 0 && last >= Items.Count)
+
+                if (_scrollViewer.ViewportHeight > 0 && last >= PackageItems.Count())
```
nkolev92: Accessing PackageItems creates an array, and then Count() allocates. 

Do we have a cheaper way of doing this?

donnie-msft (commented):
I understand there's no existing tests, and this isn't the most intuitive to write tests for as the architecture is brittle. It only partially follows some MVVM paradigms. However, it's still good to describe test strategies used, a few repro steps, and ideally a couple screenshots in PR descriptions. When reviewing now, and when we refer back to this PR later, it'll be convenient.

Can you describe the scenario a bit more?
A few questions that came to mind:
1. Why set the status here only to remove it in the following lines?
https://github.com/NuGet/NuGet.Client/blob/a4768a1942d74f72de610798082f2c23e5db392f/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs#L357
https://github.com/NuGet/NuGet.Client/blob/a4768a1942d74f72de610798082f2c23e5db392f/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs#L378

2. Should adding the vulnerability loading indicator also set the `addedLoadingIndicator` flag to `true`?
https://github.com/NuGet/NuGet.Client/blob/a4768a1942d74f72de610798082f2c23e5db392f/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs#L304

Nigusu-Allehu:
> I understand there's no existing tests, and this isn't the most intuitive to write tests for as the architecture is brittle. It only partially follows some MVVM paradigms. However, it's still good to describe test strategies used, a few repro steps, and ideally a couple screenshots in PR descriptions. When reviewing now, and when we refer back to this PR later, it'll be convenient.
> 
> Can you describe the scenario a bit more? A few questions that came to mind:
> 
> 1. Why set the status here only to remove it in the following lines?
>    https://github.com/NuGet/NuGet.Client/blob/a4768a1942d74f72de610798082f2c23e5db392f/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs#L357
>    
>    https://github.com/NuGet/NuGet.Client/blob/a4768a1942d74f72de610798082f2c23e5db392f/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs#L378
> 2. Should adding the vulnerability loading indicator also set the `addedLoadingIndicator` flag to `true`?
>    https://github.com/NuGet/NuGet.Client/blob/a4768a1942d74f72de610798082f2c23e5db392f/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs#L304

I have updated the PR description to contain gifs to give a better description.
Answer to your two questions
1. You are right, removing the indicator is not the right thing to do. That would result in the no items found status to disappear.
2. The addedLoadingIndicator flag is used to appropriately remove `_loadingStatusIndicator`, It used so that, if it is set to true, the code knows `Items` contains 1 more item (_loadingStatusIndicator) than just the packages.

donnie-msft (commented):
The code makes more sense, now. Did you also test that the vulnerability indicator does show while vulnerabilities are still loading? In other words, that nothing is regressed with the 2nd indicator?

Nigusu-Allehu:
> The code makes more sense, now. Did you also test that the vulnerability indicator does show while vulnerabilities are still loading? In other words, that nothing is regressed with the 2nd indicator?

Nothing regressed regarding how the vulnerability indicator works. This only makes sure we load more packages as soon as we reach the end of the list of all packages.

donnie-msft (approved):

PR #5571, opened 2024-01-06
dotnet list package prompts only once for authentication even though there are multiple projects in a solution

### LINKED ISSUE #12456: dotnet list package --vulnerable requires constant login to 3rd party nuget feed
### NuGet Product Used

dotnet.exe

### Product Version

.NET SDK  7.0.200

### Worked before?

_No response_

### Impact

### Repro Steps & Context

On calling `dotnet list package --interactive --vulnerable` I have to constantly sign in via the web browser for an internal nuget feed.

Without the `--vulnerable` switch it seems to use my cached login and I don't even need to login once.

Expected behavior:

I would expect to login at most once for the command to finish.

Log Output:
[log.txt](https://github.com/NuGet/Home/files/10848428/log.txt)

### Verbose Logs

_No response_

kartheekp-ms:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12456

## Description

**Problem:**  `dotnet list package --interactive --vulnerable/--outdated` prompts for authentication once for every project when executed on a solution. If there are "n" projects in a solution, then there will be "n" number of authentication prompts.

**Root cause:** In the current implementation, the `ListPackageCommandRunner.GetReportDataAsync()` method incorrectly invokes `PopulateSourceRepositoryCache()` for each project in the solution. This is problematic because source repositories are intended to be singletons for the entire operation. Updating the source repositories for every project, especially when accessing private feeds, results in NuGet invoking the Credential Provider with `isRetry = true`. Specifically for the Azure DevOps Credential Provider, this leads to the local cache being invalidated, which may prompt users to enter their credentials again, an unnecessary step.

**Solution:** This Pull Request resolves the issue by ensuring that users are prompted for authentication only once. It achieves this by calling the `PopulateSourceRepositoryCache()` method a single time, rather than for each project. This approach allows credentials to be reused while accessing the private feed, avoiding cache invalidation and repeated credential prompts. If authentication is successful, there should be no further need for authentication to access the feed during the command's execution.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added - I tested the changes manually on the patched SDK also.
  
<details>
  <summary> Current behavior which displays authentication prompt for every project (Click to expand!) </summary>
  
dotnet list package --vulnerable --interactive
    [CredentialProvider]DeviceFlow: https://org.pkgs.vs.com/org/_packaging/feedname/nuget/v3/index.json
    [CredentialProvider]ATTENTION: User interaction required.
 
    **********************************************************************
 
    To sign in, use a web browser to open the page https://microsoft.com/devicelogin and enter the code {CODE} to authenticate.
 
    **********************************************************************
    [CredentialProvider]DeviceFlow: https://org.pkgs.vs.com/org/_packaging/feedname/nuget/v3/index.json
    [CredentialProvider]ATTENTION: User interaction required.
 
    **********************************************************************
 
    To sign in, use a web browser to open the page https://microsoft.com/devicelogin and enter the code {CODE} to authenticate.
 
    **********************************************************************
 
The following sources were used:
https://org.pkgs.vs.com/org/_packaging/feedname/nuget/v3/index.json
 
The given project `authentication-demo` has no vulnerable packages given the current sources.
The given project `ClassLibrary1` has no vulnerable packages given the current sources.

</details>

<details>
  <summary> New behavior which displays only one authentication prompt irrespective of number of projects (Click to expand!) </summary>
  
dotnet list package --vulnerable --interactive
    [CredentialProvider]DeviceFlow: https://org.pkgs.vs.com/org/_packaging/feedname/nuget/v3/index.json
    [CredentialProvider]ATTENTION: User interaction required.
 
    **********************************************************************
 
    To sign in, use a web browser to open the page https://microsoft.com/devicelogin and enter the code {CODE} to authenticate.
 
    **********************************************************************
 
The following sources were used:
https://org.pkgs.vs.com/org/_packaging/feedname/nuget/v3/index.json
 
The given project `authentication-demo` has no vulnerable packages given the current sources.
The given project `ClassLibrary1` has no vulnerable packages given the current sources.

</details>

- **Documentation**
  
  - [x] N/A

### DIFF
--- a/test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/ListPackageTests.cs
+++ b/test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/ListPackageTests.cs
@@ -5,14 +5,21 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Net;
 using System.Reflection;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.CommandLineUtils;
 using Moq;
 using NuGet.CommandLine.XPlat;
+using NuGet.CommandLine.XPlat.ListPackage;
+using NuGet.Commands;
+using NuGet.Commands.Test;
 using NuGet.Common;
+using NuGet.Configuration;
 using NuGet.Protocol;
 using NuGet.Test.Utility;
+using Test.Utility;
 using Xunit;
 
 namespace NuGet.XPlat.FuncTest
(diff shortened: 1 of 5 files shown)


test/TestUtilities/Test.Utility/FileSystemBackedV3MockServer.cs
```diff
@@ -44,6 +46,18 @@ private void InitializeServer()
...
+                    string authorization = request.Headers["Authorization"];
+                    if (authorization == null)
+                    {
+                        return new Action<HttpListenerResponse>(response =>
+                        {
+                            response.StatusCode = 401;
+                            response.AddHeader("WWW-Authenticate", "Basic");
+                        });
+                    }
+                }
```
kartheekp-ms, line 60: Considering that this PR pertains to a private feed requiring an authorization header, I have incorporated logic to ensure the code utilizes credential providers to obtain credentials.

test/TestUtilities/Test.Utility/Commands/ProjectTestHelpers.cs
```diff
@@ -205,7 +205,7 @@ public static TestRestoreRequest CreateRestoreRequest(SimpleTestPathContext path
 
             return new TestRestoreRequest(projectToRestore, sources, pathContext.UserPackagesFolder, logger)
             {
-                LockFilePath = Path.Combine(projectToRestore.FilePath, LockFileFormat.AssetsFileName),
```
kartheekp-ms, line 208: The current implementation generates the `project.assets.json` file within the `projectname.csproj` folder, rather than in the project's designated intermediate path folder i.e. `obj`. For instance, it creates the assets file in `$\.test\work\guid\guid\solution\projectA\projectA.csproj\project.assets.json` instead of the expected `$\.test\work\guid\guid\solution\projectA\obj\project.assets.json`. While I'm uncertain about the correctness of hardcoding `obj`, the local tests passed, leading me to assume it's acceptable.

nkolev92: I had attempted a similar change at some point and there were *a lot of test failures*. 

The correct way to set it is actually to use the output path, since that's what real restore does. 

`projectToRestore.RestoreMetadata.OutputPath`

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/ListPackageTests.cs
```diff
@@ -156,6 +163,123 @@ public void BasicListPackage_OutputFormat_BadInput_Parsing_Fails(string outputFo
...
+
+            projectA.AddPackageToAllFrameworks(packageA100);
+            projectB.AddPackageToAllFrameworks(packageB100);
+
+            var solution = new SimpleTestSolutionContext(pathContext.SolutionRoot);
+            solution.Projects.Add(projectA);
+            solution.Projects.Add(projectB);
+            solution.Create(pathContext.SolutionRoot);
+
+            SimpleTestSettingsContext.RemoveSource(pathContext.Settings.XML, "source");
```
nkolev92, line 191: Not really a feedback on your code, since you're just using the APIs (I probably have to in the past but did not notice, hopefully I did not add them :D) 
These methods are a bit unusual. 
There's statics to operate on what's *always* the object being represented by the context :D

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/ListPackageTests.cs
```diff
@@ -156,6 +163,123 @@ public void BasicListPackage_OutputFormat_BadInput_Parsing_Fails(string outputFo
...
+                                        path: Path.Combine(pathContext.SolutionRoot, "solution.sln"),
+                                        packageSources: [new(mockServer.ServiceIndexUri)],
+                                        frameworks: ["net6.0"],
+                                        reportType: ReportType.Vulnerable,
+                                        renderer: new ListPackageConsoleRenderer(),
+                                        includeTransitive: false,
+                                        prerelease: false,
+                                        highestPatch: false,
+                                        highestMinor: false,
+                                        logger: NullLogger.Instance,
```
nkolev92: test: If you pass a `TestLogger` here, you can then show all the messages from there in line 223, if case the assertion fails.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/ListPackageTests.cs
```diff
@@ -156,6 +163,123 @@ public void BasicListPackage_OutputFormat_BadInput_Parsing_Fails(string outputFo
...
+            }
+
+            static async Task RestoreProjectAsync(ISettings settings,
+                SimpleTestPathContext pathContext,
+                SimpleTestProjectContext project,
+                IEnumerable<PackageSource> packageSources,
+                IList<string> fallbackFolders,
+                string globalpackagesFolder)
+            {
+                project.Sources = packageSources;
```
nkolev92: Aren't there some test helpers utils that set these up?
In particular https://github.com/NuGet/NuGet.Client/blob/dev/test/TestUtilities/Test.Utility/Commands/ProjectTestHelpers.cs#L284-L291

test/TestUtilities/Test.Utility/Commands/ProjectTestHelpers.cs
kartheekp-ms: All the CI checks passed with the hardcoded string. I liked your approach to use `projectToRestore.RestoreMetadata.OutputPath` property. Fixed in the latest commit.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/ListPackageTests.cs
kartheekp-ms: In this test scenario, I removed the local folder source because I wanted the requests to be directed to the file-based mock server and credential service.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/ListPackageTests.cs
kartheekp-ms: This is precisely what I did in the commit at https://github.com/NuGet/NuGet.Client/pull/5571/commits/f8f381b658fc390292525ed18be048c6d3eb33c0 to troubleshoot why the test failed on Linux. Then, I realized that there were other tests using the mock server that also failed on Linux and had the `Platform Fact Windows` configuration set. I can implement the same changes in the next commit.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/ListPackageTests.cs
kartheekp-ms: If I replace all the lines from 271 to 274 with the following 2 lines, then the test is passing. Otherwise, test fails with an error message saying `value is null` for Sources property.

```cs
project.Sources = packageSources;
var packageSpec = ProjectTestHelpers.WithSettingsBasedRestoreMetadata(project.PackageSpec, settings);
```

nkolev92: That's unusual. If you debug, why are the sources different?

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/ListPackageTests.cs
```diff
@@ -156,6 +163,125 @@ public void BasicListPackage_OutputFormat_BadInput_Parsing_Fails(string outputFo
...
+                    .Returns((Uri sourceUri, bool isProxyRequest, out ICredentials outCredentials) =>
+                    {
+                        outCredentials = cachedCredentials;
+                        return outCredentials != null;
+                    });
+            }
+
+            static async Task RestoreProjectsAsync(SimpleTestPathContext pathContext, SimpleTestProjectContext projectA, SimpleTestProjectContext projectB)
+            {
+                var settings = Settings.LoadDefaultSettings(Path.GetDirectoryName(pathContext.NuGetConfig), Path.GetFileName(pathContext.NuGetConfig), null);
```
nkolev92: The settings should be loaded wrt a project directory, not NuGet.config.

kartheekp-ms: I changed the code to load setting with respect to a solution directory, it worked. I will push the commit shortly.

```cs
var settings = Settings.LoadDefaultSettings(Path.GetDirectoryName(pathContext.SolutionRoot), Path.GetFileName(pathContext.NuGetConfig), null);
```

When I passed `projectA` directory path, the test failed because it tried to find out the `NuGet.Config` file under `projectA` folder instead of `solution-level folder`. Hence changed the code to pass solution directory which passed the test locally.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/ListPackageTests.cs
kartheekp-ms: Sources are not different. It is just that `Project.Sources` is null. Hence, accessing `Project.PackageSpec` throws a null reference exception because existing implementation has `Sources.ToList()` https://github.com/NuGet/NuGet.Client/blob/dev/test/TestUtilities/Test.Utility/SimpleTestSetup/SimpleTestProjectContext.cs#L261. 

I added a null check and changed the test code just to invoke  `ProjectTestHelpers.WithSettingsBasedRestoreMetadata(project.PackageSpec, settings);`. The tests passed locally and pushed the commit.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/ListPackageTests.cs
kartheekp-ms: Fixed in the latest commit.

aortiz-msft (changes requested):
Could you please elaborate on the PR description, why moving PopulateSourceRepositoryCache to GetReportDataAsync fixes the problem? For PR descriptions, we should include information that would help someone else understand the motivation and rationale of a code change six months later.

test/TestUtilities/Test.Utility/Commands/ProjectTestHelpers.cs
nkolev92: I think you're combining too many things, you're combining the base directory and output path. 

You only need the outputpath.

kartheekp-ms: Thanks for catching that. It was working because `Path.Combine()` normalizes the path. I addressed your comment in the latest commit.

nkolev92 (approved):

Nigusu-Allehu (approved):

aortiz-msft (approved):

PR #5572, opened 2024-01-08
Enable central package management implicitly when Directory.Packages.props exists

### LINKED ISSUE #11834: [DCR]: Central Package Management - Respect .props file as a way to opt-in to the feature.

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Home/issues/11834

## Description

This change makes it so the presence of a `Directory.Packages.props` means that central package management is implicitly on.  I've had countless people ask over the last year why you have to have the file and explicitly set `ManagePackageVersionsCentrally` to `true`.  The logic now assumes if `Directory.Packages.props` exists then `ManagePackageVersionsCentrally` should be `true` unless the user has already set that property to a value, including if its set in `Directory.Packages.props`.  This will make it easier for customers to onboard.  

This is a redo of https://github.com/NuGet/NuGet.Client/pull/5540 which was closed during the holiday break but then I forced pushed a rebase which GitHub then prevents me from re-opening.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.CommandLine/MsBuildUtility.cs
+++ b/src/NuGet.Clients/NuGet.CommandLine/MsBuildUtility.cs
@@ -23,6 +23,7 @@ namespace NuGet.CommandLine
     {
         internal const int MsBuildWaitTime = 2 * 60 * 1000; // 2 minutes in milliseconds
 
+        private const string NuGetProps = "NuGet.CommandLine.NuGet.props";
         private const string NuGetTargets = "NuGet.CommandLine.NuGet.targets";
         private static readonly XNamespace MSBuildNamespace = XNamespace.Get("http://schemas.microsoft.com/developer/msbuild/2003");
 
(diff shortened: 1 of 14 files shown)


martinrrm (approved):

nkolev92 (approved):

ViktorHofer:
Note that this broke https://github.com/dotnet/runtime/pull/98476 when upgrading to .NET 9 Preview 1. This is because previously an empty Directory.Packages.props file was created to avoid importing a root one. We understand the right fix but it might be worthwhile to write a breaking change notice for this.

jeffkl:
> Note that this broke [dotnet/runtime#98476](https://github.com/dotnet/runtime/pull/98476) when upgrading to .NET 9 Preview 1. This is because previously an empty Directory.Packages.props file was created to avoid importing a root one. We understand the right fix but it might be worthwhile to write a breaking change notice for this.

Sorry about that.  Unfortunately its to be expected any repository that's running tests where those tests are going to import the repo's build logic and the repo is using CPM but the tests are not.  We really wanted to reduce the friction with onboarding but knew that a couple repos would need to update to explicitly disable CPM for their tests.

PR #5573, opened 2024-01-08
End the restore process when a user cancels a build during static gra…

### LINKED ISSUE #13140: Cancelling static graph-based restore does not end the restore process
### NuGet Product Used

MSBuild.exe, dotnet.exe

### Product Version

All

### Worked before?

_No response_

### Impact

### Repro Steps & Context

Static graph-based restore launches an external process and waits for it to return:

https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Build.Tasks/StaticGraphRestoreTaskBase.cs#L184

However, the code passes a cancellation token which is fired when the user cancels a build via CTRL+C.  The `catch` code is there to return `true` so that cancelled builds don't "fail" but it does not kill the process.

https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Build.Tasks/StaticGraphRestoreTaskBase.cs#L211-L217

The `try...catch` which handles cancellation should be moved to surround just the `Semaphore.Wait()` since that's the only call that would throw an `OperationCanceledException`.  The proceeding code would then kill the process and return `true` in the case where the user canceled the build.

### Verbose Logs

_No response_

jeffkl:
…ph-based restore

## Bug

Fixes: https://github.com/NuGet/Home/issues/13140

## Description

During cancellation, MSBuild notifies the `RestoreTaskEx` which then signals the cancellation token.  The cancellation token is passed to the `Semaphore.Wait()` which throws an `OperationCanceledException`.  However, the code then just returns `true` and does not kill the process.

This change modifies the scope of the `try...catch` block which handles the `OperationCancelledException` so that the process is then killed and still returns `true`.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception - This is hard to test since it relies on cancellation but I did debug the change to verify the functionality
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Build.Tasks/StaticGraphRestoreTaskBase.cs
+++ b/src/NuGet.Core/NuGet.Build.Tasks/StaticGraphRestoreTaskBase.cs
@@ -145,76 +145,77 @@ namespace NuGet.Build.Tasks
 
                     process.Exited += (sender, args) => semaphore.Release();
 
+                    Log.LogMessageFromResources(MessageImportance.Low, nameof(Strings.Log_RunningStaticGraphRestoreCommand), process.StartInfo.FileName, process.StartInfo.Arguments);
+
+                    Encoding previousConsoleInputEncoding = Console.InputEncoding;
+
+                    // Set the input encoding to UTF8 without a byte order mark, the spawned process will use this encoding on .NET Framework
+                    Console.InputEncoding = utf8Encoding;
+
                     try
                     {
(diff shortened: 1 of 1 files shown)


jeffkl:
I recommend looking at the diff while ignoring whitespace since it changes indentation: https://github.com/NuGet/NuGet.Client/pull/5573/files?diff=split&w=1

nkolev92 (approved):

PR #5582, opened 2024-01-09
Update dependencies - Microsoft.ServiceHub.Framework, other VS component versions and various test dependencies

nkolev92:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2168

## Description

I'm bringing a few of our dependencies more up to date. 
Unfortunately the Visual Studio dependencies are not all consistently versioned, so updating can sometimes be a pain. 

A few notable things I've done.

- Update servicehub framework to 4.2.0. There was a breaking change.
- Update xunit and runner versions to latest compatible (note that we can't update the runner to newer versions since compatibility for netcoreapp3.1 has been dropped.
- Order the xunit analyzers suppressions and add a few more.
- Add framework assemblies where required. We were getting these framework assemblies through packages but they're not a part of that anymore.
- Addressed a few analyzer warnings in places where there was only 1 or 2 issues.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/build/test.targets
+++ b/build/test.targets
@@ -4,7 +4,7 @@
   <!-- Shared test project template -->
 
   <PropertyGroup>
-    <NoWarn>$(NoWarn);CS1701;xUnit2013;xUnit2000;xUnit2017;xUnit2006;xUnit2009;xUnit2013;xUnit2014;xUnit2003;xUnit2010;xUnit2015;xUnit2007;xUnit1013;xUnit1014;xUnit2002;xUnit2012;xUnit2004</NoWarn>
+    <NoWarn>$(NoWarn);CS1701;xUnit1012;xUnit1013;xUnit1014;xUnit1031;xUnit1041;xUnit2000;xUnit2002;xUnit2003;xUnit2004;xUnit2006;xUnit2007;xUnit2009;xUnit2010;xUnit2012;xUnit2013;xUnit2013;xUnit2014;xUnit2015;xUnit2017;xUnit2020;xUnit2021</NoWarn>
   </PropertyGroup>
   
   <ItemGroup>
--- a/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/NuGet.VisualStudio.Internal.Contracts.csproj
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/NuGet.VisualStudio.Internal.Contracts.csproj
@@ -6,6 +6,7 @@
     <IncludeInVsix>true</IncludeInVsix>
     <RootNamespace>NuGet.VisualStudio.Internal.Contracts</RootNamespace>
     <Nullable>enable</Nullable>
+    <NoWarn>$(NoWarn);CS8602</NoWarn>
   </PropertyGroup>
 
   <ItemGroup>
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -6,7 +6,7 @@
     <NewtonsoftJsonPackageVersion Condition="'$(NewtonsoftJsonPackageVersion)' == ''">13.0.3</NewtonsoftJsonPackageVersion>
     <SystemTextJsonVersion Condition="'$(SystemTextJsonVersion)' == ''">7.0.3</SystemTextJsonVersion>
     <SystemPackagesVersion Condition="'$(SystemPackagesVersion)' == ''">4.3.0</SystemPackagesVersion>
-    <VSFrameworkVersion Condition="'$(VSFrameworkVersion)' == ''">17.4.33103.184</VSFrameworkVersion>
+    <VSFrameworkVersion Condition="'$(VSFrameworkVersion)' == ''">17.6.36389</VSFrameworkVersion>
     <VSServicesVersion Condition="'$(VSServicesVersion)' == ''">16.153.0</VSServicesVersion>
     <SystemCommandLineVersion Condition="'$(SystemCommandLineVersion)' == ''">2.0.0-beta4.23307.1</SystemCommandLineVersion>
     <!-- Default MSBuild version -->
@@ -128,7 +128,7 @@
               '$(MSBuildProjectFile)' == 'NuGet.VisualStudio.Contracts.csproj' OR
               '$(MSBuildProjectFile)' == 'NuGet.VisualStudio.Interop.csproj' OR
               '$(MSBuildProjectFile)' == 'NuGet.SolutionRestoreManager.Interop.csproj'">
-    <PackageVersion Include="Microsoft.ServiceHub.Framework" Version="4.1.112" />
+    <PackageVersion Include="Microsoft.ServiceHub.Framework" Version="4.2.102" />
     <PackageVersion Include="Microsoft.VisualStudio.ComponentModelHost" Version="17.4.255" />
     <PackageVersion Update="Microsoft.VisualStudio.SDK" Version="" />
     <PackageVersion Include="Microsoft.VisualStudio.Shell.15.0" Version="$(VSFrameworkVersion)" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -32,7 +32,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageVersion Include="FluentAssertions" Version="6.6.0" />
+    <PackageVersion Include="FluentAssertions" Version="6.12.0" />
     <PackageVersion Include="ILMerge" Version="3.0.41" />
     <PackageVersion Include="Lucene.Net" Version="3.0.3" />
     <PackageVersion Include="MessagePack" Version="2.5.108" />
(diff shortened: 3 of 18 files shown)


nkolev92:
There are failing tests and I'm having some strong name validation issues in my build, so marking this as draft until I fix it.

nkolev92:
All tests are passing now.

src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/NuGet.VisualStudio.Internal.Contracts.csproj
```diff
@@ -6,6 +6,7 @@
     <IncludeInVsix>true</IncludeInVsix>
     <RootNamespace>NuGet.VisualStudio.Internal.Contracts</RootNamespace>
     <Nullable>enable</Nullable>
+    <NoWarn>$(NoWarn);CS8602</NoWarn>
```
zivkan, line 9: Why suppress it for the whole assembly?  Can the nullable warnings be suppressed at individual locations, or code changed to actually allow the compiler to help us detect null bugs?

build/test.targets
```diff
@@ -4,7 +4,7 @@
   <!-- Shared test project template -->
 
   <PropertyGroup>
-    <NoWarn>$(NoWarn);CS1701;xUnit2013;xUnit2000;xUnit2017;xUnit2006;xUnit2009;xUnit2013;xUnit2014;xUnit2003;xUnit2010;xUnit2015;xUnit2007;xUnit1013;xUnit1014;xUnit2002;xUnit2012;xUnit2004</NoWarn>
+    <NoWarn>$(NoWarn);CS1701;xUnit1012;xUnit1013;xUnit1014;xUnit1031;xUnit1041;xUnit2000;xUnit2002;xUnit2003;xUnit2004;xUnit2006;xUnit2007;xUnit2009;xUnit2010;xUnit2012;xUnit2013;xUnit2013;xUnit2014;xUnit2015;xUnit2017;xUnit2020;xUnit2021</NoWarn>
```
zivkan, line 7: I never noticed these xunit analyzer suppressions before 😞 

It looks like you re-ordered them, in addition to adding a few more, making it hard for me to figure out which ones you're adding. Certainly 2020 and 2021 are added, but I can't tell what else.

Also, going though the list carefully, I see that 2013 is listed twice.

Here are my opinions:

|code|what|opinion|
|--|--|--|
|[xunit2012](https://xunit.net/xunit.analyzers/rules/xUnit2012)|nullable checking for inline data|Seems useful to keep IMO|
|[xunit2013](https://xunit.net/xunit.analyzers/rules/xUnit1013)|public method in a class with at least 1 test should be marked as a test|From a software design point of view, I like this one. Putting shared utilities into a class without tests sounds reasonable to me.  But it won't catch bugs, so I can understand suppressing it|
|[xunit2014](https://xunit.net/xunit.analyzers/rules/xUnit1014)|using string instead of nameof to MemberData|If the string pointed to a name that doesn't exist, I hope that xunit would fail the test, rather than skip, so this probably doesn't help correctness.  But it's also really, really trivial to fix|
|[xunit2031](https://xunit.net/xunit.analyzers/rules/xUnit1031)|convert tests using `.Wait()` or `.Result` to async|I don't agree with xunit's justification for this analyzer. They're taking generic async advice and applying it to a scenario where I dont' think deadlocks are possible. Plus, they ignore the number of parallel tests that end up running when async, and then say "you can disable running tests in parallel". Well, maybe I want to run in parallel, but limited to the number of CPU cores. Anyway, I agree with suppressing this one.|
|[xunit1041](https://xunit.net/xunit.analyzers/rules/xUnit1041)|test class constructors with fixtures must have fixture sources|This sounds like a potential correctness issue to me. CI might have passed all tests, but did classes that violate this analyzer actually work correctly?|
|[xunit2000](https://xunit.net/xunit.analyzers/rules/xUnit2000)|put expected and actual arguments in correct order in asserts|I like this analyzer, because when the assert fails for a test you've never investigated before, having an error message tell you the wrong actual and wrong expected can put you in the wrong mindset and slow down investigation|
|[xunit2002](https://xunit.net/xunit.analyzers/rules/xUnit2002)|don't use null check on value type|value types will never be null, so any invalid checks will always fail, making this analyzer more of a perf suggestion. But it's also trivial to fix, since it just requires the line to be deleted.|
|[xunit2003](https://xunit.net/xunit.analyzers/rules/xUnit2003)|use Assert.Null rather than assert where expected is null|This analuzer seems very opinionated, and I doubt results in a better error message. I'd be interested in testing to see if the error message is better or not, but keeping it suppressed is probably ok. On the other hand, it's really trivial to fix|
|[xunit2004](https://xunit.net/xunit.analyzers/rules/xUnit2004)|use Assert.True and Assert.False|Just like 2003, but for true/false rather than null|
|[xunit2006](https://xunit.net/xunit.analyzers/rules/xUnit2006)|Use `Assert.Equal`, not `Assert.Equal<string>` for strings|similar as previous two, is more of a perf improvement, but trivial to fix the analyzer violation|
|[xunit2007](https://xunit.net/xunit.analyzers/rules/xUnit2007)|Use `Assert.IsType<T>`, not `Assert.IsType(typeof(T), v)`|looks like an overload was added at some point, and they want everyone to migrate to the new syntax. This suppression doesn't bother me, but also is trivial to fix|
|[xunit2009](https://xunit.net/xunit.analyzers/rules/xUnit2009)|Use `Assert.Contains` to find substrings|Error message will probably be better, so sounds like a good analyzer to me|
|[xunit2010](https://xunit.net/xunit.analyzers/rules/xUnit2010)|Use `Assert.Equals(...)` rather than `Assert.True(string.Equals(...))`|Error message will be much improved if we fix the violations. I don't understand why this should ever be suppressed|
|[xunit2012](https://xunit.net/xunit.analyzers/rules/xUnit2012)|Use `Assert.Contains` instead of LINQ `.Any`|Seems like a good analyzer to me, as it should give better failure messages|
|[xunit2013](https://xunit.net/xunit.analyzers/rules/xUnit2013)|Use Assert's `.Empty`, `.Single`, or `.NotEmpty`, rather than asserting collection length against 0 or 1|The failure message might be slightly improved, since it'll give us context that it's about a collection, rather than just "expected 1 actual 0", but overall not one of the more helpful analyzers IMO|
|[xunit2014](https://xunit.net/xunit.analyzers/rules/xUnit2014)|Use `Assert.ThrowsAsync<T>`, not `Assert.Throws<T>`|I think the example in their docs is wrong, because their lamba function isn't async, but to me this seems like a potential correctness issue. Although, I think I'd expect the error to cause xunit to always throw, meaning the test can't pass unless it's fixed. But anyway, I think this analyzer is about correctness, so should not be suppressed|
|[xunit2015](https://xunit.net/xunit.analyzers/rules/xUnit2015)|Use `Assert.Throws<T>` instead of `Assert.Throws(typeof(T), v)`|Failures probably have a better error message, but maybe the violation has a sufficient error message|
|[xunit2017](https://xunit.net/xunit.analyzers/rules/xUnit2017)|Use `Assert.Contains`, not LINQ `.Contains`|Like others, trivial to fix, maybe better error message|
|[xunit2020](https://xunit.net/xunit.analyzers/rules/xUnit2020)|Use `Assert.Fail(message)` not `Assert.True(false, message)`|Either a new API was added that didn't exist before, or this is just raising awareness of the Fail API. Not a correctness issue, but seems reasonable to me, and trivial to fix|
|[xunit2021](https://xunit.net/xunit.analyzers/rules/xUnit2021)|If you use an async assertion, you must await the result|xunit lists this as an error, not a warning. So, I'm not sure that nowarn will work. But if it does, I think it's dangerous to nowarn. I believe this is a very clear bug, and assertion failures will never cause the test to fail. This should abosolutely be fixed, not ignored|

TL;DR, 2021 should not be suppressed, and maybe 2014 as well. The others are largely about reducing tech debt, so I think it's disppointing to ignore.

src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/NuGet.VisualStudio.Internal.Contracts.csproj
nkolev92: There are 66 errors and I didn't want to accidentally introduce more bugs, so I'll leave those changes for a future PR.

I'll make that changes in a new PR to avoid potentially breaking something sometihng.

nkolev92: Issue: https://github.com/NuGet/Home/issues/13179

build/test.targets
nkolev92: The new ones are: 1012, 1031, 1041, 2020, 2021. 
I removed the duplicate. 

I think we can definitely address these, but in the interest of getting the package upgrades done, I think it'd be better to create an issue for this as well.

nkolev92: fwiw, I'm working on these, I'll just do it in a separate PR.

zivkan (approved):

PR #5585, opened 2024-01-10
Create a Apex daily test pipeline that triggered by NuGet.Client Official

SueSu01:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2020

## Description

1. Create a Apex daily test pipeline that can run a lot more tests and triggered by NuGet.Client PrivateDev.
2. We plan to add more test cases once the pipeline becomes stable.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


zivkan:
> Create a Apex daily test pipeline that can run a lot more tests and triggered by NuGet.Client Official.
> Once this PR is merged and the Apex daily pipeline starts running, we can stop the Apex Tests On Windows stage in NuGet.Client Official pipeline.

This makes sense if bugs introduced in the test scenarios are low impact/priority,so should not block the PR introducing the bug.

However, if there are VS scenarios that we should block pull requests on (for example, restore is completely broken, installing packages in VS is completely broken), then moving the VS tests to a different pipeline doesn't help, and only makes it harder to find the test results (since I now need to look in two different pipelines).

Additionally, in general the intent for "daily" tests are not to run after every commit/pull request, but to run only once a day. This way, you have your critical-scenario tests in your regular (every CI build) pipeline, and then you have the slow, exhaustive test suit in a different pipeline that runs infrequently.

So, this pull request is currently implementing something different to what the [linked issue](https://github.com/NuGet/Client.Engineering/issues/2020) specifies, and it's unclear to be what benefit there is to simply shifting the tests to another pipeline, as this PR currently does.

nkolev92:
> Once this PR is merged and the Apex daily pipeline starts running, we can stop the Apex Tests On Windows stage in NuGet.Client Official pipeline.

As @zivkan suggested, I don't think we'll want to disable the Apex Tests on Windows stage, but rather minimize the scope to critical scenarios only and migrate the more extensive tests to the daily pipeline.

eng/pipelines/templates/Apex_Tests_On_Windows_Daily.yml
```diff
@@ -0,0 +1,233 @@
...
+          displayName: 'Set RunSettingsURI variable'
+          name: SetRunSettingsURI
+          inputs:
+            targetType: 'inline'
+            script: |
+              try {
+                $branchName = "$(resources.pipeline.ComponentBuildUnderTest.sourceBranch)"
+                $branchName = $branchName.Replace('refs/heads/', '')
+
+                $RunSettingsURI = "https://vsdrop.corp.microsoft.com/file/v1/RunSettings/$(System.TeamProject)/NuGet-NuGet.Client-Trusted/$branchName/$(resources.pipeline.ComponentBuildUnderTest.runID);NuGet.Tests.Apex.runsettings"
```
nkolev92: We want to create a separate assembly for the daily tests, so we probably ened a separate run settings.

nkolev92 (commented):
@NuGet/nuget-client Do we want a PrivateDev version for the daily apex tests or do we just want to keep an official one?

eng/pipelines/dailytests.yml
```diff
@@ -0,0 +1,31 @@
+resources:
+  pipelines:
+  - pipeline: ComponentBuildUnderTest
+    source: NuGet.Client-Official
```
zivkan: > Do we want a PrivateDev version for the daily apex tests or do we just want to keep an official one?

We need a way to run the pipeline for PRs that add new tests, or attempt to fix failing tests, can be validated before merging. I think it would be terrible to need to run an official build for the PR branch, just to test the changes.

However, I don't think we need to run the scheduled tests on both official builds and PR builds. Honestly, I think we should use the PR/private builds as the source/trigger, not official builds. We're running those twice a day already, to calculate build reliability.

```suggestion
    source: NuGet.Client-PrivateDev
```

eng/pipelines/dailytests.yml
```diff
@@ -0,0 +1,31 @@
...
+    trigger:
+        branches:
+          - dev-*
```
zivkan:
```suggestion
          - dev
```

eng/pipelines/dailytests.yml
```diff
@@ -0,0 +1,31 @@
...
+  - delete
+  - stop
+
+extends:
+  template: templates/dailytests.yml
+  parameters:
+    DartLabEnvironment: ${{parameters.DartLabEnvironment}}
+    ApexAgentCleanup: ${{parameters.ApexAgentCleanup}}
+    NuGetLocalizationType: Full
+    isOfficialBuild: true
```
zivkan:
```suggestion
    isOfficialBuild: false
```

Official builds are those that create binaries that we ship to customers. Test pipelines don't create any artifacts that we ship to customers, so are never "official" builds.

For our official builds, we use the variable/parameter to prevent flaky tests breaking the build, and therefore making it impossible to create a release pipeline from the build. But since this test pipeline won't be the input of another pipeline, the same requirement doesn't exist.

eng/pipelines/dailytests.yml
nkolev92: Just to confirm what you're proposing:

- DailyTests will run based off of the PrivateDev definition as that's really the PR validation, where we're most likely to care about. 
- DailyTests will *not* run as part of the official definition.
- For "validation" purposes, we can use the DailyTests run *against* the dev.

eng/pipelines/dailytests.yml
nkolev92: So this would effectively trigger with the weekly builds?

eng/pipelines/dailytests.yml
zivkan: > DailyTests will run based off of the PrivateDev definition as that's really the PR validation, where we're most likely to care about.

Yes, I think so (I'm not 100% sure I'm reading your sentence the way you intended, but I think I am). But we also already have the dev branch scheduled twice a day in order to measure build reliability. So, the PrivateDev build definition is not just for PR builds, but also these twice a day scheduled builds.

> DailyTests will not run as part of the official definition.

Right. The daily tests are testing the product, not a specific build definition. PrivateDev and Official both build the same product code. So, in my mind, it doesn't matter which build definition we trigger from.

> For "validation" purposes, we can use the DailyTests run against the dev.

I have low confidence I understand what you mean by this. I assume by validation, you mean a pull request that modifies a daily test that is not run as part of the apex tests that run in the PrivateDev and Official pipelines. Also, if by "run against the dev" you mean "run against the PrivateDev PR build", then yes.

> Just to confirm what you're proposing

If my suggestion isn't accepted, then while I understand the intent of the new pipeline testing official builds, my question is how do we validate pull requests that modify daily tests, given the PrivateDev build definition is not going to run the daily tests?

As mentioned earlier, I'm exploiting the fact that we run dev branch builds in the PrivateDev pipeline regularly, so we can use it as a way to both automatically test the latest dev branch commits, as well as manually triggering the pipeline to test pull requests that modify daily tests.

eng/pipelines/dailytests.yml
zivkan: No, it would trigger every time a new commit is added to the `dev` branch. The way that this PR currently has it, `dev-*`, it would not trigger when the `dev` branch changes, but will only tigger on` dev-*` branch changes. Considering most of us push `dev-*` changes multiple times every day, this would be an enormous waste of resources, and then it wouldn't even test the product code that we ship (the dev/main branch).

eng/pipelines/dailytests.yml
nkolev92: > I have low confidence I understand what you mean by this. I assume by validation, you mean a pull request that modifies a daily test that is not run as part of the apex tests that run in the PrivateDev and Official pipelines. Also, if by "run against the dev" you mean "run against the PrivateDev PR build", then yes.

@zivkan I was more thinking, are we going to be able to track pass rates and avoid these extend packages to be out of sight, out of mind.

In the end, the idea with daily tests is that they should be passing on dev too, as a reference.

zivkan: Both the official build and PrivateDev pipelines build the same source code. So, the dev branch in one is the same as the dev branch in the other.

> these extend packages

I assume this is a typo, and you mean other branches?

If you want to find the test pass rate for the dev branch, ignoring all pull request branches and builds, then if you're using Azure Pipeline Analytics, use the filter and only select the dev branch.

nkolev92: I meant to say extended tests. 
I'm aware of the tools, I just wanted to make sure we're gonna get runs to validate that and we're actually gonna pay attention to them :D 
Anyway, I think we're going in circles here, I'm happy with it as is. 

We should *definitely* extend the documentation around this in the client engineering repo though. Describing basically what we discussed here.

zivkan: oh, you mean email notifications when a pipeline fails?   that's something we have to manually configure. If we had different pipelines for automatic daily runs and a different pipeline for validate a PR, then we'd set up an alert for all failires on the daily pipeline.  If we we a single pipeline for both scheduled runs, and also manually triggered PR branches, then I think we'd just need to configure where branch == dev in the email notification settings.

eng/pipelines/templates/Apex_Tests_On_Windows_Daily.yml
nkolev92: Can we please create a follow up for this?

eng/pipelines/templates/dailytests.yml
```diff
@@ -0,0 +1,52 @@
+
```
nkolev92: Do we need this extra template? Should it go into the apex tests on windows daily file.

@NuGet/nuget-client can someone with more CI authoring experience provide suggestions here?

nkolev92 (commented):
Looks good to me. 

Please create a follow up task for the run settings. 
We want to have 2 assemblies, once of a smaller/more limited set of tests and one of the more exhaustive tests.

eng/pipelines/templates/dailytests.yml
zivkan: FWIW, I agree. Our primary pipeline was split into many files because there are many (6 or 7?) stages. It used to be one long file, and IMO it was very difficult to manage, because it was so long. But this daily tests pipeline is a single stage, so I don't see the benefit in splitting it up.

eng/pipelines/templates/Apex_Tests_On_Windows_Daily.yml
SueSu01: Created a new _**NuGet.Tests.Apex.Daily.runsettings**_ for this pipeline.

eng/pipelines/templates/dailytests.yml
SueSu01: Removed the extra template _eng/pipelines/templates/dailytests.yml_ yml file, thanks.

test/NuGet.Tests.Apex/NuGet.Tests.Apex.Daily/NuGetEndToEndTests/NuGetUITestCase.cs
nkolev92, line 1: You probably don't need the `NuGetEndToEndTests` folder here. You can just put them in the root of the project folder.

nkolev92:
> Once this PR is merged and the Apex daily pipeline starts running, we can stop the Apex Tests On Windows stage in NuGet.Client Official pipeline.

Can we delete this from the description?
This is not a goal. We want to maintain a minimal apex tests test bed, not completely remove it from every build.

test/NuGet.Tests.Apex/NuGet.Tests.Apex.Daily/NuGetEndToEndTests/NuGetUITestCase.cs
```diff
@@ -0,0 +1,58 @@
...
+        private readonly SimpleTestPathContext _pathContext = new SimpleTestPathContext();
+
+        public NuGetUITestCase()
+            : base()
+        {
+        }
+
+        [TestMethod]
+        [Timeout(DefaultTimeout)]
+        public async Task TestPipeline()
```
nkolev92, line 27: I'm guessing this is something that gets deleted eventually correct?

I'd just keep the test as simple as possible, since it seems to function as a dummy test.

nkolev92 (approved):
My feedback can be addressed in a future PR. 

Thanks for getting this done. Really excited about it!

test/NuGet.Tests.Apex/NuGet.Tests.Apex.Daily/NuGetEndToEndTests/NuGetUITestCase.cs
SueSu01: > I'm guessing this is something that gets deleted eventually correct?
> 
> I'd just keep the test as simple as possible, since it seems to function as a dummy test.

Yes, it was added just to verify the new _NuGet.Tests.Apex.Daily.runsettings_ works well in Apex daily test pipeline, it will be removed when a valid test case is added.

SueSu01:
> > Once this PR is merged and the Apex daily pipeline starts running, we can stop the Apex Tests On Windows stage in NuGet.Client Official pipeline.
> 
> Can we delete this from the description? This is not a goal. We want to maintain a minimal apex tests test bed, not completely remove it from every build.

Deleted. Thanks for your reminder.

PR #5593, opened 2024-01-12
Further reduce allocations in CreateGraphNodeAsync

Erarndt:
## Bug

## Description

This PR makes additional changes to reduce allocations in `CreateGraphNodeAsync()` by addressing three issues highlighted in this trace:

image

A `LightweightList` was introduced in a previous PR to reduce the need to allocate a `List<T>` in the cases where the number of items was 4 or fewer. This value was chosen because it was seen as a safe starting point since it's the default size for `List<T>` as well as the 80th percentile in telemetry data

50 - 2,
70 - 3,
80 - 4
90 - 7
95 - 9

This trace indicates that there are still significant allocations occurring and this change allows `LightweightList` to contain up to 10 items before needing to allocate a `List<T>`. 

The second change introduces a helper method on `GraphNode` that enables callers to ensure that `InnerNodes` has sufficient capacity to add additional items to minimize resizing.

Finally, `FindRuntimeDependencies()` often returns a `List<T>` wrapped by the `IEnumerable<T>` interface which was resulting in boxing of the `List<T>.Enumerator`.  The existing `NoAllocEnumerate()` helper can be used here to avoid the boxing.

## PR Checklist

- [ ] PR has a linked issue.
- [ ] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.DependencyResolver.Core/GraphModel/GraphNode.cs
+++ b/src/NuGet.Core/NuGet.DependencyResolver.Core/GraphModel/GraphNode.cs
@@ -62,9 +62,13 @@ namespace NuGet.DependencyResolver
         /// <param name="additionalSpace">The count of additional items that will be added.</param>
         internal void EnsureInnerNodeCapacity(int additionalSpace)
         {
-            if (InnerNodes is List<GraphNode<TItem>> innerList && innerList.Capacity < innerList.Count + additionalSpace)
+            if (InnerNodes is List<GraphNode<TItem>> innerList)
             {
-                innerList.Capacity = innerList.Count + additionalSpace;
+                int requiredCapacity = innerList.Count + additionalSpace;
+                if (innerList.Capacity < requiredCapacity)
+                {
+                    innerList.Capacity = requiredCapacity;
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


nkolev92 (dismissed):

jeffkl (dismissed):

src/NuGet.Core/NuGet.DependencyResolver.Core/GraphModel/GraphNode.cs
```diff
@@ -55,6 +55,19 @@ internal bool AreAllParentsRejected()
...
+        /// <summary>
+        /// Ensures that <see cref="InnerNodes"/> has the capacity to add
+        /// <paramref name="additionalSpace"/> more items.
+        /// </summary>
+        /// <param name="additionalSpace">The count of additional items that will be added.</param>
+        internal void EnsureInnerNodeCapacity(int additionalSpace)
+        {
+            if (InnerNodes is List<GraphNode<TItem>> innerList && innerList.Capacity < innerList.Count + additionalSpace)
+            {
+                innerList.Capacity = innerList.Count + additionalSpace;
```
zivkan: `innerList.Count + additionalSpace` is calculated twice. You could assign it to a variable and use it twice.

jeffkl (approved):

PR #5601, opened 2024-01-18
Change to allow style warning to show up in build and fix all unused …

jgonz120:
## Bug
N/A

Fixes:

## Description

We currently have a set of code analyzer rules but they aren't checked during builds, this can lead to PRs with comments like "remove unnecessary using statement". In this PR I've enabled enforcing the code styles in build and removed the warning for several existing issues to help reduce the scope of this PR. So, while there are 466 file changes, they are only removing unused using statements. 

In the future you would be getting a build error in the IDE informing you about the using statements. The most difficult part of this was identifying the using statements where were conditionally used based on the framework selected. I'm not certain how we want to treat those as I saw several variations on how they were previously handled. 

### Ignored warnings
[CS1591](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs1591): Missing XML comment for publicly visible type or member
[IDE0076](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0076): Remove invalid global 'SuppressMessageAttribute' 
[CS1573](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs1573): Parameter has no matching param tag in the XML comment
[CS1574](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs1574): A string passed to a cref tag referred to a member that is not available within the current build environment.
[CS1587](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs1587): XML comment is not placed on a valid language element
[CS1572](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs1572): XML comment has a param tag for a parameter that doesn't exist
[CS1584](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs1584): One of the parameters passed to a tag for documentation comments has invalid syntax
[CS1658](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs1658): Warning text for an error that was overwritten as a warning. The errors were compiler errors found within xml comments. 
[CS0419](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0419): Cref in comment has an ambiguous reference
## PR Checklist

- [ ] PR has a linked issue.
- [X] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [X] N/A 

- **Documentation**
  
  - **OR**
  - [X] N/A


Directory.Build.props
```diff
@@ -9,6 +9,11 @@
     <Import Project="build\config.props" />
     <Import Project="build\common.project.props" />
   </ImportGroup>
+  <PropertyGroup>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+    <NoWarn>$(NoWarn);CS1591;IDE0076;CS1573;CS1574;CS1587;CS1570;CS1572;CS1584;CS1658;CS0419;IDE0055;EnableGenerateDocumentationFile</NoWarn>
```
zivkan, line 15: We already set common project properties, such as NoWarn, elsewhere in our build scripts: https://github.com/search?q=repo%3ANuGet%2FNuGet.Client+NoWarn+path%3A**%2F*.props&type=code

I'd rather we kept such build properties in a single place, even if it's not following conventions that most other .NET repos use (the file is also imported immediately before this new property group)

However, I think NoWarn on code analyzers is functionality equivalent to disabling them, right? We won't see the violations anywhere, not even underlined in VS.  Roslyn implemented a better way to configure analyzers in the `.editorconfig` file, where you can set the analyzer to info level (or some other option), which shouldn't cause build failures, but will still visibly underline the code in the IDE.

However, can we also consider keeping the analyzers, and instead suppress all existing violations, to prevent new violations from being added?

Disabling an analyzer (IMO preferably though `.editorconfig`) for analyzers we really don't ever want is a good idea, but does not appear to be intentional given this PR's title and description. For analyzers that we think are a good idea, but too much to fix all existing violations in a single PR, I hope global or inline suppressions will allow us to use the analyzer to prevent further violations.

src/NuGet.Core/NuGet.Commands/TrustedSignersCommand/TrustedSignersCommandRunner.cs
```diff
@@ -10,9 +10,11 @@
 using System.Threading.Tasks;
 using NuGet.Common;
 using NuGet.Configuration;
+#if IS_SIGNING_SUPPORTED
 using NuGet.Packaging;
-using NuGet.Packaging.Signing;
 using NuGet.Protocol;
+#endif
+using NuGet.Packaging.Signing;
```
zivkan, line 17: Personally, I'd rather see the conditional using statements to be moved after the unconditional statements, so that they're all one continuous block. This change breaks the "keep using alphabetical" rule.

Also, I'm curious if someone modifies the file and needs to add a new using, will the intellicode auto fix add the new using statement alphabetically or not? 

My comment applies to everywhere that conditional using statments are added, but this is the first one I've seen that makes the usings non-alphabetical.

src/NuGet.Core/NuGet.Packaging/Signing/Timestamp/Rfc3161TimestampUtils.cs
```diff
@@ -3,11 +3,11 @@
 // See the LICENSE file in the project root for more information.
 
 using System.Globalization;
-using System.Reflection;
-using System.Runtime.InteropServices;
 
 #if IS_DESKTOP
 using System.Security.Cryptography.Pkcs;
+using System.Reflection;
+using System.Runtime.InteropServices;
```
zivkan, line 10: The namespaces within the `#if IS_DESKTOP` block are not sorted alphabetically.

Directory.Build.props
```diff
@@ -9,6 +9,11 @@
     <Import Project="build\config.props" />
     <Import Project="build\common.project.props" />
   </ImportGroup>
+  <PropertyGroup>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
```
zivkan, line 14: Does this make `git format` in CI redundant?  If so, we should remove it from the pipeline: https://github.com/NuGet/NuGet.Client/blob/bd0764b4a5b78eea77c61f5b82078623c4fc0cfe/eng/pipelines/templates/Static_Source_Analysis.yml#L77-L81

Directory.Build.props
zivkan: Having written what I did, I believe that @nkolev92 dislikes huge commits that touch many files without functional changes, because it harms his debugging workflow. I rarely use git history to investigate bugs, and I've never used `git bisect`.  So, while I'd prefer to optimize for long term code maintenance, we should see what middle ground we can find with people who want to avoid the types of changes I porposed.

Directory.Build.props
nkolev92: We already set it to true: https://github.com/NuGet/NuGet.Client/blob/bd0764b4a5b78eea77c61f5b82078623c4fc0cfe/build/common.project.props#L113

This was already attempted before: https://github.com/NuGet/NuGet.Client/commit/d888411098c720d83be5a0ef750c51c74f1ef2b0

It doesn't fully work anyways. There are many bugs: https://github.com/dotnet/roslyn/issues?q=is%3Aissue+EnforceCodeStyleInBuild+is%3Aopen


In particular: https://github.com/dotnet/roslyn/issues/49439 and https://github.com/dotnet/roslyn/issues/41640

Tldr; style analyzers don't quite work if the documentation file is not generated, but there are many places in our builds where we explicitly disable the documentation files so that they don't get packed.

Directory.Build.props
```diff
@@ -9,6 +9,11 @@
     <Import Project="build\config.props" />
     <Import Project="build\common.project.props" />
   </ImportGroup>
+  <PropertyGroup>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
nkolev92, line 13: Don't set unconditional properties. Might not apply to this property, but overwrites can lead to issues.

The changes would also normally go to common.project.props.

Directory.Build.props
nkolev92: In the past when we added analyzers, @kartheekp-ms should be able to help out with tihs, we put all the suppressions in globalsupressions.
It basically allows us to keep the analyzers and it does not change code that's not affected. 

Re: Additional changes, my argument tends to revolve around git blame and changing lines that don't need change. As is the PR just has removals, and potentially additions with global suppressions, so I have no concerns with that.

nkolev92 (commented):
EnforceCodestyleInBuild is already enabled, but does not fully work because of roslyn limitations. 

Unfortunately it's not the most reliable thing at this point. 

We should also try not to suppress analyzers, but rather suppress violations and leave the analyzers on for future changes.

PR #5602, opened 2024-01-23
Improve dotnet package search JSON format output

### LINKED ISSUE #13158: [Dotnet Package Search] The search result of the package should be “version” instead of “latestVersion” when executing command “dotnet package search <Package Name> --exact-match --format json”
### NuGet Product Used

dotnet.exe

### Product Version

.NET SDK 9.0.100-alpha.1.24062.11

### Worked before?

It’s not a regression issue since it is a new feature.

### Impact

#### Repro Steps:  
1.Patch the latest .NET SDK version ([9.0.100-alpha.1.24062.11](https://github.com/dotnet/installer?tab=readme-ov-file#use-dotnet---help-to-see-available-commands-or-visit-httpsakamsdotnet-cli)) with NuGet Client [Pipelines - Run 6.9.0.86 artifacts (visualstudio.com)](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=8897544&view=artifacts&pathAsName=false&type=publishedArtifacts). 
   1.1 Download the [Entropy](https://github.com/NuGet/Entropy.git) to the desktop of test machine.
   1.2 Download and extract the NuGet Client build “[6.9.0.86](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=8897544&view=artifacts&pathAsName=false&type=publishedArtifacts)->nupkgs - NonRTM” (named to “nupkgs”) to the desktop of test machine.
   1.3 Update the $SDKChannel = "LTS" and $SDKVersion = "9.0.100-alpha.1.24062.11" in the “patchOnWindows.ps1” file of the downloaded “Entropy\SDKPatchTool”.
   1.4 Modify the $patchSDKFolder = "C:\Users\[UserName]\Desktop\PatchedSDK" and $nupkgsPath = "C:\Users\v-luzh\Desktop\nupkgs" (the path of downloaded nupkgs in step 1.2) in the “patchOnWindows.ps1” file of the downloaded “Entropy\SDKPatchTool”.
2. After patching finished, go to the patchedSDK folder and execute the command `.\dotnet package search NUnit --exact-match --format json` in powershell.

#### Expected Results:  
The search result of the package should be “version” instead of “latestVersion” since one package only has one latest version. Like the result of the searching the same package with “--format table” by default as below.
image

#### Actual Results:
The search result of the package shows a lot of “latestVersion” which does not make sense as below screenshot.
image

#### Note:
Repro rate: 100%

### LINKED ISSUE #13165: The search result of the package should be “totalDownloads” instead of “total downloads” when executing command “dotnet package search <Package Name> --format json”
### NuGet Product Used

dotnet.exe

### Product Version

.NET SDK 9.0.100-alpha.1.24062.11

### Worked before?

N/A

### Impact

### Repro Steps & Context

.\dotnet.exe package search <package_name> --format json
(also repros with --exact-match)

The json output of the dotnet package search should have total downloads be a single word, rather than two words with a space. This would allow accessing the property with dot notation rather than having to rely on deferencing the parsed json with brackets.

This issue occurs both when running with and without the --exact-match flag.

Expected:
output would have "totalDownloads", "total_downloads" or something similar, that is a single word.

Actual:
image

### Verbose Logs

_No response_

Nigusu-Allehu:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13165
Fixes: https://github.com/NuGet/Home/issues/13158

## Description

This PR has three things it is doing which are corelated
1. Clean up table output code as per the suggestion here : https://github.com/NuGet/NuGet.Client/pull/5525#discussion_r1442413884 
which involves removing various string that are used to represent the different possible headers we could have based on te verbosity value and using a Factory class.
2. In the json output making sure where we list out all the versions for a specific package, use "version" as a property name instead of "latestVersion"
3. In the json output user "totalDownloads" as a property name instead of "total downloads"

Json formatting was implemented using class models for each package in the search result.
I had the following types that would be used to represent packages
- Detailed verbosity, exact match json
- Detailed verbosity, non exact match
- Normal verbosity, exact match json
- Normal verbosity, non exact match
- minimal verbosity, exact match
- minimal verbosity, non exact match

 However, I made a design change in the code to use a custom JsonConverter in system.text.json instead.
The custom JsonConverter would check the verbosity and exact match values to print out the required package information.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/JsonFormat/MinimalVerbosityJsonFormatSearchResultPackage.cs
```diff
@@ -2,22 +2,21 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Text.Json.Serialization;
-using NuGet.CommandLine.XPlat.Commands.PackageSearch;
 using NuGet.Protocol.Core.Types;
 
 namespace NuGet.CommandLine.XPlat
 {
-    internal class MinimalVerbosityPackage : ISearchResultPackage
+    internal class MinimalVerbosityJsonFormatSearchResultPackage : ISearchResultPackage
```
nkolev92: This name is a handful. What's wrong with the previous name?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/JsonFormat/DetailedVerbosityJsonFormatSearchResultPackage.cs
```diff
@@ -0,0 +1,39 @@
+using System;
+using System.Linq;
+using System.Text.Json.Serialization;
+using NuGet.Protocol.Core.Types;
+
+namespace NuGet.CommandLine.XPlat
+{
+    internal class DetailedVerbosityJsonFormatSearchResultPackage : NormalVerbosityJsonFormatSearchResultPackage
```
nkolev92: This name is very verbose.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/TableFormat/DetailedTableFormatStrategy.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace NuGet.CommandLine.XPlat
+{
+    internal class DetailedTableFormatStrategy : ITableFormatStrategy
+    {
+        private readonly string[] _detailedVerbosityTableHeader = { "Package ID", "Latest Version", "Owners", "Total Downloads", "Vulnerable", "Deprecation", "Project URL", "Description" };
+        private readonly string[] _detailedVerbosityTableHeaderForExactMatch = { "Package ID", "Version", "Owners", "Total Downloads", "Vulnerable", "Deprecation", "Project URL", "Description" };
+        private readonly int[] _detailedColumnsToHighlight = { 0, 2, 6, 7 };
+
+        public Table CreateTable(IEnumerable<IPackageSearchMetadata> results, bool exactMatch)
```
nkolev92: You're not using results here.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/TableFormat/MinimalTableFormatStrategy.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace NuGet.CommandLine.XPlat
+{
+    internal class MinimalTableFormatStrategy : ITableFormatStrategy
+    {
+        private readonly string[] _minimalVerbosityTableHeader = { "Package ID", "Latest Version" };
+        private readonly string[] _minimalVerbosityTableHeaderForExactMatch = { "Package ID", "Version" };
+        private readonly int[] _minimalColumnsToHighlight = { 0 };
+
+        public Table CreateTable(IEnumerable<IPackageSearchMetadata> results, bool exactMatch)
```
nkolev92: same here.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultTableRenderer.cs
```diff
@@ -44,42 +32,8 @@ public void Add(PackageSource source, IEnumerable<IPackageSearchMetadata> comple
...
-                    table = new Table(_normalColumnsToHighlight, _normalVerbosityTableHeaderForExactMatch);
-                }
-                else
-                {
-                    table = new Table(_normalColumnsToHighlight, _normalVerbosityTableHeader);
-                }
-            }
-
+            ITableFormatStrategy strategy = TableFormatStrategyFactory.GetStrategy(_verbosity);
+            Table table = strategy.CreateTable(completedSearch, _exactMatch);
```
nkolev92: If exact match changes the columns, isn't it a strategy itself, rather than something that's part of the table creation?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/JsonFormat/JsonFormatFactorySearchResultPackage.cs
```diff
@@ -0,0 +1,38 @@
...
+
+using NuGet.Protocol.Core.Types;
+
+namespace NuGet.CommandLine.XPlat
+{
+    internal class JsonFormatFactorySearchResultPackage
+    {
+        public static ISearchResultPackage GetPackage(IPackageSearchMetadata packageSearchMetadata, PackageSearchVerbosity verbosity, bool exactMatch, string deprecation)
+        {
+            if (exactMatch)
```
nkolev92: What's the difference between exact match and non-exact match in terms of rendering?
Basically why do we need the indirection that leads to 3 extra types for exact match?

Nigusu-Allehu: Exact-match operation in a sense is totally different from non-exact match operation. Basically, exact match lists all the available versions for a package. However, a non-exact-match operation on the other hand looks up a search phrase and lists available packages. Rendering the otput from the two can be a simillar operation because in both cases we list packages. Howeve, the only difference comes in when we are rendering the version property. For exact match, the property name is "version", however, for non-exact match operations, it is "latestVersion".

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/JsonFormat/DetailedVerbosityJsonFormatSearchResultPackage.cs
Nigusu-Allehu: In the naming of this type we would like to show two properties mainly: verbosity and whether it is exact-match or not.
would `DetailedJsonSearchResultPackage` and `ExactDetailedJsonSearchResultPackage` be better names than the current `DetailedVerbosityJsonFormatSearchResultPackage ` and `ExactMatchDetailedVerbosityJsonFormatSearchResultPackage `?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/JsonFormat/NormalVerbosityJsonFormatSearchResultPackage.cs
```diff
@@ -6,17 +6,17 @@
 
 namespace NuGet.CommandLine.XPlat
 {
-    internal class NormalVerbosityPackage : MinimalVerbosityPackage
+    internal class NormalVerbosityJsonFormatSearchResultPackage : MinimalVerbosityJsonFormatSearchResultPackage
     {
-        [JsonPropertyName("total downloads")]
+        [JsonPropertyName("totalDownloads")]
```
donnie-msft: Rather than using a string, you should be able to use the JsonProperties from Protocol. 
I did this here: https://github.com/NuGet/NuGet.Client/blob/001ea1a/src/NuGet.Core/NuGet.Protocol/Model/PackageSearchMetadata.cs#L67

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/JsonFormat/NormalVerbosityJsonFormatSearchResultPackage.cs
```diff
@@ -6,17 +6,17 @@
...
         public long? Downloads { get; set; }
 
         [JsonPropertyName("owners")]
         public string Owners { get; set; }
```
donnie-msft: I recently merged a change to Protocol that you could consider echoing here. The property is `OwnersList` and provides an enumerable deserialized from the JSON. The `Owners` is representing a comma-delimited list of the items in the `OwnersList`.
https://github.com/NuGet/NuGet.Client/blob/001ea1a/src/NuGet.Core/NuGet.Protocol/Model/PackageSearchMetadata.cs#L67

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/JsonFormat/NormalVerbosityJsonFormatSearchResultPackage.cs
```diff
@@ -6,17 +6,17 @@
 
 namespace NuGet.CommandLine.XPlat
 {
-    internal class NormalVerbosityPackage : MinimalVerbosityPackage
+    internal class NormalVerbosityJsonFormatSearchResultPackage : MinimalVerbosityJsonFormatSearchResultPackage
```
donnie-msft: nit: Can you help me understand why this inheritance structure is needed? 

My thinking: 
I agree that these types build upon one another (it's probably unlikely we'd ever need to add/remove a property from a lower level, say, `Minimal` formatting, which would actually need to be available in a higher level formatting, like `Detailed`.

However, is anything gained by this structure besides avoiding copy/pasting code for the properties? I can't think of how testing would be easier, or how developing with these classes would be clear. (I'd still have to validate whether I had a NormalVerbosityJson object and then know which properties that gains me).

One suggestion would be to just keep a reference to your `IPackageSearchMetadata` and then only expose the properties that are relevant. No inheritance would be necessary.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/TableFormat/NormalTableFormatStrategy.cs
```diff
@@ -0,0 +1,28 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using NuGet.Protocol.Core.Types;
+
+namespace NuGet.CommandLine.XPlat
+{
+    internal class NormalTableFormatStrategy : ITableFormatStrategy
+    {
+
```
donnie-msft: nit: whitespace

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/TableFormat/TableFormatStrategyFactory.cs
```diff
@@ -0,0 +1,25 @@
...
+            switch (verbosity)
+            {
+                case PackageSearchVerbosity.Minimal:
+                    return new MinimalTableFormatStrategy();
+                case PackageSearchVerbosity.Normal:
+                    return new NormalTableFormatStrategy();
+                case PackageSearchVerbosity.Detailed:
+                    return new DetailedTableFormatStrategy();
+                default:
+                    throw new ArgumentException("Invalid verbosity level");
```
donnie-msft, line 21: Probably should localize this if it's intended for a response to customer typos.

Nigusu-Allehu: This is internal and it should never happen unless there is something wrong with our code.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/JsonFormat/DetailedVerbosityJsonFormatSearchResultPackage.cs
jeffkl, line 1: Team Review:  Try using a custom formatter instead of a class for each level of verbosity

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/JsonFormat/NormalVerbosityJsonFormatSearchResultPackage.cs
zivkan: I also **strongly** encourage you to try to parse the json in multiple programming languages and get a feel for how to interact with the results.  Json property names with characters that are invalid variable names are very problematic in some languages, so if you want the json to be maximally useful to customers, you need to design the json schema to be maximally compatible with all these different restrictions.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/JsonFormat/DetailedVerbosityJsonFormatSearchResultPackage.cs
Nigusu-Allehu: Used a custom JsonConverter and removed all the classes that were used for serialization.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/JsonFormat/NormalVerbosityJsonFormatSearchResultPackage.cs
Nigusu-Allehu: Yes, we could have an `ownersList` property instead of `owners`. However, should we do that in a separate PR?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/JsonFormat/SearchResultPackage.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using NuGet.Protocol.Core.Types;
+
+namespace NuGet.CommandLine.XPlat
+{
+    internal class SearchResultPackage
+    {
+        public IPackageSearchMetadata PackageSearchMetadata { get; set; }
+        public string DeprecationMessage { get; set; }
```
nkolev92: Isn't DeprecationMEssage a part of PackageSearchMetadata? 
I'm guessing we don't need it here. 

That'd mean that we probably don't need this class anymore.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/JsonFormat/SearchResultPackagesConverter.cs
```diff
@@ -0,0 +1,76 @@
...
+                {
+                    writer.WriteBoolean("vulnerable", true);
+                }
+                else
+                {
+                    writer.WriteNull("vulnerable");
+                }
+
+                writer.WriteString(JsonProperties.ProjectUrl, value.PackageSearchMetadata.ProjectUrl.ToString());
+                writer.WriteString(JsonProperties.Deprecation, value.DeprecationMessage);
```
nkolev92: Seeing how deprecation is used, I'm more confident in my feedback that you probably don't need the extra property.

nkolev92 (commented):
Minor feedback.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/JsonFormat/SearchResultPackagesConverter.cs
Nigusu-Allehu: It is because deprecation metadata needs to be retrieved asynchronously. However, the write method for a JsonConverter is synchronous. As a result, I decided to retrieve the deprecation message in the `PackageSearchResultJsonRenderer` class before running the serializer.

nkolev92: Just do a .Result. 
The calculation is only delayed in Visual Studio scenarios.

nkolev92: you can get opinions from @NuGet/nuget-client if you're not sure.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/JsonFormat/SearchResultPackage.cs
Nigusu-Allehu: Fixed

nkolev92 (dismissed):

jeffkl (dismissed):

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/JsonFormat/SearchResultPackagesConverter.cs
```diff
@@ -0,0 +1,87 @@
...
+                    writer.WriteBoolean("vulnerable", true);
+                }
+                else
+                {
+                    writer.WriteNull("vulnerable");
+                }
+
+                writer.WriteString(JsonProperties.ProjectUrl, value.ProjectUrl.ToString());
+
+                PackageDeprecationMetadata packageDeprecationMetadata = value.GetDeprecationMetadataAsync().Result;
```
kartheekp-ms, line 72: Isn't `value.GetDeprecationMetadataAsync().Result` a blocking call?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultJsonRenderer.cs
```diff
@@ -16,32 +13,30 @@ internal class PackageSearchResultJsonRenderer : IPackageSearchResultRenderer
...
             PackageSearchResult packageSearchResult = new PackageSearchResult(source.Name);
 
             foreach (IPackageSearchMetadata metadata in completedSearch)
             {
-                if (_verbosity == PackageSearchVerbosity.Minimal)
-                {
-                    packageSearchResult.Packages.Add(new MinimalVerbosityPackage(metadata));
-                }
-                else if (_verbosity == PackageSearchVerbosity.Detailed)
+                string deprecation = "";
```
kartheekp-ms: looks like this is an unused variable. Is it needed here? If yes, can we `string.Empty` instead of  `""`?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultJsonRenderer.cs
```diff
@@ -16,32 +13,30 @@ internal class PackageSearchResultJsonRenderer : IPackageSearchResultRenderer
...
+
+                if (_verbosity == PackageSearchVerbosity.Detailed)
                 {
                     PackageDeprecationMetadata packageDeprecationMetadata = await metadata.GetDeprecationMetadataAsync();
```
kartheekp-ms: Looks like we are making the network call here to `metadata.GetDeprecationMetadataAsync()` and we are not using the method result. If we can somehow address this comment, then I guess we don't need to make another blocking call in the new `SearchResultPackagesConverter` code. Please let me know If I am missing something.

kartheekp-ms (commented):
Left few comments to being with. I am still reviewing this PR.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultJsonRenderer.cs
Nigusu-Allehu: https://github.com/NuGet/NuGet.Client/pull/5602#discussion_r1491862567 
Previously I had it so that the aync call to `metadata.GetDeprecationMetadataAsync() `  is called in the `PackageSearchResultJsonRenderer.cs`. However, that mean I had to create a class containing both `IPackageSearchMetadata` and deprecation message for each package. Instead of that, the new implementation does a blocking call to `GetDeprecationMetadataAsync()` . That means we can remove the new class I had created to represent packages.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/JsonFormat/SearchResultPackagesConverter.cs
Nigusu-Allehu: @kartheekp-ms what's your opinion?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchResultJsonRenderer.cs
nkolev92: `.Result` isn't necessarily bad in this situation. 
The API probably isn't the best. There's never any actual calculation happening in the async method.
There's also no deadlock concerns here.

nkolev92 (dismissed):

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/JsonFormat/SearchResultPackagesConverter.cs
```diff
@@ -0,0 +1,87 @@
...
+
+                if (packageDeprecationMetadata != null)
+                {
+                    writer.WriteString(JsonProperties.Deprecation, value.GetDeprecationMetadataAsync().Result.Message);
```
kartheekp-ms: Can we reuse the result from line no. 72 instead of making another call to the same  `GetDeprecationMetadataAsync()` method?

kartheekp-ms (approved):
I am in the process of developing an alternative design, and it is taking a bit longer than expected. I will let you merge this pull request if you have addressed all the unresolved comments. Later today, I will add a comment with the alternative design. If the team agrees that the new design is good enough, you can address it in a follow-up pull request.

PR #5605, opened 2024-01-23
Protocol loads Owners from package sources as an enumerable type

### LINKED ISSUE #13186: Owner package metadata is an array in JSON but a string in Protocol types
### NuGet Product(s) Affected

Other/NA

### Current Behavior

Owner metadata is provided as an array, but Protocol immediately converts this to a comma-delmited string upon parsing the data.

### Desired Behavior

Preserve the array of owners so that the enumeration can be used by NuGet Client.

### Additional Context

_No response_

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13186

## Description

Read Owners collection from package sources into an enumerable type, rather than converting it to a comma-delimited string immediately upon parsing the JSON response.

There's no effective customer change in this PR, just that the enumeration is retained in Protocol types rather than only as a concatenated string. In subsequent PRs, I'll utilize this enumeration of owners for customer benefit.

Test changes:
- Ensure a package source response that has Owners is converted into a matching enumeration.
- Ensure a package source response without Owners returns null and doesn't affect other metadata.
- Update `EntityFrameworkSearch.json` (from stale 2017 data) with latest package search metadata result response from nuget.org.
- Update test with fluent assertions for all properties in the updated JSON.
- Update `test/NuGet.Core.Tests/NuGet.CommandLine.Xplat.Tests/PackageSearchRunnerTests.cs` which was failing as it attempted to assert that the Table Output contained owners. The registration JSON does not normally contain Owner, so I updated the JSON and the test accordingly.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Core/NuGet.Protocol/Model/PackageSearchMetadata.cs
```diff
@@ -63,8 +63,9 @@ public PackageIdentity Identity
         public Uri LicenseUrl { get; private set; }
 
         [JsonProperty(PropertyName = JsonProperties.Owners)]
-        [JsonConverter(typeof(MetadataFieldConverter))]
-        public string Owners { get; private set; }
+        public IEnumerable<string> OwnersEnumerable { get; private set; }
+
+        public string Owners => OwnersEnumerable != null ? string.Join(", ", OwnersEnumerable.Where(s => !string.IsNullOrWhiteSpace(s))) : null;
```
nkolev92: This computes every time it's accessed. 
Can we cached it after the first computation?

src/NuGet.Core/NuGet.Protocol/Model/IPackageSearchMetadata.cs
```diff
@@ -27,6 +27,7 @@ public interface IPackageSearchMetadata
         Uri ReportAbuseUrl { get; }
         Uri PackageDetailsUrl { get; }
         DateTimeOffset? Published { get; }
+        IEnumerable<string> OwnersEnumerable { get; }
```
nkolev92: Can we use a list? It's better for perf.

src/NuGet.Core/NuGet.Protocol/Model/PackageSearchMetadata.cs
```diff
@@ -63,8 +63,9 @@ public PackageIdentity Identity
         public Uri LicenseUrl { get; private set; }
 
         [JsonProperty(PropertyName = JsonProperties.Owners)]
-        [JsonConverter(typeof(MetadataFieldConverter))]
-        public string Owners { get; private set; }
+        public IEnumerable<string> OwnersEnumerable { get; private set; }
```
nkolev92: OwnerList?

Enumerable is expressing the type a bit too much.

nkolev92 (dismissed):
Please get reviews from others, but this looks just fine to me.

src/NuGet.Core/NuGet.Protocol/Model/IPackageSearchMetadata.cs
```diff
@@ -27,6 +27,7 @@ public interface IPackageSearchMetadata
         Uri ReportAbuseUrl { get; }
         Uri PackageDetailsUrl { get; }
         DateTimeOffset? Published { get; }
+        List<string> OwnersList { get; }
```
zivkan: All the other properties in this interface are read-only (no setters, and data types which can't be mutated). Therefore, I think `IReadOnlyList<string>` is more appropriate.

it means that we shouldn't use `foreach` to iterate. Well, maybe if we use the `NoAllocEnumerate()` utility method, but a basic `for` loop would ensure zero allocation enumeration of the list.

src/NuGet.Core/NuGet.Protocol/Model/LocalPackageSearchMetadata.cs
```diff
@@ -42,6 +42,8 @@ public LocalPackageSearchMetadata(LocalPackageInfo package)
 
         public Uri LicenseUrl => Convert(_nuspec.GetLicenseUrl());
 
+        public List<string> OwnersList => null;
```
zivkan: This can be computed from `Owners`. While we might not use it in our own code, from an API design point of view, it would look like a bug if this returns null (or even an empty list) when `Owners` returns a non-empty value.

src/NuGet.Core/NuGet.Protocol/Model/PackageSearchMetadataV2Feed.cs
```diff
@@ -85,6 +84,8 @@ public PackageSearchMetadataV2Feed(V2FeedPackageInfo package, MetadataReferenceC
 
         public Uri LicenseUrl { get; private set; }
 
+        public List<string> OwnersList => null;
```
zivkan: A considerable number of customers still use nuget.org's V2 URL.  Will future PM UI changes not work for these customers that use V2 feeds because of this?

src/NuGet.Core/NuGet.Protocol/Model/LocalPackageSearchMetadata.cs
donnie-msft: Am I misunderstanding Local metadata? I believe it's always metadata from a `nuspec` on the client machine; therefore, it should never have OwnersList, since OwnersList is a package source concept (Search API).

src/NuGet.Core/NuGet.Protocol/Model/PackageSearchMetadataV2Feed.cs
donnie-msft: V2 does not currently return Owners like V3 does. 

I checked with the server team and it seems they do not backport features to V2. So, I'm not sure I see a problem for future PM UI, since the existing behavior will not change. That is, the package source nuget.org V2 will never return Owners.

zivkan: Owners was added to both the nuspec, and also NuGet.Client's internal data model before NuGet went 1.0: https://github.com/NuGet/NuGet2/commit/a6f17b03736ba178a21fe422603442eff8ee8d65

Did you validate that the V2 protocol doesn't return owners?  It would really surprise me if it's not already being returned. After all, the concept of owners has existed ever since nuget.org has.

src/NuGet.Core/NuGet.Protocol/Model/LocalPackageSearchMetadata.cs
zivkan: There was a time when owners wasn't deprecated from the nuspec. Some customers might be using it for reasons other than what nuget.org uses it for. But if we want to maintain the line that owners is now deprecated, then perhaps this method should throw a `NotSupportedException`, and the `Owners` property should also be marked as obsolete, so customers using NuGet.Protocol in their own apps are aware.

src/NuGet.Core/NuGet.Protocol/Model/PackageSearchMetadataV2Feed.cs
donnie-msft: Yes, I know you're not available offline, but if you get a moment, I can send you a sample response from nuget.org that I captured with Fiddler. There's no Owners. 

`OwnersList` is not intended to represent `Owners` from the nuspec. It's only a Search API result of `owners` from https://learn.microsoft.com/en-us/nuget/api/search-query-service-resource#search-result.
The code you linked is for a nuspec Owner, which is not the same thing.

src/NuGet.Core/NuGet.Protocol/Model/LocalPackageSearchMetadata.cs
donnie-msft: You're talking about `Owners` but you're pointing at `OwnersList`. I think `NotSupportedException` is reasonable, as local metadata will never support a server's metadata. 
 
Marking `Owners` obsolete seems like a separate concern than what this PR does, which is _adds_ server metadata. We can discuss if we're ready to officially make the nuspec field obsolete. Regardless, that will be a separate Spec and separate PR to this one.

src/NuGet.Core/NuGet.Protocol/Model/IPackageSearchMetadata.cs
donnie-msft: Done!

src/NuGet.Core/NuGet.Protocol/Model/LocalPackageSearchMetadata.cs
donnie-msft: ~I've updated `LocalPackageSearchMetadata` to throw when accessing the list.~
EDIT: unfortunately, I can't update `LocalPackageSearchMetadata` to throw due to the same reason below...

I'd like to update `PackageSearchMetadataV2Feed` as well, but there's a design problem that needs addressed.
The utility for V2 feeds here:
https://github.com/NuGet/NuGet.Client/blob/4b618ca3c42fa0b678b28d17af2d815c1d26d28b/src/NuGet.Core/NuGet.Protocol/LegacyFeed/V2FeedUtilities.cs#L25

...causes the Cloned metadata to copy properties here (which is fine for V3 feeds, but not for V2 feeds):
https://github.com/NuGet/NuGet.Client/blob/4b618ca3c42fa0b678b28d17af2d815c1d26d28b/src/NuGet.Core/NuGet.Protocol/Model/PackageSearchMetadataBuilder.cs#L95

Since I don't want this code to throw `NotImplementedException`, a V2 feed will need to continue passing `null`.

src/NuGet.Core/NuGet.Protocol/GlobalSuppressions.cs
```diff
@@ -230,3 +230,4 @@
...
 [assembly: SuppressMessage("Globalization", "CA1308:Normalize strings to uppercase", Justification = "We convert packageId to lower case for api call.", Scope = "member", Target = "~M:NuGet.Protocol.PackageMetadataResourceV3.LoadRegistrationIndexAsync(NuGet.Protocol.HttpSource,System.Uri,System.String,NuGet.Protocol.Core.Types.SourceCacheContext,System.Func{NuGet.Protocol.HttpSourceResult,System.Threading.Tasks.Task{NuGet.Protocol.Model.RegistrationIndex}},NuGet.Common.ILogger,System.Threading.CancellationToken)~System.Threading.Tasks.Task{System.ValueTuple{NuGet.Protocol.Model.RegistrationIndex,NuGet.Protocol.Core.Types.HttpSourceCacheContext}}")]
 [assembly: SuppressMessage("Usage", "CA2237:Mark ISerializable types with serializable", Justification = "<Pending>", Scope = "type", Target = "~T:NuGet.Protocol.PackageNotFoundProtocolException")]
+[assembly: SuppressMessage("Usage", "CA2227:Collection properties should be read only", Justification = "<Pending>", Scope = "member", Target = "~P:NuGet.Protocol.Core.Types.PackageSearchMetadataBuilder.ClonedPackageSearchMetadata.OwnersList")]
```
zivkan: Since the property has been changed to `IReadOnlyList`, I expect this suppression is no longer needed.

Also, if it wasn't, the justification should be filled out. We allowed the `<Pending>` justification when we enabled the code analyzers, because there was hundreds being added in one PR, it was too much effort to justify every one. New code should explain why we have a code analyzer enabled, but being ignored.

src/NuGet.Core/NuGet.Protocol/Model/PackageSearchMetadataV2Feed.cs
zivkan: > OwnersList is not intended to represent Owners from the nuspec.
> It's only a Search API result of owners
> The code you linked is for a nuspec Owner, which is not the same thing.

These are all different serializations of the same conceptual information. And the `IPackageSearchMetadata` interface represents both HTTP package metadata and nuspec package metadata (which could be a design bug, but it also reduces memory allocations, since HTTP search results and local nuspec files have so much overlap). And when using multiple feeds with the "all" package source selected, it all gets merged, local and http feeds. Why does it matter that it's code for a different serialization?

We've fixed multiple bugs in the past, and I'm sure we'll have more in the future, due to unintuitive API design. And we've had multiple customers using the NuGet SDK in their own apps report bugs to us because our APIs are unintuitive to them, and their apps behave differently to NuGet client does. Therefore, we can hopefully reduce support workloads by trying to make our public APIs as unsurprising as possible. Customers seeing a type with both `Owners` and `OwnersList` members, one of which returns null when the other is not, is not intuitive.

As mentioned in the other comment thread, you could at least mark the type as `Obsolete`, so anyone using this type directly will get compiler warnings. Although if the type is ever downcast to `IPackageSearchMetadata` the warning will be lost at call sites, but at least some mitigation to warn developers that this property is never used in this data type would exist. 

Also, I just checked NuGet.Server, which is an implementation of the V2 protocol. nuget.org's v2 protocol might not return owners, but [NuGet.Server's implementation sure appears to](https://github.com/search?q=repo%3ANuGet%2FNuGet.Server%20owners&type=code)

src/NuGet.Core/NuGet.Protocol/Model/LocalPackageSearchMetadata.cs
zivkan: > what this PR does, which is adds server metadata

No it's not, otherwise the HTTP model wouldn't have removed the `JsonProperty` attribute from the existing `Owners` property, stopped using a JSON converter than converts between the serialized list and the data model's single string, and the new `OwnersList` property would need a different `JsonProperty` attribute, with a different property name. For the same reason, it also wouldn't have been possible to compute the `Owner` property from the `OwnerList` property in PakageSearchMetadata.cs.

Rather, this PR is adding a new representation, or a new data model, for the existing owner metadata. We've had bugs in the past, and I'm sure will continue to have bugs in the future, where someone writing some new code didn't understand the API design of something done in the past, and they might not notice the bug during development because there are too many different scenarios to test everything exhaustively (and none of the devs can probably list every single nuget feature exhaustively, I know I can't).

Plus we sometimes get customers using the NuGet SDK in their own app report bugs to us, again usually because the API is implemented unintuitively, and official NuGet clients work around the API design in other business logic.

So, the `PackageSearchMetadata` data model computing one property from the other is a kind of precedent, so I think following [the principal of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment), the other implementations of the interface should do so as well, even if we believe it's not going to be used. None of us know every single NuGet scenario, every way different customers are using NuGet, or unexpected interactions between different features.

Marking the property/properties as obsolete and/or throwing an exception was one idea to minimize "astonishment" about different APIs having different behaviours, despite being about the same conceptual data. It didn't work work out, but I still think a little extra code now will reduce future work, rather than silently returning `null` without an easy way for developers (either working on NuGet or external customers using the SDK) to discover why. PackageSearchMetadata only needed a few lines of code to compute one property from the other, so it seems reasonable that the same still be true for the local package metadata and the v2 package metadata.

donnie-msft: >  and they might not notice the bug during development

Are you seeing a bug, or think something may be a bug, or is this just hypothetical?

I've offered in a comment last week to meet with you and discuss this PR directly. I think it may be more efficient to do so. Could you reach out to me with a time that works?

src/NuGet.Core/NuGet.Protocol/GlobalSuppressions.cs
donnie-msft: It's still necessary and I've added justification. There's code that sets property references in the Cloned package metadata class.

Here's where Owners is being set today on dev. OwnersList and all other properties do the same thing.
https://github.com/NuGet/NuGet.Client/blob/98b7d827d91c055fcaa77be7f34bf66811381b5a/src/NuGet.Core/NuGet.Protocol/Model/PackageSearchMetadataBuilder.cs#L95


And here's what happens if the suppression is removed:
image

donnie-msft:
This PR is blocked since Andy has feedback and appears to be unavailable for some time. My best understanding at present is that the feedback is around potential and ongoing confusion of what "Owners" means:

> Customers seeing a type with both Owners and OwnersList members, one of which returns null when the other is not, is not intuitive.

I agree that NuGet's 3 meanings of "Owners" is not intuitive. Perhaps someone can suggest an alternative name that they'd approve?
If others can chime in with their feedback on this topic, along with any suggestions, please do so!

donnie-msft:
Chatted offline with Andy. Going to investigate V2 private Hosted Feeds and their behavior with Owners.

donnie-msft:
I wasn't able to see Owners coming through a NuGet.Server using V2 protocol. 
Andy also mentioned offline that this code is ignoring Owners, which I believe lessens his original concern.
https://github.com/NuGet/NuGet.Client/blob/98b7d827d91c055fcaa77be7f34bf66811381b5a/src/NuGet.Core/NuGet.Protocol/LegacyFeed/V2FeedParser.cs#L375

However, I'm now exposing OwnersList for V2 metadata as a preventative / forward-compatible approach.

Local metadata is also now populating the OwnersList from the serialized Owners, so to reduce any confusion by SDK consumers.

zivkan (dismissed):

src/NuGet.Core/NuGet.Protocol/Model/LocalPackageSearchMetadata.cs
```diff
@@ -50,7 +50,7 @@ public IReadOnlyList<string> OwnersList
             {
                 if (_ownersList is null)
                 {
-                    _ownersList = Owners != null ? Owners.Split(',').Select(s => s.Trim()).ToList() : null;
+                    _ownersList = Owners != null ? Owners.Split(',').Where(s => !string.IsNullOrWhiteSpace(s)).Select(s => s.Trim()).ToList() : null;
```
zivkan, line 53: You can have `string.Split` not generate the emprty elements in the first place: https://learn.microsoft.com/en-us/dotnet/api/system.string.split?view=net-8.0#system-string-split(system-char()-system-stringsplitoptions)

edit: string.split's option probably wouldn't work for the test you have, since it's not empty, just whitespace, so your code is needed to make that test pass

zivkan (approved):

PR #5609, opened 2024-01-24
Mark method that gets properties via DTE as obsolete

zivkan:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2664

Regression? no

## Description

Visual Studio's performance tests count the number of exceptions thrown (even if caught) and flag a regression if the number increases. DTE thrown an exception when a project doesn't have a value for a property. Therefore, we get flagged every time we add a new property and use the API that falls back to DTE when the newer API doesn't return a value.

So, mark the API as obsolete, and suppress all instances of the API where it's currently being called. Hopefully this discourages new features from using the API that uses DTE.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/VsProjectJsonNuGetProject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/VsProjectJsonNuGetProject.cs
@@ -43,7 +43,10 @@ namespace NuGet.PackageManagement.VisualStudio
         protected override async Task<string> GetMSBuildProjectExtensionsPathAsync()
         {
             await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+#pragma warning disable CS0618 // Type or member is obsolete
+            // Need to validate no project systems get this property via DTE, and if so, switch to GetPropertyValue
             var msbuildProjectExtensionsPath = _vsProjectAdapter.BuildProperties.GetPropertyValueWithDteFallback(ProjectBuildProperties.MSBuildProjectExtensionsPath);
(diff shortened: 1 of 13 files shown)


src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/VsProjectJsonNuGetProject.cs
```diff
@@ -72,17 +75,23 @@ protected override async Task UpdateInternalTargetFrameworkAsync()
                 var jsonTargetFramework = targetFramework as NuGetFramework;
                 if (IsUAPFramework(jsonTargetFramework))
                 {
+#pragma warning disable CS0618 // Type or member is obsolete
+                    // Need to validate no project systems get this property via DTE, and if so, switch to GetPropertyValue
```
nkolev92, line 79: We have some telemetry on when the fallback is being used correct?
Does that telemetry contain the property? 

IOW, are we attempting to validate this?

nkolev92 (approved):

jeffkl (approved):

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/VsProjectJsonNuGetProject.cs
zivkan: Just a question of time and prioritization. I created a follow up issue: https://github.com/NuGet/Client.Engineering/issues/2670

PR #5612, opened 2024-01-25
Do not allocate unnecessary lists of no warn log codes in LibraryDependency

### LINKED ISSUE #13184: LibraryDependency creates a collection that is mostly empty
### NuGet Product Used

dotnet.exe, MSBuild.exe

### Product Version

All

### Worked before?

_No response_

### Impact

None

### Repro Steps & Context

The `LibraryDependency` creates a list for warnings but in a lot of code paths it ends up being empty so there's no need to allocate the list and instead it could be an empty singleton.

https://github.com/NuGet/NuGet.Client/blob/a6fff16af6f44927a83851448a99beb1d2801600/src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs#L22

Originally reported internally: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1826369

### Verbose Logs

_No response_

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13184

## Description

The `LibraryDependency` class hold a collection of NoWarn codes and initializes it as a `List<T>` which allocates.  Some code paths then parse the user specified codes and then adds them to the list while other code paths simply replace the list allocated at the time of construction.  This change adds an overload to the constructor that accepts an `IList<NuGetLogCode>` so that callers can specify the exact list they want.  In some code paths the callers now specify `Array,Empty<NuGetLogCode>()` to avoid allocations.

I also refactored `MSBuildStringUtility.Split()` to return an `IList<T>` instead of `IEnumerable<T>` so that callers don't have to call `.ToList()` on it.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception - Existing tests cover this refactoring
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/ProjectServices/VsManagedLanguagesProjectSystemServices.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/ProjectServices/VsManagedLanguagesProjectSystemServices.cs
@@ -204,7 +204,7 @@ namespace NuGet.PackageManagement.VisualStudio
         private static LibraryDependency ToPackageLibraryDependency(PackageReference reference, bool isCpvmEnabled)
         {
             // Get warning suppressions
-            List<NuGetLogCode> noWarn = MSBuildStringUtility.GetNuGetLogCodes(GetReferenceMetadataValue(reference, ProjectItemProperties.NoWarn)).ToList();
+            IList<NuGetLogCode> noWarn = MSBuildStringUtility.GetNuGetLogCodes(GetReferenceMetadataValue(reference, ProjectItemProperties.NoWarn));
 
             var dependency = new LibraryDependency(noWarn)
             {
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/ProjectServices/VsManagedLanguagesProjectSystemServices.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/ProjectServices/VsManagedLanguagesProjectSystemServices.cs
@@ -203,7 +203,10 @@ namespace NuGet.PackageManagement.VisualStudio
 
         private static LibraryDependency ToPackageLibraryDependency(PackageReference reference, bool isCpvmEnabled)
         {
-            var dependency = new LibraryDependency
+            // Get warning suppressions
+            List<NuGetLogCode> noWarn = MSBuildStringUtility.GetNuGetLogCodes(GetReferenceMetadataValue(reference, ProjectItemProperties.NoWarn)).ToList();
+
+            var dependency = new LibraryDependency(noWarn)
             {
                 AutoReferenced = MSBuildStringUtility.IsTrue(GetReferenceMetadataValue(reference, ProjectItemProperties.IsImplicitlyDefined)),
                 GeneratePathProperty = MSBuildStringUtility.IsTrue(GetReferenceMetadataValue(reference, ProjectItemProperties.GeneratePathProperty)),
@@ -212,7 +215,7 @@ namespace NuGet.PackageManagement.VisualStudio
                 LibraryRange = new LibraryRange(
                     name: reference.Name,
                     versionRange: ToVersionRange(reference.Version, isCpvmEnabled),
-                    typeConstraint: LibraryDependencyTarget.Package)
+                    typeConstraint: LibraryDependencyTarget.Package),
             };
 
             MSBuildRestoreUtility.ApplyIncludeFlags(
@@ -220,13 +223,6 @@ namespace NuGet.PackageManagement.VisualStudio
                 GetReferenceMetadataValue(reference, ProjectItemProperties.IncludeAssets),
                 GetReferenceMetadataValue(reference, ProjectItemProperties.ExcludeAssets),
                 GetReferenceMetadataValue(reference, ProjectItemProperties.PrivateAssets));
-
-
-            // Add warning suppressions
-            foreach (var code in MSBuildStringUtility.GetNuGetLogCodes(GetReferenceMetadataValue(reference, ProjectItemProperties.NoWarn)))
-            {
-                dependency.NoWarn.Add(code);
-            }
 
             return dependency;
         }
--- a/src/NuGet.Core/NuGet.Common/MsBuildStringUtility.cs
+++ b/src/NuGet.Core/NuGet.Common/MsBuildStringUtility.cs
@@ -85,19 +85,32 @@ namespace NuGet.Common
         }
 
         /// <summary>
-        /// Splits and parses a ; or , delimited list of log codes.
-        /// Ignores codes that are unknown.
+        /// Parses the specified string as a comma or semicolon delimited list of NuGet log codes and ignores unknown codes.
         /// </summary>
-        public static IEnumerable<NuGetLogCode> GetNuGetLogCodes(string s)
+        /// <param name="s">A comma or semicolon delimited list of NuGet log codes.</param>
+        /// <returns>An <see cref="IList{T}" /> containing the <see cref="NuGetLogCode" /> values that were successfully parsed from the specified string.</returns>
+        public static IList<NuGetLogCode> GetNuGetLogCodes(string s)
         {
-            foreach (var item in MSBuildStringUtility.Split(s, ';', ','))
+            // The Split() method already checks for an empty string and returns Array.Empty<string>().
+            string[] split = MSBuildStringUtility.Split(s, ';', ',');
+
+            if (split.Length == 0)
             {
-                if (item.StartsWith("NU", StringComparison.OrdinalIgnoreCase) &&
-                    Enum.TryParse<NuGetLogCode>(value: item, ignoreCase: true, result: out var result))
+                return Array.Empty<NuGetLogCode>();
+            }
+
+            List<NuGetLogCode> logCodes = new List<NuGetLogCode>(capacity: split.Length);
+
+            for (int i = 0; i < split.Length; i++)
+            {
+                if (split[i].StartsWith("NU", StringComparison.OrdinalIgnoreCase) &&
+                    Enum.TryParse(value: split[i], ignoreCase: true, out NuGetLogCode logCode))
                 {
-                    yield return result;
+                    logCodes.Add(logCode);
                 }
             }
+
+            return logCodes;
         }
 
         /// <summary>
--- a/src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
+++ b/src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using NuGet.Common;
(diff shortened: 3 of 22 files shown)


src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/ProjectServices/VsManagedLanguagesProjectSystemServices.cs
```diff
@@ -203,7 +203,10 @@ private static string GetReferenceMetadataValue(ProjectReference reference, stri
 
         private static LibraryDependency ToPackageLibraryDependency(PackageReference reference, bool isCpvmEnabled)
         {
-            var dependency = new LibraryDependency
+            // Get warning suppressions
+            List<NuGetLogCode> noWarn = MSBuildStringUtility.GetNuGetLogCodes(GetReferenceMetadataValue(reference, ProjectItemProperties.NoWarn)).ToList();
```
nkolev92: Can the original method return a list instead to avoid the now guaranteed allocation.

nkolev92: I'm comfortable making a breaking change in that API.

src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
```diff
@@ -50,8 +51,16 @@ public class LibraryDependency : IEquatable<LibraryDependency>
...
         public LibraryDependency()
+            : this(new List<NuGetLogCode>())
         {
-            NoWarn = new List<NuGetLogCode>();
+        }
+
+        /// <summary>Initializes a new instance of the LibraryDependency class with the specified NoWarn codes.</summary>
+        /// <param name="noWarn">Specifies a <see cref="List{T}" /> containing <see cref="NuGetLogCode" /> values.</param>
+        /// <remarks>Required properties must be set when using this constructor.</remarks>
+        public LibraryDependency(IList<NuGetLogCode> noWarn)
```
nkolev92, line 61: Add SetsRequiredMembers?

nkolev92, line 61: I'm wondering if we can just require LibraryRange and NoWarn to be set.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/ProjectServices/VsManagedLanguagesProjectSystemServices.cs
jeffkl: I was considering doing this in a separate change.  Do you prefer I make that change together with this?

nkolev92: Do whatever's easier. 

Personally, I don't mind if they're 1 change.

jeffkl: Done

src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
jeffkl: This constructor does not set required members so this isn't needed

jeffkl: I'm not sure if its worth it.  By default the object will have an empty list that callers can populate.  Our codebase will always specify it.  Would it be a breaking change to make NoWarn `required`?

src/NuGet.Core/NuGet.Common/MsBuildStringUtility.cs
```diff
@@ -85,19 +85,32 @@ public static bool IsTrueOrEmpty(string? value)
...
+        /// Parses the specified string as a comma or semicolon delimited list of NuGet log codes and ignores unknown codes.
         /// </summary>
-        public static IEnumerable<NuGetLogCode> GetNuGetLogCodes(string s)
+        /// <param name="s">A comma or semicolon delimited list of NuGet log codes.</param>
+        /// <returns>An <see cref="IList{T}" /> containing the <see cref="NuGetLogCode" /> values that were successfully parsed from the specified string.</returns>
+        public static IList<NuGetLogCode> GetNuGetLogCodes(string s)
         {
-            foreach (var item in MSBuildStringUtility.Split(s, ';', ','))
+            // The Split() method already checks for an empty string and returns Array.Empty<string>().
+            string[] split = MSBuildStringUtility.Split(s, ';', ',');
```
Erarndt, line 95: I wouldn't be surprised if the allocs in `MSBuildStringUtility.Split()` show up here. Here's how it's implemented:

        public static string[] Split(string s, params char[] chars)
        {
            if (!string.IsNullOrEmpty(s))
            {
                // Split on ; and trim all entries
                // After trimming remove any entries that are now empty due to trim.
                return s.Split(chars, StringSplitOptions.RemoveEmptyEntries)
                    .Select(entry => entry.Trim())
                    .Where(entry => entry.Length != 0)
                    .ToArray();
            }
            return Array.Empty<string>();
        }

Each call that actually does any work is going to allocate:

- A `char[]` because of params
- a `string[]` returned from `Split()`
- an `int[]` the size of the input `string` allocated internally by `Split()`
- An alloc for each `string` in the `string[]`
- Boxing of the `char[]` enumerator from `Select()`
- Potentially new strings from the call to `Trim()`
- a new `string[]` from the call to `ToArray()`

We've seen lots of allocations arising from code that splits strings, and like the code below, the string may not even be needed (only needed if it starts with `"NU"`). We ended up writing a helper function that splits the string without any allocations unless you actually need to materialize the string. Instead, you get a struct enumerator of the results of the split operation as `ReadOnlySpan<char>`. If this is allocating a sufficient amount (here or cumulatively in all `Split()` calls), I'd consider porting the split helper.

src/NuGet.Core/NuGet.Common/MsBuildStringUtility.cs
```diff
@@ -85,19 +85,32 @@ public static bool IsTrueOrEmpty(string? value)
...
 
         /// <summary>
-        /// Splits and parses a ; or , delimited list of log codes.
-        /// Ignores codes that are unknown.
+        /// Parses the specified string as a comma or semicolon delimited list of NuGet log codes and ignores unknown codes.
         /// </summary>
-        public static IEnumerable<NuGetLogCode> GetNuGetLogCodes(string s)
+        /// <param name="s">A comma or semicolon delimited list of NuGet log codes.</param>
+        /// <returns>An <see cref="IList{T}" /> containing the <see cref="NuGetLogCode" /> values that were successfully parsed from the specified string.</returns>
+        public static IList<NuGetLogCode> GetNuGetLogCodes(string s)
```
Erarndt, line 92: Switching this to return `IList<NuGetLogCode>` may not be a win here. Looks like a lot of callers just want to enumerate the results (there is at least one that needs a `List`). Usage of `yield return` is going to allocate a state machine and creating a `List` allocates the list object and backing array. Depending on the size, I wouldn't be surprised if `yield return` won out.

src/NuGet.Core/NuGet.ProjectModel/PackageSpecReferenceDependencyProvider.cs
```diff
@@ -306,7 +306,7 @@ private List<LibraryDependency> GetDependenciesFromSpecRestoreMetadata(PackageSp
                 // Note: Only add in dependencies that are in the filtered list to avoid getting the wrong TxM
                 dependencies.AddRange(childReferences
                     .Where(reference => filteredExternalDependencies.Contains(reference.ProjectName))
-                    .Select(reference => new LibraryDependency
+                    .Select(reference => new LibraryDependency(noWarn: Array.Empty<NuGetLogCode>())
                     {
                         LibraryRange = new LibraryRange
```
Erarndt, line 311: Not directly related to this change, but how often do we end up parsing the same `VersionRange`? Can we have a singleton for `"1.0.0"`?

src/NuGet.Core/NuGet.Common/MsBuildStringUtility.cs
jeffkl: A lot of the code paths are enumerating the results and calling `List<T>.Add()` so this would be better right?

src/NuGet.Core/NuGet.Common/MsBuildStringUtility.cs
jeffkl: Can you please point me to the Split helper you're referring to?

zivkan: https://github.com/NuGet/NuGet.Client/blob/e520f969f09411cdbf024349cbae3a1f60c4c73c/src/NuGet.Core/NuGet.Common/MsBuildStringUtility.cs#L25-L37

src/NuGet.Core/NuGet.ProjectModel/PackageSpecReferenceDependencyProvider.cs
zivkan: Using NuGet.Insights to check all packages on nuget.org, of packages that have at least one dependency, almost exactly 1% of those dependencies are for version 1.0.0 (well, `[1.0.0, )`). That doesn't mean the same will be true for this code path (project references where the other project doesn't specify a version will end up with 1.0.0 from memory), but I think nuget.org dependencies is the only data we have.

src/NuGet.Core/NuGet.Common/MsBuildStringUtility.cs
jeffkl: Decided not to work on a change with `MsBuildStringUtility.Split()` for this pull request.  We should address it in a subsequent change.

src/NuGet.Core/NuGet.Common/MsBuildStringUtility.cs
jeffkl: Talked offline, the point of this PR is to address calls to `new LibraryDependency()` not allocating a list and instead getting the list at construction.  This change does that but doesn't address places where the result of GetLogCodes() is used as an `IEnumerable>

src/NuGet.Core/NuGet.ProjectModel/PackageSpecReferenceDependencyProvider.cs
jeffkl: Thanks for getting that data.  Sounds like we'd want to pursue any such change in a different pull request anyway.

src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
```diff
@@ -50,8 +51,16 @@ public class LibraryDependency : IEquatable<LibraryDependency>
         /// <summary>Initializes a new instance of the LibraryDependency class.</summary>
         /// <remarks>Required properties must be set when using this constructor.</remarks>
         public LibraryDependency()
+            : this(new List<NuGetLogCode>())
```
kartheekp-ms, line 54: Can this simply be `Array.Empty<NuGetLogCode>()` so that we can avoid invoking the new constructor—the one that accepts `IList<NuGetLogCode> noWarn`? This approach would eliminate the need to invoke the new overloaded constructor by just passing an empty collection in so many places.

jeffkl: Unfortunately no, if you have an existing code path that does this:

```c#
var libraryDependency = new LibraryDependency();

libraryDependency.NoWarn.Add(NuGetLogCode.NU1000);
```

it will break because the list is empty and when you call `Add()` you get an exception about no capacity.  So I left this constructor in place but all of NuGet's code uses the other constructor to signal if the NoWarn codes should be an exact list it specifies or an empy list that it won't be adding to.

kartheekp-ms: We could potentially address the above limitation by making `NoWarn` a read-only collection. Is it because of public API requirements that we do not want to change the types?

jeffkl: Yes, I was trying to avoid making a breaking change.  We could make a breaking change later, for this change I'm just trying to address unnecessary memory allocations.

kartheekp-ms (approved):
There are some unresolved comments in this pull request. Please feel free to create an issue to address them if applicable.

PR #5614, opened 2024-01-26
Improve VS project API telemetry to include property name

zivkan:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2670

Regression? No

## Description

The previously telemetry only told us if at least 1 property per project system used the ProjectBuildProperties or DTE APIs, but not which properties used which API.  This API gives us the info per property (and still per project system), so we'll know which properties we can migrate to the new API exclusively.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Telemetry/VsProjectBuildPropertiesTelemetry.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Telemetry/VsProjectBuildPropertiesTelemetry.cs
@@ -1,8 +1,11 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+#nullable enable
+
 using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.Linq;
 using NuGet.Common;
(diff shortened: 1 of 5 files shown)


src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Telemetry/VsProjectBuildPropertiesTelemetry.cs
```diff
@@ -15,49 +18,73 @@ namespace NuGet.PackageManagement.VisualStudio.Telemetry
...
+            foreach (var projectSystem in _projectSystems)
+            {
+                foreach (var property in projectSystem.Value)
+                {
+                    ApiUsage apiUsage = property.Value;
+                    items.Add(property.Value);
+                }
+            }
+
+            e.ComplexData["ProjectBuildProperties"] = items;
```
zivkan, line 81: I'm not sure if I should change the property name, so our queries can easily filter out old telemetry without the property name. Or if we should just filter out telemetry where the property name is missing.

nkolev92: I'm not too concerned with the breaking change. We can handle it on the telemetry end, but also we're very likely to make decisions based off of the latest usage of the tooling anyways, since that's the one with more actionable telemetry.

nkolev92 (approved):

PR #5616, opened 2024-01-30
Introduce TaskResultCache to handle caching of task results and fix issue with continuations happening on the same thread

### LINKED ISSUE #13206: Caching task results can cause all continuations to occur on the same thread
### NuGet Product Used

MSBuild.exe, dotnet.exe

### Product Version

All

### Worked before?

_No response_

### Impact

None

### Repro Steps & Context

During a restore, NuGet's code has several places where it caches a task for reuse by graph walks.  Some large restores which normally take 30 minutes are taking up to 3 hours.  Analysis of the memory dump shows thousands of tasks waiting to be scheduled and only one task running at a time.

One such code path is here: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.DependencyResolver.Core/ResolverUtility.cs#L238

The issue is around how we're using a `Dictionary<TKey, Task<TValue>` or `ConcurrentDictionary<TKey, Task<TValue>>`.

Consider the following code:

```C#
ConcurrentDictionary<string, Task> taskCache = new ConcurrentDictionary<string, Task>(StringComparer.OrdinalIgnoreCase);

for (int i = 0; i < Iterations; i++)
{
    int num = i;
    Task.Run(() => WorkerTask(num, taskCache));
}

Console.ReadLine();

void CachedTask()
{
    LogCacheTaskStart();
    Thread.Sleep(DelayInMilliseconds);
    LogCacheTaskStop();
}

async Task WorkerTask(int iteration, ConcurrentDictionary<string, Task> taskCache)
{
    await Task.Delay(Random.Next(100, 400));
    LogWorkerTaskStart(iteration);
    if (!taskCache.TryGetValue("1", out Task task))
    {
        task = taskCache.GetOrAdd("1", Task.Run(CachedTask));
    }
    await task;
    LogWorkerTaskWorking(iteration);
    Thread.Sleep(DelayInMilliseconds);
    LogWorkerTaskStop(iteration);
}
```

This will spawn N threads which all are waiting for a Task in the cache to complete.  All worker threads receive the same Task and all of them resume on the same thread, causing execution to all happen on the same CPU core.

Output:
```
Worker task 2 started, waiting on cached task (Thread 13)...
Starting caching task (Thread 14)...
Worker task 8 started, waiting on cached task (Thread 5)...
Worker task 7 started, waiting on cached task (Thread 7)...
Worker task 4 started, waiting on cached task (Thread 4)...
Worker task 5 started, waiting on cached task (Thread 8)...
Worker task 6 started, waiting on cached task (Thread 12)...
Worker task 9 started, waiting on cached task (Thread 6)...
Worker task 0 started, waiting on cached task (Thread 9)...
Worker task 1 started, waiting on cached task (Thread 3)...
Worker task 3 started, waiting on cached task (Thread 11)...
Cached task complete (Thread 14).
Cached task complete, worker task 2 doing work (Thread 14)...
Worker task 2 complete (Thread 14)
Cached task complete, worker task 8 doing work (Thread 14)...
Worker task 8 complete (Thread 14)
Cached task complete, worker task 7 doing work (Thread 14)...
Worker task 7 complete (Thread 14)
Cached task complete, worker task 4 doing work (Thread 14)...
Worker task 4 complete (Thread 14)
Cached task complete, worker task 5 doing work (Thread 14)...
Worker task 5 complete (Thread 14)
Cached task complete, worker task 6 doing work (Thread 14)...
Worker task 6 complete (Thread 14)
Cached task complete, worker task 9 doing work (Thread 14)...
Worker task 9 complete (Thread 14)
Cached task complete, worker task 0 doing work (Thread 14)...
Worker task 0 complete (Thread 14)
Cached task complete, worker task 1 doing work (Thread 14)...
Worker task 1 complete (Thread 14)
Cached task complete, worker task 3 doing work (Thread 14)...
Worker task 3 complete (Thread 14)
```

You can clearly see that each worker task is resuming on the same thread at the cached task completed on and the worker threads all execute serially.

The problem seems to be how the continuations happen on the same thread.  The solution is to use a `TaskCompletionSource<TValue>` instead.

```c#
Dictionary<string, Task> taskCache = new Dictionary<string, Task>(StringComparer.OrdinalIgnoreCase);

for (int i = 0; i < Iterations; i++)
{
    int num = i;
    Task.Run(() => WorkerTask(num, taskCache));
}

Console.ReadLine();

return Task.CompletedTask;

async Task CachedTaskAsync()
{
    LogCacheTaskStart();
    await Task.Delay(DelayInMilliseconds);
    LogCacheTaskStop();
}

async Task WorkerTask(int iteration, Dictionary<string, Task> taskCache)
{
    LogWorkerTaskStart(iteration);

    if (!taskCache.TryGetValue("1", out Task task))
    {
        lock (taskCache)
        {
            if (!taskCache.TryGetValue("1", out task))
            {
                TaskCompletionSource<int> taskCompletionSource = new TaskCompletionSource<int>(TaskCreationOptions.RunContinuationsAsynchronously);

                CachedTaskAsync().ContinueWith((task) =>
                {
                    if (task.IsCanceled)
                    {
                        taskCompletionSource.SetCanceled();
                    }
                    else if (task.IsFaulted)
                    {
                        taskCompletionSource.SetException(task.Exception);
                    }
                    else
                    {
                        taskCompletionSource.SetResult(0);
                    }
                });

                task = taskCache["1"] = taskCompletionSource.Task;
            }
        }
    }
    await task;
    LogWorkerTaskWorking(iteration);
    Thread.Sleep(DelayInMilliseconds);
    LogWorkerTaskStop(iteration);
}
```

That code uses `ContinueWith()` and a `TaskCompletionSource` which specifies that continuations should happen asynchronously and has the following output:
```
Worker task 7 started, waiting on cached task (Thread 4)...
Worker task 0 started, waiting on cached task (Thread 3)...
Worker task 2 started, waiting on cached task (Thread 5)...
Worker task 6 started, waiting on cached task (Thread 12)...
Worker task 3 started, waiting on cached task (Thread 7)...
Worker task 8 started, waiting on cached task (Thread 11)...
Worker task 4 started, waiting on cached task (Thread 10)...
Worker task 5 started, waiting on cached task (Thread 8)...
Worker task 9 started, waiting on cached task (Thread 14)...
Worker task 1 started, waiting on cached task (Thread 6)...
Starting caching task (Thread 4)...
Cached task complete (Thread 13).
Cached task complete, worker task 0 doing work (Thread 12)...
Cached task complete, worker task 4 doing work (Thread 10)...
Cached task complete, worker task 7 doing work (Thread 13)...
Cached task complete, worker task 2 doing work (Thread 11)...
Cached task complete, worker task 3 doing work (Thread 9)...
Cached task complete, worker task 6 doing work (Thread 7)...
Cached task complete, worker task 5 doing work (Thread 3)...
Cached task complete, worker task 8 doing work (Thread 4)...
Cached task complete, worker task 9 doing work (Thread 5)...
Cached task complete, worker task 1 doing work (Thread 8)...
Worker task 5 complete (Thread 3)
Worker task 4 complete (Thread 10)
Worker task 1 complete (Thread 8)
Worker task 9 complete (Thread 5)
Worker task 6 complete (Thread 7)
Worker task 3 complete (Thread 9)
Worker task 2 complete (Thread 11)
Worker task 7 complete (Thread 13)
Worker task 8 complete (Thread 4)
Worker task 0 complete (Thread 12)
```

Now all worker tasks resume on a different thread and continue to execute in parallel.

### Verbose Logs

_No response_

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13206

## Description

In several parts of our code, we have a caching mechanism for Task results:

```c#
Dictionary<TKey, Task<TValue>> cache;
```

This works well as multiple callers receive the same task to await.  Unfortunately, it appears that continuations can happen on the same thread causing all callers to continue serially rather than on different threads in parallel.  From my testing, it appears to only be a problem on .NET Framework.  I've developed a prototype application that shows the behavior.  I'm not 100% certain why we don't see the issue all of time as it appears to happen intermittently.  Some large repositories with thousands of projects report seeing restore times that are usually around 30 minutes take over 2 hours.  Memory dumps show thousands of tasks in the scheduler waiting to run.

The fix appears to schedule the task with a continuation flag indicating that it should happen asynchronously:

```c#
public Task<TValue> GetOrAddAsync(TKey key, bool refresh, Func<Task<TValue>> valueFactory, CancellationToken cancellationToken)
{
    if (!refresh && _cache.TryGetValue(key, out Task<TValue>? result))
    {
        return result;
    }

    lock (_cache)
    {
        if (!refresh && _cache.TryGetValue(key, out result))
        {
            return result;
        }

        result = valueFactory().ContinueWith(
                     (task, state) => task.GetAwaiter().GetResult(),
                     state: null,
                     cancellationToken,
                     TaskContinuationOptions.RunContinuationsAsynchronously,
                     TaskScheduler.Default);

        _cache[key] = result;

        return result;
    }
}
```

To accomplish this, I added a new abstraction for this functionality, a `TaskResultCache<TKey, TValue>` and updated parts of our code that were using a Task caching construct such as this. 

To test this change, I built the fix locally and deployed them to a large internal repository.  It did not seem to make restore any faster but eliminate the runaway restores which were taking several hours.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception - Existing tests cover refactoring and I deployed the changes to a repository which restores over 2,000 projects without any issues.
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


build/Shared/TaskResultCache.cs
jeffkl, line 1: @stephentoub do you have any advice here?  Our code base wants to run a bunch of work in parallel and each thread is trying to cache the result of work involved in resolving NuGet packages.  The current implementation seems to experience continuations on the same thread causing serial work.  This pull request uses a `Dictionary<TKey, Task<TValue>>` with locks and a `ContinueWith()` which seems to fix the problem.  Does this pattern look okay?  Thanks in advance for your insights.

build/Shared/TaskResultCache.cs
```diff
@@ -0,0 +1,126 @@
...
+        /// </summary>
+        /// <param name="key">The key for the async operation to get or store in the cache.</param>
+        /// <param name="refresh"><see langword="true" /> to force the specified asynchronous operation to be executed and stored in the cache even if a cached operation exists, otherwise <see langword="false" />.</param>
+        /// <param name="valueFactory">A <see cref="Func{TResult}" /> to execute asynchronously if a cached operation does not exist.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken" /> to use for signaling that an operation should be cancelled.</param>
+        /// <returns>A <see cref="Task{TResult}" /> for the specified asynchronous operation from the cache if found, otherwise the scheduled asynchronous operation to await.</returns>
+        public Task<TValue> GetOrAddAsync(TKey key, bool refresh, Func<Task<TValue>> valueFactory, CancellationToken cancellationToken)
+        {
+            if (!refresh && _cache.TryGetValue(key, out Task<TValue>? result))
```
stephentoub: This could be reading the Dictionary while another thread holding the lock is mutating the Dictionary... that's not safe.

build/Shared/TaskResultCache.cs
```diff
@@ -0,0 +1,126 @@
...
+            lock (_cache)
+            {
+                if (!refresh && _cache.TryGetValue(key, out result))
+                {
+                    return result;
+                }
+
+                result = valueFactory().ContinueWith(TaskComplete, state: null, cancellationToken, TaskContinuationOptions.RunContinuationsAsynchronously, TaskScheduler.Default);
+
+                _cache[key] = result;
```
stephentoub: Does this cache grow unbounded? Is that ok?

build/Shared/TaskResultCache.cs
```diff
@@ -0,0 +1,126 @@
...
+                {
+                    throw aggregateException.InnerException!;
+                }
+
+                throw task.Exception!;
+            }
+
+#pragma warning disable VSTHRD002 // Avoid problematic synchronous waits does not apply since we are in a continuation and we already checked if the task is complete
+            return task.Result;
+#pragma warning restore VSTHRD002
```
stephentoub:
```C#
return task.GetAwaiter().GetResult();
```
?

stephentoub:
> From my testing, it appears to only be a problem on .NET Framework.

On .NET Framework, all await continuations are registered as synchronous continuations: if there are multiple, and continuations can be run synchronously, they'll end up being run serially.  This was changed for .NET Core 3.0 in https://github.com/dotnet/coreclr/pull/25280: with that, only the first await continuation will be invoked synchronously, with others being forcefully queued so they don't back up behind the first.

build/Shared/TaskResultCache.cs
jeffkl: @davkean said it would reduce lock contention if we always look for an existing value and then lock before writing.   So this code looks for a value, gets a lock, looks again, and then writes.  Is there a better way?  We were also told that the `ConcurrentDictionary<TKey, TValue>` could cause allocations on hot paths so we should use a Dictionary with locks.

build/Shared/TaskResultCache.cs
jeffkl: It grows unbounded during a restore yes.  Each restore gets a new context which starts with a new cache.

build/Shared/TaskResultCache.cs
jeffkl: Are you saying that would be better?  Since we know the task is complete in this case, is it still not good to just access `.Result` ?

build/Shared/TaskResultCache.cs
stephentoub: > It grows unbounded during a restore yes.

And there's no risk it grows too big?

build/Shared/TaskResultCache.cs
stephentoub: I'm suggesting replacing the whole body with that, not just the .Result call.  It seems like you're explicitly checking for cancellation and exceptions in order to propagate a single exception, which task.GetAwaiter().GetResult() will already do.

build/Shared/TaskResultCache.cs
stephentoub: > said it would reduce lock contention if we always look for an existing value and then lock before writing

It will reduce lock contention, but this pattern is not safe with `Dictionary<>`.  You would need to use `ConcurrentDictionary<>` or something similar to avoid locking on the read path. Technically you could also use the non-generic Hashtable, which as an implementation detail is safe for reads concurrently with at most one write.

build/Shared/TaskResultCache.cs
jeffkl: Wow yeah that's way better.  Thanks for the suggestion!

build/Shared/TaskResultCache.cs
jeffkl: I'm assuming there is no risk since this concept of a dictionary containing cached task results has existed in NuGet for quite some time.  In some of the largest restores where NuGet is restoring thousands of packages, it hasn't been an issue.  The idea here is NuGet is restoring hundreds of projects at the same time and each target framework for each project is walking the dependency graph and each thread wants to resolve a particular package that might involved querying the feed, downloading the `.nupkg`, and unzipping it.  We want each caller to be awaiting the same task for any particular package and not duplicate the work.  Are you thinking of another way to accomplish this that wouldn't grow an unbounded cache?  Would love to hear your thoughts in general of this design.

build/Shared/TaskResultCache.cs
jeffkl: I've pushed a commit to use ConcurrentDictionary but still just call `TryGetValue` then manually lock.  Will that give the right balance of reducing locks as the cache is populated but still make the read/write operations thread safe?

davkean: > @davkean said it would reduce lock contention if we always look for an existing value and then lock before writing.

I never said this with `Dictionary<TKey, TValue>`. :)

You don't need to mix ConcurrentDictionary and the lock usage, use .GetOrAdd instead. However, I would still be interested in the contention this introduces, and you need to think about what happens when the factories overlap and one of the values is thrown away.

build/Shared/TaskResultCache.cs
davkean: The concern here is that you will continue to hold onto the memory used by these tasks. Given we see this consume 100 GB VM today in some traces, this is a real concern.

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteWalkContext.cs
```diff
@@ -46,7 +40,9 @@ public RemoteWalkContext(SourceCacheContext cacheContext, PackageSourceMapping p
         /// <summary>
         /// Library entry cache.
         /// </summary>
-        public ConcurrentDictionary<LibraryRangeCacheKey, Task<GraphItem<RemoteResolveResult>>> FindLibraryEntryCache { get; }
+        internal static TaskResultCache<LibraryRangeCacheKey, GraphItem<RemoteResolveResult>> FindLibraryEntryCache { get; } = new();
```
nkolev92: I don't think this is the right thing. This keeps all the tasks forever. 

This change has nothing to do with the continuation.

nkolev92 (changes requested):

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteWalkContext.cs
nkolev92: another notable thing is that these caches are not necessarily safe to be shared across projects as is. 

GraphItem has concepts like CentralTransitive which can only ever be project specific. 

Does restore with this change yield the same results?

davkean: While I like the enthusiam here, I agree with @nkolev92 it doesn't feel like GraphItem is at right level to be caching here across a solution.

jeffkl: Sorry this was a mistake and was just me experimenting with all projects sharing the same task cache.  I've reverted this

build/Shared/TaskResultCache.cs
```diff
@@ -0,0 +1,113 @@
...
+
+            AsyncLazy<TValue> result = refresh
+                ? _cache.AddOrUpdate(key, lazy, (_, _) => lazy)
+                : _cache.GetOrAdd(key, lazy);
+
+            return await result;
+        }
+
+#pragma warning disable VSTHRD011 // Lazy<Task<T>>.Value can deadlock. Use AsyncLazy<T> instead, which we are
+        public class AsyncLazy<T> : Lazy<Task<T>>
```
nkolev92: We have an async lazy in NuGet.Common. 

Should we have the same implementation?

src/NuGet.Core/NuGet.Commands/RestoreCommand/SourceRepositoryDependencyProvider.cs
```diff
@@ -204,11 +201,12 @@ internal static SemaphoreSlim GetThrottleSemaphoreSlim(IEnvironmentVariableReade
             }
 
             cancellationToken.ThrowIfCancellationRequested();
-            Task<LibraryIdentity> result = FindLibraryIdentityAsync(libraryRange, cacheContext);
 
             try
             {
-                return await result;
+                LibraryIdentity result = await _libraryMatchCache.GetOrAddAsync(libraryRange, cacheContext.RefreshMemoryCache, () => FindLibraryCoreAsync(libraryRange, cacheContext, logger, cancellationToken), cancellationToken);
```
nkolev92: Isn't this a change from the previous implementation where we were calling it using Task.Run?

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteWalkContext.cs
```diff
@@ -26,7 +22,8 @@ public RemoteWalkContext(SourceCacheContext cacheContext, PackageSourceMapping p
             RemoteLibraryProviders = new List<IRemoteDependencyProvider>();
             PackageSourceMapping = packageSourceMapping ?? throw new ArgumentNullException(nameof(packageSourceMapping));
 
-            FindLibraryEntryCache = new ConcurrentDictionary<LibraryRangeCacheKey, Task<GraphItem<RemoteResolveResult>>>();
+            FindLibraryEntryCache = new TaskResultCache<LibraryRangeCacheKey, GraphItem<RemoteResolveResult>>();
+            ResolvePackageLibraryMatchCache = new TaskResultCache<LibraryRange, Tuple<LibraryRange, RemoteMatch>>();
```
nkolev92, line 26: This will probably be empty most of the time. 
That's why I originally had the cache outside of the remote walk context. 
It's not doing so much walking, since PackageDownload is kind of an extra hook to restore, so logically I had kept it out of the regular restore flow on purpose. 

You don't have to change this, I'm just explaining the reasoning, as their separation was intentional.

src/NuGet.Core/NuGet.DependencyResolver.Core/ResolverUtility.cs
```diff
@@ -229,13 +227,14 @@ public static class ResolverUtility
...
-        public static Task<Tuple<LibraryRange, RemoteMatch>> FindPackageLibraryMatchCachedAsync(
-            ConcurrentDictionary<LibraryRange, Task<Tuple<LibraryRange, RemoteMatch>>> cache,
+        public static async Task<Tuple<LibraryRange, RemoteMatch>> FindPackageLibraryMatchCachedAsync(
             LibraryRange libraryRange,
-           RemoteWalkContext remoteWalkContext,
+            RemoteWalkContext remoteWalkContext,
             CancellationToken cancellationToken)
         {
-            return cache.GetOrAdd(libraryRange, (cacheKey) => ResolvePackageLibraryMatchAsync(libraryRange, remoteWalkContext, cancellationToken));
+            Tuple<LibraryRange, RemoteMatch> result = await remoteWalkContext.ResolvePackageLibraryMatchCache.GetOrAddAsync(libraryRange, () => ResolvePackageLibraryMatchAsync(libraryRange, remoteWalkContext, cancellationToken), cancellationToken);
```
nkolev92: you just return the task here to avoid the extra state machine.

src/NuGet.Core/NuGet.DependencyResolver.Core/ResolverUtility.cs
```diff
@@ -229,13 +227,14 @@ public static class ResolverUtility
         /// <param name="remoteWalkContext">remote Providers (all sources, including file sources)</param>
         /// <param name="cancellationToken">cancellation token</param>
         /// <returns>The requested range and remote match.</returns>
-        public static Task<Tuple<LibraryRange, RemoteMatch>> FindPackageLibraryMatchCachedAsync(
-            ConcurrentDictionary<LibraryRange, Task<Tuple<LibraryRange, RemoteMatch>>> cache,
+        public static async Task<Tuple<LibraryRange, RemoteMatch>> FindPackageLibraryMatchCachedAsync(
```
nkolev92: If possible, consider reverting the async part.

src/NuGet.Core/NuGet.Protocol/RemoteRepositories/RemoteV2FindPackageByIdResource.cs
```diff
@@ -433,27 +434,18 @@ public RemoteV2FindPackageByIdResource(PackageSource packageSource, HttpSource h
             return packageInfos.FirstOrDefault(p => p.Identity.Version == version);
         }
 
-        private Task<IEnumerable<PackageInfo>> EnsurePackagesAsync(
+        private async Task<IEnumerable<PackageInfo>> EnsurePackagesAsync(
```
nkolev92, line 437: Same as the others, A bunch of these intentionally do not await the task.

src/NuGet.Core/NuGet.DependencyResolver.Core/ResolverUtility.cs
```diff
@@ -18,22 +17,18 @@ namespace NuGet.DependencyResolver
 {
     public static class ResolverUtility
     {
-        public static Task<GraphItem<RemoteResolveResult>> FindLibraryCachedAsync(
-            ConcurrentDictionary<LibraryRangeCacheKey, Task<GraphItem<RemoteResolveResult>>> cache,
+        public static async Task<GraphItem<RemoteResolveResult>> FindLibraryCachedAsync(
```
nkolev92: I'd check all of these, and see where we can avoid adding the extra state machine.

build/Shared/TaskResultCache.cs
jeffkl: At the moment I have `TaskResultCache` in `build\Shared` so its compiled into everything and it can be `internal`.  Not all of our assemblies reference `NuGet.Common` so those won't compile if they reference that `AsyncLazy<T>` of course.  I could put `TaskResultCache` into `NuGet.Common` but then it would have to be public.  Thoughts?

src/NuGet.Core/NuGet.Commands/RestoreCommand/SourceRepositoryDependencyProvider.cs
jeffkl: I'm certainly no expert in TPL for these advance cases but I thought calling `MethodAsync()` and `Task.Run(() => MethodAsync())` is essentially the same thing.  So there's no need to do so since this is calling a method which receives the task factory delegate which actually schedules the task and returns a task that can be awaited and shared by callers to all await.  In all of my testing, things seem to work?

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteWalkContext.cs
jeffkl: I mainly liked this change because I didn't think the caches should be `public` since they're only used as an implementation detail and callers don't need to concern themselves with that.  Then since this Task cache was here, I figured I'd put the other one here as well.  Just my throught process.

src/NuGet.Core/NuGet.DependencyResolver.Core/ResolverUtility.cs
jeffkl: Done, I've updated all the places where it wasn't needed.  I wish we had an analyzer to tell us.

src/NuGet.Core/NuGet.DependencyResolver.Core/ResolverUtility.cs
jeffkl: Done

src/NuGet.Core/NuGet.DependencyResolver.Core/ResolverUtility.cs
jeffkl: Done

src/NuGet.Core/NuGet.Protocol/RemoteRepositories/RemoteV2FindPackageByIdResource.cs
jeffkl: Done

build/Shared/TaskResultCache.cs
nkolev92: Let's leave it as is. 
Let's just have a targetted fix are better, as long as you're aware of the other implementation.

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteWalkContext.cs
nkolev92: Yeah, I'm fine with that. 

When I get around to do: https://github.com/NuGet/Home/issues/13077, it'll make a bunch of our restore changes straightforward.

src/NuGet.Core/NuGet.Commands/RestoreCommand/SourceRepositoryDependencyProvider.cs
davkean: I don't have a comment on the change, however:

> but I thought calling MethodAsync() and Task.Run(() => MethodAsync()) is essentially the same thing

No, that is not the same thing. 

Given the following method:

``` C#
Task MethodAsync() 
{
    File.Delete("Foo.cs");
    
    await FooAsync();
}
```

`await MethodAsync();` will synchronously call File.Delete on the same thread, yielding the calling thread (releasing back to the pool) at `await FooAsync`.

`await Task.Run(() => MethodAsync());` will schedule a new item onto the thread pool where File.Delete will execute when there is a free thread, yielding the calling thread immediately.

jeffkl: Ah so is it safe to say that `Task.Run()` should always be used in case the called method has a mixture of async/sync code?

davkean: No, in fact it can [hurt your throughput](https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/24027/Threading?anchor=%E2%9D%8C-**avoid**-unneeded-thread-pool-queuing-via-%60task.run%60).  Calling Task.Run and immediately awaiting on it will only hurt (you saying "add this task to the back of the queue").

You want to use Task.Run when scheduling multiple mostly synchronous/CPU tasks to enable them to overlap and run at the same time.

nkolev92: I think in this situation, it probably won't help us whether we have an extra Task.Run, since we're in an async codepath and are always going to be on the threadpool already.

nkolev92: @Erarndt The Task.Run was added recently here in https://github.com/NuGet/NuGet.Client/pull/5556.

nkolev92 (dismissed):

src/NuGet.Core/NuGet.Commands/RestoreCommand/SourceRepositoryDependencyProvider.cs
Erarndt: @davkean If we want to avoid blocking the caller in an async path, is it preferrable to do `Task.Run()` or use `Task.Yield()`?

davkean: In this exact case (running on a thread pool thread), they are mostly identical. If you are running on something different than a thread pool, they will be different (Task.Yield on a UI thread for example, would get readded to back of message queue),

jeffkl:
I've pushed a commit that gets rid of `AsyncLazy<T>` with an implementation from @Erarndt that allocates less.

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteWalkContext.cs
Erarndt, line 26: A bit outside the scope of this change, but I wonder if it'd be worth it to switch off of `System.Tuple` and use `ValueTuple` instead.

src/NuGet.Core/NuGet.DependencyResolver.Core/ResolverUtility.cs
```diff
@@ -19,21 +18,18 @@ namespace NuGet.DependencyResolver
...
+            LibraryRangeCacheKey key = new(libraryRange, framework);
 
-            if (cache.TryGetValue(key, out var graphItem))
-                return graphItem;
-
-            graphItem = cache.GetOrAdd(key, FindLibraryEntryAsync(key.LibraryRange, framework, runtimeIdentifier, context, cancellationToken));
-
-            return graphItem;
+            return context.FindLibraryEntryCache.GetOrAddAsync(key,
+                state => FindLibraryEntryAsync(state.LibraryRange, state.framework, state.runtimeIdentifier, state.context, state.cancellationToken),
```
Erarndt: Can this be marked `static`? This helps catch cases where there's an unintended capture.

Erarndt (dismissed):

nkolev92 (dismissed):

jeffkl:
This should be the last change, per Eric's findings, we decided to institute a per-key lock object so that multiple threads can get/add to the cache if they're looking for another key.  This reduces lock contention and makes things faster.

nkolev92 (approved):

PR #5625, opened 2024-02-06
Eliminate System.CommandLine prebuilt package

NikolaMilosavljevic:
Source-build is showing a prebuilt package for `System.CommandLine`, version 2.0.0-beta4.23307.1.

Source-build needs to use a single version of all packages, that are built from source.

The change enables source-build to override version of this package that is used by the repo.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -30,6 +30,10 @@
       <Uri>https://github.com/aspnet/xdt</Uri>
       <Sha>d4d088b6a9c793525b1a27a119cb66ba4587bb39</Sha>
     </Dependency>
+    <Dependency Name="System.CommandLine" Version="2.0.0-beta4.23307.1">
+      <Uri>https://github.com/dotnet/command-line-api</Uri>
+      <Sha>02fe27cd6a9b001c8feb7938e6ef4b3799745759</Sha>
+    </Dependency>
     <Dependency Name="System.ComponentModel.Composition" Version="4.5.0">
       <Uri>https://github.com/dotnet/corefx</Uri>
       <Sha>30ab651fcb4354552bd4891619a0bdd81e0ebdbf</Sha>
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -30,9 +30,9 @@
       <Uri>https://github.com/aspnet/xdt</Uri>
       <Sha>d4d088b6a9c793525b1a27a119cb66ba4587bb39</Sha>
     </Dependency>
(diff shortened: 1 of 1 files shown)


NikolaMilosavljevic:
@ellahathaway @MichaelSimons @mthalman @zivkan

nkolev92 (dismissed):

mthalman (approved):

eng/Version.Details.xml
```diff
@@ -30,6 +30,10 @@
       <Uri>https://github.com/aspnet/xdt</Uri>
       <Sha>d4d088b6a9c793525b1a27a119cb66ba4587bb39</Sha>
     </Dependency>
+    <Dependency Name="System.CommandLine" Version="2.0.0-beta4.23307.1">
```
MichaelSimons: Additionally, wouldn't it be best to actually upgrade NuGet to be on the same version as .NET?

NikolaMilosavljevic: Probably, yes, However, there isn't a single version used by other repos. There are at least 3 different versions used today across 11 repos that consume this package. `arcade` uses the same version as `nuget.client`. The version is 6 months old.

`installer` repo uses the most recent version, from couple weeks ago. I'll update `nuget.client` to use the same.

While other repos can use the dependency flow to get on latest, `nuget.client` repo does not have that support.

NikolaMilosavljevic: Fixed with https://github.com/NuGet/NuGet.Client/pull/5625/commits/a06168918fc250a2088a0ce4f18808bd98f6d1d3

NikolaMilosavljevic:
@nkolev92 it doesn't seem that checks were run for this PR - is there a way to trigger them? Otherwise, this is ready for merging.

eng/Version.Details.xml
```diff
@@ -30,6 +30,10 @@
       <Uri>https://github.com/aspnet/xdt</Uri>
       <Sha>d4d088b6a9c793525b1a27a119cb66ba4587bb39</Sha>
     </Dependency>
+    <Dependency Name="System.CommandLine" Version="2.0.0-beta4.24068.1">
```
nkolev92, line 33: This doesn't match our built against version. 2.0.0-beta4.23307.1

Is that intentional? 
We matched our version to the one in the .NET SDK recently.

NikolaMilosavljevic: That's OK - versions in this file are used for source-build only.

nkolev92 (approved):

ellahathaway (approved):

PR #5628, opened 2024-02-09
Consolidate packages.config restore implementations by minimizing discrepancies

### LINKED ISSUE #13233: Consolidate packages.config restore implementations by minimizing discrepancies
Unfortunately we have 3 different implementations of packages.config restore.
A lot of that is due to the difference in reading the packages and part of it is not wanting to make public APIs for PC restore.

Those 3 implementations are:

nuget.exe restore
msbuild.exe /t:restore
VS restore
There's a bunch of differences in how the PackageRestoreManager gets called. It's made my work for PC restore vulnerability checking challenging, because the calls are not unified.

nkolev92:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13233

## Description

Unfortunately we have 3 different implementations of packages.config restore. 
A lot of that is due to the difference in reading the packages and part of it is not wanting to make public APIs for PC restore.

Those 3 implementations are: 
- nuget.exe restore
- msbuild.exe /t:restore
- VS restore

There's a bunch of differences in how the PackageRestoreManager gets called. It's made my work for PC restore vulnerability checking challenging, because the calls are not unified. 

The change here is primarily 1, always create a PackageRestoreData list with *all* packages rather than just the missing ones. The same amount of work is being done, the difference being that now we have PackageRestoreData list with all the data, which I can later use for vulnerability checking. 

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception - refactoring - existing tests actually helped ensuring I got it right :D 
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Clients/NuGet.CommandLine/Commands/RestoreCommand.cs
```diff
@@ -424,6 +444,20 @@ private async Task<IReadOnlyList<RestoreSummary>> PerformNuGetV2RestoreAsync(Pac
             }
         }
 
+        private static Dictionary<string, string> GetPackagesConfigToProjectPath(PackageRestoreInputs packageRestoreInputs)
+        {
+            Dictionary<string, string> configToProjectPath = [];
```
jeffkl: What is this syntax?  I prefer 
```suggestion
            Dictionary<string, string> configToProjectPath = new();
```

I couldn't find any instance of this syntax in our existing code and suggest we don't start using it.

nkolev92: These are largely the suggestions applied when you change the constructor. 
I think it used to do `new()` but now it does `[]`. 

I can change it.

jeffkl: It confused me, I thought it was somehow set to an empty array?  But it must be short hand for `new()`?  I've never seen it and think `new()` is short hand enough while still being concise.

src/NuGet.Core/NuGet.Build.Tasks/BuildTasksUtility.cs
```diff
@@ -449,9 +447,15 @@ private static async Task<RestoreSummary> PerformNuGetV2RestoreAsync(Common.ILog
...
                 string packagesConfigPath = GetPackagesConfigFilePath(pcRestoreMetadata.ProjectPath);
 
-                firstPackagesConfigPath = firstPackagesConfigPath ?? packagesConfigPath;
-
-                installedPackageReferences.AddRange(GetInstalledPackageReferences(packagesConfigPath, allowDuplicatePackageIds: true));
+                foreach (PackageReference packageReference in GetInstalledPackageReferences(packagesConfigPath))
+                {
+                    if (!packageReferenceToProjects.TryGetValue(packageReference, out List<string> value))
+                    {
+                        value ??= [];
```
jeffkl:
```suggestion
                        value ??= new();
```

src/NuGet.Clients/NuGet.CommandLine/Commands/RestoreCommand.cs
```diff
@@ -264,12 +264,46 @@ private async Task<IReadOnlyList<RestoreSummary>> PerformNuGetV2RestoreAsync(Pac
...
+                    foreach (PackageReference packageReference in GetInstalledPackageReferences(configFile))
+                    {
+                        if (configToProjectPath.TryGetValue(configFile, out string projectPath))
+                        {
+                            projectPath = configFile;
+                        }
+
+                        if (!packageReferenceToProjects.TryGetValue(packageReference, out List<string> value))
+                        {
+                            value ??= [];
```
jeffkl:
```suggestion
                            value ??= new();
```

jeffkl (dismissed):

src/NuGet.Core/NuGet.Build.Tasks/BuildTasksUtility.cs
nkolev92: These are actually arrays/list, so I thought here it made sense.
Do you want these changed as well?

jeffkl: I guess I don't understand this syntax then.  Does this say "set my list to an empty array"?  Or does this say "set my list to a new list"?  If it is the former, I would think that's bad since later on we call `.Add()` and the backing storage would need to be resized.  If it is the later, I prefer just `new()`.

nkolev92: It's a new list/new array. 

I'll merge the suggestions.

jeffkl (approved):
Thanks!

PR #5631, opened 2024-02-13
All calls to SourceRepository.GetResource must pass a cancellation token

### LINKED ISSUE #11813: [Bug]: Canceling msbuild restore is slow when invalid/unreachable source configured
### NuGet Product Used

MSBuild.exe

### Product Version

msbuild 17.2

### Worked before?

_No response_

### Impact

### Repro Steps & Context

1. Find a solution that does not restore quickly. The longer it takes, the better. For example, try https://github.com/OrchardCMS/OrchardCore
2. Run `dotnet nuget add source https://invalid.nuget.org/v1234/index.json -n invalidSource`
3. Run `msbuild -t:restore -v:m`
4. Wait for "error NU1301: Unable to load the service index for source ..." messages start appearing
5. Press ctrl+C
6. See that "Attempting to cancel the build" appears immediately, but it can take seconds, or minutes, for the command prompt to return (restore to end), during which time a bunch more NU1301 errors will be reported.

By comparison, if you press ctrl-c quickly after the "Determining projects to restore" message is displayed, the command line returns immeidately. Similarly, when using `-p:RestoreUseStaticGraphEvaluation=true`, ctrl-c also exits immediately.

### Verbose Logs

_No response_

### LINKED ISSUE #13234: SourceRepository.GetResourceAsync should be cancellable.
### NuGet Product(s) Affected

NuGet SDK

### Current Behavior

Calling `SourceRepository.GetResourceAsync<FindPackageByIdResource>(cancellationToken);` without an internet connection throws FatalProtocolException after a timeout of 5 seconds, even if cancellationToken.IsCancellationRequested is true.

### Desired Behavior

If cancellation is requested before the timeout, the task should cancel early and throw TaskCancelledException instead of FatalProtocolException.

### Additional Context

Using NuGet.Protocol v6.8.0

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13234
Fixes: https://github.com/NuGet/Home/issues/11813

Regression? No

## Description

`CancellationToken`s weren't being passed around everywhere needed. In the case of the two linked issues, it was 

I don't know what NuGet.Protocol looked like before the V3 protocol was added, but since it was added, there are resource providers that need the service index, so it no longer makes sense to call `GetResourceAsync` or `GetResource` without passing a cancellation token. Therefore I marked these two overloads (without the cancellation token parameter) as obsolete. **This is a breaking change** from an API point of view. What's worse is that these are commonly used APIs by anyone using the NuGet.Protocol package (part of the NuGet Client SDK), therefore I'm expecting high impact. However, the fix is trivial to implement.

All the other changes in PR are just fixing compile errors (since we treat warnings as errors). I passed in cancellation tokens where possible, but there are some methods that don't have a cancellation token already, but mostly this was in NuGet.CommandLine or NuGet.CommandLine.XPlat, where we don't handle ctrl-c directly anyway.

Honestly, it would be great to ban `CancellationToken.None` in our non-test code, but it will require a bunch of breaking API changes where existing public APIs need to start taking in a cancellation token, which is more effort than I'm willing to make for this quick fix.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A: No UX changes, just "backend" changes. 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.PowerShellCmdlets/NuGetPowerShellBaseCommand.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.PowerShellCmdlets/NuGetPowerShellBaseCommand.cs
@@ -384,7 +384,6 @@ namespace NuGet.PackageManagement.PowerShellCmdlets
         {
             var packageSource = new PackageSource(source);
             var repository = _sourceRepositoryProvider.CreateRepository(packageSource);
-            var resource = repository.GetResource<PackageSearchResource>();
 
             return repository;
         }
--- a/src/NuGet.Core/NuGet.Commands/CommandRunners/DeleteRunner.cs
+++ b/src/NuGet.Core/NuGet.Commands/CommandRunners/DeleteRunner.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Globalization;
+using System.Threading;
 using System.Threading.Tasks;
 using NuGet.Common;
 using NuGet.Configuration;
@@ -34,7 +35,7 @@ namespace NuGet.Commands
             {
                 logger.LogWarning(string.Format(CultureInfo.CurrentCulture, Strings.Warning_HttpServerUsage, "delete", packageSource.Source));
             }
-            var packageUpdateResource = await CommandRunnerUtility.GetPackageUpdateResource(sourceProvider, packageSource);
+            var packageUpdateResource = await CommandRunnerUtility.GetPackageUpdateResource(sourceProvider, packageSource, CancellationToken.None);
 
             await packageUpdateResource.Delete(
                 packageId,
--- a/src/NuGet.Core/NuGet.Commands/CommandRunners/PushRunner.cs
+++ b/src/NuGet.Core/NuGet.Commands/CommandRunners/PushRunner.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Threading;
 using System.Threading.Tasks;
 using NuGet.Common;
 using NuGet.Configuration;
@@ -39,7 +40,7 @@ namespace NuGet.Commands
                 timeoutSeconds = 5 * 60;
             }
             PackageSource packageSource = CommandRunnerUtility.GetOrCreatePackageSource(sourceProvider, source);
-            var packageUpdateResource = await CommandRunnerUtility.GetPackageUpdateResource(sourceProvider, packageSource);
+            var packageUpdateResource = await CommandRunnerUtility.GetPackageUpdateResource(sourceProvider, packageSource, CancellationToken.None);
 
             // Only warn for V3 style sources because they have a service index which is different from the final push url.
             if (packageSource.IsHttp && !packageSource.IsHttps && !packageSource.AllowInsecureConnections &&
@@ -58,7 +59,7 @@ namespace NuGet.Commands
                 && !sourceUri.IsFile
                 && sourceUri.IsAbsoluteUri)
             {
-                symbolPackageUpdateResource = await CommandRunnerUtility.GetSymbolPackageUpdateResource(sourceProvider, source);
+                symbolPackageUpdateResource = await CommandRunnerUtility.GetSymbolPackageUpdateResource(sourceProvider, source, CancellationToken.None);
                 if (symbolPackageUpdateResource != null)
                 {
                     symbolSource = symbolPackageUpdateResource.SourceUri.AbsoluteUri;
--- a/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpSource.cs
+++ b/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpSource.cs
@@ -435,7 +435,7 @@ namespace NuGet.Protocol
                 throw new ArgumentNullException(nameof(throttle));
             }
 
-            Func<Task<HttpHandlerResource>> factory = () => source.GetResourceAsync<HttpHandlerResource>();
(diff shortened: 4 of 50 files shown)


src/NuGet.Core/NuGet.Protocol/SourceRepository.cs
```diff
@@ -135,6 +136,7 @@ public virtual async Task<FeedType> GetFeedType(CancellationToken token)
         /// </summary>
         /// <typeparam name="T">Expected resource type</typeparam>
         /// <returns>Null if the resource does not exist</returns>
+        [Obsolete("Use the overload that takes a CancellationToken. If you don't want to support cancelation, use CancellationToken.None.")]
```
zivkan: As mentioned in the PR description, this is effectively a breaking change in our public APIs. And it's a commonly used API, so this will be "highly" impactful. However, the fix is trivial, as the message explains.

I think it's worthwhile to make this breaking change, as it's a long term win, minimizing risk of bugs similar to the two linked to this PR. Please reply if you don't think we should mark these APIs as obsolete, so as not to impact customers using the APIs in their own apps.

src/NuGet.Core/NuGet.Commands/CommandRunners/DeleteRunner.cs
```diff
@@ -34,7 +35,7 @@ public static class DeleteRunner
             {
                 logger.LogWarning(string.Format(CultureInfo.CurrentCulture, Strings.Warning_HttpServerUsage, "delete", packageSource.Source));
             }
-            var packageUpdateResource = await CommandRunnerUtility.GetPackageUpdateResource(sourceProvider, packageSource);
+            var packageUpdateResource = await CommandRunnerUtility.GetPackageUpdateResource(sourceProvider, packageSource, CancellationToken.None);
```
zivkan, line 38: Currently `delete` is only available in CLI tools, and we don't handle ctrl-c in any of them, so technically the lack of cancellation passthough won't affect NuGet tooling, but if customers are using NuGet.Commands in their own apps, there's no way for them to cancel this request.

src/NuGet.Core/NuGet.Commands/CommandRunners/PushRunner.cs
```diff
@@ -58,7 +59,7 @@ public static class PushRunner
                 && !sourceUri.IsFile
                 && sourceUri.IsAbsoluteUri)
             {
-                symbolPackageUpdateResource = await CommandRunnerUtility.GetSymbolPackageUpdateResource(sourceProvider, source);
+                symbolPackageUpdateResource = await CommandRunnerUtility.GetSymbolPackageUpdateResource(sourceProvider, source, CancellationToken.None);
```
zivkan, line 62: Similar to delete, currently `push` is only available though NuGet's CLI tools, where we don't handle ctrl-c ourselves, so the .NET runtime will kill the process immediately. However, customers calling this method in their own apps won't be able to cancel the push.

src/NuGet.Core/NuGet.Protocol/HttpSource/HttpSource.cs
```diff
@@ -435,7 +435,7 @@ public static HttpSource Create(SourceRepository source, IThrottle throttle)
                 throw new ArgumentNullException(nameof(throttle));
             }
 
-            Func<Task<HttpHandlerResource>> factory = () => source.GetResourceAsync<HttpHandlerResource>();
+            Func<Task<HttpHandlerResource>> factory = () => source.GetResourceAsync<HttpHandlerResource>(CancellationToken.None);
```
zivkan, line 438: While this is in a Client SDK package, anything that wants to make an HTTP request has to get the HttpHandlerResource, hence we can be confident that the HttpHandlerResourceProvider isn't going to be making HTTP requests (or do any other async work), so using `CancellationToken.None` here won't cause any blocking on cancellation requested.

jeffkl (dismissed):
I'm on the fence about the breaking change.  I guess we'll find out...

donnie-msft (dismissed):
Should 11813 be labelled as Performance?

src/NuGet.Clients/NuGet.PackageManagement.PowerShellCmdlets/NuGetPowerShellBaseCommand.cs
```diff
@@ -384,7 +384,6 @@ private SourceRepository CreateRepositoryFromSource(string source)
         {
             var packageSource = new PackageSource(source);
             var repository = _sourceRepositoryProvider.CreateRepository(packageSource);
-            var resource = repository.GetResource<PackageSearchResource>();
```
nkolev92, line 387: do you why was the search resource fetched here? I'm guessing just a bug?

nkolev92 (dismissed):
We can probably do this change without obsoleting the methods. 
For example, we can ban the usage of those methods using https://github.com/NuGet/NuGet.Client/blob/dev/build/BannedSymbols.txt. 

Just an option.

src/NuGet.Clients/NuGet.PackageManagement.PowerShellCmdlets/NuGetPowerShellBaseCommand.cs
zivkan: It certainly looks like a bug. `resource` was never used, and `GetResource` isn't supposed to have side-effects, so I don't see any reason why the line should exist. Hence I'm deleting it.  We have a fair few PMC tests, so I'm hopeful that if deleting this line does somehow cause problems, the tests should catch it.

jeffkl (approved):

nkolev92 (approved):

PR #5632, opened 2024-02-14
Yaml uses 'include' templates instead of 'extends' templates

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2689

## Description

Moving our templates to be `include` type templates means that they no longer inherit the `pipeline.yml` but instead copy the contents of that yaml.

`resources` had to be copied to the parent pipeline. Unfortunately, it's not supported within templates.
https://developercommunity.visualstudio.com/t/unexpected-value-resources-in-yaml-template/728151

I've tested changes to the template, `pipeline.yml`, and the modified pipelines by using the Validate option in Azure DevOps, and running the private dev CI. 

### Tests and validation:
- build/vsts_build.yaml - **Deleted. No longer needed per Andy**
- eng/pipelines/ci.yml
  - image
- eng/pipelines/dailytests.yml
  - image
- eng/pipelines/official.yml
  - image
- eng/pipelines/pull_request.yml
  - image

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


donnie-msft:
@zivkan could you please take a look at my progress & testing question in the description before I open this Draft PR? Thanks

(note if you try to select my branch in AzDo, make sure it's `dev-donnie-msft-yamlExtends2`. I rebased this PR's branch and now AzDo isn't getting new commits to the branch)

build/vsts_build.yaml
zivkan, line 1: This file is no longer used and can be deleted. It used to be our "entry point" yaml, before I split into separate official and pull_request files, and moved to `/eng/pipelines`.

I kept it around for a while, in case my splitting into multiple files caused problems, we'd be able to quickly revert back. But then I forgot about this file, so never deleted it myself.

eng/pipelines/ci.yml
```diff
@@ -70,21 +70,25 @@ parameters:
   displayName: Run Mono tests on Mac
   type: boolean
   default: true
+- name: isOfficialBuild
+  type: boolean
+  default: false
```
zivkan: There's no reason for someone manually queuing a build to be allowed to set `isOfficialBuild` true. In fact, trying to do so will fail, because the pipeline then try to request real signing, rather than testing signing, and failing because the pipeline definition isn't approved for real signing.

Therefore, remove this parameter, and when "invoking" the `pipeline.yml` template, use `isOfficialBuild: false`

eng/pipelines/official.yml
```diff
@@ -64,21 +64,21 @@ parameters:
...
-    RunBuildForPublishing: ${{parameters.RunBuildForPublishing}}
-    RunCrossFrameworkTestsOnWindows: ${{parameters.RunCrossFrameworkTestsOnWindows}}
-    RunEndToEndTests: ${{parameters.RunEndToEndTests}}
-    RunFunctionalTestsOnWindows: ${{parameters.RunFunctionalTestsOnWindows}}
-    RunSourceBuild: ${{parameters.RunSourceBuild}}
-    RunTestsOnLinux: ${{parameters.RunTestsOnLinux}}
-    RunTestsOnMac: ${{parameters.RunTestsOnMac}}
-    RunMonoTestsOnMac: ${{parameters.RunMonoTestsOnMac}}
+stages:
+  - template: templates/pipeline.yml
```
zivkan: I'm looking at the PR with "show whitespace" disabled, and so these are the only two lines that show as changed for this file.

I believe every file yaml that currently has `extends:` should have these two changes made, and none of the other changes in this PR are needed. In fact, in `dailytests.yml` you didn't add a `-` before the template. If that works, then it's not needed here either, and it might be just a simple find/replace `extends` to `stages` in the yaml files. However, if the `-` is needed, then it's needed in `dailytests.yml` as well. You can verify by queueing a new build of the daily test pipeline (you don't need to actually queue the pipeline, just check that it "compiles" and shows you the parameters and variables to change before actually running it).

But we can also check with the DartLab team, to confirm that this is sufficient to fix whatever issue caused them to ask us to switch to templates instead of extends.

eng/pipelines/pull_request.yml
```diff
@@ -70,21 +70,52 @@ parameters:
...
+resources:
+  pipelines:
+  - pipeline: DartLab
+    source: DartLab
+    branch: main
+  repositories:
+  - repository: DartLabTemplates
+    type: git
+    name: DartLab.Templates
+    ref: refs/heads/main
```
zivkan, line 83: This yaml file doesn't reference any of these resources, so I don't believe they need to be defined here. They should be defined in the same yaml where they're used.

eng/pipelines/pull_request.yml
```diff
@@ -70,21 +70,52 @@ parameters:
   displayName: Run Mono tests on Mac
   type: boolean
   default: true
+- name: isOfficialBuild
+  type: boolean
+  default: false
```
zivkan: Like a comment on another file, we don't need the ability to set `isOfficialBuild` to true when manually queueing the build. Therefore we don't need this parameter on this yaml.

eng/pipelines/templates/pipeline.yml
```diff
@@ -1,107 +1,3 @@
...
-resources:
-  pipelines:
-  - pipeline: DartLab
-    source: DartLab
-    branch: main
-  repositories:
-  - repository: DartLabTemplates
-    type: git
-    name: DartLab.Templates
-    ref: refs/heads/main
```
zivkan, line 87: I believe this is why the pipeline is failing to "compile" (your question from the PR description). This yaml file references the DartLab pipeline and the DartLabTemplates repository, but it's no longer defined in this file.

eng/pipelines/templates/pipeline.yml
```diff
@@ -1,107 +1,3 @@
...
-
-variables:
```
zivkan, line 89: I don't know if variables need to be set in the "entry point" or "top level" yaml. If it does, then these variables need to be copied into `official.yaml` as well.

However, try reverting all the changes to this file, and removing the resources and variables from `pull_request.yaml`.

I believe that this file shouldn't need to be modified at all for this PR.

eng/pipelines/official.yml
donnie-msft: The `-` is needed. Here's an example from the docs: https://learn.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops&pivots=templates-includes#stage-reuse

eng/pipelines/pull_request.yml
donnie-msft: That's a great thought, but unfortunately, it's not supported.
https://developercommunity.visualstudio.com/t/unexpected-value-resources-in-yaml-template/728151

eng/pipelines/official.yml
donnie-msft: image

eng/pipelines/templates/pipeline.yml
donnie-msft: Answered in https://github.com/NuGet/NuGet.Client/pull/5632/files#r1498420606

eng/pipelines/ci.yml
```diff
@@ -71,9 +71,21 @@ parameters:
...
+  repositories:
+  - repository: DartLabTemplates
+    type: git
+    name: DartLab.Templates
+    ref: refs/heads/main
+
+stages:
+- template: templates/pipeline.yml
   parameters:
+    isOfficialBuild: false
```
kartheekp-ms, line 88: Is it intentional to set the official build as false here?

eng/pipelines/pull_request.yml
```diff
@@ -70,9 +70,39 @@ parameters:
...
+resources:
+  pipelines:
+  - pipeline: DartLab
+    source: DartLab
+    branch: main
+  repositories:
+  - repository: DartLabTemplates
+    type: git
+    name: DartLab.Templates
+    ref: refs/heads/main
```
kartheekp-ms, line 83: This YAML code is duplicated in three files in this PR. Can we consolidate this code into a single location and reference it as needed across all three places? The reason behind this request is that sometimes we may need to change the DartLab branch name to test if a fix on their side unblocks us. Generally, we avoid creating a pull request in our repo solely for changing the DartLab branch name. Instead, we queue a PR build to verify if the fix on their side resolves our issue, and then we revert our changes. DartLab team pushes their changes to the main branch eventually, which we rely on.

eng/pipelines/ci.yml
donnie-msft: Yes, see https://github.com/NuGet/NuGet.Client/pull/5632#discussion_r1490558324

eng/pipelines/pull_request.yml
donnie-msft: That's a great thought, but unfortunately resources are not supported in YAML templates:
https://developercommunity.visualstudio.com/t/unexpected-value-resources-in-yaml-template/728151

If you know another way to accomplish this, do let me know.

donnie-msft: Here's an internal example of another repo which also has explicit resources defined. It does use a template for parameters, but not resources.
https://dev.azure.com/devdiv/DevDiv/_git/VSUnitTesting?path=/.vsts-ci-validation.yml&version=GBmain&line=16&lineEnd=17&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

eng/pipelines/pull_request.yml
zivkan: I unresolved this conversation, as I still see `isOfficialBuild` as a parameter that can changed on manual queue. Although, I don't see it actually passed through to the `pipeline.yml` template.

zivkan (commented):
LGTM, except pull_request.yml defines `isOfficialBuild` parameter (which also isn't used).

zivkan (approved):

PR #5635, opened 2024-02-17
Add nullable annotations to NuGet.Configuration

### LINKED ISSUE #13250: Add nullable annotations to NuGet.Configuration
[NuGet.Configuration](https://www.nuget.org/packages/NuGet.Configuration) is a package that is part of the [NuGet Client SDK](https://learn.microsoft.com/en-us/nuget/reference/nuget-client-sdk). We are adding null annotations, so projects using [nullable reference types](https://learn.microsoft.com/dotnet/csharp/nullable-references) will benefit from compiler null checking, leading to improved software reliability (fewer NullReferenceExceptions).

Generally when annotating nulls we try to avoid making any breaking API or ABI changes. However, the following changes have been made in order to allow the compiler to provide everyone with the most useful nullable warnings.

Projects that use this package and enable nullable checks might find that upgrading to a version of the package that now has nullable annotations will generate warnings that previously did not generate warnings. This is a natural consequence of annotating the assembly where it previously lacked annotations.

In addition, the following breaking changes were implemented:
 
## Changes to `SettingsBase` and `SettingsElement`

To make the compiler's null checking more accurate, some properties were changed, and some abstract class constructors changed. This should only affect classes that extend the abstract classes. NuGet.Configuration does not provide extensibility to allow custom configuration sections via `Settings.Load*`. Therefore, while these are technically breaking API changes, it's not expected to affect common usage of NuGet.Configuration.

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13250

Regression? No

## Description

Enabled `nullable` in NuGet.Configuration and its test project.

Dealing with `SettingBase.Origin` and `SettingElement.ElementName` was particularly difficult (without suppressing the warnings), and I introduced some breaking API changes in these abstract classes as a result. However, this only affect classes that extend the configuration, but we don't provide extensibility other assemblies to add custom configuration classes. Therefore, it's unlikely that anyone should be affected by those API changes (witnessed by the fact that none of NuGet's other projects needed changes due to those API changes).

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A: no functionality changes, only adding null annotations. 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceCredential.cs
+++ b/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceCredential.cs
@@ -44,7 +44,7 @@ namespace NuGet.Configuration
         /// Comma-delimited list of authentication types the credential is valid for as stored in the config file.
         /// If null or empty, all authentication types are valid. Example: 'basic,negotiate'
         /// </summary>
-        public string ValidAuthenticationTypesText { get; }
+        public string? ValidAuthenticationTypesText { get; }
 
         /// <summary>
         /// Retrieves password in clear text. Decrypts on-demand.
@@ -67,7 +67,7 @@ namespace NuGet.Configuration
                 }
                 else
                 {
-                    return PasswordText;
+                    return PasswordText!;
                 }
             }
         }
@@ -94,7 +94,7 @@ namespace NuGet.Configuration
         /// Comma-delimited list of authentication types the credential is valid for as stored in the config file.
         /// If null or empty, all authentication types are valid. Example: 'basic,negotiate'
         /// </param>
-        public PackageSourceCredential(string source, string username, string passwordText, bool isPasswordClearText, string validAuthenticationTypesText)
+        public PackageSourceCredential(string source, string username, string passwordText, bool isPasswordClearText, string? validAuthenticationTypesText)
         {
             Source = source ?? throw new ArgumentNullException(nameof(source));
             Username = username ?? throw new ArgumentNullException(nameof(username));
@@ -134,7 +134,7 @@ namespace NuGet.Configuration
             string username,
             string password,
             bool storePasswordInClearText,
-            string validAuthenticationTypesText)
+            string? validAuthenticationTypesText)
         {
             if (source == null)
             {
@@ -194,11 +194,11 @@ namespace NuGet.Configuration
         /// <returns>
         /// Enumeration of valid authentication types. If empty, all authentication types are valid.
         /// </returns>
-        private static IEnumerable<string> ParseAuthTypeFilterString(string str)
+        private static IEnumerable<string> ParseAuthTypeFilterString(string? str)
         {
             return string.IsNullOrWhiteSpace(str)
                 ? Enumerable.Empty<string>()
-                : str.Split(',').Select(x => x.Trim()).Where(x => x != string.Empty);
+                : str!.Split(',').Select(x => x.Trim()).Where(x => x != string.Empty);
         }
 
 
@@ -207,7 +207,7 @@ namespace NuGet.Configuration
             return new CredentialsItem(Source, Username, PasswordText, IsPasswordClearText, ValidAuthenticationTypesText);
         }
 
-        public bool Equals(PackageSourceCredential other)
+        public bool Equals(PackageSourceCredential? other)
         {
             if (other == null)
             {
@@ -225,7 +225,7 @@ namespace NuGet.Configuration
                 IsPasswordClearText == other.IsPasswordClearText;
         }
 
-        public override bool Equals(object other)
+        public override bool Equals(object? other)
         {
             return Equals(other as PackageSourceCredential);
         }
--- a/src/NuGet.Core/NuGet.Configuration/Proxy/ProxyCache.cs
+++ b/src/NuGet.Core/NuGet.Configuration/Proxy/ProxyCache.cs
@@ -46,7 +46,7 @@ namespace NuGet.Configuration
             _environment = environment;
         }
 
-        public IWebProxy GetProxy(Uri sourceUri)
+        public IWebProxy? GetProxy(Uri sourceUri)
         {
             // Check if the user has configured proxy details in settings or in the environment.
             var configuredProxy = GetUserConfiguredProxy();
@@ -77,14 +77,14 @@ namespace NuGet.Configuration
             return _cachedCredentials.TryAdd(configuredProxy.ProxyAddress, proxyCredentials);
         }
 
-        public WebProxy GetUserConfiguredProxy()
+        public WebProxy? GetUserConfiguredProxy()
         {
             // Try reading from the settings. The values are stored as 3 config values http_proxy, http_proxy.user, http_proxy.password
             var host = SettingsUtility.GetConfigValue(_settings, ConfigurationConstants.HostKey);
             if (!string.IsNullOrEmpty(host))
             {
                 // The host is the minimal value we need to assume a user configured proxy.
-                var webProxy = new WebProxy(host);
+                var webProxy = new WebProxy(host!);
 
                 if (RuntimeEnvironmentHelper.IsWindows)
                 {
@@ -101,7 +101,7 @@ namespace NuGet.Configuration
                 if (!string.IsNullOrEmpty(noProxy))
                 {
                     // split comma-separated list of domains
-                    webProxy.BypassList = noProxy.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
+                    webProxy.BypassList = noProxy!.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
                 }
 
                 return webProxy;
@@ -109,7 +109,7 @@ namespace NuGet.Configuration
 
             // Next try reading from the environment variable http_proxy. This would be specified as http://<username>:<password>@proxy.com
             host = _environment.GetEnvironmentVariable(ConfigurationConstants.HostKey);
-            Uri uri;
+            Uri? uri;
             if (!string.IsNullOrEmpty(host)
                 && Uri.TryCreate(host, UriKind.Absolute, out uri))
             {
@@ -129,7 +129,7 @@ namespace NuGet.Configuration
                 if (!string.IsNullOrEmpty(noProxy))
                 {
                     // split comma-separated list of domains
-                    webProxy.BypassList = noProxy.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
+                    webProxy.BypassList = noProxy!.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
                 }
 
                 return webProxy;
@@ -150,9 +150,9 @@ namespace NuGet.Configuration
                 updateValueFactory: (_, __) => { Version = Guid.NewGuid(); return credentials; });
         }
 
-        public NetworkCredential GetCredential(Uri proxyAddress, string authType)
+        public NetworkCredential? GetCredential(Uri proxyAddress, string authType)
         {
-            ICredentials cachedCredentials;
+            ICredentials? cachedCredentials;
             if (_cachedCredentials.TryGetValue(proxyAddress, out cachedCredentials))
             {
                 return cachedCredentials.GetCredential(proxyAddress, authType);
@@ -162,10 +162,10 @@ namespace NuGet.Configuration
         }
 
         [Obsolete("Retained for backcompat only. Use UpdateCredential instead")]
-        public void Add(IWebProxy proxy)
+        public void Add(IWebProxy? proxy)
         {
             var webProxy = proxy as WebProxy;
-            if (webProxy != null)
+            if (webProxy != null && webProxy.Credentials is not null)
             {
                 _cachedCredentials.TryAdd(webProxy.ProxyAddress, webProxy.Credentials);
             }
--- a/src/NuGet.Core/NuGet.Configuration/Proxy/ProxyCache.cs
+++ b/src/NuGet.Core/NuGet.Configuration/Proxy/ProxyCache.cs
@@ -165,7 +165,7 @@ namespace NuGet.Configuration
         public void Add(IWebProxy? proxy)
         {
             var webProxy = proxy as WebProxy;
-            if (webProxy != null && webProxy.Credentials is not null)
+            if (webProxy?.Credentials is not null)
             {
                 _cachedCredentials.TryAdd(webProxy.ProxyAddress, webProxy.Credentials);
             }
--- a/src/NuGet.Core/NuGet.Configuration/Resources.resx
+++ b/src/NuGet.Core/NuGet.Configuration/Resources.resx
@@ -309,4 +309,7 @@
 1 - config file path
 </comment>
   </data>
+  <data name="ShowError_UnknownOrigin" xml:space="preserve">
+    <value>File location not known</value>
+  </data>
 </root>
--- a/src/NuGet.Core/NuGet.Configuration/Settings/Items/AuthorItem.cs
+++ b/src/NuGet.Core/NuGet.Configuration/Settings/Items/AuthorItem.cs
@@ -27,7 +27,7 @@ namespace NuGet.Configuration
 
         public override SettingBase Clone()
         {
-            var newItem = new AuthorItem(Name, Certificates.Select(c => c.Clone() as CertificateItem).ToArray());
+            var newItem = new AuthorItem(Name, Certificates.Select(c => (CertificateItem)c.Clone()).ToArray());
 
             if (Origin != null)
             {
@@ -59,7 +59,7 @@ namespace NuGet.Configuration
             return element;
         }
 
-        public override bool Equals(object other)
+        public override bool Equals(object? other)
         {
             if (other is AuthorItem author)
             {
--- a/src/NuGet.Core/NuGet.Configuration/Settings/Items/CredentialsItem.cs
+++ b/src/NuGet.Core/NuGet.Configuration/Settings/Items/CredentialsItem.cs
@@ -146,7 +146,7 @@ namespace NuGet.Configuration
             var elementDescendants = element.Elements();
             var countOfDescendants = elementDescendants.Count();
 
-            var parsedItems = elementDescendants.Select(e => (AddItem)SettingFactory.Parse(e, origin));
+            var parsedItems = elementDescendants.Select(e => SettingFactory.Parse(e, origin)).OfType<AddItem>();
 
             foreach (var item in parsedItems)
             {
--- a/src/NuGet.Core/NuGet.Configuration/Settings/Items/CredentialsItem.cs
+++ b/src/NuGet.Core/NuGet.Configuration/Settings/Items/CredentialsItem.cs
@@ -146,7 +146,7 @@ namespace NuGet.Configuration
             var elementDescendants = element.Elements();
             var countOfDescendants = elementDescendants.Count();
 
-            var parsedItems = elementDescendants.Select(e => SettingFactory.Parse(e, origin) as AddItem).Where(i => i != null).Cast<AddItem>();
+            var parsedItems = elementDescendants.Select(e => (AddItem)SettingFactory.Parse(e, origin));
 
             foreach (var item in parsedItems)
             {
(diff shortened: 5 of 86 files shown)


src/NuGet.Core/NuGet.Configuration/Proxy/ProxyCache.cs
```diff
@@ -162,10 +162,10 @@ public NetworkCredential GetCredential(Uri proxyAddress, string authType)
...
 
         [Obsolete("Retained for backcompat only. Use UpdateCredential instead")]
-        public void Add(IWebProxy proxy)
+        public void Add(IWebProxy? proxy)
         {
             var webProxy = proxy as WebProxy;
-            if (webProxy != null)
+            if (webProxy != null && webProxy.Credentials is not null)
             {
                 _cachedCredentials.TryAdd(webProxy.ProxyAddress, webProxy.Credentials);
```
zivkan: Maybe this could technically be a breaking change?  Adding a null credential into the cache doesn't seem useful, hence my code change. But it also means the dictionary key will no longer be added.

src/NuGet.Core/NuGet.Configuration/Settings/SettingBase.cs
```diff
@@ -8,11 +8,11 @@ namespace NuGet.Configuration
...
     public abstract class SettingBase
     {
-        internal XNode Node { get; private set; }
+        internal XNode? Node { get; private set; }
 
-        internal ISettingsGroup Parent { get; set; }
+        internal ISettingsGroup? Parent { get; set; }
 
-        internal SettingsFile Origin { get; private set; }
+        internal SettingsFile? Origin { get; private set; }
```
zivkan, line 15: A lot of code assumes this property is never null, and will throw NullReferenceException if it is. Yet, many classes set this to null in `Update` methods.

In the end it was too hard to figure out an ideal fix, so I annotated the property as it's used (nullable), and suppressed the warning in a few places where NullReferenceException will be thrown (unless there was an obvious/easy way to handle the null).  This PR is trying to add null annotations, not redesign the class design.

src/NuGet.Core/NuGet.Configuration/Settings/SettingFactory.cs
```diff
@@ -176,7 +176,7 @@ public int GetHashCode(SettingElement obj)
             if (xElement.Name.LocalName.Equals(ConfigurationConstants.PackageSourceMapping, StringComparison.OrdinalIgnoreCase) && duplicatedDescendants != null)
             {
                 var duplicatedKey = string.Join(", ", duplicatedDescendants.Select(d => d.Attributes["key"]));
-                var source = duplicatedDescendants.Select(d => d.Origin.ConfigFilePath).First();
+                var source = duplicatedDescendants.Select(d => d.Origin?.ConfigFilePath).Where(d => d is not null).First();
```
zivkan: In theory `d.Origin` will never be null, because that means that the instance wasn't loaded from a file. But we don't have APIs to load settings from an in-memory stream, and other parts of the code throw NullReferenceException if settings classes are created programatically, rather than from loading from a file.

But technically this could be a change in behaviour, whereas the goal of this PR was no behaviour changes, only null annotations.

src/NuGet.Core/NuGet.Configuration/Settings/SettingsGroup.cs
```diff
@@ -15,15 +16,17 @@ public abstract class SettingsGroup<T> : SettingElement, ISettingsGroup where T
...
+        public override string ElementName { get; }
+
+        protected SettingsGroup(string name)
+            : this(name, attributes: null, children: null)
         {
-            Children = new List<T>();
         }
 
-        protected SettingsGroup(IReadOnlyDictionary<string, string> attributes, IEnumerable<T> children)
+        protected SettingsGroup(string name, IReadOnlyDictionary<string, string>? attributes, IEnumerable<T>? children)
```
zivkan, line 26: Technically a breaking API change. But it only affects classes that extend this class, which should only happen in this assembly.

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceCredential.cs
```diff
@@ -67,7 +67,7 @@ public string Password
                 }
                 else
                 {
-                    return PasswordText;
+                    return PasswordText!;
```
jeffkl, line 70: Should the Password property be marked as nullable instead?

src/NuGet.Core/NuGet.Configuration/Proxy/ProxyCache.cs
```diff
@@ -162,10 +162,10 @@ public NetworkCredential GetCredential(Uri proxyAddress, string authType)
         }
 
         [Obsolete("Retained for backcompat only. Use UpdateCredential instead")]
-        public void Add(IWebProxy proxy)
+        public void Add(IWebProxy? proxy)
         {
             var webProxy = proxy as WebProxy;
-            if (webProxy != null)
+            if (webProxy != null && webProxy.Credentials is not null)
```
jeffkl:
```suggestion
            if (webProxy?.Credentials is not null)
```

src/NuGet.Core/NuGet.Configuration/Resources.resx
```diff
@@ -309,4 +309,7 @@
 1 - config file path
 </comment>
   </data>
+  <data name="ShowError_UnknownOrigin" xml:space="preserve">
+    <value>File location not known</value>
```
jeffkl, line 313: Should this end in a period?
```suggestion
    <value>File location not known.</value>
```

src/NuGet.Core/NuGet.Configuration/Settings/Items/CredentialsItem.cs
```diff
@@ -149,12 +140,13 @@ public CredentialsItem(string name, string username, string password, bool isPas
...
         {
-            ElementName = element.Name.LocalName;
+            // ElementName is read from XML file, so it must be decoded.
+            ElementName = XmlConvert.DecodeName(element.Name.LocalName);
 
             var elementDescendants = element.Elements();
             var countOfDescendants = elementDescendants.Count();
 
-            var parsedItems = elementDescendants.Select(e => SettingFactory.Parse(e, origin) as AddItem).Where(i => i != null);
+            var parsedItems = elementDescendants.Select(e => SettingFactory.Parse(e, origin) as AddItem).Where(i => i != null).Cast<AddItem>();
```
jeffkl: While you're here we might as well change this to use `OfType<T>()` instead

```suggestion
            var parsedItems = elementDescendants.Select(e => SettingFactory.Parse(e, origin)).OfType<AddItem>();
```

src/NuGet.Core/NuGet.Configuration/Settings/Items/FileClientCertItem.cs
```diff
@@ -66,16 +73,16 @@ internal FileClientCertItem(XElement element, SettingsFile origin)
                                                                     Origin?.ConfigFilePath ?? "<Config file path>"));
             }
 
-            AddAttribute(ConfigurationConstants.PathAttribute, path);
+            AddAttribute(ConfigurationConstants.PathAttribute, path!);
 
             if (!string.IsNullOrWhiteSpace(password))
             {
-                AddAttribute(ConfigurationConstants.PasswordAttribute, password);
+                AddAttribute(ConfigurationConstants.PasswordAttribute, password!);
```
jeffkl, line 80: I really wish the compiler was smarter here, is there a way for us to tell the compiler that `string.IsNullOrWhiteSpace()` already checked for `null`?

src/NuGet.Core/NuGet.Configuration/Settings/Items/RepositoryItem.cs
```diff
@@ -201,7 +206,13 @@ internal override void Update(SettingItem other)
                 }
                 else
                 {
-                    _owners.Update(repository._owners);
+#pragma warning disable CS8604 // Possible null reference argument.
```
jeffkl, line 209: 🤣

src/NuGet.Core/NuGet.Configuration/Settings/SettingFactory.cs
jeffkl: Not a hot path so it doesn't really matter but this might as well just call `First()`

```suggestion
                var source = duplicatedDescendants.Select(d => d.Origin?.ConfigFilePath).First(d => d is not null);
```

src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs
```diff
@@ -436,7 +439,8 @@ private static ISettings LoadSettingsGivenSettingsFiles(List<SettingsFile> setti
                             }
                             return settingsLoadingContext.GetOrCreateSettingsFile(f);
                         })
-                        .Where(f => f != null));
+                        .Where(f => f != null)
+                        .Cast<SettingsFile>());
```
jeffkl:
```suggestion
                        .OfType<SettingsFile>();
```

src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs
```diff
@@ -723,15 +736,18 @@ private static string ResolveRelativePath(string originDirectoryPath, string ori
         {
             if (string.IsNullOrEmpty(originDirectoryPath) || string.IsNullOrEmpty(originFilePath))
             {
+#pragma warning disable CS8603 // Possible null reference return.
+                // This code path doesn't seem possible, but if I'm wrong, then deleting this block will change the behavior of the code.
```
jeffkl, line 738: Would it be better for this method to return `string?`?

src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs
```diff
@@ -839,10 +855,11 @@ private static string GetSettingsFileNameFromDir(string directory)
 
         private static IEnumerable<string> GetSettingsFilePaths(string root)
```
jeffkl, line 854: Why not have this method accept nullable?

```suggestion
        private static IEnumerable<string> GetSettingsFilePaths(string? root)
```

src/NuGet.Core/NuGet.Configuration/Settings/SettingsFile.cs
```diff
@@ -201,13 +207,14 @@ private XDocument CreateDefaultConfig()
             return new XDocument(configurationElement.AsXNode());
         }
 
-        private void ExecuteSynchronized(Action ioOperation)
+        private T ExecuteSynchronized<T>(Func<T> ioOperation)
```
jeffkl, line 210: I think declaring this as nullable will fix the warning on line 245 below

```suggestion
        private T? ExecuteSynchronized<T>(Func<T?> ioOperation)
```

jeffkl (commented):
Looks great overall, thanks for the effort involved in this.

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceCredential.cs
zivkan: I think this is one of the places I was getting frustrated with the design/implementation, so I just suppressed the issue.

The statis `FromUserInput` throws if `password` is null. So, if an instance is created that way, `Password` and `PasswordText` can't be null.

However, the class constructor doesn't throw if password is null. So, maybe a null is allowed?  But after I finished annotating everything, there weren't any calls to the constructor where the compiler detected that a null value might be used.

So, in the end I just suppressed it, because that's how the implementation appears to work, even if the way it's written confuses the compiler. If you do a null check in an if statement, even if the symbol is not marked as nullable, the compiler will treat it as nullable. So, the `if (PasswordText != null && otherCondition)` tricks the compiler into thinking that `PasswordText` might be null when the `if` expression is false, even though `PasswordText` is not marked as nullable.

In theory it should be safe to remove the `PasswordText != null` part of the expression, but my goal is primarily to add nullable annotations, and I want to minimize risk of adding bugs.

src/NuGet.Core/NuGet.Configuration/Resources.resx
zivkan: No, I don't think it. It gets used inside another message: https://github.com/NuGet/NuGet.Client/blob/5502baa16a294bb1290b6fd2a135694ed2884a59/src/NuGet.Core/NuGet.Configuration/Resources.resx#L307

In theory this should never be shown to customers, but because of how `SettingsBase` is implemented, `.Origin` has to be marked as nullable because of an obscure design choice, but in practise when saving changes to nuget.config files, it will always be not null.

src/NuGet.Core/NuGet.Configuration/Settings/Items/CredentialsItem.cs
zivkan: This comment made me realize my PR has a bug.

The old code could theoretically end up with nulls in the enumeration, and then when it's used, throw a NullReferenceException.

Your suggestion, and my original PR, would silently ignore the bad values, therefore changing product behaviour.

What I'm changing locally, and will push soon, is changing the `as AddItem` to a `(AddItem)` cast, so that an exception will be thrown (so similar to the original code), but it'll be an `InvalidCastException` which will tell us the actual type of the object that couldn't be cast.

src/NuGet.Core/NuGet.Configuration/Settings/Items/FileClientCertItem.cs
zivkan: At some point I thought I saw a NuGet package that had .NET Framework reference assemblies with nullable annotations backported from .NET (Core). In theory, if we used that then the compiler would know. But 1. I can't find it and 2. I feel like it's somewhat risky. We'd need to carefully inspect the package contents to make sure it's really only reference assemblies and not adding a shim that ends up enabling an interception attack.

So yeah, as long as we target .NET Framework or .NET Standard, it sucks that we have to do a whole lot of nullable suppressions. But I don't know of a better solution.

src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs
zivkan: Then the caller would have to handle nulls. IntelliCode shows there's only 1 reference. If we make that one return a nullable string too, then there's 6 references who now need to handle nulls.

But as the comment says, this doesn't seem possible. This method is only called in 1 place:
https://github.com/NuGet/NuGet.Client/blob/5502baa16a294bb1290b6fd2a135694ed2884a59/src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs#L712-L717

So, as we can see, it already verifies that both parameters are not null or empty before calling the method, so it's not clear why this method checks it again.

As mentioned in another comment, my goal was to annotate nulls, so I was trying to avoid product changes to reduce risk of regressions.

src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs
zivkan: because to me it's non-sensical to call a method that finds all parent directories of a null path. Yes, it would allow us to have one less line, but we don't win anything for code golf, and I feel like correctness is nicer than fewer lines.  But that's my personal preference 😁

src/NuGet.Core/NuGet.Configuration/Settings/SettingsFile.cs
zivkan: As mentioned in another comment, my preference is for method signatures to be "correct", rather than convenient. By putting the nullable annotation in the method signates, now callers will have to handle nulls. But the implementation of the method is such that it can never return a null. It either returns a non-null value, or it throws.

In order to fix this "properly", we need to make an overload of `ConcurrencyUtilities.ExecuteWithFileLocked` that takes a `Func<T>` and return the `T`, this this `ExecuteSynchronzied` will be able to use the return value, rather than a nullable variable that gets set inside a delegate function. But that would require new public APIs that I didn't feel like creating.

zivkan:
> Looks great overall, thanks for the effort involved in this.

@jeffkl  thank you for actually reviewing it and providing feedback!   It took me many hours, of largely "mechanical" work, so it was very easy for my concentration to slip and to do the wrong thing. It's great that someone double checked my work.

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceCredential.cs
jeffkl: Okay

src/NuGet.Core/NuGet.Configuration/Settings/Items/CredentialsItem.cs
```diff
@@ -149,12 +140,13 @@ public CredentialsItem(string name, string username, string password, bool isPas
...
         {
-            ElementName = element.Name.LocalName;
+            // ElementName is read from XML file, so it must be decoded.
+            ElementName = XmlConvert.DecodeName(element.Name.LocalName);
 
             var elementDescendants = element.Elements();
             var countOfDescendants = elementDescendants.Count();
 
-            var parsedItems = elementDescendants.Select(e => SettingFactory.Parse(e, origin) as AddItem).Where(i => i != null);
+            var parsedItems = elementDescendants.Select(e => (AddItem)SettingFactory.Parse(e, origin));
```
jeffkl: Before the logic was using a safe cast (`as AddItem`) which would return `null` if the item was not of that type.  It would not throw an exception right?  Then it would check for `null` and finally cast the item.  `.OfType<T>()` does all of that in a single call and is used in other places in this assembly.

Here's a sample of the three implementations: https://dotnetfiddle.net/w8kV4G

```suggestion
            var parsedItems = elementDescendants.Select(e => SettingFactory.Parse(e, origin)).OfType<AddItem>();
```

src/NuGet.Core/NuGet.Configuration/Settings/Items/AuthorItem.cs
```diff
@@ -27,7 +27,7 @@ internal AuthorItem(XElement element, SettingsFile origin)
 
         public override SettingBase Clone()
         {
-            var newItem = new AuthorItem(Name, Certificates.Select(c => c.Clone() as CertificateItem).ToArray());
+            var newItem = new AuthorItem(Name, Certificates.Select(c => (CertificateItem)c.Clone()).ToArray());
```
jeffkl, line 30: I doubt it matters in this code path but before the array of certificates would be a clone and a safe cast to `CertificateItem` so the array _could_ have `null`s?  Now it will throw?  Again, probably doesn't matter, just wanted to point it out in case it does.

src/NuGet.Core/NuGet.Configuration/Settings/Items/FileClientCertItem.cs
jeffkl: Okay thanks for looking into it!

src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs
jeffkl: True, in a perfect world this should throw an `ArgumentNullException` but that's probably not possible to change behavior.

jeffkl (commented):
Reviewing this pull request nearly crashes my browser, let me know if I missed a follow-up to any comment or if you want to discuss offline

src/NuGet.Core/NuGet.Configuration/Settings/Items/CredentialsItem.cs
zivkan: You're right. Fixed in the next commit.

src/NuGet.Core/NuGet.Configuration/Settings/Items/AuthorItem.cs
zivkan: It's not possible, because `Certificates` has type `IList<CertificateItem>`. The implementation of `CertificateItem.Clone()` returns a `new CertificateItem(...)`, but the `Clone` method is defined in `SettingsBase`, so `Clone()`'s return type is `SettingBase`.  So, IMO it's an example of incorrect usage of the `as` keyword.

Personally, I think every class should use the pattern

```c#
public SettingsBase SettingsBase.Clone() => Clone();

public new CertificateItem Clone() { /* implementation */ }
```

This way callers won't need to cast, and people reading the code would be less confused.  But I didn't want to spend a lot of time changing the assembly, just trying to annotate "as quickly as possible".

jeffkl (approved):
🤞

PR #5638, opened 2024-02-21
[`dotnet package search`] Don't show help output for source errors.

### LINKED ISSUE #13161: dotnet package search format shows help output in addition to a json file
### NuGet Product Used

dotnet.exe

### Product Version

8.0.2xx

### Summary

There is an issue where the help output gets mixed with the JSON output under specific circumstances when using `dotnet package search`.

Steps to Reproduce:

- Use a command line interface to run a search command using the SDK that shells out to NuGet.
- Ensure that the command and parameters are correctly formatted.
- Include a --source parameter with an invalid source. The invalid source could be one that is not specified in a configuration file or is not recognized as a valid source.

Example command: `dotnet package search --source my-source`
Expected Behavior:

The SDK should handle invalid sources gracefully, providing a clear error message without interfering with the expected JSON output format.

Current Problem:

When the provided source is invalid (e.g., my-source does not exist in any config file or is not a valid source by itself), the SDK outputs both the help message and the JSON output together.

                {
                  "version": 1,
                  "problems": [
                    {
                      "text": "The specified source "my-source" is invalid. Provide a valid source.",
                      "problemType": "Error"
                    }
                  ],
                  "searchResult": []
                }
                Description:
                  Searches one or more package sources for packages that match a search term. If no sources are specified, all sources defined in the NuGet.Config are used.
                
                Usage:
                  dotnet package search [<SearchTerm>] [options]
                
                Arguments:
                  <SearchTerm>  Search term to filter package names, descriptions, and tags. Used as a literal value. Example: `dotnet package search some.package`. See also `--exact-match`.
                
                Options:
                  --source <Source>          The package source to search. You can pass multiple `--source` options to search multiple package sources. Example: `--source https://api.nuget.org/v3/index.json`.
                  --take <Take>              Number of results to return. Default 20.
                  --skip <Skip>              Number of results to skip, to allow pagination. Default 0.
                  --exact-match              Require that the search term exactly match the name of the package. Causes `--take` and `--skip` options to be ignored.
                  --interactive              Stop and wait for user input or action (for example to complete authentication).
                  --prerelease               Include prerelease packages.
                  --configfile <ConfigFile>  The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used. For more information, see https://docs.microsoft.com/nuget/consume-packages/configuring-nuget-behavior
                  --format <Format>          Format the output accordingly. Either `table`, or `json`. The default value is `table`.
                  --verbosity <Verbosity>    Display this amount of details in the output: `normal`, `minimal`, `detailed`. The default is `normal`
                  -?, -h, --help             Show command line help.

Nigusu-Allehu:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13161

## Description

Example command: `dotnet package search --source my-source`
Expected Behavior:

The SDK should handle invalid sources gracefully, providing a clear error message without interfering with the expected JSON output format.

Current Problem:

When the provided source is invalid (e.g., my-source does not exist in any config file or is not a valid source by itself), the SDK outputs both the help message and the JSON output together.

                {
                  "version": 1,
                  "problems": [
                    {
                      "text": "The specified source "my-source" is invalid. Provide a valid source.",
                      "problemType": "Error"
                    }
                  ],
                  "searchResult": []
                }
                Description:
                  Searches one or more package sources for packages that match a search term. If no sources are specified, all sources defined in the NuGet.Config are used.
                
                Usage:
                  dotnet package search [<SearchTerm>] [options]
                
                Arguments:
                  <SearchTerm>  Search term to filter package names, descriptions, and tags. Used as a literal value. Example: `dotnet package search some.package`. See also `--exact-match`.
                
                Options:
                  --source <Source>          The package source to search. You can pass multiple `--source` options to search multiple package sour....

This SDK and NuGet side of the `dotnet package search` command communicate via exit codes.
- if the exit code is 0, help output is not shown.
- if the exit code is 1, help output is shown.
This PR makes sure the exit code 1 only for errors caused by problems from parsing user command input. 

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchCommand.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchCommand.cs
@@ -122,7 +122,7 @@ namespace NuGet.CommandLine.XPlat
                 catch (ArgumentException ex)
                 {
                     logger.LogError(ex.Message);
-                    return 1;
+                    return ExitCodes.EXIT_COMMANDLINE_ARGUMENT_PRSING_FAILURE;
                 }
             });
 
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchCommand.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchCommand.cs
@@ -122,7 +122,7 @@ namespace NuGet.CommandLine.XPlat
                 catch (ArgumentException ex)
                 {
                     logger.LogError(ex.Message);
-                    return ExitCodes.EXIT_COMMANDLINE_ARGUMENT_PARSING_FAILURE;
+                    return ExitCodes.InvalidArguments;
                 }
             });
 
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchCommand.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchCommand.cs
@@ -122,7 +122,7 @@ namespace NuGet.CommandLine.XPlat
                 catch (ArgumentException ex)
                 {
                     logger.LogError(ex.Message);
-                    return ExitCodes.EXIT_COMMANDLINE_ARGUMENT_PRSING_FAILURE;
+                    return ExitCodes.EXIT_COMMANDLINE_ARGUMENT_PARSING_FAILURE;
                 }
             });
 
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchRunner.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchRunner.cs
@@ -50,7 +50,7 @@ namespace NuGet.CommandLine.XPlat
                 packageSearchResultRenderer.RenderProblem(new PackageSearchProblem(PackageSearchProblemType.Error, ex.Message));
                 packageSearchResultRenderer.Finish();
 
-                return 1;
+                return 0;
             }
 
             WarnForHTTPSources(listEndpoints, packageSearchArgs.Logger);
@@ -61,7 +61,7 @@ namespace NuGet.CommandLine.XPlat
                 packageSearchResultRenderer.RenderProblem(new PackageSearchProblem(PackageSearchProblemType.Error, Strings.Error_NoSource));
                 packageSearchResultRenderer.Finish();
 
-                return 1;
+                return 0;
             }
 
             Func<PackageSource, Task<IEnumerable<IPackageSearchMetadata>>> searchPackageSourceAsync =
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchRunner.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchRunner.cs
@@ -50,7 +50,7 @@ namespace NuGet.CommandLine.XPlat
                 packageSearchResultRenderer.RenderProblem(new PackageSearchProblem(PackageSearchProblemType.Error, ex.Message));
                 packageSearchResultRenderer.Finish();
 
-                return 0;
+                return -1;
             }
 
             WarnForHTTPSources(listEndpoints, packageSearchArgs.Logger);
@@ -61,7 +61,7 @@ namespace NuGet.CommandLine.XPlat
                 packageSearchResultRenderer.RenderProblem(new PackageSearchProblem(PackageSearchProblemType.Error, Strings.Error_NoSource));
                 packageSearchResultRenderer.Finish();
 
-                return 0;
+                return -1;
             }
 
             Func<PackageSource, Task<IEnumerable<IPackageSearchMetadata>>> searchPackageSourceAsync =
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchRunner.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchRunner.cs
@@ -50,7 +50,7 @@ namespace NuGet.CommandLine.XPlat
                 packageSearchResultRenderer.RenderProblem(new PackageSearchProblem(PackageSearchProblemType.Error, ex.Message));
                 packageSearchResultRenderer.Finish();
 
-                return ExitCodes.EXIT_COMMAND_EXECUTION_FAILURE;
+                return ExitCodes.Error;
             }
 
             WarnForHTTPSources(listEndpoints, packageSearchArgs.Logger);
@@ -61,7 +61,7 @@ namespace NuGet.CommandLine.XPlat
                 packageSearchResultRenderer.RenderProblem(new PackageSearchProblem(PackageSearchProblemType.Error, Strings.Error_NoSource));
                 packageSearchResultRenderer.Finish();
 
-                return ExitCodes.EXIT_COMMAND_EXECUTION_FAILURE;
+                return ExitCodes.Error;
             }
 
             Func<PackageSource, Task<IEnumerable<IPackageSearchMetadata>>> searchPackageSourceAsync =
@@ -126,7 +126,7 @@ namespace NuGet.CommandLine.XPlat
             }
 
             packageSearchResultRenderer.Finish();
-            return ExitCodes.EXIT_SUCCESS;
+            return ExitCodes.Success;
         }
 
         /// <summary>
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchRunner.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchRunner.cs
@@ -50,7 +50,7 @@ namespace NuGet.CommandLine.XPlat
                 packageSearchResultRenderer.RenderProblem(new PackageSearchProblem(PackageSearchProblemType.Error, ex.Message));
                 packageSearchResultRenderer.Finish();
 
-                return -1;
+                return ExitCodes.EXIT_COMMAND_EXECUTION_FAILURE;
             }
 
             WarnForHTTPSources(listEndpoints, packageSearchArgs.Logger);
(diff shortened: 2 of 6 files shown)


src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchRunner.cs
```diff
@@ -50,7 +50,7 @@ internal static class PackageSearchRunner
                 packageSearchResultRenderer.RenderProblem(new PackageSearchProblem(PackageSearchProblemType.Error, ex.Message));
                 packageSearchResultRenderer.Finish();
 
-                return 1;
+                return 0;
```
zivkan: process exit code 0 means that the command ran successfully. Therefore, this change will break people who use `dotnet pacakge search` in scripts and need to detect errors.

I think this needs to be fixed at a general level, not specific to one command: https://github.com/NuGet/Home/issues/13251

Nigusu-Allehu: That is correct. We have three scenarios and here is my suggestion: 
1. code parsed and run successfully : set exit code to 0
2. code did not parse successfully: set exit code to 1
3. code parsed successfully, but did not run successfully : set exit code to -1

Now, whenever exit code is 1, we printout help messages in addition to the error otherwise we printout the error as it is.
what do you think?

zivkan: My gut reaction is that magic numbers are bad. There are many Stack Overflow, blog posts, and other sources discussing why magic numbers are bad.

When I debug locally and test invalid sub-command names, or invalid arguments, the command line parsing library we use throws a `CommandParsingException`. Rather than having catch-all exception handing [1](https://github.com/NuGet/NuGet.Client/blob/b49d8914861cf18287302ab311fc9d2566e06fe1/src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs#L100)  [2](https://github.com/NuGet/NuGet.Client/blob/b49d8914861cf18287302ab311fc9d2566e06fe1/src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs#L143), could we catch this exception type specifically and show help only then?

That can allow us to fix not only package search, but also all other commands, as requested in this issue: https://github.com/NuGet/Home/issues/13251

Quick fixes are fine when "doing the right thing" is very time consuming, but so far I'm unconvinced that handling this in a very explicit way (not using magic numbers) is going to be very time consuming.

Something else to consider is this:
https://github.com/NuGet/NuGet.Client/blob/b49d8914861cf18287302ab311fc9d2566e06fe1/src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs#L200-L204

jeffkl: I would declare a `const` with a comment indicating why this code path can't return `1` and return that.

Nigusu-Allehu: I added a class for the exit codes. In the future, we can use it for the other commands in XPlat as well. What do you think @zivkan ?

src/NuGet.Core/NuGet.CommandLine.XPlat/ExitCodes.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace NuGet.CommandLine.XPlat
+{
+    internal static class ExitCodes
+    {
+        // Exit code use when the command line arguments are parsed successfully and the command is run successfully
+        internal const int EXIT_SUCCESS = 0;
+
+        // Exit code used when te command line arguments are not parsed successfully.
+        internal const int EXIT_COMMANDLINE_ARGUMENT_PRSING_FAILURE = 1;
```
donnie-msft:
```suggestion
        internal const int EXIT_COMMANDLINE_ARGUMENT_PARSING_FAILURE = 1;
```

src/NuGet.Core/NuGet.CommandLine.XPlat/ExitCodes.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.CommandLine.XPlat
+{
+    internal static class ExitCodes
```
donnie-msft, line 6: I'm not sure if we really need this given the finite, well-known set of values. I couldn't find anywhere in System.Commandline docs where they use constants (https://learn.microsoft.com/en-us/dotnet/standard/commandline/model-binding#set-exit-codes)

Nigusu-Allehu: It is because there is three exit codes(0, 1, 2) we are using here. They all serve different purposes and allow communication between Nuget.CommandLineXPlat and dotnet sdk. It would be using magic numbers (https://github.com/NuGet/NuGet.Client/pull/5638#discussion_r1502202212) if we start returning 0, 1, 2 without any explanation.

src/NuGet.Core/NuGet.CommandLine.XPlat/ExitCodes.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        // Exit code use when the command line arguments are parsed successfully and the command is run successfully
+        internal const int EXIT_SUCCESS = 0;
+
+        // Exit code used when te command line arguments are not parsed successfully.
+        internal const int EXIT_COMMANDLINE_ARGUMENT_PARSING_FAILURE = 1;
```
zivkan: There are no tests in this PR that validate that the bug reported is actually fixed (and therefore doesn't regress again in the future).

I don't think it will be difficult to write a `Dotnet.Integration.Test` that captures the process output and validates that the result is valid JSON.  I strongly recommend that when it comes to bug fixes, try to practise [test driven development](https://en.wikipedia.org/wiki/Test-driven_development) when possible, and write a test that fails first (to ensure the test actually works), then fix the product code to make the test pass. Some things are too hard to test, and TDD is a pain for new features, but for "little" bugs like this, I think it's quite important.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchRunner.cs
zivkan: Honestly, I think it's disappointing that the problem we know exists for multiple commands isn't being addressed. But I'm just one person, you only need one approval for someone else in the team to merge.  If you did investigate fixing the problem in the general sense, but found it was going to be too much effort, please briefly explain what was difficult and going to take a lot of time, and then I'll be happy to approve.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchCommand.cs
zivkan, line 1: `dotnet list package --format json` also outputs json.  Does it have the same problem?

src/NuGet.Core/NuGet.CommandLine.XPlat/ExitCodes.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.CommandLine.XPlat
+{
+    internal static class ExitCodes
+    {
+        // Exit code use when the command line arguments are parsed successfully and the command is run successfully
+        internal const int EXIT_SUCCESS = 0;
```
jeffkl: nit: There is no reason for these to be all caps, I would just name them:
* `ExitCodes.Success`
* `ExitCodes.Error`
* `ExitCodes.InvalidArguments`

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/PackageSearchCommand.cs
Nigusu-Allehu: Not exactly the same problem, but
1. It prints out a help output when there is a problem with a source. Which also does not make sense.
2. The help output it prints out is a little confusing and diffrent from the help output printed out when doing `dotnet list package -h`
Here is an output from `dotnet list package` if there is a problem with a source
```
error: The specified source 'abab' is invalid. Provide a valid source.


Usage: NuGet.CommandLine.XPlat.dll package list [arguments] [options]

Arguments:
  <PROJECT | SOLUTION>  A path to a project, solution file or directory.

Options:
  -h|--help               Show help information
  --force-english-output  Forces the application to run using an invariant, English-based culture.
  --framework             Specifies the target framework for which the packages will be listed.
  --deprecated            Displays only the packages marked deprecated by the authors.
  --outdated              Displays only the packages that need updates with the latest version from the sources.
  --vulnerable            Displays only the packages flagged as vulnerable.
  --include-transitive    Includes transitive packages too in the result.
  --include-prerelease    Considers prerelease versions when looking for latest. Works only with `--outdated`.
  --highest-patch         Considers only the versions with matching minor and major. Works only with `--outdated`.
  --highest-minor         Considers only the versions with matching major. Works only with `--outdated`.
  --source                Sources to lookup for latest versions. Works only with `--outdated`.
  --config                A path to a config file to specify sources. Works only with `--outdated`.
  --format                Set the report output format.
  --output-version        The version of report output.
  --interactive           Allow the command to block and require manual action for operations like authentication.
  -v|--verbosity          Set the verbosity level of the command. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
```
This help output makes it look like the user should be running the command `NuGet.CommandLine.XPlat.dll package list [arguments] [options]`. A dotnet user would definitely be confused by this.

In general, almost all nuget commands in dontet have this problem when it comes to error handling.

Nigusu-Allehu: This PR is just trying to address https://github.com/NuGet/Home/issues/13161, and make sure we have less tech debt by removing the problem from the new command,: `dotnet package search`. However, I agree, we should work on addressing the issue for all the nuget commands in dotnet.

src/NuGet.Core/NuGet.CommandLine.XPlat/ExitCodes.cs
donnie-msft: I'm not sure they're "magic numbers", as they are standard exit codes and they don't change like a MAX_HTTP_RETRY or timeout or something would. Regardless of what they're called, if it helps write cleaner tests or something, that's fine.

jeffkl (approved):

PR #5639, opened 2024-02-21
Enable vulnerability checking for packages.config projects during commandline restore

### LINKED ISSUE #13253: Warn when vulnerabilities are detected during packages.config restore in CLI scenarios.
Child issue of https://github.com/NuGet/Home/issues/12307

nkolev92:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13253
Fixes: https://github.com/NuGet/Client.Engineering/issues/2698

## Description
See https://github.com/NuGet/Home/pull/13154 for the main design.

- The Visual Studio side requires a lot of changes, and as such to make the review easier, I've split up the enablement of audit checking for VS scenarios. Despite that, there are changes/ideas here that are driven by that. Note that even the public APIs here aren't fully committed, but simply a good enough version as of right now.
- [PackageRestoreManager ](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.PackageManagement/IDE/PackageRestoreManager.cs)is where the main work for packages.config restore happens.
- It's a short lived service based on https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.PackageManagement/IDE/IPackageRestoreManager.cs. 
Counterintuitively the service part is used more frequently in non-VS scenarios, which is why I have this length explanation.
- There are 4 methods in this class, 2 for solution based, 2 for package driven restoring, I have deprecated the 2 unused methods as they just add complexity to adding any changes like vulnerability checking, since I now need to check many inputs.
- Given that in the design we decided to use properties to enable/disable audit, I have reused the RestoreAuditProperties class.
- Extract 2 methods for parsing the restore audit properties, in particular the 2 that we need for packages.config restore.
- Create a public API for vulnerability running. 
- The check returns a type that can be "stored" and used for no-op in Visual Studio scenarios. In particular, in a future version, if the restore in VS has no-op, we'll just "replay" the warnings raised by the previous restore if any.
- The type knows how to make additions to a telemetry event. Note that this isn't final, as I don't have the complete Visual Studio implementation yet.
- CLI scenarios, such as restore, install from nuget.exe and `msbuild /t:restore` will run audit by default, even if packages are downloaded.
  
## PR Checklist

- [ ] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [x] Documentation PR or issue filled - https://github.com/NuGet/docs.microsoft.com-nuget/issues/3247
  - **OR**
  - [ ] N/A


src/NuGet.Clients/NuGet.CommandLine/Commands/InstallCommand.cs
```diff
@@ -190,6 +190,8 @@ private async Task PerformV2RestoreAsync(string packagesConfigFilePath, string i
                 packageRestoreFailedEvent: (sender, args) => { failedEvents.Enqueue(args); },
                 sourceRepositories: packageSources.Select(sourceRepositoryProvider.CreateRepository),
                 maxNumberOfParallelTasks: DisableParallelProcessing ? 1 : PackageManagementConstants.DefaultMaxDegreeOfParallelism,
+                disableNuGetAudit: false,
```
zivkan: generally it's better to have booleans use a "positive" name, to avoid double negatives like this

src/NuGet.Core/NuGet.PackageManagement/Audit/AuditCheckResult.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        internal int Severity0VulnerabilitiesFound { get; set; }
+        internal int Severity1VulnerabilitiesFound { get; set; }
+        internal int Severity2VulnerabilitiesFound { get; set; }
+        internal int Severity3VulnerabilitiesFound { get; set; }
+        internal int InvalidSeverityVulnerabilitiesFound { get; set; }
+        internal List<PackageIdentity>? Packages { get; set; }
+        internal double? DownloadDurationInSeconds { get; set; }
+        internal double? CheckPackagesDurationInSeconds { get; set; }
+        internal int? SourcesWithVulnerabilities { get; set; }
```
zivkan, line 26: Considering this type is a `record`, and therefore a `struct`, I think it's important to make all these properties `get; init;`, rather than `get; set;`. 

The problem with `get; set;` in a struct, is if one is passed in as a method parameter (but not as a `ref`, changing the value of properties will cause the values to be lost when the method returns. So, it's significant risk of bugs if people use it wrong.

Making it only possible to set at initialization time reduces the risk that it gets set wrong.

src/NuGet.Core/NuGet.PackageManagement/Audit/AuditCheckResult.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.Collections.Generic;
+using NuGet.Common;
+using NuGet.Packaging.Core;
+
+namespace NuGet.PackageManagement
+{
+    public record AuditCheckResult
+    {
+        public IReadOnlyList<ILogMessage> Warnings { get; }
+        internal bool IsAuditEnabled { get; set; } = true;
```
zivkan, line 16: From a public API design point of view, personally I don't like that this is a public class, and therefore part of the public API, but with properties that other assemblies can't set. It makes it impossible to use in unit tests for a start, except for assemblies that this assembly lists as `InternalsVisibleTo`.

If the intention is that this type should only be created from within the assembly, never by others (even for unit tests), then I suggest adding an internal constructor, so it doesn't have a public constructor.

But if this type needs to be public to be able to get the warnings and set the telemetry, then I think the properties should be public too.  As I mentioned in another comment, the properties can change `set` to `init`, to prevent tampering, making it a "read only" (immutable, depending on your definition of immutable) type.

src/NuGet.Core/NuGet.PackageManagement/Audit/AuditChecker.cs
```diff
@@ -168,18 +215,55 @@ internal static void CreateWarningsForPackagesWithVulnerabilities(Dictionary<Pac
...
+                    packagesWithReportedAdvisories.Add(package);
                 }
             }
+            return warnings;
         }
 
         internal static Dictionary<PackageIdentity, PackageAuditInfo>? FindPackagesWithKnownVulnerabilities(
             IReadOnlyList<IReadOnlyDictionary<string, IReadOnlyList<PackageVulnerabilityInfo>>> knownVulnerabilities,
-            IEnumerable<PackageRestoreData> packages, PackageVulnerabilitySeverity minSeverity)
+            IEnumerable<PackageRestoreData> packages)
```
zivkan, line 268: since `packages` is an `IEnumerable<T>`, change the `foreach` look it's used in to use `NoAllocEnumerator()`, or whatever the method is called.  Alternatively, change the type to something that has a `struct` enumerator, or something like `IList<T>`, so a `for` loop can be used to avoid a heap allocated enumerator.

src/NuGet.Core/NuGet.PackageManagement/Audit/AuditChecker.cs
```diff
@@ -156,10 +193,20 @@ static bool IsAnyVulnerabilityDataFound(IReadOnlyList<IReadOnlyDictionary<string
...
+            ref int Sev1Matches,
+            ref int Sev2Matches,
+            ref int Sev3Matches,
+            ref int InvalidSevMatches,
+            ref List<PackageIdentity> packagesWithReportedAdvisories)
         {
+            var warnings = new List<LogMessage>();
             foreach ((PackageIdentity package, PackageAuditInfo auditInfo) in packagesWithKnownVulnerabilities.OrderBy(p => p.Key.Id))
             {
+                bool isVulnerabilityReported = false; // TODO NK - Am I trying to count packages?
```
zivkan: well, for PackageReference a package that is used in multiple projects will be re-counted in each project, so if we aggregate all the projects in a solution, it will be counted multiple times. The telemetry that lists package IDs with known vulnerabilities gives us the opportunity to de-duplicate and find the unique package list.

src/NuGet.Core/NuGet.PackageManagement/Audit/AuditChecker.cs
```diff
@@ -168,18 +215,55 @@ internal static void CreateWarningsForPackagesWithVulnerabilities(Dictionary<Pac
...
                         severityLabel,
                         vulnerability.Url);
-                    var restoreLogMessage =
-                        RestoreLogMessage.CreateWarning(logCode,
-                        message,
-                        package.Id);
-                    logger.Log(restoreLogMessage);
+
+                    bool counted = false;
+                    foreach (var projectPath in auditInfo.Projects)
```
zivkan: use `NoAllocEnumerate()` or whatever it is, or change the type to something without a heap allocated enumerator, or change to `IList<T>` and use a `for` loop.

src/NuGet.Clients/NuGet.CommandLine/Commands/RestoreCommand.cs
```diff
@@ -350,6 +370,15 @@ private async Task<IReadOnlyList<RestoreSummary>> PerformNuGetV2RestoreAsync(Pac
                     packagesFolderPath,
                     restoreSummaries);
 
+                using SourceCacheContext cacheContext = new();
```
zivkan, line 373: Should we care about the `-NoCache` CLI argument, with regards to checking the vulnerability index?

zivkan (commented):
I only looked at the product code so far, not tests, but generally looks good. I have a meeting now, so don't have time to finish reviewing today.

src/NuGet.Clients/NuGet.CommandLine/Commands/InstallCommand.cs
nkolev92: My reasoning was that we'd want to not disable audit by default, so having the boolean with a natural false default was safer. 

wyt? I can change it, I'm not tied to it.

src/NuGet.Core/NuGet.PackageManagement/Audit/AuditCheckResult.cs
nkolev92: These are not always known/available in every audit situation. 

I have tests ensuring things get set in the expected scenarios. Obviously that doesn't guard against future additions. 

I can have a bunch of internal constructors I guess if you'd prefer that.

src/NuGet.Core/NuGet.PackageManagement/Audit/AuditCheckResult.cs
nkolev92: > From a public API design point of view, personally I don't like that this is a public class, and therefore part of the public API, but with properties that other assemblies can't set. It makes it impossible to use in unit tests for a start, except for assemblies that this assembly lists as InternalsVisibleTo.
If the intention is that this type should only be created from within the assembly, never by others (even for unit tests), then I suggest adding an internal constructor, so it doesn't have a public constructor.

I'll change it to an internal constructor.

> But if this type needs to be public to be able to get the warnings and set the telemetry, then I think the properties should be public too. As I mentioned in another comment, the properties can change set to init, to prevent tampering, making it a "read only" (immutable, depending on your definition of immutable) type.

I feel this overexposes internals that are rather flexible. Metrics we'd want to collect would change but they don't fundamentally change the result of the vulnerability check.

src/NuGet.Core/NuGet.PackageManagement/Audit/AuditChecker.cs
nkolev92: The VS side has a lot of different ways of collecting the packages and they're almost always IEnumerable. The type that comes in within IPackageRestoreManager is IEnumerable. 
It was just easier to have that type. 

I can change if but it'd require creating a list in a bunch of places. 
Given that this is PC, and the restore is lightweight, I wasn't too worried.

src/NuGet.Core/NuGet.PackageManagement/Audit/AuditChecker.cs
nkolev92: I ended up counting packages. I'll remove the TODO. I thought I had gotten all of those, mb.

src/NuGet.Clients/NuGet.CommandLine/Commands/RestoreCommand.cs
nkolev92: Hmm, that's interesting.

We have the same concerns in the PR restore right? Can we handle those in a separate issue/PR?

nkolev92: I'd be happy to create an issue.

nkolev92: I created https://github.com/NuGet/Home/issues/13261.

src/NuGet.Core/NuGet.PackageManagement/Audit/AuditChecker.cs
nkolev92: Changed to List. While the preexisting API that sets this is IENumerable, it's always a list that's used.

nkolev92:
@zivkan Thanks for taking a look! 
I've addressed or responded too all of your feedback and it's ready for another look.

src/NuGet.Clients/NuGet.CommandLine/Commands/InstallCommand.cs
jeffkl: My vote is `enableNuGetAudit: true`

src/NuGet.Clients/NuGet.CommandLine/Commands/RestoreCommand.cs
```diff
@@ -324,15 +325,34 @@ private async Task<IReadOnlyList<RestoreSummary>> PerformNuGetV2RestoreAsync(Pac
 
                     throw new InvalidOperationException(message);
                 }
+                restoreAuditProperties = new();
```
jeffkl: Would this dictionary need to be case-insensitive?
```suggestion
                restoreAuditProperties = new(StringComparer.OrdinalIgnoreCase);
```

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -414,27 +403,23 @@ private NuGetAuditMode ParseAuditMode()
...
-                || string.Equals(value, "default", StringComparison.OrdinalIgnoreCase))
+            if (value == null)
             {
                 return true;
+
             }
-            if (string.Equals(value, bool.FalseString, StringComparison.OrdinalIgnoreCase)
-                || string.Equals(value, "disable", StringComparison.OrdinalIgnoreCase))
+
+            if (!value!.TryParseEnableAudit(out bool result))
```
jeffkl: This nullability suppression confused me, is it needed since above this the `value` is checked for `null`?
```suggestion
            if (!value.TryParseEnableAudit(out bool result))
```

src/NuGet.Core/NuGet.PackageManagement/Audit/AuditChecker.cs
```diff
@@ -168,22 +216,60 @@ internal static void CreateWarningsForPackagesWithVulnerabilities(Dictionary<Pac
...
-                        package.Id);
-                    logger.Log(restoreLogMessage);
+
+                    bool counted = false;
+                    for (int i = 0; i < auditInfo.Projects.Count; i++)
+                    {
+                        string projectPath = auditInfo.Projects[i];
+                        auditSettings.TryGetValue(projectPath, out (bool, PackageVulnerabilitySeverity) auditSetting);
+
+                        if (auditSetting == default || auditSetting.Item1 && (int)vulnerability.Severity >= (int)auditSetting.Item2)
```
jeffkl: Please use named Tuples, I'm not a fan of seeing `Item1`, `Item2`, `Item3` etc

src/NuGet.Core/NuGet.PackageManagement/IDE/PackageRestoreContext.cs
```diff
@@ -27,6 +30,19 @@ public class PackageRestoreContext
             EventHandler<PackageRestoreFailedEventArgs> packageRestoreFailedEvent,
             IEnumerable<SourceRepository> sourceRepositories,
             int maxNumberOfParallelTasks,
+            ILogger logger) : this(nuGetPackageManager, packages, token, packageRestoredEvent, packageRestoreFailedEvent, sourceRepositories, maxNumberOfParallelTasks, true, new Dictionary<string, RestoreAuditProperties>(), logger)
```
jeffkl: Can this pass an empty dictionary instead of new'ing one up?

nkolev92: I'll create a static.

src/NuGet.Clients/NuGet.CommandLine/Commands/RestoreCommand.cs
nkolev92: It needs the path aware version. There's only ever going to be 1 value here, so I think I wasn't too worried, but I should've done the path aware.

src/NuGet.Clients/NuGet.CommandLine/Commands/InstallCommand.cs
nkolev92: I'll address that then. :)

src/NuGet.Core/NuGet.PackageManagement/Audit/AuditChecker.cs
nkolev92: TryGetValue was finnicky last time I tried, but I was able to figure out the syntax file. It looks a bit dumb, but it works :D

src/NuGet.Clients/NuGet.CommandLine/Commands/RestoreCommand.cs
```diff
@@ -283,7 +283,7 @@ private async Task<IReadOnlyList<RestoreSummary>> PerformNuGetV2RestoreAsync(Pac
                 {
                     foreach (PackageReference packageReference in GetInstalledPackageReferences(configFile))
                     {
-                        if (configToProjectPath.TryGetValue(configFile, out string projectPath))
+                        if (!configToProjectPath.TryGetValue(configFile, out string projectPath))
```
jeffkl, line 286: This looks like you fixed a bug?

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -414,27 +403,23 @@ private NuGetAuditMode ParseAuditMode()
...
-                || string.Equals(value, "default", StringComparison.OrdinalIgnoreCase))
+            if (value == null)
             {
                 return true;
+
             }
-            if (string.Equals(value, bool.FalseString, StringComparison.OrdinalIgnoreCase)
-                || string.Equals(value, "disable", StringComparison.OrdinalIgnoreCase))
+
+            if (value!.TryParseEnableAudit(out bool result))
```
jeffkl: I still don't know why this `!` is needed.  the `if (value == null)` should make it not required?
```suggestion
            if (value.TryParseEnableAudit(out bool result))
```

src/NuGet.Core/NuGet.PackageManagement/Audit/AuditChecker.cs
```diff
@@ -168,22 +216,60 @@ internal static void CreateWarningsForPackagesWithVulnerabilities(Dictionary<Pac
...
-                        RestoreLogMessage.CreateWarning(logCode,
-                        message,
-                        package.Id);
-                    logger.Log(restoreLogMessage);
+
+                    bool counted = false;
+                    for (int i = 0; i < auditInfo.Projects.Count; i++)
+                    {
+                        string projectPath = auditInfo.Projects[i];
+                        auditSettings.TryGetValue(projectPath, out (bool IsAuditEnabled, PackageVulnerabilitySeverity MinimumSeverity) auditSetting);
```
jeffkl, line 224: 💯 so much better, thank you.

jeffkl (dismissed):

src/NuGet.Clients/NuGet.CommandLine/Commands/RestoreCommand.cs
nkolev92: yep, details about where a package was coming were never truly relevant before. 

Now that we have a config that's project specific that needs looked up after restore, it is relevant.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
nkolev92: I some version of the changes it was probably necessary. Not anymore though.
Nice catch.

jeffkl (dismissed):

src/NuGet.Core/NuGet.PackageManagement/Audit/AuditCheckResult.cs
zivkan: `init` does not mean `required`, it just means that the type is read only, and needs to have a new "instance" created to change a value. Using the [`with` expression](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/with-expression) makes this very easy. Therefore multiple constructors are not needed, unless you want to validate that when "property 1" is set "property 2" must be as well.

And when I reviewed this PR last week, I didn't see any examples of the type having any values changed after initialization. At least in the product code, I only saw properties being set on init.

> Obviously that doesn't guard against future additions.

This is the point of my suggestion. It does prevent future incorrect usage that might otherwise go unnoticed.

src/NuGet.Clients/NuGet.CommandLine/Commands/RestoreCommand.cs
```diff
@@ -458,6 +483,20 @@ private async Task<IReadOnlyList<RestoreSummary>> PerformNuGetV2RestoreAsync(Pac
             return configToProjectPath;
         }
 
+        private static Dictionary<string, RestoreAuditProperties> GetRestoreAuditProperties(PackageRestoreInputs packageRestoreInputs)
+        {
+            Dictionary<string, RestoreAuditProperties> restoreAuditProperties = new(PathUtility.GetStringComparerBasedOnOS());
+            foreach (PackageSpec project in packageRestoreInputs.ProjectReferenceLookup.Projects)
```
zivkan, line 489: Unless I'm using VSCode wrong, `.Projects` has type `IReadOnlyList<T>`, and therefore `foreach` will allocation.  I
```suggestion
            foreach (PackageSpec project in packageRestoreInputs.ProjectReferenceLookup.Projects.NoAllocateEnumerate())
```

src/NuGet.Core/NuGet.PackageManagement/Audit/AuditChecker.cs
```diff
@@ -55,18 +71,44 @@ public async Task CheckPackageVulnerabilitiesAsync(CancellationToken cancellatio
...
+            Dictionary<PackageIdentity, PackageAuditInfo>? packagesWithKnownVulnerabilities = FindPackagesWithKnownVulnerabilities(allVulnerabilityData.KnownVulnerabilities!, packages);
+            int Sev0Matches = 0, Sev1Matches = 0, Sev2Matches = 0, Sev3Matches = 0, InvalidSevMatches = 0;
+
+            List<PackageIdentity> packagesWithReportedAdvisories = new(packagesWithKnownVulnerabilities?.Count ?? 0);
+
+            IReadOnlyList<ILogMessage> warnings = packagesWithKnownVulnerabilities is not null ?
+                CreateWarnings(packagesWithKnownVulnerabilities, auditSettings, ref Sev0Matches, ref Sev1Matches, ref Sev2Matches, ref Sev3Matches, ref InvalidSevMatches, ref packagesWithReportedAdvisories) :
+                Array.Empty<ILogMessage>();
+
+            foreach (var warning in warnings)
```
zivkan, line 91:
```suggestion
            foreach (var warning in warnings.NoAllocEnumerate())
```

`warnings` is `IReadOnlyList<T>`, hence `foreach` will do a heap alloc by default.

zivkan (approved):
Two opportunities to follow best perf practises, but otherwise LGTM.

nkolev92:
thanks @zivkan

I'll address these in the next PR. Merging the PR allows me to unblock my VS work.

PR #5645, opened 2024-02-26
Do not clone objects in DependencyGraphSpec during command-line restore

### LINKED ISSUE #9041: Restore:  excessive deep cloning of ProjectSpec
When restoring ([`NuGet.Commands.RestoreRunner.RunAsync(...)`](https://github.com/NuGet/NuGet.Client/blob/74d1935e4e25a90df536ae7cac6cf7c73c6ea678/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreRunner.cs#L25)) NuGet makes wasteful use of memory with respect to `PackageSpec` objects.  This is especially noticable on large solutions.

[`DependencyGraphSpec.WithProjectClosure(...)`](https://github.com/NuGet/NuGet.Client/blob/74d1935e4e25a90df536ae7cac6cf7c73c6ea678/src/NuGet.Core/NuGet.ProjectModel/DependencyGraphSpec.cs#L125-L136) creates `PackageSpec` clones by default.  [`PackageSpec.Clone()`](https://github.com/NuGet/NuGet.Client/blob/74d1935e4e25a90df536ae7cac6cf7c73c6ea678/src/NuGet.Core/NuGet.ProjectModel/PackageSpec.cs#L192-L224) performs a fairly deep and expensive clone.  It's not clear that cloning by default is necessary.  Note that https://github.com/NuGet/NuGet.Client/pull/3094 was a targeted hack in that `DependencyGraphSpec` readonly-ness is not actually enforced and the optimization only applied to a specific command-line scenario.

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Home/issues/9041

## Description

@KirillOsenkov recently helped us out and mentioned that the linked issue is of great importance to his current efforts.  This change specifies the `readOnly` flag to the `DependencyGraphSpec` constructor indicating that the objects it contains do not need to be cloned.  I added this constructor for static graph-based restore but didn't add it to regular restore so this change makes them the same.

I added an overload to prevent a breaking API change even though I doubt anyone would be affected.  The existing constructor just keeps the existing behavior.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception - Existing test coverage will suffice
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
@@ -30,9 +30,19 @@ namespace NuGet.Commands
         private const string DoubleSlash = "//";
 
         /// <summary>
-        /// Convert MSBuild items to a DependencyGraphSpec.
+        /// Creates a <see cref="DependencyGraphSpec" /> from the specified MSBuild items.
         /// </summary>
+        /// <param name="items">An <see cref="IEnumerable{T}" /> of <see cref="IMSBuildItem" /> objects representing the MSBuild items gathered for restore.</param>
         public static DependencyGraphSpec GetDependencySpec(IEnumerable<IMSBuildItem> items)
+        {
+            return GetDependencySpec(items, readOnly: false);
+        }
+        /// <summary>
+        /// Creates a <see cref="DependencyGraphSpec" /> from the specified MSBuild items.
+        /// </summary>
+        /// <param name="items">An <see cref="IEnumerable{T}" /> of <see cref="IMSBuildItem" /> objects representing the MSBuild items gathered for restore.</param>
+        /// <param name="readOnly"><see langword="true" /> to indicate that the <see cref="DependencyGraphSpec" /> is considered read-only and won't be changed, otherwise <see langword="false" />.</param>
+        public static DependencyGraphSpec GetDependencySpec(IEnumerable<IMSBuildItem> items, bool readOnly)
         {
             if (items == null)
             {
@@ -44,7 +54,7 @@ namespace NuGet.Commands
             // To workaround this unique names should be compared based on the OS.
             var uniqueNameComparer = PathUtility.GetStringComparerBasedOnOS();
 
-            var graphSpec = new DependencyGraphSpec();
+            var graphSpec = new DependencyGraphSpec(readOnly);
             var itemsById = new Dictionary<string, List<IMSBuildItem>>(uniqueNameComparer);
             var restoreSpecs = new HashSet<string>(uniqueNameComparer);
             var validForRestore = new HashSet<string>(uniqueNameComparer);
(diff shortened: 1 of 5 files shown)


src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
```diff
@@ -30,9 +30,19 @@ public static class MSBuildRestoreUtility
         private const string DoubleSlash = "//";
 
         /// <summary>
-        /// Convert MSBuild items to a DependencyGraphSpec.
+        /// Creates a <see cref="DependencyGraphSpec" /> from the specified MSBuild items.
         /// </summary>
+        /// <param name="items">An <see cref="IEnumerable{T}" /> of <see cref="IMSBuildItem" /> objects representing the MSBuild items gathered for restore.</param>
         public static DependencyGraphSpec GetDependencySpec(IEnumerable<IMSBuildItem> items)
```
nkolev92, line 36: Any reason not to make this the default?

It's only used in CLI scenarios right?
It's VS where mutation is a concern correct?

nkolev92 (approved):
I'm comfortable with changing the behavior of the API altogether if it's only called from the CLI. 

fwiw, I really thought we were already doing this in both static graph and standard restore.

nkolev92:
@jeffkl Should this "close" the issue it's linked? 

Given that we're not fixing the VS side, I think that issue should remain open.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
zivkan: It's not clear to me why VS would need to mutate a DGSpec. If projects are loaded/unloaded, a new DGSpec can be created with the new project list.

jeffkl:
> @jeffkl Should this "close" the issue it's linked?
> 
> Given that we're not fixing the VS side, I think that issue should remain open.

I almost opened a new issue but hesitated because I don't think we'll ever "fix" the Visual Studio side of things since it needs to mutate the dgspec right?  The person who opened the original issue really only cares about command-line based restore at the moment so I figured this pull request would fix the issue and if we wanted we could open a new one for Visual Studio restores?  Or I can open a new one for command-line based restores?  I do not have a strong opinion.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
jeffkl: I would also like a better understanding of why the items in the DG spec are cloned?

nkolev92: The readonly part can be a bit misleading.
Readonly means that the PackageSpec is not cloned. 

The primary reason around package spec cloning is defensiveness. In Visual Studio as a long lived process, the PackageSpec is not guaranteed to be immutable. 

After this, we're down to 2 clones, 1 in the RestoreRunner, 1 in the CPSPackageReferenceProject, the latter one is needed, but the previous I'm 99% sure we don't need anymore. 

Long term the approach should be; 

1. Introduce a PackageSpec lock mechanism, and the dg spec would only accept locked PackageSpec. Then it can just avoid cloning.

jeffkl:
I've opened https://github.com/NuGet/Home/issues/13270 as a way of tracking the removal of the cloning in Visual Studio code paths.

PR #5655, opened 2024-02-29
Add 64-bit non crypto hash algo for dgspec uniqueness computation

### LINKED ISSUE #13214: ProjectModel.HashObjectWriter.OnFlush is using a SHA512 hash versus a cheaper hash which seems like overkill
### NuGet Product(s) Affected

MSBuild.exe, dotnet.exe

### Current Behavior

Originally filed by the VS Perf team.  The `ProjectModel.HashObjectWriter` currently uses a SHA512 hash algorithm and we're wondering if a faster hash algorithm would still work and speed things up.

### Desired Behavior

A faster hashing algorithm that still yields the same outcome.

### Additional Context

Originally reported at https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1987168

GenelleM:
## Bug
Fixes: https://github.com/NuGet/Home/issues/13214

## Description
Switch calculating dependency graph spec json to a 64-bit non-crypto hash from SHA512 hash to reduce comparison/restore time but maintaining accuracy of need to restore when determining if NuGet needs to regenerate the dgspec.json file. 

We've also added an escape hatch a user can set the environment variable `NUGET_ENABLE_LEGACY_DGSPEC_HASH_FUNCTION` to `true` to bring back the SHA512 hash function.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.ProjectModel/DependencyGraphSpec.cs
+++ b/src/NuGet.Core/NuGet.ProjectModel/DependencyGraphSpec.cs
@@ -14,6 +14,11 @@ namespace NuGet.ProjectModel
 {
     public class DependencyGraphSpec
     {
+        /// <summary>
+        /// Allows a user to enable the legacy SHA512 hash function for dgSpec files which is used by no-op.
+        /// </summary>
+        private static readonly bool UseLegacyHashFunction = string.Equals(Environment.GetEnvironmentVariable("NUGET_ENABLE_LEGACY_DGSPEC_HASH_FUNCTION"), bool.TrueString, StringComparison.OrdinalIgnoreCase);
+
         private const string DGSpecFileNameExtension = "{0}.nuget.dgspec.json";
 
         private readonly SortedSet<string> _restore = new(PathUtility.GetStringComparerBasedOnOS());
@@ -300,7 +305,8 @@ namespace NuGet.ProjectModel
 
         public string GetHash()
         {
-            using (var hashFunc = new FnvHash64Function())
+            // Use the faster FNV hash function for hashing unless the user has specified to use the legacy SHA512 hash function
+            using (IHashFunction hashFunc = UseLegacyHashFunction ? new Sha512HashFunction() : new FnvHash64Function())
             using (var writer = new HashObjectWriter(hashFunc))
             {
                 Write(writer, hashing: true, PackageSpecWriter.Write);
--- a/src/NuGet.Core/NuGet.ProjectModel/DependencyGraphSpec.cs
+++ b/src/NuGet.Core/NuGet.ProjectModel/DependencyGraphSpec.cs
@@ -300,7 +300,7 @@ namespace NuGet.ProjectModel
 
         public string GetHash()
         {
-            using (var hashFunc = new Sha512HashFunction())
+            using (var hashFunc = new FnvHash64Function())
             using (var writer = new HashObjectWriter(hashFunc))
             {
                 Write(writer, hashing: true, PackageSpecWriter.Write);
--- a/test/NuGet.Core.Tests/NuGet.ProjectModel.Test/FnvHash64FunctionTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.ProjectModel.Test/FnvHash64FunctionTests.cs
@@ -1,0 +1,46 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Text;
+using Xunit;
+
+namespace NuGet.ProjectModel.Test
+{
+    public class FnvHash64FunctionTests
+    {
+        private const string ExpectedResult = "BSgChkz/Y68=";
+        private static readonly byte[] Input = Encoding.UTF8.GetBytes("BeachClubExtraAvocadoSpread");
+
+        [Fact]
+        public void Update_SupportsIncrementalUpdates()
+        {
+            using var hashFunc = new FnvHash64Function();
+
+            for (var i = 0; i < Input.Length; ++i)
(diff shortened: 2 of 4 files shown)


GenelleM:
@jeffkl

src/NuGet.Core/NuGet.Packaging/FnvHash64Function.cs
```diff
@@ -0,0 +1,92 @@
...
+using System;
+
+namespace NuGet.Packaging
+{
+    /// <summary>
+    /// An Fnv 64-bit hash function that supports non-cryptographic hashing to optimize speed and minimize collisions.
+    /// 
+    /// This is non-private only to facilitate unit testing.
+    /// </summary>
+    public sealed class FnvHash64Function : IHashFunction
```
nkolev92: This type is only used in the NuGet.ProjectModel assembly. It should be moved there. 

Given that it's specific to that one assembly, I think we can make the type internal as well. 
Some guidance about adding public APIs: https://github.com/NuGet/NuGet.Client/blob/dev/docs/nuget-sdk.md#changes-to-apis.

src/NuGet.Core/NuGet.Packaging/PackagesConfigReader.cs
```diff
@@ -144,7 +144,7 @@ public IEnumerable<PackageReference> GetPackages(bool allowDuplicatePackageIds)
             {
                 string id = null;
                 if (!PackagesConfig.TryGetAttribute(package, "id", out id)
-                    || String.IsNullOrEmpty(id))
```
nkolev92: The only changes in this file are non-functional, so you can probably just revert.

nkolev92 (commented):
How do the numbers look like before/after the change?
How much of a difference is this making?

test/NuGet.Core.Tests/NuGet.ProjectModel.Test/FnvHash64FunctionTests.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Text;
+using NuGet.Packaging;
+using Xunit;
+
+namespace NuGet.ProjectModel.Test
+{
+    public class FnvHash64FunctionTests
+    {
+        private const string ExpectedResult = "BSgChkz/Y68=";
+        private static readonly byte[] Input = Encoding.UTF8.GetBytes("BeachClubExtraAvocadoSpread");
```
jeffkl, line 13: 🤣

src/NuGet.Core/NuGet.Packaging/FnvHash64Function.cs
jeffkl: I agree, let's move this to `NuGet.ProjectModel` and make it `internal`.

src/NuGet.Core/NuGet.ProjectModel/DependencyGraphSpec.cs
```diff
@@ -300,7 +300,7 @@ public void Save(Stream stream)
 
         public string GetHash()
         {
-            using (var hashFunc = new Sha512HashFunction())
+            using (var hashFunc = new FnvHash64Function())
```
jeffkl: It would be a ton of work to have an escape hatch other than env var to bring back the old behavior.  Should we check for an env var to bring back the SHA512 hash just in case?

src/NuGet.Core/NuGet.Packaging/FnvHash64Function.cs
jeffkl: I've address feedback for Genelle, please re-review

src/NuGet.Core/NuGet.ProjectModel/DependencyGraphSpec.cs
nkolev92: I think that's a great idea.

nkolev92 (dismissed):
Let's add the env var just in case and 🚢 it.

src/NuGet.Core/NuGet.ProjectModel/DependencyGraphSpec.cs
jeffkl: I've added an escape hatch `NUGET_ENABLE_LEGACY_DGSPEC_HASH_FUNCTION=true`

src/NuGet.Core/NuGet.ProjectModel/DependencyGraphSpec.cs
```diff
@@ -14,6 +14,11 @@ namespace NuGet.ProjectModel
 {
     public class DependencyGraphSpec
     {
+        /// <summary>
+        /// Allows a user to enable the legacy SHA512 hash function for dgSpec files which is used by no-op.
+        /// </summary>
+        private static readonly bool UseLegacyHashFunction = string.Equals(Environment.GetEnvironmentVariable("NUGET_ENABLE_LEGACY_DGSPEC_HASH_FUNCTION"), bool.TrueString, StringComparison.OrdinalIgnoreCase);
```
nkolev92, line 20: This can be done later, but we tend to prefer EnvironmentVariableReaderWrapper for checks like this.

nkolev92 (approved):

src/NuGet.Core/NuGet.ProjectModel/DependencyGraphSpec.cs
jeffkl: Yeah I wasn't going to bother with unit tests so for now I just directly read the value.  If needed, I can change it later on as you mentioned.

PR #5656, opened 2024-02-29
Use error handler succeeded instead of throwing when we fail to get solution information

nkolev92:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2647

## Description

In some cases, the solution properties cannot be obtained for various reasons. 
I can't think of why this wouldn't be safe for us to do. Having the fault reporting isn't really helping us do anything actionable anyways.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception - There's already plenty of tests here. Every scenario we have runs this codepath.
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSSolutionManager.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSSolutionManager.cs
@@ -542,8 +542,11 @@ namespace NuGet.PackageManagement.VisualStudio
         private static object GetVSSolutionProperty(IVsSolution vsSolution, int propId)
         {
             ThreadHelper.ThrowIfNotOnUIThread();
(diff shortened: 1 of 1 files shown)


zivkan (approved):

jeffkl (approved):

donnie-msft (approved):
This may cause new NullReferenceExceptions
eg, https://github.com/NuGet/NuGet.Client/blob/b1da05f5eef1a62aad556566ebcb4e96286aa935/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSSolutionManager.cs#L512

nkolev92:
yeah, it could in theory.

Somehow that one always succeeds.

The API is inconsistent or I at least I don't follow the pattern :)

PR #5662, opened 2024-03-01
Use StringBuilder.AppendInt() in more places to reduce allocations when appending integers to a StringBuilder

### LINKED ISSUE #13282: Usage of StringBuilder.Append() allocates when appending an int
`StringBuilder.Append()` calls `ToString()` which can be avoided by using our `AppendInt()` extension method.

Erarndt:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13282

## Description

`StringBuilder.Append(int)` allocates a string in order to append an integer so this code adds an extension method which uses `char`s instead to avoid allocating unnecessary strings.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception - Existing coverage
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/build/Shared/SharedExtensions.cs
+++ b/build/Shared/SharedExtensions.cs
@@ -6,6 +6,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text;
 
 namespace NuGet.Shared
 {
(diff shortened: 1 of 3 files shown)


jeffkl (approved):

build/Shared/SharedExtensions.cs
```diff
@@ -62,5 +63,68 @@ public static void ForEach<T>(this IEnumerable<T> enumeration, Action<T> action)
...
+
+            // remember when we found our first digit so we can keep adding intermediate zeroes
+            bool digitFound = false;
+            while (divisor > 0)
+            {
+                if (digitFound || value >= divisor)
+                {
+                    digitFound = true;
+                    int digit = value / divisor;
+                    value -= digit * divisor;
```
zivkan, line 116: Consider using `Math.DivRem` instead. Some (all?) CPU architectures have the divide instruction calculate both the quotient and the remainder, but when you only use one of the two, the other is discarded. Typically, (native) compilers will detect when you do both `a / b` followed by `a % b`, and then only execute a single divide CPU instruction, but the .NET BCL has this `Math.DivRem` method which I assume is also very likely to use the optimized implementation.

PR #5666, opened 2024-03-02
Fix sources equality check in ProjectRestoreMetadata

### LINKED ISSUE #13269: [.NET 9 Preview 3] no-op restore is no longer a no-op
### NuGet Product Used

dotnet.exe

### Product Version

.NET SDK 9.0.100-preview.3.24126.2

### Worked before?

.NET SDK 9.0.100-preview.2.24122.3

### Impact

### Repro Steps & Context

We discovered here a test failure: https://github.com/xamarin/xamarin-android/pull/8763

The underlying cause appears to be:
```
The restore inputs for 'UnnamedProject' have changed. Continuing restore.
```

This was an incremental build with no changes to the project.

Digging deeper, the offending build seems to have some NuGet sources listed twice:
```diff
++https://api.nuget.org/v3/index.json
++https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8/nuget/v3/index.json
https://api.nuget.org/v3/index.json
https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8/nuget/v3/index.json
```
So there must be something that adds extra feeds, making the no-op restore skip? so it runs a full restore?

### Verbose Logs

Here are two `.binlog` files, `install2.binlog` restores as expected, `install3.binlog` does not:
[nuget-noop.zip](https://github.com/NuGet/Home/files/14421956/nuget-noop.zip)

### LINKED ISSUE #13294: Duplicate sources can lead to unnecessary assets file writes.
The recent changes from https://github.com/NuGet/NuGet.Client/commit/1942c17866695ae1175ec3d8b42a6b41bcd6a9c9#diff-93482a8631efd30f5df833baa903bcf342fb99a2e93cf9a3dea6a672b44d3a61L81 and the changes moving from Newtonsoft.Json to System.Text.Json for assets file reading. 

If there are duplicate objects, Newtonsoft.Json dedups, while System.Text.Json *does not*. 

```json
      "sources": {
        "C:\\Program Files (x86)\\Microsoft SDKs\\NuGetPackages\\": {},
        "C:\\Program Files\\dotnet\\library-packs": {},
        "https://api.nuget.org/v3/index.json": {},
        "https://api.nuget.org/v3/index.json": {}
      }
```

This lead to perf regressions.

nkolev92:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13294
Fixes: https://github.com/NuGet/Home/issues/13269

## Description

The recent changes from https://github.com/NuGet/NuGet.Client/commit/1942c17866695ae1175ec3d8b42a6b41bcd6a9c9#diff-93482a8631efd30f5df833baa903bcf342fb99a2e93cf9a3dea6a672b44d3a61L81 and the changes moving from Newtonsoft.Json to System.Text.Json for assets file reading. 

If there are duplicate objects, Newtonsoft.Json dedups, while System.Text.Json *does not*. 

```json
      "sources": {
        "C:\\Program Files (x86)\\Microsoft SDKs\\NuGetPackages\\": {},
        "C:\\Program Files\\dotnet\\library-packs": {},
        "https://api.nuget.org/v3/index.json": {},
        "https://api.nuget.org/v3/index.json": {}
      }
```

Deduping the sources based on the source value will allow the comparison to work in *both* STJ and NJ scenarios. 
Eventually, we'd probably want to remove the deduplication, but given that we want this codepath to work with both, this is the best approach. 

Note that previous ordering and current deduplication costs are comparable.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.ProjectModel/ProjectRestoreMetadata.cs
+++ b/src/NuGet.Core/NuGet.ProjectModel/ProjectRestoreMetadata.cs
@@ -190,7 +190,7 @@ namespace NuGet.ProjectModel
                    osStringComparer.Equals(OutputPath, other.OutputPath) &&
                    osStringComparer.Equals(ProjectName, other.ProjectName) &&
                    osStringComparer.Equals(ProjectUniqueName, other.ProjectUniqueName) &&
-                   Sources.Distinct().OrderedEquals(other.Sources.Distinct(), source => source.Source, StringComparer.OrdinalIgnoreCase) &&
+                   GetSources(Sources).SetEqualsWithNullCheck(GetSources(other.Sources), StringComparer.OrdinalIgnoreCase) &&
                    osStringComparer.Equals(PackagesPath, other.PackagesPath) &&
                    ConfigFilePaths.OrderedEquals(other.ConfigFilePaths, filePath => filePath, osStringComparer, osStringComparer) &&
                    FallbackFolders.OrderedEquals(other.FallbackFolders, fallbackFolder => fallbackFolder, osStringComparer, osStringComparer) &&
@@ -208,6 +208,16 @@ namespace NuGet.ProjectModel
                    EqualityUtility.EqualsWithNullCheck(CentralPackageVersionOverrideDisabled, other.CentralPackageVersionOverrideDisabled) &&
                    EqualityUtility.EqualsWithNullCheck(CentralPackageTransitivePinningEnabled, other.CentralPackageTransitivePinningEnabled) &&
                    RestoreAuditProperties == other.RestoreAuditProperties;
+        }
+
+        private HashSet<string> GetSources(IList<PackageSource> sources)
+        {
+            var setSources = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            for (var i = 0; i < sources.Count; i++)
+            {
+                setSources.Add(sources[i].Source);
+            }
+            return setSources;
         }
 
         public virtual ProjectRestoreMetadata Clone()
--- a/src/NuGet.Core/NuGet.ProjectModel/ProjectRestoreMetadata.cs
+++ b/src/NuGet.Core/NuGet.ProjectModel/ProjectRestoreMetadata.cs
@@ -212,7 +212,7 @@ namespace NuGet.ProjectModel
 
         private HashSet<string> GetSources(IList<PackageSource> sources)
         {
-            var setSources = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            var setSources = new HashSet<string>(sources.Count, StringComparer.OrdinalIgnoreCase);
             for (var i = 0; i < sources.Count; i++)
             {
                 setSources.Add(sources[i].Source);
--- a/src/NuGet.Core/NuGet.ProjectModel/ProjectRestoreMetadata.cs
+++ b/src/NuGet.Core/NuGet.ProjectModel/ProjectRestoreMetadata.cs
@@ -212,7 +212,11 @@ namespace NuGet.ProjectModel
 
         private HashSet<string> GetSources(IList<PackageSource> sources)
         {
+#if NETSTANDARD2_0
+            var setSources = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+#else
             var setSources = new HashSet<string>(sources.Count, StringComparer.OrdinalIgnoreCase);
+#endif
             for (var i = 0; i < sources.Count; i++)
             {
                 setSources.Add(sources[i].Source);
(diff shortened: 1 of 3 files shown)


src/NuGet.Core/NuGet.ProjectModel/ProjectRestoreMetadata.cs
```diff
@@ -190,7 +190,7 @@ public bool Equals(ProjectRestoreMetadata other)
                    osStringComparer.Equals(OutputPath, other.OutputPath) &&
                    osStringComparer.Equals(ProjectName, other.ProjectName) &&
                    osStringComparer.Equals(ProjectUniqueName, other.ProjectUniqueName) &&
-                   Sources.OrderedEquals(other.Sources.Distinct(), source => source.Source, StringComparer.OrdinalIgnoreCase) &&
+                   Sources.Distinct().OrderedEquals(other.Sources.Distinct(), source => source.Source, StringComparer.OrdinalIgnoreCase) &&
```
zivkan: Many (most?) repos probably won't have duplicates, but doing `Distinct` twice is going to cause a bunch of additional allocations.  Would it be better to remove the `Distinct` instead, to get the (minor) perf benefit?

nkolev92: NJ deduplicates at deserialization, while STJ does not. 
This is the only implementation that'll work for both.

zivkan: Won't that only affect people on the first restore after switching between NJ and STJ (including upgrading NuGet from a version that didn't previously support it).

Is avoiding one-time "warm" restore worth the every time perf cost?

Sometimes new NuGet features that modify the PackageSpec cause the DGSpec hash to change, so the first restore after a new NuGet version upgrade will fail the no-op check anyway. But it's only the first time after an upgrade, afterwards no-op still works.

nkolev92: I'm not sure I follow the one time switch comments. 
The hashing is actually not affected, since we never hash a package spec from disk, it's only assets file dirty checks that are broken. 

I'll more details in the issue, but my intention is to switch to just use comparison without distinct once we've permanently switched to System.Text.Json.

nkolev92: Ready for review.

jgonz120: Where does NJ do the deduplication? Just curious where I missed this, the code where in JsonPackageSpecReader doesn't look it would dedup the sources.

jgonz120 (dismissed):

src/NuGet.Core/NuGet.ProjectModel/ProjectRestoreMetadata.cs
nkolev92: ReadObject seems to do it automatically. 
Couldn't figure out where/why, but it was just happening.

kartheekp-ms (approved):

jeffkl:
Does it make sense to add a new PackageSource comparer?  I'm not sure how many other places in our code could benefit from something like [this](https://dotnetfiddle.net/Bt3r5y).
```c#
internal class PackageSourceComparer : IEqualityComparer<PackageSource>
{
	private enum Mode
	{
		All,
		NameOnly,
		SourceOnly,
	}

	private readonly Mode _mode;

	private readonly IEqualityComparer<string> _stringComparer;

	public static PackageSourceComparer SourceOnlyOrdinalIgnoreCase { get; } = new PackageSourceComparer(Mode.SourceOnly, StringComparer.OrdinalIgnoreCase);

	PackageSourceComparer(Mode mode, StringComparer stringComparer)
	{
		_mode = mode;
		 _stringComparer = stringComparer;
	}

	public bool Equals(PackageSource x, PackageSource y)
	{
		switch (_mode)
		{
			case Mode.All :
				return x.Equals(y);
				
			case Mode.NameOnly :
				return _stringComparer.Equals(x.Name, y.Name);

			case Mode.SourceOnly :
				return _stringComparer.Equals(x.Source, y.Source);
		}

		return false;
	}

	public int GetHashCode(PackageSource obj)
	{
		switch (_mode)
		{
			case Mode.All :
				return obj.GetHashCode();

			case Mode.NameOnly :
				return _stringComparer.GetHashCode(obj.Name);

			case Mode.SourceOnly :
				return _stringComparer.GetHashCode(obj.Source);
		}

		return 0;
	}
}
```

PR #5669, opened 2024-03-05
Set ngen architecture via property

zivkan:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2722

Regression? no

## Description

Change all `<NgenArchitecture>all` item metadata to instead get the value from a property with the same name. When the property doesn't exist, then the item effectively won't either. Makes makes it a simple 1 line change in the future to set/unset when we need, but for now we won't set it.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception: no way to test within this repo that the outcome is as desired.
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.VisualStudio.Client/NuGet.VisualStudio.Client.csproj
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Client/NuGet.VisualStudio.Client.csproj
@@ -100,31 +100,31 @@
     <ProjectReference Include="..\..\NuGet.Core\NuGet.Common\NuGet.Common.csproj">
       <Ngen>True</Ngen>
       <NgenApplication>$(NgenApplicationDefault)</NgenApplication>
-      <NgenArchitecture>All</NgenArchitecture>
+      <NgenArchitecture>$(NgenArchitecture)</NgenArchitecture>
       <NgenPriority>$(NgenPriorityDefault)</NgenPriority>
     </ProjectReference>
     <ProjectReference Include="..\..\NuGet.Core\NuGet.Configuration\NuGet.Configuration.csproj">
       <Ngen>True</Ngen>
       <NgenApplication>$(NgenApplicationDefault)</NgenApplication>
-      <NgenArchitecture>All</NgenArchitecture>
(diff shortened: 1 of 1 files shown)


davkean (approved):

jeffkl (approved):

src/NuGet.Clients/NuGet.VisualStudio.Client/NuGet.VisualStudio.Client.csproj
```diff
@@ -100,31 +100,31 @@
     <ProjectReference Include="..\..\NuGet.Core\NuGet.Common\NuGet.Common.csproj">
       <Ngen>True</Ngen>
       <NgenApplication>$(NgenApplicationDefault)</NgenApplication>
-      <NgenArchitecture>All</NgenArchitecture>
+      <NgenArchitecture>$(NgenArchitecture)</NgenArchitecture>
```
nkolev92, line 103: What does not having this imply?

I see the recommendation in the original issue, but completely sure why.

nkolev92 (approved):

src/NuGet.Clients/NuGet.VisualStudio.Client/NuGet.VisualStudio.Client.csproj
zivkan: I don't know 🤷 the team recommending this knows more about ngen than I do. They say that with the `all` value we currently have it tells ngen (well, vsn.exe, I think) to create both 32 and 64 native images, which wastes time and makes VS updates slower unnecesserily.  I don't know how ngen/vsn.exe knows which platform to ngen when there is no value, but if it works, it works.

davkean: My recommendation was to delete it to use the default, this is basically equivalent of that because NgenAchitecture won't be set in most builds.

PR #5504, opened 2023-11-11
Add `disableTLSCertificateValidation` property to nuget.config

Nigusu-Allehu:
## Bug 

Fixes: https://github.com/NuGet/Home/issues/12996

Regression? Last working version:

## Description

We need to add a `disableTLSCertificateValidation` property into `packageSources` section in NuGet.Config files, as below:
```
<!-- Disables certification validation on a specific https source -->
<packageSources>
    <add key="Contoso" value="https://contoso.com/packages/" disableTLSCertificateValidation="true" />
</packageSources>
```
spec : https://github.com/NuGet/Home/blob/dev/proposed/2023/InsecureConnectionsDisableCertificateValidation.md#package-source-nuget-config 
## PR Checklist

- [x] PR has a meaningful title
- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [ ] Documentation PR or issue filled : 
  - **OR**
  - [x] N/A

PR #5674, opened 2024-03-08
Allow HTTPS sources that have SSL certificate problems

### LINKED ISSUE #4387: NuGet cannot restore from HTTPS sources that have SSL certificate problems
## Idea

We should add a way for users to bypass certain HTTPS certificate errors. If the user has a source that does not have a valid certificate chain, they should still have some way of getting NuGet to interact with this source.

## "Inspirations"

npm and yarn have this:
```
npm config set strict-ssl false
yarn config set strict-ssl false
```

Maven has some extra knobs and switches:
```
-Dmaven.wagon.http.ssl.allowall=true
-Dmaven.wagon.http.ssl.insecure=true
-Dmaven.wagon.http.ssl.ignore.validity.dates=true
```

PHP's Composer
```json
{
    "config": {
        "disable-tls": true,
        "secure-http": false
    }
}
```

curl has it:
```
--no-check-certificate
```

## Specific Scenario

nuget.org server side runs functional tests against the staging slot of DEV, INT, and PROD that use nuget.exe. The DNS for this staging slot is some random blah.cloudapp.net. The certificate used is associated with the production DNS (e.g. *.nuget.org). This means nuget.exe fails SSL checks when running against this staging slot. 

## Proposal

Add a NuGet.config `trustedCertificates` section with standard `<add>` and `<remove>` tags. The value should be the SHA1 fingerprint (a.k.a. thumbprint) of the certificate. This allows developers to specifically opt-in into the tenuous state of SSL validation failures. The risk is mitigated since the developer must opt into it and even then only on a case-by-case basic.

Any certificate added to the `trustedCertificates` set is excluded from SSL validations -- I am thinking ALL validations exposed to use via the .NET API (untrusted root, name mistmatch, etc).

To improve discovery, we can add a log which writes out the fingerprint of the certificate that failed SSL validation at the `error` level.

We should also consider adding an environment variable to set this value. I believe npm has this and this  sibling environment variable would match other NuGet.config keys, such as the location of the user packages directory.

## Implementation

`HttpClientHandler` in `System.Net.Http` 4.3.0 (maybe earlier) has a way to set the certification validation check. This allows an app to configure the certificate validation on a per-request basis, instead of a per-AppDomain basis previously provided by `ServicePointManager` static properties.

With this approach, we should be able to implement the settings for all of our entry points: nuget.exe, .NET CLI, and (most tricky) Visual Studio.

The only possible problem is TFM compatibility with the `System.Net.Http` package. This requires investigation.

## Issues

- [x] https://github.com/NuGet/Home/issues/12996
- [x] https://github.com/NuGet/Home/issues/13023

Nigusu-Allehu:
## Bug

Fixes: https://github.com/NuGet/Home/issues/4387

## Description

This PR is merging a feature branch that:
- Adds disableTLSCertificateValidation option : https://github.com/NuGet/NuGet.Client/pull/5504
- Makes use of it to allow insecure connection with sources with no valid TLS certificate : https://github.com/NuGet/NuGet.Client/pull/5514
## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [ ] Documentation PR or issue filled (https://github.com/NuGet/Home/issues/13478)
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetSourcesService.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetSourcesService.cs
@@ -94,6 +94,7 @@ namespace NuGet.PackageManagement.VisualStudio
                         && packageSource.Source.Equals(packageSourceContextInfo.Source, StringComparison.InvariantCulture)
                         && packageSource.ProtocolVersion == packageSourceContextInfo.ProtocolVersion
                         && packageSource.AllowInsecureConnections == packageSourceContextInfo.AllowInsecureConnections
+                        && packageSource.DisableTLSCertificateValidation == packageSourceContextInfo.DisableTLSCertificateValidation
                         && packageSource.IsEnabled == packageSourceContextInfo.IsEnabled)
                     {
                         newPackageSources.Add(packageSource);
@@ -113,6 +114,7 @@ namespace NuGet.PackageManagement.VisualStudio
                             Description = packageSource.Description,
                             ProtocolVersion = packageSourceContextInfo.ProtocolVersion,
                             AllowInsecureConnections = packageSourceContextInfo.AllowInsecureConnections,
+                            DisableTLSCertificateValidation = packageSourceContextInfo.DisableTLSCertificateValidation,
                             MaxHttpRequestsPerSource = packageSource.MaxHttpRequestsPerSource,
                         };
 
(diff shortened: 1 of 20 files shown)


nkolev92:
Let's hold this for 6.11, we're in P3, so we're feature complete already.

src/NuGet.Clients/NuGet.CommandLine/Commands/ProjectFactory.cs
```diff
@@ -761,6 +761,7 @@ private PackageDependency CreateDependencyFromProject(dynamic project, Dictionar
 
                 projectFactory.InitializeProperties(builder);
 
+#pragma warning disable CS0612 // Type or member is obsolete
```
kartheekp-ms: These changes are not relevant to this pull-request. May be revert the https://github.com/NuGet/NuGet.Client/pull/5674/commits/c27cbfc95402afb1a1771f9a7324cf35f08c93cd commit and rebase your branch again on top of dev.

nkolev92 (approved):

kartheekp-ms (approved):

PR #5679, opened 2024-03-09
Enable advisory suppressions in CLI restore for PackageReference projects

### LINKED ISSUE #13313: `MSBuildRestoreUtility.GetRestoreAuditProperties` needs a breaking change to read `NuGetAuditSuppress` items
### NuGet Product(s) Affected

NuGet.exe, Visual Studio Package Management UI, Visual Studio Package Manager Console, MSBuild.exe, dotnet.exe, NuGet SDK

### Current Behavior

The `GetRestoreAuditProperties()` method in `NuGet.Commands.MSBuildRestoreUtility` currently just takes a single `IMSBuildItem specItem` input, which contains all the relevant NuGetAudit MSBuild Properties.

### Desired Behavior

We're adding a new MSBuild Item for NuGetAudit with `NuGetAuditSuppress`. To read this item, we will need to update the method signature for `GetRestoreAuditProperties()` to accept an additional parameter, `System.Collections.Generic.IList<string> suppressionItems`.

advay26:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2720
Fixes: https://github.com/NuGet/Home/issues/13313

## Description

Design spec: https://github.com/NuGet/Home/blob/dev/accepted/2023/NuGetAudit-supress-advisory.md

We're adding a new MSBuild item called `NuGetAuditSuppress` that allows users to suppress individual advisories during restore. These items will have the advisory URL in the `Include=` field, looking something like this:

`<NuGetAuditSuppress Include="https://github.com/advisories/GHSA-5crp-9r3c-p9vr" />`.

This PR implements suppressions for CLI restore paths (dotnet/MSBuild restore, NuGet.exe restore, Static graph restore) for PackageReference projects. The packages.config implementation, as well as the VS restore implementation for PackageReference projects, will follow later.

These are the main pieces:
- To read the new items, we added a new 'Collect' target called `CollectNuGetAuditSuppressions` (`src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets`)
- We also added a new target called `_CollectRestoreInputs`, which will run all the individual 'Collect' item targets. This is meant to reduce duplication in our code. Going forward, instead of adding new 'Collect' targets to lists in multiple places, we will be able to add it to the `DependsOnTargets=` attribute in `_CollectRestoreInputs`, and use this target everywhere. (See [_CollectRestoreInputs target](https://github.com/NuGet/Home/blob/dev/accepted/2023/NuGetAudit-supress-advisory.md#_collectrestoreinputs-target))
- The new `GetRestoreNuGetAuditSuppressionsTask` parses the 'NuGetAuditSuppress' items into the `_RestoreGraphEntry` items that we ultimately read in restore.
- `src/NuGet.Core/NuGet.ProjectModel/` holds all the changes related to reading and writing the dgspec.
- A new `HashSet<string> SuppressedAdvisories` member was added `src/NuGet.Core/NuGet.ProjectModel/RestoreAuditProperties.cs` to store the suppression inputs.
- The business logic for suppressions is in the `src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs` file.
- I've added tests too, but I'm not sure if they cover everything. Let me know if I need to add tests anywhere else, and I'll go back and make the changes.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [x] Documentation PR or issue filled: https://github.com/NuGet/Home/issues/13493
  - **OR**
  - [ ] N/A


src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
```diff
@@ -37,14 +37,11 @@ internal sealed class MSBuildStaticGraphRestore : IDisposable
...
-        /// Represents the small list of targets that must be executed in order for PackageReference, PackageDownload, and FrameworkReference items to be accurate.
+        /// Represents the small list of targets that must be executed in order for various restore input items to be accurate.
         /// </summary>
         private static readonly string[] TargetsToBuild =
         {
-            "CollectPackageReferences",
-            "CollectPackageDownloads",
-            "CollectFrameworkReferences",
-            "CollectCentralPackageVersions"
+            "_CollectRestoreInputs"
```
zivkan, line 44: I see that @nkolev92 provided a comment in the spec to rename it this, but, seeing this pull request, it got me wondering. Would we even need `CollectPackageReference`, `CollectPackageDownload`, etc if we had a single `CollectRestoreInputs`?

Unfortunately we'd need to keep the existing targets for backwards compatibility, but if `CollectRestoreInputs` became the new supported target, then we'd never need to introduce a new target even if we added more items.

But this is starting to feel out of scope for this PR's title.

src/NuGet.Core/NuGet.Build.Tasks/GetRestoreNuGetAuditSuppressionsTask.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        /// <summary>
+        /// Output items
+        /// </summary>
+        [Output]
+        public ITaskItem[] RestoreGraphItems { get; set; }
+
+        public override bool Execute()
+        {
+            var entries = new List<ITaskItem>();
```
zivkan: Always initialise lists with an expected capacity when you know it.  We don't generally expect duplicates, so we can use `NuGetAuditSuppressions.Length`.

Technically the same is true for the hashset, it should have a ctor overload with initial capacity. However, my memory is that the smallest hashset capacity is quite large for most of our use cases, so it's unlikely we'll  get backing array resizes for hash sets.

Additionally, you can special case `if (NuGetAuditSuprpessions.Length == 0) return;` to avoid memory allocations when we know there won't be any work to do.

src/NuGet.Core/NuGet.Build.Tasks/GetRestoreNuGetAuditSuppressionsTask.cs
```diff
@@ -0,0 +1,67 @@
...
+            var seenIds = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
zivkan: This task is called in `_GenerateProjectRestoreGraphPerFramework`, which `DependsOnTargets="_GetRestoreProjectStyle;_CollectRestoreInputs"`, and `_collectRestoreInputs` does 

```xml
<CheckForDuplicateNuGetItemsTask
      ...
      >
      <Output TaskParameter="DeduplicatedItems" ItemName="DeduplicatedNuGetAuditSuppressions" />
```

Therefore, I feel like this hashset, and the deduplication that happens in this method, is duplicated and unnecessary work.

@nkolev92 @jeffkl what do you think?

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
```diff
@@ -329,6 +330,53 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+  -->
+  <Target Name="CollectNuGetAuditSuppressions" Returns="@(NuGetAuditSuppress)">
+    <!-- NOTE for design-time builds we need to ensure that we continue on error. -->
+    <PropertyGroup>
+      <CollectNuGetAuditSuppressionsContinueOnError>$(ContinueOnError)</CollectNuGetAuditSuppressionsContinueOnError>
+      <CollectNuGetAuditSuppressionsContinueOnError Condition="'$(ContinueOnError)' == '' ">false</CollectNuGetAuditSuppressionsContinueOnError>
+    </PropertyGroup>
+
+    <CheckForDuplicateNuGetItemsTask
+      Condition="'$(DisableCheckingDuplicateNuGetItems)' != 'true' "
```
zivkan: when we added duplicate item checks for `PackageReference`, it was a new check, so this opt-out made sense in case it broke people and they needed a way out.

Since `NuGetAuditSuppress` is a new item, we don't have the same backwards compatibility concerns. Is it worthwhile to remove the condition, and always warn customers when they're doing what we consider the wrong thing?

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
```diff
@@ -1065,6 +1113,19 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+    <GetRestoreNuGetAuditSuppressionsTask
+      Condition=" '$(PackageReferenceCompatibleProjectStyle)' == 'true' "
+      ProjectUniqueName="$(MSBuildProjectFullPath)"
+      NuGetAuditSuppressions="@(NuGetAuditSuppress)"
+      TargetFrameworks="$(TargetFramework)"
+      >
+
+      <Output
+        TaskParameter="RestoreGraphItems"
+        ItemName="_RestoreGraphEntry" />
```
zivkan, line 1125: Question for @nkolev92  and @jeffkl, is there a good reason we do this as a C# task, rather than MSBuild script?  I didn't look at the implementation of other `GetRestore*Task`, just `GetRestoreNuGetAuditSuppressionsTask`, but it seems to be the C# code could be replaced by something like
```xml
<ItemGroup>
  <_RestoreGraphEntry Include="$([System.Guid]::NewGuid())">
    <Type>NuGetAuditSuppress</Type>
    <ProjectUniqueName>$(MSBuildProjectFullPath)</ProjectUniqueName>
    <Id>@(NuGetAuditSuppress->Identity)</Id>
    <TargetFrameworks>$(TargetFramework)</TargetFrameworks>
  </_RestoreGraphEntry>
</ItemGroup>
```

I can never remember when to use `@(..)` vs `%(..)`.

If this is functionally equivalent, I'm not sure what we gain by having the C# tasks for all of these items, except for deduplication, but now we do that in the `Collect*` targets anyway.

src/NuGet.Core/NuGet.Commands/GlobalSuppressions.cs
```diff
@@ -193,3 +193,5 @@
...
 [assembly: SuppressMessage("Globalization", "CA1305:Specify IFormatProvider", Justification = "Source code from FileSystem submodule. We don't intent to modify it since it's an archived project", Scope = "member", Target = "~M:Microsoft.Extensions.FileSystemGlobbing.Abstractions.DirectoryInfoWrapper.GetDirectory(System.String)~Microsoft.Extensions.FileSystemGlobbing.Abstractions.DirectoryInfoBase")]
+[assembly: SuppressMessage("Performance", "CA1827:Do not use Count() or LongCount() when Any() can be used", Justification = "<Pending>", Scope = "member", Target = "~M:NuGet.Commands.MSBuildRestoreUtility.GetRestoreAuditProperties(NuGet.Commands.IMSBuildItem,System.Collections.Generic.IList{System.String})~NuGet.ProjectModel.RestoreAuditProperties")]
+[assembly: SuppressMessage("Performance", "CA1827:Do not use Count() or LongCount() when Any() can be used", Justification = "<Pending>", Scope = "member", Target = "~M:NuGet.Commands.RestoreCommand.PerformAuditAsync(System.Collections.Generic.IEnumerable{NuGet.Commands.RestoreTargetGraph},NuGet.Common.TelemetryActivity,System.Threading.CancellationToken)~System.Threading.Tasks.Task")]
```
zivkan: Please don't suppress these analyzer warnings. This specific analyzer is telling you that the code is doing something that is low performance, and restore in NuGet's hotpath, so we need to take extra care to be high performance in restore.

Additionally, even if we were going to suppress the analyzer, don't leave the default `<Pending>` justification. The others do because when we introduced all our analyzers there were hundreds, or thousands, of violations and it was infeasible to justify each individually. But now new suppressions should justify why they're not following "best practises"

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -499,6 +500,7 @@ private async Task PerformAuditAsync(IEnumerable<RestoreTargetGraph> graphs, Tel
 
             telemetry.TelemetryEvent[AuditLevel] = (int)audit.MinSeverity;
             telemetry.TelemetryEvent[AuditMode] = AuditUtility.GetString(audit.AuditMode);
+            if (audit.SuppressedAdvisories.Count() > 0) { telemetry.TelemetryEvent[SuppressedAdvisories] = ConcatAsString(audit.SuppressedAdvisories); }
```
zivkan:
```suggestion
            telemetry.TelemetryEvent[SuppressedAdvisories] = ConcatAsString(audit.SuppressedAdvisories);
```

There's value in knowing when the count is zero vs telemetry from an older version of VS before this property/feature exists.

Also, never use LINQ when you don't need to. HashSet has a Count property, so `.Count > 0` is going to cause fewer allocations, therefore fewer garbage collections, and just generally have less CPU work to do than `Count()`.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -62,6 +63,7 @@ internal class AuditUtility
 
             MinSeverity = ParseAuditLevel();
             AuditMode = ParseAuditMode();
+            SuppressedAdvisories = new HashSet<string>(restoreAuditProperties?.SuppressedAdvisories ?? new List<string>());
```
zivkan: Restore is NuGet's hot-path, so you should avoid memory allocations when they're not needed. Make `SuppressedAdvisories` nullable, and only create the `HashSet` when there are values to be used.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -92,6 +92,7 @@ public class RestoreCommand
         private const string AuditEnabled = "Audit.Enabled";
         private const string AuditLevel = "Audit.Level";
         private const string AuditMode = "Audit.Mode";
+        private const string SuppressedAdvisories = "Audit.SuppressedAdvisories";
```
zivkan: I think the name is unclear if this is the number of suppressions the project defined, or how many were used/hit during audit.  I'm also wondering if it's worthwhile recording both.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -272,6 +274,11 @@ private static (string severityLabel, NuGetLogCode code) GetSeverityLabelAndCode
                                 continue;
                             }
 
+                            if (SuppressedAdvisories.Contains(knownVulnerability.Url.ToString()))
```
zivkan:
```suggestion
                            if (SuppressedAdvisories.Contains(knownVulnerability.Url.OriginalString))
```

Unfortunately it was a big mistake for me to set `Url` to `System.Uri` rather than `string`. But we can still avoid obvious allocations.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
```diff
@@ -295,7 +295,7 @@ public static PackageSpec GetPackageSpec(IEnumerable<IMSBuildItem> items)
                         );
                     }
                     pcRestoreMetadata.RestoreLockProperties = GetRestoreLockProperties(specItem);
-                    pcRestoreMetadata.RestoreAuditProperties = GetRestoreAuditProperties(specItem);
+                    pcRestoreMetadata.RestoreAuditProperties = GetRestoreAuditProperties(specItem, suppressionItems: new List<string>());
```
zivkan: Restore is NuGet's hot path, do don't create allocations when they're not needed. Make arguments nullable and be careful the handle nulls correctly.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
```diff
@@ -920,25 +920,34 @@ private static RestoreLockProperties GetRestoreLockProperties(IMSBuildItem specI
...
-        public static RestoreAuditProperties GetRestoreAuditProperties(IMSBuildItem specItem)
+        public static RestoreAuditProperties GetRestoreAuditProperties(IMSBuildItem specItem, IList<string> suppressionItems)
         {
             string enableAudit = specItem.GetProperty("NuGetAudit");
             string auditLevel = specItem.GetProperty("NuGetAuditLevel");
             string auditMode = specItem.GetProperty("NuGetAuditMode");
 
-            if (enableAudit != null || auditLevel != null || auditMode != null)
+            if (enableAudit != null || auditLevel != null || auditMode != null
+                || (suppressionItems != null && suppressionItems.Count() > 0))
```
zivkan:
```suggestion
                || (suppressionItems != null && suppressionItems.Count > 0))
```

Never call LINQ methods when not needed. They cause memory allocations, plus require a bunch of CPU work to calculate the value that already exists.

src/NuGet.Core/NuGet.ProjectModel/RestoreAuditProperties.cs
```diff
@@ -28,6 +29,11 @@ public class RestoreAuditProperties : IEquatable<RestoreAuditProperties>
         /// <value>direct, all</value>
         public string? AuditMode { get; set; }
 
+        /// <summary>
+        /// Gets or sets values indicating which advisories to suppress.
+        /// </summary>
+        public IList<string> SuppressedAdvisories { get; set; } = new List<string>();
```
zivkan: Restore is on NuGet's hotpath, so allocations should be avoided unless needed. Make the property nullable, and use null to signal that the project didn't define any suppressions.

src/NuGet.Core/NuGet.ProjectModel/RestoreAuditProperties.cs
```diff
@@ -127,6 +135,7 @@ internal RestoreAuditProperties Clone()
                 EnableAudit = EnableAudit,
                 AuditLevel = AuditLevel,
                 AuditMode = AuditMode,
+                SuppressedAdvisories = SuppressedAdvisories,
```
zivkan: Since you made the list a mutable data structure, this makes this line "wrong", because if someone clones the `RestoreAuditProperties`, then modifies `SuppressedAdvisories` in one of the two, the other one will be affected.

You should change `SuppressedAdvisories` to a type like `IReadOnlyList<T>` to make it safe to clone this class without needing to also clone the collection.

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.Utf8JsonStreamReader.cs
```diff
@@ -1070,6 +1071,8 @@ private static void ReadMSBuildMetadata(ref Utf8JsonStreamReader jsonReader, Pac
                     else if (jsonReader.ValueTextEquals(RestoreAuditPropertiesPropertyName))
                     {
                         string enableAudit = null, auditLevel = null, auditMode = null;
+                        var suppressedAdvisories = new List<string>();
```
zivkan: restore is on NuGet's hotpath so we shouldn't allocate unless there's a value. In this case, it's "worse", because if there is a value, the list that's allocated here is thrown away, never used.

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.Utf8JsonStreamReader.cs
```diff
@@ -1086,6 +1089,10 @@ private static void ReadMSBuildMetadata(ref Utf8JsonStreamReader jsonReader, Pac
                                 {
                                     auditMode = jsonReader.ReadNextTokenAsString();
                                 }
+                                else if (jsonReader.ValueTextEquals(AuditSuppressionsPropertyName))
+                                {
+                                    suppressedAdvisories = jsonReader.ReadNextStringOrArrayOfStringsAsReadOnlyList().ToList();
```
zivkan: I don't think we should allow a single value to be a string literal vs an array with only 1 value. I think it's fine if there's just a single value that it must always be an array.

However, if at some point we add per-package filtering for suppressions, then if this is an array of strings, then it will require a schema change.  But if `suppressedAdvisories` is an object where the url is the key and value is `null`, then in the future we should change that hardcoded `null` to something else without a breaking schema change. For example:
```json
"suppressedAdvisories": {
  "https://cve/": null
}
```

Also, the fact that the method called returns a read only list and then you need to call `ToList` should be a hint that there's something wrong with the design. But I already mentioned read only list in another comment.

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.cs
```diff
@@ -1075,9 +1075,10 @@ private static void ReadMSBuildMetadata(JsonTextReader jsonReader, PackageSpec p
 
                     case "restoreAuditProperties":
                         string enableAudit = null, auditLevel = null, auditMode = null;
+                        var suppressedAdvisories = new List<string>();
```
zivkan: As mentioned elsewhere, don't allocate unless you need to.

src/NuGet.Core/NuGet.ProjectModel/GlobalSuppressions.cs
```diff
@@ -86,3 +86,4 @@
 [assembly: SuppressMessage("Build", "CA2227:Change 'Imports' to be read-only by removing the property setter.", Justification = "<Pending>", Scope = "member", Target = "~P:NuGet.ProjectModel.TargetFrameworkInformation.Imports")]
 [assembly: SuppressMessage("Build", "CA2237:Add [Serializable] to FileFormatException as this type implements ISerializable", Justification = "<Pending>", Scope = "type", Target = "~T:NuGet.ProjectModel.FileFormatException")]
 [assembly: SuppressMessage("Usage", "CA2227:Collection properties should be read only", Justification = "<Pending>", Scope = "member", Target = "~P:NuGet.ProjectModel.LockFile.CentralTransitiveDependencyGroups")]
+[assembly: SuppressMessage("Build", "CA2227:Change 'SuppressedAdvisories' to be read-only by removing the property setter.", Justification = "<Pending>", Scope = "member", Target = "~P:NuGet.ProjectModel.RestoreAuditProperties.SuppressedAdvisories")]
```
zivkan: When we added analyzers to the repo, there were hundreds of violations, so it was infeasible to justify each one.  But new suppressions should be justified, although it's better to find a different design that doesn't violate the analyzer.

src/NuGet.Core/NuGet.ProjectModel/PackageSpecWriter.cs
```diff
@@ -235,6 +235,11 @@ private static void WriteNuGetAuditProperties(IObjectWriter writer, RestoreAudit
             SetValueIfNotNull(writer, "auditLevel", auditProperties.AuditLevel);
             SetValueIfNotNull(writer, "auditMode", auditProperties.AuditMode);
 
+            if (auditProperties.SuppressedAdvisories.Any())
```
zivkan:
```suggestion
            if (auditProperties.SuppressedAdvisories.Count > 0)
```

Never use LINQ when it's not needed. It causes allocations and has higher CPU cost than checking properties that types already provide.

src/NuGet.Core/NuGet.ProjectModel/PackageSpecWriter.cs
```diff
@@ -235,6 +235,11 @@ private static void WriteNuGetAuditProperties(IObjectWriter writer, RestoreAudit
...
+            {
+                writer.WriteNameArray("suppressedAdvisories", auditProperties.SuppressedAdvisories);
```
zivkan: As mentioned elsewhere, I'm wondering if it's worthwhile to make this an object where the advisory URL is the key, and the value is hardcoded to `null` for now, so that in the future if we add per-package filtering on suppressions, it doesn't require a breaking schema change.

src/NuGet.Core/NuGet.ProjectModel/RestoreAuditProperties.cs
```diff
@@ -90,7 +96,8 @@ public bool Equals(RestoreAuditProperties? other)
 
             return EnableAudit == other.EnableAudit &&
                 AuditLevel == other.AuditLevel &&
-                AuditMode == other.AuditMode;
+                AuditMode == other.AuditMode &&
+                SuppressedAdvisories.OrderedEquals(other.SuppressedAdvisories, advisory => advisory, StringComparer.Ordinal, EqualityComparer<string>.Default);
```
zivkan: You're calling `OrderedEquals` here in `Equals`, but in `GetHashCode` you're calling `AddUnorderedSequence`. These should be consistent.

Since we don't care what order customers define the suppressions, it makes sense to make it consistently unordered.

test/NuGet.Core.Tests/NuGet.Build.Tasks.Test/GetRestoreNuGetAuditSuppressionsTaskTests.cs
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Build.Framework;
+using Xunit;
+
+namespace NuGet.Build.Tasks.Test
+{
+    public class GetRestoreNuGetAuditSuppressionsTaskTests
```
zivkan, line 9: I know you haven't had a lot of time to write tests yet, but the product class that this test class is testing has business logic to deduplicate as well, so assuming we actually keep that (see my comments on the product code), there should be tests for that.

test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/RestoreCommandTests.cs
```diff
@@ -2873,6 +2874,7 @@ public async Task ExecuteAsync_WithSinglePackage_PopulatesCorrectTelemetry()
                 ["Audit.Enabled"] = value => value.Should().Be("enabled"),
                 ["Audit.Level"] = value => value.Should().Be(0),
                 ["Audit.Mode"] = value => value.Should().Be("Unknown"),
+                ["Audit.SuppressedAdvisories"] = value => value.Should().Be("https://cve-1"),
```
zivkan: when looking at the product code, I didn't realize it was going to send the actual advisory URLs. Personally, I don't think that we'd use the advisory URLs to make product decisions, so I think a count would be sufficient.

If we do send actual advisory URLs, we might have to make it PII, because we can't be sure that a company isn't using NuGetAudit internally for their own private packages, or hosting their own vulnerability database.

test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/Utility/AuditUtilityTests.cs
```diff
@@ -438,12 +438,46 @@ static async Task<RestoreTargetGraph> CreateGraphAsync(DependencyProvider packag
         }
     }
 
+    [Fact]
+    public async Task Check_ProjectWithSuppressions_SuppressesExpectedVulnerabilities()
+    {
+        // Arrange
+        var context = new AuditTestContext();
+        string cveUrl1 = "https://cve.test.suppressed/1";
```
zivkan: test url domains should end in `.test`, since that's s reserved TLD, and it ensures that HTTP traffic can't accidentally go out. Whereas if someone registers the TLD `.suppressed`, then depending on what the test code does, it might actually send HTTP requests to someone's web server, or at least the DNS server.

zivkan (commented):
Great job!

Since you're new to the client team, you're not aware of our performance requirements, especially on restore, so there's a bunch of unnecessary memory allocations that should be removed.

src/NuGet.Core/NuGet.Build.Tasks/GetRestoreNuGetAuditSuppressionsTask.cs
jeffkl: Duplicates certainly matter for PackageReference since you could have:
```xml
<PackageReference Include="A" Version="1.0.0" />
...
<PackageReference Include="A" Version="2.0.0" />
```
So we'd warn a user that they have specified A twice or more and they might not get the version they expect since MSBuild and NuGet have to decided which single one to use.  Would duplicate suppressions for advisories have this problem?  If not, I wouldn't bother de-duplicating them since it adds overhead.  But if a user could accidentally specify more than one and potentially not get the desired behavior, we'll have to warn them.

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
nkolev92: > Unfortunately we'd need to keep the existing targets for backwards compatibility, but if CollectRestoreInputs became the new supported target, then we'd never need to introduce a new target even if we added more items.

We need to ask the CPS/Project-system about this.

src/NuGet.Core/NuGet.Build.Tasks/GetRestoreNuGetAuditSuppressionsTask.cs
nkolev92: I think what @zivkan is saying that by the time we get to [GetRestoreNuGetAuditSuppressionsTask.cs](https://github.com/NuGet/NuGet.Client/pull/5679/files/5e35b7e4c5577af50cf50e108043b8010a12f04d#diff-8bebcf007d7d681e507ab067017c52c1fbeffb425cff02a79d105271b9da9a97), things will be deduplicated (warnings and all) already, so we shouldn't need to check it twice..

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
nkolev92: I think the SDK has a switch on their end, not sure if it captures all deduplication.

IMO, I think it's ok not have a condition for the new item.

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
nkolev92: I think for the most part, it had been easier to write in C# code for people :) 

For some of the items, I think it's an overkill to have a task.

As long as this works, I'm ok with replacing it.

src/NuGet.Core/NuGet.Commands/GlobalSuppressions.cs
nkolev92: I agree. We should resolve all performance warnings in new code.

src/NuGet.Core/NuGet.Commands/PublicAPI/netcoreapp5.0/PublicAPI.Shipped.txt
```diff
@@ -947,7 +947,6 @@ override NuGet.Commands.WarningPropertiesCollection.GetHashCode() -> int
 ~static NuGet.Commands.MSBuildRestoreUtility.GetDependencySpec(System.Collections.Generic.IEnumerable<NuGet.Commands.IMSBuildItem> items) -> NuGet.ProjectModel.DependencyGraphSpec
 ~static NuGet.Commands.MSBuildRestoreUtility.GetMessageForUnsupportedProject(string path) -> NuGet.Common.RestoreLogMessage
 ~static NuGet.Commands.MSBuildRestoreUtility.GetPackageSpec(System.Collections.Generic.IEnumerable<NuGet.Commands.IMSBuildItem> items) -> NuGet.ProjectModel.PackageSpec
-~static NuGet.Commands.MSBuildRestoreUtility.GetRestoreAuditProperties(NuGet.Commands.IMSBuildItem specItem) -> NuGet.ProjectModel.RestoreAuditProperties
```
nkolev92, line 953: This an API breaking change. 
Sounds like no one is concerned about it, since this is arguably something that ideally wasn't public, but has to be, so I think it's ok to make the breaking change.

However, we should make sure we disclose the API breaking change in our release notes. 
You can do that by creating a specific issue with a label `Category:BreakingChange` that talks about the breaking change and is fixed by this PR (alongside the other one).

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
nkolev92: Should the original type of the advisories be hashset? 

For example, package versions is a dictionary.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
nkolev92: I'd just add GetAuditSuppressions(items)) here is well. 

We'll get PC to support this soon.

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
zivkan: If CPS/project-system have hardcoded the list of Collect* targets for DTBs, then I think it's even more reason to have a single CollectRestoreInputs, so they don't need to be modified every time we add a new target.  But yes, it sounds like we need to talk to them (again) about it.

src/NuGet.Core/NuGet.Commands/PublicAPI/netcoreapp5.0/PublicAPI.Shipped.txt
zivkan: From memory, it's a public API so that both static graph restore, as well as "normal" CLI restore, can share the same code. But the method name, the method parameter type and return type are so specific, I just don't see it being useful to anyone else.

I wish we had some way to mark an API as "public because NuGet needs it internally, but we're going to make no effort to avoid breaking changes". `InternalsVisibleTo` feels like a bad solution. But yes, I noticed this when reviewing the PR, but because of the above justification, I just don't see it impacting anyone.

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
advay26: I've removed this check now

src/NuGet.Core/NuGet.Commands/GlobalSuppressions.cs
advay26: Yeah, that was a miss on my part. I've fixed it now. Thanks!

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
advay26: I've made both `RestoreAuditProperties` and `AuditUtility` use HashSets now, both nullable.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
advay26: I've updated this so we're now recording both, the number of defined NuGetAuditSuppress items (`Audit.SuppressedAdvisories.Defined.Count`), as well as the number of advisories that were actually skipped during the audit (`Audit.SuppressedAdvisories.Used.Count`).

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
advay26: I'm now passing `GetAuditSuppressions(items)` for PC projects too, but I've also updated the `GetRestoreAuditProperties()` method to handle null parameters correctly

src/NuGet.Core/NuGet.ProjectModel/RestoreAuditProperties.cs
advay26: We're using a nullable HashSet<string> here now

src/NuGet.Core/NuGet.ProjectModel/RestoreAuditProperties.cs
advay26: We changed to using HashSets instead of Lists, and while an `IReadOnlySet` exists, it is only available on .NET 5+ frameworks, so we'll need to use a `HashSet` instead. 

I've changed the `Clone()` method so it should work correctly now, but let me know if it looks wrong.

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.Utf8JsonStreamReader.cs
advay26: Thanks for pointing that out! I've fixed this now. Let me know if it's still not as it should be.

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.Utf8JsonStreamReader.cs
advay26: Good point! I've incorporated your suggestion, and updated tests to reflect that as well.

src/NuGet.Core/NuGet.ProjectModel/GlobalSuppressions.cs
advay26: I looked into `IReadOnlySet`, but it's only available in .NET 5+, so we'll need to use HashSets instead. I've added a justification here now.

src/NuGet.Core/NuGet.ProjectModel/PackageSpecWriter.cs
advay26: I've updated the code with your suggestion, thanks!

src/NuGet.Core/NuGet.ProjectModel/RestoreAuditProperties.cs
advay26: `OrderedEquals` first 'orders' the unordered sequence, and then checks the two sequences for equality.

Now that we're using HashSets, I need to use `SetEqualsWithNullCheck` anyway, so I've made that change.

test/NuGet.Core.Tests/NuGet.Build.Tasks.Test/GetRestoreNuGetAuditSuppressionsTaskTests.cs
advay26: I've added a duplicate test case in for now, since I haven't made any changes to the original task. If we decide to get rid of the task, or remove the duplication part of the task, I'll update/remove the tests accordingly.

test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/RestoreCommandTests.cs
advay26: Makes sense. We'll just be recording the counts now.

src/NuGet.Core/NuGet.Commands/PublicAPI/netcoreapp5.0/PublicAPI.Shipped.txt
advay26: Sounds good. I've created an issue here: https://github.com/NuGet/Home/issues/13313, and referenced it in this PR as you mentioned. Let me know if I need to change anything else here

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
advay26: It sounds like this is something we'll look to do later once we've had further discussions, so I'll leave these targets as they are right now

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
advay26: It seems like the MSBuild code @zivkan proposed is the simplest way to wrap these suppression items, but I'll look to make these changes in a later PR (discussed offline with Andy) to keep the scope of this PR more focused.

I'll keep this Task as it is for now, and I've created a tracking item to follow up on it later: https://github.com/NuGet/Client.Engineering/issues/2731. If people prefer that I just make the change with this PR itself, let me know and I can make the changes.

src/NuGet.Core/NuGet.Build.Tasks/GetRestoreNuGetAuditSuppressionsTask.cs
advay26: Done.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
advay26: I've fixed this now, thanks!

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
advay26: Fixed

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
advay26: Fixed.

src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.cs
advay26: I believe I've fixed this now, but let me know if it's still not right, and I'll make the changes.

src/NuGet.Core/NuGet.ProjectModel/PackageSpecWriter.cs
advay26: Fixed

test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/Utility/AuditUtilityTests.cs
advay26: Fixed this now. Thanks for explaining!

src/NuGet.Core/NuGet.Build.Tasks/GetRestoreNuGetAuditSuppressionsTask.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// </summary>
+        [Output]
+        public ITaskItem[] RestoreGraphItems { get; set; }
+
+        public override bool Execute()
+        {
+            if (NuGetAuditSuppressions.Length == 0) return true;
+
+            var entries = new List<ITaskItem>(NuGetAuditSuppressions.Length);
+            var seenIds = new HashSet<string>(NuGetAuditSuppressions.Length, StringComparer.OrdinalIgnoreCase);
```
zivkan:
```suggestion
            var seenIds = new HashSet<string>(NuGetAuditSuppressions.Length, StringComparer.Ordinal);
```

I don't think it's a safe assumption that URLs are always case insensitive. While GitHub Advisories Database is, a generic web server can serve static files from a filesystem. And some filesystems are case sensitive. Therefore, if the filesystem contains both `index.html` and `Index.html`, which file should be served if the request comes in for `INDEX.HTML`?

zivkan (commented):
Looks great to me!

Just one thing where my gut reaction makes me feel it's a correctness issue, although testing a bunch of popular websites, and all the CVE websites I can find, they're all case insensitive, so maybe my suggestion doesn't actually matter.

src/NuGet.Core/NuGet.Build.Tasks/GetRestoreNuGetAuditSuppressionsTask.cs
advay26: That's fair. I've made this change now, and added a test for it as well

zivkan (dismissed):

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -272,6 +275,12 @@ private static (string severityLabel, NuGetLogCode code) GetSeverityLabelAndCode
                                 continue;
                             }
 
+                            if (SuppressedAdvisories?.Contains(knownVulnerability.Url.OriginalString) == true)
+                            {
+                                SuppressedAdvisoriesUsedCount++;
```
nkolev92: Do you expect to compare used vs defined at all?


Because the way this is calculated right now, it doesn't handle multi targeting, or scenarios where 1 advisory refers to multiple packages etc. 

I think it'd be  helpful defining what you want this number to be

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
```diff
@@ -920,25 +920,34 @@ private static RestoreLockProperties GetRestoreLockProperties(IMSBuildItem specI
                 IsPropertyTrue(specItem, "RestoreLockedMode"));
         }
 
-        public static RestoreAuditProperties GetRestoreAuditProperties(IMSBuildItem specItem)
+        public static RestoreAuditProperties GetRestoreAuditProperties(IMSBuildItem specItem, IList<string> suppressionItems)
```
nkolev92: Should this type take a hash set if we're creating a hashset?

nkolev92:
Removing the label since the target branch is the feature branch anyways

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
advay26: Instead of the single 'Used' metric, we're now recording two separate metrics here: the total number of warnings we suppressed during audit: `"Audit.SuppressedAdvisories.TotalWarningsSuppressed.Count"`, and the number of distinct advisories we suppressed: `"Audit.SuppressedAdvisories.DistinctAdvisoriesSuppressed.Count"`

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
advay26: Made the change so this is using HashSets too

nkolev92 (approved):

PR #5680, opened 2024-03-11
Make NuGetAuditMode default to all with .NET 9 SDK

### LINKED ISSUE #13293: NuGetAudit should check transitive packages by default when the .NET 9 SDK is installed
### NuGet Product(s) Affected

NuGet.exe, Visual Studio Package Management UI, Visual Studio Package Manager Console, MSBuild.exe, dotnet.exe

### Current Behavior

Currently `NuGetAuditMode` defaults to `direct`, which means that NuGetAudit will only create warnings when direct package references have a known vulnerability.

### Desired Behavior

`NuGetAuditMode` should default to `all`, so both direct and transitive package references should be scanned for known vulnerabilities.

Note this will create additional warnings, which were not present before. When a project treats warnings as errors, it will cause breaks. We know this is impactful to everyone, but since this is a security feature we believe it is an overall good for the ecosystem.

Our [docs on NugetAudit](https://learn.microsoft.com/en-us/nuget/concepts/auditing-packages) explain multiple ways to avoid NuGetAudit breaking builds, such as `WarningsNotAsErrors` or disabling NuGetAudit.

### Additional Context

Due to how NuGet flows from the NuGet.Client repo into both Visual Studio (incuding MSBuild) and the .NET SDK:

* This will affect SDK style projects in Visual Studio when the .NET 9 SDK is installed, even if your project does not target `net9.0`
* Non-SDK style projects will not be affected (until the .NET 9 SDK goes GA)

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13293

Regression? no

## Description

The .NET SDK introduced the concept of SdkAnalysisLevel, to be similar to how the C# compiler has analysis levels, allowing customers to install newer tooling without getting breaking changes while their project continues to use the older analysis level.  So, we'll use SdkAnalysisLevel `9.0.100` to detect anyone using the .NET 9 or higher SDK, and default to NuGetAuditMode all. Everything else (projects using .NET 8 SDK or below, and non-SDK style projects), will continue to use the `direct` default for now.

We tried a similar thing when introducing NuGetAudit initially, and we got feedback from customers using nuget.exe with Visual Studio 2019 still that it broke. So, in addition, NuGet.exe needs to detect when it's using a version of MSBuild that doesn't support version comparison intrinsic functions, so that our MSBuild script can use MSBuild's lazy evaluation to avoid using the function on MSBuild versions that don't support it.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception: Impractical to have a bunch of different MSBuild versions and .NET SDK versions
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A: There's going to be a LOT of NuGetAudit documentation and blog posts and announcements needed, but I'm not creating a tracking issue just for this.

### DIFF
--- a/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
+++ b/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
@@ -73,6 +73,10 @@ Copyright (c) .NET Foundation. All rights reserved.
     <!-- Report on low severity vulnerabilities, and higher. Allowed values are: low, moderate, high, critical -->
     <NuGetAuditLevel Condition=" '$(NuGetAuditLevel)' == '' ">low</NuGetAuditLevel>
     <!-- Report known vulnerabilities on direct dependencies only. Allowed values are: direct, all -->
+    <NuGetAuditMode Condition="'$(NuGetAudit)' == ''
+                    AND '$(NuGetExeSkipSDKCheck)' != 'true'
+                    AND $([MSBuild]::VersionGreaterThanOrEquals($([MSBuild]::ValueOrDefault('$(NETCoreSdkVersion)', '0.0')), '9.0'))"
+                    >all</NuGetAuditMode>
     <NuGetAuditMode Condition=" '$(NuGetAuditMode)' == '' ">direct</NuGetAuditMode>
   </PropertyGroup>
 
--- a/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
+++ b/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
@@ -73,9 +73,9 @@ Copyright (c) .NET Foundation. All rights reserved.
     <!-- Report on low severity vulnerabilities, and higher. Allowed values are: low, moderate, high, critical -->
     <NuGetAuditLevel Condition=" '$(NuGetAuditLevel)' == '' ">low</NuGetAuditLevel>
     <!-- Report known vulnerabilities on direct dependencies only. Allowed values are: direct, all -->
-    <NuGetAuditMode Condition="'$(NuGetAudit)' == ''
-                    AND '$(NuGetExeSkipSDKCheck)' != 'true'
-                    AND $([MSBuild]::VersionGreaterThanOrEquals($([MSBuild]::ValueOrDefault('$(NETCoreSdkVersion)', '0.0')), '9.0'))"
+    <NuGetAuditMode Condition="'$(NuGetAuditMode)' == ''
+                    AND '$(NuGetExeSkipSdkAnalysisLevelCheck)' != 'true'
+                    AND $([MSBuild]::VersionGreaterThanOrEquals($([MSBuild]::ValueOrDefault('$(NETCoreSdkVersion)', '0.0')), '9.0.100'))"
                     >all</NuGetAuditMode>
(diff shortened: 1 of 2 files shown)


jebriede (dismissed):

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
```diff
@@ -73,6 +73,10 @@ Copyright (c) .NET Foundation. All rights reserved.
     <!-- Report on low severity vulnerabilities, and higher. Allowed values are: low, moderate, high, critical -->
     <NuGetAuditLevel Condition=" '$(NuGetAuditLevel)' == '' ">low</NuGetAuditLevel>
     <!-- Report known vulnerabilities on direct dependencies only. Allowed values are: direct, all -->
+    <NuGetAuditMode Condition="'$(NuGetAuditMode)' == ''
+                    AND '$(NuGetExeSkipSdkAnalysisLevelCheck)' != 'true'
+                    AND $([MSBuild]::VersionGreaterThanOrEquals($([MSBuild]::ValueOrDefault('$(NETCoreSdkVersion)', '0.0')), '9.0.100'))"
```
kartheekp-ms: Are we supposed to use the `SdkAnalysisLevel` property instead of `NETCoreSdkVersion`? PR description has `So, we'll use SdkAnalysisLevel 9.0.100 to detect anyone using the .NET 9 or higher SDK, and default to NuGetAuditMode all. Everything else (projects using .NET 8 SDK or below, and non-SDK style projects), will continue to use the direct default for now.` information. Hence, I am double-checking.

zivkan: yes. If a customer is using analysis level 1 by explicitly setting it in their Directory.Build.props, and they don't want a new .NET SDK to add or change warnings, then we need to use their requested analysis level, not the SDK version. This is how it works with C# analysis levels, where you want your project targetnet `net8.0` to compile the same regardless if you're using the .NET 8 SDK or .NET 9 SDK.

kartheekp-ms: Thanks, Andy, for the explanation. Does this mean that we should be using `SdkAnalysisLevel` property instead of `NETCoreSdkVersion` property?

```suggestion
                    AND $([MSBuild]::VersionGreaterThanOrEquals($([MSBuild]::ValueOrDefault('$(SdkAnalysisLevel)', '0.0')), '9.0.100'))"
```

zivkan: Oops 🤦

I didn't notice when I was reviving this PR, or from your first comment. Sorry!   Good catch, I've just pushed an update that fixes this oversight.

jebriede (approved):

zivkan:
@dotnet-policy-bot /rerun

zivkan:
@dnfclas /rerun

zivkan:
@dnfclas /help

kartheekp-ms (approved):

zivkan:
@dotnet-policy-bot /rerun

zivkan:
@dnfclas /rerun

zivkan:
@dotnet-policy-bot rerun

zivkan:
@dotnet-policy-service rerun

PR #5681, opened 2024-03-11
Improve package spec hashing - share package spec hashes during the hashing operation

### LINKED ISSUE #10068: Improve the no-op hashing performance
Currently the no-op hash is calculated by creating a json string of the dependency graph spec for said project. 
This involves a lot of duplicate + using jsonwriter for hash is not ideal.

This issue tracks the investigation for improving that hashing performance.

I discussed with @cristinamanum and @zivkan 

The suggestion is move to a normal hashing function instead of using the package specwriter. 
The challenge is that we don't know if the package spec is mutated during restore.

Given that here are the options. 

* Consider setting a Locked property, to disable mutation after a certain point.
* Use an intermediate class that can generate the hashcode on demand (this will require coordination to ensure the hash generation is synchronized).
* Restore Runner before invoking the restore command can prepopulate the hash code for all the package specs. 

This is expected to be a long investigation, because while perf is one goal, we need to ensure this doesn't get regressed in the future. 

I did prototype these changes and it showed a significant improvement for large slns with deep project graphs, up to 2-4x faster hash generation. The hash generation right now is about ~30% of the time spent in the restore part (not dg spec generation). 

So cmd could see numbers like 7s to 6.8s, but VS will relatively better perf, example, 1.1s -> 900ms.

nkolev92:
## Bug

Fixes: https://github.com/NuGet/Home/issues/10068

## Description

* In a large restore operation, a single package spec will be used in multiple restores
* The hash of a specific project does not change.
* We'll reduce the frequency a hash of a project is calculated by caching the value and simply looking it up from a dictionary.
* In addition to that, there's a discrepancy between the type exposed for the list of projects and the internally used type. Currently we create a list anytime the property is accessed. To reduce that we'll use a simple cache. I attempted to write an iterator over a sorted dictionary, but implementing `[ ]` accessor was a bunch of work, so the simple list cache is probably a good enough balance.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Core/NuGet.ProjectModel/DependencyGraphSpec.cs
```diff
@@ -340,14 +407,49 @@ private void Write(RuntimeModel.IObjectWriter writer, bool hashing, Action<Packa
...
+            writer.WriteObjectEnd();
+        }
+
+        private void WriteProject(IObjectWriter writer, bool hashing, Action<PackageSpec, IObjectWriter, bool, IEnvironmentVariableReader> writeAction, PackageSpec project, Dictionary<string, string>? projectNameToHashCode)
+        {
+            if (hashing && projectNameToHashCode != null)
+            {
+                string? projectHash = null;
+
+                lock (projectNameToHashCode)
```
jeffkl, line 430: For large restores, I wonder if it would be better to use a per-project lock instead of locking the entire dictionary like we have in TaskResultCache?

nkolev92: A per project lock would involve some sort of a lock lookup which itself would have the same issue. 
Looking up is going to be really fast, so I don't expect much contention here.

jeffkl (dismissed):

src/NuGet.Core/NuGet.ProjectModel/DependencyGraphSpec.cs
```diff
@@ -43,34 +54,64 @@ public DependencyGraphSpec(bool isReadOnly)
...
         /// Projects to restore.
         /// </summary>
-        public IReadOnlyList<string> Restore => _restore.ToList();
+        public IReadOnlyList<string> Restore
+        {
+            get
+            {
+                if (_restoreList == null || _refreshRestoreList)
+                {
+                    _restoreList = _restore.ToList();
```
zivkan, line 70: Does this property get called once per restore, or can it get called multiple times?

Specifically what I'm thinking about is if this method is going to get called multiple times concurrently when `_restoreList == null || _refreshRestoreList`.

This method is idempotent, so this implementation won't cause correctness problems. But `.ToList()` will be called my multiple threads concurrently, causing the same list to be allocated multiple times. If the backing store doesn't implement `ICollection`, then `.ToList()` is going to cause multiple array allocations for the list's backing store, so multiple threads doing so concurrently will have a particularly bad allocation effect.

Basicly, is it worth wrapping the inside of the `if` statement in a `lock`, to reduce memory allocations?

src/NuGet.Core/NuGet.ProjectModel/DependencyGraphSpec.cs
```diff
@@ -43,34 +54,64 @@ public DependencyGraphSpec(bool isReadOnly)
...
         /// All project specs.
         /// </summary>
-        public IReadOnlyList<PackageSpec> Projects => _projects.Values.ToList();
+        public IReadOnlyList<PackageSpec> Projects
+        {
+            get
+            {
+                if (_restoreProjectList == null || _refreshProjectsList)
+                {
+                    _restoreProjectList = _projects.Values.ToList();
```
zivkan, line 89: Same question here as I had for the `Restore` property.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
```diff
@@ -86,6 +87,7 @@ private IReadOnlyList<RestoreSummaryRequest> GetRequestsFromItems(RestoreArgs re
                 // Limiting to processor count reduces task context switching which is better
                 MaxDegreeOfParallelism = Environment.ProcessorCount
             };
+            var dictionary = new Dictionary<string, string>(PathUtility.GetStringComparerBasedOnOS());
```
zivkan: I think this variable's name can be improved. I can see from its type that it's a dictionary, but currently the variable name doesn't give a hint what it's a dictionary of, or how it's used.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
```diff
@@ -94,6 +96,7 @@ private IReadOnlyList<RestoreSummaryRequest> GetRequestsFromItems(RestoreArgs re
                 {
                     IReadOnlyList<PackageSpec> closure = dgFile.GetClosure(projectNameToRestore);
                     DependencyGraphSpec projectDependencyGraphSpec = dgFile.CreateFromClosure(projectNameToRestore, closure);
+                    projectDependencyGraphSpec.SetProjectNameToHashCode(dictionary);
```
zivkan: public API design feedback: is there a reason to pass the dictionary in a method, rather than the `projectDependencyGraphSpec` constructor, and therefore through the `CreateFromClosure` method?

Are there any restore codepaths where the `DependencyGraphSpec` will not, or should not, have one of these caches?

src/NuGet.Core/NuGet.ProjectModel/DependencyGraphSpec.cs
```diff
@@ -27,6 +30,14 @@ public class DependencyGraphSpec
         private const int Version = 1;
 
         private readonly bool _isReadOnly;
+        // Internal for testing purposes
+        internal Dictionary<string, string>? _projectNameToHashCode;
+
+        public void SetProjectNameToHashCode(Dictionary<string, string> projectNameToHashCode)
```
zivkan:
```suggestion
        public void SetProjectNameToHashCodeCache(Dictionary<string, string> projectNameToHashCode)
```

src/NuGet.Core/NuGet.ProjectModel/DependencyGraphSpec.cs
```diff
@@ -27,6 +30,14 @@ public class DependencyGraphSpec
         private const int Version = 1;
 
         private readonly bool _isReadOnly;
+        // Internal for testing purposes
+        internal Dictionary<string, string>? _projectNameToHashCode;
```
zivkan:
```suggestion
        internal Dictionary<string, string>? _projectNameToHashCodeCache;
```

Or Lookup?  The name feels confusing to me at the moment. The name suggests it should be a single value, not a collection

src/NuGet.Core/NuGet.ProjectModel/DependencyGraphSpec.cs
```diff
@@ -340,14 +407,49 @@ private void Write(RuntimeModel.IObjectWriter writer, bool hashing, Action<Packa
...
+                {
+                    using IHashFunction hashFunc = UseLegacyHashFunction ? new Sha512HashFunction() : new FnvHash64Function();
+                    using var projectWriter = new HashObjectWriter(hashFunc);
+                    writeAction.Invoke(project, projectWriter, hashing, EnvironmentVariableWrapper.Instance);
+                    projectHash = projectWriter.GetHash();
+                }
+
+                lock (projectNameToHashCode)
+                {
+                    if (!projectNameToHashCode.ContainsKey(project.RestoreMetadata.ProjectUniqueName))
```
zivkan: can this lock be moved inside the `if (projectHash == null)` block, so it doesn't need to run on cache hits?

src/NuGet.Core/NuGet.ProjectModel/DependencyGraphSpec.cs
zivkan: The lock is held for a very short time, because all it's doing is a lookup, but if multiple threads concurrently have a cache miss, then they're all going to take hashes of the object at the same time.

Is hashing zero alloc, or low alloc?  If not, blocking the thread might be less impactful than having multiple threads do duplicate work with allocations that might cause GC, since GC will block all threads.

nkolev92: I don't want hashing 1 project to stop restore from hashing other projects, so I'd need per project locks. 
I think that would theoretically be lower allocations, but with the newer hashing algorithm they're already significantly lower, so the biggest benefit in the testing is time.

I can try a more complex implementation if you feel too strongly, but I think this is one of those where I don't perfect to be the enemy of really really good.

src/NuGet.Core/NuGet.ProjectModel/DependencyGraphSpec.cs
nkolev92: It gets called multiple times in restore, but it's the population is not concurrent (each project has it's own dg spec, so each project enumerates early). 

I haven't seen it called concurrently, it's just called a bunch of times during restore cause we didn't know any better.

src/NuGet.Core/NuGet.ProjectModel/DependencyGraphSpec.cs
nkolev92: This isn't used during the core of restore, but outside. 
All accesses are sequential

src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
nkolev92: This makes the caching obvious and scopes it down to restore only and avoids unexpected behaviors. 

I thought about adding it in CreateFromClosure and it'd avoid the public method, but it'll lead to potential fragility. 
I don't feel too strongly about that though, as fragility is the name of the game in this PR as we basically assume no one will ever change the PackageSpec within the core of restore.

src/NuGet.Core/NuGet.ProjectModel/DependencyGraphSpec.cs
nkolev92: Yeah, the 2nd check would help.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RequestFactory/DependencyGraphSpecRequestProvider.cs
nkolev92: You know, I think I prefer the new public API-less approach, even if CreateFromClosure becomes a bit funky. 

I'll document it and only do when the dg spec is read only.

zivkan (dismissed):

nkolev92:
Unfortunately the change as is will break scenarios like no-op after dotnet add package, or no-op after package installation. 

I'm trying to fix that but it's revealing a bunch of other issues. 

I'll have to deprioritize this to get other things in P3, before looking into this change again.

PR #5683, opened 2024-03-12
Implement a dynamically wrapping table for `dotnet package search`

### LINKED ISSUE #13162: dotnet package search --verbosity detailed output table too wide
### NuGet Product(s) Affected

dotnet.exe

### Current Behavior

the search command outputs a table that is as wide as its content

### Desired Behavior

table format should wrap instead of create a table that's the width of the result. 

### Additional Context

_No response_

Nigusu-Allehu:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13162

## Description

This PR changes the table implementation previously used by dotnet package search.
- This takes the console window width into consideration
- Based on the window width a wrapping table is rendered.
- In addition, after each row a line separator is added.
### Previously `dotnet package search microsoft --verbosity detailed` would look like this
image
### Now `dotnet package search microsoft --verbosity detailed` looks like the following
image

#### Step 1: Determine Total Window Width
- **Description**: Find out the total width available for rendering the table. This width is measured in characters.
- **Example**: Suppose our CLI window is 100 characters wide.

#### Step 2: Calculate Maximum Column Width
- **Description**: Divide the total window width by the number of columns in the table to get the maximum possible width for each column.
- **Example**: For a table with 5 columns, `_maxColumnWidth = Total Window Width / Number of Columns = 100 / 5 = 20`. So, the maximum width available for each column is 20 characters.

#### Step 3: Establish Maximum Column Width Constraint
- **Description**: The calculated maximum column width represents the upper limit for the width of any column, ensuring that the table fits within the available window width.

#### Step 4: Determine Required Space for Each Column
- **Description**: Calculate the minimum necessary width for each column by finding the longest piece of content (in terms of character length) that must fit within a single row of that column.
- **Example**: If the longest content in the first column is "column 1" (8 characters), for the second column is "column 2 column 2 column 2" (26 characters), and so on, these values represent the minimum width required for each column to display its content without breaking onto a new line.

#### Step 5: Calculate Actual Column Widths
- **Description**: For each column, compare the maximum column width from step 2 with the column's required width from step 4. The actual width of a column is the smaller of these two values. This ensures that the column is as wide as it needs to be to fit its content, without exceeding the maximum width limit.
- **Example**: If the maximum column width is 20 characters, but a column only needs 8 characters to display its longest piece of content, the width of that column will be set to 8 characters. However if the column needs 26 characters, set the column width to 20, as it is beyond the _maxColumnWidth value.

This algorithm ensures that each column is allocated a width that balances between accommodating its longest content and adhering to the total available window width, resulting in a neatly rendered table in the CLI.
## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/WrappingTable.cs
```diff
@@ -0,0 +1,235 @@
...
+namespace NuGet.CommandLine.XPlat
+{
+    internal class Column
+    {
+        public string Header { get; set; }
+        public int Width { get; set; }
+        public bool Highlight { get; set; }
+    }
+
+    internal class WrappingTable
```
nkolev92: you probably don't need the table to be of a different type. 
We only have 1 table, so keeping the type name would minimize the changes in other files.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/TableFormat/DetailedTableFormatStrategy.cs
```diff
@@ -8,9 +8,9 @@ internal class DetailedTableFormatStrategy : ITableFormatStrategy
         private readonly string[] _detailedVerbosityTableHeader = { "Package ID", "Latest Version", "Owners", "Total Downloads", "Vulnerable", "Deprecation", "Project URL", "Description" };
         private readonly int[] _detailedColumnsToHighlight = { 0, 2, 6, 7 };
 
-        public Table CreateTable()
```
nkolev92: Similar feedback as in the WrappingTable class. 
If you keep the type Table, then none of these changes are needed.

nkolev92 (commented):
One suggestion is keeping the old type names since we have only 1 table. 

it'd allow the reviewer to more easily differentiate between: 
- new changes
- old behavior

Right now it's more challenging to review it because I have to approach the wrapping table as a completely new implementation rather than an improvement of an existing one.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageSearch/Table.cs
```diff
@@ -4,119 +4,232 @@
...
-            if (row.Length != _columnWidths.Length)
+            if (row.Length != _columns.Count)
             {
                 throw new InvalidOperationException("Row column count does not match header column count.");
             }
 
             for (int i = 0; i < row.Length; i++)
             {
-                _columnWidths[i] = Math.Max(_columnWidths[i], row[i]?.Length ?? 0);
+                _columns[i].Width = Math.Min(_maxColumnWidth, Math.Max(_columns[i].Width, row[i]?.Length ?? 0));
```
nkolev92, line 81: Try to understand the algorithm here, the order of columns would guide how much space is taken correct?

Nigusu-Allehu: just updated the description to contain the algorithm. Please let me know if you have any questions

nkolev92 (approved):

nkolev92:
Thanks for the excellent description @Nigusu-Allehu. 

Nicely done.

PR #5702, opened 2024-03-21
Deprecate http usage: `source` commands, promote warning to error for http sources

Nigusu-Allehu:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13333

## Description

- promotes http warnings to errors in the following commands
     - source add
     - source update
     - source enable
- removes http warning in the following commands
     - source list : I believe the list command should list out all the available sources and allow users to observe what they have rather than showing an error and blocking them. Instead warn them about the HTTP sources
    
## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
```diff
@@ -63,12 +62,12 @@ public static void Run(AddSourceArgs args, Func<ILogger> getLogger)
 
             var newPackageSource = new Configuration.PackageSource(args.Source, args.Name);
 
-            if (newPackageSource.IsHttp && !newPackageSource.IsHttps && !newPackageSource.AllowInsecureConnections)
+            if (newPackageSource.IsHttp && !newPackageSource.IsHttps)
```
nkolev92: When we're adding, can the error message be more empowering and tell the user what they need to pass as an extra option?

This command is special enough where the generic error message can be improved.

idk if that conflicts with the other motivation behind deprecating http, so curious what others think

Nigusu-Allehu: Currently, the error message they get is `You are running the 'source add' operation with an 'HTTP' source: {1}. NuGet requires HTTPS sources. Please refer to https://aka.ms/nuget-https-everywhere.` I removed the information about editing the config file as that would not make sense when trying to add a new source. 

From my understanding, I do not think we provide users with the option to add an http source using this command. I was considering of adding a new option that would allow users to add an http source. May be `-allowInsecureConnections`, similar to our config settings.

nkolev92: Let's have a discussion about this, but I think the new option might be necessary.

As this is, we'd effectively block people from adding http sources period, with no automation based workaround available.

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
```diff
@@ -64,12 +63,12 @@ public static void Run(AddSourceArgs args, Func<ILogger> getLogger)
             var newPackageSource = new Configuration.PackageSource(args.Source, args.Name);
             newPackageSource.AllowInsecureConnections = args.AllowInsecureConnections;
 
-            if (newPackageSource.IsHttp && !newPackageSource.IsHttps && !newPackageSource.AllowInsecureConnections)
+            if (newPackageSource.IsHttp && !newPackageSource.IsHttps && !args.AllowInsecureConnections)
```
nkolev92: I guess you have the option to check with the package source since that's what gets persisted. You're already assigning it anyways: args.AllowInsecureConnections. 

Not sure it matters too much. :D

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
```diff
@@ -17,6 +17,8 @@ namespace Dotnet.Integration.Test
     public class DotnetSourcesTests
     {
         private readonly DotnetIntegrationTestFixture _fixture;
+        string _httpErrorSingleShort = "You are running the '{0}' operation with an 'HTTP' source: {1}. NuGet requires HTTPS sources. Please refer to https://aka.ms/nuget-https-everywhere.";
```
nkolev92: Might've missed this in other places and previous PRs, but I think you should use the resource not the string to allow the tests to work on other locales.

nkolev92 (commented):
Some test feedback. 
Rest looks great!
Nicely done.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
Nigusu-Allehu: It's because the Nuget.Commands locals(internal) are not accessible from here.

nkolev92: I see, can you add an internals visible to?
Maybe just match https://aka.ms/nuget-https-everywhere?
Check exit code?

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
```diff
@@ -66,10 +65,10 @@ public static void Run(AddSourceArgs args, Func<ILogger> getLogger)
 
             if (newPackageSource.IsHttp && !newPackageSource.IsHttps && !newPackageSource.AllowInsecureConnections)
             {
-                getLogger().LogWarning(
+                throw new ArgumentException(
                     string.Format(CultureInfo.CurrentCulture,
-                        Strings.Warning_HttpServerUsage,
-                        "add source",
+                        Strings.Error_HttpSource_Single_Short,
+                        "source add",
```
kartheekp-ms: The command name in docs is [dotnet nuget add source](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-nuget-add-source), hence I assume `add source` is correct. Is there any reason why we changed the string from `add source` to `source add`?

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
```diff
@@ -162,8 +161,6 @@ public static void Run(ListSourceArgs args, Func<ILogger> getLogger)
                                 source.IsEnabled ? string.Format(CultureInfo.CurrentCulture, Strings.SourcesCommandEnabled) : string.Format(CultureInfo.CurrentCulture, Strings.SourcesCommandDisabled)));
                             getLogger().LogMinimal(string.Format(CultureInfo.CurrentCulture, "{0}{1}", sourcePadding, source.Source));
                         }
-
-                        WarnForHttpSources(sourcesList, getLogger);
```
kartheekp-ms: I am actually thinking if it makes sense for us to still display warnings for HTTP sources in `list source` command.

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
```diff
@@ -286,10 +245,10 @@ public static void Run(UpdateSourceArgs args, Func<ILogger> getLogger)
                 existingSource = new Configuration.PackageSource(args.Source, existingSource.Name);
                 existingSource.AllowInsecureConnections = args.AllowInsecureConnections;
 
-                // If the existing source is not http, warn the user
-                if (existingSource.IsHttp && !existingSource.IsHttps && !existingSource.AllowInsecureConnections)
+                // If the new source is not http, throw an error
+                if (existingSource.IsHttp && !existingSource.IsHttps && !existingSource.AllowInsecureConnections && !existingSource.AllowInsecureConnections)
```
kartheekp-ms:
```suggestion
                if (existingSource.IsHttp && !existingSource.IsHttps && !existingSource.AllowInsecureConnections)
```

removed duplicate condition from the if block.

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
```diff
@@ -286,10 +245,10 @@ public static void Run(UpdateSourceArgs args, Func<ILogger> getLogger)
...
                 {
-                    getLogger().LogWarning(string.Format(CultureInfo.CurrentCulture, Strings.Warning_HttpServerUsage, "update source", args.Source));
+                    throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_HttpSource_Single, "source update", args.Source));
```
kartheekp-ms: the command name is `dotnet nuget update source` should we continue using the old code i.e. `update source`?

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
```diff
@@ -383,7 +342,7 @@ public static void EnableOrDisableSource(PackageSourceProvider sourceProvider, s
                     Strings.SourcesCommandSourceEnabledSuccessfully, name));
                 if (packageSource.IsHttp && !packageSource.IsHttps && !packageSource.AllowInsecureConnections)
                 {
-                    getLogger().LogWarning(string.Format(CultureInfo.CurrentCulture, Strings.Warning_HttpServerUsage, "enable source", packageSource.Source));
+                    throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_HttpSource_Single, "source enable", packageSource.Source));
```
kartheekp-ms: can we use `enable source` to align with the command name?

src/NuGet.Core/NuGet.Commands/Strings.resx
```diff
@@ -1107,4 +1107,8 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
     <value>You are running the '{0}' operation with an 'HTTP' source: {1}. NuGet requires HTTPS sources. To use an HTTP source, you must explicitly set 'allowInsecureConnections' to true in your NuGet.Config file. Please refer to https://aka.ms/nuget-https-everywhere.</value>
     <comment>0 - The command name. Ex. Push/Delete. 1 - server URI.</comment>
   </data>
+  <data name="Error_HttpSource_Single_Short" xml:space="preserve">
```
kartheekp-ms, line 1110: Can we the existing `Error_HttpSource_Single` resource instead of adding a new one?

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
kartheekp-ms: aka.ms link seems to be broken. I found https://devblogs.microsoft.com/nuget/https-everywhere/ incase if it helps.

kartheekp-ms (commented):
Once all the PRs related to this effort are merged, I think we should remove [NU1803](https://learn.microsoft.com/nuget/reference/errors-and-warnings/nu1803) error code from the product. If we still would like to warn for HTTP sources in the `list source` command, then we can keep the error code.

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
Nigusu-Allehu: Oh i was looking at nuget.exe which has it as "source add". I will revert it since both can be valid

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
Nigusu-Allehu: I just created the aka.ms link and it does not redirect to a page yet but that will be worked on alongside the docs.

src/NuGet.Core/NuGet.Commands/Strings.resx
Nigusu-Allehu: It is because they are a little different. One tells the user to configure `allowInsecureConnections` while the other one does not.

kartheekp-ms: There is duplication in the strings for `Error_HttpSource_Single`, `Error_HttpSources_Multiple` and `Error_HttpSource_Single_Short` resources. If we ever have to change the error message, we have to update in all the places. I guess we should be okay good for now.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
```diff
@@ -126,23 +127,35 @@ public void Sources_WarnWhenAdding(string source, bool shouldWarn)
                     settings.ConfigPath
                 };
 
-                // Act
-                CommandRunnerResult result = _fixture.RunDotnetExpectSuccess(workingPath, string.Join(" ", args));
+                if (shouldFail)
```
kartheekp-ms: I think it would be better to split the test cases into two. My understanding is that `[InlineData]` is meant to pass multiple values for the same test case, not for having different code paths based on the value passed to the test case. This comment applies to other tests following the same pattern in this PR.

donnie-msft: I agree, since you've made a very lean setup for your test case (which is great!), it's cheap and cleaner to just copy the test and modify the arrange and assertions.

I recommend: 
`Sources_WhenAddingHttpSource_ShouldError`
`Sources_WhenAddingHttpsSource_ShouldSucceed`

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
```diff
@@ -184,24 +197,34 @@ public void Sources_WarnWhenUpdatingHttpSource(string updateSource, bool shouldW
...
 
         [PlatformTheory(Platform.Windows)]
-        [InlineData("http://source.test", "http://source.test.2", true, "warn : You are running the 'list source' operation with 'HTTP' source")]
-        [InlineData("https://source.test", "http://source.test.2", true, "warn : You are running the 'list source' operation with an 'HTTP' source")]
-        [InlineData("https://source.test", "https://source.test.2", false, "warn")]
-        public void Sources_WarnWhenListHttpSource(string initialSource, string secondSource, bool shouldWarn, string warningMessage)
+        [InlineData("http://source.test", "http://source.test.2")]
+        [InlineData("https://source.test", "http://source.test.2")]
+        [InlineData("https://source.test", "https://source.test.2")]
+        public void Sources_WhenListHttpSource(string initialSource, string secondSource)
```
donnie-msft:
```suggestion
        public void ListSources_WhenHttpSources_ErrorsForEachHttpSource(string initialSource, string secondSource)
```

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
```diff
@@ -236,40 +259,38 @@ public void Sources_WarnWhenListHttpSource(string initialSource, string secondSo
                 CommandRunnerResult result = _fixture.RunDotnetExpectSuccess(configFileDirectory, string.Join(" ", args));
 
                 // Assert
-                Assert.Equal(shouldWarn, result.Output.Contains(warningMessage));
+                Assert.Contains(initialSource, result.AllOutput);
```
donnie-msft, line 335: Is this scenario supposed to be an Error? If so, this assertion should be like the others and look for the Error output, not All Output.

src/NuGet.Core/NuGet.Commands/Strings.resx
donnie-msft: I think it's okay to do this, and the repo has a *lot* of duplicate strings in resource files. My understanding was that it's not a significant difference for the localization team. Perhaps I'm remembering it's even preferred to just create a duplicate resource?!

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/NuGetSourcesCommandTest.cs
```diff
@@ -347,15 +361,16 @@ public void SourcesCommandTest_DisableSource_NoWarnWhenUsingHttp()
...
 
         [Theory]
-        [InlineData("http://source.test", "http://source.test.2", true, "WARNING: You are running the 'list source' operation with 'HTTP' source")]
-        [InlineData("https://source.test", "http://source.test.2", true, "WARNING: You are running the 'list source' operation with an 'HTTP' source")]
-        [InlineData("https://source.test", "https://source.test.2", false, "WARNING")]
-        public void SourcesList_WithDefaultFormat_UsesDetailedFormat(string source, string secondSource, bool shouldWarn, string warningMessage)
+        [InlineData("http://source.test", "http://source.test.2")]
+        [InlineData("https://source.test", "http://source.test.2")]
+        [InlineData("https://source.test", "https://source.test.2")]
+        public void SourcesList_WithDefaultFormat_UsesDetailedFormat(string source, string secondSource)
```
donnie-msft, line 437: When I read this test's assertions, I don't understand where the "Detailed format" comes in. 
Also, shouldn't the test assert that no warning exists in the output?

donnie-msft (commented):
Looks pretty good! I think the test content is well done, just needs a little separation/renaming.

src/NuGet.Core/NuGet.Commands/Properties/AssemblyInfo.cs
```diff
@@ -9,3 +9,4 @@
...
 [assembly: InternalsVisibleTo("NuGet.PackageManagement.VisualStudio.Test, PublicKey=002400000480000094000000060200000024000052534131000400000100010007d1fa57c4aed9f0a32e84aa0faefd0de9e8fd6aec8f87fb03766c834c99921eb23be79ad9d5dcc1dd9ad236132102900b723cf980957fc4e177108fc607774f29e8320e92ea05ece4e821c0a5efe8f1645c4c0c93c1ab99285d622caa652c1dfad63d745d6f2de5f17e5eaf0fc4963d261c8a12436518206dc093344d5ad293")]
 [assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7")]
+[assembly: InternalsVisibleTo("Dotnet.Integration.Test, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
```
jeffkl, line 12: Be aware that this will conflict with my recent change: https://github.com/NuGet/NuGet.Client/pull/5744

You'll need to add an `<InternalsVisibleTo />` item to the `.csproj` instead.  I wonder if it would be worth rebasing your feature branch.

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
Nigusu-Allehu: I think that's a good idea. I have added a new warning

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSourcesTests.cs
Nigusu-Allehu: The nuget list command, allows users to list configured sources. However, just because a user has an http source, we should not be throwing an error for the list command. Instead, we should list all the configured sources and warn them about it. As a result, I have not added an error for the list command.

src/NuGet.Core/NuGet.Commands/Properties/AssemblyInfo.cs
Nigusu-Allehu: Thanks, I will rebase

src/NuGet.Core/NuGet.Commands/SourcesCommands/SourceRunners.cs
```diff
@@ -163,6 +163,7 @@ public static void Run(ListSourceArgs args, Func<ILogger> getLogger)
                             getLogger().LogMinimal(string.Format(CultureInfo.CurrentCulture, "{0}{1}", sourcePadding, source.Source));
                         }
 
+                        // Warn for non Https sources
```
kartheekp-ms: IMO, comment is redundant because the method name `WarnForHttpSources` is self-explanatory.

src/NuGet.Core/NuGet.Commands/Strings.Designer.cs
```diff
@@ -385,7 +385,7 @@ internal class Strings {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to You are running the &apos;{0}&apos; operation with an &apos;HTTP&apos; source: {1}. NuGet requires HTTPS sources. To use an HTTP source, you must explicitly set &apos;allowInsecureConnections&apos; to true in your NuGet.Config file. Please refer to https://aka.ms/nuget-https-everywhere..
+        ///   Looks up a localized string similar to You are running the &apos;{0}&apos; operation with an &apos;HTTP&apos; source: {1}. NuGet requires HTTPS sources. To use an HTTP source, you must explicitly set &apos;allowInsecureConnections&apos; to true in your NuGet.Config file. Please refer to https://aka.ms/nuget-https-everywhere for more information..
```
kartheekp-ms, line 388: We have an open issue to avoid using please in error messages. https://github.com/NuGet/Home/issues/8141

_Per the Microsoft style guide:_
Avoid please except in situations where the customer is asked to do something inconvenient, or the application or site is to blame for the situation.

This comment applies to the all the error messages that has the word `Please`.

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/NuGetSourcesCommandTest.cs
```diff
@@ -41,12 +47,23 @@ public void SourcesCommandTest_AddSource(string source, bool shouldWarn)
                 CommandRunnerResult result = CommandRunner.Run(nugetexe, workingPath, string.Join(" ", args));
 
                 // Assert
-                Assert.Equal(0, result.ExitCode);
-                ISettings loadedSettings = Configuration.Settings.LoadDefaultSettings(workingPath, null, null);
-                SettingSection packageSourcesSection = loadedSettings.GetSection("packageSources");
-                SourceItem sourceItem = packageSourcesSection?.GetFirstItemWithAttribute<SourceItem>("key", "test_source");
-                Assert.Equal(source, sourceItem.GetValueAsPath());
-                Assert.Equal(shouldWarn, result.Output.Contains("WARNING: You are running the 'add source' operation with an 'HTTP' source"));
+                if (shouldFail)
```
kartheekp-ms: Similar feedback as https://github.com/NuGet/NuGet.Client/pull/5702#discussion_r1578665526. This feedback applies to all the occurrences of this pattern introduced in this pull-request.

src/NuGet.Core/NuGet.Commands/Strings.Designer.cs
Nigusu-Allehu: https://github.com/NuGet/Home/issues/13430. Thanks for bringing this to my attention. We can merge this and I will make a separate PR to address this in all the errors that have been merged to the feature branch.

Nigusu-Allehu:
@NuGet/nuget-client Any suggestions?

nkolev92 (approved):

jeffkl (approved):

PR #5712, opened 2024-03-23
Render the ReadMe in PM UI

### LINKED ISSUE #12583: Support viewing README files in Visual Studio
### NuGet Product(s) Involved

### The Elevator Pitch

Last December, it was announced that you can edit and render markdown files now in Visual Studio. 

https://devblogs.microsoft.com/visualstudio/write-markdown-without-leaving-visual-studio/

It is a regular ask from the developer community to be able to read READMEs of packages while they are inside the PM UI whether browsing or getting started with a NuGet package. 

Developers should see a README in a tab similar to the [NuGet.org re-designed tabs](https://devblogs.microsoft.com/nuget/introducing-the-new-nuget-org-package-details-page/):

image

Inside of Visual Studio:

image

Please 👍 or 👎 this comment to help us with the direction of this feature & leave as much feedback/questions/concerns as you'd like on this issue itself and we will get back to you shortly.

Thank You 🎉

jgonz120:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12583

## Description

Added functionality in the PM UI to render the ReadMe for the selected package. Currently the ReadMe only renders the ReadMe for items which are available on the user's machine. 

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A


jgonz120:
@jebriede @donnie-msft Can you take a look at this draft PR? I have it set as draft right now because there are some questions about the packages. I'm hoping to get those answered next week but I would like to get feedback on the rest of the code. I'm working off the assumption that we'll be able to use IMarkdownPreview. 

In particular I want to make sure the way I'm getting the readme and showing/hiding the ReadMe tab are okay.

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
```diff
@@ -613,19 +660,22 @@ public DisplayVersion SelectedVersion
...
-
-                        NuGetUIThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+                        else
                         {
-                            try
-                            {
-                                await SelectedVersionChangedAsync(_searchResultPackage, _selectedVersion.Version, loadCts.Token).AsTask();
-                            }
-                            catch (OperationCanceledException) when (loadCts.IsCancellationRequested)
+
```
jebriede: Nit: remove extra blank line

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataControl.xaml.cs
```diff
@@ -17,69 +19,58 @@ namespace NuGet.PackageManagement.UI
     /// </summary>
     public partial class PackageMetadataControl : UserControl
     {
+        public ReadMePreviewViewModel ReadMePreviewViewModel { get; set; }
```
jebriede: Unused code

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataControl.xaml.cs
```diff
@@ -17,69 +19,58 @@ namespace NuGet.PackageManagement.UI
...
-        private void ViewLicense_Click(object sender, RoutedEventArgs e)
+#pragma warning disable VSTHRD100 // Avoid async void methods
+        private void PackageMetadataControl_DataContextChanged(object sender, DependencyPropertyChangedEventArgs e)
+#pragma warning restore VSTHRD100 // Avoid async void methods
         {
-            if (DataContext is DetailedPackageMetadata metadata)
+            if (DataContext is DetailControlModel detailControlModel)
             {
-                var window = new LicenseFileWindow()
+                detailControlModel.PropertyChanged += PackageMetadataControl_DataContext_PropertyChanged;
```
jebriede: Be sure to unregister the old detailControlModel's PropertyChanged before replacing it and registering the new one's listener.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/MultiSourcePackageMetadataProvider.cs
```diff
@@ -4,8 +4,10 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection.Metadata.Ecma335;
```
jebriede: Is this intentional?

jebriede (commented):
Great work here! We did a live review together and to summarize my feedback:
1) The Readme control should wrap the MarkdownPreview control and handle all the view-related interactions, including when it should make the call to update the content and when it should be visible.
2) The ReadmeViewModel should handle how to load and update the readme. The Readme control ideally will tell the ViewModel when to fetch the readme markdown and will listen for the readme's property changed to invoke the update of the content on the MarkdownPreview control for rendering.
3) Move the readme logic out of the DetailControlModel and into the ReadmeViewModel to separate the concerns and encapsulate the readme logic in its own view model.
4) The ReadmeViewModel should ideally not have to be a property on the DetailControlModel.

Those are just some code reorganizations to use the MVVM pattern. Let me know when the next review is available!

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadMeControl.xaml.cs
```diff
@@ -19,11 +15,84 @@ namespace NuGet.PackageManagement.UI
...
+            if (_disposed)
+            {
+                return;
+            }
+
+            if (disposing)
+            {
+                _markdownPreview?.Dispose();
+                var viewModel = (ReadMePreviewViewModel)DataContext;
+                viewModel.PropertyChanged -= ViewModel_PropertyChangedAsync;
```
jebriede, line 71: This will only unregister the event handler for the last viewModel when the PMRMC disposes. You should instead make sure this unregister happens when the viewModel changes, unregister the old viewModel, and register the new one.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadMeControl.xaml.cs
```diff
@@ -19,11 +15,84 @@ namespace NuGet.PackageManagement.UI
...
+            }
+
+            _disposed = true;
+        }
+
+        private void PackageMetadataReadMeControl_DataContextChanged(object sender, DependencyPropertyChangedEventArgs e)
+        {
+            var viewModel = (ReadMePreviewViewModel)DataContext;
+            viewModel.PropertyChanged += ViewModel_PropertyChangedAsync;
```
jebriede: This presents a memory leak. Mentioned in another comment, when you register this new view model to listen for events, you have to unregister the old one first or it will continue to hold on to a reference to this control and it will never be able to be garbage collected.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadMeControl.xaml.cs
```diff
@@ -19,11 +15,84 @@ namespace NuGet.PackageManagement.UI
...
+        }
+
+#pragma warning disable VSTHRD100 // Avoid async void methods
+        private async void ViewModel_PropertyChangedAsync(object sender, System.ComponentModel.PropertyChangedEventArgs e)
+#pragma warning restore VSTHRD100 // Avoid async void methods
+        {
+            var viewModel = (ReadMePreviewViewModel)DataContext;
```
jebriede: Instead of casting the DataContext as the expected type whenever it's referenced, I like to have a private property that returns the DataContext as the expected type and use that instead.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadMePreviewViewModel.cs
```diff
@@ -50,30 +29,88 @@ public bool IsErrorWithReadMe
...
+            set
+            {
+                _isReadMeAvailable = value;
+                RaisePropertyChanged(nameof(IsReadMeAvailable));
+            }
+        }
+
+        private string _rawReadMe;
+
+        public string RawReadMe
```
jebriede:
```suggestion
        public string ReadmeMarkdown
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataControl.xaml.cs
jebriede: This still appears in this iteration. Is this needed?

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataControl.xaml.cs
```diff
@@ -2,76 +2,67 @@
...
-                    }
-                    else
-                    {
-                        tabReadMe.Visibility = Visibility.Visible;
-                    }
-
-                }
-
+                tabReadMe.IsSelected = true;
+                tabPackageDetails.IsSelected = false;
```
jebriede: Do we have to set all other tabs as IsSelected = false? Is it sufficient to set the tab we want selected to true? I'm not certain, but I believe that's how the tab control works. You'd need to test to be sure, but I don't believe it allows for more than one tab to be selected at a time.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataControl.xaml.cs
```diff
@@ -2,76 +2,67 @@
...
 
     /// <summary>
     /// Interaction logic for PackageMetadataControl.xaml
     /// </summary>
-    public partial class PackageMetadataControl : UserControl
+    public partial class PackageMetadataControl : UserControl, IDisposable
     {
         public ReadMePreviewViewModel ReadMePreviewViewModel { get; set; }
 
+        private bool _initialVisibilitySet = false;
```
jebriede: I don't quite understand what this initialVisibilitySet is for. Could you leave a comment explaining its purpose? It looks like it's used to determine when to select the Readme tab. A comment would help explain what behavior this enforces.

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
```diff
@@ -555,19 +538,7 @@ public DetailedPackageMetadata PackageMetadata
...
-                                    IsReadMeAvailable = !string.IsNullOrWhiteSpace(readMe);
-                                }
-                                catch (Exception)
-                                {
-                                    await ReadMePreviewViewModel.UpdateMarkdownAsync("");
-                                    ReadMePreviewViewModel.IsErrorWithReadMe = true;
-                                    IsReadMeAvailable = true;
-                                    throw;
-                                }
+                                await ReadMePreviewViewModel.UpdateReadMe(_packageMetadata?.PackagePath, _packageMetadata?.Id);
```
jebriede: Could the Readme view call on update readme rather than the setter of the PackageMetadata? It looks like the data context is being set on the view to the new ReadmePreviewViewModel already, so is that a good time to update the readme which will invoke a property changed event on the VM and the view can react by then rendering the markdown? This would remove the responsibility from the DetailControlModel of having to drive the ReadMeViewModel.

jebriede (commented):
This is headed in the right direction. Initial comments were not all addressed, and I left a few more. Let me know if you have any questions and want to chat offline!

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadMeControl.xaml.cs
jgonz120: I like this :O

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataControl.xaml.cs
jgonz120: Hmm, before adding a comment, what I'm trying to do is make sure the right tab is selected when the control is first rendered. If there is a readme present then I was the ReadMe tab to be selected first. Otherwise I want to have the details tab selected. The problem I have is that on initial load of the component the DetailsModel doesn't have the package path set and so the ReadMe tab is disabled. Then it gets populated and the TabReadMe_IsVisibleChanged is fired. SO the first time it fires I want to set the selected tab correctly and I use this flag to determine if it's the first time that event is fired. 

If you have any suggestions let me know bc I don't love this solution. Otherwise I can add a comment explaining

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/MultiSourcePackageMetadataProvider.cs
jgonz120: nope

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadMeControl.xaml.cs
jgonz120: Okay, added it to the context change and unregistered it here too, needs to be removed in both instances right?

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataControl.xaml.cs
jgonz120: yup looks like it's not needed

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadMeControl.xaml.cs
jebriede: Yeah, you're right. It should also be unregistered in the dispose.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataControl.xaml.cs
jebriede: Is there an event when the package path gets set on the DetailsModel? Whenever the package path is populated, the UI could respond by setting the is visible on the Readme tab.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerTopPanel.xaml
```diff
@@ -180,18 +180,16 @@
...
-
       <Grid
         Grid.Column="0">
         <Grid.ColumnDefinitions>
           <ColumnDefinition Width="*" MaxWidth="320" MinWidth="118"/>
           <ColumnDefinition Width="auto" />
-          <ColumnDefinition Width="auto" />
-          <ColumnDefinition Width="auto" />
+          <ColumnDefinition Width="*" MaxWidth="125" />
+          <ColumnDefinition Width="*" MaxWidth="150"/>
```
donnie-msft: Can omit the defaults. Same applies to other usages.
```suggestion
          <ColumnDefinition MaxWidth="125" />
          <ColumnDefinition MaxWidth="150"/>
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataControl.xaml
```diff
@@ -5,9 +5,12 @@
   xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
   xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
   xmlns:nuget="clr-namespace:NuGet.PackageManagement.UI"
+  xmlns:vm="clr-namespace:NuGet.PackageManagement.UI.ViewModels"
   xmlns:imaging="clr-namespace:Microsoft.VisualStudio.Imaging;assembly=Microsoft.VisualStudio.Imaging"
   xmlns:catalog="clr-namespace:Microsoft.VisualStudio.Imaging;assembly=Microsoft.VisualStudio.ImageCatalog"
+  xmlns:diag="clr-namespace:System.Diagnostics;assembly=WindowsBase"
```
donnie-msft: I'd check if some of these namespaces are really being used in XAML.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerTopPanel.xaml
```diff
@@ -246,7 +244,7 @@
           Unchecked="_checkboxVulnerabilities_Unchecked"
           Visibility="Hidden"
           IsChecked="False">
-          <TextBlock Text="{x:Static nuget:Resources.Checkbox_Show_Vulnerable_Only}" />
+          <TextBlock TextTrimming="CharacterEllipsis" Text="{x:Static nuget:Resources.Checkbox_Show_Vulnerable_Only}" />
```
donnie-msft, line 247: Thanks for working to improve the reflow layout issues in PM UI. I believe you could split this into a separate PR and get it merged and separated from the README work.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/MultiSourcePackageMetadataProvider.cs
```diff
@@ -247,6 +257,25 @@ private static IEnumerable<PackageVulnerabilityMetadata> MergeVulnerabilityMetad
...
+                    {
+                        return localPackageSearchMetadata.PackagePath;
+                    }
+                    if (packageMetadata is PackageSearchMetadataBuilder.ClonedPackageSearchMetadata clonedPackage)
+                    {
+                        return clonedPackage.PackagePath;
+                    }
+                    return string.Empty;
+                })
+                .Where(packagePath => !string.IsNullOrWhiteSpace(packagePath));
```
donnie-msft: The `where` can be put before the `select`.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/MultiSourcePackageMetadataProvider.cs
```diff
@@ -247,6 +257,25 @@ private static IEnumerable<PackageVulnerabilityMetadata> MergeVulnerabilityMetad
             return vulnerabilityMetadatas.FirstOrDefault(v => v != null && v.Any());
         }
 
+        private static string MergePackagePath(IEnumerable<IPackageSearchMetadata> packages)
+        {
+            var packagePaths = packages
+                .Select(packageMetadata =>
+                {
+                    if (packageMetadata is LocalPackageSearchMetadata localPackageSearchMetadata)
```
donnie-msft: I guessed we had this logic somewhere, but I couldn't find it. Perhaps it could be put in a utility class, but this provider seems reasonable to me.

src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/ContextInfos/PackageSearchMetadataContextInfo.cs
```diff
@@ -67,7 +67,7 @@ public static PackageSearchMetadataContextInfo Create(IPackageSearchMetadata pac
                 PackageDetailsUrl = packageSearchMetadata.PackageDetailsUrl,
                 PackagePath =
                     (packageSearchMetadata as LocalPackageSearchMetadata)?.PackagePath ??
-                    (packageSearchMetadata as ClonedPackageSearchMetadata)?.PackagePath,
+                    (packageSearchMetadata as PackageSearchMetadataBuilder.ClonedPackageSearchMetadata)?.PackagePath,
```
donnie-msft, line 70: Any idea what made this necessary?

NuGet.Config
```diff
@@ -8,10 +8,15 @@
     <add key="myget-legacy@Local" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/myget-legacy%40Local/nuget/v3/index.json" />
     <add key="nuget-build" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/nuget-build/nuget/v3/index.json" />
     <add key="vside" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/msft_consumption/nuget/v3/index.json" />
+    <add key="vs-impl" value="https://devdiv.pkgs.visualstudio.com/_packaging/vs-impl/nuget/v3/index.json" />
   </packageSources>
   <packageSourceMapping>
     <clear />
-    <packageSource key = "dotnet-public">
+    <packageSource key="vs-impl">
```
donnie-msft, line 16: We can't have private feeds in our OSS repo. I believe this feed is an upstream for msft_consumption (`vside`).

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
```diff
@@ -613,19 +609,21 @@ public DisplayVersion SelectedVersion
                         {
                             PackageMetadata = detailedPackageMetadata;
                         }
-
-                        NuGetUIThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+                        else
```
donnie-msft, line 612: Is this fixing another bug with the version dropdown?

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
```diff
@@ -1042,4 +1042,10 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
     <value>The package `{0}` is available in the Global packages folder, but the source it came from `{1}` is not one of the configured sources.</value>
     <comment>{0} is the package ID. {1} is the URI of the package source found in the Global packages folder.</comment>
   </data>
+  <data name="Label_PackageDetails" xml:space="preserve">
+    <value>Package Details</value>
+  </data>
+  <data name="Error_UnableToLoadReadme" xml:space="preserve">
+    <value>An error occured while trying to load the ReadMe.</value>
```
donnie-msft: Is this generally all caps?

```suggestion
    <value>An error occurred while trying to load the README.</value>
```

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadMePreviewViewModel.cs
```diff
@@ -0,0 +1,120 @@
...
+        {
+            get => _rawReadMe;
+            set
+            {
+                _rawReadMe = value;
+                RaisePropertyChanged(nameof(ReadMeMarkdown));
+            }
+        }
+
+        public async Task UpdateReadMe(string packagePath, string packageId)
```
donnie-msft: nit: the term update makes me think of writing content to disk.
Reload or Load, maybe?

```suggestion
        public async Task LoadReadMe(string packagePath, string packageId)
```

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadMePreviewViewModel.cs
```diff
@@ -0,0 +1,120 @@
...
+            {
+                if (!string.IsNullOrEmpty(packagePath))
+                {
+                    var fileInfo = new FileInfo(packagePath);
+                    if (fileInfo.Exists)
+                    {
+                        using var stream = fileInfo.OpenRead();
+                        using var archive = new ZipArchive(stream, ZipArchiveMode.Read);
+                        var nuspecZipArchiveEntry = archive.Entries.FirstOrDefault(zipEntry => string.Equals(UnescapePath(zipEntry.Name), $"{packageId}{PackagingCoreConstants.NuspecExtension}", StringComparison.OrdinalIgnoreCase));
+                        if (nuspecZipArchiveEntry is not null)
```
donnie-msft, line 61: The else here is that the README wasn't able to be read, but I don't see how the `isErrorWithReadMe` would become true.
I think the rendering would show no README, but shouldn't it show an error? At least it'd be good to fire telemetry so we can see how common it is that we can't read the archives.

donnie-msft (commented):
This feature is going to be awesome! We talked and I know you're working on some edge cases, but wanted to go ahead and read through your draft.

Biggest thing: go ahead and pull out the reflow work you've done. That doesn't need to wait on the README changes, as it's an active bug you've already addressed here. I understand why it's more relevant as part of this work. 

If you don't want to change the reflow separately, you could still pull out the ellipses change and merge that to fix the Show Only Vulnerable checkbox overlapping bug.

src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/ContextInfos/PackageSearchMetadataContextInfo.cs
jgonz120: ClonedPackageSearchMetadata is a class defined in PackageSearchMetadataBuilder. 
I can do 
```using static NuGet.Protocol.Core.Types.PackageSearchMetadataBuilder```

File: NuGet.Config (hunk 18 above)
Comment (jgonz120) [reply NuGet.Config:16]: yup! Working on getting the package published to the public vs-impl feed

File: src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs (hunk 19 above)
Comment (jgonz120) [reply src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs:612]: No, not fixing a bug persay. If you look at SelectedVersionChangedAsync it ends setting the PackageMetadata property again. 

I didn't think it was necessary to try to retrieve the metadata information and update the _metadataDict cache after we've already used it.

File: src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx (hunk 20 above)
Comment (jgonz120) [reply src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx]: looks like we do "Readme" in other locations, but it's "README" on nuget.org.

File: src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/MultiSourcePackageMetadataProvider.cs (hunk 15 above)
Comment (jgonz120) [reply src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/MultiSourcePackageMetadataProvider.cs]: I changed the linq query completely, I felt like the select was a little confusing. Lemme know what you think

File: src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadMePreviewViewModel.cs (hunk 22 above)
Comment (jgonz120) [reply src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadMePreviewViewModel.cs:61]: So the else here would mean that a nuspec wasn't found in the nupkg, which would be weird but I didn't see it as a problem specifically with rendering the readme information and with the package itself. 

I think logging telemetry for it makes sense since it's a strange scenario, I was more so trying to avoid a null reference exception.

**Comment by mckennabarlow** [comment]
I know I'm late to the party, but I was wondering what the plan is to show in the PMUI when the NuGet package does not contain a README file. As we are continuing our effort to increase README adoption and incentivize package owners to include a README, this feature of showing in the PMUI will be very helpful as an incentive to add a README to your package, as well as an opportunity for a warning if you don't. I'm wondering if there is no README file included in the package if we could show some warning in the PMUI instead of just leaving it blank. Hopefully if we had some sort of visual reminder to encourage folks to include a README this could organically increase README adoption. Is this something that might be possible in the PM UI in VS?

**Comment by donnie-msft** [comment]
@mckennabarlow there is a spec for this feature where I voted for the option of showing (not hiding) the tab when there is no README. I like that UX better in general, but it also would allow for showing something like "Help me create a README" affordance right there within the tab.

Maybe you can check review the spec and help answer some of those open questions. https://github.com/NuGet/Home/pull/13200/files/753100cd4507a5620e0de77bd6a10985ac944a33#r1490237966

File: src/NuGet.Clients/NuGet.VisualStudio.Client/.vsixinclude [hunk 23]
```diff
@@ -38,7 +38,11 @@ Microsoft.Extensions.FileSystemGlobbing.dll
...
 Newtonsoft.Json.dll
-System.Buffers.dll
-System.Memory.dll
 System.Numerics.Vectors.dll
-System.Runtime.CompilerServices.Unsafe.dll
+Microsoft.VisualStudio.Linux.ConnectionManager.Store.dll
+Microsoft.VisualStudio.Markdown.Core.dll
+Microsoft.VisualStudio.Markdown.Platform.dll
+System.Drawing.Common.dll
+Microsoft.Internal.VisualStudio.Shell.Framework.resources.dll
```
donnie-msft: Seems like you're on the right track moving some of these to the vsixignore. 
Have you tried moving `Microsoft.Internal.VisualStudio.Shell.Framework.resources.dll` as well?

src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/ContextInfos/PackageSearchMetadataContextInfo.cs
donnie-msft: That type existed prior to this PR. Can the whole file just be reverted?

jgonz120:
Merging into a feature branch instead, https://github.com/NuGet/NuGet.Client/pull/5920

PR #5714, opened 2024-03-25
NuGetAuditSuppress for LegacyPackageReferenceProject in VS

zivkan:
## Bug

Related: https://github.com/NuGet/Client.Engineering/issues/2709

Regression? no

## Description

* Make `LegacyPackageReferenceProject.GetPackageSpecs` get `NuGetAuditSuppress` items from the project system, to populate `RestoreAuditProperties`.
* change async `IVsProjectAdapter.GetBuildItemInformationAsync` to non-async `IVsProjectAdapter.GetBuildItemInformation`, in order to prevent unnecessary async state machine allocations. It was being called by another method which in turn had an unnecessary async state machine allocation, so we're saving multiple allocations per call.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
@@ -128,10 +128,12 @@ namespace NuGet.PackageManagement.VisualStudio
             return (new[] { packageSpec }, null);
         }
 
-        private async Task<Dictionary<string, CentralPackageVersion>> GetCentralPackageVersionsAsync()
+        private Dictionary<string, CentralPackageVersion> GetCentralPackageVersions()
         {
+            ThreadHelper.ThrowIfNotOnUIThread();
+
             IEnumerable<(string PackageId, string Version)> packageVersions =
-                        (await _vsProjectAdapter.GetBuildItemInformationAsync(ProjectBuildProperties.PackageVersion, ProjectBuildProperties.Version))
+                        _vsProjectAdapter.GetBuildItemInformation(ProjectBuildProperties.PackageVersion, ProjectBuildProperties.Version)
                         .Select(item => (PackageId: item.ItemId, Version: item.ItemMetadata.FirstOrDefault()));
 
             return packageVersions
@@ -154,6 +156,65 @@ namespace NuGet.PackageManagement.VisualStudio
             }
 
             return new CentralPackageVersion(packageId, VersionRange.Parse(version));
+        }
+
+        private RestoreAuditProperties GetRestoreAuditProperties()
+        {
+            ThreadHelper.ThrowIfNotOnUIThread();
+
+            string enableAudit = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAudit);
+            string auditLevel = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditLevel);
+            string auditMode = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditMode);
+            HashSet<string> suppressedAdvisories = GetSuppressedAdvisories();
+
+            bool hasAnyValue = !string.IsNullOrWhiteSpace(enableAudit)
+                || !string.IsNullOrWhiteSpace(auditLevel)
+                || !string.IsNullOrWhiteSpace(auditMode)
+                || suppressedAdvisories?.Count > 0;
+
+            RestoreAuditProperties auditProperties = hasAnyValue
+                ? new RestoreAuditProperties()
+                {
+                    EnableAudit = enableAudit,
+                    AuditLevel = auditLevel,
+                    AuditMode = auditMode,
+                    SuppressedAdvisories = suppressedAdvisories,
+                }
+                : null;
+            return auditProperties;
+        }
+
+        private HashSet<string> GetSuppressedAdvisories()
+        {
+            ThreadHelper.ThrowIfNotOnUIThread();
+
+            IEnumerable<(string ItemId, string[] ItemMetadata)> buildItems = _vsProjectAdapter.GetBuildItemInformation(ProjectBuildProperties.NuGetAuditSuppress);
+            if (buildItems is null)
+            {
+                return null;
+            }
+            else if (buildItems is ICollection<(string, string[])> collection)
+            {
+                if (collection.Count == 0) return null;
+
+                var suppressedAdvisories = new HashSet<string>(collection.Count, StringComparer.OrdinalIgnoreCase);
+                foreach ((string itemId, _) in buildItems.NoAllocEnumerate())
+                {
+                    suppressedAdvisories.Add(itemId);
+                }
+
+                return suppressedAdvisories;
+            }
+            else
+            {
+                var suppressedAdvisories = new HashSet<string>(StringComparer.Ordinal);
+                foreach ((string itemId, _) in buildItems.NoAllocEnumerate())
+                {
+                    suppressedAdvisories.Add(itemId);
+                }
+
+                return suppressedAdvisories.Count == 0 ? null : suppressedAdvisories;
+            }
         }
 
         #endregion
@@ -396,7 +457,7 @@ namespace NuGet.PackageManagement.VisualStudio
             if (isCpvmEnabled)
             {
                 // Add the central version information and merge the information to the package reference dependencies
-                projectTfi.CentralPackageVersions.AddRange(await GetCentralPackageVersionsAsync());
+                projectTfi.CentralPackageVersions.AddRange(GetCentralPackageVersions());
                 LibraryDependency.ApplyCentralVersionInformation(projectTfi.Dependencies, projectTfi.CentralPackageVersions);
             }
 
@@ -441,17 +502,7 @@ namespace NuGet.PackageManagement.VisualStudio
                 }
             }
 
-            string enableAudit = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAudit);
-            string auditLevel = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditLevel);
-            string auditMode = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditMode);
-            RestoreAuditProperties auditProperties = !string.IsNullOrEmpty(enableAudit) || !string.IsNullOrEmpty(auditLevel)
-                ? new RestoreAuditProperties()
-                {
-                    EnableAudit = enableAudit,
-                    AuditLevel = auditLevel,
-                    AuditMode = auditMode,
-                }
-                : null;
+            RestoreAuditProperties auditProperties = GetRestoreAuditProperties();
 
             var msbuildProjectExtensionsPath = await GetMSBuildProjectExtensionsPathAsync();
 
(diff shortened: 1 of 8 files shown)


src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/VsProjectAdapter.cs
```diff
@@ -220,18 +222,11 @@ public async Task<IEnumerable<(string ItemId, string[] ItemMetadata)>> GetBuildI
                 throw new ArgumentNullException(nameof(itemName));
             }
 
-            await _threadingService.JoinableTaskFactory.SwitchToMainThreadAsync();
-
-            var itemStorage = VsHierarchy as IVsBuildItemStorage;
-            if (itemStorage != null)
```
zivkan: I'm in two minds about this change. Let's assume someone is building a new project system, they make NuGet think it supports NuGet and PackageReference, but isn't a CPS project so ends up here, but most importantly the project doesn't implement `IVsBuildItemStorage`.

Before this change, NuGet will silently ignore the failure to get items, always telling the rest of our code "no, this project doesn't define any items of that type".  If the project does in fact define items of that type, then this will be **EXTREMELY** difficult for us to diagnose.

With my change, the code will throw an exception telling us exactly what the problem is. However, it will throw even when the project doesn't define items with the specified name, and therefore this codepath isn't even needed. This will completely break NuGet with this project system, making it impossible to restore, even if the customer tries to avoid using CPM or NuGetAuditSuppress (the only two places this method is used currently).

What do others think?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
```diff
@@ -156,6 +158,65 @@ private CentralPackageVersion ToCentralPackageVersion(string packageId, string v
...
+            string auditLevel = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditLevel);
+            string auditMode = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditMode);
+            HashSet<string> suppressedAdvisories = GetSuppressedAdvisories();
+
+            bool hasAnyValue = !string.IsNullOrWhiteSpace(enableAudit)
+                || !string.IsNullOrWhiteSpace(auditLevel)
+                || !string.IsNullOrWhiteSpace(auditMode)
+                || suppressedAdvisories?.Count > 0;
+
+            RestoreAuditProperties auditProperties = hasAnyValue
```
nkolev92: Does this affect no-op, or is this skipped if there's no values?
What does CLI do?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/VsProjectAdapter.cs
nkolev92: We don't really know about the project systems there and whether they implement IVsBuildItemStorage. 

Maybe I'm more paranoid, but I'm leaning towards having a check, because we're likely to only see these issues *after* we ship. 

You can maybe feature flag it or something, if you want to do an experiment.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
zivkan: The reason this code exists is to ensure that no-op is as consistent as possible. I tried to make this method as easy as possible to understand, so I'm not sure how else to make it easier. `GetSuppressedAdvisories()` is a little more complex, but that's because it's try to do as many early-exits as possible to minimize memory allocations and therefore increase performance.

nkolev92: Sounds like this makes no-op consistent.

Based on the fact that we always set these properties however, this check is likely never going to be true.

nkolev92 (dismissed):

nkolev92:
@NuGet/nuget-client 
Feedback's been addressed. 
Can you please take a look so this can get merged.  
I had to push to this branch so I can't complete it myself.

zivkan (commented):
LGTM, but I created the PR, so GitHub won't let me approve it.

nkolev92:
@NuGet/nuget-client - Can I please get a review.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/VsProjectAdapter.cs
```diff
@@ -220,9 +222,7 @@ public async Task<IEnumerable<(string ItemId, string[] ItemMetadata)>> GetBuildI
                 throw new ArgumentNullException(nameof(itemName));
             }
 
-            await _threadingService.JoinableTaskFactory.SwitchToMainThreadAsync();
-
-            var itemStorage = VsHierarchy as IVsBuildItemStorage;
+            var itemStorage = (IVsBuildItemStorage)VsHierarchy;
             if (itemStorage != null)
```
jebriede, line 229: This null check isn't necessary anymore if we are doing an explicit cast in the line above, now. We are asserting that VsHierarchy could never be null and it will always be castable to IVsBuildItemStorage. In that case, we can remove the null check for itemStorage.

jebriede (commented):
Look good! Left one comment that should be addressed to clean up a null check.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/VsProjectAdapter.cs
nkolev92: You actually caught a bug. 

The cast should be `as`. 
I'll fix that. We don't have a guarantee that type is implemented.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
```diff
@@ -128,10 +128,12 @@ public override async Task<(IReadOnlyList<PackageSpec> dgSpecs, IReadOnlyList<IA
             return (new[] { packageSpec }, null);
         }
 
-        private async Task<Dictionary<string, CentralPackageVersion>> GetCentralPackageVersionsAsync()
```
kartheekp-ms, line 131: Thanks for taking the time to remove several unnecessary async methods in this PR.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
```diff
@@ -156,6 +158,57 @@ private CentralPackageVersion ToCentralPackageVersion(string packageId, string v
...
+                foreach ((string itemId, _) in buildItems.NoAllocEnumerate())
+                {
+                    suppressedAdvisories.Add(itemId);
+                }
+
+                return suppressedAdvisories;
+            }
+            else
+            {
+                var suppressedAdvisories = new HashSet<string>(StringComparer.Ordinal);
```
kartheekp-ms: Should this also be `StringComparer.OrdinalIgnoreCase` just like line no. 192? 

```suggestion
                var suppressedAdvisories = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
```diff
@@ -156,6 +158,57 @@ private CentralPackageVersion ToCentralPackageVersion(string packageId, string v
...
+            if (buildItems is null)
+            {
+                return null;
+            }
+            else if (buildItems is ICollection<(string, string[])> collection)
+            {
+                if (collection.Count == 0) return null;
+
+                var suppressedAdvisories = new HashSet<string>(collection.Count, StringComparer.OrdinalIgnoreCase);
+                foreach ((string itemId, _) in buildItems.NoAllocEnumerate())
```
kartheekp-ms, line 193: I am curious why are we discarding the `string[]` collection?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
```diff
@@ -156,6 +158,57 @@ private CentralPackageVersion ToCentralPackageVersion(string packageId, string v
...
+            string auditMode = _vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.NuGetAuditMode);
+            HashSet<string> suppressedAdvisories = GetSuppressedAdvisories();
+
+            return new RestoreAuditProperties()
+            {
+                EnableAudit = enableAudit,
+                AuditLevel = auditLevel,
+                AuditMode = auditMode,
+                SuppressedAdvisories = suppressedAdvisories,
+            };
```
kartheekp-ms, line 176: In the existing code that is deleted from this PR in favor of this method, restore audit properties are only returned when audit is enabled, and audit level is set. If these properties are not set, then `null` is returned. However, in this case my understanding is that we always return `RestoreAuditProperties`. I am assuming that should be okay because if audit is disabled then these properties will have their values configured accordingly but just want to check.

```cs
RestoreAuditProperties auditProperties = !string.IsNullOrEmpty(enableAudit) || !string.IsNullOrEmpty(auditLevel)
                ? new RestoreAuditProperties()
                {
                    EnableAudit = enableAudit,
                    AuditLevel = auditLevel,
                    AuditMode = auditMode,
                }
                : null;
```

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
nkolev92: Yeah nice catch. 

Fortunately, I don't expect the 3rd tree to ever be hit :D

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
nkolev92: That represents that attribute on each item and we don't use the attributes from the item, we only use the include value.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
nkolev92: A few things:

* This ensures no-op consistency across VS/CLI. 
* The NuGet.Targets file has default, so in practice those values will never be empty: https://github.com/NuGet/NuGet.Client/blob/252bbee32a839623dc9af267f950d21cde5e9497/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets#L70-L77

nkolev92:
Ready for another look @jebriede @kartheekp-ms

jebriede (approved):

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
kartheekp-ms: (unresolved the conversation) If we don't expect the 3rd tree to ever be executed, I am wondering if we really need that code. I don't want team members in the future to think that every time we read MSBuild items, we should consider this as a pattern that may result in adding some dead code into the product. Please correct if I am missing something.

nkolev92: There's a chance the type returned is not ICollection (IEnumerable by contract), so by contract the 3rd branch is necessary. 
We just know there's 1 implementation of that service and we don't expect there to be another one, but that's not a assumption our code can make safely.

kartheekp-ms: Thanks for clarifying.

PR #5716, opened 2024-03-25
Make OptProf MicroBuild plug-in work again for Official Builds

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2749
 
## Description

Follow-up to https://github.com/NuGet/NuGet.Client/pull/5706

Copies pipeline parameter to a template variable in the parent YAML so that `Conditions` work. Most relevantly, this means OptProf MicroBuild plugin will be invoked by the 1ES templates.

Also remove outputs with conditions would never be true (ie, values never satisfy conditions for BuildRTM and IsOfficialBuild).

[Sample official build](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=9306988&view=logs&j=d80b2849-30d4-52dd-74cd-d6536ba98fe0&t=834cc09b-7de8-53f7-7930-d1ac3e92b356)

[Sample OptProf](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=9306989&view=results)
[Sample OptProf 2](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=9307253&view=results)

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/eng/pipelines/templates/pipeline.yml
+++ b/eng/pipelines/templates/pipeline.yml
@@ -107,11 +107,14 @@ stages:
         swix:
           enabled: true
         optprof:
-          enabled: "and(succeeded(), eq(variables['IsOfficialBuild'], 'true'))"
-          OptimizationInputsLookupMethod: DropPrefix
-          DropNamePrefix: OptimizationInputs/$(System.TeamProject)/$(Build.Repository.Name)
-          ShouldSkipOptimize: $(ShouldSkipOptimize)
-          AccessToken: $(System.AccessToken)
+        ${{ if and(succeeded(), eq(variables['IsOfficialBuild'], 'true')) }}:
+            enabled: true
+            OptimizationInputsLookupMethod: DropPrefix
+            DropNamePrefix: OptimizationInputs/$(System.TeamProject)/$(Build.Repository.Name)
+            ShouldSkipOptimize: $(ShouldSkipOptimize)
+            AccessToken: $(System.AccessToken)
+          ${{ else }}:
+            enabled: false
       outputs:
       - output: pipelineArtifact
         displayName: 'Publish buildinfo.json as an artifact'
--- a/eng/pipelines/templates/pipeline.yml
+++ b/eng/pipelines/templates/pipeline.yml
@@ -108,13 +108,13 @@ stages:
           enabled: true
         optprof:
         ${{ if and(succeeded(), eq(variables['IsOfficialBuild'], 'true')) }}:
-            enabled: true
-            OptimizationInputsLookupMethod: DropPrefix
-            DropNamePrefix: OptimizationInputs/$(System.TeamProject)/$(Build.Repository.Name)
-            ShouldSkipOptimize: $(ShouldSkipOptimize)
-            AccessToken: $(System.AccessToken)
-          ${{ else }}:
-            enabled: false
+          enabled: true
+          OptimizationInputsLookupMethod: DropPrefix
+          DropNamePrefix: OptimizationInputs/$(System.TeamProject)/$(Build.Repository.Name)
+          ShouldSkipOptimize: $(ShouldSkipOptimize)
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/pipeline.yml
```diff
@@ -106,39 +107,39 @@ stages:
         swix:
           enabled: true
         optprof:
-          enabled: "and(succeeded(), eq(variables['BuildRTM'], 'false'), eq(variables['IsOfficialBuild'], 'true'))"
```
nkolev92, line 109: Why is the BuildRTM unnecessary? 
Wouldn't it now run  in 2 phases while it previous ran in 1?

nkolev92: line: 95 hardcodes it.

nkolev92 (dismissed):

eng/pipelines/templates/pipeline.yml
zivkan: Line 95 sets up a job variable, but it doesn't prevent this optprof step to skip on official build's RTM job.

eng/pipelines/templates/pipeline.yml
```diff
@@ -106,39 +108,39 @@ stages:
...
+          enabled: ${{ variables['isOfficialBuild'] }}
```
zivkan, line 111: you probably don't want to remove `succeeded()`, as there's no point installing the optprof plugin if a previous step already failed.

donnie-msft:
> Line 95 sets up a job variable, but it doesn't prevent this optprof step to skip on official build's RTM job.

@zivkan could you rephrase this, please? Optprof only runs on NonRTM

eng/pipelines/templates/pipeline.yml
donnie-msft: Does this look right to you? I can't find documentation that this is supported within a variable. I also didn't see that the templates pass in a condition - but this is "compiling" at least:
https://github.com/NuGet/NuGet.Client/pull/5716/commits/81376bd6c90cf5206e1a542fa443617cb438f9fa#diff-ff6085ab5cf199ecbf9b59ba9a5854fdc078bdc1c7516d16c95f6bff5ecd070aR112

jeffkl: The plug-ins are installed nearly first, after clone but before any of our tasks run.  I don't think we need a condition on this step.

eng/pipelines/templates/pipeline.yml
zivkan: > @zivkan could you rephrase this, please? Optprof only runs on NonRTM

Github is showing me the following diff on the files tab, but for some reason not on the conversation tab:

```diff
-enabled: "and(succeeded(), eq(variables['BuildRTM'], 'false'), eq(variables['IsOfficialBuild'], 'true'))"
+enabled: ${{ variables['isOfficialBuild'] }}
```

So previously it would only run on the non-RTM job, and only on official builds.

Now, it looks like it's running on official builds for any job.

However, while gathering evidence, I now realize this is part of the nonRTM job defined on line 87, whereas the RTM job definition starts on 271.  When I originally split up the jobs into different files, I intended the individual yamls to contain "all" the steps, including publishing, so now that pipeline.yml has all the outputs, it's completely thrown off my mental model for how the CI jobs are defined.

zivkan (dismissed):

eng/pipelines/templates/pipeline.yml
donnie-msft: Discussed offline. No condition needed.

zivkan (approved):

PR #5725, opened 2024-03-27
[release-6.8.x] Use 1ES Templates in CI Pipelines

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2755

## Description

Cherry-pick of https://github.com/NuGet/NuGet.Client/pull/5706

[Sample Official Build](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=9328610&view=results)
[Sample OptProf Run](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=9328611&view=results)

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


eng/pipelines/templates/Build_and_UnitTest.yml
```diff
@@ -115,9 +86,6 @@ steps:
         exit 1
       }
 
-- task: ComponentGovernanceComponentDetection@0
```
zivkan, line 118: I didn't notice this getting removed from the dev branch. The "problem" with not running it explictly in our pipeline is that it runs at the end of the pipeline, after tests have created a whole lot of test nupkgs, and then CG complains about packages with no licenses, and other issues..

Also, since the NonRTM pipeline create SemVer pre-release versions of our packages, if the latest release was a security fix, then CG will complain that it found vulnerable versions of our packages, even though the package was created in the pipeline. We can get around this by always incrementing config.prop's version after a fix ships, but that's a manual step, and change in process that we chose not to do a few months ago when I added in this CG step in a carefuly chosen location in the pipeline.

Do we no longer have control when the CG steps runs in our pipeline?

zivkan (approved):

eng/pipelines/templates/Build_and_UnitTest.yml
donnie-msft: Yes, it was requested that I remove it and did so in dev here: https://github.com/NuGet/NuGet.Client/pull/5706/commits/9336d020b01658afbfc8faaa5817dc655bb0e899

I'll see if I can configure the 1ES step. If not, I could put our steps back in - @jeffkl thoughts?

donnie-msft: https://github.com/NuGet/Client.Engineering/issues/2765

PR #5732, opened 2024-03-27
Deprecate http: Promote from warning to error in VS Options NuGet Package Manager

Nigusu-Allehu:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13354

## Description

- PMUI shows an http warning when an http source is used. Show an error instead.
- Show warning for a source if `allowInsecureConnections` attribute is enabled.
- Allow users to add non-https sources after requesting user to confirm
- Link to explanation how to work with NuGet configs
### Here is what the PMUI would look like

https://github.com/NuGet/NuGet.Client/assets/59111203/55a7528b-b62b-476c-bd04-4dbe56ae3010

### Fixed Padding
image

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Options/CheckedListBoxAccessibleObject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Options/CheckedListBoxAccessibleObject.cs
@@ -37,7 +37,7 @@ namespace NuGet.PackageManagement.UI.Options
                 if (packageSource.IsHttp && !packageSource.IsHttps && !packageSource.AllowInsecureConnections)
                 {
                     var sourceMessage = string.Concat(
-                        Resources.Warning_HTTPSource,
+                        Resources.Error_HttpSource_Single,
                         packageSource.Source);
                     return new CheckedListBoxItemAccessibleObject(this, packageSource.Name, index, sourceMessage);
                 }
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Options/CheckedListBoxAccessibleObject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Options/CheckedListBoxAccessibleObject.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Globalization;
 using System.Windows.Forms;
 using NuGet.Configuration;
 using NuGet.VisualStudio.Internal.Contracts;
@@ -36,7 +37,8 @@ namespace NuGet.PackageManagement.UI.Options
                 PackageSource packageSource = new PackageSource(item.Source, item.Name);
                 if (packageSource.IsHttp && !packageSource.IsHttps && !packageSource.AllowInsecureConnections)
                 {
-                    var sourceMessage = string.Concat(
+                    var sourceMessage = string.Format(
+                        CultureInfo.CurrentCulture,
                         Resources.Error_HttpSource_Single,
                         packageSource.Source);
                     return new CheckedListBoxItemAccessibleObject(this, packageSource.Name, index, sourceMessage);
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Options/CheckedListBoxAccessibleObject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Options/CheckedListBoxAccessibleObject.cs
@@ -44,6 +44,15 @@ namespace NuGet.PackageManagement.UI.Options
                     return new CheckedListBoxItemAccessibleObject(this, packageSource.Name, index, sourceMessage);
                 }
 
+                if (packageSource.AllowInsecureConnections)
+                {
+                    var insecureConnectionMessage = string.Format(
+                        CultureInfo.CurrentCulture,
+                        Resources.Warning_HTTPSource,
+                        packageSource.Source);
+                    return new CheckedListBoxItemAccessibleObject(this, packageSource.Name, index, insecureConnectionMessage);
+                }
+
                 return new CheckedListBoxItemAccessibleObject(this, packageSource.Name, index, packageSource.Source);
             }
             else
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Options/PackageSourceCheckedListBox.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Options/PackageSourceCheckedListBox.cs
@@ -20,13 +20,13 @@ namespace NuGet.PackageManagement.UI.Options
     {
         public Size CheckBoxSize { get; set; }
 
-        private static Icon WarningIcon { get; set; }
+        private static Icon ErrorIcon { get; set; }
 
-        private Icon GetWarningIcon()
+        private Icon GetErrorIcon()
         {
             ThreadHelper.ThrowIfNotOnUIThread();
 
-            if (WarningIcon == null)
+            if (ErrorIcon == null)
             {
                 ImageAttributes attributes = new ImageAttributes
                 {
@@ -39,12 +39,12 @@ namespace NuGet.PackageManagement.UI.Options
                 };
 
                 IVsImageService2 imageService = (IVsImageService2)Microsoft.VisualStudio.Shell.Package.GetGlobalService(typeof(SVsImageService));
-                IVsUIObject uIObj = imageService.GetImage(KnownMonikers.StatusWarning, attributes);
+                IVsUIObject uIObj = imageService.GetImage(KnownMonikers.StatusError, attributes);
 
-                WarningIcon = (Icon)GelUtilities.GetObjectData(uIObj);
+                ErrorIcon = (Icon)GelUtilities.GetObjectData(uIObj);
             }
 
-            return WarningIcon;
+            return ErrorIcon;
         }
 
         public override int ItemHeight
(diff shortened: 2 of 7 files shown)


nkolev92:
Can you upload a screenshot?

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/PackageSourceCheckedListBox.cs
```diff
@@ -137,23 +137,23 @@ protected override void OnDrawItem(DrawItemEventArgs e)
 
                         var packageSource = new PackageSource(currentItem.Source, currentItem.Name);
                         packageSource.AllowInsecureConnections = currentItem.AllowInsecureConnections;
-                        var shouldShowHttpWarningIcon = packageSource.IsHttp && !packageSource.IsHttps && !packageSource.AllowInsecureConnections;
-                        Rectangle warningBounds = default;
+                        var shouldShowHttpErrorIcon = packageSource.IsHttp && !packageSource.IsHttps && !packageSource.AllowInsecureConnections;
```
zivkan, line 167: Do we want to keep showing the warning icon when `allowInsecureConnections` is set to true? Our VS package sources options page doesn't show the cusotmer that `allowInsecureConnections` is set, so at least a warning would be a push to customers to start asking why they can't use secure communications, even if it's "internal" to their company.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/CheckedListBoxAccessibleObject.cs
```diff
@@ -36,8 +37,9 @@ public override AccessibleObject GetChild(int index)
                 PackageSource packageSource = new PackageSource(item.Source, item.Name);
                 if (packageSource.IsHttp && !packageSource.IsHttps && !packageSource.AllowInsecureConnections)
                 {
-                    var sourceMessage = string.Concat(
-                        Resources.Warning_HTTPSource,
```
zivkan, line 40: Are there any other references to `Warning_HTTPSource`?  You added a new string to the resx, but you didn't remove this one, and my intuition makes me think that nothing else would use this string.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/PackageSourceCheckedListBox.cs
zivkan, line 167: What happens if you start with a nuget.config that uses http and has `allowInsecureConnections` set to true, then you edit the URL, for example from http://contoso.test/nuget to http://contoso.test/nuget/v3/index.json?  Does the `allowInsecureConections` get lost?  What about changing http to https?  What about changing the source name?

Basically, I'm thinking about all the customer experiences around editing the source. Are customers going to complain to us, increasing our support workload?

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
```diff
@@ -1042,4 +1042,7 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
     <value>The package `{0}` is available in the Global packages folder, but the source it came from `{1}` is not one of the configured sources.</value>
     <comment>{0} is the package ID. {1} is the URI of the package source found in the Global packages folder.</comment>
   </data>
+  <data name="Error_HttpSource_Single" xml:space="preserve">
+    <value>NuGet requires HTTPS sources. To use an HTTP source, you must explicitly set 'allowInsecureConnections' to true in your NuGet.Config file.</value>
```
zivkan, line 1043: I'm not sure this is enough information for people new to .NET. They might not know where nuget.config is, how to edit it, etc. I think it would be worthwhile having a clickable link to go to docs, or just tell customers on the package sources options page that they can open the nuget.config file using the "configuration files" page that can be selected on the left (if we don't already, I think it would be good to have a link from the config files page to the nuget.config reference docs, so it's easy and quick for customers to find info, but obviously adding that link is out of scope for this PR. I'm just making a point that we should be looking for opportunities to improve customer experience)

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/PackageSourceCheckedListBox.cs
Nigusu-Allehu: Editing the URL or changing from http to https would not affect the value of `allowInsecureConnections` for the specific source. However, changing the source name would completely replace the existing source as a result the value `allowInsecureConnections` would be lost. 

I feel like we should add a way for users to see all the properties of a source without having to open the config file. For `allowInsecureConnections` property, we could do that as follows(by adding a checkbox)
image


This would allow users to know learn about the sources and update them without having to open the config file. What do you think @NuGet/nuget-client ? CC : @JonDouglas

Nigusu-Allehu: wrote a spec for this here :https://github.com/NuGet/Home/pull/13376

jeffkl: I would pop up a dialog with something like "NuGet does not recommend using HTTP sources because blah blah.  Do you want to add this HTTP source and disable protections?" with Yes/No buttons.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/CheckedListBoxAccessibleObject.cs
Nigusu-Allehu: removed now

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/PackageSourceCheckedListBox.cs
Nigusu-Allehu: I added a warning icon that would be shown when users have set the `allowInsecureConnections` attribute

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
Nigusu-Allehu: @JonDouglas what do you think?

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
```diff
@@ -933,10 +933,7 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
     <comment>{0} is a version number</comment>
   </data>
   <data name="Warning_HTTPSource" xml:space="preserve">
-    <value>Warning: Non-HTTPS access will be removed in a future version. Consider migrating to an 'HTTPS' source.</value>
-  </data>
-  <data name="Warning_NewHTTPSource_VSOptions" xml:space="preserve">
-    <value>Non-HTTPS access will be removed in a future version. Consider migrating to an 'HTTPS' source.</value>
+    <value>`AllowInsecureConnections` attribute has been enabled for this source which would allow it connecting to a NON-HTTPS source.</value>
```
Nigusu-Allehu: @JonDouglas  any suggestions on this. It is a warning shown to users when the `allowInsecureConnections` attribute is set to true for a source. See video.

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
donnie-msft: Configuration files page already has a docs link and this page could mimic that, but point to a section about HTTP sources.
image

donnie-msft (commented):
The spacing for the warning/error is a little too indented. Can you play around with the layout a bit? I believe previously the warning wasn't being pushed so far too the right. Also, be sure to test resizing the window to a smaller size.

image

Nigusu-Allehu:
> The spacing for the warning/error is a little too indented. Can you play around with the layout a bit? I believe previously the warning wasn't being pushed so far too the right. Also, be sure to test resizing the window to a smaller size.
> 
> image

Thanks. I have fixed the indentation. Look in the description. I made sure it looks like how the warning (Insecure connections) used to look except now it's an error

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/PackageSourcesOptionsControl.cs
```diff
@@ -94,8 +76,25 @@ public PackageSourcesOptionsControl(IAsyncServiceProvider asyncServiceProvider)
...
+                Flags = (uint)_ImageAttributesFlags.IAF_RequiredFlags
+            };
+
+            IVsUIObject uIObj = ImageService.GetImage(KnownMonikers.StatusWarning, attributes);
+
+            WarningIcon = (Image)GelUtilities.GetObjectData(uIObj);
+
+            uIObj = ImageService.GetImage(KnownMonikers.StatusError, attributes);
+
+            ErrorIcon = (Image)GelUtilities.GetObjectData(uIObj);
```
jebriede: nit: remove some of the blank lines between each line. A blank line can be used to group related lines of code, but if there's a blank line between each line, it is equivalent to not having the blank lines, which would be preferred.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/PackageSourcesOptionsControl.cs
```diff
@@ -483,13 +482,37 @@ private TryUpdateSourceResults TryUpdateSource()
...
 
-            selectedPackageSource.Name = name;
-            selectedPackageSource.Source = source;
-            _packageSources.ResetCurrentItem();
+            PackageSource packageSource = new PackageSource(source, name);
+
+            if (packageSource.IsHttp && !packageSource.IsHttps)
+            {
+                // Updating to an http source. Request user confirmation
+                bool? addHttpSource = MessageHelper.ShowQueryMessage(Resources.Warn_Adding_HttpSource, "HTTP Source", true);
```
jebriede: The dialog title, "HTTP Source" needs to be localized.

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/PackageSourcesOptionsControl.cs
```diff
@@ -715,21 +738,32 @@ private static bool IsPathRootedSafe(string path)
             return path.IndexOfAny(Path.GetInvalidPathChars()) == -1 && Path.IsPathRooted(path);
         }
 
-        private void SetHttpWarningVisibilityForSelectSource(PackageSourceContextInfo selectedSource)
+        private void SetHttpErrorVisibilityForSelectSource(PackageSourceContextInfo selectedSource)
```
jebriede: Typo in method name.

```suggestion
        private void SetHttpErrorVisibilityForSelectedSource(PackageSourceContextInfo selectedSource)
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Options/PackageSourcesOptionsControl.cs
```diff
@@ -715,21 +738,32 @@ private static bool IsPathRootedSafe(string path)
...
         {
             var source = new PackageSource(selectedSource.Source, selectedSource.Name);
             source.AllowInsecureConnections = selectedSource.AllowInsecureConnections;
 
-            // Warn if the selected source is http, and the AllowInsecureConnections for this source is set to false. 
+            HttpErrorOrWarning.Visible = true;
+            HttpErrorOrWarningIcon.Visible = true;
+            configurationLink.Visible = true;
```
jebriede: By setting these to true before checking if the source is http, not http, and insecure connections are not allowed, does the error flash on the screen before reaching the else case where they're set to false? I prefer to err on the side of only showing error messages when we've confirmed the state is an error state. That way, on slow or busy machines, you wouldn't see the errors show visible then disappear a brief moment later.

My suggestion then would be to not modify the current visibility state until after we've confirmed there's an error or not to prevent a potential blinking effect.

jebriede (changes requested):
Looks great! Main change requested is for the localization of the dialog title. Others are more optional.

jebriede (approved):

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
JonDouglas: `AllowInsecureConnections` is enabled and allows http access. Note: This method is not secure. For secure options, see https://aka.ms/nuget-https-everywhere.

PR #5735, opened 2024-04-01
Update frameworks for unit and functional tests

### LINKED ISSUE #8452: Remove .NET Framework TFM from NuGet.CommandLine.XPlat
NuGet.CommandLine.XPlat is intended to be used only by the dotnet cli, although searching github we can find multiple customers who use it and call our `Program.Main()` in their code. I can't verify that all existing usage is used by netcoreapp projects, so we'll wait until the next major release before removing the .NET Framework TFM.

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2747
Fixes: https://github.com/NuGet/Home/issues/8452

## Description

This updates the target frameworks that are unit and functional tests target to ensure we're testing the correct runtime.  It also installs a different version of the .NET SDK for the functional tests to patch so that we're testing unreleased bits in an effort to find issues before public releases.

Previously, our tests were targeting a mixture of .NET Core 3.1, .NET 5.0, .NET 7.0, and .NET 8.0.  Ideally, all of our class libraries should target .NET Framework v4.7.2 and .NET Standard 2.0.  However, some of our libraries need to use cryptograph APIs that are not available in .NET Standard.  Our solution to this problem is to have those libraries target .NET Framework v4.7.2, .NET Standard 2.0, and .NET 5.0.  When we build those libraries for .NET Standard, all of the calls to cryptography APIs are #ifdef'd out but when we build for .NET 5.0, those calls are included in the assemblies.  To test these libraries, our tests need to target .NET Framework v4.7.2, .NET Core 3.1, and .NET 8.0.

<table>
  <tr><th /><th colspan="3" align="center">Library</td></tr>
  <tr><th>Test</th><th>net472</th><th>netstandard2.0</th><th>net5.0</th></tr>
  <tr><td>net472</td><td align="center">✅</td><td/><td/></tr>
  <tr><td>netcoreapp3.1</td><td/><td align="center">✅</td><td/></tr>
  <tr><td>net8.0</td><td/><td/><td align="center">✅</td></tr>
</table>

So for any given library, it will target either `net472`, and `netstandard2.0`, or `net472`, `netstandard2.0`, and `net5.0`.  Corresponding unit tests need to target the correct runtime to test the correct library.  The reason the test targets `netcoreapp3.1`, is so that at runtime the `netstandard2.0` library is loaded instead of the `net5.0` or `.net8.0` one.

I also added a property so that some runnable apps like `NuGet.Build.Tasks.Console.exe` can target `net8.0`.  This results in a few main properties:

| Property | Value |
|---|---|
| `NETCoreTargetFramework` | `net8.0` |
| `NETCoreLegacyTargetFramework` | `netcoreapp3.1` |
| `NETCoreLegacyTargetFrameworkForSigning` | `net5.0` |

So anything that's a runnable app, we try to target `net8.0` otherwise we target `netcoreapp3.1` unless cryptography APIs are needed in which case we target `net5.0`.

For multi-targeting libraries and EXEs, we target the following:

| Property | Value | Purpose |
|---|---|---|
| `TargetFrameworksLibrary` | `net472;netstandard2.0` | Class library |
| `TargetFrameworksLibraryForSigning` | `net472;netstandard2.0;net5.0` | Class library that needs cryptography APIs |
| `TargetFrameworksExe` | `net472;netcoreapp3.1` | Runnable app |

Then for tests we have:
| Property | Value | Purpose |
|---|---|---|
| `TargetFrameworksUnitTest` | `net472;net8.0` | Tests for libraries that only target `net472` and `netstandard2.0` |
| `TargetFrameworksUnitTestForSigning` | `net472;8.0;net5.0` | Tests for libraries that need cryptography APIs and build target `net472`, `netstandard2.0`, and `net5.0`  |

I also have all of the logic so when building on Linux and Mac, our build won't target `net472` since we're not running tests against that framework on that platform. 

With this change, we only need to install the .NET 8 SDK to build and the .NET Core 3.1 runtime to run some tests which reduces the overhead of our configuration scripts.

Other changes of note:
1. The .NET SDK we install for the functional tests is for the `8.0.3xx` channel and `daily` quality.  This is specified in `build\DotNetSdkTestVersions.txt`.  When this is branched for a release, we'll want to update this value so that any particular branch is always testing the corresponding release of the .NET SDK.
2. The .NET SDK that we install for functional tests is placed under `.test\dotnet` and I updated the configuration scripts for Windows, Linux, and Mac to install it there.
3. I updated the pipeline YAML for the cross framework tests to only build the one project needed so it speeds things up and also removed the execution of .NET 7.0 since we don't support that anymore
4. I updated every project so that its using the correct value for `<TargetFramework>` or `<TargetFrameworks>`
5. I added a target for projects that need to copy files so that they support incremental builds and retriable file copies.
6. I removed any `<Service />` items in projects since they are no longer needed.
7. I cleaned up any obvious extra `<Reference />` items while I was in there.
8. I updated any places with `CopyToOutputDirectory=Always` to use `CopyToOutputDirectory=PreserveNewest` since copying always breaks incremental builds.
9. I had to update some project templates we were testing since some of them were deprecated in .NET 8.0
10. Some tests now fail on Mac, I'm assuming because they are now running on .NET 8.0 so I updated them to only run on Windows and Linux.
11. Updated the logic of finding the .NET SDK for testing to look in `.test\dotnet`.
12. I had to fix some nullable checks now that some tests target newer frameworks.
13. Updated some code paths that were using the compiler constants.
14. Fixed a bug in CommandRunner.cs that would mean that a timed out process would not actually exit.  Instead, an exception was being thrown and we were perpetually waitiing on `Process.WaitForExit()`.  This would lead to a timeout in the pipeline and no logs.
15. Updated some tools to use `$(NETCoreTargetFramework)` instead of having a hard-coded value of `net7.0` so we don't need older runtimes to run these tools.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/build/DotNetSdkTestVersions.txt
+++ b/build/DotNetSdkTestVersions.txt
@@ -1,3 +1,2 @@
 # Each line represents arguments for the .NET SDK installer script (https://learn.microsoft.com/dotnet/core/tools/dotnet-install-script)
--Channel 8.0
--Channel 3.1 -Runtime dotnet
+-Channel 8.0.3xx -Quality daily
--- a/build/common.project.props
+++ b/build/common.project.props
@@ -25,18 +25,10 @@
     <TargetFrameworksLibraryForSigning>$(TargetFrameworksLibrary)</TargetFrameworksLibraryForSigning>
     <TargetFrameworksLibraryForSigning Condition="'$(DotNetBuildFromSource)' != 'true'">$(TargetFrameworksLibraryForSigning);$(NETCoreLegacyTargetFrameworkForSigning)</TargetFrameworksLibraryForSigning>
     
-    <!-- Target framework for .NET Core runnable apps -->
-    <TargetFrameworksExe>$(NETCoreLegacyTargetFramework)</TargetFrameworksExe>
+    <!-- Target framework for runnable apps -->
+    <TargetFrameworksExe>$(NETCoreTargetFramework)</TargetFrameworksExe>
     <TargetFrameworksExe Condition="'$(IsXPlat)' != 'true' And '$(DotNetBuildFromSource)' != 'true'">$(NETFXTargetFramework);$(TargetFrameworksExe)</TargetFrameworksExe>
 
-    <!-- Target frameworks for class libraries which require signing APIs which need to target NET 5.0 -->
-    <TargetFrameworksExeForSigning>$(TargetFrameworksExe)</TargetFrameworksExeForSigning>
-    <TargetFrameworksExeForSigning Condition="'$(IsXPlat)' != 'true' And '$(DotNetBuildFromSource)' != 'true'">$(TargetFrameworksExeForSigning);$(NETCoreLegacyTargetFrameworkForSigning)</TargetFrameworksExeForSigning>
-
-    <!-- Target frameworks for applications that don't need to use modern frameworks -->
-    <MinimalTargetFrameworksExe>$(NETCoreTargetFramework)</MinimalTargetFrameworksExe>
-    <MinimalTargetFrameworksExe Condition="'$(IsXPlat)' != 'true' And '$(DotNetBuildFromSource)' != 'true'">$(NETFXTargetFramework);$(MinimalTargetFrameworksExe)</MinimalTargetFrameworksExe>
-    
     <!-- Target frameworks for unit tests -->
     <TargetFrameworksUnitTest>$(NETCoreTargetFramework)</TargetFrameworksUnitTest>
     <TargetFrameworksUnitTest Condition="'$(IsXPlat)' != 'true' And '$(DotNetBuildFromSource)' != 'true'">$(NETFXTargetFramework);$(TargetFrameworksUnitTest)</TargetFrameworksUnitTest>
@@ -44,16 +36,10 @@
     <!-- Target frameworks for unit tests that test libaries using signing APIs -->
     <TargetFrameworksUnitTestForSigning>$(NETCoreTargetFramework)</TargetFrameworksUnitTestForSigning>
     <TargetFrameworksUnitTestForSigning Condition="'$(IsXPlat)' != 'true' And '$(DotNetBuildFromSource)' != 'true'">$(NETFXTargetFramework);$(TargetFrameworksUnitTestForSigning);$(NETCoreLegacyTargetFramework)</TargetFrameworksUnitTestForSigning>
-
-    <!-- Target framework for functional tests -->
-    <TargetFrameworksFunctionalTest>net8.0</TargetFrameworksFunctionalTest>
-    <TargetFrameworksFunctionalTest Condition="'$(IsXPlat)' != 'true' And '$(DotNetBuildFromSource)' != 'true'">$(NETFXTargetFramework);$(TargetFrameworksFunctionalTest)</TargetFrameworksFunctionalTest>
-  
   </PropertyGroup>
 
   <!-- Common -->
   <PropertyGroup>
-    <RestoreProjectStyle>PackageReference</RestoreProjectStyle>
     <RepositoryRootDirectory>$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), 'README.md'))\</RepositoryRootDirectory>
     <BuildCommonDirectory>$(RepositoryRootDirectory)build\</BuildCommonDirectory>
     <SolutionFile>$(RepositoryRootDirectory)$(RepositoryName).sln</SolutionFile>
--- a/build/common.project.props
+++ b/build/common.project.props
@@ -6,28 +6,54 @@
     <IsXPlat Condition=" $(MSBuildProjectFullPath.StartsWith('/')) == 'true' OR $(MSBuildProjectFullPath.StartsWith('\')) == 'true' ">true</IsXPlat>
   </PropertyGroup>
 
+  <!-- Target Frameworks -->
+  <PropertyGroup>
+    <NETFXTargetFrameworkVersion>v4.7.2</NETFXTargetFrameworkVersion>
+    <NETFXTargetFramework>net472</NETFXTargetFramework>
+    <NetStandardVersion>netstandard2.0</NetStandardVersion>
+    <NETCoreTargetFramework>net8.0</NETCoreTargetFramework>
+    <NETCoreTargetFramework Condition="'$(DotNetBuildFromSource)' == 'true'">net9.0</NETCoreTargetFramework>
+    <NETCoreLegacyTargetFramework>netcoreapp3.1</NETCoreLegacyTargetFramework>
+    <NETCoreLegacyTargetFramework Condition="'$(DotNetBuildFromSource)' == 'true'">net9.0</NETCoreLegacyTargetFramework>
+    <NETCoreLegacyTargetFrameworkForSigning>netcoreapp5.0</NETCoreLegacyTargetFrameworkForSigning>
+    
+    <!-- Target frameworks for class libraries-->
+    <TargetFrameworksLibrary>$(NetStandardVersion)</TargetFrameworksLibrary>
+    <TargetFrameworksLibrary Condition="'$(IsXPlat)' != 'true' And '$(DotNetBuildFromSource)' != 'true'">$(NETFXTargetFramework);$(TargetFrameworksLibrary)</TargetFrameworksLibrary>
+
+    <!-- Target frameworks for class libraries which require signing APIs which need to target NET 5.0 -->
+    <TargetFrameworksLibraryForSigning>$(TargetFrameworksLibrary)</TargetFrameworksLibraryForSigning>
+    <TargetFrameworksLibraryForSigning Condition="'$(DotNetBuildFromSource)' != 'true'">$(TargetFrameworksLibraryForSigning);$(NETCoreLegacyTargetFrameworkForSigning)</TargetFrameworksLibraryForSigning>
+    
+    <!-- Target framework for .NET Core runnable apps -->
+    <TargetFrameworksExe>$(NETCoreLegacyTargetFramework)</TargetFrameworksExe>
+    <TargetFrameworksExe Condition="'$(IsXPlat)' != 'true' And '$(DotNetBuildFromSource)' != 'true'">$(NETFXTargetFramework);$(TargetFrameworksExe)</TargetFrameworksExe>
+
+    <!-- Target frameworks for class libraries which require signing APIs which need to target NET 5.0 -->
+    <TargetFrameworksExeForSigning>$(TargetFrameworksExe)</TargetFrameworksExeForSigning>
+    <TargetFrameworksExeForSigning Condition="'$(IsXPlat)' != 'true' And '$(DotNetBuildFromSource)' != 'true'">$(TargetFrameworksExeForSigning);$(NETCoreLegacyTargetFrameworkForSigning)</TargetFrameworksExeForSigning>
+
+    <!-- Target frameworks for applications that don't need to use modern frameworks -->
+    <MinimalTargetFrameworksExe>$(NETCoreTargetFramework)</MinimalTargetFrameworksExe>
+    <MinimalTargetFrameworksExe Condition="'$(IsXPlat)' != 'true' And '$(DotNetBuildFromSource)' != 'true'">$(NETFXTargetFramework);$(MinimalTargetFrameworksExe)</MinimalTargetFrameworksExe>
+    
+    <!-- Target frameworks for unit tests -->
+    <TargetFrameworksUnitTest>$(NETCoreTargetFramework)</TargetFrameworksUnitTest>
+    <TargetFrameworksUnitTest Condition="'$(IsXPlat)' != 'true' And '$(DotNetBuildFromSource)' != 'true'">$(NETFXTargetFramework);$(TargetFrameworksUnitTest)</TargetFrameworksUnitTest>
+
+    <!-- Target frameworks for unit tests that test libaries using signing APIs -->
+    <TargetFrameworksUnitTestForSigning>$(NETCoreTargetFramework)</TargetFrameworksUnitTestForSigning>
+    <TargetFrameworksUnitTestForSigning Condition="'$(IsXPlat)' != 'true' And '$(DotNetBuildFromSource)' != 'true'">$(NETFXTargetFramework);$(TargetFrameworksUnitTestForSigning);$(NETCoreLegacyTargetFramework)</TargetFrameworksUnitTestForSigning>
+
+    <!-- Target framework for functional tests -->
+    <TargetFrameworksFunctionalTest>net8.0</TargetFrameworksFunctionalTest>
+    <TargetFrameworksFunctionalTest Condition="'$(IsXPlat)' != 'true' And '$(DotNetBuildFromSource)' != 'true'">$(NETFXTargetFramework);$(TargetFrameworksFunctionalTest)</TargetFrameworksFunctionalTest>
+  
+  </PropertyGroup>
+
   <!-- Common -->
   <PropertyGroup>
     <RestoreProjectStyle>PackageReference</RestoreProjectStyle>
-    <NETFXTargetFrameworkVersion>v4.7.2</NETFXTargetFrameworkVersion>
-    <NETFXTargetFramework>net472</NETFXTargetFramework>
-    <NETCoreTargetFramework>netcoreapp3.1</NETCoreTargetFramework>
-    <NETCoreTargetFramework Condition="'$(DotNetBuildFromSource)' == 'true'">net9.0</NETCoreTargetFramework>
-    <NETCoreTestTargetFrameworks>net8.0</NETCoreTestTargetFrameworks>
-    <NetStandardVersion>netstandard2.0</NetStandardVersion>
-    <TargetFrameworksExe>$(NETFXTargetFramework);$(NETCoreTargetFramework)</TargetFrameworksExe>
-    <TargetFrameworksExe Condition="'$(IsXPlat)' == 'true'">$(NETCoreTargetFramework)</TargetFrameworksExe>
-    <TargetFrameworksExeForSigning>$(TargetFrameworksExe);netcoreapp5.0</TargetFrameworksExeForSigning>
-    <TargetFrameworksExeForSigning Condition=" '$(IsXPlat)' == 'true' ">$(NETCoreTargetFramework);netcoreapp5.0</TargetFrameworksExeForSigning>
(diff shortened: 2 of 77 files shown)


test/NuGet.Clients.Tests/NuGet.CommandLine.Test/NuGet.CommandLine.Test.csproj
```diff
@@ -55,8 +52,23 @@
...
+          AfterTargets="PrepareForRun"
+          Inputs="$(ArtifactsDirectory)$(VsixOutputDirName)\NuGet.exe"
+          Outputs="$(OutputPath)NuGet\NuGet.exe"
+          Condition="'$(SkipILMergeOfNuGetExe)' != 'true'">
     <Copy SourceFiles="$(ArtifactsDirectory)$(VsixOutputDirName)\NuGet.exe"
-          DestinationFolder="$(OutputPath)NuGet\" />
+          DestinationFiles="$(OutputPath)NuGet\NuGet.exe"
+          SkipUnchangedFiles="$(SkipCopyUnchangedFiles)"
+          OverwriteReadOnlyFiles="$(OverwriteReadOnlyFiles)"
+          Retries="$(CopyRetryCount)"
```
donnie-msft, line 64: It looks like most but not all of the options have been provided. How did you determine which ones to add in this PR?

build/common.targets
```diff
@@ -16,7 +16,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <SigningNotSupported Condition=" '$(TargetFramework)' == 'netstandard2.0' OR '$(TargetFramework)' == 'netcoreapp2.1' OR '$(TargetFramework)' == 'netstandard2.1' OR '$(TargetFramework)' == 'netcoreapp3.1'">true</SigningNotSupported>
+    <SigningNotSupported Condition=" '$(TargetFramework)' == '$(NetStandardVersion)' Or '$(TargetFramework)' == '$(NETCoreLegacyTargetFramework)'">true</SigningNotSupported>
```
donnie-msft:
```suggestion
    <SigningNotSupported Condition=" '$(TargetFramework)' == '$(NetStandardVersion)' OR '$(TargetFramework)' == '$(NETCoreLegacyTargetFramework)'">true</SigningNotSupported>
```

donnie-msft (dismissed):
Thanks for the detailed writeup of what all this does. Great work! This looks like it took a lot of work but improves several aspects of the build and tests.

build/DotNetSdkTestVersions.txt
zivkan, line 1: This is only used by Dotnet.Integration.Tests, right?  I find it confusing because the file name sounds like all tests will use this, hence I was surprised that it doesn't include 3.1, to test the netstandard 2.0 binaries.

Therefore, I suggest that `DotnetIntegrationTestSdkVersions.txt` would be a more clear name.

build/common.project.props
```diff
@@ -6,28 +6,54 @@
...
+    <NETCoreLegacyTargetFramework>netcoreapp3.1</NETCoreLegacyTargetFramework>
+    <NETCoreLegacyTargetFramework Condition="'$(DotNetBuildFromSource)' == 'true'">net9.0</NETCoreLegacyTargetFramework>
+    <NETCoreLegacyTargetFrameworkForSigning>netcoreapp5.0</NETCoreLegacyTargetFrameworkForSigning>
+    
+    <!-- Target frameworks for class libraries-->
+    <TargetFrameworksLibrary>$(NetStandardVersion)</TargetFrameworksLibrary>
+    <TargetFrameworksLibrary Condition="'$(IsXPlat)' != 'true' And '$(DotNetBuildFromSource)' != 'true'">$(NETFXTargetFramework);$(TargetFrameworksLibrary)</TargetFrameworksLibrary>
+
+    <!-- Target frameworks for class libraries which require signing APIs which need to target NET 5.0 -->
+    <TargetFrameworksLibraryForSigning>$(TargetFrameworksLibrary)</TargetFrameworksLibraryForSigning>
```
zivkan, line 25: Would it be worthwhile changing the name to `WithSigning`, rather than `ForSigning`?  The word signing has different meanings in different contexts, and if you don't know the context in which this property is being used, the name might be misleading.

build/common.project.props
```diff
@@ -6,28 +6,54 @@
...
+    <TargetFrameworksLibraryForSigning Condition="'$(DotNetBuildFromSource)' != 'true'">$(TargetFrameworksLibraryForSigning);$(NETCoreLegacyTargetFrameworkForSigning)</TargetFrameworksLibraryForSigning>
+    
+    <!-- Target framework for .NET Core runnable apps -->
+    <TargetFrameworksExe>$(NETCoreLegacyTargetFramework)</TargetFrameworksExe>
```
zivkan:
```suggestion
    <TargetFrameworksExe>$(NETCoreTargetFramework)</TargetFrameworksExe>
```

build/common.project.props
```diff
@@ -6,28 +6,54 @@
...
+
+    <!-- Target framework for functional tests -->
+    <TargetFrameworksFunctionalTest>net8.0</TargetFrameworksFunctionalTest>
+    <TargetFrameworksFunctionalTest Condition="'$(IsXPlat)' != 'true' And '$(DotNetBuildFromSource)' != 'true'">$(NETFXTargetFramework);$(TargetFrameworksFunctionalTest)</TargetFrameworksFunctionalTest>
+  
+  </PropertyGroup>
+
+  <!-- Common -->
+  <PropertyGroup>
+    <RestoreProjectStyle>PackageReference</RestoreProjectStyle>
```
zivkan: All our projects are SDK style now. We can delete this.

build/common.project.props
```diff
@@ -6,28 +6,54 @@
...
+    
+    <!-- Target framework for .NET Core runnable apps -->
+    <TargetFrameworksExe>$(NETCoreLegacyTargetFramework)</TargetFrameworksExe>
+    <TargetFrameworksExe Condition="'$(IsXPlat)' != 'true' And '$(DotNetBuildFromSource)' != 'true'">$(NETFXTargetFramework);$(TargetFrameworksExe)</TargetFrameworksExe>
+
+    <!-- Target frameworks for class libraries which require signing APIs which need to target NET 5.0 -->
+    <TargetFrameworksExeForSigning>$(TargetFrameworksExe)</TargetFrameworksExeForSigning>
+    <TargetFrameworksExeForSigning Condition="'$(IsXPlat)' != 'true' And '$(DotNetBuildFromSource)' != 'true'">$(TargetFrameworksExeForSigning);$(NETCoreLegacyTargetFrameworkForSigning)</TargetFrameworksExeForSigning>
+
+    <!-- Target frameworks for applications that don't need to use modern frameworks -->
```
zivkan:
```suggestion
    <!-- Target frameworks for applications that don't need to use legacy runtimes -->
```

build/common.project.props
```diff
@@ -6,28 +6,54 @@
...
+    <!-- Target frameworks for unit tests -->
+    <TargetFrameworksUnitTest>$(NETCoreTargetFramework)</TargetFrameworksUnitTest>
+    <TargetFrameworksUnitTest Condition="'$(IsXPlat)' != 'true' And '$(DotNetBuildFromSource)' != 'true'">$(NETFXTargetFramework);$(TargetFrameworksUnitTest)</TargetFrameworksUnitTest>
+
+    <!-- Target frameworks for unit tests that test libaries using signing APIs -->
+    <TargetFrameworksUnitTestForSigning>$(NETCoreTargetFramework)</TargetFrameworksUnitTestForSigning>
+    <TargetFrameworksUnitTestForSigning Condition="'$(IsXPlat)' != 'true' And '$(DotNetBuildFromSource)' != 'true'">$(NETFXTargetFramework);$(TargetFrameworksUnitTestForSigning);$(NETCoreLegacyTargetFramework)</TargetFrameworksUnitTestForSigning>
+
+    <!-- Target framework for functional tests -->
+    <TargetFrameworksFunctionalTest>net8.0</TargetFrameworksFunctionalTest>
```
zivkan:
```suggestion
    <TargetFrameworksFunctionalTest>$(NETCoreTargetFramework)</TargetFrameworksFunctionalTest>
```

build/common.targets
```diff
@@ -16,7 +16,7 @@
...
     <SigningNotSupported Condition=" '$(SigningNotSupported)' != 'true'">false</SigningNotSupported>
```
zivkan: Can we take the opportunity to make it `SigningSupported`, rather than `SigningNotSupported`?  Double negatives lead to confusion.

eng/pipelines/templates/CrossFramework_Tests_On_Windows.yml
```diff
@@ -42,16 +41,6 @@ steps:
     arguments: '--no-restore --no-build --framework net472 --settings:$(Build.Repository.LocalPath)\build\xunit.runsettings /noautorsp /property:Configuration=$(BuildConfiguration) "/binarylogger:$(Build.StagingDirectory)\\binlog\\03.Test-net472.binlog"'
     testRunTitle: 'NuGet.Client Cross Verify Tests On Windows (.NET Framework)'
 
-- task: DotNetCoreCLI@2
-  displayName: "Run Cross Verify Tests (.NET 7.0)"
-  continueOnError: ${{ eq(variables['IsOfficialBuild'], 'true') }}
-  inputs:
-    command: 'test'
-    projects: '$(Build.Repository.LocalPath)\test\NuGet.Core.FuncTests\NuGet.Signing.CrossFramework.Test\NuGet.Signing.CrossFramework.Test.csproj'
-    arguments: '--no-restore --no-build --framework net7.0 --settings:$(Build.Repository.LocalPath)\build\xunit.runsettings /noautorsp /property:Configuration=$(BuildConfiguration) "/binarylogger:$(Build.StagingDirectory)\\binlog\\03.Test-net7.0.binlog"'
```
zivkan, line 51: Do you know the reason why we run one TFM at a time, rather than just `dotnet test the.csproj` and let it automatically detect and run all the project TFMs?

src/NuGet.Core/NuGet.Configuration/NuGet.Configuration.csproj
```diff
@@ -27,7 +21,7 @@
     <Reference Include="System.Xml.Linq" />
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(TargetFramework)' == '$(NetStandardVersion)' or '$(TargetFramework)' == 'net7.0' or '$(TargetFramework)' == 'net8.0' or '$(TargetFramework)' == 'net9.0' ">
+  <ItemGroup Condition=" '$(TargetFramework)' != '$(NETFXTargetFramework)' ">
     <PackageReference Include="System.Security.Cryptography.ProtectedData" />
```
zivkan: I don't know if it works on the item group, but if the condition was on the PackageReferene item itself, we could use `'$(TargetFrameworkIdentifier)' == '.NETFramework'`, which in my opinion makes it more obvious it's not specific to one particular .NET Framework version, but all of them.

build/common.ps1
```diff
@@ -219,6 +228,44 @@ Function Install-DotnetCLI {
...
+        if ($SdkItem.StartsWith("#") -or $SdkItem.Equals("")) {
+            continue
+        }
+
+        if ([Environment]::Is64BitOperatingSystem) {
+            $arch = "x64";
+        }
+        else {
+            $arch = "x86";
+        }
```
zivkan, line 250: Doesn't `dotnet-install.ps1` detect this automatically when `-Architeture` is not provided?

Or do you think this is designed to intentionally avoid arm64 binaries if someone clones our repo on an arm64 machine? (In which case a comment would be useful)

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSignTests.cs
```diff
@@ -36,7 +36,7 @@ public DotnetSignTests(DotnetIntegrationTestFixture buildFixture, SignCommandTes
             _signFixture.SetFallbackCertificateBundles(buildFixture.SdkDirectory);
         }
 
-        [Fact]
+        [PlatformFact(Platform.Windows, Platform.Linux)]
```
zivkan: Is this because of test failures that you mentioned in the PR description?  If so, a comment with a link to a tracking useful would be helpful, to increase the chance that we can unskip the tests on Mac at a later date.

test/NuGet.Core.Tests/NuGet.Packaging.Test/PackageExtraction/ZipArchiveExtensionsTests.cs
```diff
@@ -63,7 +63,7 @@ public void UpdateFileTimeFromEntry_FileBusyForShortTime_Retries()
                 File.Delete(tempFile);
             }
         }
-
+#if !NET8_0_OR_GREATER
         // Trying to change a file timestamp when the file is open only throws on Windows
         [PlatformFact(Platform.Windows)]
         public async Task UpdateFileTimeFromEntry_FileBusyForLongTime_Throws()
```
zivkan, line 69: Do you remember why this test fails on .NET 8?

build/DotNetSdkTestVersions.txt
jeffkl: Yeah good point, when I originally started, I thought that it was used by more projects.  But as far as I can tell, its only used by this one.  Do you think it makes sense to move this file to the directory containing the project?

Also, since we'll already have bootstrapped a version of the .NET SDK to the machine, the installation for this one used by tests _could_ be done via MSBuild logic.  Do you think that would be a good change?  It would essentially mean that building the `Dotnet.Integration.Tests` project would go and get what it needs at that time instead of in `configure.ps1`.

build/common.project.props
jeffkl: It took me a long time to understand that the point of this is that a particular library needs cryptography APIs not available in .NET Standard. Maybe the name should be something like: `TargetFrameworksLibraryThatUsesCryptoAPIs` ?  I tried to address this confusion with a comment but its hard to explain.  I'm fine with whatever naming convention that you think expresses the intent best.

build/common.project.props
jeffkl: I just realized that now `TargetFrameworksExe` isn't even used.  I've gotten rid of both `TargetFrameworksExeForSigning` and `MinimalTargetFrameworksExe` because they confuse me and we don't really need all three.

build/common.targets
jeffkl: I personally prefer the proper case syntax of `And` and `Or` instead of `AND` and `OR`.  What is everyone's opinion?

eng/pipelines/templates/CrossFramework_Tests_On_Windows.yml
jeffkl: This is for reporting the results as separate runs which is my personal preference.  That way its obvious which framework a failure is happening on.

image

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/NuGet.CommandLine.Test.csproj
jeffkl: I just copy/pasted what MSBuild/.NET SDK uses when copying the assembly from the intermediate directory to the output directory.

https://github.com/dotnet/msbuild/blob/8fa24d94b5a0a0b164e04ea3eeb8fd102584ee5b/src/Tasks/Microsoft.Common.CurrentVersion.targets#L4806

test/NuGet.Core.Tests/NuGet.Packaging.Test/PackageExtraction/ZipArchiveExtensionsTests.cs
jeffkl: Yeah for some reason the exception is never thrown and the file's last updated time is updated.  I decided to just #ifdef it out to get me unblocked, not sure if this functionality really needs to be tested on all frameworks.

build/common.project.props
jeffkl: Ended up removing this property and the corresponding comment, thanks for the suggestion.

build/common.project.props
jeffkl: Thanks for pointing this out, turns out I didn't end up using this as the functional tests either use the `TargetFrameworksUnitTestForSigning` or only target one framework.  I've removed this property.

build/common.project.props
jeffkl: Done!

build/common.ps1
jeffkl: The script definitely figures this out automatically, I just cut/pasted.  But I'll remove it since its not needed.

build/DotNetSdkTestVersions.txt
nkolev92: I think the xplat integration tests also use the downloaded .NET SDK. 

fwiw, I filled https://github.com/NuGet/Client.Engineering/issues/2172. 

What would it mean for build time to get the SDK it needs?

nkolev92 (dismissed):
Excellent work @jeffkl!

I'm really excited about the test vs build SDK separation and that our tests are primarily targeting net8.0 right now instead of netcoreapp3.1!

I think we should create a tracking issue for the mac tests as @zivkan suggested, and 🚢

src/NuGet.Core/NuGet.Configuration/NuGet.Configuration.csproj
jeffkl: It unfortunately didn't work, the conditions using properties only works on items themselves.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetSignTests.cs
jeffkl: Thanks.  https://github.com/NuGet/Client.Engineering/issues/2781

build/DotNetSdkTestVersions.txt
jeffkl: I think for now, I'm going to leave it as-is.  I like that the two files are next to each other and follow the same convention.  But we can definitely change it later.  If we had it happen as part of the build, it could make things strange in Visual Studio but it would mean you no longer have to run `configure.ps1` ahead of time.  I want to play around with alternatives without holding up this change.

nkolev92: > I want to play around with alternatives without holding up this change.

Yup, I agree :D We're just discussing ideas.

nkolev92 (approved):

PR #3462, opened 2020-06-18
Replace JObjects in PackageMetadataResourceV3 with strong type for better performance, lower memory consumption

erdembayar:
## Bug

Fixes: https://github.com/NuGet/Home/issues/9719
Regression: No  
* Last working version:   
* How are we preventing it in future:   

## Fix

Details: 
Using VS nuget GUI to consolodate different versions of a package crashes the instance of VS. Occurs in 2017, 2019 and 2019 int preview.
One of main problem was PM UI creates giant Jobject and millions of Nuget related objects and ballooning memory.
I am replacing this strong type instead of JObject.

## Testing/Validation

Tests Added: Yes
Reason for not adding tests:  
Validation:  Private unit test.

PR #5741, opened 2024-04-05
Fix NRE and Disallow Missing Registration Page Leaves

### LINKED ISSUE #13377: NRE When Package Metadata CatalogPage Leaf returns 404
### NuGet Product Used

dotnet.exe, NuGet.exe, NuGet SDK

### Product Version

dotnet sdk 8 and almost all versions of NuGet

### Worked before?

Issue was introduced in [#3462](https://github.com/NuGet/NuGet.Client/pull/3462)

### Impact

### Repro Steps & Context

You need to simulate a catalog page list where at least one of the range url 404's.

Here is the JSON my NuGet repo returns:
```
{
    "@id": "https://my.nuget.repo/v3/registration3-gz-semver2/my.package/index.json",
    "@type": [
        "catalog:CatalogRoot",
        "PackageRegistration",
        "catalog:Permalink"
    ],
    "count": 2,
    "items": [
        {
            "@id": "https://my.nuget.repo/v3/registration3-gz-semver2/my.package/page/7.4.3388/2018.4.0-enh-workersinclien13.json",
            "@type": "catalog:CatalogPage",
            "count": 128,
            "lower": "7.4.3388",
            "upper": "2018.4.0-enh-workersinclien13"
        },
        {
            "@id": "https://my.nuget.repo/v3/registration3-gz-semver2/my.package/page/7.4.3150/7.4.3387.json",
            "@type": "catalog:CatalogPage",
            "count": 128,
            "lower": "7.4.3150",
            "upper": "7.4.3387"
        }
    ]
}
``` 

The first page url returns a valid list of versions in that range ✔️ However the second one 404's and the leaf page ends up being null in the code here: https://github.com/NuGet/NuGet.Client/blob/7ad6fcc9c56c960975c37b2416c7eae1d53ba3fd/src/NuGet.Core/NuGet.Protocol/Resources/PackageMetadataResourceV3.cs#L122-L124

And because the if check is checking the wrong variable I get a NRE.

### Verbose Logs

```shell
The issue happens when running `dotnet tool restore` with the following `dotnet-tools.json`

{
  "version": 1,
  "isRoot": true,
  "tools": {
    "my.package": {
      "version": "9.1.7",
      "commands": [
        "dotnet-tool"
      ]
    }
  }
}


[NuGet Manager] [Info]   GET https://my.nuget.repo/v3/registration3-gz-semver2/my.package/index.json
[NuGet Manager] [Info]   OK https://my.nuget.repo/v3/registration3-gz-semver2/my.package/index.json 237ms
[NuGet Manager] [Info]   GET https://my.nuget.repo/v3/registration3-gz-semver2/my.package/page/7.4.3388/2018.4.0-enh-workersinclien13.json
[NuGet Manager] [Info]   OK https://my.nuget.repo/v3/registration3-gz-semver2/my.package/page/7.4.3388/2018.4.0-enh-workersinclien13.json 108ms
[NuGet Manager] [Info]   GET https://my.nuget.repo/v3/registration3-gz-semver2/my.package/page/7.4.3150/7.4.3387.json
[NuGet Manager] [Info]   NotFound https://my.nuget.repo/v3/registration3-gz-semver2/my.package/page/7.4.3150/7.4.3387.json 131ms
Unhandled exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at NuGet.Protocol.PackageMetadataResourceV3.ProcessRegistrationPage(RegistrationPage registrationPage, List`1 results, VersionRange range, Boolean includePrerelease, Boolean includeUnlisted, MetadataReferenceCache metadataCache)
   at NuGet.Protocol.PackageMetadataResourceV3.GetMetadataAsync(String packageId, Boolean includePrerelease, Boolean includeUnlisted, VersionRange range, SourceCacheContext sourceCacheContext, ILogger log, CancellationToken token)
   at NuGet.Protocol.PackageMetadataResourceV3.GetMetadataAsync(String packageId, Boolean includePrerelease, Boolean includeUnlisted, SourceCacheContext sourceCacheContext, ILogger log, CancellationToken token)
   at Microsoft.DotNet.Cli.NuGetPackageDownloader.NuGetPackageDownloader.GetPackageMetadataAsync(PackageSource source, String packageIdentifier, Boolean includePrerelease, Boolean includeUnlisted, CancellationToken cancellationToken)
   at Microsoft.DotNet.Cli.NuGetPackageDownloader.NuGetPackageDownloader.GetPackageMetadataAsync(String packageIdentifier, NuGetVersion packageVersion, IEnumerable`1 sources, CancellationToken cancellationToken, Boolean includeUnlisted)
   at Microsoft.DotNet.Cli.NuGetPackageDownloader.NuGetPackageDownloader.GetPackageSourceAndVersion(PackageId packageId, NuGetVersion packageVersion, PackageSourceLocation packageSourceLocation, Boolean includePreview, Boolean includeUnlisted, PackageSourceMapping packageSourceMapping)
   at Microsoft.DotNet.Cli.NuGetPackageDownloader.NuGetPackageDownloader.DownloadPackageAsync(PackageId packageId, NuGetVersion packageVersion, PackageSourceLocation packageSourceLocation, Boolean includePreview, Boolean includeUnlisted, Nullable`1 downloadFolder, PackageSourceMapping packageSourceMapping)
   at Microsoft.DotNet.Cli.ToolPackage.ToolPackageDownloader.DownloadAndExtractPackage(PackageLocation packageLocation, PackageId packageId, INuGetPackageDownloader nugetPackageDownloader, String packagesRootPath, IToolPackageStore toolPackageStore, NuGetVersion packageVersion, PackageSourceLocation packageSourceLoca
tion, Boolean includeUnlisted)
   at Microsoft.DotNet.Cli.ToolPackage.ToolPackageDownloader.<>c__DisplayClass8_0.<InstallPackage>b__0()
   at Microsoft.DotNet.Cli.TransactionalAction.Run[T](Func`1 action, Action commit, Action rollback)
   at Microsoft.DotNet.Tools.Tool.Restore.ToolRestoreCommand.InstallPackages(ToolManifestPackage package, Nullable`1 configFile)
   at Microsoft.DotNet.Tools.Tool.Restore.ToolRestoreCommand.<>c__DisplayClass10_0.<Execute>b__0(ToolManifestPackage package)
   at System.Linq.Enumerable.SelectArrayIterator`2.ToArray()
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at Microsoft.DotNet.Tools.Tool.Restore.ToolRestoreCommand.Execute()
   at System.CommandLine.Invocation.InvocationPipeline.Invoke(ParseResult parseResult)
   at Microsoft.DotNet.Cli.Program.ProcessArgs(String[] args, TimeSpan startupTime, ITelemetry telemetryClient)
```

mukunku:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13377

Regression? Last working version: The NRE has been there since the code was written in [#3462](https://github.com/NuGet/NuGet.Client/pull/3462)

## Description

The original issue is just the wrong variable being used in the `if` check. However for some reason when doing `dotnet tool restore` in the dotnet 6 sdk the issue doesn't happen for us. But once we upgraded our project to use dotnet 8 the NRE started happening every time we ran `dotnet tool restore`. So this is sort of a "regression" for us because our in-house NuGet repo (cloudsmith) returns 404 if there are no active packages in a version range.

I think it makes sense to gracefully skip any missing leaf pages. We are already allowing 404's with the `IgnoreNotFounds = true,` flag. So I think not throwing an exception in a recoverable situation makes more sense here and it would help resolve our issue as well.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Protocol/Resources/PackageMetadataResourceV3.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Resources/PackageMetadataResourceV3.cs
@@ -120,12 +120,6 @@ namespace NuGet.Protocol
                     {
                         var rangeUri = registrationPage.Url;
                         var leafRegistrationPage = await GetRegistratioIndexPageAsync(_client, rangeUri, packageId, lower, upper, httpSourceCacheContext, log, token);
-
-                        if (leafRegistrationPage == null)
-                        {
-                            throw new InvalidDataException(rangeUri);
-                        }
-
                         ProcessRegistrationPage(leafRegistrationPage, results, range, includePrerelease, includeUnlisted, metadataCache);
                     }
                     else
--- a/src/NuGet.Core/NuGet.Protocol/Resources/PackageMetadataResourceV3.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Resources/PackageMetadataResourceV3.cs
@@ -121,9 +121,10 @@ namespace NuGet.Protocol
                         var rangeUri = registrationPage.Url;
                         var leafRegistrationPage = await GetRegistratioIndexPageAsync(_client, rangeUri, packageId, lower, upper, httpSourceCacheContext, log, token);
 
-                        if (registrationPage == null)
+                        if (leafRegistrationPage == null)
                         {
-                            throw new InvalidDataException(registrationUri.AbsoluteUri);
+                            log.LogDebug($"{rangeUri} returned 404 NotFound. No versions in this range are available.");
+                            continue;
                         }
 
                         ProcessRegistrationPage(leafRegistrationPage, results, range, includePrerelease, includeUnlisted, metadataCache);
--- a/src/NuGet.Core/NuGet.Protocol/Resources/PackageMetadataResourceV3.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Resources/PackageMetadataResourceV3.cs
@@ -123,8 +123,7 @@ namespace NuGet.Protocol
 
                         if (leafRegistrationPage == null)
                         {
-                            log.LogDebug($"{rangeUri} returned 404 NotFound. No versions in this range are available.");
-                            continue;
+                            throw new InvalidDataException(rangeUri);
                         }
 
                         ProcessRegistrationPage(leafRegistrationPage, results, range, includePrerelease, includeUnlisted, metadataCache);
@@ -234,7 +233,7 @@ namespace NuGet.Protocol
                                 $"list_{packageIdLowerCase}_range_{lower.ToNormalizedString()}-{upper.ToNormalizedString()}",
                                 httpSourceCacheContext)
                             {
-                                IgnoreNotFounds = true,
+                                IgnoreNotFounds = false,
                             },
                             httpSourceResult => DeserializeStreamDataAsync<RegistrationPage>(httpSourceResult.Stream, token),
                             log,
--- a/test/NuGet.Core.Tests/NuGet.Protocol.Tests/PackageMetadataResourceV3Tests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Protocol.Tests/PackageMetadataResourceV3Tests.cs
@@ -11,7 +11,6 @@ using System.Threading.Tasks;
 using NuGet.Packaging;
 using NuGet.Packaging.Core;
 using NuGet.Protocol.Core.Types;
-using NuGet.Test.Utility;
 using NuGet.Versioning;
 using Test.Utility;
 using Xunit;
@@ -302,6 +301,37 @@ namespace NuGet.Protocol.Tests
                 //Assert
                 Assert.Null(metadata);
             }
+        }
+
+        [Fact]
+        public async Task PackageMetadataResourceV3_GetMetadataAsync_RegistrationPageLeaf_NotFound()
+        {
+            // Arrange
+            var source = "https://api.nuget.org/v3/index.json";
+            var dummyPackage = "dummy.package";
+            var responses = new Dictionary<string, string>
+            {
+                { "https://api.nuget.org/v3/index.json", JsonData.RepoSignIndexJsonData },
+                { $"https://api.nuget.org/v3/registration0/{dummyPackage}/index.json", "{\"ok\": false, \"error\": \"This feature is not supported.\"}" },
+                { $"https://api.nuget.org/v3-registration3-gz-semver2/{dummyPackage}/index.json", JsonData.PackageRegistrationCatalogWithLeafPagesUpperLower },
+                { $"https://api.nuget.org/v3/registration3-gz-semver2/{dummyPackage}/page/2.0.1/3.0.0.json", string.Empty /*404*/ },
+                { $"https://api.nuget.org/v3/registration3-gz-semver2/{dummyPackage}/page/1.0.0/2.0.0.json", string.Empty /*404*/ },
+            };
+
+            var repo = StaticHttpHandler.CreateSource(source, Repository.Provider.GetCoreV3(), responses);
+            var resource = await repo.GetResourceAsync<PackageMetadataResource>(CancellationToken.None);
+
+            //Act
+            using var sourceCacheContext = new SourceCacheContext();
+            var packages = await resource.GetMetadataAsync(dummyPackage,
+                false,
+                false,
+                sourceCacheContext,
+                Common.NullLogger.Instance,
+                CancellationToken.None);
+
+            //Assert 
+            Assert.Empty(packages);
         }
     }
 }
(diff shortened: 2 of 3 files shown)


mukunku:
@dotnet-policy-service agree company="StackOverflow"

src/NuGet.Core/NuGet.Protocol/Resources/PackageMetadataResourceV3.cs
```diff
@@ -121,9 +121,10 @@ public class PackageMetadataResourceV3 : PackageMetadataResource
                         var rangeUri = registrationPage.Url;
                         var leafRegistrationPage = await GetRegistratioIndexPageAsync(_client, rangeUri, packageId, lower, upper, httpSourceCacheContext, log, token);
 
-                        if (registrationPage == null)
+                        if (leafRegistrationPage == null)
                         {
-                            throw new InvalidDataException(registrationUri.AbsoluteUri);
+                            log.LogDebug($"{rangeUri} returned 404 NotFound. No versions in this range are available.");
```
nkolev92: Why does the server return a leaf page that 404s?
I think that'd be bad data, so I'm not sure we should continue quietly. 

You mentioned https://github.com/NuGet/NuGet.Client/commit/7532d57276a3e2ace5346f5a255111049620bff4 had the first bug. 

Do you know what was the behavior in 5.6 before this change?

mukunku: No not sure how it behaved before that. It 404's because all the versions within the leaf have been unlisted at some point. I can't comment on why cloudsmith opts to 404 when there are no active versions in a range but just based on the `IgnoreNotFound` flag that's being set here, I feel like this is a valid way of handling this maybe?

mukunku: I can force cloudsmith to synchronize at least one package from the range that 404's by explicitly installing an older version. That fixes those urls so they don't 404 for me but I feel like this behavior change between .net sdk 6 and .net sdk 8 is not desirable. Whether that's NuGet's responsibility I'm not sure but this seemed like a reasonable implementation to me.

nkolev92: >  It 404's because all the versions within the leaf have been unlisted

I think there's a couple of things that kind of seem incorrect. 

- Registration should not be filtering unlisted packages. 
- If all packages from a package don't "exist", then the page should not exist either. 

In general, I think this change won't really lead to any issues with current servers, since most servers won't have empty pages, but it still feels like this should be a server side fix.

mukunku: We tried but couldn't find any way to disable this behavior on the Cloudsmith side. Since we have a workaround in place and this isn't a blocker for us I'm happy to change this PR to just check the correct variable if that's what we want. But I still think this change makes sense in terms of defensive programming.

nkolev92: I think the check on line 124 makes sense. Currently it's checking the wrong value. 

I'm not sure I'd call continue instead of throwing defensive programming. There's a client/server contract and the server is not honoring it. Throwing is a controlled failure.

mukunku: If we're going to consider this a data contract error would it make sense to set `IgnoreNotFounds` to `false` ?
https://github.com/NuGet/NuGet.Client/blob/3088606e8ab4b499adb8b645d056029411826e6d/src/NuGet.Core/NuGet.Protocol/Resources/PackageMetadataResourceV3.cs#L237

Not sure why we're handling 404 differently in this case. We can treat it like any other http error code.

mukunku: @nkolev92 Let me know what you think of my latest changes.

I set `IgnoreNotFounds` to false and adjusted my unit test accordingly. I also removed the null check since we [ensure a successful http status code](https://github.com/NuGet/NuGet.Client/blob/7ad6fcc9c56c960975c37b2416c7eae1d53ba3fd/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpSource.cs#L155) in the http handler.

mukunku: @nkolev92 Any thoughts on the latest version of the code?

mukunku:
Seems like folks don't like my fix so going to close this out.

PR #5744, opened 2024-04-08
Add option to disable all strong name signing during build of NuGet.Client repository

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2790

## Description

This change makes it possible to build our repository without any strong name signing enabled via `SkipSigning=true`.  This will allow us to have a build pipeline that doesn't require disabling strong name validation and external contributors can build and run tests easier.

* Moved all of the `InternalsVisibleTo` logic from `AssemblyInfo.cs` to the projects so we can adjust the public key token depending on if strong name signing is enabled
* Added a new strong name key to sign tests with so that all `InternalsVisibleTo` are using the same public key.
* Updated logic used during ILMerge to use the same key as the containing project.

This doesn't currently change any behavior other than tests using a different key.  You can just now build with `/property:SkipSigning=true` at the command-line to build without any signing.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


jeffkl:
I've scheduled an official build to make sure I didn't break anything: https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=9382138&view=results

build/config.props
```diff
@@ -90,11 +90,6 @@
     <PackageIcon>icon.png</PackageIcon>
   </PropertyGroup>
 
-  <PropertyGroup>
-    <NUGET_PFX_PATH>$(MSBuildThisFileDirectory)..\keys\NuGetKey.snk</NUGET_PFX_PATH>
-    <MS_PFX_PATH>$(MSBuildThisFileDirectory)..\keys\35MSSharedLib1024.snk</MS_PFX_PATH>
```
zivkan, line 95: build.ps1 override these when `-SkipDelaySign` was passed as an argument. It should be updated.

It possibly also makes this PR redundant, since there's already a different way to achieve the same outcome?

zivkan:
> * [x] PR has a meaningful title

it does not at the moment.

build/config.props
jeffkl: Thanks for pointing that out. I've updated the script to set `SkipSigning=true` to just disable all strong name signing.  This pull request will make it easier for our CI on all platforms to disable signing.  I'm thinking in the near future to instead use `SignType` values of `None` (default), `Test`, and `Real`.  This would be more in line with MicroBuild.  But from a developer's perspective, we just build and run tests.  I need to find a way to make it so we don't have to do anything special to debug our VSIX (if possible) and have the default signing to None.  Otherwise I'll default signing to "Test" which will delay sign things and require strong name signing validation exclusions.  But for Linux and Mac CI, we'll just build and run tests as unsigned.

test/TestUtilities/Test.Utility/Test.Utility.csproj
```diff
@@ -3,42 +3,27 @@
     <TargetFrameworks>$(TargetFrameworksUnitTestForSigning)</TargetFrameworks>
     <NoWarn>$(NoWarn);CS1591</NoWarn>
     <NoWarn Condition=" $(TargetFramework.StartsWith('netcoreapp')) ">$(NoWarn);CS1998</NoWarn>
-    <PackProject>true</PackProject>
-    <Shipping>true</Shipping>
-    <IsPackable>true</IsPackable>
+    <PackProject>false</PackProject>
+    <Shipping>false</Shipping>
+    <IsPackable>false</IsPackable>
```
zivkan: NuGet server team uses this Test.Utility package for signed package testing of nuget.org ingestion pipelines. This needs to keep being a package for them.

jeffkl: It looks like its only used in NuGet.Jobs and version 6.4?  Would it be better to get them a better system or continue to "ship" this test utility?  They obviously don't need an update, unless there are more usages I couldn't find.

https://github.com/search?q=org%3ANuGet%20%22Test.Utility%22&type=code

nkolev92: They're using the package creating utilities. 
I think we should keep it for now and not create potentially extra work for them. 
IMO we can remove if they're ok it.

src/NuGet.Clients/NuGet.CommandLine/NuGet.CommandLine.csproj
```diff
@@ -12,6 +12,7 @@
     <ApplicationManifest>app.manifest</ApplicationManifest>
     <Shipping>true</Shipping>
     <OutputType>Exe</OutputType>
+    <SignWithMicrosoftKey>true</SignWithMicrosoftKey>
```
nkolev92, line 15: Shouldn't shipping imply signed with ms key?

build/sign.targets
```diff
@@ -1,27 +1,38 @@
...
+
+  <PropertyGroup Condition="'$(SignWithNuGetKey)' == 'true' Or '$(SignWithMicrosoftKey)' == 'true'">
     <DelaySign Condition="'$(IsXplat)' != 'true'">true</DelaySign>
     <PublicSign Condition="'$(IsXplat)' == 'true'">true</PublicSign>
-    <AssemblyOriginatorKeyFile Condition = " '$(AssemblyOriginatorKeyFile)' == '' ">$(StrongNameKey)</AssemblyOriginatorKeyFile>
-    <StrongNameCert Condition="'$(StrongNameCert)' == ''">$(DefaultStrongNameCert)</StrongNameCert>
   </PropertyGroup>
 
+  <ItemGroup>
+    <InternalsVisibleTo Update="@(InternalsVisibleTo)" PublicKey="0024000004800000940000000602000000240000525341310004000001000100A5276DF8650A58CB43396DC7B3D395F30A82D0D1FA98FBCFE3ABEAD5DE0B1DB6764347A0F6BF0B060A27C202CCD122DB5DED8F596CEBE2ECC3A6629015EEB96C94F6B9E8185D4ACC84C376FF6B1C3147431A4D55CB5736DB97A9E88FCC47D9193F4DB5896DC5817E5D0CBD2641726E7431990BCD2DD7FA1D28493D0CFD9DCFA4" />
```
nkolev92, line 36: I love this trickery :D

src/NuGet.Clients/NuGet.CommandLine/NuGet.CommandLine.csproj
jeffkl: We sign assemblies with two keys, NuGet and Microsoft.  By default, if not specified, everything is signed with the NuGet key.  But these assemblies get the Microsoft key:

* `src\NuGet.Core\**\*.csproj`
* `src\NuGet.Clients\NuGet.CommandLine\NuGet.CommandLine.csproj`
* `src\NuGet.Clients\NuGet.Indexing\NuGet.Indexing.csproj`

I think the Directory.Build.props I added should actually be in `src\NuGet.Core` so I'm going to move it.

test/TestUtilities/Test.Utility/Test.Utility.csproj
jeffkl: Okay I'll bring it back

src/NuGet.Clients/NuGet.CommandLine/NuGet.CommandLine.csproj
nkolev92: damn, it's way more complicated than I thought. 

I'll resolve this thread.

jeffkl:
Official build is now passing: https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=9389345

I also verified a few assemblies to make sure their public key token didn't change.  We can merge this and if it causes any issues we just can just revert and try again.

nkolev92 (approved):

PR #5751, opened 2024-04-12
Protocol loads Owners from package sources as an enumerable type

### LINKED ISSUE #13186: Owner package metadata is an array in JSON but a string in Protocol types

### LINKED ISSUE #13392: UIActionEngine test assuming task cancellation is flaky
`CreateInstallAction_OnInstallingProject_EmitsPkgWasTransitiveTelemetryAndTabAndIsSolutionPropertiesAsync` assumes it will generate a cancelled telemetry event, but can find a successful event.

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13186
Fixes: https://github.com/NuGet/Home/issues/13392 (unrelated flaky tests)

## Description

_The original PR https://github.com/NuGet/NuGet.Client/pull/5605 was reverted in https://github.com/NuGet/NuGet.Client/pull/5709 was reverted._

**This PR's changes to PR 5605:**

- Handle Search response with a `string` JSON type for `owners:`.
  - A converter will put a string response into a single item array. 
  - I've updated Search API docs to clarify with customers that a string response is expected to be a single owner username in https://github.com/NuGet/docs.microsoft.com-nuget/pull/3280
- Tests changes are a new JSON with string `owner` which is deserialized in PackageSearchResourceV3Tests.cs, and direct unit tests for the new converter.
- Unrelated flaky test fixes in NuGet.PackageManagement.UI.Test.UIActionEngineTests:
   -`CreateInstallAction_OnInstallingProject_EmitsPkgWasTransitiveTelemetryAndTabAndIsSolutionPropertiesAsync`
  - `CreateInstallAction_OnInstallingProject_WithNewSourceMapping_DoesNotLogTelemetry`

**Original PR 5605 description:**

Read Owners collection from package sources into an enumerable type, rather than converting it to a comma-delimited string immediately upon parsing the JSON response.

There's no effective customer change in this PR, just that the enumeration is retained in Protocol types rather than only as a concatenated string. In subsequent PRs, I'll utilize this enumeration of owners for customer benefit.

Test changes:
- Ensure a package source response that has Owners is converted into a matching enumeration.
- Ensure a package source response without Owners returns null and doesn't affect other metadata.
- Update `EntityFrameworkSearch.json` (from stale 2017 data) with latest package search metadata result response from nuget.org.
- Update test with fluent assertions for all properties in the updated JSON.
- Update `test/NuGet.Core.Tests/NuGet.CommandLine.Xplat.Tests/PackageSearchRunnerTests.cs` which was failing as it attempted to assert that the Table Output contained owners. The registration JSON does not normally contain Owner, so I updated the JSON and the test accordingly.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [x] Documentation PR or issue filled https://github.com/NuGet/docs.microsoft.com-nuget/pull/3280
  - **OR**
  - [ ] N/A


src/NuGet.Core/NuGet.Protocol/Converters/MetadataStringOrArrayConverter.cs
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Newtonsoft.Json;
+
+namespace NuGet.Protocol
+{
+    public class MetadataStringOrArrayConverter : JsonConverter
```
nkolev92: Does this need to be public?

If Ctrl + F didn't betray me, I think it's only used it Protocol. 

Nice fix to the problem btw :)_

donnie-msft: Only our parsing of `PackageSearchMetadata` depends on it.

All our JSON converters are public, as well as the metadata type which will be utilizing this converter. Can you help me understand the risk of it remaining public?

nkolev92: I'm not sure risk is the correct word, there's no functional risk here, it's all about support.

https://github.com/NuGet/NuGet.Client/blob/dev/docs/nuget-sdk.md#changes-to-apis 

> Do all of your public APIs need to be public, or can any be made internal?

In general, we try to minimize public APIs because as said in the doc, we have to support them forever. 
Personally, if we could take it back, I'd say all converters should have remained internal.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/NuGet.Protocol.Tests.csproj
```diff
@@ -16,6 +16,10 @@
     <None Remove="RemoteRepositories\NewFolder\**" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Remove="compiler\resources\EntityFrameworkSearchWithStringTypes.json" />
```
nkolev92: This is already done on line 9.

nkolev92 (commented):
I think there's an extra change in csproj.


Rest looks great, nice fix, and excellent tests!

test/NuGet.Core.Tests/NuGet.Protocol.Tests/NuGet.Protocol.Tests.csproj
donnie-msft: Weird, it seems VS must have added this when I added a new file.

nkolev92 (approved):

PR #5757, opened 2024-04-17
Include NuGet.Tests.Apex.dll in NuGet.Tests.Apex.Daily.runsettings

SueSu01:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2778

## Description

Add all other existing Apex test cases to the new daily pipeline and in order to avoid the risk of duplication so include the NuGet.Tests.Apex.dll in NuGet.Tests.Apex.Daily.runsettings.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/build/template.runsettingsproj
+++ b/build/template.runsettingsproj
@@ -48,7 +48,7 @@
   </ItemGroup>
 
   <ItemGroup Label="Test Containers">
-    <TestContainer Include="$(FileName)" /> 
+    <TestContainer Include="$(FileName)" />
     <InstallationConfiguration Include="$(VsConfigPath)" />
   </ItemGroup>
 
--- a/build/template.runsettingsproj
+++ b/build/template.runsettingsproj
@@ -48,8 +48,7 @@
   </ItemGroup>
 
   <ItemGroup Label="Test Containers">
-    <TestContainer Include="$(FileName)" />
-    <TestContainer Include="NuGet.Tests.Apex.dll" Condition=" '$(RunName)' == 'NuGet.Tests.Apex.Daily' " />  
+    <TestContainer Include="$(FileName)" /> 
     <InstallationConfiguration Include="$(VsConfigPath)" />
   </ItemGroup>
 
--- a/build/template.runsettingsproj
+++ b/build/template.runsettingsproj
@@ -49,6 +49,7 @@
 
   <ItemGroup Label="Test Containers">
     <TestContainer Include="$(FileName)" />
+    <TestContainer Include="NuGet.Tests.Apex.dll" Condition=" '$(RunName)' == 'NuGet.Tests.Apex.Daily' " />  
     <InstallationConfiguration Include="$(VsConfigPath)" />
   </ItemGroup>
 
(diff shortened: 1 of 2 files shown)


build/template.runsettingsproj
```diff
@@ -49,6 +49,7 @@
 
   <ItemGroup Label="Test Containers">
     <TestContainer Include="$(FileName)" />
+    <TestContainer Include="NuGet.Tests.Apex.dll" Condition=" '$(RunName)' == 'NuGet.Tests.Apex.Daily' " />
```
nkolev92: Can this be passed from runsettings.proj instead?

FileName is our own property, so we can make different, like for example, we can split by `,` or something and include all of that as a test container.

SueSu01: > Can this be passed from runsettings.proj instead?
> 
> FileName is our own property, so we can make different, like for example, we can split by `,` or something and include all of that as a test container.

@nkolev92 I tried split the multiple .dlls with `&` and ` ,` respectively and assign it to FileName, the run failed at `Deploy Tests Install-TestContainers` . 
image
I'm not sure if a test container can include mutiple dlls. Do you have any suggesstion or idea?

zivkan: MSBuild syntax is to use a semicolon: https://learn.microsoft.com/en-us/visualstudio/msbuild/how-to-select-the-files-to-build?view=vs-2022#to-declare-multiple-items

SueSu01: @zivkan @nkolev92 Thanks for your guidance. Updated the code with your suggestion.

nkolev92 (approved):

build/template.runsettingsproj
zivkan: Looking at the daily test pipeline, I don't see a manual run of this branch to validate that this PR works as intended.  Have you tested this change somewhere else?

SueSu01: > Looking at the daily test pipeline, I don't see a manual run of this branch to validate that this PR works as intended. Have you tested this change somewhere else?

I verified that this PR works on this [daily run](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=9445583&view=results) triggered by [private build](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=9445925&view=results) which include my changes.

zivkan: @SueSu01 that CI run looks like it only ran the apex tests, but not the daily tests assembly. So, it only ran one of the two test assemblies.

SueSu01: > @SueSu01 that CI run looks like it only ran the apex tests, but not the daily tests assembly. So, it only ran one of the two test assemblies.

Currently daily tests include 42 P0 and P1 apex cases. The NuGet.Tests.Apex.dll assembly has 101 apex tests and I checked that CI run has 143 tests totally, so it ran two test assemblies.

nkolev92: image

This shows the multiple assemblies were run too. 

Conveniently there was a failure in each :D

v-luzh: Yes, we will try to investigate the failures in the recent runs during the following week.

PR #5760, opened 2024-04-22
HTTPS everywhere: Throw an error for non https service endpoints

### LINKED ISSUE #13364: Show an error when a non https source is used in a resource in a service index
### NuGet Product(s) Involved

NuGet.exe

### The Elevator Pitch

In the scenarios where the service index is https, but through the index.json the links for certain resources in there are http, show an error message.

_No response_

Nigusu-Allehu:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13364

## Description

An https NuGet source could have a non https service endpoint.
```json
{
    "version": "3.0.0",
    "resources": [
        {
            "@id": "http://source",
        }]
}
```
This PR makes sure we throw an error for such endpoints unless the source is set to `AllowInsecureConnections` in the Nuget config file.
## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A


src/NuGet.Core/NuGet.Protocol/Providers/PackageSearchResourceV3Provider.cs
```diff
@@ -23,6 +25,16 @@ public PackageSearchResourceV3Provider()
             if (serviceIndex != null)
             {
                 var endpoints = serviceIndex.GetServiceEntryUris(ServiceTypes.SearchQueryService);
+
+                foreach (Uri endpoint in endpoints)
+                {
+                    PackageSource serviceSource = new PackageSource(endpoint.ToString(), endpoint.ToString());
```
zivkan: We need to minimize memory allocations to minimize performance slowdowns. Uri has a property `OriginalString`, which I expect can be used and shouldn't allocate.

src/NuGet.Core/NuGet.Protocol/RemoteRepositories/HttpFileSystemBasedFindPackageByIdResourceProvider.cs
```diff
@@ -23,6 +25,15 @@ public HttpFileSystemBasedFindPackageByIdResourceProvider()
             var serviceIndexResource = await sourceRepository.GetResourceAsync<ServiceIndexResourceV3>(token);
             var packageBaseAddress = serviceIndexResource?.GetServiceEntryUris(ServiceTypes.PackageBaseAddress);
 
+            foreach (Uri endpoint in packageBaseAddress)
+            {
+                PackageSource serviceSource = new PackageSource(endpoint.ToString(), endpoint.ToString());
```
zivkan: We need to minimize memory allocations to minimize performance slowdowns. Uri has a property `OriginalString`, which I expect can be used and shouldn't allocate.

src/NuGet.Core/NuGet.Protocol/Providers/RawSearchResourceV3Provider.cs
```diff
@@ -27,6 +29,15 @@ public RawSearchResourceV3Provider()
             {
                 var endpoints = serviceIndex.GetServiceEntryUris(ServiceTypes.SearchQueryService);
 
+                foreach (Uri endpoint in endpoints)
+                {
+                    PackageSource serviceSource = new PackageSource(endpoint.ToString(), endpoint.ToString());
```
zivkan: We need to minimize memory allocations to minimize performance slowdowns. Uri has a property `OriginalString`, which I expect can be used and shouldn't allocate.

src/NuGet.Core/NuGet.Protocol/Providers/DownloadResourceV3Provider.cs
```diff
@@ -33,7 +33,14 @@ public DownloadResourceV3Provider()
 
                 // If index.json contains a flat container resource use that to directly
                 // construct package download urls.
-                var packageBaseAddress = serviceIndex.GetServiceEntryUri(ServiceTypes.PackageBaseAddress)?.AbsoluteUri;
+                var endpoint = serviceIndex.GetServiceEntryUri(ServiceTypes.PackageBaseAddress);
+                var packageBaseAddress = endpoint?.AbsoluteUri;
+
+                // Check for a not HTTPS source
+                if (endpoint.Scheme == Uri.UriSchemeHttp && endpoint.Scheme != Uri.UriSchemeHttps && !source.PackageSource.AllowInsecureConnections)
```
jebriede: If the endpoint scheme is equal to UriSchemeHttp, it cannot be equal to Uri.UriSchemeHttps so it's redundant to check that it's both equal to UriSchemeHttp and not equal to UriSchemeHttps.

src/NuGet.Core/NuGet.Protocol/Providers/DownloadResourceV3Provider.cs
```diff
@@ -33,7 +33,14 @@ public DownloadResourceV3Provider()
 
                 // If index.json contains a flat container resource use that to directly
                 // construct package download urls.
-                var packageBaseAddress = serviceIndex.GetServiceEntryUri(ServiceTypes.PackageBaseAddress)?.AbsoluteUri;
+                var endpoint = serviceIndex.GetServiceEntryUri(ServiceTypes.PackageBaseAddress);
+                var packageBaseAddress = endpoint?.AbsoluteUri;
+
+                // Check for a not HTTPS source
```
jebriede:
```suggestion
                // Check if the source is not HTTPS.
```

src/NuGet.Core/NuGet.Protocol/Providers/PackageDetailsUriResourceV3Provider.cs
```diff
@@ -24,6 +25,11 @@ public PackageDetailsUriResourceV3Provider()
             if (serviceIndex != null)
             {
                 var uri = serviceIndex.GetServiceEntryUri(ServiceTypes.PackageDetailsUriTemplate);
+                // Check for a not HTTPS source
```
jebriede:
```suggestion
                // Check if the source is not HTTPS.
```

src/NuGet.Core/NuGet.Protocol/Providers/PackageDetailsUriResourceV3Provider.cs
```diff
@@ -24,6 +25,11 @@ public PackageDetailsUriResourceV3Provider()
...
+                if (uri.Scheme == Uri.UriSchemeHttp && uri.Scheme != Uri.UriSchemeHttps && !source.PackageSource.AllowInsecureConnections)
```
jebriede: Same as my above comment, there is a redundant condition here.

src/NuGet.Core/NuGet.Protocol/Providers/PackageSearchResourceV3Provider.cs
```diff
@@ -23,6 +25,16 @@ public PackageSearchResourceV3Provider()
...
+                    if (serviceSource.IsHttp && !serviceSource.IsHttps && !source.PackageSource.AllowInsecureConnections)
```
jebriede: Is it possible for isHttp and isHttps both ever evaluate to true? If not, this is also redudant.

src/NuGet.Core/NuGet.Protocol/Providers/PackageUpdateResourceV3Provider.cs
```diff
@@ -30,7 +29,11 @@ public PackageUpdateResourceV3Provider()
             if (serviceIndex != null)
             {
                 var baseUrl = serviceIndex.GetServiceEntryUri(ServiceTypes.PackagePublish);
-
+                // Check for a not HTTPS source
+                if (baseUrl.Scheme == Uri.UriSchemeHttp && baseUrl.Scheme != Uri.UriSchemeHttps && !source.PackageSource.AllowInsecureConnections)
```
jebriede: Redundant condition. There are a few more occurrences of checking for that the scheme is equal to both http and not https in this PR so this comment applies to all those instances.

src/NuGet.Core/NuGet.Protocol/Providers/RegistrationResourceV3Provider.cs
```diff
@@ -27,7 +27,11 @@ public RegistrationResourceV3Provider()
             {
                 //This will come back as null if there are no matching RegistrationsBaseUrl types
                 var baseUrl = serviceIndex.GetServiceEntryUri(ServiceTypes.RegistrationsBaseUrl);
-
+                // Check for a not HTTPS source
+                if (baseUrl.Scheme == Uri.UriSchemeHttp && baseUrl.Scheme != Uri.UriSchemeHttps && !source.PackageSource.AllowInsecureConnections)
+                {
+                    throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_HttpServiceIndexUsage, source.PackageSource.SourceUri, baseUrl));
+                }
```
jebriede: Should the code that validates the endpoint / url be promoted to the ResourceProvider so we reduce code duplication?

src/NuGet.Core/NuGet.Protocol/Providers/DownloadResourceV3Provider.cs
```diff
@@ -33,7 +33,14 @@ public DownloadResourceV3Provider()
...
                 // If index.json contains a flat container resource use that to directly
                 // construct package download urls.
-                var packageBaseAddress = serviceIndex.GetServiceEntryUri(ServiceTypes.PackageBaseAddress)?.AbsoluteUri;
+                var endpoint = serviceIndex.GetServiceEntryUri(ServiceTypes.PackageBaseAddress);
+                var packageBaseAddress = endpoint?.AbsoluteUri;
+
+                // Check if the source is not HTTPS.
+                if (endpoint.Scheme == Uri.UriSchemeHttp && endpoint.Scheme != Uri.UriSchemeHttps && !source.PackageSource.AllowInsecureConnections)
+                {
+                    throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_HttpServiceIndexUsage, source.PackageSource.SourceUri, endpoint));
```
zivkan: These checks are being performed in every resource provider, which has the risk that we accidentally miss a provider, or in the future someone adds a new provider and doesn't add this AllowInsecureConnections check.

If instead we check only in the service index provider, when we load the service index, then I think it can reduce risk of human errors.  However, it would have the side-effect that if a custom nuget feed has a custom resource in the service index that NuGet doesn't use, we might throw exceptions. But it's a tradeoff that you might find acceptable.

Nigusu-Allehu: Great idea!

src/NuGet.Core/NuGet.Protocol/Providers/PackageSearchResourceV3Provider.cs
Nigusu-Allehu: https is a subset of http. Http would be true for both https and non https

src/NuGet.Core/NuGet.Protocol/PublicAPI/net472/PublicAPI.Shipped.txt
```diff
@@ -14,7 +14,6 @@ NuGet.Protocol.AutoCompleteResourceV2Feed
 NuGet.Protocol.AutoCompleteResourceV2FeedProvider
 NuGet.Protocol.AutoCompleteResourceV2FeedProvider.AutoCompleteResourceV2FeedProvider() -> void
 NuGet.Protocol.AutoCompleteResourceV3
-~NuGet.Protocol.AutoCompleteResourceV3.AutoCompleteResourceV3(NuGet.Protocol.HttpSource client, NuGet.Protocol.ServiceIndexResourceV3 serviceIndex, NuGet.Protocol.RegistrationResourceV3 regResource) -> void
```
zivkan: Don't break existing APIs unless it's really needed. Instead, add a new overload, and have the old overload call the new one with a "sensible" default for the new parameters.

Mark the old method as obsolete if there's a good enough justification, or you can use the banned API analyzer to just ban the old API within our own repo, without affecting custoemrs who might be calling it in their own apps.

src/NuGet.Core/NuGet.Protocol/Resources/ServiceIndexResourceV3.cs
```diff
@@ -119,6 +136,16 @@ public virtual Uri GetServiceEntryUri(params string[] orderedTypes)
             return GetServiceEntryUris(clientVersion, orderedTypes).FirstOrDefault();
         }
 
+        /// <summary>
+        /// Get the best match service URI. Give the option to allow insecure connections or non https sources.
+        /// Default is false.
+        /// </summary>
+        public virtual Uri GetServiceEntryUri(bool allowInsecureConnections, params string[] orderedTypes)
+        {
+            _allowInsecureConnections = allowInsecureConnections;
```
zivkan: This is not "thread safe", and I believe that currently `SourceRepository` and therefore `ServiceIndexResourceV3` can safely be called on multiple threads at the same time. In fact, I'm pretty sure the design of our VS integration actively uses it.

Also, from a public API design point of view, I wouldn't allow callers to specify whether they want to allow insecure connections. "Normal" SourceRepository usage is to create a SourceRepository object from a PackageSource, and then all "resources" use the settings from that initial creation. Therefore, I think it would be better that when this `ServiceIndexResourveV3` object is created, the `allowInsecureConnections` bool is passed from the `packageSource`, and then you don't need new public `GetServiceEntryUri` methods with the parameter.

src/NuGet.Core/NuGet.Protocol/Strings.resx
```diff
@@ -530,4 +530,9 @@ The "s" should be localized to the abbreviation for seconds.</comment>
 {1} - number of vulerabilitiy files reported
 {2} - max count allowed</comment>
   </data>
-</root>
+  <data name="Error_HttpServiceIndexUsage" xml:space="preserve">
+    <value>You have used a non HTTPS service index: '{0}'. NuGet requires HTTPS sources. To use HTTP sources, you must explicitly set 'allowInsecureConnections' to true in your NuGet.Config file. Please refer to https://aka.ms/nuget-https-everywhere for more information.</value>
```
zivkan, line 534: This message is saying that the service index is not https, but where this message is being used is checking the service index's entries, not the service index itself.

```suggestion
    <value>There are non-HTTPS resources defined by your package source: '{0}'. NuGet requires HTTPS endpoints. To allow unencrypted HTTP, you must explicitly set 'allowInsecureConnections' to true in your NuGet.Config file, or talk to the administrator of your source and ask them to migrate to a secure HTTPS URL. Please refer to https://aka.ms/nuget-https-everywhere for more information.</value>
```

src/NuGet.Core/NuGet.Protocol/Resources/ServiceIndexResourceV3.cs
```diff
@@ -105,7 +107,22 @@ private IReadOnlyList<ServiceIndexEntry> GetBestVersionMatchForType(NuGetVersion
...
-                return entries.Where(e => e.ClientVersion == bestMatch.ClientVersion).ToList();
+                List<ServiceIndexEntry> matchingEntries = new List<ServiceIndexEntry>();
+
+                foreach (var entry in entries)
+                {
+                    if (entry.ClientVersion == bestMatch.ClientVersion)
+                    {
+                        if (entry.Uri.Scheme == Uri.UriSchemeHttp && entry.Uri.Scheme != Uri.UriSchemeHttps && !_allowInsecureConnections)
+                        {
+                            throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_HttpServiceIndexUsage, entry.Uri));
```
kartheekp-ms, line 119: There is some context around why we didn't add warnings for HTTP resources in a feed in the past. It is documented [here](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1510220).

src/NuGet.Core/NuGet.Protocol/Resources/ServiceIndexResourceV3.cs
```diff
@@ -105,7 +107,22 @@ private IReadOnlyList<ServiceIndexEntry> GetBestVersionMatchForType(NuGetVersion
...
             {
                 // Find all entries with the same version.
-                return entries.Where(e => e.ClientVersion == bestMatch.ClientVersion).ToList();
+                List<ServiceIndexEntry> matchingEntries = new List<ServiceIndexEntry>();
+
+                foreach (var entry in entries)
+                {
+                    if (entry.ClientVersion == bestMatch.ClientVersion)
+                    {
+                        if (entry.Uri.Scheme == Uri.UriSchemeHttp && entry.Uri.Scheme != Uri.UriSchemeHttps && !_allowInsecureConnections)
```
kartheekp-ms, line 117: Are we performing this check even if the service index resource is cached?

kartheekp-ms, line 117: I am curious to understand if we perform this check once per every resource in the feed or once per resource and per package. If it is once per every resource in the feed, we should be okay I guess but if it is once per resource and per package then it may impact restore performance.

src/NuGet.Core/NuGet.Protocol/Resources/ServiceIndexResourceV3.cs
```diff
@@ -105,7 +108,22 @@ private IReadOnlyList<ServiceIndexEntry> GetBestVersionMatchForType(NuGetVersion
             else
             {
                 // Find all entries with the same version.
-                return entries.Where(e => e.ClientVersion == bestMatch.ClientVersion).ToList();
+                List<ServiceIndexEntry> matchingEntries = new List<ServiceIndexEntry>();
```
donnie-msft, line 111:
```suggestion
                List<ServiceIndexEntry> matchingEntries = new List<ServiceIndexEntry>(capacity: entries.Count);
```

src/NuGet.Core/NuGet.Protocol/Resources/ServiceIndexResourceV3.cs
```diff
@@ -22,6 +23,8 @@ public class ServiceIndexResourceV3 : INuGetResource
         private static readonly IReadOnlyList<ServiceIndexEntry> _emptyEntries = new List<ServiceIndexEntry>();
         private static readonly SemanticVersion _defaultVersion = new SemanticVersion(0, 0, 0);
 
+        internal bool _allowInsecureConnections = false; //by default don't allow non https service index
```
donnie-msft, line 26: Nits

```suggestion
        internal bool _allowInsecureConnections; // By default, don't allow non-HTTPS service index
```

src/NuGet.Core/NuGet.Protocol/Resources/ServiceIndexResourceV3.cs
```diff
@@ -105,7 +108,22 @@ private IReadOnlyList<ServiceIndexEntry> GetBestVersionMatchForType(NuGetVersion
             else
             {
                 // Find all entries with the same version.
-                return entries.Where(e => e.ClientVersion == bestMatch.ClientVersion).ToList();
+                List<ServiceIndexEntry> matchingEntries = new List<ServiceIndexEntry>();
+
+                foreach (var entry in entries)
+                {
+                    if (entry.ClientVersion == bestMatch.ClientVersion)
```
donnie-msft, line 115: I'm not following why this is a loop. Could there be a single (if) check above the loop to see if any `entry` has a mismatch with HTTPS/allowInsecureConnections ?

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetListPackageTests.cs
```diff
@@ -732,7 +730,38 @@ public async Task ListPackage_WithHttpSourceAndAllowInsecureConnections_WarnsCor
...
             using var mockServer = new FileSystemBackedV3MockServer(pathContext.PackageSource);
             mockServer.Start();
-            pathContext.Settings.AddSource("http-source", mockServer.ServiceIndexUri, allowInsecureConnections);
+            pathContext.Settings.AddSource("http-source", mockServer.ServiceIndexUri, allowInsecureConnectionsValue: "false");
+
+            _fixture.RunDotnetExpectSuccess(Directory.GetParent(projectA.ProjectPath).FullName, $"add package A --version 1.0.0", testOutputHelper: _testOutputHelper);
+
+            // Act & Assert
+            CommandRunnerResult listResult = _fixture.RunDotnetExpectFailure(Directory.GetParent(projectA.ProjectPath).FullName, $"list package --outdated", testOutputHelper: _testOutputHelper);
+            mockServer.Stop();
```
donnie-msft, line 748: Or you can use this above and we have less tests that might accidentally not stop the server.
`using var server = new MockServer();`

Nigusu-Allehu:
Closing: Too many conflicts. Issue being resolved here: https://github.com/NuGet/NuGet.Client/pull/5915

PR #5761, opened 2024-04-23
Add `dotnet nuget why` command

### LINKED ISSUE #11943: [Feature]: add dotnet nuget why to dotnet CLI
### NuGet Product(s) Involved

dotnet.exe

### The Elevator Pitch

This is a tracking issue for feature request https://github.com/NuGet/Home/issues/11782

Design spec - [dotnet nuget why command](https://github.com/NuGet/Home/blob/dev/accepted/2022/dotnet-nuget-why-proposal.md) ([Spec PR](https://github.com/NuGet/Home/pull/11875))

Proof of concept: https://github.com/NuGet/NuGet.Client/tree/pragnya17-dotnet-nuget-why ([compare](https://github.com/NuGet/NuGet.Client/compare/dev...pragnya17-dotnet-nuget-why))

- [x] Add `dotnet nuget why` command - https://github.com/NuGet/Client.Engineering/issues/2862
- [x] Add docs to `dotnet/docs` - https://github.com/NuGet/Home/issues/13499
- [x] Add integration to `dotnet/sdk` - https://github.com/NuGet/Client.Engineering/issues/2861

advay26:
## Bug

Fixes: https://github.com/NuGet/Home/issues/11943

## Description

Design spec: [dotnet nuget why command](https://github.com/NuGet/Home/blob/dev/accepted/2022/dotnet-nuget-why-proposal.md)

Adds a new `dotnet nuget why` command to the Dotnet CLI, allowing users to see the dependency graph for a given package. They can use this to easily identify the top-level packages that are bringing in a given transitive dependency.

`dotnet nuget why -h`
image

`dotnet nuget why C:\TestProject.sln System.Text.Json`
image

This was built on the proof-of-concept made by Pragnya and Kartheek: [pragnya17-dotnet-nuget-why](https://github.com/NuGet/NuGet.Client/compare/dev...pragnya17-dotnet-nuget-why)

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [x] Documentation PR or issue filled: https://github.com/NuGet/Home/issues/13499
  - **OR**
  - [ ] N/A


src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand.cs
```diff
@@ -0,0 +1,107 @@
...
+using System.Linq;
+using Microsoft.Extensions.CommandLineUtils;
+using NuGet.Common;
+using NuGet.Frameworks;
+
+namespace NuGet.CommandLine.XPlat
+{
+    internal static class WhyCommand
+    {
+        public static void Register(CommandLineApplication app, Func<ILogger> getLogger, Func<IWhyCommandRunner> getCommandRunner)
```
Nigusu-Allehu: I am not sure if we are still working on it but there is a plan to migrate from `Microsoft.Extensions.CommandLineUtils.CommandLineApplication` to `System.CommandLine.CliCommand`. As a result, when we add new commands, it's recommended to use `System.CommandLine.CliCommand` instead. @martinrrm do you know if that is still the case?

nkolev92: I agree, we should do System.Commandline.CLiCommand.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand.cs
```diff
@@ -0,0 +1,107 @@
...
+                throw new ArgumentException(
+                    string.Format(CultureInfo.CurrentCulture,
+                    Strings.WhyCommand_Error_ArgumentCannotBeEmpty,
+                    package.Name));
+            }
+        }
+
+        private static void ValidateFrameworksOption(CommandOption framework)
+        {
+            var frameworks = framework.Values.Select(f =>
```
nkolev92: This one is interesting, I think we need to use the full string and use https://github.com/NuGet/NuGet.Client/blob/2669e3770fbc99c3019d4f899a1c5438f26bd352/src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFile.cs#L107 instead. 

Instead of actual frameworks, we use the alias. This will be in support of https://github.com/NuGet/Home/pull/12124.

nkolev92: fwiw, this also allows people to specify `net8.0-windows` if that's in their csproj.

people won't always know the exact framework since the SDK sets that property.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -0,0 +1,548 @@
...
+                // if the assets file was not found, print an error message and continue to the next project
+                if (!File.Exists(assetsPath))
+                {
+                    Console.Error.WriteLine(
+                        string.Format(
+                            CultureInfo.CurrentCulture,
+                            Strings.Error_AssetsFileNotFound,
+                            projectPath));
+
+                    ProjectCollection.GlobalProjectCollection.UnloadProject(project);
```
nkolev92: In general, if you're gonna have something that needs done at the end of a loop, I think it's better to just have it in one spot. 

I think it leads to fewer issues down the line if you don't need to copy around cleanup/recovery code.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -0,0 +1,548 @@
...
+                            CultureInfo.CurrentCulture,
+                            Strings.WhyCommand_Error_CannotReadAssetsFile,
+                            assetsPath));
+
+                    ProjectCollection.GlobalProjectCollection.UnloadProject(project);
+                    continue;
+                }
+
+                // get all resolved package references for a project
+                List<FrameworkPackages> packages = msBuild.GetResolvedVersions(project, whyCommandArgs.Frameworks, assetsFile, transitive: true);
```
nkolev92: I think if you use some methods to separate this part from the assets file figuring out and reading part, it'd enable more unit tests.

nkolev92: Seeing this GetResolvedVersions method now, I was wondering why it needs a project, but it probably doesn't, it just uses the full path off of that. 
I think long term we need a better location for that method, since it's not really an MSBUild specific ut

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -0,0 +1,548 @@
...
+            IList<LockFileTarget> targets,
+            string targetPackage,
+            string projectName)
+        {
+            bool doesProjectHaveDependencyOnPackage = false;
+            var dependencyGraphPerFramework = new Dictionary<string, List<DependencyNode>>(targets.Count);
+
+            foreach (var frameworkPackage in frameworkPackages)
+            {
+                LockFileTarget target = targets.FirstOrDefault(f => f.TargetFramework.GetShortFolderName() == frameworkPackage.Framework);
```
nkolev92: We should be using strings instead of NugetFramework here right? 

We should make sure that why works with maui projects, which is where you'd see these challenges. 

We can look together if you need help around this topic.

nkolev92: I think I linked a method up above that might be the one we need to use. GetLockFileTarget or something like that. 

using that might do enough.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -0,0 +1,548 @@
...
+        /// Traverses the dependency graph for a given top-level package, looking for a path to the target package.
+        /// Returns the top-level package node if a path was found, and null if no path was found.
+        /// </summary>
+        /// <param name="topLevelPackage">Top-level package.</param>
+        /// <param name="packageLibraries">All package libraries for a given framework.</param>
+        /// <param name="visited">HashSet tracking every package node that we've traversed.</param>
+        /// <param name="dependencyNodes">Dictionary tracking all packageIds that were added to the graph, mapped to their DependencyNode objects.</param>
+        /// <param name="versions">Dictionary mapping packageIds to their resolved versions.</param>
+        /// <param name="targetPackage">The package we want the dependency paths for.</param>
+        /// <returns></returns>
```
nkolev92: documentation: The guidance tends to not leave empty tags. I'd just document this here. 

Apply this to any other places in this changeset.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatWhyTests.cs
```diff
@@ -0,0 +1,122 @@
...
+    public class XPlatWhyTests
+    {
+        private static readonly string ProjectName = "Test.Project.DotnetNugetWhy";
+        private static MSBuildAPIUtility MsBuild => new MSBuildAPIUtility(new TestCommandOutputLogger());
+
+        [Fact]
+        public async void WhyCommand_BasicFunctionality_Succeeds()
+        {
+            // Arrange
+            using (var pathContext = new SimpleTestPathContext())
```
donnie-msft: nit: consider inlining the using statement to reduce indentation

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/IWhyCommandRunner.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+
+namespace NuGet.CommandLine.XPlat
+{
+    internal interface IWhyCommandRunner
+    {
+        Task ExecuteCommand(WhyCommandArgs whyCommandArgs);
```
donnie-msft:
```suggestion
        Task ExecuteCommandAsync(WhyCommandArgs whyCommandArgs);
```

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatWhyTests.cs
```diff
@@ -0,0 +1,122 @@
...
+
+                var whyCommandRunner = new WhyCommandRunner();
+                var whyCommandArgs = new WhyCommandArgs(
+                        project.ProjectPath,
+                        packageY.PackageName,
+                        [projectFramework],
+                        logger);
+
+                // Act
+                var result = whyCommandRunner.ExecuteCommand(whyCommandArgs);
```
donnie-msft:
```suggestion
                var result = await whyCommandRunner.ExecuteCommand(whyCommandArgs);
```

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatWhyTests.cs
```diff
@@ -0,0 +1,122 @@
...
+
+                // Assert
+                var output = logger.ShowMessages();
+
+                Assert.Equal(System.Threading.Tasks.Task.CompletedTask, result);
```
donnie-msft: This can go away if you agree with my suggestion to `await` above. If not, please help me understand why not to await in this case.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatWhyTests.cs
```diff
@@ -0,0 +1,122 @@
...
+namespace NuGet.XPlat.FuncTest
+{
+    [Collection("NuGet XPlat Test Collection")]
+    public class XPlatWhyTests
+    {
+        private static readonly string ProjectName = "Test.Project.DotnetNugetWhy";
+        private static MSBuildAPIUtility MsBuild => new MSBuildAPIUtility(new TestCommandOutputLogger());
+
+        [Fact]
+        public async void WhyCommand_BasicFunctionality_Succeeds()
```
donnie-msft: I prefer names that are more explicit. Just a suggestion:
```suggestion
        public async void WhyCommand_WhenProjectHasDependencyGraph_Succeeds()
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -912,4 +912,46 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
...
     <value>Gets the NuGet configuration settings that will be applied.</value>
   </data>
+  <data name="WhyCommand_Description" xml:space="preserve">
+    <value>Lists the dependency graph for a particular package for a given project or solution.</value>
+  </data>
+  <data name="WhyCommand_PathArgument_Description" xml:space="preserve">
+    <value>A path to a project, solution file or directory.</value>
+  </data>
+  <data name="WhyCommand_PackageArgument_Description" xml:space="preserve">
+    <value>The package name to look-up in the dependency graph.</value>
```
donnie-msft:
```suggestion
    <value>The package name to lookup in the dependency graph.</value>
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -912,4 +912,46 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
   <data name="ConfigGetCommandDescription" xml:space="preserve">
     <value>Gets the NuGet configuration settings that will be applied.</value>
   </data>
+  <data name="WhyCommand_Description" xml:space="preserve">
+    <value>Lists the dependency graph for a particular package for a given project or solution.</value>
+  </data>
+  <data name="WhyCommand_PathArgument_Description" xml:space="preserve">
+    <value>A path to a project, solution file or directory.</value>
```
donnie-msft: nit
```suggestion
    <value>A path to a project, solution file, or directory.</value>
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -912,4 +912,46 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
...
+  <data name="WhyCommand_Error_CannotReadAssetsFile" xml:space="preserve">
+    <value>Unable to read the assets file `{0}`.</value>
+    <comment>{0} - Assets file path</comment>
+  </data>
+  <data name="WhyCommand_Message_NoPackagesFoundForGivenFrameworks" xml:space="preserve">
+    <value>No packages were found for the project '{0}' for the given framework(s).</value>
+    <comment>{0} - Project name</comment>
+  </data>
+  <data name="WhyCommand_Message_NoDependencyGraphsFoundInProject" xml:space="preserve">
+    <value>Project '{0}' does not have any dependency graph(s) for '{1}'</value>
```
donnie-msft:
```suggestion
    <value>Project '{0}' does not have any dependency graph(s) for '{1}'.</value>
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -912,4 +912,46 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
...
+    <comment>{0} - Project name, {1} - Target package</comment>
+  </data>
+  <data name="WhyCommand_Message_DependencyGraphsFoundInProject" xml:space="preserve">
+    <value>Project '{0}' has the following dependency graph(s) for '{1}':</value>
+    <comment>{0} - Project name, {1} - Target package</comment>
+  </data>
+  <data name="WhyCommand_Message_NoDependencyGraphsFoundForFramework" xml:space="preserve">
+    <value>No dependency graph(s) found</value>
```
donnie-msft:
```suggestion
    <value>No dependency graph(s) found.</value>
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -912,4 +912,46 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
...
+  <data name="WhyCommand_Error_ArgumentCannotBeEmpty" xml:space="preserve">
+    <value>Unable to run 'dotnet nuget why'. The '{0}' argument cannot be empty.</value>
+    <comment>{0} - Argument that was not provided</comment>
+  </data>
+  <data name="WhyCommand_Error_PathIsMissingOrInvalid" xml:space="preserve">
+    <value>Unable to run 'dotnet nuget why'. Missing or invalid project/solution file '{0}'.</value>
+    <comment>{0} - Project/solution file path</comment>
+  </data>
+  <data name="WhyCommand_Error_InvalidFramework" xml:space="preserve">
+    <value>Failed to parse one of the given frameworks. Please make sure the given frameworks are valid.</value>
```
donnie-msft: What about pointing to the CLI reference docs in this scenario, similar to [CommandInvalidArgumentException](https://github.com/NuGet/NuGet.Client/blob/19d9b0f57a4355e8398c0f6e8895053888f9fd88/src/NuGet.Clients/NuGet.CommandLine/NuGetCommand.resx#L738) ?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -0,0 +1,548 @@
...
+        public Task ExecuteCommand(WhyCommandArgs whyCommandArgs)
+        {
+            var msBuild = new MSBuildAPIUtility(whyCommandArgs.Logger);
+
+            string targetPackage = whyCommandArgs.Package;
+
+            IEnumerable<string> projectPaths = Path.GetExtension(whyCommandArgs.Path).Equals(".sln")
+                                                    ? MSBuildAPIUtility.GetProjectsFromSolution(whyCommandArgs.Path).Where(f => File.Exists(f))
+                                                    : new List<string>() { whyCommandArgs.Path };
```
kartheekp-ms: can we specify capacity 1 while creating a new list object?

```cs
new List<string>(1);
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -0,0 +1,548 @@
...
+
+            foreach (var projectPath in projectPaths)
+            {
+                Project project = MSBuildAPIUtility.GetProject(projectPath);
+
+                // if the current project is not a PackageReference project, print an error message and continue to the next project
```
kartheekp-ms: I think this comment is redundant because the if block logic is easy to understand.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -0,0 +1,548 @@
...
+        /// <param name="whyCommandArgs">CLI arguments for the 'why' command.</param>
+        /// <returns></returns>
+        public Task ExecuteCommand(WhyCommandArgs whyCommandArgs)
+        {
+            var msBuild = new MSBuildAPIUtility(whyCommandArgs.Logger);
+
+            string targetPackage = whyCommandArgs.Package;
+
+            IEnumerable<string> projectPaths = Path.GetExtension(whyCommandArgs.Path).Equals(".sln")
+                                                    ? MSBuildAPIUtility.GetProjectsFromSolution(whyCommandArgs.Path).Where(f => File.Exists(f))
```
donnie-msft, line 36: I'm guessing that MSBuild only returns project files which exist. @jeffkl would know for sure.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -0,0 +1,548 @@
...
+                                                    : new List<string>() { whyCommandArgs.Path };
+
+            foreach (var projectPath in projectPaths)
+            {
+                Project project = MSBuildAPIUtility.GetProject(projectPath);
```
donnie-msft, line 41: Maybe we can enhance the MSBuildAPIUtility to just return the ProjectAndSolution type instead of selecting only the paths in `GetProjectsFromSolution`. That way, you wouldn't need convert the paths back to projects, here. 

I'm not familiar with these types, so let's see if @jeffkl has opinions on this.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -0,0 +1,548 @@
...
+                            assetsPath));
+
+                    ProjectCollection.GlobalProjectCollection.UnloadProject(project);
+                    continue;
+                }
+
+                // get all resolved package references for a project
+                List<FrameworkPackages> packages = msBuild.GetResolvedVersions(project, whyCommandArgs.Frameworks, assetsFile, transitive: true);
+
+                if (packages?.Count > 0)
```
donnie-msft: Packages can't be null.
```suggestion
                if (packages.Count > 0)
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -0,0 +1,548 @@
...
+                else
+                {
+                    currentPrefix = current.Prefix + ChildNodeSymbol;
+                    childPrefix = current.Prefix + ChildPrefixSymbol;
+                }
+
+                // print current node
+                if (current.Node.Id == targetPackage)
+                {
+                    /*
```
donnie-msft: 👀

donnie-msft (commented):
Really cool seeing this come together. Nice work!

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
kartheekp-ms, line 1: I think this class has multiple responsibilities such as performing graph walk to identify the paths that lead to the given package, format and print to the output. Please consider refactoring this class. I am happy to brainstorm ideas.

kartheekp-ms (commented):
I am still reviewing this PR but left couple of comments to begin with.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatWhyTests.cs
jebriede: I agree and I'd also suggest adding a more descriptive suffix explaining what it means for this test to succeed.

Example:
Let's say we have a test for an indexOf method and we call our test IndexOf_WhenCharacterExistsInString_Succeeds, we can instead name it something more descriptive like what it means for this test to succeed. Something like this:
IndexOf_WhenCharacterExistsInString_ReturnsFirstFoundPosition

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand.cs
```diff
@@ -0,0 +1,107 @@
...
+            if (string.IsNullOrEmpty(path.Value))
+            {
+                throw new ArgumentException(
+                    string.Format(CultureInfo.CurrentCulture,
+                    Strings.WhyCommand_Error_ArgumentCannotBeEmpty,
+                    path.Name));
+            }
+
+            if (!File.Exists(path.Value)
+                || (!path.Value.EndsWith("proj", StringComparison.OrdinalIgnoreCase) && !path.Value.EndsWith(".sln", StringComparison.OrdinalIgnoreCase)))
```
jebriede: If we're splitting the line based on logical operators, it makes it easier to read if it's split on all of them.

```suggestion
                || (!path.Value.EndsWith("proj", StringComparison.OrdinalIgnoreCase)
                && !path.Value.EndsWith(".sln", StringComparison.OrdinalIgnoreCase)))
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand.cs
advay26: I've changed this to CliCommand now. Please let me know if it doesn't look right or needs further changes

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand.cs
advay26: I'm using this GetTarget() method in the actual execution code in `WhyCommandRunner.cs` now.

Here, I've still left the validation as it was, as it just attempts to parse the input frameworks string as a NuGetFramework to see if it's a supported framework or not.

When we do this validation, it doesn't actually change the inputs, and we run the 'why' command with the original input strings.

@nkolev92 I might have misunderstood what you meant here, so please let me know and I can remove this validation and make changes wherever necessary.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
advay26: I've refactored this method, and we only unload at a single point in the code now.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
advay26: I've attempted to refactor this somewhat, such that we're now reading and validating the assets file separately, and then running the command with all the loaded inputs separately. Let me know if you'd like to see this refactored further, and I'd be happy to go back and make changes!

advay26: This method does currently use the project in other places too, like https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs#L763, but I'll reach out to you offline to see if there's ways for us to refactor this method.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
advay26: I'm now using the GetTarget() method that you referred to above.

The code here also uses the `GetResolvedVersions()` method from MSBuildAPIUtility, and that extracts information from the assets file's targets based on user input framework strings. This code might need changes too, but I wasn't sure how to implement them: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs#L686-L712

I'll reach out to you offline to see what changes this might need.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
advay26: Thanks! I've fixed this now.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/IWhyCommandRunner.cs
advay26: I realized that I might have been using a `Task` return type unnecessarily here. I don't believe this method needs to be async at all, but I could be wrong. I've now changed it to return an `int` directly, instead of a `Task`. Let me know if I've got this wrong

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatWhyTests.cs
advay26: This method doesn't need to be async, so we shouldn't have to do this here. See https://github.com/NuGet/NuGet.Client/pull/5761#discussion_r1598583957. Let me know if I've misunderstood something!

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatWhyTests.cs
advay26: The method now returns an `int` directly, instead of a `Task`, so this Assert statement checks for the exit code instead.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatWhyTests.cs
advay26: Thanks! I've taken another run at naming these tests. Please let me know if you still think it could do be better, and I'd be happy to make further changes!

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
advay26: I've updated the error message to point to https://learn.microsoft.com/en-us/nuget/reference/target-frameworks#supported-frameworks

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
advay26: `msBuild.GetResolvedVersions` says it returns null on errors, so these 'packages' (`List<FrameworkPackages>`) may well be null here in some cases: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs#L661

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
advay26: Forgot to remove this! Fixed it now!

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
advay26: I'm using `: [whyCommandArgs.Path]` here instead now. https://github.com/NuGet/NuGet.Client/blob/5c3ad520829d9e6adac8ad5f0cae2eb7ec41ea8c/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs#L54

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
advay26: I've broken up some of the methods into smaller pieces in response to some other feedback, but I'm happy to refactor this further. I've reached out to you to discuss ideas offline

src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs
```diff
@@ -74,18 +74,27 @@ public static int MainInternal(string[] args, CommandOutputLogger log)
...
                     log.HidePrefixForInfoAndMinimal = true;
                     return log;
                 };
 
                 CliCommand rootCommand = new CliCommand("package");
+
+                if (args[0] == "why")
+                {
+                    rootCommand = new CliCommand("nuget");
+                }
```
jebriede: The code in this method is starting to become a bit cumbersome. It would benefit from a more scalable design. With the current approach, we set up the "package" CliCommand, but then override after checking again if the arg is "why". We briefly discussed offline a potential design and we can follow up on that in more detail. I'm happy to help with ideas. That said, you're following an existing design here and maybe it should be improved upon in a follow-up PR or separate one before this implementation.

src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs
```diff
@@ -74,18 +74,27 @@ public static int MainInternal(string[] args, CommandOutputLogger log)
...
-                // We are executing command `dotnet package search`
                 Func<ILoggerWithColor> getHidePrefixLogger = () =>
                 {
                     log.HidePrefixForInfoAndMinimal = true;
                     return log;
                 };
 
                 CliCommand rootCommand = new CliCommand("package");
+
+                if (args[0] == "why")
```
jebriede: I don't quite understand how "why" is the first argument (arg[0]) when the command is `dotnet nuget why` while "package" is the arg[0] when the command is `dotnet package search`. Wouldn't "why" be arg[1] or is there some other reason why "nuget" is removed from the args? If that's the case, maybe a comment could help explain that as this is a bit confusing.

src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs
```diff
@@ -74,18 +74,27 @@ public static int MainInternal(string[] args, CommandOutputLogger log)
...
+                {
+                    rootCommand = new CliCommand("nuget");
+                }
+
                 PackageSearchCommand.Register(rootCommand, getHidePrefixLogger);
                 ConfigCommand.Register(rootCommand, getHidePrefixLogger);
+                WhyCommand.Register(rootCommand, getHidePrefixLogger);
```
jebriede: If the command is a "why" command, we are still registering package search and config as sub commands of the "nuget" CliCommand (because that's what's created when we encounter "why" as arg[0]). This doesn't sound right to me. Should we only be registering the package search and config as subcommands of the root command if this wasn't a Why command? Or are we supposed to be registering the sub commands in all cases for some reason (maybe for --help?), but to the appropriate base command? If so, then we have a bug here, because it looks like when we have a "nuget why" command we are registering "search" as a sub-command of "nuget".

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -0,0 +1,606 @@
...
+
+            foreach (var projectPath in projectPaths)
+            {
+                Project project = MSBuildAPIUtility.GetProject(projectPath);
+                LockFile assetsFile = GetProjectAssetsFile(project, whyCommandArgs.Logger);
+
+                if (assetsFile != null)
+                {
+                    FindAllDependencyGraphs(whyCommandArgs, msBuild, project, assetsFile);
+                }
```
zivkan: Add a comment saying that GetProjectAssetsFile logs errors as needed. I originally wrote a comment here asking why error conditions were being ignored, so while I think that it's a fine design for the method to return null for errors, it's not intuitive when reading this code that calls the method.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -0,0 +1,606 @@
...
+            }
+        }
+
+        private static void ValidateFrameworksOption(List<string> frameworks)
+        {
+            var parsedFrameworks = frameworks.Select(f =>
+                                    NuGetFramework.Parse(
+                                        f.Split(new char[] { '/' }, StringSplitOptions.RemoveEmptyEntries)
+                                            .Select(s => s.Trim())
+                                            .ToArray()[0]));
```
zivkan: Why would a customer provide a framework with a `/`?  It's not intuitive to me.

In the assets file, and maybe some other places, we use the syntax "tfm/rid", but I would not expect a customer to do `dotnet nuget why -f net8.0/win-x64`. In other commands, like build, if you want to specific a rid, you need to use the seprate `--rid` or `-r` option.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -0,0 +1,606 @@
...
+        }
+
+        /// <summary>
+        /// Finds dependency graphs for a given project, and prints output to the console.
+        /// </summary>
+        /// <param name="whyCommandArgs">CLI arguments for the 'why' command</param>
+        /// <param name="msBuild">MSBuild utility</param>
+        /// <param name="project">Current project</param>
+        /// <param name="assetsFile">Assets file for current project</param>
+        private static void FindAllDependencyGraphs(
```
zivkan: naming suggestion:  I expect a method named `Find` to return a data structure with the 'search' results.  Perhaps `OutputAllDependencyGraphs` better sets expectations.

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -912,4 +912,46 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
   <data name="ConfigGetCommandDescription" xml:space="preserve">
     <value>Gets the NuGet configuration settings that will be applied.</value>
   </data>
+  <data name="WhyCommand_Description" xml:space="preserve">
+    <value>Lists the dependency graph for a particular package for a given project or solution.</value>
```
zivkan:
```suggestion
    <value>Shows the dependency graph for a particular package for a given project or solution.</value>
```
```suggestion
    <value>Lists the dependency graph for a particular package for a given project or solution.</value>
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -912,4 +912,46 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
...
+  <data name="WhyCommand_Error_ArgumentCannotBeEmpty" xml:space="preserve">
+    <value>Unable to run 'dotnet nuget why'. The '{0}' argument cannot be empty.</value>
+    <comment>{0} - Argument that was not provided</comment>
+  </data>
+  <data name="WhyCommand_Error_PathIsMissingOrInvalid" xml:space="preserve">
+    <value>Unable to run 'dotnet nuget why'. Missing or invalid project/solution file '{0}'.</value>
+    <comment>{0} - Project/solution file path</comment>
+  </data>
+  <data name="WhyCommand_Error_InvalidFramework" xml:space="preserve">
+    <value>Failed to parse one of the given frameworks. Please make sure the given frameworks are valid. For a list of accepted values, please visit https://learn.microsoft.com/en-us/nuget/reference/target-frameworks#supported-frameworks.</value>
```
zivkan: some shells detect URLs, and make them clickable. Some of those shells don't trim off trailing periods, making the URL invalid when opened in the browser.

```suggestion
    <value>Failed to parse one of the given frameworks. Please make sure the given frameworks are valid. For a list of accepted values, please visit https://learn.microsoft.com/en-us/nuget/reference/target-frameworks#supported-frameworks</value>
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -912,4 +912,46 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
...
+  </data>
+  <data name="WhyCommand_Error_CannotReadAssetsFile" xml:space="preserve">
+    <value>Unable to read the assets file `{0}`.</value>
```
zivkan: Does this error get appended to other errors?  because by itself, it's not actionable by customers.

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -912,4 +912,46 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
...
+    <comment>{0} - Assets file path</comment>
+  </data>
+  <data name="WhyCommand_Message_NoPackagesFoundForGivenFrameworks" xml:space="preserve">
+    <value>No packages were found for the project '{0}' for the given framework(s).</value>
+    <comment>{0} - Project name</comment>
+  </data>
+  <data name="WhyCommand_Message_NoDependencyGraphsFoundInProject" xml:space="preserve">
+    <value>Project '{0}' does not have any dependency graph(s) for '{1}'.</value>
```
zivkan: I find this message confusing, and I've been working on NuGet for years 😕 

Does it mean that the package wasn't found in the dependency graph for a specific TFM?  If so, we can make this message much more clear to customers.

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -912,4 +912,46 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
...
+  </data>
+  <data name="WhyCommand_Error_InvalidFramework" xml:space="preserve">
+    <value>Failed to parse one of the given frameworks. Please make sure the given frameworks are valid. For a list of accepted values, please visit https://learn.microsoft.com/en-us/nuget/reference/target-frameworks#supported-frameworks.</value>
+  </data>
+  <data name="WhyCommand_Error_CannotReadAssetsFile" xml:space="preserve">
+    <value>Unable to read the assets file `{0}`.</value>
+    <comment>{0} - Assets file path</comment>
+  </data>
+  <data name="WhyCommand_Message_NoPackagesFoundForGivenFrameworks" xml:space="preserve">
+    <value>No packages were found for the project '{0}' for the given framework(s).</value>
```
zivkan: Given that `dotnet nuget why` is about searching for a specific package, does the "this project doesn't have any packages" need a specific message?  It could be grouped in a "Package 'x' is not used for the given framework(s)."

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
zivkan: I think it would be good to tell the customer which framework string couldn't be parsed, in case they provided multiple. It makes the error message much more actionable.

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -912,4 +912,46 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
...
+  <data name="WhyCommand_Message_NoDependencyGraphsFoundInProject" xml:space="preserve">
+    <value>Project '{0}' does not have any dependency graph(s) for '{1}'.</value>
+    <comment>{0} - Project name, {1} - Target package</comment>
+  </data>
+  <data name="WhyCommand_Message_DependencyGraphsFoundInProject" xml:space="preserve">
+    <value>Project '{0}' has the following dependency graph(s) for '{1}':</value>
+    <comment>{0} - Project name, {1} - Target package</comment>
+  </data>
+  <data name="WhyCommand_Message_NoDependencyGraphsFoundForFramework" xml:space="preserve">
+    <value>No dependency graph(s) found.</value>
```
zivkan: I'm not sure what scenario this is referring to without searching the code, so I expect customers won't understand what the message means either.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -0,0 +1,606 @@
...
+            WhyCommandArgs whyCommandArgs,
+            MSBuildAPIUtility msBuild,
+            Project project,
+            LockFile assetsFile)
+        {
+            // get all resolved package references for a project
+            List<FrameworkPackages> frameworkPackages = msBuild.GetResolvedVersions(project, whyCommandArgs.Frameworks, assetsFile, transitive: true, includeProjectReferences: true);
```
zivkan: The assets file already has a list of direct package references, and project references, for a project. It's not intuitive to me why we'd get it from the msbuild API, or why we'd ask the MSBuild API to recurse through project references.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -0,0 +1,606 @@
...
+                else
+                {
+                    logger.LogMinimal($"{currentPrefix}{current.Node.Id} (v{current.Node.Version})");
+                }
+
+                if (current.Node.Children?.Count > 0)
+                {
+                    // push all the node's children onto the stack
+                    counter = 0;
+                    foreach (var child in current.Node.Children)
```
zivkan: I think the children should be enumerated in StringComparison.CurrentCulture order, so that the packages are listed in alphabetical order.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -0,0 +1,606 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
```
zivkan, line 6: While many of the classes this class will depend on don't have nullable annotations (and there we won't get full benefit), it might be good to add `#nullable enable` in this new file (honestly, all new files), to reduce the risk of null reference exceptions at runtime.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -0,0 +1,606 @@
...
+                    Strings.WhyCommand_Error_InvalidFramework));
+            }
+        }
+
+        /// <summary>
+        /// Validates and returns the assets file for the given project.
+        /// </summary>
+        /// <param name="project">Evaluated MSBuild project</param>
+        /// <param name="logger">Logger for the 'why' command</param>
+        /// <returns>Assets file for given project</returns>
```
zivkan: The code appears to return null in error condtions, but these xmldoc comments don't explain that.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -0,0 +1,606 @@
...
+        /// </summary>
+        /// <param name="frameworks">The list of frameworks that share this dependency graph.</param>
+        /// <param name="topLevelNodes">The top-level package nodes of the dependency graph.</param>
+        /// <param name="targetPackage">The package we want the dependency paths for.</param>
+        private static void PrintDependencyGraphPerFramework(List<string> frameworks, List<DependencyNode> topLevelNodes, string targetPackage, ILoggerWithColor logger)
+        {
+            // print framework header
+            foreach (var framework in frameworks)
+            {
+                logger.LogMinimal($"\t[{framework}]");
```
zivkan: As far as I can tell, different terminals can use a different number of spaces for a tab character. So, using `\t` will have different results depending on what terminal software each customer is using.  Perhaps we should use spaces instead, so it's more consistent?

Looking at your screenshots, I think the graphs are too much indented. For deep graphs, or packages wil long names, it's at increased risk of wrapping, so using less indenting for our own formatting reduces the risk.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand.cs
nkolev92: The user provided input must not be parsed at all. 

It should be a string match.

src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs
jeffkl: Team Review:  We agree that the registration of the `why` command should be in the `if` and the other commands in this method should be registered in an `else` instead.

jeffkl: Existing tech debt issues to address this in the future:
https://github.com/NuGet/Home/issues/13089
https://github.com/NuGet/Home/issues/11997
https://github.com/NuGet/Home/issues/11996

src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs
martinrrm: Yes, we register all the commands for the --help, or if the user typed the command wrong it will also invoke the help option. The other commands are registered at runtime too, I think https://github.com/NuGet/Home/issues/13089 could help with this.

zivkan (commented):
Oops, I forgot to submit these comments yesterday.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
zivkan: You could consider splitting the file, so methods related to outputting to the console is in one file. Methods related to creating the data structure from the assets file is a in a different file. Then main `WhyCommandRunner` class will then be a fairly light class that orchostrates the two other classes that do the main logic.  This can also help think of ways to write unit tests, rather than slow integration tests.

src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs
advay26: Looking at the previous code where all the other commands are registered, we didn't register all the commands together there. We registered all the 'dotnet package *' commands in one place, and then the 'dotnet nuget *' commands separately: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs#L239-L258

We can do something similar here too:
image
Let me know if this can be improved upon

src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs
advay26: Since the config command is also a `dotnet nuget *` command, we should probably register the 'config' and 'why' commands in the same block, and the 'search' command separately, something like this:
image

And as discussed offline, I'll add a comment to the code linking all the related tech debt issues.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
kartheekp-ms: Just to give you some ideas. I asked VS Copilot to explain all the responsibilities of this class. It gave the following response.

Here's a breakdown of the class and its methods:

- **ExecuteCommand(WhyCommandArgs whyCommandArgs):** This is the main method that executes the 'why' command. It takes in WhyCommandArgs which contains the arguments for the 'why' command. It loads the project or solution file, checks if it's a PackageReference project, reads the project's assets file, and then finds all the dependency graphs for the target package in the project. It also handles unloading the project after it's done.
- **FindAllDependencyGraphs(...):** This method finds all the dependency graphs for a given package in a project. It checks if the project has a dependency on the target package and if it does, it gets the dependency graph for each framework in the project.
- **GetDependencyGraphPerFramework(...):** This method finds all dependency paths from the top-level packages to the target package for a given framework. It uses depth-first search to find these paths.
- **FindDependencyPath(...):** This method traverses the dependency graph for a given top-level package, looking for a path to the target package. It uses a stack to keep track of the nodes it needs to visit.
- **AddToGraph(...):** This method adds a dependency path to the graph, starting from the target package and traversing up to the top-level package.
- **PrintAllDependencyGraphs(...):** This method prints the dependency graphs for all target frameworks. It first deduplicates the dependency graphs and then prints each unique graph.
- **GetDeduplicatedFrameworks(...):** This method deduplicates dependency graphs and returns groups of frameworks that share the same graph.
- **PrintDependencyGraphPerFramework(...):** This method prints the dependency graph for a given framework or list of frameworks.
- **GetDependencyGraphHashCode(...):** This method returns a hash for a given dependency graph. It's used to deduplicate dependency graphs for different frameworks.

The class also contains several private classes (DependencyNode, StackDependencyData, StackOutputData) that are used to represent nodes in the dependency graph and to keep track of data during the depth-first search and while printing the dependency graph.

As a follow-up, I asked how to refactor this class so that the new design adheres to the SOLID principles. It came up with following design.

- **Project Loading:** This responsibility involves loading the project or solution file and checking if it's a PackageReference project. This can be encapsulated in a ProjectLoader class.

```cs
public interface IProjectLoader
{
    Task<Project> LoadProject(string projectPath);
}

public class ProjectLoader : IProjectLoader
{
    // Implementation goes here
}
```
- **Dependency Graph Finding:** This responsibility involves finding all the dependency graphs for a given package in a project. This can be encapsulated in a DependencyGraphFinder class.

```cs
public interface IDependencyGraphFinder
{
    void FindAllDependencyGraphs(IEnumerable<FrameworkPackages> frameworkPackages, IList<LockFileTarget> targets, string targetPackage, string projectName);
}

public class DependencyGraphFinder : IDependencyGraphFinder
{
    // Implementation goes here
}
```
- **Dependency Graph Printing:** This responsibility involves printing the dependency graphs for all target frameworks. This can be encapsulated in a DependencyGraphPrinter class.

```cs
public interface IDependencyGraphPrinter
{
    void PrintAllDependencyGraphs(Dictionary<string, List<DependencyNode>> dependencyGraphPerFramework, string targetPackage);
}

public class DependencyGraphPrinter : IDependencyGraphPrinter
{
    // Implementation goes here
}
```

- **Command Execution:** This responsibility involves coordinating the execution of the 'why' command. This can remain in the WhyCommandRunner class, but now it will delegate the other responsibilities to the classes above.

```cs
public class WhyCommandRunner
{
    private IProjectLoader _projectLoader;
    private IDependencyGraphFinder _graphFinder;
    private IDependencyGraphPrinter _graphPrinter;

    public WhyCommandRunner(IProjectLoader projectLoader, IDependencyGraphFinder graphFinder, IDependencyGraphPrinter graphPrinter)
    {
        _projectLoader = projectLoader;
        _graphFinder = graphFinder;
        _graphPrinter = graphPrinter;
    }

    public async Task ExecuteCommand(WhyCommandArgs whyCommandArgs)
    {
        // Use _projectLoader, _graphFinder, and _graphPrinter
    }
}
```

I am not saying we should consider whatever the design Copilot recommends but it just gives an idea because we can visualize the code even before refactoring.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand.cs
advay26: I've now removed the framework parsing, and we're using the user input frameworks to get the targets from the assets file

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
advay26: @kartheekp-ms @zivkan Thank you for the suggestions! I've attempted to refactor the code, moving some of the code into 2 new classes: `DependencyGraphFinder` and `DependencyGraphPrinter`. I've also been able to write new unit tests for the Graph finder class with these changes. Please take a look and let me know if this can be improved upon

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
advay26: I've now removed the validation on input frameworks completely, and we're using the input strings as provided to get the targets from the assets file

advay26: I'm still not sure if I'm dealing with the runtime identifiers correctly, so please take a look and let me know if it looks wrong

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
advay26: After the refactoring, this method (now in the `DependencyGraphFinder` class) returns the dependency graph data structure, which is then separately output by the `DependencyGraphPrinter` class. Let me know if this doesn't look okay

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
advay26: I've changed the error message to say "The assets file {0} is invalid. Please run restore for project '{1}' before running this command."

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -0,0 +1,168 @@
...
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using Microsoft.Build.Evaluation;
+using NuGet.CommandLine.XPlat.WhyCommandUtility;
+using NuGet.ProjectModel;
+
+namespace NuGet.CommandLine.XPlat
```
zivkan, line 14: nitpick: Just wondering why the namespace is different to the project's default namespace + directory structure?

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
advay26: This error message now says "Project '{0}' does not have a dependency on '{1}'." Let me know if I can improve on this further

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
advay26: I've removed the framework validation since we match on target aliases instead. Now, we're not doing any validation on the input frameworks strings. Let me know if I should add any validations for these inputs

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandUtility/DependencyGraphFinder.cs
zivkan, line 1: nitpick: Many NuGet.Client projects have very poor organization, so it's good that this PR isn't adding more files into an already overloaded directory, but there's already a `WhyCommand` subdirectory that doesn't have many files in it.  I think these utilitity classes will be just fine in the `WhyCommand` directory, and doesn't need an additional subdirectory.

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
advay26: This was meant to be framework-specific, so I've changed this to "No dependency graph(s) found for this target framework."

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
advay26: Instead of using this method from the MSBuildAPIUtility class, I've written a more minimal method to get the top-level package and project references (in `DependencyGraphFinder`), and this doesn't need anything other than the AssetsFile and the user input frameworks

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
advay26: I've changed the indentation, and added some new screenshots to the PR description. Let me know if this looks good now

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -0,0 +1,168 @@
...
+                if (assetsFile != null)
+                {
+                    Dictionary<string, List<DependencyNode>?>? dependencyGraphPerFramework = DependencyGraphFinder.GetAllDependencyGraphs(
+                        assetsFile,
+                        whyCommandArgs.Package,
+                        whyCommandArgs.Frameworks);
+
+                    if (dependencyGraphPerFramework != null)
+                    {
+                        whyCommandArgs.Logger.LogMinimal(
```
nkolev92: These feel like they should be debug level instead of minimal. 

Minimal is pretty much gonna show it every time.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandUtility/DependencyGraphFinder.cs
```diff
@@ -0,0 +1,306 @@
...
+                {
+                    LockFileTarget target = assetsFile.GetTarget(targetFrameworkAlias, runtimeIdentifier: null);
+
+                    if (target != null)
+                    {
+                        // get all package libraries for the framework
+                        IList<LockFileTargetLibrary> packageLibraries = target.Libraries;
+
+                        // if the project has a dependency on the target package, get the dependency graph
+                        if (packageLibraries.Any(l => l.Name == targetPackage))
```
nkolev92: package ids are case insensitive.

Change to ignorecase comparer and please add a test (I've had regressions due to casing misses, that Andy had to fix: https://github.com/NuGet/NuGet.Client/commit/b49d8914861cf18287302ab311fc9d2566e06fe1).

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandUtility/DependencyGraphFinder.cs
```diff
@@ -0,0 +1,306 @@
...
+        /// <summary>
+        /// Finds all dependency paths from the top-level packages to the target package for a given framework.
+        /// </summary>
+        /// <param name="topLevelReferences">All top-level package and project references for the framework.</param>
+        /// <param name="packageLibraries">All package libraries for the framework.</param>
+        /// <param name="targetPackage">The package we want the dependency paths for.</param>
+        /// <returns>
+        /// List of all top-level package nodes in the dependency graph.
+        /// </returns>
+        private static List<DependencyNode>? GetDependencyGraphPerFramework(
```
nkolev92: naming: this doesn't make it obvious that we're getting the dependency paths for a particular package. 

It makes me think we're getting an object that represents the full graph.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandUtility/DependencyGraphFinder.cs
```diff
@@ -0,0 +1,306 @@
...
+            var stack = new Stack<StackDependencyData>();
+            stack.Push(new StackDependencyData(topLevelPackage, null));
+
+            while (stack.Count > 0)
+            {
+                var currentPackageData = stack.Pop();
+                var currentPackageId = currentPackageData.Id;
+
+                // if we reach the target node, or if we've already traversed this node and found dependency paths, add it to the graph
+                if (currentPackageId == targetPackage
```
nkolev92: Similar as earlier, case insensitive

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandUtility/DependencyGraphFinder.cs
```diff
@@ -0,0 +1,306 @@
...
+                    || dependencyNodes.ContainsKey(currentPackageId))
```
nkolev92: Dictionary looks ups as well. 

We need to make sure the comparer is the right one.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandUtility/DependencyGraphFinder.cs
```diff
@@ -0,0 +1,306 @@
...
+            // get all top-level package and project references for the project, categorized by target framework alias
+            Dictionary<string, List<string>> topLevelReferencesByFramework = GetTopLevelPackageAndProjectReferences(assetsFile, userInputFrameworks);
+
+            if (topLevelReferencesByFramework.Count > 0)
+            {
+                foreach (var (targetFrameworkAlias, topLevelReferences) in topLevelReferencesByFramework)
+                {
+                    LockFileTarget target = assetsFile.GetTarget(targetFrameworkAlias, runtimeIdentifier: null);
+
+                    if (target != null)
```
nkolev92: Is the idea that we would've validated that this target exists in the inputs? 

I think we might just want to assume it exists here and let is NRE in case made a mistake earlier. 

This potentially hides it.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/DependencyNode.cs
```diff
@@ -0,0 +1,36 @@
...
+    internal class DependencyNode
+    {
+        public string Id { get; set; }
+        public string Version { get; set; }
+        public HashSet<DependencyNode> Children { get; set; }
+
+        public DependencyNode(string id, string version)
+        {
+            Id = id;
+            Version = version;
```
jebriede:
```suggestion
            Id = id ?? throw new ArgumentNullException(nameof(id));
            Version = version ?? throw new ArgumentNullException(nameof(version));;
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -0,0 +1,168 @@
...
+            try
+            {
+                ValidatePathArgument(whyCommandArgs.Path);
+                ValidatePackageArgument(whyCommandArgs.Package);
+            }
+            catch (ArgumentException ex)
+            {
+                whyCommandArgs.Logger.LogError(ex.Message);
+                return ExitCodes.InvalidArguments;
+            }
```
jebriede: It's better to avoid throwing an exception from the private validation methods and catching it in the calling method. Since this is the same class, we don't need an exception to indicate that the validation failed. The validations could simply return a boolean and/or an error message or log the error themselves. Throwing an exception has additional overhead.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandUtility/DependencyGraphFinder.cs
```diff
@@ -0,0 +1,306 @@
...
+                }
+
+                if (i > 0)
+                {
+                    var childNode = dependencyNodes[dependencyPath[i - 1]];
+
+                    if (dependencyNodes[currentPackageId].Children.Any(p => p.Id == childNode.Id))
+                    {
+                        continue;
+                    }
```
jebriede: Children is a HashSet which enforces uniqueness, so you don't need to first check if the key exists before adding it. The HashSet takes care of that. https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1.add?view=net-8.0

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -912,4 +912,42 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
...
+    <value>Unable to run 'dotnet nuget why'. The '{0}' argument cannot be empty.</value>
+    <comment>{0} - Argument that was not provided</comment>
+  </data>
+  <data name="WhyCommand_Error_PathIsMissingOrInvalid" xml:space="preserve">
+    <value>Unable to run 'dotnet nuget why'. Missing or invalid project/solution file '{0}'.</value>
+    <comment>{0} - Project/solution file path</comment>
+  </data>
+  <data name="WhyCommand_Error_InvalidFramework" xml:space="preserve">
+    <value>Failed to parse one of the given frameworks. Please make sure the given frameworks are valid. For a list of accepted values, please visit https://learn.microsoft.com/en-us/nuget/reference/target-frameworks#supported-frameworks</value>
+  </data>
```
zivkan: Is this string still used? I think it can be deleted.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
advay26: We're currently printing **"project {0} has the following dependency graph for package {1}:"** followed by its graph. This is the only place that shows the current project's name, so it makes sense to show this for all levels of output.

The next print statement here says **"project {0} does not have a dependency on package {1}."** We could potentially change this output to LogDebug, but in larger solutions, it likely still helps to be explicit about which projects have a dependency graph and which don't.

Spoke to @nkolev92 offline, and we decided to keep this as it is for now, unless we get further feedback that we should change it.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandUtility/DependencyGraphFinder.cs
advay26: Fixed, and added a unit test for it too

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandUtility/DependencyGraphFinder.cs
advay26: I've changed the name to GetDependencyGraph<ins>ForTarget</ins>PerFramework. Let me know if this sounds better

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandUtility/DependencyGraphFinder.cs
advay26: Fixed

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandUtility/DependencyGraphFinder.cs
advay26: Fixed this where we first declare the HashSet

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandUtility/DependencyGraphFinder.cs
advay26: We're now validating that the input frameworks have a corresponding target at an earlier point, so I've removed this check from here. I've also added a test for the warning output from the validation

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/DependencyNode.cs
advay26: Fixed

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
advay26: I think I've addressed this now with the changes. Please let me know if it doesn't look right or needs anything further

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandUtility/DependencyGraphFinder.cs
advay26: I needed to implement a custom comparer for this to work right, which I've now done. I've removed this .Any() check from here now as it's no longer needed.

zivkan (commented):
I forgot to submit these comments last week!   Maybe some of them are no longer relevant due to updates made since I wrote the comments locally.

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
advay26: You're right, this wasn't needed anymore so I've removed it. We are validating frameworks again earlier in the code, where we check if the assets file has a corresponding target for each of the input framework aliases, so I created a new warning message called "_WhyCommand_Warning_AssetsFileDoesNotContainSpecifiedTarget_".

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandUtility/DependencyGraphFinder.cs
advay26: That's fair. I've now moved these 2 utility files into the `WhyCommand` subdirectory alongside the other files.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
advay26: All commands and files in this project fall in the `NuGet.CommandLine.XPlat` namespace, so I've removed other namespaces like `NuGet.CommandLine.XPlat.WhyCommandUtility` and moved everything under `NuGet.CommandLine.XPlat` to keep things consistent. Let me know if you think this needs changes or can be improved upon

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
advay26: We're not using the `GetResolvedVersions()` method any longer, so the return type doesn't need a null check here anymore. We're just checking `.Count > 0` now.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
zivkan: IMO the reasons to organize code on the filesystem is identical to the reasons to organize code in namespaces. Therefore, I consider it bad when they're out of sync. But I recognise that NuGet's existing code does this. I just think it's a bad pattern that we should stop copying. But I'm not going to hold up this PR because of this, and I don't know if anyone else in the team share my opinion. But for what it's worth, since this file's class is internal, there's no public API risk of moving it to a "more appropriate" namespace.

zivkan (approved):

nkolev92 (approved):

PR #5763, opened 2024-04-24
Create OwnerDetailsUriTemplateResourceV3 and provider

### LINKED ISSUE #13409: Consume the OwnerDetailsUriTemplate resource
### NuGet Product(s) Affected

NuGet.exe, Visual Studio Package Management UI, Visual Studio Package Manager Console, dotnet.exe, NuGet SDK

### Current Behavior

The OwnerDetailsUriTemplate resource is a new feature being added to NuGet.

### Desired Behavior

The OwnerDetailsUriTemplate resource is consumed by NuGet.Protocol.

### Additional Context

Part of https://github.com/NuGet/Home/issues/12501
Resource documented in https://github.com/NuGet/docs.microsoft.com-nuget/pull/3282
Resource added to nuget.org service index in https://github.com/NuGet/NuGet.Services.Index/pull/235/files

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13409

## Description

**_Note: this resource is not yet consumed by the Client, but will be in https://github.com/NuGet/NuGet.Client/pull/5766_**

Adds a resource, `OwnerDetailsUriTemplateResourceV3`, similar to the [PackageDetailsUriResourceV3](https://github.com/NuGet/NuGet.Client/blob/784274cc9b2002ca73055e8134064261118d8a01/src/NuGet.Core/NuGet.Protocol/Resources/PackageDetailsUriResourceV3.cs).
- A template must use a Uri scheme of HTTPS and be an Absolute Uri.
- `GetUri` will replace a template's substring, `{owner}`, with each Owner returned from the NuGet Search API.
  - Client will not enforce that the template must have this substring. The linked documentation change will suggest that practice.

A new resource provider, `OwnerDetailsUriResourceV3Provider`, retrieves the resource from the service index, if available.

I've tested in Visual Studio by consuming this resource from package source, https://apidev.nugettest.org/v3/index.json, which @martinrrm has been developing for nuget.org (Thank you! 🙏).

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [x] Documentation PR or issue filled https://github.com/NuGet/docs.microsoft.com-nuget/pull/3282
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Protocol/Providers/OwnerDetailsUriResourceV3Provider.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Providers/OwnerDetailsUriResourceV3Provider.cs
@@ -29,8 +29,12 @@ namespace NuGet.Protocol.Providers
             ServiceIndexResourceV3? serviceIndex = await source.GetResourceAsync<ServiceIndexResourceV3>(token);
             if (serviceIndex != null)
             {
-                Uri uriTemplate = serviceIndex.GetServiceEntryUri(ServiceTypes.OwnerDetailsUriTemplate);
-                resource = OwnerDetailsUriTemplateResourceV3.CreateOrNull(uriTemplate);
+                Uri? uriTemplate = serviceIndex.GetServiceEntryUri(ServiceTypes.OwnerDetailsUriTemplate);
+
+                if (uriTemplate != null)
+                {
+                    resource = OwnerDetailsUriTemplateResourceV3.CreateOrNull(uriTemplate);
+                }
             }
 
             return new Tuple<bool, INuGetResource?>(resource != null, resource);
--- a/src/NuGet.Core/NuGet.Protocol/Providers/OwnerDetailsUriResourceV3Provider.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Providers/OwnerDetailsUriResourceV3Provider.cs
@@ -1,0 +1,39 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+#nullable enable
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using NuGet.Protocol.Core.Types;
+using NuGet.Protocol.Resources;
+
+namespace NuGet.Protocol.Providers
+{
+    /// <summary>NuGet.Protocol resource provider for <see cref="OwnerDetailsUriTemplateResourceV3"/> in V3 HTTP feeds.</summary>
+    /// <remarks>When successful, returns an instance of <see cref="OwnerDetailsUriTemplateResourceV3"/>.</remarks>
+    public class OwnerDetailsUriResourceV3Provider : ResourceProvider
+    {
+        public OwnerDetailsUriResourceV3Provider()
+            : base(typeof(OwnerDetailsUriTemplateResourceV3),
+                  nameof(OwnerDetailsUriTemplateResourceV3),
+                  NuGetResourceProviderPositions.Last)
+        {
+        }
+
+        /// <inheritdoc cref="ResourceProvider.TryCreate(SourceRepository, CancellationToken)"/>
+        public override async Task<Tuple<bool, INuGetResource?>> TryCreate(SourceRepository source, CancellationToken token)
+        {
+            OwnerDetailsUriTemplateResourceV3? resource = null;
+            ServiceIndexResourceV3? serviceIndex = await source.GetResourceAsync<ServiceIndexResourceV3>(token);
+            if (serviceIndex != null)
+            {
+                Uri uriTemplate = serviceIndex.GetServiceEntryUri(ServiceTypes.OwnerDetailsUriTemplate);
+                resource = OwnerDetailsUriTemplateResourceV3.CreateOrNull(uriTemplate);
+            }
+
+            return new Tuple<bool, INuGetResource?>(resource != null, resource);
+        }
+    }
+}
--- a/src/NuGet.Core/NuGet.Protocol/Resources/OwnerDetailsUriTemplateResourceV3.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Resources/OwnerDetailsUriTemplateResourceV3.cs
@@ -15,7 +15,7 @@ namespace NuGet.Protocol.Resources
     {
         private readonly string _template;
 
-        public OwnerDetailsUriTemplateResourceV3(string template)
+        private OwnerDetailsUriTemplateResourceV3(string template)
         {
             _template = template ?? throw new ArgumentNullException(nameof(template));
         }
--- a/src/NuGet.Core/NuGet.Protocol/Resources/OwnerDetailsUriTemplateResourceV3.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Resources/OwnerDetailsUriTemplateResourceV3.cs
@@ -1,0 +1,76 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+#nullable enable
+
+using System;
+using NuGet.Protocol.Core.Types;
+
+namespace NuGet.Protocol.Resources
+{
+    /// <summary>Owner Details Uri Template for NuGet V3 HTTP feeds.</summary>
+    /// <remarks>Not intended to be created directly. Use <see cref="SourceRepository.GetResourceAsync{T}(CancellationToken)"/>
+    /// with <see cref="OwnerDetailsUriTemplateResourceV3"/> for T, and typecast to this class.
+    public class OwnerDetailsUriTemplateResourceV3 : INuGetResource
+    {
+        private readonly string _template;
+
+        public OwnerDetailsUriTemplateResourceV3(string template)
+        {
+            _template = template ?? throw new ArgumentNullException(nameof(template));
+        }
+
+        /// <summary>
+        /// Creates the specified Owner Details Uri template provided by the server if it exists and is valid.
+        /// </summary>
+        /// <param name="uriTemplate">The Absolute Uri template provided by the server.</param>
+        /// <returns>A valid Owner Details Uri template, or null.</returns>
+        public static OwnerDetailsUriTemplateResourceV3? CreateOrNull(Uri? uriTemplate)
+        {
+            if (uriTemplate == null || uriTemplate.OriginalString.Length == 0 || !uriTemplate.IsAbsoluteUri)
+            {
+                return null;
+            }
+
+            string absoluteUri = uriTemplate.OriginalString;
+            if (string.IsNullOrWhiteSpace(absoluteUri)
+                || !IsValidUriTemplate(absoluteUri))
+            {
+                return null;
+            }
+
+            return new OwnerDetailsUriTemplateResourceV3(absoluteUri);
+        }
+
+        private static bool IsValidUriTemplate(string absoluteUri)
+        {
+            Uri? uri;
+            var isValidUri = Uri.TryCreate(absoluteUri, UriKind.Absolute, out uri);
+
+            // Only allow HTTPS owner details URLs.
+            if (isValidUri && uri?.Scheme.Equals("https", StringComparison.OrdinalIgnoreCase) != true)
+            {
+                return false;
+            }
+
+            return isValidUri;
+        }
+
+        /// <summary>
+        /// Gets a URL for viewing package Owner URL outside of Visual Studio. The URL will not be verified to exist.
+        /// </summary>
+        /// <param name="owner">The owner username.</param>
+        /// <returns>The first URL from the resource, with the URI template applied.</returns>
+        public Uri GetUri(string owner)
+        {
+            var uriString = _template
+#if NETCOREAPP
+               .Replace("{owner}", owner, StringComparison.OrdinalIgnoreCase);
+#else
+               .Replace("{owner}", owner);
+#endif
+
+            return new Uri(uriString);
+        }
+    }
+}
--- a/src/NuGet.Core/NuGet.Protocol/Resources/OwnerDetailsUriTemplateResourceV3.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Resources/OwnerDetailsUriTemplateResourceV3.cs
@@ -25,35 +25,21 @@ namespace NuGet.Protocol.Resources
         /// </summary>
         /// <param name="uriTemplate">The Absolute Uri template provided by the server.</param>
         /// <returns>A valid Owner Details Uri template, or null.</returns>
-        public static OwnerDetailsUriTemplateResourceV3? CreateOrNull(Uri? uriTemplate)
+        public static OwnerDetailsUriTemplateResourceV3? CreateOrNull(Uri uriTemplate)
         {
-            if (uriTemplate == null || uriTemplate.OriginalString.Length == 0 || !uriTemplate.IsAbsoluteUri)
+            if (uriTemplate is null)
+            {
+                throw new ArgumentNullException(nameof(uriTemplate));
+            }
+
+            if (uriTemplate.OriginalString.Length == 0
+                || !uriTemplate.IsAbsoluteUri
+                || !uriTemplate.Scheme.Equals("https", StringComparison.OrdinalIgnoreCase))
             {
                 return null;
             }
 
-            string absoluteUri = uriTemplate.OriginalString;
-            if (string.IsNullOrWhiteSpace(absoluteUri)
-                || !IsValidUriTemplate(absoluteUri))
-            {
-                return null;
-            }
-
-            return new OwnerDetailsUriTemplateResourceV3(absoluteUri);
-        }
-
-        private static bool IsValidUriTemplate(string absoluteUri)
-        {
-            Uri? uri;
-            var isValidUri = Uri.TryCreate(absoluteUri, UriKind.Absolute, out uri);
-
-            // Only allow HTTPS owner details URLs.
-            if (isValidUri && uri?.Scheme.Equals("https", StringComparison.OrdinalIgnoreCase) != true)
-            {
-                return false;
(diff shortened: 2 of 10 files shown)


src/NuGet.Core/NuGet.Protocol/Resources/OwnerDetailsUriTemplateResourceV3.cs
```diff
@@ -0,0 +1,76 @@
...
+                return null;
+            }
+
+            return new OwnerDetailsUriTemplateResourceV3(absoluteUri);
+        }
+
+        private static bool IsValidUriTemplate(string absoluteUri)
+        {
+            Uri? uri;
+            var isValidUri = Uri.TryCreate(absoluteUri, UriKind.Absolute, out uri);
```
kartheekp-ms: We are converting the URI parameter value to a string in the CreateOrNull method, which then calls the IsValidUriTemplate method, passing in that string. I am not sure why we are converting that string again to verify if it is a valid URI. How about just passing the URI object that was passed to the CreateOrNull method instead of creating another URI object here?

[Security considerations section in docs](https://learn.microsoft.com/dotnet/fundamentals/runtime-libraries/system-uri#security-considerations) suggest that, You can check a URI string for validity by calling the [Uri.IsWellFormedOriginalString](https://learn.microsoft.com/en-us/dotnet/api/system.uri.iswellformedoriginalstring) method. Can this method be used in this case?

src/NuGet.Core/NuGet.Protocol/Resources/OwnerDetailsUriTemplateResourceV3.cs
```diff
@@ -0,0 +1,76 @@
...
+
+            return isValidUri;
+        }
+
+        /// <summary>
+        /// Gets a URL for viewing package Owner URL outside of Visual Studio. The URL will not be verified to exist.
+        /// </summary>
+        /// <param name="owner">The owner username.</param>
+        /// <returns>The first URL from the resource, with the URI template applied.</returns>
+        public Uri GetUri(string owner)
```
kartheekp-ms, line 50: I am assuming this public method will be used in future pull requests for this effort, correct? I don't see this method being used in this changeset, so I wanted to confirm.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/Resources/OwnerDetailsUriTemplateResourceV3Tests.cs
```diff
@@ -0,0 +1,122 @@
...
+using System;
+using FluentAssertions;
+using NuGet.Protocol.Resources;
+using Xunit;
+
+namespace NuGet.Protocol.Tests.Resources
+{
+    public class OwnerDetailsUriTemplateResourceV3Tests
+    {
+        private readonly Uri _template = new Uri("https://nuget.test/profiles/{owner}?_src=template");
```
kartheekp-ms, line 15: I am curious, why did we append `_src=template` to the URL? Is this defined to track the caller of the URL on the server side?

test/NuGet.Core.Tests/NuGet.Protocol.Tests/Resources/OwnerDetailsUriTemplateResourceV3Tests.cs
```diff
@@ -0,0 +1,122 @@
...
+
+            // Assert
+            target.Should().BeNull();
+        }
+
+        [Fact]
+        public void CreateOrNull_WhenTemplateNotHttps_CreatesNullResource()
+        {
+            // Arrange
+            var template = new Uri("http://nuget.test/profiles/{owner}?_src=template");
```
kartheekp-ms, line 44: How about using the read-only URI field declared in the class?

```cs
private readonly Uri _template = new Uri("https://nuget.test/profiles/{owner}?_src=template");
```

src/NuGet.Core/NuGet.Protocol/Resources/OwnerDetailsUriTemplateResourceV3.cs
donnie-msft: Yes and that PR is open https://github.com/NuGet/NuGet.Client/pull/5766

test/NuGet.Core.Tests/NuGet.Protocol.Tests/Resources/OwnerDetailsUriTemplateResourceV3Tests.cs
donnie-msft: You can talk to the nuget.org team about that, but my understanding is that it's a way to identify hits from Visual Studio.

src/NuGet.Core/NuGet.Protocol/Resources/OwnerDetailsUriTemplateResourceV3.cs
donnie-msft: I noticed that method as well, but unfortunately, it returns `false` because of the braces in the URI template provided by the server resource, as shown below.

My suspicion is this is why the [PackageDetailsUriResourceV3](https://github.com/NuGet/NuGet.Client/blob/784274cc9b2002ca73055e8134064261118d8a01/src/NuGet.Core/NuGet.Protocol/Resources/PackageDetailsUriResourceV3.cs#L19) does this custom validation, and it's why I copied that for the `OwnerDetailsUriTemplateResourceV3` as well. 

If you have other suggestions, let me know. I'm a little hesitant to modify the `PackageDetailsUriResourceV3` in this PR, but could consider refactoring it as well separately.

image

test/NuGet.Core.Tests/NuGet.Protocol.Tests/Resources/OwnerDetailsUriTemplateResourceV3Tests.cs
donnie-msft: Done. Good catch! Thank you

src/NuGet.Core/NuGet.Protocol/Resources/OwnerDetailsUriTemplateResourceV3.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            _template = template ?? throw new ArgumentNullException(nameof(template));
+        }
+
+        /// <summary>
+        /// Creates the specified Owner Details Uri template provided by the server if it exists and is valid.
+        /// </summary>
+        /// <param name="uriTemplate">The Absolute Uri template provided by the server.</param>
+        /// <returns>A valid Owner Details Uri template, or null.</returns>
+        public static OwnerDetailsUriTemplateResourceV3? CreateOrNull(Uri? uriTemplate)
```
nkolev92: I like that this class implements nullable. 

I think the even more complete approach would be to just be null safe and throw Uri is null, and make the call ensure a Uri is passed when available. 

I know it may seem stylistic a bit, but the general idea is you want to minimize the places in your code that can work around null values.

src/NuGet.Core/NuGet.Protocol/Providers/OwnerDetailsUriResourceV3Provider.cs
```diff
@@ -0,0 +1,39 @@
...
+        }
+
+        /// <inheritdoc cref="ResourceProvider.TryCreate(SourceRepository, CancellationToken)"/>
+        public override async Task<Tuple<bool, INuGetResource?>> TryCreate(SourceRepository source, CancellationToken token)
+        {
+            OwnerDetailsUriTemplateResourceV3? resource = null;
+            ServiceIndexResourceV3? serviceIndex = await source.GetResourceAsync<ServiceIndexResourceV3>(token);
+            if (serviceIndex != null)
+            {
+                Uri uriTemplate = serviceIndex.GetServiceEntryUri(ServiceTypes.OwnerDetailsUriTemplate);
```
nkolev92:
```suggestion
                Uri? uriTemplate = serviceIndex.GetServiceEntryUri(ServiceTypes.OwnerDetailsUriTemplate);
```

src/NuGet.Core/NuGet.Protocol/Resources/OwnerDetailsUriTemplateResourceV3.cs
```diff
@@ -0,0 +1,76 @@
...
+            if (string.IsNullOrWhiteSpace(absoluteUri)
+                || !IsValidUriTemplate(absoluteUri))
+            {
+                return null;
+            }
+
+            return new OwnerDetailsUriTemplateResourceV3(absoluteUri);
+        }
+
+        private static bool IsValidUriTemplate(string absoluteUri)
```
nkolev92: You have a Uri type in the method calling this, so you shouldn't need to pass a string. 

I'd just all the validations in 1 method and use the Uri type that's already there.

nkolev92 (commented):
Minor feedback.

src/NuGet.Core/NuGet.Protocol/Resources/OwnerDetailsUriTemplateResourceV3.cs
donnie-msft: Hm, I was trying not to invent any new validation strategy, as mentioned in https://github.com/NuGet/NuGet.Client/pull/5763#discussion_r1581410514, [PackageDetailsUriResourceV3](https://github.com/NuGet/NuGet.Client/blob/784274cc9b2002ca73055e8134064261118d8a01/src/NuGet.Core/NuGet.Protocol/Resources/PackageDetailsUriResourceV3.cs#L19) does this custom validation, and it's why I copied that for the `OwnerDetailsUriTemplateResourceV3` as well. 

I'll see if I can reduce this validation.

src/NuGet.Core/NuGet.Protocol/Resources/OwnerDetailsUriTemplateResourceV3.cs
kartheekp-ms: `CreateOrNull` method in [PackageDetailsUriResourceV3](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Protocol/Resources/PackageDetailsUriResourceV3.cs#L19) takes a `string` as parameter but in this resource the same method has `Uri` as parameter. Is this could be a reason why we are converting the Uri to string and then string to Uri for validation purposes?

src/NuGet.Core/NuGet.Protocol/Providers/OwnerDetailsUriResourceV3Provider.cs
donnie-msft: `GetServiceEntryUri` returns FirstOrDefault, therefore, it can be null. 
Changed

src/NuGet.Core/NuGet.Protocol/Resources/OwnerDetailsUriTemplateResourceV3.cs
donnie-msft: Ok, I don't like this any better than what was there, but now I'm throwing if a null is provided, and now must check for null in the Provider. We don't want providers to throw when they can't find the resource, because it's not required for a package source to implement this resource.

src/NuGet.Core/NuGet.Protocol/Resources/OwnerDetailsUriTemplateResourceV3.cs
donnie-msft: I've reduced some of the validation logic there, have a look and see if that addresses your concern.

src/NuGet.Core/NuGet.Protocol/Resources/OwnerDetailsUriTemplateResourceV3.cs
donnie-msft: See https://github.com/NuGet/NuGet.Client/pull/5763#discussion_r1581639227

nkolev92 (approved):

src/NuGet.Core/NuGet.Protocol/Providers/OwnerDetailsUriResourceV3Provider.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            OwnerDetailsUriTemplateResourceV3? resource = null;
+            ServiceIndexResourceV3? serviceIndex = await source.GetResourceAsync<ServiceIndexResourceV3>(token);
+            if (serviceIndex != null)
+            {
+                Uri? uriTemplate = serviceIndex.GetServiceEntryUri(ServiceTypes.OwnerDetailsUriTemplate);
+
+                if (uriTemplate != null)
+                {
+                    resource = OwnerDetailsUriTemplateResourceV3.CreateOrNull(uriTemplate);
```
kartheekp-ms, line 36: I am also wondering if we should invoke `uriTemplate.IsWellFormedOriginalString()` as per [security considerations](https://learn.microsoft.com/dotnet/fundamentals/runtime-libraries/system-uri#security-considerations) in the docs.

src/NuGet.Core/NuGet.Protocol/Resources/OwnerDetailsUriTemplateResourceV3.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// <summary>
+        /// Gets a URL for viewing package Owner URL outside of Visual Studio. The URL will not be verified to exist.
+        /// </summary>
+        /// <param name="owner">The owner username.</param>
+        /// <returns>The first URL from the resource, with the URI template applied.</returns>
+        public Uri GetUri(string owner)
+        {
+            var uriString = _template
+#if NETCOREAPP
+               .Replace("{owner}", owner, StringComparison.OrdinalIgnoreCase);
```
kartheekp-ms, line 54: The `string.Replace()` method in this function replaces the `{owner}` substring only if it exists; otherwise, it does nothing and returns the same string. Given that nuget.org includes the `{owner}` substring in the template URL and we are planning to make the placeholder mandatory as per [this discussion](https://github.com/NuGet/docs.microsoft.com-nuget/pull/3282/files#r1576870144), we seem to lack necessary client-side validations.

I also suggest adding an additional check to ensure that the actual URL derived from replacing the `{owner}` placeholder remains within the confines of the specified template URL, which could help protect against path traversal issues. Here's an illustrative example demonstrating this:

```cs
Uri templateUri = new Uri("https://nuget.test/profiles/{owner}?_src=template");
Uri actualUri = new Uri("https://nuget.test/profiles/kp?_src=template");
Console.WriteLine("IsBaseOf: " + templateUri.IsBaseOf(actualUri)); // returns true

Uri templateUri = new Uri("https://nuget.test/profiles/{owner}?_src=template");
Uri actualUri = new Uri("https://nuget.test/profiles/../kp?_src=template");
Console.WriteLine("IsBaseOf: " + templateUri.IsBaseOf(actualUri)); // returns false because "../kp" string is specified as owner which changes the URL to "https://nuget.test/kp?_src=template" after resolving.
```

This additional check using `IsBaseOf` could be crucial for maintaining the integrity and security of URL generation. I found out about the `IsBaseOf` method in [Security considerations](https://learn.microsoft.com/dotnet/fundamentals/runtime-libraries/system-uri#security-considerations) from the docs.

kartheekp-ms (commented):
I know this PR has already been merged, but I would like to provide some feedback and am happy to discuss it further offline. If my comments are valid, perhaps we could consider opening a follow-up issue.

src/NuGet.Core/NuGet.Protocol/Providers/OwnerDetailsUriResourceV3Provider.cs
donnie-msft: I thought I had addressed this concern in https://github.com/NuGet/NuGet.Client/pull/5763#discussion_r1581410514. Maybe we can have a chat offline so you can show me how this would work?

src/NuGet.Core/NuGet.Protocol/Resources/OwnerDetailsUriTemplateResourceV3.cs
donnie-msft: The outcome of the discussion is in [this comment](https://github.com/NuGet/docs.microsoft.com-nuget/pull/3282/files#r1576889184) is to document that it's required, but that Client will not require it. Therefore, it's acceptable that a Uri template not contain `{owner}`.

>  2) we say it's required in the docs because we think it should really be there, but we make no effort to enforce it in the client.

Would the `actualUri` in your example fail to be an absolute URI in that case? I'll have to test this and follow-up if necessary. Note that Package Details Uri has used this pattern for the template since it was created.

src/NuGet.Core/NuGet.Protocol/Providers/OwnerDetailsUriResourceV3Provider.cs
donnie-msft: I ran a test and I'm still seeing that the template will never be considered well formed, and that the generated URI, even with `...` is indeed considered well formed. Can you provide a test case here that shows what this will catch?

> URL: https://nuget.test/profiles/{owner}?_src=template
> Valid?True
> IsWellFormed?False
> 
> URL: https://nuget.test/profiles/donnie?_src=template
> Valid?True
> IsWellFormed?True
> 
> URL: https://nuget.test/profiles/../?_src=template
> Valid?True
> IsWellFormed?True

donnie-msft: Talked offline - resolving

src/NuGet.Core/NuGet.Protocol/Resources/OwnerDetailsUriTemplateResourceV3.cs
kartheekp-ms: Talked offline. We should be good for now with additional changes.

PR #5766, opened 2024-04-26
Show Package Owner profile links on Browse Tab

### LINKED ISSUE #12501: Swap authors for owners in Visual Studio Package Manager UI

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12501
Spec: https://github.com/NuGet/Home/blob/dev/accepted/2023/owner-author-pmui.md

### Key points
- On the Browse Tab, when a single package source is selected and it provides a resource, `OwnerDetailsUriTemplateResourceV3`, hyperlinks will be shown for each package owner in the PM UI Packages List. This is referred to as **Supporting Known Owners.**
- The owner profile URLs are built from the template from that source and the link's ToolTip shows the full URL.
- No telemetry is added in this PR.
_Screenshot of single package source Supporting Known Owners_:
image

- Authors will never be shown in the packages list when the source Supports Known Owners.
- In all other scenarios, Author will continue to be shown when Known Owners is not supported.
_Screenshot of single package source which does not Support Known Owners_:
image

### Commentary
- I'm introducing new ViewModels here, but no software design pattern or architectural improvements/refactoring is being done to existing Views or ViewModels in this PR. I'm happy to contribute to that type of refactoring work in a separate effort. 
- A new type, `IOwnerDetailsUriService`, provides an indicator for support for known owners, and a means to obtain an Owner Profile URL.
  - Currently, only a `MultiSourcePackageMetadataProvider` implements and provides this service:
    - A single package source is required to Support Known Owners. 
    - No metadata merging decisions are being made in this PR.
- `PackageSearchMetadataContextInfo` is the coming together of two things: Search results and package metadata. Owners belong to Search, therefore this context is where I'm materializing `KnownOwner` models.
- Package Item ToolTips:
  - Still show "by `author`" when there's no Supported Known Owners. 
  - Show "by `owner`" when there are Supported Known Owners.
    - I'm using the String here rather than calculating it from the `KnownOwnerViewModels`, as they should match. I kept this pattern here as I'd like to eventually refactor this class and use a more MVVM approach for the AuthorAndDownloadCount's view model.
- Hyperlink events are handled by the `ExecuteOpenExternalLink` event similar to the Details Pane event. Telemetry can be added for owner profile links by simply providing a new property which this event will emit.
- `KnownOwnerViewModel` represents known owners within Visual Studio UIs.
- `LastItemToVisibilityConverter` allows for adding commas after each hyperlink, except the last one.

### Accessibility

- I did not test Psuedo-localization (PLOC) since the reflow design of the packages list is already handled - that is, it's truncated if it exceeds the width of the packages list.
- Keyboard navigation visits each hyperlink
- Screen-readers announce each hyperlink as "Hyperlink <owner name>", similar to other hyperlinks.
- **Follow-up**: Add explanatory text into the ToolTip or explain that these profile URLs are owners from a particular source. https://github.com/NuGet/Home/issues/13414
- **Accessibility Insights** identified 1 bug which is a known WPF bug. I filed this to track it for when it's found by the test team:  https://github.com/NuGet/Client.Engineering/issues/2801

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [x] Documentation PR or issue filled https://github.com/NuGet/docs.microsoft.com-nuget/issues/3287
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
@@ -13,8 +13,8 @@ using Microsoft.ServiceHub.Framework;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Threading;
 using NuGet.Common;
+using NuGet.PackageManagement.UI.ViewModels;
 using NuGet.PackageManagement.VisualStudio;
-using NuGet.Packaging.Core;
 using NuGet.Protocol.Core.Types;
 using NuGet.Versioning;
 using NuGet.VisualStudio;
@@ -262,9 +262,9 @@ namespace NuGet.PackageManagement.UI
                 return Enumerable.Empty<PackageItemViewModel>();
             }
 
-            var listItemViewModels = new List<PackageItemViewModel>();
+            var listItemViewModels = new List<PackageItemViewModel>(capacity: _state.ItemsCount);
 
-            foreach (PackageSearchMetadataContextInfo metadata in _state.Results.PackageSearchItems)
+            foreach (PackageSearchMetadataContextInfo metadataContextInfo in _state.Results.PackageSearchItems)
             {
                 VersionRange allowedVersions = VersionRange.All;
                 VersionRange versionOverride = null;
@@ -272,7 +272,7 @@ namespace NuGet.PackageManagement.UI
                 // get the allowed version range and pass it to package item view model to choose the latest version based on that
                 if (_packageReferences != null)
                 {
-                    IEnumerable<IPackageReferenceContextInfo> matchedPackageReferences = _packageReferences.Where(r => StringComparer.OrdinalIgnoreCase.Equals(r.Identity.Id, metadata.Identity.Id));
+                    IEnumerable<IPackageReferenceContextInfo> matchedPackageReferences = _packageReferences.Where(r => StringComparer.OrdinalIgnoreCase.Equals(r.Identity.Id, metadataContextInfo.Identity.Id));
                     VersionRange[] allowedVersionsRange = matchedPackageReferences.Select(r => r.AllowedVersions).Where(v => v != null).ToArray();
                     VersionRange[] versionOverrides = matchedPackageReferences.Select(r => r.VersionOverride).Where(v => v != null).ToArray();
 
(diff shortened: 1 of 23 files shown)


jeffkl (approved):

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
```diff
@@ -262,17 +262,17 @@ public IEnumerable<PackageItemViewModel> GetCurrent()
                 return Enumerable.Empty<PackageItemViewModel>();
             }
 
-            var listItemViewModels = new List<PackageItemViewModel>();
+            var listItemViewModels = new List<PackageItemViewModel>(capacity: _state.ItemsCount);
 
-            foreach (PackageSearchMetadataContextInfo metadata in _state.Results.PackageSearchItems)
+            foreach (PackageSearchMetadataContextInfo metadataContextInfo in _state.Results.PackageSearchItems)
```
nkolev92, line 267: nit, this rename here did cause a lot of changes that made it harder to gauge which lines are actually changed.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/KnownOwnerViewModel.cs
```diff
@@ -0,0 +1,24 @@
...
+        private string _name;
+        private Uri _link;
+
+        public KnownOwnerViewModel(KnownOwner knownOwner)
+        {
+            _name = knownOwner.Name;
+            _link = knownOwner.Link;
+        }
+
+        public string Name => _name;
```
nkolev92, line 20: public string Name {get;} is syntactic sugar over having a private and public field.

src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/ContextInfos/PackageSearchMetadataContextInfo.cs
```diff
@@ -22,9 +24,38 @@ public sealed class PackageSearchMetadataContextInfo
         public string? Tags { get; internal set; }
         public Uri? LicenseUrl { get; internal set; }
         public Uri? ReadmeUrl { get; internal set; }
-        public string? Owners { get; internal set; }
         public Uri? ProjectUrl { get; internal set; }
         public DateTimeOffset? Published { get; internal set; }
+        public IReadOnlyList<string>? OwnersList { get; internal set; }
+        public string? Owners { get; internal set; }
+        public IReadOnlyList<KnownOwner>? KnownOwners
```
nkolev92, line 31: This class is going from a ~model~ DTO (post discussion), to something that has action logic in addition to being a model. 

i don't have a clear recommendation, but something that stood out.

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
donnie-msft: Yes it did, and I apologize this makes it a little harder to read the diff, but it needs to be named accurately. Metadata is something different than metadataContextInfo.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/KnownOwnerViewModel.cs
donnie-msft: True, I created a field at some point during development, but it's not really necessary, I'll revise that.

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
nkolev92: I think the history is that the type was IPackageSearchMetadata and then became PackageSearchMetadataContextInfo, which is the serializable model. 

They're almost fully equivalent, I'd argue metadata doesn't change too much. 
This has been merged, so no longer relevant though. Resolving.

src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/ContextInfos/PackageSearchMetadataContextInfo.cs
donnie-msft: Talked offline. This is kind of a DTO versus Model, so I'll move the calculation logic down into `SearchObject`.

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
donnie-msft: We discussed offline. The reasoning is that I want to emphasize the difference between metadata and contextInfo which is more than just metadata.

src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/ContextInfos/PackageSearchMetadataContextInfo.cs
donnie-msft: https://github.com/NuGet/NuGet.Client/pull/5782

PR #5742, opened 2024-04-05
Add `allowinsecureconnections` option

Nigusu-Allehu:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13378

Regression? Last working version:

## Description

`nuget add` and `nuget update` commands do not allow for adding a new package source with an attribute `allowInsecureConnections`.  Users would have to go to the config file manually to add this attribute. 
* This PR adds a `-allowInsecureConeections` option to these commands.

## PR Checklist

- [x] PR has a meaningful title
- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  https://github.com/NuGet/Home/issues/13498 , https://github.com/dotnet/docs/issues/40488
  - [x] Documentation PR or issue filled
  - **OR**
  - [ ] N/A

PR #5767, opened 2024-04-26
Fix `PackageSourceProvider.UpdatePackageSource` doesn't respect a value from `AllowInsecureConnection` field

### LINKED ISSUE #13418: `PackageSourceProvider.UpdatePackageSource` doesn't respect a value from `AllowInsecureConnection` field
### NuGet Product Used

NuGet SDK

### Product Version

6.9.1

### Worked before?

No, this is not a regression

### Impact

### Repro Steps & Context

Hi folks, a `PackageSourceProvider.UpdatePackageSource` method from the `NuGet.Configuration` package doesn't respect the `PackageSource.AllowInsecureConnections` property value.

This is due to a missing expression in the `if` statement here:
https://github.com/NuGet/NuGet.Client/blob/3ebcf80e47ce6aaca1d017acd8750e27c8208822/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs#L616-L617

To reproduce the issue you could use a simple code snippet:

```csharp
using NuGet.Configuration;

var configContents = """
    <?xml version="1.0" encoding="utf-8"?>
    <configuration>
        <packageSources>
            <add key="default-http" value="http://api.nuget.org/v3/index.json" />
        </packageSources>
    </configuration>
    """;

File.WriteAllText(Path.Combine(Environment.CurrentDirectory, "NuGet.Config"), configContents);

var settings = new Settings(Environment.CurrentDirectory);
var packageSourceProvider = new PackageSourceProvider(settings);

var source = packageSourceProvider.GetPackageSourceByName("default-http")!;
source.AllowInsecureConnections = true;
packageSourceProvider.UpdatePackageSource(source, false, false);

// Loading brand new Settings instance to ensure we are not checking for cached value
settings = new Settings(Environment.CurrentDirectory);
packageSourceProvider = new PackageSourceProvider(settings);
source = packageSourceProvider.GetPackageSourceByName("default-http")!;

var value = source.AllowInsecureConnections; // false, should be true
```

### Verbose Logs

_No response_

seclerp:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13418

## Description

Hi, so this is quite a simple fix. The missing if check prevented update logic if only AllowInsecureConnections was changed.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
+++ b/src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
@@ -614,7 +614,8 @@ namespace NuGet.Configuration
             if (string.Equals(newSource.Name, existingSource.Name, StringComparison.OrdinalIgnoreCase))
             {
                 if ((!string.Equals(newSource.Source, existingSource.Source, StringComparison.OrdinalIgnoreCase) ||
-                    newSource.ProtocolVersion != existingSource.ProtocolVersion) && newSource.IsPersistable)
+                    newSource.ProtocolVersion != existingSource.ProtocolVersion ||
+                    newSource.AllowInsecureConnections != existingSource.AllowInsecureConnections) && newSource.IsPersistable)
                 {
                     Settings.AddOrUpdate(ConfigurationConstants.PackageSources, newSource.AsSourceItem());
                     isDirty = true;
--- a/test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Configuration.Test/PackageSourceProviderTests.cs
@@ -1388,6 +1388,44 @@ namespace NuGet.Configuration.Test
             AssertPackageSource(values[2], "three", "threesource", true);
         }
 
+        [Fact]
+        public void UpdatePackageSource_ShouldUpdateAllowInsecureConnections()
+        {
+            using var directory = TestDirectory.Create();
+
+            // Arrange
+            var configContents = """
+                <?xml version="1.0" encoding="utf-8"?>
+                <configuration>
+                    <packageSources>
+                        <add key="default-http" value="http://api.nuget.org/v3/index.json" />
+                    </packageSources>
+                </configuration>
+                """;
+
+            File.WriteAllText(Path.Combine(directory.Path, "NuGet.Config"), configContents);
+
+            var settings = new Settings(directory);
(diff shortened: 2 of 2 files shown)


Nigusu-Allehu (commented):
Good catch! We're adding `-allowInsecureConnections` option to the `update` and `add` commands. With it we have also addressed this problem: https://github.com/NuGet/NuGet.Client/pull/5742. However, it is not ready for release yet as it is just a sub part of another project.

seclerp:
Nice, I didn't check that PR for an already implemented fix, thanks for noticing! So what about this one? Maybe let's proceed with it to address the problem quicker?

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
```diff
@@ -614,7 +614,8 @@ public void UpdatePackageSource(PackageSource source, bool updateCredentials, bo
             if (string.Equals(newSource.Name, existingSource.Name, StringComparison.OrdinalIgnoreCase))
             {
                 if ((!string.Equals(newSource.Source, existingSource.Source, StringComparison.OrdinalIgnoreCase) ||
-                    newSource.ProtocolVersion != existingSource.ProtocolVersion) && newSource.IsPersistable)
+                    newSource.ProtocolVersion != existingSource.ProtocolVersion ||
```
nkolev92, line 617: cc @Nigusu-Allehu 

I'm guessing DisableTLSCertificateValidation has the same concern?
Can you please double check?

nkolev92 (approved):

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
Nigusu-Allehu: Yes that is right. I will create a new issue for that

Nigusu-Allehu (approved):

src/NuGet.Core/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
Nigusu-Allehu: https://github.com/NuGet/Home/issues/13423

PR #5787, opened 2024-05-06
Run all unit and functional tests in a new pipeline

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2789

## Description

This change adds a new pipeline for running all unit and functional tests.  I moved the logic from `build\build.proj` to `Directory.Solution.targets` so that the build pipeline is just building `NuGet.sln`.  This reduces complexity and makes it a lot easier to replicate the hosted build.

Locally, you can now run `dotnet test` at the root and it will restore, build, and run all unit tests.  To run the functional tests, you simply run `dotnet test -Property:TestType=Functional`.  You can also run tests for individual target frameworks via `dotnet test --framework net8.0`.

The build pipeline splits the execution across platform and target framework so that the matrix includes:
image

This provides a lot of parallelism and its easy to re-run individual runs in case of a flaky test failure.  Additionally, tests are grouped together so its easy to see where a failure occurred:

image

When manually scheduling the pipeline, you can deselect jobs in case you want to just run one or more test runs.

Some other change notes:
* I updated `CommandRunner` to make it more reliable
* I updated a bunch of tests to log the output of executions of tools like `nuget.exe`, `msbuild.exe`, and `dotnet.exe`.  This helped me diagnose problems when getting this pipeline working
* Each folder under `test` specifies the `TestProjectType` as `Unit` or `Functional` and each project can override the value if necessary.
* Target frameworks are discovered automatically but the list of frameworks we run tests on is hard-coded into the pipeline.  This is good for now as a particular branch needs to test specific frameworks but perhaps in the future I can make it more dynamic.
* The pipeline is on the `dnceng-public` tenant so the results will be viewable by the general public which will be great for external contributors (https://dev.azure.com/dnceng-public/public/_build?definitionId=289&_a=summary).
* I left the existing pipeline in place for now as we transition to using this one.  I'll send a follow-up PR to clean up the legacy implementation.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


Directory.Solution.targets
jeffkl, line 1: This file essentially does what `build\build.proj` does but is more dynamic and runs logic for you when you run `dotnet test` at the root of the repo.  I'm hoping to pursue some changes in .NET SDK to make it so that this logic ships as part of the SDK instead of existing only in our repo.

build/common.project.props
```diff
@@ -36,7 +36,8 @@
 
     <!-- Target frameworks for unit tests that test libaries using signing APIs -->
     <TargetFrameworksUnitTestForSigning>$(NETCoreTargetFramework)</TargetFrameworksUnitTestForSigning>
-    <TargetFrameworksUnitTestForSigning Condition="'$(IsXPlat)' != 'true' And '$(DotNetBuildSourceOnly)' != 'true'">$(NETFXTargetFramework);$(TargetFrameworksUnitTestForSigning);$(NETCoreLegacyTargetFramework)</TargetFrameworksUnitTestForSigning>
+    <TargetFrameworksUnitTestForSigning Condition="'$(DotNetBuildSourceOnly)' != 'true'">$(TargetFrameworksUnitTestForSigning);$(NETCoreLegacyTargetFramework)</TargetFrameworksUnitTestForSigning>
```
jeffkl, line 39: This fixes an issue where some projects are missing the `netcoreapp3.1` framework 🤦

eng/pipelines/pr.job.yml
jeffkl, line 1: This file represents a test execution on a platform and a test type (unit or functional).  It restores and builds with one task and runs the tests with another.

eng/pipelines/pr.yml
jeffkl, line 1: This is the new pipeline and is as simple as possible, just schedules jobs for each platform and framework.

test/NuGet.Clients.FuncTests/NuGet.CommandLine.FuncTest/CommandRunnerTest.cs
```diff
@@ -24,7 +34,7 @@ public async Task Run_AlwaysCaptureStdout()
                 {
                     return Task.Run(() =>
                     {
-                        for (int i = 0; i < 1000; i++)
+                        for (int i = 0; i < 100; i++)
```
jeffkl, line 37: This test was taking a minute to run so I reduced this value to get it to finish in a few seconds.  It still catches problems and is way faster.

test/NuGet.Core.Tests/NuGet.Credentials.Test/PluginCredentialProviderTests.cs
```diff
@@ -569,10 +571,14 @@ public async Task OnPluginUnexpectedStatusException_NoRetryIfVerbosityFlagWasNot
         [Fact]
         public void SetsIdBasedOnTypeAndFilename()
         {
+            FileInfo providerFileInfo = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
+                ? new FileInfo(@"c:\some\path\provider.exe")
+                : new FileInfo("/some/path/provider");
+
             var mockLogger = new Mock<Common.ILogger>();
-            var provider = new PluginCredentialProvider(mockLogger.Object, @"c:\some\path\provider.exe", 5, "Normal");
+            var provider = new PluginCredentialProvider(mockLogger.Object, providerFileInfo.FullName, 5, "Normal");
```
jeffkl, line 579: This makes the test pass on non-Windows so its now being run on other platforms.

test/NuGet.Core.Tests/NuGet.PackageManagement.Test/NuGet.PackageManagement.Test.csproj
```diff
@@ -3,6 +3,8 @@
     <TargetFrameworks>$(TargetFrameworksUnitTest)</TargetFrameworks>
     <UseParallelXunit>true</UseParallelXunit>
     <Description>Unit tests for NuGet.PackageManagement.</Description>
+    <TestProjectType Condition="'$(IsXPlat)' != 'true'">Functional</TestProjectType>
```
jeffkl, line 6: This is how a project conveys that its a functional test but only on Windows.

test/NuGet.Core.Tests/NuGet.PackageManagement.Test/NuGetPackageManagerTests.cs
```diff
@@ -6257,7 +6257,7 @@ public async Task PreviewInstallPackage_PackagesConfig_RaiseTelemetryEvents()
                     CancellationToken.None);
 
                 // Assert
-                Assert.Equal(3, telemetryEvents.Count);
+                telemetryEvents.Count.Should().BeGreaterThan(0);
```
jeffkl, line 6260: This test is flaky so this "fix" isn't ideal but it works for now.

https://github.com/NuGet/Client.Engineering/issues/2817

test/TestUtilities/Test.Utility/CommandRunnerResult.cs
```diff
@@ -32,9 +30,8 @@ public class CommandRunnerResult
         /// </summary>
         public string AllOutput => Output + Environment.NewLine + Errors;
 
-        internal CommandRunnerResult(Process process, int exitCode, string output, string error)
+        internal CommandRunnerResult(int exitCode, string output, string error)
```
jeffkl, line 33: The `Process` is in a `using` so there's no point in passing a reference to it since we only needed it's exit code.  This might make things more reliable...

test/TestUtilities/Test.Utility/MockServer.cs
```diff
@@ -104,15 +104,13 @@ public void Stop()
         {
             try
             {
-                _listener.Abort();
+                _listener?.Abort();
```
jeffkl, line 107: This logic was causing a `NullReferenceException` sometimes and would crash tests.  This also might make things more reliable...

jeffkl:
Most of the changes are to tests to increase diagnostics by making them all send EXE output to the `ITestOutputHelper`.  You can review the first commit to just see the actual implementation.  I can revert the test changes if we feel like its too much.

test/NuGet.Core.FuncTests/Directory.Build.props
```diff
@@ -1,6 +1,7 @@
 <Project>
   <PropertyGroup Condition="'$(_WasCommonPropsImported)' != 'true'">
     <TestProject Condition="'$(TestProject)' == ''">true</TestProject>
+    <TestProjectType Condition="'$(TestProjectType)' == ''">Functional</TestProjectType>
```
donnie-msft, line 4: nit: My brain wants Unit tests to be the default and functional to be the opt-in but it looks like there's more Functional tests.

test/NuGet.Clients.FuncTests/NuGet.CommandLine.FuncTest/Commands/PushCommandTest.cs
```diff
@@ -145,20 +154,23 @@ public void PushCommand_Server_SkipDuplicate_NotSpecified_PushHalts()
...
+                        timeOutInMilliseconds: 120 * 1000,
+                        testOutputHelper: _testOutputHelper); // 120 seconds
 
                     result3 = CommandRunner.Run(
                        nuget,
                        packageDirectory,
                        $"push {sourcePath2} -Source {server.Uri}push -Timeout 110",
-                       timeOutInMilliseconds: 120 * 1000); // 120 seconds
+                       timeOutInMilliseconds: 120 * 1000,
+                        testOutputHelper: _testOutputHelper); // 120 seconds
```
donnie-msft, line 173:
```suggestion
                       testOutputHelper: _testOutputHelper); // 120 seconds
```

test/NuGet.Clients.FuncTests/NuGet.CommandLine.FuncTest/Commands/PushCommandTest.cs
```diff
@@ -208,21 +220,24 @@ public void PushCommand_Server_SkipDuplicate_IsSpecified_PushProceeds()
...
+                        timeOutInMilliseconds: 120 * 1000,
+                        testOutputHelper: _testOutputHelper); // 120 seconds
 
                     //Third run with a different package.
                     result3 = CommandRunner.Run(
                         nuget,
                         packageDirectory,
                         $"push {sourcePath2} -Source {server.Uri}push -Timeout 110 -SkipDuplicate",
-                        timeOutInMilliseconds: 120 * 1000); // 120 seconds
+                        timeOutInMilliseconds: 120 * 1000,
```
donnie-msft, line 239: Is this a good time to just make all the timeouts literals?

test/NuGet.Clients.FuncTests/NuGet.CommandLine.FuncTest/Commands/PushCommandTest.cs
```diff
@@ -845,7 +874,8 @@ public void PushCommand_WhenPushingToAnHttpServerV3_WithSymbols_Warns()
                 nuget,
                 packageDirectory,
                 $"push {snupkgFileName} -Source {sourceName} -Timeout 110 -Verbosity detailed",
-                timeOutInMilliseconds: 120000); // 120 seconds
+                timeOutInMilliseconds: 120000,
+                        testOutputHelper: _testOutputHelper); // 120 seconds
```
donnie-msft:
```suggestion
                testOutputHelper: _testOutputHelper); // 120 seconds
```

test/NuGet.Clients.FuncTests/NuGet.CommandLine.FuncTest/Commands/TrustedSignersCommandTests.cs
```diff
@@ -137,7 +140,8 @@ public void TrustedSignersCommand_Failure_InvalidArguments_HelpMessage(string ar
             var result = CommandRunner.Run(
                 _nugetExePath,
                 Directory.GetCurrentDirectory(),
-                $"trusted-signers {args}");
+                $"trusted-signers {args}",
+                    testOutputHelper: _testOutputHelper);
```
donnie-msft, line 144: There's a few indentations that are off...

test/NuGet.Core.FuncTests/Msbuild.Integration.Test/MsbuildIntegrationTestFixture.cs
```diff
@@ -8,14 +8,22 @@
...
     {
-        internal readonly string _testDir;
-        private readonly Dictionary<string, string> _processEnvVars = new Dictionary<string, string>();
+        internal readonly string _testDir = Directory.GetCurrentDirectory();
+
+        private readonly Dictionary<string, string> _processEnvVars = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
+        {
+            ["DOTNET_MULTILEVEL_LOOKUP"] = "0",
+            // Uncomment to debug the msbuild call
+            //["DEBUG_RESTORE_TASK"] = bool.TrueString,
```
donnie-msft, line 23: neat!

test/NuGet.Core.Tests/NuGet.PackageManagement.Test/NuGet.PackageManagement.Test.csproj
donnie-msft: Nice - just to be clear, the default from Directory.Build.props is Functional, then this project can override that default?

test/TestUtilities/Test.Utility/CommandRunnerResult.cs
donnie-msft: Let's hope!

test/TestUtilities/Test.Utility/MockServer.cs
donnie-msft: 💖

donnie-msft (dismissed):

test/NuGet.Core.Tests/NuGet.PackageManagement.Test/NuGet.PackageManagement.Test.csproj
jeffkl: Yes, in this case the `Directory.Build.props` actually indicates that every project in this folder is a unit test.  But this test project is actually a Functional test.  Its currently declared like this:

https://github.com/NuGet/NuGet.Client/blob/dev/build/common.project.props#L210-L216

test/NuGet.Core.FuncTests/Directory.Build.props
jeffkl: The folder is named `NuGet.Core.FuncTests` so by default these projects are functional tests.

Directory.Solution.targets
```diff
@@ -0,0 +1,251 @@
...
+    -->
+    <VSTestNoLogo Condition="'$(VSTestNoLogo)' == ''">true</VSTestNoLogo>
+    <VSTestVerbosity Condition="'$(SYSTEM_DEBUG)' == 'true'">detailed</VSTestVerbosity>
+    <VSTestVerbosity Condition="'$(VSTestVerbosity)' == ''">quiet</VSTestVerbosity>
+    <VSTestSetting Condition="'$(VSTestSetting)' == '' And '$(CI)' == 'true'">$([System.IO.Path]::Combine('$(MSBuildThisFileDirectory)', 'build', 'xunit.runsettings'))</VSTestSetting>
+    <VSTestResultsDirectory Condition="'$(VSTestResultsDirectory)' == ''">$([System.IO.Path]::Combine('$(MSBuildThisFileDirectory)', 'TestResults'))</VSTestResultsDirectory>
+    <VSTestLogger Condition="'$(VSTestLogger)' == ''">console%3Bverbosity=$(VSTestVerbosity)</VSTestLogger>
+
+    <!--
+      Build tests for all supported target frameworks by default, otherwise build tests for just the specified target framework.
```
nkolev92, line 17: This is not really feedback, but just a shout out how much these comments are appreciated. 

MSBuild doesn't always have a clear correlation as to why a certain value is imported, so this is incredibly helpful.

build/common.project.props
nkolev92: At some point I created: https://github.com/NuGet/NuGet.Client/blob/dev/docs/updating-target-frameworks.md. 

Can you please update it with a pointer to the new location of the targets?

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/NuGet.CommandLine.Test.csproj
```diff
@@ -3,6 +3,7 @@
     <TargetFramework>$(NETFXTargetFramework)</TargetFramework>
     <UseParallelXunit>true</UseParallelXunit>
     <Description>An end-to-end test suite for NuGet.CommandLine. Contains tests for every nuget.exe CLI command. Overlaps in tests with NuGet.CommandLine.FuncTest.</Description>
+    <TestProjectType Condition="'$(IsXPlat)' != 'true'">Functional</TestProjectType>
```
nkolev92, line 6: How does this run on mono now?

nkolev92 (dismissed):
👏 👏 👏 👏 👏 
Minor feedback. I think it needs addressed, but it can be a future PR.

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/NuGet.CommandLine.Test.csproj
jeffkl: No tests are currently running under Mono for this pipeline.  This functional test only runs on Windows.  To make it work on Mac for Mono, I'll need to explore building the `NuGet.CommandLine` project on Mac where I think the only technical challenge is running `ilmerge.exe`.  If that worked, I could get Mono tests running again but considering how deprecated Mono is and how similar it is to .NET Framework, I don't foresee us really needing Mono tests anymore.

build/common.project.props
jeffkl: Yeah I saw this, I want to come up with a better system, so I filed a tracking issue: https://github.com/NuGet/Client.Engineering/issues/2827

test/NuGet.Clients.FuncTests/NuGet.CommandLine.FuncTest/Commands/PushCommandTest.cs
jeffkl: Not in my opinion, I really want to circle back and improve these tests at a later date.

nkolev92 (dismissed):

eng/pipelines/pr.job.yml
```diff
@@ -0,0 +1,104 @@
...
+  variables:
+    - name: TestType
+      value: ${{ parameters.testType }}
+    - name: TestTargetFramework
+      value: ${{ parameters.testTargetFramework }}
+    - name: TestResultsDirectory
+      ${{ if eq(parameters.osName, 'Windows') }}:
+        value: $(Agent.TempDirectory)\TestResults
+      ${{ else }}:
+        value: $(Agent.TempDirectory)/TestResults
```
zivkan: Windows supports `/`. Is there a reason we can't use that everywhere, and avoid a lot of these "if windows .. else .." blocks?

eng/pipelines/pr.job.yml
zivkan, line 1: Is it worthwhile moving this file to a subdirectory, so that all the yml files in the `eng/pipelines` directory are "targets" or entry points of Azure Pipeline definitions, and everything that's imported from one of these entry ymls are elsewhere?

eng/pipelines/pr.yml
```diff
@@ -0,0 +1,241 @@
...
+variables:
+  DOTNET_NOLOGO: 1
+  NUGET_EXPERIMENTAL_CHAIN_BUILD_RETRY_POLICY: 3,1000
+  SkipSigning: true
+  CI: true
+
+trigger:
+  branches:
+    include:
+    - dev
```
zivkan, line 36: Wouldn't we want to run CI on all PRs, not just the dev branch?  CI on release branches are useful, but moreso if we have some `dev-feature-branch`, we want to run CI so we can test PRs before the whole feature branch is ready to go into dev.

eng/pipelines/pr.yml
```diff
@@ -0,0 +1,241 @@
...
+    dependsOn: []
+    jobs:
+    - template: pr.job.yml
+      parameters:
+        displayName: Functional Tests on Windows (.NET Framework 4.7.2)
+        osName: Windows
+        agentPool: NetCore-Public
+        agentDemands: ImageOverride -equals Windows.VS2022Preview.Amd64.Open
+        testType: Functional
+        testTargetFramework: net472
```
zivkan, line 104: I'm not a fan of testing each TFM in a separate job. I know different people have different tolerances for waiting, but I will find this pipeline breaking when we change target frameworks more annoying than having CI jobs take a little longer  (but there will be fewer of them too). I always go do something else while I wait for CI to finish.

Directory.Solution.targets
zivkan: Have you tested the perf of this build on a high CPU count machine?

I always felt like NuGet.Client was slower to `msbuild -t:BuildVS build/build.proj` than `msbuild nuget.sln`.  I never figured out why, but it seemed to me like build.proj was making msbuild do bad choices in what it could build in parallel vs serially. I hope modifying this metaproj doesn't make building the sln file "slow" as well.

zivkan: Ok, on the dev branch, running `msbuild NuGet.sln` a few times, so all projects are up to date, it takes 11 second on my machine. Checking out this branch, building the sln still takes 11 seconds. So it seems to be all good 👍 

`msbuild build/build.proj -t:BuildVS` takes 25 seconds, so it wasn't just my imagination, build.proj really does make building slower for some reason.

eng/pipelines/pr.yml
jeffkl: The main advantages of having them fanned out is more parallelism (the machines we get for Linux and Mac have 2 cores, Windows machines have 4) and you can re-run a smaller stage in case of a flaky test.  I'm more than happy to tweak this configuration down the road but I believe this is the best balance for the current state of things.

eng/pipelines/pr.yml
jeffkl: I'm not sure, we currently have the PrivateDev branch set up to just build all branches that start with `dev-`.  These settings will only trigger on PRs to the `dev` branch as you say.  I'm fine with just triggering it for all branches or something like:
```yaml
pr:
  branches:
    include:
    - dev
    - release-*
    - feature-*
    - dev-feature-*
```

Definitely easy to change as we go as well.

eng/pipelines/pr.job.yml
jeffkl: Certainly up for debate. This particular template is only used by this one pull request pipeline so I felt like not hiding it in a subfolder.  Makes it easier to open both in my opinion...

eng/pipelines/pr.job.yml
jeffkl: The value here gets passed to `dotnet test` and that appears to not support a single value on Windows.  I might be able to get rid of this logic as I was originally running multiple `dotnet test` instances for each test assembly and then publishing the results as merged which requires an extra pipeline task.  The current iteration just runs all tests under a single `dotnet test` invocation so maybe I can get away without needing this.  I'll test it tomorrow...

Directory.Solution.targets
jeffkl: For at least this build pipeline, it doesn't build packages or the VSIX since they're not needed to run the unit and functional tests.  I'm not sure if `build.proj` builds those?  I know its doing some glob expansion as well.  Oh well, one of my goals is to get rid of `build.proj` and bake all of our logic directly into the .NET SDK or something like Arcade.

eng/pipelines/pr.job.yml
jeffkl: And it turns out it wasn't necessary due to other changes, thanks for pointing it out.  I've removed it and the publish test result task to make the job yaml even simpler!

nkolev92 (approved):

PR #5796, opened 2024-05-10
Enforce style rules during build

jgonz120:
## Bug

Fixes:

## Description

Updated two projects to include style rules enforcement during build, fixing the build errors for those two projects. 
## PR Checklist

- [ ] PR has a linked issue.
- [ ] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A


build/common.targets
```diff
@@ -8,6 +8,10 @@
     <DefineConstants Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' Or ('$(TargetFrameworkIdentifier)' == '.NETCoreApp' and $([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), 5.0)))">$(DefineConstants);IS_SIGNING_SUPPORTED</DefineConstants>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(EnforceCodeStyleInBuild)' != '' And '$(IsOfficialBuild)' != 'true'">
+    <WarningsNotAsErrors>CS1591</WarningsNotAsErrors>
```
jgonz120: This warning is for missing xml comments for type members.  
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs1591

src/NuGet.Clients/NuGet.CommandLine/CommandLineMachineWideSettings.cs
```diff
@@ -8,17 +8,23 @@
 namespace NuGet.CommandLine
 {
     [Export(typeof(Configuration.IMachineWideSettings))]
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
```
nkolev92, line 11: We've historically added our suppressions in a GlobalSuppressions file, since this makes the file really busy.

src/NuGet.Core/NuGet.Versioning/NuGet.Versioning.csproj
```diff
@@ -7,6 +7,9 @@
     <PackProject>true</PackProject>
     <Shipping>true</Shipping>
     <IncludeInVSIX>true</IncludeInVSIX>
+    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+    <GenerateDocumentationFile Condition="'$(IsOfficialBuild)' == 'true'">false</GenerateDocumentationFile>
```
nkolev92, line 11: Wouldn't this cause it to only fail in the official build?

src/NuGet.Clients/NuGet.CommandLine/NuGet.CommandLine.csproj
```diff
@@ -7,7 +7,9 @@
     <AssemblyTitle>NuGet Command Line</AssemblyTitle>
     <PackageId>NuGet.CommandLine</PackageId>
     <TargetFramework>$(NETFXTargetFramework)</TargetFramework>
-    <GenerateDocumentationFile>false</GenerateDocumentationFile>
+    <GenerateDocumentationFile Condition="'$(IsOfficialBuild)' == 'true'">false</GenerateDocumentationFile>
+    <GenerateDocumentationFile Condition="'$(IsOfficialBuild)' != 'true'">true</GenerateDocumentationFile>
+    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
```
nkolev92, line 12: Isn't this globally set already?

https://github.com/NuGet/NuGet.Client/blob/9b1ed6bf04ebc23342a69c283c1fb98bb219524d/build/common.project.props#L127

src/NuGet.Clients/NuGet.CommandLine/CommandLineMachineWideSettings.cs
jgonz120: It didn't give me the option to in the "Quick actions" even after creating a global suppression file. Lemme see just force the value in there and see if it works

src/NuGet.Clients/NuGet.CommandLine/NuGet.CommandLine.csproj
jgonz120: I read that as saying, if $(EnforceCodeStyleInBuild) is set to anything other than empty string, then it sets the property to true. So if we don't set it somewhere it won't put any value, right?

src/NuGet.Core/NuGet.Versioning/NuGet.Versioning.csproj
jgonz120: I think the opposite, right now the documents are only generated if the project has "PackProject" is set to true, and there's nothing set in "GenerateDocumentationFile".

I wanted to make sure that nothing changes for the official builds. Without the Generated documentation files, some of the rules don't get enforced.

src/NuGet.Clients/NuGet.CommandLine/NuGet.CommandLine.csproj
nkolev92: It effectively sets a default. 
If not set, it'll set true, and since we don't set it, it's gonna be true. 

It's a common msbuild pattern, since properties have ordering considerations, and you don't want to overwrite something set earlier in targets.

src/NuGet.Core/NuGet.Versioning/NuGet.Versioning.csproj
nkolev92: I might be missing some other motivation here, but I tend to be of the opinion that build rules should remain the same for simplicity. 

There's no harm in getting xml files for projects we don't ship as packages, as we have other logic (such as VSIX), to ensure we only ship dlls.

src/NuGet.Clients/NuGet.CommandLine/NuGet.CommandLine.csproj
jgonz120: Shouldn't it be 
```xml 
<EnforceCodeStyleInBuild Condition=" '$(EnforceCodeStyleInBuild)' == '' " >true</EnforceCodeStyleInBuild> 
```

For setting it default to true if nothing is set?

nkolev92: See my brain abstracted that detail.
You're absolutely correct :D 

We should absolutely start by enable that everywhere. 

Locally on my machine, if I set that to true, I only get 2 different errors. 

That could be a good start, as we ensure that EnforceCodeStyleInBuild is actually respected everywhere.

jgonz120: So I did that as a separate PR: https://github.com/NuGet/NuGet.Client/pull/5809

src/NuGet.Clients/NuGet.CommandLine/CommandLineMachineWideSettings.cs
zivkan: I 100% disagree with putting all the suppressions in a GlobalSuppressions file. It being busy in the code is a feature, because when I'm reviewing a future PR and I see they're modifying a line within a `#pragma disable` block, or a line right next to it, I can comment on the PR saying please remove this tech debt.  If the suppression is in the GlobalSuppresions file, it becomes much more difficult to know that there's an analyzer warning that's being suppressed and that improvements could be made.

If this is specifically about VS1591 or other XML comment warnings, then maybe we should disable this analyzer for proejcts that don't ship to nuget.org as packages.

I just don't see GlobalSuppressions.cs as a positive thing, with the possible exception of WPF code and the "keep the object heirarchy lower than X" warning.

PR #5800, opened 2024-05-16
Allow Numbers that look like letters in package ID

Nigusu-Allehu:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2826

## Description

This PR updates the regular expression used for validating package IDs
new regex `[\w\p{N}]+([.-][\w\p{N}]+)*` has two parts
1 .`[\w\p{N}]`  
* \w :A character in the input string can belong to any of the Unicode categories that are appropriate for characters in words. It is a union of:
  * Li: Letter lowercase
  * Lu: Letter UpperCase
  * Lt: Letter TitleCase
  * Lo: Letter Other
  * Lm: Letter, Modifier
  * Mn: Mark, Nonspacing
  * Nd: Number, Decimal Digit
  * Pc: Punctuation Connector
 * \p{N} allows us to include any kind of numeric character in any script. That include letter looking like numbers
2. `([.-][\w\p{N}]+)*` Allows repetition of what I wrote in one by proceeding it with . or -
## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/test/NuGet.Core.Tests/NuGet.Packaging.Test/PackageIdValidatorTest.cs
+++ b/test/NuGet.Core.Tests/NuGet.Packaging.Test/PackageIdValidatorTest.cs
@@ -19,6 +19,23 @@ namespace NuGet.Packaging.Test
         }
 
         [Fact]
+        public void ValidatePackageId_OnGB18030Characters_Succeeds()
+        {
+            // Arrange
+            string packageId = "G龦鿃C啊阿鼾齄D2丂丄狚狛D3狜狝﨨﨩D4ⅫㄨㄩD1ˊˋ〇D5U1U2U3㐀㐁䶴䶵ABCDEFKS";
+
+            // Act & Assert
+            try
+            {
+                PackageIdValidator.ValidatePackageId(packageId);
+            }
+            catch (Exception ex)
+            {
+                Assert.Fail(ex.Message);
+            }
+        }
+
+        [Fact]
         public void EmptyIsNotValid()
         {
             // Arrange
--- a/src/NuGet.Core/NuGet.Packaging/PackageCreation/Utility/PackageIdValidator.cs
+++ b/src/NuGet.Core/NuGet.Packaging/PackageCreation/Utility/PackageIdValidator.cs
@@ -15,7 +15,7 @@ namespace NuGet.Packaging
         /// In case update this value please update in src\NuGet.Core\NuGet.Configuration\PackageSourceMapping\PackageSourceMapping.cs too.
         /// </summary>
         public const int MaxPackageIdLength = 100;
-        private static readonly Regex IdRegex = new Regex(pattern: @"^\w+([.-]\w+)*$",
+        private static readonly Regex IdRegex = new Regex(pattern: @"^[\p{L}\p{M}\p{N}\p{Pc}]+([.-][\p{L}\p{M}\p{N}\p{Pc}]+)*$",
             options: RegexOptions.IgnoreCase | RegexOptions.ExplicitCapture | RegexOptions.CultureInvariant,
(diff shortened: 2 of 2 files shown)


test/NuGet.Core.Tests/NuGet.Packaging.Test/PackageIdValidatorTest.cs
```diff
@@ -18,6 +18,23 @@ public void ValidatePackageIdInvalidIdThrows()
             ExceptionAssert.ThrowsArgumentException(() => PackageIdValidator.ValidatePackageId(packageId), "The package ID '  Invalid  . Woo   .' contains invalid characters. Examples of valid package IDs include 'MyPackage' and 'MyPackage.Sample'.");
         }
 
+        [Fact]
+        public void ValidatePackageId_OnGB18030Characters_Succeeds()
```
jebriede, line 22: Test name ends with "Succeeds" but the Act & Assert step does not validate that anything succeeds. It actually only asserts if an exception is thrown, and in that case, verifies that a failure occurred. Can we update the suffix to describe what we expect to assert in this test?

test/NuGet.Core.Tests/NuGet.Packaging.Test/PackageIdValidatorTest.cs
```diff
@@ -18,6 +18,23 @@ public void ValidatePackageIdInvalidIdThrows()
...
+
+            // Act & Assert
+            try
+            {
+                PackageIdValidator.ValidatePackageId(packageId);
+            }
+            catch (Exception ex)
+            {
+                Assert.Fail(ex.Message);
+            }
```
jebriede, line 35: We shouldn't use try/catch blocks in tests. It seems that you may be trying to validate that an exception will be thrown when calling `PackageIdValidator.ValidatePackageId(packageId);` with the specified input. If that's the case, you'll want to use `Assert.Throws<TypeOfExceptionHere>(() => PackageIdValidator.ValidatePackageId(packageId));` That will make sure that the test verifies that this code path, given the input provided, will always throw an exception.

The try/catch block we have in the test right now will only execute the assertion if the method under tests actually throws an exception. If it doesn't, the test will do nothing and not assert anything which could obscure a bug.

jebriede (changes requested):

test/NuGet.Core.Tests/NuGet.Packaging.Test/PackageIdValidatorTest.cs
Nigusu-Allehu: `ValidatePackageId` method throws an exception if the id is valid otherwise it does not. In this test case, I am trying to assert that an exception is not thrown when using GB18030 characters. The idea is to use a try catch. Try running `ValidatePackageId` and if it throws an exception we have an Assert.Fail in the catch block.

test/NuGet.Core.Tests/NuGet.Packaging.Test/PackageIdValidatorTest.cs
Nigusu-Allehu: It's trying to assert an exception is not thrown rather

martinrrm: I think we should do what JeanPierre is suggesting, if I'm not wrong if by any chance the code changes and `PackageIdValidator.ValidatePackageId(packageId);` doesn't throw the test will succeed and won't catch the exception assert.

nkolev92:
We should close this now.

PR #5801, opened 2024-05-16
New VS nomination data structure and API with generic item bag

### LINKED ISSUE #13443: VS project nominations should use generic data structure
For project systems in Visual Studio that support both CPS and PackageReference capabilities (so in effect just the SDK style .NET projects) use a NuGet data flow where the project system pushes project item and properties to NuGet via an API called NominateProjectAsync.

This API has a data structure where the interface has properties `ProjectReferences`, `PackageReferences`, and so on. So, every time NuGet wants to add a new MSBuild item, we need to create a new iterface for the new item type, publish the package, then edit dotnet/project-system to use the new package and set the new property for the item.

Project properties and item metadata already use a generic property bag, which is more or less equivalent to `IReadOnlyDictionary<string, string>` (but is different for COM compatibility). Items should also use a generic data structure, so that future additions to which MSBuild items NuGet reads no longer requires project system to update package versions to implement the change.

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13443

Regression? no

## Description

Basically, the goal of this PR is to solve the problem described in the linked issue. A new interface/data structure that will prevent the need to make any more future interface changes. https://xkcd.com/927/

Instead of individual `ProjectReference`, `PackageReference`, `PackageDownload`, `PackageVersion`, and `FrameworkReference` properties, each of which is a list of items, there's now one `IReadOnlyDictionary<string, IReadOnlyList<IVsItemReference2>> Items { get; }` property, where the dictionary key is the item type.

* Talking to Nikolche, we believe the only implementation is dotnet/project-system, which is written in .NET, hence we don't believe we need to support COM implementations of project systems any more. Therefore, the new interfaces no longer have `[COMImport]` or GUIDs. Additionally, I removed all the interfaces that were just COM compatible equivalents of read only collections to make them easier to use.
* Modified the VS restore service to use the new data type only. That had a big impact on the test code, so many changes in the test files. But there aren't test behaviour changes, it's just adapting the tests to the new API shape.
* To avoid breaking existing users of the earlier `IVsProjectRestoreInfo` interface versions, the adapter pattern was used to transform the older data structures into the new data structure. This also let some existing tests work without migrating to the new API.
* I had some null bugs, caught by tests. However, since the tests run a LOT of code in each test, it wasn't obvious where the bug was, so I added nullable annotations, which made the PR quite a lot bigger. Sorry, but it seemed easier in the long term to do the null annotations and have the compiler tell me where all the null bugs were, rather than debugging the tests without nullable annotations.
* I wanted to use `record`, `required` and `init` keywords in a test project, but test projects don't import `build/Shared/*.cs`, so I copied the two files that provide the compiler attributes to `build/TestShared/`.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception: As explained in the description, there isn't exactly behaviour changes in this PR. There is a new API, but the old API usage gets transformed into the new API via an adapter. So in effect existing tests cover all the scenarios.
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [x] Documentation PR or issue filled: https://github.com/NuGet/Home/issues/13501
  - **OR**
  - [ ] N/A


src/NuGet.Clients/NuGet.SolutionRestoreManager/ProjectRestoreInfo3Adapter.cs
```diff
@@ -0,0 +1,210 @@
...
+                    IReadOnlyList<IVsReferenceItem2> packageReferences = ToReferenceItem2ReadOnlyList(targetFrameworkInfo.PackageReferences);
+                    result[ProjectItems.PackageReference] = packageReferences;
+                }
+
+                if (targetFrameworkInfo is IVsTargetFrameworkInfo2 targetFrameworkInfo2)
+                {
+                    if (targetFrameworkInfo2.PackageDownloads?.Count > 0)
+                    {
+                        IReadOnlyList<IVsReferenceItem2> packageDownloads = ToReferenceItem2ReadOnlyList(targetFrameworkInfo2.PackageDownloads);
+                        result["PackageDownload"] = packageDownloads;
```
kartheekp-ms:
```suggestion
                        result[ProjectItems.PackageDownload] = packageDownloads;
```

Replaced hardcoded strings with constants. I see a few other hardcoded strings in the changeset; maybe we can replace them with constants.

src/NuGet.Clients/NuGet.SolutionRestoreManager/ProjectRestoreInfo3Adapter.cs
```diff
@@ -0,0 +1,210 @@
...
+
+#nullable enable
+
+using System;
+using System.Collections.Generic;
+using NuGet.ProjectManagement;
+
+namespace NuGet.SolutionRestoreManager
+{
+    internal record ProjectRestoreInfo3Adapter : IVsProjectRestoreInfo3
```
kartheekp-ms: The adapter is adapting both the `IVsProjectRestoreInfo` and `IVsProjectRestoreInfo2` interfaces. Given the different implementations for each, splitting into two classes would improve readability and maintainability. This way, changes to one adapter won't affect the other, adhering to the Single Responsibility Principle.

While this increases the number of classes, we can mitigate code duplication by refactoring common functionality into shared helper methods. This balances the benefits of separation with maintainability and readability.

src/NuGet.Clients/NuGet.VisualStudio/SolutionRestoreManager/IVsProjectRestoreInfo3.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// <summary>
+    /// Contains project metadata needed for project restore operation.
+    /// </summary>
+    public interface IVsProjectRestoreInfo3
+    {
+        /// <summary>
+        /// The MSBuildProjectExtensionsPath of the project (originally BaseIntermediateOutputPath was used,
+        /// but rather than create a new interface, we changed the meaning of this property).
+        /// </summary>
+        string BaseIntermediatePath { get; }
```
nkolev92: MSBuildProjectExtensionsPath. 

If we're making a breaking change, might as well fully commit to it :D

src/NuGet.Clients/NuGet.VisualStudio/SolutionRestoreManager/IVsProjectRestoreInfo3.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        /// <summary>
+        /// Collection of tool references.
+        /// </summary>
+        IReadOnlyList<IVsReferenceItem2>? ToolReferences { get; }
+
+        /// <summary>
+        /// Original raw value of TargetFrameworks property as set in a project file.
+        /// </summary>
+        string? OriginalTargetFrameworks { get; }
```
nkolev92, line 37: We get the TargetFramework property in each property bag per framework. 

I wonder if we still need to duplicate this.

src/NuGet.Clients/NuGet.VisualStudio/SolutionRestoreManager/IVsTargetFrameworkInfo.cs
```diff
@@ -21,17 +23,17 @@ public interface IVsTargetFrameworkInfo
         /// <summary>
         /// Collection of project references.
         /// </summary>
-        IVsReferenceItems ProjectReferences { get; }
+        IVsReferenceItems? ProjectReferences { get; }
```
nkolev92, line 27: comment: Not really feedback, but it's unfortunate that everything is nullable.

src/NuGet.Clients/NuGet.VisualStudio/SolutionRestoreManager/IVsTargetFrameworkInfo4.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Collections.Generic;
+
+namespace NuGet.SolutionRestoreManager
+{
+    public interface IVsTargetFrameworkInfo4
+    {
+        /// <summary>
+        /// Target framework name in full format.
+        /// </summary>
+        string TargetFrameworkMoniker { get; }
```
nkolev92: I don't think this is used anywhere anymore is it?

Should it be the free from TargetFramework value instead?

src/NuGet.Clients/NuGet.SolutionRestoreManager/ProjectRestoreInfo3Adapter.cs
```diff
@@ -0,0 +1,210 @@
...
+    {
+        public required string BaseIntermediatePath { get; init; }
```
nkolev92: Similar throughout, call is MSBuildProjectExtensionsPath. 

Since we're completely reimplementing the API, might as well get the names current (note that I didn't say "correct", cause that BaseIntermediatePath was correct at the point of implementation.).

src/NuGet.Clients/NuGet.SolutionRestoreManager/ProjectRestoreInfo3Adapter.cs
```diff
@@ -0,0 +1,210 @@
...
+
+        public string? OriginalTargetFrameworks { get; init; }
```
nkolev92: If this is unused, I'd remove it from the adapter as well.

src/NuGet.Clients/NuGet.SolutionRestoreManager/VSNominationUtilities.cs
```diff
@@ -90,72 +91,65 @@ internal static RuntimeGraph GetRuntimeGraph(IEnumerable targetFrameworks)
...
             {
-                tfi.Dependencies.AddRange(
-                    targetFrameworkInfo.PackageReferences
-                        .Cast<IVsReferenceItem>()
-                        .Select(pr => ToPackageLibraryDependency(pr, cpvmEnabled)));
+                return tfi;
             }
 
-            if (targetFrameworkInfo is IVsTargetFrameworkInfo2 targetFrameworkInfo2)
+            if (targetFrameworkInfo.Items.TryGetValue(ProjectItems.PackageReference, out var packageReferences))
```
nkolev92, line 121: nit: Use type instead of var.

Improves readability.

src/NuGet.Clients/NuGet.SolutionRestoreManager/VSNominationUtilities.cs
```diff
@@ -90,72 +91,65 @@ internal static RuntimeGraph GetRuntimeGraph(IEnumerable targetFrameworks)
...
             {
-                if (targetFrameworkInfo2.PackageDownloads != null)
-                {
-                    tfi.DownloadDependencies.AddRange(
-                       targetFrameworkInfo2.PackageDownloads
-                           .Cast<IVsReferenceItem>()
-                           .SelectMany(ToPackageDownloadDependency));
-                }
+                tfi.Dependencies.AddRange(packageReferences.Select(pr => ToPackageLibraryDependency(pr, cpvmEnabled)));
```
nkolev92, line 123: This should come in a future PR, but all these AddRange and Select in nomination codepaths can probably be rewritten. 

We probably just need an issue.

src/NuGet.Clients/NuGet.SolutionRestoreManager/VSNominationUtilities.cs
```diff
@@ -90,72 +91,65 @@ internal static RuntimeGraph GetRuntimeGraph(IEnumerable targetFrameworks)
...
-                if (targetFrameworkInfo2.FrameworkReferences != null)
-                {
-                    PopulateFrameworkDependencies(tfi, targetFrameworkInfo2);
-                }
+            if (cpvmEnabled && targetFrameworkInfo.Items.TryGetValue("PackageVersion", out var centralPackageVersions))
+            {
+                tfi.CentralPackageVersions.AddRange(
+                    centralPackageVersions
+                       .Select(ToCentralPackageVersion)
+                       .Distinct(CentralPackageVersionNameComparer.Default)
```
nkolev92, line 136: Do we still need the distinct here, given that we have a dictionary of id?

src/NuGet.Clients/NuGet.SolutionRestoreManager/VSNominationUtilities.cs
```diff
@@ -510,10 +514,10 @@ private static ProjectRestoreReference ToProjectRestoreReference(IVsReferenceIte
             return dependency;
         }
 
-        private static bool TryGetVersionRange(IVsReferenceItem item, string propertyName, out VersionRange versionRange)
+        private static bool TryGetVersionRange(IVsReferenceItem2 item, string propertyName, [NotNullWhen(true)] out VersionRange? versionRange)
```
nkolev92: Looking at this method, I think it'd throw if the provided version range is unparsable right? 

It makes the "try" name confusing.

src/NuGet.Clients/NuGet.SolutionRestoreManager/VsSolutionRestoreService.cs
```diff
@@ -40,7 +41,8 @@ namespace NuGet.SolutionRestoreManager
     [Export(typeof(IVsSolutionRestoreService2))]
     [Export(typeof(IVsSolutionRestoreService3))]
     [Export(typeof(IVsSolutionRestoreService4))]
-    public sealed class VsSolutionRestoreService : IVsSolutionRestoreService, IVsSolutionRestoreService2, IVsSolutionRestoreService3, IVsSolutionRestoreService4
+    [Export(typeof(IVsSolutionRestoreService5))]
```
nkolev92, line 44: What an overachiever this class is.

src/NuGet.Clients/NuGet.SolutionRestoreManager/VsSolutionRestoreService.cs
```diff
@@ -178,35 +206,48 @@ public Task<bool> NominateProjectAsync(string projectUniqueName, IVsProjectResto
...
             }
 
-            if (projectRestoreInfo == null && projectRestoreInfo2 == null)
+            if (projectRestoreInfo == null)
             {
                 throw new ArgumentNullException(nameof(projectRestoreInfo));
             }
 
-            if (projectRestoreInfo != null && projectRestoreInfo2 != null)
+            if (string.IsNullOrEmpty(projectRestoreInfo.BaseIntermediatePath))
```
nkolev92: Guessing this is supposed to be an improvement?

nkolev92 (commented):
Looking great. 

A few suggestions that I think make the APIs even better. 
We know more than what we did when these APIs were originally written, so we can use that.

src/NuGet.Clients/NuGet.SolutionRestoreManager/ProjectRestoreInfo3Adapter.cs
zivkan: It also adapts the `IVsProjectRestoreInfo3` interface, but the way it was implemented was by extending the `2` interface and casting, so that it didn't need to add a new `IVsProjectRestoreServiceX` interface. If you consider the responsibility to be "convert older ProjectRestoreInfo interfaces into the new data structure", then it is a single responsibility. Especially since:
* IVsProjectRestoreInfo1-3 were strictly supersets of each prior version.
* External components implement these interfaces in order to pass data back to NuGet, hence we can't make any changes to any of these interfaces, as it would break nomination as soon as we insert any change into VS.

For these reasons, the `IVsProjectRestoreInfo*` interfaces are adhere strictly to the Open/Close Principal. Therefore, I don't believe maintainability is a concern because we can't change the interfaces.

I don't agree that readability is a concern, because all the differences for the `2` interface are handled within a single `if (pri is ...)` statement, and similarly there's only 1 type check/cast for the `3` interface. Splitting this adapter into two or three different implementations, either needs duplicate code, or needs composition of methods, which harms readability since you no longer have a single place to see all the work that's happening.

Finally, as mentioned in the PR description, once the dotnet/project-system PR is merged to start using the new `IVsProjectRestoreInfo4` interface, then this adpater will become unused/dead code. We have telemetry on invocation counts of all NuGet's VS extensibility APIs, including all the nomination APIs. So we can query the telemetry to validate and then delete all the old interfaces and code at the next major VS version. But for now, we temporarily need this adapter to allow a transition because dotnet/project-system doesn't insert into VS in the same commit as NuGet does, plus in theory other project systems could be using the API.

src/NuGet.Clients/NuGet.VisualStudio/SolutionRestoreManager/IVsProjectRestoreInfo3.cs
zivkan: The `PackageSpec` class has a property `OriginalTargetFrameworks` which gets populated by this. We could generate an equivalent list by iterating each `TargetFramework`. If you have an opinion, please share. Otherwise, I'll think about it while I'm addressing all the other feedback.

src/NuGet.Clients/NuGet.VisualStudio/SolutionRestoreManager/IVsTargetFrameworkInfo.cs
zivkan: This gets into a more philosophical decision. Much like deserializing input, this gets provided to us by an external component. While we trust dotnet/project-system to do the right thing, we can't prevent 3rd party VS project systems from trying to call this API as well. If they don't enable nullable in their own repo and don't set this value to a non-null value, if we don't handle the nulls in our own code, then we'll throw a NullReferenceException and it might be challenging to determine the root cause.  If we say this could be null and we handle the null in our own code, then we won't get unhandled exceptions.

Additionally, there's a question about performance. If we force this collection to be non-null, even when there are no project references, package references, etc, then it forces the project system to make allocations. If we allow these properties to be null when there are zero items to report, then project system has the choice to not allocate. If they do allocate, and the VS Perf team complain about zero length collections, at least NuGet isn't causing the allocations through our design.

src/NuGet.Clients/NuGet.SolutionRestoreManager/VSNominationUtilities.cs
zivkan: The interface in this PR has `IReadOnlyDictionary<string, IReadOnlyList<IVsReferenceItem2>> Items` on the per-TFM data structure, but the key is the item type (PackageReference, PackageVersion, etc). The collection of items isn't a dictionary, just a list, like MSBuild fundamentally allowed. Now, we know that CPS and therefore dotnet/project-system collect items in per-type dictionaries. So maybe NuGet using a list, like MSBuild, is a leaky abstraction. I'm not sure how worried I should be about CPS/project-system in the future changing their mind and changing back to the MSBuild "shape" API, and treating them as lists that allow duplicate identifiers. This way NuGet won't need to change nomination API again.

Having said that, `Distinct` has a perf impact, so maybe we should avoid it anyway. We also de-duplicate in the `Collect*` targets, plus we know that for now CPS/project-system use a dictionary so they're not technically capable of providing us duplicates at this time.

src/NuGet.Clients/NuGet.SolutionRestoreManager/VSNominationUtilities.cs
zivkan: I guess whoever named the method did so because it can return null when the metadata isn't defined. I'll change it.

src/NuGet.Clients/NuGet.SolutionRestoreManager/VsSolutionRestoreService.cs
zivkan: It's "insurance" against a project system with a bad implementation.

If the value is null, it'll be passed into `Path.Combine` and throw an `ArgumentNullException`, but the exception message will be non-obvious that it's a problem from the caller violating NuGet's API contract. By checking and providing a clear error message, it'll be easier for 3rd party project systems to see that they're not meeting our API's expectations.

If the value is an empty string, we're going to write the assets file, and all NuGet's other output files, to the project directory, which is also clearly wrong.

As I mentioned elsewhere, just because we annotate the property as not-null, we don't know if external (to NuGet.Client) callers have enabled nullable checks in their project.

src/NuGet.Clients/NuGet.VisualStudio/SolutionRestoreManager/IVsProjectRestoreInfo3.cs
zivkan: We also use this property to set a `CrossTargeting` property in the `PackageSpec`, even if the project's `TargetFrameworks` only contains a single framework.

Therefore, there's no other way to reproduce this behaviour, except to keep this `OriginalTargetFrameworks` property.

src/NuGet.Clients/NuGet.SolutionRestoreManager/VSNominationUtilities.cs
zivkan: * https://github.com/NuGet/Home/issues/13449

src/NuGet.Clients/NuGet.SolutionRestoreManager/ProjectRestoreInfo3Adapter.cs
kartheekp-ms: Maybe my understanding is wrong, but I still think this adapter class has multiple responsibilities as mentioned below:
- Converts `IVsProjectRestoreInfo` or `IVsProjectRestoreInfo2` into `IVsProjectRestoreInfo3`.
- Converts `IVsTargetFrameworks` and `IVsTargetFrameworks2` into `IVsTargetFrameworkInfo4`.
- Converts `IVsReferenceItems` into `IVsReferenceItem2`.
- Converts properties `IVsProjectProperties` and `IVsReferenceProperties` into dictionary formats.

While these tasks are related to the main responsibility of adapting project restore information, each of them can be considered a distinct responsibility.

> Once the dotnet/project-system PR is merged to start using the new `IVsProjectRestoreInfo4` interface, this adapter will become unused/dead code. 

Given this, please consider creating a follow-up issue to track the telemetry and remove this adapter code in the next major VS version.

src/NuGet.Clients/NuGet.VisualStudio/SolutionRestoreManager/IVsTargetFrameworkInfo4.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// A type representing all project inputs (items and properties) for the given target framework.
+    /// It is implemented by project systems sending nominations to NuGet.
+    /// </summary>
+    public interface IVsTargetFrameworkInfo4
+    {
+        /// <summary>
+        /// Collection of item types. The dictionary key is the MSBuild item type (for example, PackageReference), and
+        /// the value is the list of items of that item type.
+        /// </summary>
+        IReadOnlyDictionary<string, IReadOnlyList<IVsReferenceItem2>>? Items { get; }
```
kartheekp-ms, line 20: I am sure you may have thought about it, but I am curious if we are losing any type-safety benefits by using a bag of properties populated by the project system while nominating a project for restore. I understand the limitations of the existing design where we have to add new interfaces every time, we want the project system to pass a new MSBuild property during nomination, as it requires changes on the project system side as well.

src/NuGet.Clients/NuGet.VisualStudio/SolutionRestoreManager/IVsProjectRestoreInfo3.cs
nkolev92: You are right. 

OriginalTargetFrameworks tells us whether `TargetFramework` or `TargetFrameworks` was used to define the frameworks. 

Thanks for digging into this.

src/NuGet.Clients/NuGet.SolutionRestoreManager/VSNominationUtilities.cs
nkolev92: I think it's less risky to let it be at this point. 

Your call though. As far as I'm concerned, this is ok with no change.

nkolev92: fwiw, there are other ways to improve the perf of the above.

For example, create a set with the right size, then add everything there instead of select and then distinct.

src/NuGet.Clients/NuGet.SolutionRestoreManager/ProjectRestoreInfo3Adapter.cs
```diff
@@ -11,7 +11,7 @@ namespace NuGet.SolutionRestoreManager
 {
     internal record ProjectRestoreInfo3Adapter : IVsProjectRestoreInfo3
     {
-        public required string BaseIntermediatePath { get; init; }
+        public required string MsBuildProjectExtensionsPath { get; init; }
```
nkolev92: from a casing perspective, don't we use `MSBuildProjectExtensionsPath`?

Should it be:
```suggestion
        public required string MSBuildProjectExtensionsPath { get; init; }
```

nkolev92 (dismissed):

src/NuGet.Clients/NuGet.SolutionRestoreManager/ProjectRestoreInfo3Adapter.cs
zivkan: I see that `RestoreRequest` does use `MS`, rather than `Ms`. However, the .NET naming convention, that NuGet mostly adheres to, is to lowercase initialisms and abbreviations. Hence [SqlClient](https://learn.microsoft.com/en-us/dotnet/api/system.data.sqlclient?view=net-8.0), not SQLClient. [XmlReader](https://learn.microsoft.com/en-us/dotnet/api/system.xml.xmlreader?view=net-8.0), not XMLReader. [JsonSerializer](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializer?view=net-8.0), not JSONSerializer.

The adapter class that this comment is made on implements the `IVsProjectRestoreInfo3` interface, that is defined in NuGet.VisualStudio, which is the interface contract that project system implements. Since this is our external contract, I thought it would be better to follow the .NET naming convention more closely, rather than be consistent with NuGet's "incorrect" existing usage.

nkolev92: My logic is the follow the documented property: https://learn.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-properties?view=vs-2022

MSBuild tends to be capitalized in their APIs too, example: https://learn.microsoft.com/en-us/dotnet/api/microsoft.build.locator.msbuildlocator?view=msbuildlocator-1.5. 

I think you'll see MSBuild more commonly referred to instead of MsBuild. 

Their Microsoft.Build.Framework types seem to be avoiding using of msbuild as a word though: https://learn.microsoft.com/en-us/dotnet/api/microsoft.build.framework?view=msbuild-17-netcore.

kartheekp-ms (dismissed):

src/NuGet.Clients/NuGet.SolutionRestoreManager/ProjectRestoreInfo3Adapter.cs
zivkan: I made some minor changes, but structurally it's still similar to before. Now, the classes that implement the new interfaces are POCOs, no logic at all, not even in the constructor. The `ProjectRestoreInfo3Adapter` class is now just a basic Factory pattern, creating instances of the POCOs, by transforming the older interfaces.

Each method in the factory is single responsibility, transforming just a single type, delegating to other methods when one type contains references to another type.

I still feel that it's most maintainable and readable in one file, rather than split across multiple files, since it's all in one place.

src/NuGet.Clients/NuGet.VisualStudio/SolutionRestoreManager/IVsTargetFrameworkInfo4.cs
zivkan: All our other restore paths are already use similar code, because MSBuild's APIs are similar to this PR's generic item bag. For example, see:

* https://github.com/NuGet/NuGet.Client/pull/5679/files#diff-69bcc4cc18dd020ef093e419c357cd2a3452094ec27be239a03c665b32de863b
* https://github.com/NuGet/NuGet.Client/pull/5714/files#diff-1c1b03f2151db1b5a3e33e52b3eb6421948e0cc949a4333acb21e960e9b2f06c

There might be more risk by using a generic bag, rather than properties with specific names, but that was just offloading the responsibility to dotnet/project-system. It doesn't avoid some code somewhere needing to pick out items by name out of a generic API.

It may warrant closer scrutiny in pull requests, but it's similar code to what we're doing for non-SDK projects in VS, and also all projects on the command line.

src/NuGet.Clients/NuGet.SolutionRestoreManager/ProjectRestoreInfo3Adapter.cs
zivkan: > My logic is the follow the documented property

That page lists the property names in MSBuild scripting language, which has different conventions to C#/.NET. Therefore, I don't consider it precedent for .NET names.

> MSBuild tends to be capitalized in their APIs too

Two wrongs don't make a right 😆   It's also an outlier. Neither of us can find other examples of .NET APIs that use any "msbuild"

> I think you'll see MSBuild more commonly referred to instead of MsBuild

In what context?  I agree in English you'll see people write MSBuild, but English isn't C#. In English you'll see SQL or sql, but not so much Sql. Same for XML. Although I feel like Json is common, though technically wrong as it's technically JSON, as JSON is an initialism, and the ECMA spec uses all caps, as does most formal documentation.

Anyway, this is an API that in theory only a single repo is going to implement, so it doesn't really matter. I just think that in a .NET API that `MsBuild` complies with the naming convention, but `MSBuild` does not.

nkolev92 (approved):

src/NuGet.Clients/NuGet.SolutionRestoreManager/ProjectRestoreInfo3Adapter.cs
nkolev92: I'm fine either way. My logic has been to align with how people are used to writing the property.

Don't want to be a sticking point on something that's amazing, and will stay amazing regardless of the casing of s. 

I see you have MSBuild right now, but I'm comfortable if you want to change it.
I've approved the current iteration, happy to approve a different one as well if you feel strongly about it.

PR #5809, opened 2024-05-17
Enable EnforceCodeStyleInBuild by default

jgonz120:
## Bug

Fixes:

## Description

Updated the common.project.props to properly default the EnforceCodeStyleInBuild property to "true". Fixed build errors that arose from this

## PR Checklist

- [ ] PR has a linked issue.
- [X] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [X] N/A 

- **Documentation**
  
  - **OR**
  - [X] N/A


build/common.project.props
```diff
@@ -124,13 +124,17 @@
     <MinimumVisualStudioVersion>15.0</MinimumVisualStudioVersion>
     <VsixOutputDirName>VS15</VsixOutputDirName>
     <CodeAnalysisRuleSet Condition="'$(CodeAnalysisRuleSet)' == ''">$(BuildCommonDirectory)NuGet.ruleset</CodeAnalysisRuleSet>
-    <EnforceCodeStyleInBuild Condition=" '$(EnforceCodeStyleInBuild)' != '' " >true</EnforceCodeStyleInBuild>
+    <WarningsNotAsErrors>EnableGenerateDocumentationFile</WarningsNotAsErrors>
```
nkolev92:
```suggestion
    <WarningsNotAsErrors>$(WarningsNotAsErrors);EnableGenerateDocumentationFile</WarningsNotAsErrors>
```

We should add not overwrite.

fwiw, I had no clue there's a warning identified as  `EnableGenerateDocumentationFile` and that it can be suppressed like that :D

build/common.project.props
```diff
@@ -124,13 +124,17 @@
...
   </PropertyGroup>
 
   <PropertyGroup>
     <VsixPublishDestination>$(ArtifactRoot)$(VsixOutputDirName)\</VsixPublishDestination>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(EnforceCodeStyleInBuild)' != 'false'">
```
nkolev92: `== ''` should have been the pattern. 

Any reason not to update that on line 127?

Many properties are defined there.

jgonz120: The documentation on conditions shows that only certain elements support the condition statement.  https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-conditions?view=vs-2022#supported-elements

build/common.project.props
jgonz120: I wasn't sure it would work tbh xD

build/common.project.props
jgonz120: But ya, switched it back to == ''

jgonz120: Oh I guess "Property" counts as any property? whatever lol

build/TestShared/GlobalSuppressions.cs
```diff
@@ -5,17 +5,3 @@
 // attributes that are applied to this project.
 // Project-level suppressions either have no target or are given 
 // a specific target and scoped to a namespace, type, member, etc.
```
zivkan: This file is now effectively empty (it only has comments). You should delete it.

nkolev92: Makes sense, we should delete this.

nkolev92 (dismissed):

build/common.project.props
```diff
@@ -124,7 +124,8 @@
     <MinimumVisualStudioVersion>15.0</MinimumVisualStudioVersion>
     <VsixOutputDirName>VS15</VsixOutputDirName>
     <CodeAnalysisRuleSet Condition="'$(CodeAnalysisRuleSet)' == ''">$(BuildCommonDirectory)NuGet.ruleset</CodeAnalysisRuleSet>
-    <EnforceCodeStyleInBuild Condition=" '$(EnforceCodeStyleInBuild)' != '' " >true</EnforceCodeStyleInBuild>
+    <EnforceCodeStyleInBuild Condition="'$(EnforceCodeStyleInBuild)' == ''">true</EnforceCodeStyleInBuild>
+    <WarningsNotAsErrors>$(WarningsNotAsErrors);EnableGenerateDocumentationFile</WarningsNotAsErrors>
```
kartheekp-ms: I am curious to learn more about the benefits of `WarningsNotAsErrors` and `EnableGenerateDocumentationFile` properties.

build/common.project.props
```diff
@@ -124,7 +124,8 @@
     <MinimumVisualStudioVersion>15.0</MinimumVisualStudioVersion>
     <VsixOutputDirName>VS15</VsixOutputDirName>
     <CodeAnalysisRuleSet Condition="'$(CodeAnalysisRuleSet)' == ''">$(BuildCommonDirectory)NuGet.ruleset</CodeAnalysisRuleSet>
-    <EnforceCodeStyleInBuild Condition=" '$(EnforceCodeStyleInBuild)' != '' " >true</EnforceCodeStyleInBuild>
+    <EnforceCodeStyleInBuild Condition="'$(EnforceCodeStyleInBuild)' == ''">true</EnforceCodeStyleInBuild>
```
kartheekp-ms, line 127: If I am reading the [document](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/overview?tabs=net-8#latest-updates) correctly, we should also set the `AnalysisLevel` MSBuild property to a specific value to lock the warnings to a particular version of SDK. Otherwise, we may get frequent build failures as we keep building on the preview versions of SDK.

kartheekp-ms, line 127: If I am reading [document](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/overview?tabs=net-8#enable-on-build) correctly, it looks like we are relying on the default severity set by the .NET SDK Analyzers. I am okay with this but just wanted to share that there is an option to set severity levels of these rules. I think we should start with default severity set by the analyzers and adjust as we progress.

build/common.project.props
jgonz120: EnableGenerateDocumentationFile is a warning that causes the build to fail when EnforceCodeStyleInBuild is enabled and IDE0005 is set to warning or greater. IDE0005 sets a warning for unused import statements. 

So I'm setting EnableGenerateDocumentationFile to not error on compile.

build/common.project.props
jgonz120: I like this, at least we can be intentional about taking in new rules as new versions of the SDK are released. Added.

jgonz120: We have an .editorconfig file with a bunch of values already too.

build/common.project.props
```diff
@@ -124,7 +124,9 @@
...
+    <WarningsNotAsErrors>$(WarningsNotAsErrors);EnableGenerateDocumentationFile</WarningsNotAsErrors>
+    <AnalysisLevel>8.0</AnalysisLevel>
```
nkolev92: I think the spirit of that is to only add it if you need it. 

See details about https://github.com/dotnet/designs/blob/main/proposed/sdk-analysis-level.md, that we collaborated on.

This is a team decision ofc, but I think in general, we want to get a steady stream of warnings, instead of all at once.

jgonz120: You know, I added this and now we're getting more errors 😅

jgonz120: Okay so if I run it with level "Latest" it gives a new set of errors.

But the docs say, "If your project targets .NET 5 or later, or if you've added the [AnalysisMode](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#analysismode) property, the default value is latest." 

It looks like we target net472 for most of our projects, so I guess that adds up. I'm going to remove it for now bc I don't want to go about fixing all those changes in all those projects 😅

build/common.project.props
jeffkl: Shouldn't this be 

```suggestion
    <WarningsNotAsErrors>$(WarningsNotAsErrors);IDE0005</WarningsNotAsErrors>
```

I see a bunch of warnings on the build https://dev.azure.com/dnceng-public/public/_build/results?buildId=683379&view=results

jgonz120: That would make it so "Remove unnecessary using directives" won't give an error on build which won't actually enforce that rule. 

But I'm okay with it for this PR since I want to start enabling those types of rules project by project in subsequent PRs.

jgonz120: Actually second thought. I switched it to NoWarn, that way IDE0005 can still warn for projects as they are enabled.

nkolev92 (dismissed):

build/DotNetSdkVersions.txt
```diff
@@ -1,3 +1,3 @@
 # Each line represents arguments for the .NET SDK installer script (https://learn.microsoft.com/dotnet/core/tools/dotnet-install-script)
--Channel 8.0
+-Channel 8.0.2xx
```
dtivel, line 2: This seems unrelated to enforcing code style in build.

Can you elaborate on this change?  `8.0` will use the current released version, which is currently 8.0.3xx.  Hard coding this to `8.0.2xx` will lock us into that feature band until someone bumps it.

jgonz120: There's a known issue that was causing the build to fail. If we use 8.0.3xx then there is a mismatch between the compiler versions. We will need to update the value back once the latest version is available in CI. 

https://github.com/dotnet/core/blob/21da990bab4170059c6e9ed84feba7e912fa56bc/release-notes/8.0/known-issues.md#802xx-sdk-is-not-compatible-with-178-for-some-scenarios

donnie-msft (dismissed):

nkolev92 (approved):

build/DotNetSdkVersions.txt
jgonz120: @dtivel is it okay to resolve this comment?

PR #5810, opened 2024-05-20
Change the testMachineImageName's value of NuGet.Client Daily Tests pipeline

SueSu01:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2840

## Description

For recently runs in [NuGet.Client Daily Tests](https://devdiv.visualstudio.com/DevDiv/_build?definitionId=18097&_a=summary) pipeline failed due to an error "There is not enough space on the disk". So changing the testMachineImageName's value to "Windows-11-Enterprise-23H2-WinBuildDevFiles", and it creates a bigger disk(255GB) than the default image(126GB).

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/eng/pipelines/dailytests.yml
+++ b/eng/pipelines/dailytests.yml
@@ -32,6 +32,9 @@ parameters:
 - name: isOfficialBuild
   type: boolean
   default: false
+- name: testMachineImageName
+  type: string
+  default: Windows-11-Enterprise-23H2-WinBuildDevFiles
 
 variables:
   DOTNET_NOLOGO: 1
@@ -46,3 +49,4 @@ stages:
     dartLabEnvironment: ${{parameters.DartLabEnvironment}}
     testExecutionJobTimeoutInMinutes: 150
     testMachineCleanUpStrategy: ${{parameters.ApexAgentCleanup}}
+    testMachineImageName: ${{parameters.testMachineImageName}}
--- a/eng/pipelines/templates/Apex_Tests_On_Windows_Daily.yml
+++ b/eng/pipelines/templates/Apex_Tests_On_Windows_Daily.yml
@@ -12,6 +12,8 @@ parameters:
   - name: isOfficialBuild
     type: boolean
     default: false
+  - name: testMachineImageName
+    type: string
 
 stages:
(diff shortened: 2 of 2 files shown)


eng/pipelines/dailytests.yml
```diff
@@ -32,6 +32,9 @@ parameters:
 - name: isOfficialBuild
   type: boolean
   default: false
+- name: testMachineImageName
+  type: string
+  default: Windows-11-Enterprise-23H2-WinBuildDevFiles
```
nkolev92, line 37: Can someone who has better understanding of the machine pools review this: @NuGet/nuget-client

zivkan: FWIW, I onboarded NuGet onto DartLab, but I don't know anything about their VM images. The default always worked for us for the Apex & E2E machines, and by using their defaults, it also meant we didn't need to maintain the image name over the months & years when new versions of Windows were released, when compliance required changes to image names, etc.

Setting a specific image name here might be a fine temporary mitigation, but I believe we should contact the DartLab team to understand why there's no longer enough disk space, and to see if they're going to change the default image, so that we don't have to maintain the VM image name in our yaml long term.

nkolev92: Thanks @zivkan 
Let's start with that. 

Maybe we can try to find the contact info...start with: https://github.com/NuGet/Client.Engineering/blob/main/docs/CI/build-pipeline.md#dartlab, and file this with them.

nkolev92:
Closing as we want to avoid taking explicit dependencies ourselves.

eng/pipelines/dailytests.yml
nkolev92: @SueSu01 Can you try contacting?
If you prefer, someone from the team can do it instead.

SueSu01: > @SueSu01 Can you try contacting? If you prefer, someone from the team can do it instead.

@nkolev92 I sent an email to the ImageFactory team. Thanks.

PR #5815, opened 2024-05-23
Add new ctor for PackageIdentity

jairbubbles:
Syntaxic sugar allowing to instantiate like that

```csharp
var package = new PackageIdentity("Newtonsoft.json", "13.0.1");
```

Rather than

```csharp
var package = new PackageIdentity("Newtonsoft.json",  new NuGetVersion("13.0.1"));
```

### DIFF
--- a/src/NuGet.Core/NuGet.Packaging/Core/PackageIdentity.cs
+++ b/src/NuGet.Core/NuGet.Packaging/Core/PackageIdentity.cs
@@ -20,8 +20,8 @@ namespace NuGet.Packaging.Core
         /// Creates a new package identity.
         /// </summary>
         /// <param name="id">name</param>
-        /// <param name="version">version</param>
-        public PackageIdentity(string id, string version)
+        /// <param name="versionStr">version</param>
+        public PackageIdentity(string id, string versionStr)
         {
             if (id == null)
             {
@@ -29,7 +29,7 @@ namespace NuGet.Packaging.Core
             }
 
             _id = id;
-            _version = new NuGetVersion(version);
+            _version = NuGetVersion.Parse(versionStr);
         }
 
         /// <summary>
--- a/src/NuGet.Core/NuGet.Packaging/Core/PackageIdentity.cs
+++ b/src/NuGet.Core/NuGet.Packaging/Core/PackageIdentity.cs
@@ -21,6 +21,22 @@ namespace NuGet.Packaging.Core
         /// </summary>
         /// <param name="id">name</param>
         /// <param name="version">version</param>
+        public PackageIdentity(string id, string version)
+        {
+            if (id == null)
+            {
+                throw new ArgumentNullException(nameof(id));
+            }
+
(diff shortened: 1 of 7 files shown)


nkolev92 (changes requested):
Given that there's only 1 constructor, and the simplification is small, I don't think we should take this. 

@NuGet/nuget-client @zivkan Thoughts?

Nigusu-Allehu:
I second @nkolev92 . In addition, since the existing constructor is called with a null version value in multiple places in our tests, this will lead to multiple ambiguities. You can see that in the failed job of the pipeline

donnie-msft (commented):
Since there isn't a linked issue, could you describe the problem that this constructor solves?There's already `NuGetVersion.Parse(value)` to help in creating the type.

zivkan:
In general, some of our APIs are not easy to use, and I don't think it's a bad thing to make it easier to use.

However, this specific example can be achieved rather easily by creating a helper class/method, so instead of `new PackageIdentity("package.id", NuGetVersion.Parse("1.2.3")` you'd call your own `NuGetHelper.CreateIdendity("package.id", "1.2.3")`. It does mean that every repo that wants such a helper needs to re-implement it, but it's also trivial. The .NET BCL doesn't try to do everything and provide every overload, so we can use the BCL as prior art in not accepting such a new overload.

I could go either way, but I'd lean towards not accepting this new public API. Thanks to C#'s expression-bodied members, it's literally a one-liner to create your own helper method.

jairbubbles:
> You can see that in the failed job of the pipeline

I guess next time, I should compile before sending the PR. 😅

> Could you describe the problem that this constructor solves?

It's really just to simplify using the API. The context is that we were previously using a custom NuGet library where we re-implemented a lot of things. We're now switching to a new one, which is just a thin wrapper over the official NuGet library but to simplify its usage.

In the old library we had a `NuGetPackage` class which was practical but not well designed as it grew over time and it was doing too much things.

The idea is to switch to `PackageIdentity` and add some extensions methods over this type to simplify the use. Example:

```csharp
var client = new EzNuGetClient();

var package = new PackageIdentity("Newtonsoft.Json", new NuGetVersion("13.0.1"));

await client.InstallPackageInGlobalCache(logger, package);

var installationFolder = package.GetInstallationPathInGlobalCache(client);
```

All this is internal code but I'd be happy to share more info if you're interested.

> However, this specific example can be achieved rather easily by creating a helper class/method, so instead of new PackageIdentity("package.id", NuGetVersion.Parse("1.2.3") you'd call your own NuGetHelper.CreateIdendity("package.id", "1.2.3"). It does mean that every repo that wants such a helper needs to re-implement it, but it's also trivial. The .NET BCL doesn't try to do everything and provide every overload, so we can use the BCL as prior art in not accepting such a new overload.

Yes I agree but it's less discoverable.

nkolev92:
Closing per the discussion.

PR #5822, opened 2024-05-28
NuGet.VisualStudio should depend on packages that exist on nuget.org

zivkan:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2847

Regression? Yes
This commit broke our CI step that should have validated the dependencies: https://github.com/NuGet/NuGet.Client/commit/ff7695dfeafa05fcf1af15e7f880c7bf8507304d
This commit actually changed the dependency to a version that doesn't exist on nuget.org: https://github.com/NuGet/NuGet.Client/commit/e4899ee48ff3d7787ee345f546c818ce6b962807

## Description

* Fix `tools-local/ensure-nupkg-dependencies-on-source`, so it will return a non-zero exit code when it finds a dependency that doesn't exist on nuget.org.
* The latest version of Microsoft.VisualStudio.TemplateWizardInterface has a different version to Microsoft.VisualStudio.Sdk, so the template wizard package is no longer using the `VsFrameworkVersion` property for a version number.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -74,7 +74,7 @@
     <PackageVersion Include="Microsoft.VisualStudio.Sdk.TestFramework.Xunit" Version="17.6.32" />
     <PackageVersion Include="Microsoft.VisualStudio.Services.InteractiveClient" Version="$(VSServicesVersion)" />
     <PackageVersion Include="Microsoft.VisualStudio.Setup.Configuration.Interop" Version="3.4.2244" />
-    <PackageVersion Include="Microsoft.VisualStudio.TemplateWizardInterface" Version="$(VSFrameworkVersion)" />
+    <PackageVersion Include="Microsoft.VisualStudio.TemplateWizardInterface" Version="17.10.40170" />
     <PackageVersion Include="Microsoft.VisualStudio.VCProjectEngine" Version="$(VSFrameworkVersion)" />
     <PackageVersion Include="Microsoft.VisualStudio.Workspace.VSIntegration" Version="17.0.7-preview-0001-g5492e466a9" />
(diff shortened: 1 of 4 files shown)


jeffkl (approved):

nkolev92 (approved):

Directory.Packages.props
```diff
@@ -74,7 +74,7 @@
     <PackageVersion Include="Microsoft.VisualStudio.Sdk.TestFramework.Xunit" Version="17.6.32" />
     <PackageVersion Include="Microsoft.VisualStudio.Services.InteractiveClient" Version="$(VSServicesVersion)" />
     <PackageVersion Include="Microsoft.VisualStudio.Setup.Configuration.Interop" Version="3.4.2244" />
-    <PackageVersion Include="Microsoft.VisualStudio.TemplateWizardInterface" Version="$(VSFrameworkVersion)" />
+    <PackageVersion Include="Microsoft.VisualStudio.TemplateWizardInterface" Version="17.10.40170" />
```
donnie-msft, line 77: We have many different VS versions in play. Should we start finding a way to consolidate and centralize these versions?

donnie-msft (approved):

Directory.Packages.props
zivkan: We depend on the various different Microsoft.VisualStudio.* packages having the same version, which sadly was not the case for 17.10, as mentioned in the PR description.

PR #5827, opened 2024-05-29
Use the local copy of the consumption feed - avoid unwanted auth prompt

nkolev92:
## Bug

Fixes:

## Description

https://github.com/NuGet/NuGet.Client/commit/5cb34e019ee00d70af207b666aaebb3fe25f8638 introduced a new package version that leads to indeterminism with the consumption feed which is an upstream and can trigger lookups. 

This should resolve that concurrency. 

## PR Checklist

- [ ] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -7,7 +7,7 @@
     <add key="dotnet-libraries" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries/nuget/v3/index.json" />
     <add key="myget-legacy@Local" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/myget-legacy%40Local/nuget/v3/index.json" />
     <add key="nuget-build" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/nuget-build/nuget/v3/index.json" />
-    <add key="vside" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/msft_consumption/nuget/v3/index.json" />
+    <add key="vside" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/msft_consumption%40Local/nuget/v3/index.json" />
   </packageSources>
   <packageSourceMapping>
     <clear />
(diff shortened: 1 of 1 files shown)


zivkan (dismissed):

NuGet.Config
```diff
@@ -7,7 +7,7 @@
     <add key="dotnet-libraries" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries/nuget/v3/index.json" />
     <add key="myget-legacy@Local" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/myget-legacy%40Local/nuget/v3/index.json" />
     <add key="nuget-build" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/nuget-build/nuget/v3/index.json" />
-    <add key="vside" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/msft_consumption/nuget/v3/index.json" />
+    <add key="vside" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/msft_consumption%40Local/nuget/v3/index.json" />
```
kartheekp-ms, line 10: Thanks for creating this PR. I thought local copy of the consumption feed is accessed first before reaching to upstream feeds.

May be should we update the package source mapping settings?

nkolev92: It works differently depending on the resource. 
For example, listing from flat container is different from downloading the nupkg.
| | list package versions | download |
|-|-|-|
| local view | Only ingested packages | Only ingested packages, matches the listed list |
| main view | All packages | Only ingested, but can download with auth, which we don't have |

nkolev92: https://learn.microsoft.com/en-us/azure/devops/artifacts/how-to/public-feeds-upstream-sources?view=azure-devops&tabs=nuget#q-im-trying-to-restore-my-packages-but-i-keep-getting-a-401-unauthorized-error sort of talks about it, but not in detail.

kartheekp-ms (approved):

PR #5844, opened 2024-06-07
Bump VisualStudio dependencies to the 17.10 versions

donnie-msft:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2873

## Description

- Bump VS Packages to 17.10 which we will need in order to onboard to Unified Settings.
- No longer need `@local` for Azure Artifacts feeds as I discovered the problem was missing permissions for the team.
- Added explicit source mappings so that package versions that only exist on certain feeds are not checked for on other feeds (as a 401 will result).
- 1 assembly was added to the VSIX project which isn't used by our product.
- Suppressed a bunch of new analyzers. I can add whatever justification or tracking issue is preferred, so will wait on feedback.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


NuGet.Config
```diff
@@ -7,7 +7,7 @@
     <add key="dotnet-libraries" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries/nuget/v3/index.json" />
     <add key="myget-legacy@Local" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/myget-legacy%40Local/nuget/v3/index.json" />
     <add key="nuget-build" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/nuget-build/nuget/v3/index.json" />
-    <add key="vside" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/msft_consumption%40Local/nuget/v3/index.json" />
```
kartheekp-ms: I am wondering, in the worst-case scenario, if having the `%40Local` would be helpful because we don't run restore in interactive mode, the tooling can't respond to 401 in the `main` view. If we experience this issue again in the future and we don't have `%40Local` in the feed name, then someone from the team has to run the restore locally so that the feed can save the package from the upstream to the local view. Once the local restore succeeds, the restore will also succeed on CI because the local view has the package injected from the upstream.

If the feed name has `%40Local`, then my understanding is that the consumption feed returns a 404 when a package is not found instead of 401, and based on the package source mapping settings, NuGet could search for the same package from other feeds.

donnie-msft: I've un-reverted the change, so we will keep `@local` as it was.

jeffkl (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/GlobalSuppressions.cs
```diff
@@ -113,3 +113,4 @@
...
 [assembly: SuppressMessage("Build", "CA1501:'ConfigurationFilesControl' has an object hierarchy '9' levels deep within the defining module. If possible, eliminate base classes within the hierarchy to decrease its hierarchy level below '6': 'UserControl, ContentControl, Control, FrameworkElement, UIElement, Visual, DependencyObject, DispatcherObject, Object'", Justification = "Default WPF class hierarchy", Scope = "type", Target = "~T:NuGet.PackageManagement.UI.Options.ConfigurationFilesControl")]
+[assembly: SuppressMessage("Usage", "VSTHRD003:Avoid awaiting foreign Tasks", Justification = "<Pending>", Scope = "member", Target = "~M:NuGet.PackageManagement.UI.PackageItemViewModel.GetDetailedPackageSearchMetadataAsync~System.Threading.Tasks.Task{System.ValueTuple{NuGet.VisualStudio.Internal.Contracts.PackageSearchMetadataContextInfo,NuGet.VisualStudio.Internal.Contracts.PackageDeprecationMetadataContextInfo}}")]
```
zivkan: I'm pretty sure Nikolche generally disagrees with me on this, but I would much prefer the suppressions be done inline with `#pragma` statements, rather than GlobalSuppressions.cs.

The reason is that when I'm working on some code with a supressed analyzer warning, I can immeidately fix the issue and reduce tech debt as part of my other work.  Similarly if someone else creates a PR and I see the `#pragma warning disable` just above or below their changes, I can comment asking them to make a small tech debt improvement, since they're already modifying that code anyway.

When suppressions are hidden in GlobalSuppression.cs, then it's much more difficult to know that there's even tech debt to improve.

Plus, global suppressions work on entire methods, so if someone adds new code that also violates the same analyzer in the same method, they won't get warned with GlobalSuppresions, but they will with inline suppressions.

Directory.Packages.props
```diff
@@ -6,7 +6,7 @@
     <NewtonsoftJsonPackageVersion Condition="'$(NewtonsoftJsonPackageVersion)' == ''">13.0.3</NewtonsoftJsonPackageVersion>
     <SystemTextJsonVersion Condition="'$(SystemTextJsonVersion)' == ''">7.0.3</SystemTextJsonVersion>
     <SystemPackagesVersion Condition="'$(SystemPackagesVersion)' == ''">4.3.0</SystemPackagesVersion>
-    <VSFrameworkVersion Condition="'$(VSFrameworkVersion)' == ''">17.8.37297</VSFrameworkVersion>
+    <VSFrameworkVersion Condition="'$(VSFrameworkVersion)' == ''">17.10.39524</VSFrameworkVersion>
```
zivkan: Further down the file, there are some project-dependent `PackageVersion` items, just for our VS Extensibility packages that depend on specific packages, rather than the whole VS SDK. It'd be good to update these packages to the same versions that this VS SDK meta package references:

https://github.com/NuGet/NuGet.Client/blob/de9c314080687bfe4da4baf4b3d1a9d38b6ace3a/Directory.Packages.props#L119-L135

src/NuGet.Clients/NuGet.PackageManagement.UI/GlobalSuppressions.cs
donnie-msft: Ok, could you find a way to add this to our coding guidance so that we aren't informing PR authors that are already in code review?

Directory.Packages.props
donnie-msft: > These individual package versions should be equal to or less than the package version within the VS SDK that our
>     customers use, otherwise they might get NU1605 or MSB3277 warnings.

Do you know the answer to the question of which SDK version our customers use?

After changing these package versions to match the version used by the VS SDK, the first error I'm getting seems to want me to pick a version. I'm not sure if I pick the one the VS SDK references, or if there's some other lowest common denominator that customers would use.

> Severity	Code	Description	Project	File	Line	Suppression State	Details
> Error (active)	NU1603	Warning As Error: Microsoft.VisualStudio.ComponentModelHost 17.10.126-preview depends on Microsoft.VisualStudio.Interop (>= 17.10.39408-preview.3) but Microsoft.VisualStudio.Interop 17.10.39408-preview.3 was not found. An approximate best match of Microsoft.VisualStudio.Interop 17.10.39524 was resolved.	NuGet.VisualStudio.Client	C:\NuGet.Client\src\NuGet.Clients\NuGet.VisualStudio.Client\NuGet.VisualStudio.Client.csproj	1		
>

donnie-msft: It seems there's no way to satisfy the condition of:

> These individual package versions should be equal to or less than the package version within the VS SDK that our
>     customers use, otherwise they might get NU1605 or MSB3277 warnings.

The VS SDK dependency, `"Microsoft.VisualStudio.ComponentModelHost" version="17.10.126-preview"`
...depends on Package  `Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime (>= 17.10.39365-preview.3)`
...but VS SDK depends on `"Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime" version="17.10.39500"`

Package `Microsoft.VisualStudio.ComponentModelHost 17.10.133-preview` is the version which also depends on `17.10.39500` as above. However, `17.10.133-preview` is greater than `17.10.126-preview` which violates the rule in the comment.

If you know a way to reconcile this, let me know.

zivkan: I think this is where using public versions of the VS SDK published on nuget.org is helpful, since I think there's some additional process to increase the probability that the VS SDK is "coherent".

On nuget.org, there exists [version 17.10.40171 of the VS SDK](https://www.nuget.org/packages/Microsoft.VisualStudio.SDK/17.10.40171), which is higher than the version that this PR has been proposing.

This version of the VS SDK depends on [Microsoft.VisualStudio.ComponentModelHost 17.10.191](https://www.nuget.org/packages/Microsoft.VisualStudio.ComponentModelHost/17.10.191#dependencies-body-tab), which itself depends on Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime version 17.10.39546, while the VSSDK itself depends on version 17.10.40152, which is higher. Therefore, since the VS SDK is bringing in a higher version of that transitive dependency, I wouldn't expect any NU1603 (package downgrade) issues.

Honestly, I'm wondering if we should make NuGet.VisualStudio.Contracts depend on the VS SDK rather than individual packages. I'm no longer sure that the circular reference concern in the comments is correct. But this may not be feasible if we end up adding brokered services for the C# Dev Kit in order to share implementation between VSCode and VS. Plus it would mean that we couldn't/shouldn't ever use a pre-release VS SDK to integrate with VS features that are not yet GA. But having NuGet use a GA version of the VS SDK is compatible with this PR, since you're not trying to use a 17.11 VS SDK in this PR.

> Do you know the answer to the question of which SDK version our customers use?

I think that external customers are very unlikely to use the VSIDE feed, and probably only use the nuget.org packages. Since our dev branch is always targeting VS version vNext, by using the last stable (GA) version, we'll never release a version of NuGet.VisualStudio packages that have "bad" dependencies, since our package will always depend on the previous minor version of the VS SDK, not the same release.

donnie-msft: I believe 17.10.40171 is the version I tried first, since intuitively as you found, it's the most recent public version.
However, the version in this PR is one I found which had dependencies that existed on our repo's feeds.

I just [ran a build with that version](https://dev.azure.com/dnceng-public/public/_build/results?buildId=703801&view=logs&j=0ab14b9f-e499-56d5-97b1-fd98b70ea339&t=32a24672-ae5e-583b-9364-ae46e0d3f980&l=16 )again to recall which dependency it was, and here's an example. 

>  Error NU1603: Warning As Error: NuGet.VisualStudio.Common depends on Microsoft.Internal.VisualStudio.Shell.Framework (>= 17.10.40171) but Microsoft.Internal.VisualStudio.Shell.Framework 17.10.40171 was not found. An approximate best match of Microsoft.Internal.VisualStudio.Shell.Framework 17.10.40177 was resolved.

Where will we find all these dependencies?

zivkan: I pushed changes to use the versions of the packages actually on nuget.org, even if they aren't the same as what the VS SDK meta package depend on.

src/NuGet.Clients/NuGet.PackageManagement.UI/GlobalSuppressions.cs
donnie-msft: Moved all of the suppressions to pragma statements

NuGet.Config
```diff
@@ -54,12 +59,102 @@
...
+      <package pattern="MessagePack" />
+      <package pattern="Microsoft.AspNetCore.App.Ref" />
+      <package pattern="Microsoft.Build" />
+      <package pattern="Microsoft.Build.*" />
+      <package pattern="Microsoft.Build.Artifacts" />
+      <package pattern="Microsoft.Build.Framework" />
+      <package pattern="Microsoft.Build.Locator" />
+      <package pattern="Microsoft.Build.Runtime" />
+      <package pattern="Microsoft.Build.Tasks.Core" />
+      <package pattern="Microsoft.Build.Utilities.Core" />
```
zivkan, line 74: all these that don't start with `Microsoft.`, are already mapped in `dotnet-public`, so mapping them here to vside makes these packages "less deterministic", since they now might be restored from two sources, rather than just one. MessagePack ***maybe*** has a special version just for VS, but the other ones are very, very unlikely to be used by anything else in VS, so it's not clear to me why mappings for these packaegs were added to vside, especially without removing the mapping from dotnet-public as well.

The MSBuild packages, I'm feeling very confident that there are not special versions in the vside feed, but like MessagePack, it's at least feasible that it could possibly have a VS specific build.

NuGet.Config
```diff
@@ -54,12 +59,102 @@
...
+      <package pattern="Microsoft.CodeAnalysis.BannedApiAnalyzers" />
+      <package pattern="Microsoft.CodeAnalysis.Common" />
+      <package pattern="Microsoft.CodeAnalysis.CSharp" />
+      <package pattern="Microsoft.CodeAnalysis.CSharp.Workspaces" />
+      <package pattern="Microsoft.CodeAnalysis.EditorFeatures.Text" />
+      <package pattern="Microsoft.CodeAnalysis.Features" />
+      <package pattern="Microsoft.CodeAnalysis.PublicApiAnalyzers" />
+      <package pattern="Microsoft.CodeAnalysis.Scripting.Common" />
+      <package pattern="Microsoft.CodeAnalysis.VisualBasic.Workspaces" />
+      <package pattern="Microsoft.CodeAnalysis.Workspaces.Common" />
```
zivkan, line 87: When I had an attempt myself to upgrade to the 17.10 VS SDK, I found there was at least one or two code analysis packages that needed to come from the vside feed, but my memory is that most did not.

Some of these, the BannedApiAnalyzers and PublicApiAnalyzers, just feel infeasible that there are vside special versions, and I'm confident that we're using nuget.org versions available on dotnet-public.

NuGet.Config
```diff
@@ -54,12 +59,102 @@
...
+      <package pattern="Microsoft.CSharp" />
+      <package pattern="Microsoft.DataAI.NuGetRecommender.Contracts" />
+      <package pattern="Microsoft.DiaSymReader" />
+      <package pattern="Microsoft.DotNet.Build.Tasks.Feed" />
+      <package pattern="Microsoft.DotNet.Maestro.Tasks" />
+      <package pattern="Microsoft.Extensions.FileProviders.Abstractions" />
+      <package pattern="Microsoft.Extensions.FileSystemGlobbing" />
+      <package pattern="Microsoft.Extensions.Primitives" />
```
zivkan, line 95: DiaSymReader is a package published from the dotnet/runtime repo, so I can't imagine it needs to come from the vside feed (VS doesn't have a special version).

The Microsoft.Dotnet packages sound like dnceng packages that probably should be restored from dotnet-eng. In fact, the Build.Tasks.Feed package id this PR also adds an explicit mapping to for the dotnet-eng feed in addition to here for the vside feed.

The Microsoft.Extensions.* packages also are dotnet/* repo packages, so seem infeasible that they have VS specific versions.

NuGet.Config
```diff
@@ -54,12 +59,102 @@
...
       <package pattern="microsoft.internal.visualstudio.*" />
+      <package pattern="Microsoft.Internal.VisualStudio.Shell.Framework" />
+      <package pattern="Microsoft.IO.Redist" />
+      <package pattern="Microsoft.Net.Compilers.netcore" />
+      <package pattern="Microsoft.Net.Compilers.Toolset" />
+      <package pattern="Microsoft.NET.StringTools" />
+      <package pattern="Microsoft.NET.Test.Sdk" />
+      <package pattern="Microsoft.NETCore.Platforms" />
+      <package pattern="Microsoft.PowerShell.3.ReferenceAssemblies" />
```
zivkan, line 104: Another set of packages I find very unlikely to have VS specific versions

NuGet.Config
```diff
@@ -54,12 +59,102 @@
...
       <package pattern="microsoft.servicehub.*" />
+      <package pattern="Microsoft.ServiceHub.Framework" />
+      <package pattern="Microsoft.TeamFoundationServer.ExtendedClient" />
       <package pattern="microsoft.test.apex.visualstudio" />
+      <package pattern="Microsoft.TestPlatform.Portable" />
+      <package pattern="Microsoft.TestPlatform.TestHost" />
```
zivkan, line 110: These two sound like dotnet/* packages, so I'd expect them available on dotnet-public.

NuGet.Config
```diff
@@ -54,12 +59,102 @@
...
+      <package pattern="Microsoft.Win32.Registry" />
+      <package pattern="Moq" />
+      <package pattern="MSTest.TestAdapter" />
+      <package pattern="MSTest.TestFramework" />
+      <package pattern="Newtonsoft.Json" />
+      <package pattern="NuGet.Client.EndToEnd.TestData" />
+      <package pattern="NuGet.Core" />
+      <package pattern="NuGetValidator" />
+      <package pattern="Portable.BouncyCastle" />
+      <package pattern="SharpZipLib" />
```
zivkan, line 137: NuGet.Client.EndToEnd.TestData and NuGetValidator has close to zero percent chance of being on the vside feed. There's no way vside (well, msft_consumption) upstreams from our nuget-build feed, and this package should only exist in that nuget-build feed. This pattern already exists for the nuget-build feed, so I'm scratching my head why you needed to add the same pattern to the vside feed.

The others are not even authored by Microsoft (well MSTest is, but I feel like the justification is the same), so it just seems infeasible that there are VS specific versions, so the mapping that already exist for the dotnet-public feed should suffice. It's certainly not clear to me why these mappings were added to vside without removing them from dotnet-public.

NuGet.Config
```diff
@@ -54,12 +59,102 @@
...
+      <package pattern="System.Runtime.Extensions" />
+      <package pattern="System.Runtime.InteropServices" />
+      <package pattern="System.Security.Cryptography.Pkcs" />
+      <package pattern="System.Security.Cryptography.ProtectedData" />
+      <package pattern="System.Security.Cryptography.Xml" />
+      <package pattern="System.Text.Encoding.Extensions" />
+      <package pattern="System.Text.Json" />
+      <package pattern="System.Threading" />
+      <package pattern="System.Threading.Tasks" />
+      <package pattern="System.Threading.Tasks.Dataflow" />
```
zivkan, line 157: Like other comments, I don't beleive any of these system.* packages, all of which should come fromt he dotnet/runtime repo, have VS specific versions. The system.* mapping that already exists for the dotnet-public feed should restore these successfully.

NuGet.Config
```diff
@@ -72,9 +167,12 @@
       <package pattern="vslangproj80" />
       <package pattern="vslangproj90" />
       <package pattern="vswebsite.interop" />
+      <package pattern="xunit" />
+      <package pattern="xunit.runner.visualstudio" />
+      <package pattern="Xunit.StaFact" />
```
zivkan, line 172: Like several other comments, these patterns already exist in the dotnet-public feed, and there's no logical reason why the VS SDK would list any of these packages as a dependency, let alone have some VS specific version that doesn't exist on the dotnet-public feed.

NuGet.Config
```diff
@@ -42,7 +46,8 @@
       <package pattern="xunitxml.testlogger" />
     </packageSource>
     <packageSource key = "dotnet-eng">
-      <package pattern="microsoft.*" />
+      <package pattern="Microsoft.DotNet.*" />
+      <package pattern="Microsoft.DotNet.Build.Tasks.Feed " />
```
zivkan, line 50: the second pattern, the package specific pattern, isn't needed here if the same package ID isn't mapped to another feed as well.

NuGet.Config
```diff
@@ -21,8 +22,11 @@
       <package pattern="messagepack" />
       <package pattern="messagepack.annotations" />
       <package pattern="messagepackanalyzer" />
-      <package pattern="microsoft.*" />
-      <package pattern="Microsoft.VisualStudio.*" />
+      <package pattern="Microsoft.*" />
+      <package pattern="Microsoft.NETCore.App.Host.linux-x64" />
+      <package pattern="Microsoft.NETCore.App.Host.osx-x64" />
+      <package pattern="Microsoft.NETFramework.ReferenceAssemblies.net45" />
+      <package pattern="Microsoft.NETFramework.ReferenceAssemblies.net461" />
```
zivkan, line 29: All the package specific patterns would already be matched by microsoft.* if they're not mapped by another source, and it doesn't make sense to me for any Microsoft.NetCore.* or Microsoft.NETFramework.* pattern to be mapped to any other source.

NuGet.Config
```diff
@@ -54,12 +59,102 @@
...
+      <package pattern="Microsoft.VisualStudio.ProjectSystem.Managed" />
+      <package pattern="Microsoft.VisualStudio.ProjectSystem.Managed.VS" />
+      <package pattern="Microsoft.VisualStudio.ProjectSystem.VS" />
+      <package pattern="Microsoft.VisualStudio.SDK" />
+      <package pattern="Microsoft.VisualStudio.Sdk.TestFramework.Xunit" />
+      <package pattern="Microsoft.VisualStudio.Setup.Configuration.Interop" />
+      <package pattern="Microsoft.VisualStudio.Shell.15.0" />
+      <package pattern="Microsoft.VisualStudio.VCProjectEngine" />
+      <package pattern="Microsoft.VisualStudio.Workspace.VSIntegration" />
+      <package pattern="Microsoft.VisualStudioEng.MicroBuild.Core" />
```
zivkan, line 125: Since Microsoft.VisualStudio.* was removed from dotnet-public, none of these new mappings are needed since they match the microsoft.visualstudio.* pattern.

NuGet.Config
```diff
@@ -54,12 +59,102 @@
...
+      <package pattern="Microsoft.VSSDK.BuildTools" />
+      <package pattern="Microsoft.Web.Xdt" />
+      <package pattern="Microsoft.Win32.Registry" />
```
zivkan, line 128: I think I missed these two from all my other comments, but if the VS SDK have a transitive dependency on either of these packages (registry seems feasible, web.xdt feels very unlikely), then I'd expect them to depend on versions on nuget.org, not VS specific versions.

zivkan (approved):
I know how difficult it has been to get to this point. But there's a whole lot of packages added to the vside feed source mapping config that just don't seem logical to me. The vast majority are packages that alreday exist on nuget.org, and therefore are already mapped to dotnet-public, while you've added a few packages to the vside feed that I just don't believe actually exist in any of vside's upstream sources.

I don't want to artificially cause you further delays, but this PR at the time I'm doing the review, adds a whole lot of complexity and duplication to nuget.config that I think is unnecessary. As you know, I like to 'do things right', so I'd prefer if you simplified and improved nuget.config, but if you need to merge this now, then I'm giving you the approval to unblock the merge button.

NuGet.Config
donnie-msft: Could we go through these offline? I checked the `Microsoft.VisualStudio.ProjectSystem.Managed.VS` package as an example, which has just a few versions on dotnet-public, but appears to have all the recent published packages on vside.

Either I'm misunderstanding what you're stating, or perhaps you're restoring from your Global Packages Folder or something and thinking they are coming from dotnet-public?

image

donnie-msft:
> don't want to artificially cause you further delays, but this PR at the time I'm doing the review, adds a whole lot of complexity and duplication to nuget.config that I think is unnecessary. As you know, I like to 'do things right', so I'd prefer if you simplified and improved nuget.config, but if you need to merge this now, then I'm giving you the approval to unblock the merge button.

Let's talk about these offline and I'll work to address them separately.  https://github.com/NuGet/Home/issues/13548

For example, taking your suggestion, I get hundreds of NU1603 errors. Just picking one, I can see that the packages are *not* on dotnet-public. So, we might want to work with the publishing teams and make sure packages are going to feeds we expect. Right now, it seems like vside is the most common feed we can expect to find most VS packages.

_(Here's one of the errors when removing the source mapping for vside)_

> error NU1102: Unable to find package Microsoft.VisualStudio.ProjectSystem.Managed.VS with version (>= 17.2.0-beta1-20502-01) [C:\NuGet.Client\build\restorehelper.targets]
> C:\NuGet.Client\src\NuGet.Clients\NuGet.VisualStudio.Implementation\NuGet.VisualStudio.Implementation.csproj : error NU1102:   - Found 1 version(s) in dotnet-public [ Nearest version: 2.0.6142705 ] [C:\NuGet.Client\build\restorehelper.targets]
> C:\NuGet.Client\src\NuGet.Clients\NuGet.VisualStudio.Implementation\NuGet.VisualStudio.Implementation.csproj : error NU1102:   - Versions from dotnet-eng were not considered [C:\NuGet.Client\build\restorehelper.targets]
> C:\NuGet.Client\src\NuGet.Clients\NuGet.VisualStudio.Implementation\NuGet.VisualStudio.Implementation.csproj : error NU1102:   - Versions from dotnet-libraries were not considered [C:\NuGet.Client\build\restorehelper.targets]
> C:\NuGet.Client\src\NuGet.Clients\NuGet.VisualStudio.Implementation\NuGet.VisualStudio.Implementation.csproj : error NU1102:   - Versions from myget-legacy@Local were not considered [C:\NuGet.Client\build\restorehelper.targets]
> C:\NuGet.Client\src\NuGet.Clients\NuGet.VisualStudio.Implementation\NuGet.VisualStudio.Implementation.csproj : error NU1102:   - Versions from nuget-build were not considered [C:\NuGet.Client\build\restorehelper.targets]
> C:\NuGet.Client\src\NuGet.Clients\NuGet.VisualStudio.Implementation\NuGet.VisualStudio.Implementation.csproj : error NU1102:   - Versions from vside were not considered [C:\NuGet.Client\build\restorehelper.targets]

PR #5846, opened 2024-06-09
'dotnet nuget why' works for directory inputs for the 'Path' argument

### LINKED ISSUE #13527: 'dotnet nuget why' does not work when a directory is provided for the 'Path' argument
### NuGet Product Used

dotnet.exe

### Product Version

9.0.1xx-preview.6

### Worked before?

No

### Impact

### Repro Steps & Context

1. Run `dotnet nuget why .\Directory\Path\ Newtonsoft.Json`

It should be able to detect a solution or project file in the directory and run the command on it, following the same behavior as other Dotnet and NuGet CLI commands.

### Verbose Logs

_No response_

advay26:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13527

## Description

`dotnet nuget why` only worked when a solution or project file was provided for the 'Path' argument. It should also work for directory inputs.

This PR includes:

* Enabling directory inputs
    - Search for a solution file or project file in the given directory. Return an error if there were multiple solution or project files in the directory, or if there were zero solution or project files (same behavior as commands like `dotnet list package` and `dotnet add package`). See the `GetListOfProjectsFromPathArgument()` method in `MSBuildAPIUtility.cs`.
* Unit tests and an integration test for the change
* New test logger that I needed for unit testing

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandArgs.cs
```diff
@@ -33,5 +37,191 @@ internal class WhyCommandArgs
...
+
+            Logger.LogError(
+                string.Format(
+                    CultureInfo.CurrentCulture,
+                    Strings.WhyCommand_Error_NoProjectOrSolutionFilesInDirectory,
+                    directory));
+            return null;
+        }
+
+        private static bool IsSolutionFile(string fileName)
```
zivkan: I'd expect methods like this to already exist for `dotnet list package`, and `dotnet add package`.  Did you search or debug those to find out how they do the same outcome?

While `dotnet nuget why`'s use of MSBuild was sufficiently different to `dotnet list package` to warrant not reusing the same method, this logic of choosing solutions over projects, and finding when there's a single project in a directory vs multiple project files, should be identical across all these commands.

advay26: I found these implementations in the `dotnet/sdk` repo. `dotnet add package` ([1](https://github.com/dotnet/sdk/blob/main/src/Cli/dotnet/commands/dotnet-add/dotnet-add-package/Program.cs#L29)  -> [2](https://github.com/dotnet/sdk/blob/main/src/Cli/dotnet/MsbuildProject.cs#L78)) only searches for project files in the current directory. `dotnet list package` ([1](https://github.com/dotnet/sdk/blob/main/src/Cli/dotnet/commands/dotnet-list/dotnet-list-package/ListPackageReferencesCommand.cs#L45) -> [2](https://github.com/dotnet/sdk/blob/main/src/Cli/dotnet/commands/dotnet-list/dotnet-list-package/ListPackageReferencesCommand.cs#L68)) searches for a solution file first, and then a project file (same as what we need).

The behavior here matches what those commands do - Search for solution files and project files. Return an error if there are 0 or 2+ project/solution files. Else, return a single project or solution file.

If we want to reuse code rather than writing new code here, I will need to move some more of the parsing for this command to the `dotnet/sdk` repo and then move some of the above methods (for dotnet list package) to more accessible locations within the repo. Would that work better here?

We can also move the code changes in this PR to a more accessible place in NuGet.Client like the [MSBuildAPIUtility](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs) file so that it can be reused in the future.

advay26: I moved some of these helper methods to XPlatUtility.cs so that they can be reused in the future. I also added unit tests for them.

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/XPlatUtility.cs
```diff
@@ -84,5 +87,86 @@ internal static ISettings ProcessConfigFile(string configFile)
...
+            {
+                return projectFiles[0];
+            }
+
+            throw new ArgumentException(
+                string.Format(
+                    CultureInfo.CurrentCulture,
+                    Strings.Error_NoProjectOrSolutionFilesInDirectory,
+                    directory));
+        }
```
jebriede: This needs formatting and testing (I did not test it). This should achieve the same objective of validating that there is exactly one and only one solution OR project file (not both) and return it, otherwise throw the appropriate error while saving allocations (solution and project file arrays) and having to iterate through the files list twice. Please be sure to test and format the code and fix any bugs (I wrote this in the browser).

```suggestion
    var topLevelFiles = Directory.GetFiles(directory, "*.*", SearchOption.TopDirectoryOnly);
    string candidateFile = null;
    foreach (string file in topLevelFiles)
    {
        if (IsSolutionFile(file) || IsProjectFile(file))
        {
            // We already found a solution or project file and now we found another one.
            if (candidateFile != null)
            {
                throw new ArgumentException(
                        string.Format(
                            CultureInfo.CurrentCulture,
                            Strings.Error_MultipleProjectOrSolutionFilesInDirectory,
                            directory));
            }
        
            candidateFile = file;
        }
    }
    
    // Still being null after visiting all the files means we didn't find a solution or project file.
    if (candidateFile == null)
    {
            throw new ArgumentException(
                string.Format(
                    CultureInfo.CurrentCulture,
                    Strings.Error_NoProjectOrSolutionFilesInDirectory,
                    directory));
    }
    
    return candidateFile;
```

test/NuGet.Core.Tests/NuGet.CommandLine.Xplat.Tests/MSBuildAPIUtilityTests.cs
```diff
@@ -431,5 +433,142 @@ public void UpdateVersionOverrideInPropsFileWhenItExists_Success()
...
+                net8);
+
+            projectA.Save();
+
+            // Act
+            var projectList = MSBuildAPIUtility.GetListOfProjectsFromPathArgument(projectA.ProjectPath);
+
+            // Assert
+            Assert.Equal(projectList.Count(), 1);
+
```
jebriede: Nit: remove extra blank line between assertions in all tests.

test/NuGet.Core.Tests/NuGet.CommandLine.Xplat.Tests/MSBuildAPIUtilityTests.cs
```diff
@@ -431,5 +433,142 @@ public void UpdateVersionOverrideInPropsFileWhenItExists_Success()
...
+            var pathContext = new SimpleTestPathContext();
+
+            var net8 = NuGetFramework.Parse("net8.0");
+
+            var projectA = SimpleTestProjectContext.CreateNETCore(
+                "a",
+                pathContext.SolutionRoot,
+                net8);
+
+            projectA.Save();
```
jebriede: We can remove extra blank lines to save vertical space, especially when each line of code has a blank line between it and the next line of code. This comment applies to all tests.

test/NuGet.Core.Tests/NuGet.CommandLine.Xplat.Tests/MSBuildAPIUtilityTests.cs
```diff
@@ -431,5 +433,142 @@ public void UpdateVersionOverrideInPropsFileWhenItExists_Success()
...
+            var pathContext = new SimpleTestPathContext();
+
+            var solution = new SimpleTestSolutionContext(pathContext.SolutionRoot);
+
+            var net8 = NuGetFramework.Parse("net8.0");
+
+            var projectA = SimpleTestProjectContext.CreateNETCore(
+                "a",
+                pathContext.SolutionRoot,
+                net8);
```
jebriede:
```suggestion
            var projectA = SimpleTestProjectContext.CreateNETCore("a", pathContext.SolutionRoot, net8);
```

test/NuGet.Core.Tests/NuGet.CommandLine.Xplat.Tests/Utility/XPlatTestLogger.cs
```diff
@@ -0,0 +1,117 @@
...
+            }
+        }
+
+        public ConcurrentQueue<string> ErrorMessages
+        {
+            get
+            {
+                return Logger.ErrorMessages;
+            }
+        }
```
jebriede:
```suggestion
        public ConcurrentQueue<string> ErrorMessages => Logger.ErrorMessages;
```

or

```suggestion
        public ConcurrentQueue<string> ErrorMessages
        {
            get => Logger.ErrorMessages;
        }
```

jebriede (commented):
Great work! Left a few suggestions.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -89,6 +99,9 @@ public static int ExecuteCommand(WhyCommandArgs whyCommandArgs)
             return ExitCodes.Success;
         }
 
+        /// <summary>
+        /// Validates that the input 'PATH' argument is a valid path to a directory, solution file or project file.
```
zivkan: nitpick, writing PATH in capital letters immediately makes me think of the PATH environment variable. I'd write this comment as lowercase, to match the casing the parameter actually uses, especially since C# symbols are case sensitive

zivkan (dismissed):

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/XPlatUtility.cs
advay26: Thanks @jebriede! I've changed the implementation now

test/NuGet.Core.Tests/NuGet.CommandLine.Xplat.Tests/Utility/XPlatTestLogger.cs
advay26: This logger was no longer being used after some refactoring that I did, so I've removed the file entirely now

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
advay26: That makes sense. We're also using lowercase 'path' everywhere else, so I've changed this one too

jebriede (approved):

PR #5854, opened 2024-06-14
Enforce IDE0005 during build by enabling GenerateDocumentationFile

jgonz120:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2845

## Description

Updated two projects to include style rules enforcement during build, fixing the build errors for those two projects. Currently certain rules are not enforced unless XML documents are generated, https://github.com/dotnet/roslyn/issues/41640. 
## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/build/common.project.props
+++ b/build/common.project.props
@@ -134,6 +134,12 @@
     <VsixPublishDestination>$(ArtifactRoot)$(VsixOutputDirName)\</VsixPublishDestination>
   </PropertyGroup>
 
+  <!-- Write out .XML files for projects that will be packed. -->
+  <PropertyGroup>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <DocumentationFile Condition=" '$(DocumentationFile)' == '' AND '$(GenerateDocumentationFile)' == 'true' AND '$(IsNetCoreProject)' != 'true' ">$(OutputPath)\$(AssemblyName).xml</DocumentationFile>
+  </PropertyGroup>
+  
   <!-- Set the output location for all non-test projects -->
   <!-- Test projects currently fail when the output dir is moved -->
   <PropertyGroup Condition=" '$(TestProject)' != 'true' OR '$(Shipping)' == 'true'">
--- a/build/common.project.props
+++ b/build/common.project.props
@@ -71,6 +71,12 @@
     <MSBuildTreatWarningsAsErrors Condition=" '$(MSBuildTreatWarningsAsErrors)' == '' And '$(IsOfficialBuild)' != 'true' ">true</MSBuildTreatWarningsAsErrors>
   </PropertyGroup>
 
+  <!-- Write out .XML files for projects that will be packed. -->
+  <PropertyGroup Condition=" '$(PackProject)' == 'true' ">
+    <GenerateDocumentationFile Condition=" '$(GenerateDocumentationFile)' == '' ">true</GenerateDocumentationFile>
+    <DocumentationFile Condition=" '$(DocumentationFile)' == '' AND '$(GenerateDocumentationFile)' == 'true' AND '$(IsNetCoreProject)' != 'true' ">$(OutputPath)\$(AssemblyName).xml</DocumentationFile>
+  </PropertyGroup>
+
   <!-- Default project configuration -->
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
--- a/build/common.project.props
+++ b/build/common.project.props
@@ -71,12 +71,6 @@
     <MSBuildTreatWarningsAsErrors Condition=" '$(MSBuildTreatWarningsAsErrors)' == '' And '$(IsOfficialBuild)' != 'true' ">true</MSBuildTreatWarningsAsErrors>
   </PropertyGroup>
 
-  <!-- Write out .XML files for projects that will be packed. -->
-  <PropertyGroup Condition=" '$(PackProject)' == 'true' ">
-    <GenerateDocumentationFile Condition=" '$(GenerateDocumentationFile)' == '' ">true</GenerateDocumentationFile>
-    <DocumentationFile Condition=" '$(DocumentationFile)' == '' AND '$(GenerateDocumentationFile)' == 'true' AND '$(IsNetCoreProject)' != 'true' ">$(OutputPath)\$(AssemblyName).xml</DocumentationFile>
-  </PropertyGroup>
-
   <!-- Default project configuration -->
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
--- a/build/common.targets
+++ b/build/common.targets
@@ -35,12 +35,6 @@
     <LocTargets>GetBuildOutputWithLocMetadata</LocTargets>
   </PropertyGroup>
 
-  <!-- Write out .XML files for projects that will be packed. -->
-  <PropertyGroup Condition=" '$(PackProject)' == 'true' ">
-    <GenerateDocumentationFile Condition=" '$(GenerateDocumentationFile)' == '' ">true</GenerateDocumentationFile>
-    <DocumentationFile Condition=" '$(DocumentationFile)' == '' AND '$(GenerateDocumentationFile)' == 'true' AND '$(IsNetCoreProject)' != 'true' ">$(OutputPath)\$(AssemblyName).xml</DocumentationFile>
-  </PropertyGroup>
-
   <ImportGroup Condition=" '$(PackProject)' == 'true' AND '$(NuGetBuildTasksPackTargets)' != '' ">
       <Import Project="$(NuGetBuildTasksPackTargets)" />
   </ImportGroup>
--- a/build/common.targets
+++ b/build/common.targets
@@ -8,10 +8,6 @@
     <DefineConstants Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' Or ('$(TargetFrameworkIdentifier)' == '.NETCoreApp' and $([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), 5.0)))">$(DefineConstants);IS_SIGNING_SUPPORTED</DefineConstants>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(EnforceCodeStyleInBuild)' != '' And '$(IsOfficialBuild)' != 'true'">
-    <WarningsNotAsErrors>CS1591</WarningsNotAsErrors>
-  </PropertyGroup>
-
   <!-- Include shared files for netcore projects -->
   <ItemGroup Condition=" ('$(IsNetCoreProject)' == 'true' AND '$(SkipShared)' != 'true' AND '$(TestProject)' != 'true') OR '$(IncludeNuGetSharedFiles)' == 'true'">
     <Compile Include="$(SharedDirectory)\*.cs" Exclude="bin\**;obj\**;**\*.xproj;packages\**" />
@@ -59,7 +55,7 @@
     <None Include="$(MSBuildProjectDirectory)\README.md" Pack="true" PackagePath="\"/>
   </ItemGroup>
 
-    <!-- Don't use PublicApiAnalyzer on source-build .NET -->
+  <!-- Don't use PublicApiAnalyzer on source-build .NET -->
   <PropertyGroup Condition=" '$(DotNetBuildSourceOnly)' == 'true' ">
     <UsePublicApiAnalyzer>false</UsePublicApiAnalyzer>
   </PropertyGroup>
--- a/build/common.targets
+++ b/build/common.targets
@@ -35,6 +35,12 @@
     <LocTargets>GetBuildOutputWithLocMetadata</LocTargets>
   </PropertyGroup>
 
+  <!-- Write out .XML files for projects that will be packed. -->
+  <PropertyGroup Condition=" '$(PackProject)' == 'true' ">
+    <GenerateDocumentationFile Condition=" '$(GenerateDocumentationFile)' == '' ">true</GenerateDocumentationFile>
+    <DocumentationFile Condition=" '$(DocumentationFile)' == '' AND '$(GenerateDocumentationFile)' == 'true' AND '$(IsNetCoreProject)' != 'true' ">$(OutputPath)\$(AssemblyName).xml</DocumentationFile>
+  </PropertyGroup>
+
   <ImportGroup Condition=" '$(PackProject)' == 'true' AND '$(NuGetBuildTasksPackTargets)' != '' ">
       <Import Project="$(NuGetBuildTasksPackTargets)" />
   </ImportGroup>
--- a/build/common.targets
+++ b/build/common.targets
@@ -42,7 +42,7 @@
   </PropertyGroup>
 
   <ImportGroup Condition=" '$(PackProject)' == 'true' AND '$(NuGetBuildTasksPackTargets)' != '' ">
-    <Import Project="$(NuGetBuildTasksPackTargets)" />
+      <Import Project="$(NuGetBuildTasksPackTargets)" />
   </ImportGroup>
 
   <!-- Readme file in shipping packages -->
@@ -55,7 +55,7 @@
     <None Include="$(MSBuildProjectDirectory)\README.md" Pack="true" PackagePath="\"/>
   </ItemGroup>
 
-  <!-- Don't use PublicApiAnalyzer on source-build .NET -->
+    <!-- Don't use PublicApiAnalyzer on source-build .NET -->
   <PropertyGroup Condition=" '$(DotNetBuildSourceOnly)' == 'true' ">
     <UsePublicApiAnalyzer>false</UsePublicApiAnalyzer>
   </PropertyGroup>
@@ -78,7 +78,7 @@
   <ImportGroup Condition=" '$(TestProject)' == 'true' ">
     <Import Project="test.targets" />
   </ImportGroup>
-
+  
 
   <!-- Allow WPF projects to run under NETCore SDK -->
   <!-- Errors occur if the output path is not set correctly -->
@@ -150,7 +150,7 @@
     ============================================================
   -->
   <Target Name="GetSymbolsToIndex" DependsOnTargets="$(GetSymbolsToIndexDependsOn)" Returns="@(SymbolFilesToIndex)" />
-
+  
   <Target Name="GetSymbolsToIndexDefault" DependsOnTargets="GetTargetFrameworkSet" Returns="@(SymbolFilesToIndex)" Condition=" '$(Shipping)' == 'true' ">
     <MSBuild
       Projects="$(MSBuildProjectFullPath)"
@@ -172,8 +172,8 @@
     ============================================================
   -->
   <Target Name="GetSigningInputs" DependsOnTargets="$(GetSigningInputsDependsOn)" Returns="@(DllsToSign)" />
-
-  <!--
+ 
+ <!--
     ============================================================
     The default target to get signing inputs for a particular project
     ============================================================
@@ -258,8 +258,7 @@
       <!-- remove all net45 assemblies, otherwise loc tools have problem loading two assemblies of same name in same appdomain. net45 doesn't ship in VSIX, so doesn't need loc. -->
       <DllsToLocalizeWithMetadata Include="@(BuiltProjectOutputGroupOutput->'%(FinalOutputPath)')"
 Condition=" ('%(Extension)' == '.dll' OR '%(Filename)' == 'NuGet.CommandLine.XPlat' OR '%(Filename)' == 'NuGet') AND !($([System.String]::new('%(FullPath)').Contains('\net45\'))) " KeepDuplicates="false">
-        <TranslationFile>$(LocalizationWorkDirectory)\{Lang}\15\%(Filename)%(Extension).lcl</TranslationFile>
-        <!--Required: translation file-->
+        <TranslationFile>$(LocalizationWorkDirectory)\{Lang}\15\%(Filename)%(Extension).lcl</TranslationFile>    <!--Required: translation file-->
         <LciCommentFile>$(LocalizationWorkDirectory)\comments\15\%(Filename)%(Extension).lci</LciCommentFile>
         <HasLceComments>false</HasLceComments>
       </DllsToLocalizeWithMetadata>
@@ -378,7 +377,7 @@ Condition=" ('%(Extension)' == '.dll' OR '%(Filename)' == 'NuGet.CommandLine.XPl
   </Target>
 
   <Target Name="GetAllTargetFrameworks">
-    <PropertyGroup>
+  <PropertyGroup>
       <EffectiveFramework Condition=" '$(TargetFrameworks)' != '' ">$(TargetFrameworks)</EffectiveFramework>
       <EffectiveFramework Condition=" '$(EffectiveFramework)' == '' ">$(TargetFramework)</EffectiveFramework>
     </PropertyGroup>
--- a/build/common.targets
+++ b/build/common.targets
@@ -8,6 +8,10 @@
     <DefineConstants Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' Or ('$(TargetFrameworkIdentifier)' == '.NETCoreApp' and $([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), 5.0)))">$(DefineConstants);IS_SIGNING_SUPPORTED</DefineConstants>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(EnforceCodeStyleInBuild)' != '' And '$(IsOfficialBuild)' != 'true'">
+    <WarningsNotAsErrors>CS1591</WarningsNotAsErrors>
+  </PropertyGroup>
+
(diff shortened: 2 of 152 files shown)


src/NuGet.Clients/NuGet.CommandLine/CommandLineMachineWideSettings.cs
```diff
@@ -8,17 +8,23 @@
 namespace NuGet.CommandLine
 {
     [Export(typeof(Configuration.IMachineWideSettings))]
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
```
jgonz120: @zivkan @nkolev92 continuing the conversation here since GitHub wouldn't let me reopen the PR. 

Since this project is shipped on nuget.org, we should keep the warning right?

src/NuGet.Core/NuGet.Versioning/NuGet.Versioning.csproj
```diff
@@ -7,6 +7,8 @@
     <PackProject>true</PackProject>
     <Shipping>true</Shipping>
     <IncludeInVSIX>true</IncludeInVSIX>
+    <GenerateDocumentationFile Condition="'$(IsOfficialBuild)' == 'true'">false</GenerateDocumentationFile>
```
jgonz120: @nkolev92 So from the previous PR, are you suggesting we enable document generation for all of them?

build/common.targets
```diff
@@ -258,7 +258,8 @@
       <!-- remove all net45 assemblies, otherwise loc tools have problem loading two assemblies of same name in same appdomain. net45 doesn't ship in VSIX, so doesn't need loc. -->
       <DllsToLocalizeWithMetadata Include="@(BuiltProjectOutputGroupOutput->'%(FinalOutputPath)')"
 Condition=" ('%(Extension)' == '.dll' OR '%(Filename)' == 'NuGet.CommandLine.XPlat' OR '%(Filename)' == 'NuGet') AND !($([System.String]::new('%(FullPath)').Contains('\net45\'))) " KeepDuplicates="false">
-        <TranslationFile>$(LocalizationWorkDirectory)\{Lang}\15\%(Filename)%(Extension).lcl</TranslationFile>    <!--Required: translation file-->
+        <TranslationFile>$(LocalizationWorkDirectory)\{Lang}\15\%(Filename)%(Extension).lcl</TranslationFile>
+        <!--Required: translation file-->
```
zivkan: I interpret the comment before this PR to apply to the line for `lcl` files (`<TranslationFile>`), whereas after your change to apply to the `lci` files (`LciCommentFiles>`)

If you want the comment on a separate line, I'd put it before the `<TranslationFile>` element

src/NuGet.Clients/NuGet.CommandLine/CommandLineMachineWideSettings.cs
zivkan: The NuGet.CommandLine package is a transport vehicle for nuget.exe, which is used by `nuget.exe update -self`.

It's not a library that customers can reference in their project and call NuGet APIs on. Therefore, XML comments only benefit people working within this repo, but will no benefit anyone else.

src/NuGet.Core/NuGet.Versioning/NuGet.Versioning.csproj
zivkan: Official builds create "real signed" binaries after a PR is merged. By disabling GenerateDocumentationFile for non-official builds, it means that pull requests can have green builds that break official builds when merged.  This absolutely should not be allowed.

src/NuGet.Core/NuGet.Versioning/VersionComparer.cs
```diff
@@ -12,7 +12,9 @@ namespace NuGet.Versioning
     /// </summary>
     public sealed class VersionComparer : IVersionComparer
     {
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
         public static IVersionComparer Get(VersionComparison versionComparison)
+#pragma warning restore CS1591 // Missing XML comment for publicly visible type or member
```
zivkan: I think this one is sufficiently simple to document, just from guessing based on symbol names. Using the IDE to see the definition of the VersionComparison type, I think you'll quickly have a good understanding of what this API does, or just read this method's body below.

This PR, as it currently stands, has very few suppressions outside of NuGet.CommandLine, so I'd like to see the analyzer warnings being fixed, rather than suppressed.

src/NuGet.Core/NuGet.Versioning/VersionRangeComparer.cs
```diff
@@ -56,7 +56,9 @@ public VersionRangeComparer(IVersionComparer versionComparer)
 
         internal static IVersionRangeComparer VersionReleaseMetadata { get; } = new VersionRangeComparer(VersionComparer.VersionReleaseMetadata);
 
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
         public static IVersionRangeComparer Get(VersionComparison versionComparison)
+#pragma warning restore CS1591 // Missing XML comment for publicly visible type or member
```
zivkan: Same as the other Nuget.Versioning file, this is easy to document, which I'd rather see than kicking the can down the line.

zivkan (commented):
The PR checklist is not filled out, and personally I find the PR title misleading, as it's just enabling XMLDoc generation for two projects, but it's not enabling style rules more broadly (style rules other than XMLDoc comments to the two projects, and it's not clear to me why just these two projects, rather than most `src/NuGet.Core/*` projects)

src/NuGet.Clients/NuGet.CommandLine/CommandLineMachineWideSettings.cs
jgonz120: So this is not the same as this package? https://www.nuget.org/packages/NuGet.CommandLine

build/common.targets
jgonz120: I think all I did was run the default formatter for the file. Let me just undo it since I have no clue what the context was for the comment.

src/NuGet.Core/NuGet.Versioning/NuGet.Versioning.csproj
jgonz120: It should be the opposite; it's disabling document generation during official builds right now. Enabling them otherwise.

src/NuGet.Clients/NuGet.CommandLine/CommandLineMachineWideSettings.cs
zivkan: it is the same package. My previous comment still applies 😄

src/NuGet.Core/NuGet.Versioning/NuGet.Versioning.csproj
zivkan: Yes, you're right. But doesn't setting `GenerateDocumentFile` to false for official builds mean our nupkgs that we publish to nuget.org won't have the xmldoc, so people using our NuGet.Versioning package won't get these xmldoc intellisense?

jgonz120: Oof yes! I missed the "PackProject" element. Gonna just set it to always generate the document file.

zivkan (dismissed):

src/NuGet.Clients/NuGet.CommandLine/CommandLineMachineWideSettings.cs
nkolev92: Should we disable the xml comments for NuGet.Commandline since it's not really shipping API?

jgonz120: I added a new .editorconfig to the project and set this warning as a suggestion. When I added the .editorconfig file it made a couple of changes to the csproj so I removed them... 

Not sure why it was adding a `<EditorConfigFiles Remove=".editorconfig"/>` by default O.o

Do I need the `<None Include=".editorconfig" />` though? I removed it and everything seems fine 🤷

nkolev92: Personally, idk. I didn't know editor configs could nest. 

Is there a way to change that through msbuild potentially?

jgonz120: oo duh, added a nowarn, much less confusion :)

src/NuGet.Clients/NuGet.CommandLine/NuGet.CommandLine.csproj
```diff
@@ -24,7 +24,8 @@
     <UsePublicApiAnalyzer>false</UsePublicApiAnalyzer>
     <GetSigningInputsDependsOn>GetSigningInputsCustom</GetSigningInputsDependsOn>
     <GetSymbolsToIndexDependsOn>GetSymbolsToIndexCustom</GetSymbolsToIndexDependsOn>
-</PropertyGroup>
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
```
nkolev92: Do we still want to keep the cs changes?

tbh, I don't see too much value in documenting NuGet.Commandline in general, so I don't care about the supressions personally, but I just wanted to make sure things are consistent (either suppress locally, or completely suppress, currently we have both0

src/NuGet.Core/NuGet.Versioning/NuGet.Versioning.csproj
```diff
@@ -7,6 +7,7 @@
     <PackProject>true</PackProject>
     <Shipping>true</Shipping>
     <IncludeInVSIX>true</IncludeInVSIX>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
zivkan, line 10: The linked issue isn't clearly defined, making it sound like the intention is to enforce all style rules, not just documentation files, and also across all projects, not just one or two. The PR description also doesn't explain either why only XMLDoc, or why only 1 or 2 projects.

Most of the `src/NuGet.Core/*` projects are nupkgs published to nuget.org.  Do the other project already generate XMLDoc, and NuGet.Versioning was the only one that wasn't?

Should the linked issue be changed to explicitly mention it's only about XMLDoc, and not other style issues?

src/NuGet.Clients/NuGet.CommandLine/NuGet.CommandLine.csproj
jgonz120: Yea that makes sense, If we don't want to enforce xml comments for this project I think doing a nowarn makes more sense than suppressing it locally.

src/NuGet.Core/NuGet.Versioning/NuGet.Versioning.csproj
nkolev92: I agree with Andy that we should clarify the scope of the PR/issue.

jgonz120: Okay I updated the issue. I don't really have a reason why I picked these two projects, I just needed to start somewhere. 

One thing though, a know that <GenerateDocumentationFile> is set to true for projects with <PackProject> set to true https://github.com/NuGet/NuGet.Client/blob/15ad272450c71cdc32b8877a90ca5e6a3cb5c8e3/build/common.targets#L39-L42

Despite that I still don't get build errors for using statements until I specify GenerateDocumentationFile in the csproj.

nkolev92: can you upload a binlog for the project in question?
You can do it in teams if that's easier.

jgonz120: Okay I moved this pack from the common.targets to common.project.props, and that worked. Is there a better location for it? Basically, we need to set it in a different location because it's getting set by another config.

nkolev92: You said common.project.props, but I don't see that change in this PR.

nkolev92 (dismissed):

src/NuGet.Core/NuGet.Versioning/NuGet.Versioning.csproj
jgonz120: Okay that change didn't actually work but now I think I properly understand what's happening.

The <GenerateDocumentationFile> property is being set in Microsoft.NET.Sdk.BeforeCommon.targets 

```xml
  <!-- Handle XML documentation file settings -->
  <PropertyGroup Condition="'$(GenerateDocumentationFile)' == ''" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
    <GenerateDocumentationFile Condition="'$(DocumentationFile)' == ''">false</GenerateDocumentationFile>
    <GenerateDocumentationFile Condition="'$(DocumentationFile)' != ''">true</GenerateDocumentationFile>
  </PropertyGroup>
  <PropertyGroup Condition="'$(GenerateDocumentationFile)' == 'true' and '$(DocumentationFile)' == ''" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
    <DocumentationFile Condition="'$(MSBuildProjectExtension)' == '.vbproj'">$(AssemblyName).xml</DocumentationFile>
    <DocumentationFile Condition="'$(MSBuildProjectExtension)' != '.vbproj'">$(IntermediateOutputPath)$(AssemblyName).xml</DocumentationFile>
  </PropertyGroup>
  <PropertyGroup Condition="'$(GenerateDocumentationFile)' != 'true'" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
    <DocumentationFile />
  </PropertyGroup>
```

Looking at the msbuild generated properties file the next file up the tree we control is the individual csproj, which is where we set the <PackProject> property.

Moving the code to the common.project.props file doesn't work because when it goes to evaluate the condition <PackProject> is empty, so <GenerateDocumentationFile> is set to false in Microsoft.NET.Sdk.BeforeCommon.targets. 

Even if we set the <GenerateDocumentationFile> value in the csproj the value for<DocumentationFile> will be set by Microsoft.NET.Sdk.BeforeCommon.targets instead of what we have in common.targets
```xml
<PropertyGroup Condition=" '$(PackProject)' == 'true' ">
    <GenerateDocumentationFile Condition=" '$(GenerateDocumentationFile)' == '' ">true</GenerateDocumentationFile>
    <DocumentationFile Condition=" '$(DocumentationFile)' == '' AND '$(GenerateDocumentationFile)' == 'true' AND '$(IsNetCoreProject)' != 'true' ">$(OutputPath)\$(AssemblyName).xml</DocumentationFile>
</PropertyGroup>
```

Is there a reason to put the xml in the intermediate output path instead of the output path? If not then I think we could delete the block above from common.targets and set <GenerateDocumentationFile> for all projects we are also packing and currently don't have it set to false.

nkolev92: > Moving the code to the common.project.props file doesn't work because when it goes to evaluate the condition is empty, so is set to false in Microsoft.NET.Sdk.BeforeCommon.targets.

Which condition is false?

nkolev92: > Is there a reason to put the xml in the intermediate output path instead of the output path? If not then I think we could delete the block above from common.targets and set for all projects we are also packing and currently don't have it set to false.

No clue. I'm guessing it's missing defaults based on the is net core check.

zivkan: Keep in mind that the NuGet.Client repo has existed for years, before the .NET (Core) SDK was a thing, so while the SDK might now have defaults to make it work out of the box, perhaps in the past it didn't, or pre-SDK style projects might have worked differently as well.

I prefer a common config to apply to all projects, rather than opting individual projects into docs.

build/common.project.props
```diff
@@ -71,6 +71,12 @@
     <MSBuildTreatWarningsAsErrors Condition=" '$(MSBuildTreatWarningsAsErrors)' == '' And '$(IsOfficialBuild)' != 'true' ">true</MSBuildTreatWarningsAsErrors>
   </PropertyGroup>
 
+  <!-- Write out .XML files for projects that will be packed. -->
+  <PropertyGroup Condition=" '$(PackProject)' == 'true' ">
+    <GenerateDocumentationFile Condition=" '$(GenerateDocumentationFile)' == '' ">true</GenerateDocumentationFile>
+    <DocumentationFile Condition=" '$(DocumentationFile)' == '' AND '$(GenerateDocumentationFile)' == 'true' AND '$(IsNetCoreProject)' != 'true' ">$(OutputPath)\$(AssemblyName).xml</DocumentationFile>
+  </PropertyGroup>
```
zivkan: As you wrote in another comment, the condition on the group will always evaluate to false, because common props get evaluated BEFORE the project's own properties, which is why this was previously in the targets file, since common targets get evaluated after the project.

Therefore, this code here is dead code, and should be removed.  If the repo works as expected (projects are generating xml doc files, and we're getting warnings about missing public methods without docs), then we don't need the config at all.

src/NuGet.Core/NuGet.Versioning/NuGet.Versioning.csproj
jgonz120: > > Moving the code to the common.project.props file doesn't work because when it goes to evaluate the condition is empty, so is set to false in Microsoft.NET.Sdk.BeforeCommon.targets.
> 
> Which condition is false?

This one, since PackProject is set later in the file it has no value when this is evaluated. 
```xml
<PropertyGroup Condition=" '$(PackProject)' == 'true' ">
```

jgonz120: > I prefer a common config to apply to all projects, rather than opting individual projects into docs.

Agreed but where that common config should live is difficult because of how `<GenerateDocumentationFile>` is set by `Microsoft.NET.Sdk.BeforeCommon.targets`. 

Since we now know that without `<GenerateDocumentationFile>` we can't enforce all the style rules, is it worth just enabling it for everyone?

nkolev92: I think it's ok to enable it for everyone. 
We'd probably need to disable the IDE0005 in some projects where it fails,.
I can't think of anything else.

src/NuGet.Clients/NuGet.CommandLine/NuGet.CommandLine.csproj
```diff
@@ -24,7 +24,8 @@
     <UsePublicApiAnalyzer>false</UsePublicApiAnalyzer>
     <GetSigningInputsDependsOn>GetSigningInputsCustom</GetSigningInputsDependsOn>
     <GetSymbolsToIndexDependsOn>GetSymbolsToIndexCustom</GetSymbolsToIndexDependsOn>
-</PropertyGroup>
+    <NoWarn>$(NoWarn);CS1591;IDE0059;IDE1006;IDE0049</NoWarn>
```
zivkan, line 27: CS1591 is related to XMLDoc, and therefore the title of this PR, but the other 3 codes are not.

We want to keep IDE1006, so disabling for the whole project feels wrong. I know we've got a bunch of noise on static properties, but we want to keep warnings (ideally enforce) public properties being Pascal Case, private members being _camelCase, methods being Pascal Case.

An in my opinion, IDE0059 seems like a good analyzer we should keep.

I think IDE0049 is very reasonable too, though we've got a lot of violations in our repo. But suppressing the code increases risk that future code changes will continue to add more violations, making it harder to enforce in the future.

test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/NuGet.Protocol.FuncTest.csproj
```diff
@@ -44,4 +45,4 @@
               ItemName="ReferencesCopiedInThisBuild" />
     </Copy>
   </Target>
-</Project>
+</Project>
```
zivkan, line 48: let's keep trailing new line. I thought our editorconfig actually specifies this

build/common.project.props
```diff
@@ -134,6 +134,12 @@
     <VsixPublishDestination>$(ArtifactRoot)$(VsixOutputDirName)\</VsixPublishDestination>
   </PropertyGroup>
 
+  <!-- Write out .XML files for projects that will be packed. -->
+  <PropertyGroup>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <DocumentationFile Condition=" '$(DocumentationFile)' == '' AND '$(GenerateDocumentationFile)' == 'true' AND '$(IsNetCoreProject)' != 'true' ">$(OutputPath)\$(AssemblyName).xml</DocumentationFile>
```
zivkan, line 140: I think it's a great thing to enforce XMLDoc for the packages we publish to nuget.org. However,

1. This issue is titled enforce IDE0005, but [the docs say IDE0005 is about removing unnecessary usings](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0005?pivots=lang-csharp-vb), which has nothing to do with XMLDoc, and therefore the title "enforce IDE005 by enabling GenerateDocumentationFile" doesn't make sense
2. I don't see anything in this PR that changes enforcement of IDE005. Our CI builds will continue to pass after this merges if people add new violations in future PRs.
3. To be honest, I'm not a fan of adding NoWarn. I see NoWarn as tech debt, so now that I see the consequence of enabling `GenerateDocumentationFile` to all projects in the solution, I prefer the old implementation where it's only selectively enabled for projects that are packed.

src/NuGet.Clients/NuGet.Console/NuGet.Console.csproj
```diff
@@ -9,6 +9,7 @@
     <TargetFramework>$(NETFXTargetFramework)</TargetFramework>
     <UseWpf>true</UseWpf>
     <ExcludeFromSourceOnlyBuild>true</ExcludeFromSourceOnlyBuild>
+    <NoWarn>$(NoWarn);CS1572;CS1573;CS1591;IDE0005;IDE1006</NoWarn>
```
zivkan, line 12: The PR title says it's enforcing IDE0005, but it's just being suppressed here, not enforced.

test/NuGet.Tests.Apex/NuGet.OptProf/NuGet.OptProf.csproj
```diff
@@ -24,11 +25,7 @@
   
   <ItemGroup>
     <!-- Manually consuming compile-time assets from Microsoft.Test.Apex.VisualStudio and excluding ones that cause compile warnings -->
-    <Reference Include="$(PkgMicrosoft_Test_Apex_VisualStudio)\lib\net46\*.dll"
-               Exclude="$(PkgMicrosoft_Test_Apex_VisualStudio)\lib\net46\Microsoft.Test.Apex.PrismIntegration.dll;
-                        $(PkgMicrosoft_Test_Apex_VisualStudio)\lib\net46\Microsoft.Test.Apex.RemoteCodeInjector.dll;
-                        $(PkgMicrosoft_Test_Apex_VisualStudio)\lib\net46\Microsoft.Test.Apex.RemoteCodeInjector.x64.dll"
-               Name="%(Filename)" />
+    <Reference Include="$(PkgMicrosoft_Test_Apex_VisualStudio)\lib\net46\*.dll" Exclude="$(PkgMicrosoft_Test_Apex_VisualStudio)\lib\net46\Microsoft.Test.Apex.PrismIntegration.dll;&#xD;&#xA;                        $(PkgMicrosoft_Test_Apex_VisualStudio)\lib\net46\Microsoft.Test.Apex.RemoteCodeInjector.dll;&#xD;&#xA;                        $(PkgMicrosoft_Test_Apex_VisualStudio)\lib\net46\Microsoft.Test.Apex.RemoteCodeInjector.x64.dll" Name="%(Filename)" />
```
zivkan, line 28: This looks like a bad change.

src/NuGet.Core/NuGet.Versioning/NuGet.Versioning.csproj
jgonz120: @zivkan to focus this conversation here because this has the most context. So, the latest change I pushed is what it looks like if we enable `GenerateDocumentationFile` for all projects all at once. I kept flipping back and forth between enabling it per project then once all projects have had it enabled moving it to the central build props project. 

Currently we do not generate the documentation file for any of our projects because of the way the value is determined. So, I believe options for moving forward are:
1. Enable `GenerateDocumentationFile` for all projects, using `NoWarn` to suppress build errors, then overtime remove the suppressions per project and fix the errors that come up. (current state)
2.  Enable `GenerateDocumentationFile` for all projects, using `WarningsNotAsErrors` to suppress build errors, then overtime remove the suppressions per project and fix the errors that come up.
3. Enable `GenerateDocumentationFile` for individual projects, fixing the build errors as it is enabled. Once all projects have been updated to generate documentation files, we can move it to a central location. 
4. Force `GenerateDocumentationFile` to true for projects which have `PackProject` set to true. This will limit the scope of how many projects need to be fixed upfront, however it will still be a lot. We would need to selectively turn on `GenerateDocumentationFile` for all projects in order to get the style fixes like IDE0005 to show up as errors in the build.

PR #5857, opened 2024-06-17
Reduce memory allocations during solution load in VS

### LINKED ISSUE #13558: PERF: LockFileFormat is filled completely when common callers only need some of the data
### NuGet Product Used

Other/NA

### Product Version

VS17

### Worked before?

N/A

### Impact

Other

### Repro Steps & Context

See https://github.com/NuGet/NuGet.Client/pull/5857 for more details

Open VS
Wait for things to calm down
Start profiling
Open Roslyn.sln, open a file it none opened
Wait for things to calm down
Stop profiling

### Verbose Logs

_No response_

ToddGrun:
## Bug
-- 

Fixes: https://github.com/NuGet/Home/issues/13558
-- Only allocates information in LockFileFormat that is needed by the caller

-- N/A

## Description

-- Introduces LockFileReadFlags that is optionally passed into LockFileUtilities.GetLockFile. This flag controls the data which is populated by that method. This can have a meaningful memory reduction as the common caller during solution load doesn't need the majority of the information.
 
## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception
-- Will to add automation if desired. Should be no functional changes if called with defaults.
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


ToddGrun:
Repro: 
1) Open VS
2) Wait for things to calm down
3) Start profiling
4) Open Roslyn.sln, open a file it none opened
5) Wait for things to calm down
6) Stop profiling

Allocations during LockFileUtilities.GetLockFile:

*** Previous ***
image

*** With changes ***
image

src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileFormat.cs
```diff
@@ -207,21 +218,48 @@ private LockFile Utf8JsonRead(Stream stream, ILogger log, string path)
         }
 
         [Obsolete]
-        private static LockFile ReadLockFile(JObject cursor, string path)
+        private static LockFile ReadLockFile(JObject cursor, string path, LockFileReadFlags flags)
```
kartheekp-ms, line 221: This method is marked as `Obsolete`. I think it is better not to modify obsolete methods because when we plan on deleting them in the next .NET major version, it will create more work in the future.

**EDIT**: It looks like currently we don't have any alternative because the product code is also invoking Obsolete methods [here](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileFormat.cs#L104-L106) enclosed in a pragma block to suppress the compiler warning.

kartheekp-ms:
Thanks for your contribution. Can we add some tests for these changes in https://github.com/NuGet/NuGet.Client/blob/dev/test/NuGet.Core.Tests/NuGet.ProjectModel.Test/LockFileFormatTests.cs to begin with?

ToddGrun:
> Can we add some tests for these changes in https://github.com/NuGet/NuGet.Client/blob/dev/test/NuGet.Core.Tests/NuGet.ProjectModel.Test/LockFileFormatTests.cs to begin with?

Tests added. Was a bit non-standard in it's use of TestEnvironmentVariableReader, but I didn't see an easy way to do it with the InlineData I wanted otherwise.

src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileFormat.cs
nkolev92: This is the fallback for reading the assets file with newtonsoft.json, so until we're ready to remove that code, we need to maintain this codepath as well.

ToddGrun: Just to make sure I'm parsing these comments correctly, all is good here, right?

kartheekp-ms (dismissed):

kartheekp-ms:
> > Can we add some tests for these changes in https://github.com/NuGet/NuGet.Client/blob/dev/test/NuGet.Core.Tests/NuGet.ProjectModel.Test/LockFileFormatTests.cs to begin with?
> 
> Tests added. Was a bit non-standard in it's use of TestEnvironmentVariableReader, but I didn't see an easy way to do it with the InlineData I wanted otherwise.

Thanks for adding the tests. I think it was added in this way to ensure tests were covering Newtonsoft.Json and System.Text.Json code paths. Tests added in this PR are only covering Newtonsoft.Json code path. 

https://github.com/NuGet/NuGet.Client/blob/dev/test/NuGet.Core.Tests/NuGet.ProjectModel.Test/LockFileParsingEnvironmentVariable.cs#L31-L59

I guess we could add one more test similar to what is added in this PR to test System.Text.Json code path also. This can be achived by changing the following line in the new test from 

```cs
var environmentVariableReader = LockFileParsingEnvironmentVariable.TestEnvironmentVariableReader().First()[0] as IEnvironmentVariableReader;
// Returns "NUGET_EXPERIMENTAL_USE_NJ_FOR_FILE_PARSING: true" which means Newtonsoft.Json code path is used to parse the file
```

to 

```cs
var environmentVariableReader = LockFileParsingEnvironmentVariable.TestEnvironmentVariableReader().Skip(1).First()[0] as IEnvironmentVariableReader;
// Returns "NUGET_EXPERIMENTAL_USE_NJ_FOR_FILE_PARSING: false" which means System.Text.Json code path is used to parse the file
```

ToddGrun:
> This can be achived by changing the following line in the new test from

OK, how does the test change I made look? I didn't want to duplicate the whole test again just for that purpose.

kartheekp-ms (approved):

ToddGrun:
@kartheekp-ms -- I don't have merge rights, if you think this is ready, can you or someone from the nuget team push the button for me? Thanks!

PR #5858, opened 2024-06-17
deprecate non https sources: Promote warning to Error

### LINKED ISSUE #13289: [Epic]Deprecate http usage: Promote from warning to error
### NuGet Product(s) Affected

### Current Behavior

We have [NU1803](https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1803) warnings across NuGet. Promote from warning to error.

- [x] [Promote from warning to error in restore scenarios](https://github.com/NuGet/Home/issues/13353)
- [x] [Promote from warning to error in push operations](https://github.com/NuGet/Home/issues/13342)
- [x] [Promote from warning to error in delete operations](https://github.com/NuGet/Home/issues/13341)
- [x] [Promote from warning to error in search package](https://github.com/NuGet/Home/issues/13316)
- [x] [Promote from warning to error in list package](https://github.com/NuGet/Home/issues/13323)
- [x] [Promote from warning to error in source command](https://github.com/NuGet/Home/issues/13333)
- [x] [Promote from warning to error in VS Options NuGet Package Manager](https://github.com/NuGet/Home/issues/13354)
- [ ] [Show an error when a non https source is used in a resource in a service index](https://github.com/NuGet/Home/issues/13364)
- [x] [Add `--allow-insecure-connections` option to dotnet.exe](https://github.com/NuGet/Home/issues/13396)

Nigusu-Allehu:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13289

## Description

- NuGet has warnings for http sources and this PR promotes these warnings to errors.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A


src/NuGet.Core/NuGet.Commands/NuGet.Commands.csproj
```diff
@@ -64,5 +64,6 @@
     <InternalsVisibleTo Include="NuGet.ProjectModel.Test" />
     <InternalsVisibleTo Include="NuGet.SolutionRestoreManager.Test" />
     <InternalsVisibleTo Include="Test.Utility" />
+    <InternalsVisibleTo Include="Dotnet.Integration.Test" />
```
zivkan, line 67: dotnet.integration.test writes project files to disk, executes the dotnet CLI as a child process, then reads files off disk to validate.  There's no reason why it should need access to internal classes or methods.

src/NuGet.Core/NuGet.Common/Errors/NuGetLogCode.cs
```diff
@@ -222,6 +222,11 @@ public enum NuGetLogCode
         /// </summary>
         NU1301 = 1301,
 
+        /// <summary>
+        /// HTTP Source specified.
```
zivkan:
```suggestion
        /// Insecure HTTP Source specified.
```

Without context of this PR, just browsing the list of error codes, it sounds normal to specify HTTP sources, so it's not obvious why this would be a warning or error.

src/NuGet.Core/NuGet.Commands/NuGet.Commands.csproj
Nigusu-Allehu: That is for testing purposes. To allow access to the resex file which is internal

src/NuGet.Core/NuGet.PackageManagement/NuGet.PackageManagement.csproj
```diff
@@ -55,5 +55,8 @@
     <InternalsVisibleTo Include="DynamicProxyGenAssembly2" />
     <InternalsVisibleTo Include="NuGet.PackageManagement.Test" />
     <InternalsVisibleTo Include="NuGet.SolutionRestoreManager.Test" />
+    <InternalsVisibleTo Include="NuGet.CommandLine.FuncTest" />
+    <InternalsVisibleTo Include="Msbuild.Integration.Test" />
+    <InternalsVisibleTo Include="NuGet.Commands.FuncTest" />
```
zivkan, line 60: Needing `InternalsVisibleTo` for a test proejct other than the test project of the same name as the csproj being modified, is a significant sign of possible tech debt.

Having a quick look over the tests from these three test projects, it's not clear to me why any of those test projects need access to NuGet.PackageManagement internals, especially since this PR/feature didn't add any public APIs to the NuGet.PackageManagement assembly.

test/NuGet.Clients.FuncTests/NuGet.CommandLine.FuncTest/Caching/CachingTestContext.cs
```diff
@@ -252,18 +252,18 @@ private void InitializeFiles()
...
-        private void CreateNuGetConfig(string workingDirectory)
+        internal void CreateNuGetConfig(string workingDirectory, string source)
         {
             string nugetConfigContent =
-                @"<?xml version='1.0' encoding='utf-8'?>
+                $@"<?xml version='1.0' encoding='utf-8'?>
 <configuration>
   <packageSources>
     <clear />
+    <add key=""http-feed"" value=""{source}"" allowInsecureConnections=""true""/>
```
zivkan, line 266: I really wish someone caught this in the PR that made this change in the first place.

This change is modifying the design of how these "test contexts" work. By default they're supposed to create an isolated environment for the test to run in. Since the test directory is in a `.test` subdirectory of the NuGet.Client repo, the test directory needs a nuget.config with `<clear/>` elements to avoid getting the NuGet.Client repo nuget.config package sources.

By removing the call to `CreateNuGetConfig` from the class constructor, it's now mandatory for every test using this class to now also call `CreateNuGetConfig` to meet the original design goals (and eliminate risks of downloading packages from public feeds that have a name clash with packages used by tests, and eliminate risks of writing a new push test that accidentally pushes to a real package source).  This violates the principal of least surprise, especially since `SimpleTestConext` has a similar name and before this PR had similar semantics.

src/NuGet.Core/NuGet.Commands/NuGet.Commands.csproj
nkolev92: Can we assert the error code only in those tests? 

That can be done in a follow-up ofc.

test/NuGet.Clients.FuncTests/NuGet.CommandLine.FuncTest/Caching/CachingTestContext.cs
nkolev92: This class isn't using SimpleTestPathContext, which is probably one problem.

I think moving *all* of the config logic into the CachingTestContext is the right thing too: 

https://github.com/NuGet/NuGet.Client/pull/5858/files?file-filters%5B%5D=.cs&file-filters%5B%5D=.csproj&show-viewed-files=true#diff-2110e5d1ccc026550e4eb2aa5c587597d79f23161b6a79c565e01c17879692f4R26-R31

already had a bunch of logic that probably should've been part of the constructor anyways. 

Any setters in that class should probably be removed and just passed in the constructor.

src/NuGet.Core/NuGet.Commands/NuGet.Commands.csproj
nkolev92: After writing this, I see that we have sources tests which don't really have error codes.

test/TestUtilities/Test.Utility/SimpleTestSetup/SimpleTestSettingsContext.cs
```diff
@@ -164,6 +164,16 @@ public static void AddEntry(XElement section, string key, string value, string a
             section.Add(setting);
         }
 
+        public static void AddEntry(XElement section, string key, string value, string additionalAtrributeName, string additionalAttributeValue, string additionalAtrributeName2, string additionalAttributeValue2)
```
nkolev92, line 167: In retrospect this probably could've been a params of KVPs.

nkolev92 (dismissed):
My feedback is things that can be addressed later.

nkolev92 (approved):

PR #5862, opened 2024-06-18
Reduce allocations for version / versionranges

### LINKED ISSUE #13559: PERF: Version and VersionRange allocations are very prevalent in profiles of Roslyn solution load
### NuGet Product Used

Other/NA

### Product Version

VS17

### Worked before?

N/A

### Impact

Other

### Repro Steps & Context

See https://github.com/NuGet/NuGet.Client/pull/5862 for more details

Open VS
Wait for things to calm down
Start profiling
Open Roslyn.sln, open a file it none opened
Wait for things to calm down
Stop profiling

### Verbose Logs

_No response_

ToddGrun:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13559

No

## Description

Adds caching to both the `VersionRange` and `NuGetVersion` classes. These types (and the supporting types they create for their construction) were showing up in profiles take of Roslyn solution load.  This implementation keeps a dictionary strings and returns already parsed objects so that the same string is not parsed over and over.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception
-- Changes are trivial enough that tests wouldn't be useful
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


ToddGrun:
Repro:

Open VS
Wait for things to calm down
Start profiling
Open Roslyn.sln, open a file if none opened
Wait for things to calm down
Stop profiling

This reduces allocations by about 3% of Roslyn solution load (or about 1.5% if one of my earlier PRs is also in the payload)

Allocations images below are filtered to allocations under VersionRange.TryParse or NugetVersion.TryParse:

*** Previous ***
image

*** With changes ***
image

jeffkl:
@Erarndt do you mind taking a look and making any recommendations for us?

ToddGrun:
@Erarndt -- would love to get your thoughts!

ToddGrun:
Is there any data or explanation that I can provide to help out with the review?

ToddGrun:
gentle ping to @jeffkl or @Erarndt

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
```diff
@@ -11,6 +11,10 @@ namespace NuGet.Versioning
 {
     public partial class NuGetVersion
     {
+        // Cached mappings from string => NuGetVersion. On cache hit, avoids allocations during TryParse.
+        private static Dictionary<string, NuGetVersion> ParsedNuGetVersionsMapping = new Dictionary<string, NuGetVersion>();
```
zivkan: Dictionary doesn't have a `Capacity` property, so I can't check with C# Interactive.   What's the capacity of the dictionary's backing store when you add the first item to the dictionary?

My concern is that the cost of resizing the dictionary will be greater than the cost of overallocating a dictionary. I think we should create the dictionary passing in `MaxEntries` as the initial capacity.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
```diff
@@ -11,6 +11,10 @@ namespace NuGet.Versioning
...
+        private const int ParsedNuGetVersionsMappingMaxEntries = 500;
```
zivkan, line 15: We know the implementation of `Dictionary` uses prime numbers for backing store capacity. Therefore, I think we should set max entries to the closest prime number to avoid "wasted space". If it's going to be allocated anyway, might as well use it.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
ToddGrun: First add to a dictionary (on fx at least), allocates a backing store of size 3.

I went ahead and set the initial capacity to ParsedNuGetVersionsMappingMaxEntries, per your suggestion.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
ToddGrun: Done!

src/NuGet.Core/NuGet.Versioning/VersionRangeFactory.cs
```diff
@@ -13,6 +13,10 @@ public partial class VersionRange
     {
         // Static factory methods for creating version range objects.
 
+        // Cached mappings from (string value, bool allowFloating) => VersionRange. On cache hit, avoids allocations during TryParse.
+        private static Dictionary<(string, bool), VersionRange> ParsedVersionRangeMapping = new Dictionary<(string, bool), VersionRange>();
+        private const int ParsedVersionRangeMappingMaxEntries = 500;
```
zivkan, line 17: Same comments about the max entries const and dictionary initial capacity as I had in the other file.

ToddGrun: Sorry about that, I should have fixed that as part of the last commit.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
```diff
@@ -87,6 +101,16 @@ public static bool TryParse(string? value, [NotNullWhen(true)] out NuGetVersion?
                             metadata: buildMetadata ?? string.Empty,
                             originalVersion: originalVersion);
 
+                        lock (ParsedNuGetVersionsMapping)
+                        {
+                            if (ParsedNuGetVersionsMapping.Count >= ParsedNuGetVersionsMappingMaxEntries)
```
Erarndt, line 106: Out of curiosity, what's the reason we clear the cache above the threshold?

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
```diff
@@ -87,6 +101,16 @@ public static bool TryParse(string? value, [NotNullWhen(true)] out NuGetVersion?
                             metadata: buildMetadata ?? string.Empty,
                             originalVersion: originalVersion);
 
+                        lock (ParsedNuGetVersionsMapping)
```
Erarndt, line 104: Do you have measurements on approximately how much contention we'd expect here? Would it be worth considering `ConcurrentDictionary`?

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
Erarndt: Just an FYI, `Dictionary` will snap the passed in capacity to the smallest prime <= to the passed in capacity, so there wouldn't be any "wasted" space. Here's the relevant code in Framework:

    public Dictionary(int capacity)
        : this(capacity, (IEqualityComparer<TKey>)null)
    {
    }

    public Dictionary(int capacity, IEqualityComparer<TKey> comparer)
    {
        if (capacity < 0)
        {
            ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.capacity);
        }

        if (capacity > 0)
        {
            Initialize(capacity);
        }

        this.comparer = comparer ?? EqualityComparer<TKey>.Default;
    }

    private void Initialize(int capacity)
    {
        int prime = HashHelpers.GetPrime(capacity);
        buckets = new int[prime];
        for (int i = 0; i < buckets.Length; i++)
        {
            buckets[i] = -1;
        }

        entries = new Entry[prime];
        freeList = -1;
    }

    public static int GetPrime(int min)
    {
        if (min < 0)
        {
            throw new ArgumentException(Environment.GetResourceString("Arg_HTCapacityOverflow"));
        }

        for (int i = 0; i < primes.Length; i++)
        {
            int num = primes[i];
            if (num >= min)
            {
                return num;
            }
        }

        for (int j = min | 1; j < int.MaxValue; j += 2)
        {
            if (IsPrime(j) && (j - 1) % 101 != 0)
            {
                return j;
            }
        }

        return min;
    }

    public static readonly int[] primes = new int[72]
    {
        3, 7, 11, 17, 23, 29, 37, 47, 59, 71,
        89, 107, 131, 163, 197, 239, 293, 353, 431, 521,
        631, 761, 919, 1103, 1327, 1597, 1931, 2333, 2801, 3371,
        4049, 4861, 5839, 7013, 8419, 10103, 12143, 14591, 17519, 21023,
        25229, 30293, 36353, 43627, 52361, 62851, 75431, 90523, 108631, 130363,
        156437, 187751, 225307, 270371, 324449, 389357, 467237, 560689, 672827, 807403,
        968897, 1162687, 1395263, 1674319, 2009191, 2411033, 2893249, 3471899, 4166287, 4999559,
        5999471, 7199369
    };

Erarndt:
@ToddGrun Changes look reasonable to me. I added a few minor comments.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
ToddGrun: I didn't want unbounded growth. I saw this grow to a little over two hundred when opening Roslyn.sln, so I can't imagine many solutions will exceed this, but it felt safer to bound the growth just in case.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
ToddGrun: This particular call on the lock shouldn't hit much contention, however, the one at the beginning of the method could. Keep in mind that the code executed in that case is very light, just a quick dictionary lookup.

From my understanding, ConcurrentDictionary.Count is pretty bad performance wise. The usage of the lock was simple enough, combined with the Count() concern, that this seemed like the right approach.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
Erarndt: Totally onboard with bounding the growth. Looks like we clear out the cache entirely though once we hit the bound. Am I missing something?

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
Erarndt: Sounds good to me :)

jeffkl: Would this be a good use case for a per-key lock?  During a large restore, there could be thousands of total calls and lots of threads doing parsing.  I suppose the actual parsing is pretty fast and doesn't not require I/O so maybe the contention would be low?

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
ToddGrun: No, that's exactly what is done. Keeping track of request recency seemed like overkill, when it's fairly inexpensive to just let the cache fill back up naturally. My thought is that we should really only exceed this count when different solutions are opened inside the same VS session, so kicking out any versions requested by the old solutions seemed reasonable.

Erarndt: Gotcha, I was thinking the cache was purged between sessions. The code I was imagining would just avoid adding the item to the cache if we're at the limit. Something like:

                            
    if (ParsedNuGetVersionsMapping.Count < ParsedNuGetVersionsMappingMaxEntries)
    {
        ParsedNuGetVersionsMapping[value] = version;
    }

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
Erarndt: @jeffkl  I'm not sure there'd be much savings in this instance. We decided to take that approach with the `TaskResultCache` since kicking off the tasks can have arbitrary synchronous code execute before the task yields. The dictionary lookups/adds will be rather fast so there should be minimal contention unless this specific path gets absolutely hammered.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
ToddGrun: I think that it's best to just evacuate the cache if we hit the limit and let it start over. With the code you mention, I would have concerns about old and currently no longer used version information being in the cache over more relevant version information that was just requested.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
ToddGrun: Tracking at a per-key level, feels like re-implementing ConcurrentDictionary. If you feel fairly strong, I could just switch to ConcurrentDictionary and manually keep a best effort count manually.

Erarndt: My vote would be to keep the current implementation.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
Erarndt: Totally fair. The usage pattern is really going to determine the "right" approach here. I don't know enough to make a call with any confidence.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
ToddGrun: Very open to whatever the consensus is here. (I personally like the original approach, as I don't see us hitting the 500 entry limit as likely, and it seems a bit less complex to understand)

My understanding of the original request was that if there ended up being between 501-521, we would end up vacating the cache when adding to it wouldn't have required it to resize. I don't know if that's exactly true, as the dictionary code is a bit more complex than that (they keep track of collisiion count when walking the buckets).

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
Erarndt: Additional context I forgot to mention. The reason we went with a lock per key approach in the `TaskResultCache` was because of the guarantees we needed to provide:

1. Only one canonical task can be created per key
2. We want to be flexible enough to handle somewhat arbitrary use
3. Lock contention should be minimized.

A "quirk" of `ConcurrentDictionary` is the value factory is run *outside* of the lock internally when calling `GetOrAdd()`. You're guaranteed that each caller gets the same item back, but the item can be created multiple times. It was implemented this way to avoid holding the lock for longer running factory delegates. In `TaskResultCache`, we use a canonical lock object per key to allow more concurrency without creating multiple tasks for the same key.

src/NuGet.Core/NuGet.Versioning/NuGetVersionFactory.cs
Erarndt: From what I understand, the code ends up being identical if you specify 500 or 521. It uses the user specified capacity to set the initial bucket size, but it snaps the user specified capacity to the smallest prime that is greater than or equal to the specified capacity. I didn't intend for there to be any work as a result of my comment. Thought the background info would be interesting :)

ToddGrun: The code from the dictionary creation side is the same, whether that constant is 500 or 521. The code on my side that does the bounded size check is different based on that value. I'm just going to change back to 500 and remove the corresponding comment as it seems like more complication than that is worth.

Erarndt: Oh right. I totally overlooked that. No wonder I was confused :)

jeffkl (approved):

PR #5863, opened 2024-06-18
Make sure narrator reads source errors and warnings correctly when listing package sources in PMUI

### LINKED ISSUE #13555: Narator does not read the value of  `allowInsecureConnections`
How to replicate

1. create dotnet project
2. add a nuget.config file with an http source
3. set the http source allowInsecureConnections value to true
4. Now press ctrl+windows+enter to start narrator
5. Open PMUI and select the http package source
6. Listen to the narrator, it reads a warning even though there should not be a warning.

This is caused because the allowInsecureConnections value is not propagated down to the accessibility object

Nigusu-Allehu:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13555

## Description

This PR makes sure the narrator reads the right error message based on the value `allowInsecureConnections`
- previously the information was not being propagated as a result the narrator always reads http warnings even when it is not supposed to.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Options/CheckedListBoxAccessibleObject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Options/CheckedListBoxAccessibleObject.cs
@@ -35,6 +35,8 @@ namespace NuGet.PackageManagement.UI.Options
             {
                 var item = (PackageSourceContextInfo)CheckedListBox.Items[index];
                 PackageSource packageSource = new PackageSource(item.Source, item.Name);
+                packageSource.AllowInsecureConnections = item.AllowInsecureConnections;
+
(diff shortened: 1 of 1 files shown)


jeffkl (approved):

nkolev92 (approved):

jebriede (approved):

donnie-msft (approved):
We talked about this a few days ago and my understanding was that the problem being fixed is nothing specific to Narrator, just a logic bug due to the missing value which applies to all screen-readers.

PR #5865, opened 2024-06-19
Change the project template for test cases using IOS to Maui

SueSu01:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2828

## Description

1. Change the project template for test cases using IOS to Maui template.
2. Add the nuget.org to package source, since the following errors occur when installing a local package when package source only contains a local source.
image

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/test/NuGet.Tests.Apex/NuGet.Tests.Apex.Daily/NuGetConsoleTestCase.cs
+++ b/test/NuGet.Tests.Apex/NuGet.Tests.Apex.Daily/NuGetConsoleTestCase.cs
@@ -98,14 +98,15 @@ namespace NuGet.Tests.Apex.Daily
         [DataTestMethod]
         [DynamicData(nameof(GetMauiTemplates), DynamicDataSourceType.Method)]
         [Timeout(DefaultTimeout)]
-        public void InstallPackageForMauiProjectInPMC(ProjectTemplate projectTemplate)
+        public async Task InstallPackageForMauiProjectInPMC(ProjectTemplate projectTemplate)
         {
             EnsureVisualStudioHost();
             using (var simpleTestPathContext = new SimpleTestPathContext())
             {
                 // Arrange
-                var packageName = "log4net";
-                var v100 = "2.0.17";
+                var packageName = "IOSTestPackage";
+                var v100 = "1.0.0";
+                await CommonUtility.CreatePackageInSourceAsync(simpleTestPathContext.PackageSource, packageName, v100);
                 simpleTestPathContext.Settings.AddSource(NuGetConstants.NuGetHostName, NuGetConstants.V3FeedUrl);
 
                 using (var testContext = new ApexTestContext(VisualStudio, projectTemplate, Logger, simpleTestPathContext: simpleTestPathContext))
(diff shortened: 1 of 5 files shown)


test/NuGet.Tests.Apex/NuGet.Tests.Apex.Daily/NuGetConsoleTestCase.cs
```diff
@@ -202,7 +205,7 @@ public async Task UninstallPackageForIOSProjectInPMC(ProjectTemplate projectTemp
                     testContext.SolutionService.Build();
                     testContext.NuGetApexTestService.WaitForAutoRestore();
 
-                    //Asset
+                    // Asset
```
donnie-msft:
```suggestion
                    // Assert
```

test/NuGet.Tests.Apex/NuGet.Tests.Apex.Daily/NuGetConsoleTestCase.cs
```diff
@@ -96,17 +96,18 @@ public async Task UpdateAllPackagesInPMC(ProjectTemplate projectTemplate, string
...
             EnsureVisualStudioHost();
             using (var simpleTestPathContext = new SimpleTestPathContext())
             {
-                // Arrange
-                var packageName = "IOSTestPackage";
+                // Arrange
+                var packageName = "IOSTestPackage";
                 var v100 = "1.0.0";
                 await CommonUtility.CreatePackageInSourceAsync(simpleTestPathContext.PackageSource, packageName, v100);
+                simpleTestPathContext.Settings.AddSource(NuGetConstants.NuGetHostName, NuGetConstants.V3FeedUrl);
```
donnie-msft, line 110: Just a thought that I haven't tried, but looks feasible, is instead of adding a remote feed (nuget.org), could the necessary packages just be created in the simpleTestPathContext?

For example, 
`await CommonUtility.CreatePackageInSourceAsync(simpleTestPathContext.PackageSource, "Xamarin.Android.AndroidX.Common", v100);`

If this works, it seems better to me. The test's goal is that the console opens and installs a package, that package is in the Output pane, and no errors exist. It doesn't matter whether we download anything from a live feed or not, or even whether it's a real package with any assemblies.

donnie-msft (commented):
What are your thoughts on my suggested alternative to nuget.org? Thank you for working on this!

test/NuGet.Tests.Apex/NuGet.Tests.Apex.Daily/NuGetConsoleTestCase.cs
nkolev92: That'd be a constantly moving target. The auto referenced packages change.

donnie-msft: I see, so we want to test the side effect of this test that the real MAUI template can be downloaded from nuget.org.

SueSu01:
> What are your thoughts on my suggested alternative to nuget.org? Thank you for working on this!

@donnie-msft Thanks for your suggestion. I created the packages in local source and got the following errors after successfully installing the package.
image

donnie-msft (approved):
Looks like it's really loading the project template content, so makes sense now why you needed nuget.org. Thanks!

PR #5870, opened 2024-06-21
Remove NuGet.Packaging.Core

### LINKED ISSUE #13385: Remove NuGet.Packaging.Core code
Follow-up to https://github.com/NuGet/Home/issues/12495.

Once https://github.com/NuGet/Home/issues/12495 is done without repercussions, we can remove the code as well.

nkolev92:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13385

## Description

In https://github.com/NuGet/Home/issues/12495 we stopped publishing NuGet.Packaging.Core. 
There's been no feedback on that, so I think it's safe to remove.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


NuGet.sln
```diff
@@ -634,10 +632,6 @@ Global
 		{1EDFC48B-2CEC-49DD-90DA-3A69BDDAB8CF}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{1EDFC48B-2CEC-49DD-90DA-3A69BDDAB8CF}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{1EDFC48B-2CEC-49DD-90DA-3A69BDDAB8CF}.Release|Any CPU.Build.0 = Release|Any CPU
-		{02B7C422-C592-4BA6-89A7-ACBD7923B34D}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
```
jeffkl, line 637: Looks like the GUID needs to be scrubbed from the solution as well.

jeffkl (dismissed):

jeffkl (dismissed):

donnie-msft (dismissed):
You win Most Deleted Code PR for the week! 
Can this apply to servicing branches as well? I know the team didn't want to put a servicing branch checklist, so please put it in the description or a comment why we are/are not servicing this change.

nkolev92:
I'm not sure I see a reason why this needs to be serviced. It's like any code removal.

src/NuGet.Core/NuGet.Packaging.Core/NuGet.Packaging.Core.csproj
zivkan, line 1: I'm pretty sure we have scripts that validate that pack creates the correct number of nupkgs, so the devdiv PrivateBuild should fail (if it doesn't, I think we have a regression in our build scripts).  But you also need to check our release pipeline, because the job to push packages to nuget.org should also validate the expected packages. We want to make sure when this commit makes it to a GA release, we don't have a broken release pipeline.

nkolev92: The release definition has been taken care of already since we stopped publishing in 6.10. 

Updated the validation script. 
It seems like it runs, but the CI never kicked off due to an outage, so I never got the build validation for the particular commit.

donnie-msft (approved):

zivkan (approved):

PR #5627, opened 2024-02-07
Parse the assets file with System.Text.Json

jgonz120:
## Bug

Fixes: https://github.com/NuGet/Home/issues/12715

Regression? Last working version:

## Description
Currently we use JObject to parse the assets file. This library reads the full file into memory causing performance issues with the GC. This PR switches it to use System.Text.Json, allowing up to read the file in pieces and parse it without having to load the full file into memory. This work was done in a feature branch and this is the PR to merge that feature branch into dev
## PR Checklist

- [x] PR has a meaningful title
- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [x] Documentation PR or issue filled - https://github.com/NuGet/Client.Engineering/issues/2686
  - **OR**
  - [ ] N/A

PR #5872, opened 2024-06-26
Remove PackageDependencyGroupConverter 

### LINKED ISSUE #13445: allocation: nuget.protocol.dll!NuGet.Protocol.HttpCacheUtility+<CreateCacheFileAsync>d__.MoveNext|nuget.protocol.dll!NuGet.Protocol.PackageDependencyGroupConverter.ReadJson
Internal issue: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1864821
Code in: 

Some ideas: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs

* Avoid using LoadJson, iterate by token.
* Intern the dependency groups /deduplicate
* Ensure GetShortFolderName is not computed continously. The framework parsing tends to be cached, but another ideas to cache that as well.

martinrrm:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13445

## Description

The goal of this PR is to avoid using `JsonUtilities.LoadJson`, for that we are removing the converter and let Newtonsoft.Json handle the deserialization of PackageDependencyGroup

This converter is used in the PM UI when we are deserializing the registration API, specifically when we select a package and we retrieve all the packages versions. https://learn.microsoft.com/en-us/nuget/api/registration-base-url-resource#package-dependency

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [x] Test exception - Perf
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A


martinrrm:
This is my first time doing a perf issue, what should do to confirm this is an improvement?

Also, I tried doing something similar like [Parse the assets file with System.Text.Json by jgonz120 · Pull Request #5627 · NuGet/NuGet.Client (github.com)](https://github.com/NuGet/NuGet.Client/pull/5627), but unfortunately if we want to stop using Newtonsoft.json it will require more work in `PackageSearchMetadata`
.

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
```diff
@@ -19,9 +18,47 @@ public class PackageDependencyGroupConverter : JsonConverter
...
+                                }
+                            }
 
-            var fxName = set.Value<string>(JsonProperties.TargetFramework);
+                            if (id != null)
+                            {
+                                packages.Add(new Packaging.Core.PackageDependency(id, string.IsNullOrEmpty(version) ? null : VersionRange.Parse(version)));
+                            }
+                        }
+                        reader.Read();
```
jgonz120:
```
{
  "TargetFramework": "SomeFramework",
  "Dependencies": [
    {
      "PackageId": "SomeId",
      "Range": "SomeRange"
    }
  ]
}
```

If this is the file we're getting, on line 53 we will move the reader to the end of object, and then on line 23 call read again, I think read should be fine, and return false. 

However, if the part we're reading here is part of a larger file though it will continue reading right? 
```
{ 
  "PackageDependencyGroup": {
    "TargetFramework": "SomeFramework",
    "Dependencies": [
      {
        "PackageId": "SomeId",
        "Range": "SomeRange"
      }
    ]
  },
  "SomethingElse": {
    "uhoh": []
  }
}
```

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
```diff
@@ -19,9 +18,47 @@ public class PackageDependencyGroupConverter : JsonConverter
...
+                    }
+                    if (reader.Value.Equals(JsonProperties.TargetFramework))
```
kartheekp-ms: how about `else if` instead of `if`? My understanding is that the token value will be either `TargetFramework` or `Dependencies`.

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
martinrrm: Yeah this also fixed the failing tests

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
```diff
@@ -19,9 +18,47 @@ public class PackageDependencyGroupConverter : JsonConverter
 
         public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
         {
-            var set = JsonUtility.LoadJson(reader);
+            string fxName = null;
+            List<Packaging.Core.PackageDependency> packages = new List<Packaging.Core.PackageDependency>();
+            while (reader.Read() && reader.TokenType != JsonToken.EndObject)
```
kartheekp-ms: I was wondering if we could replace the nested while loops with the following code. This implementation uses Newtonsoft.Json's JArray.Load API to load the dependencies into a collection. I believe the following snippet is more concise, readable, and utilizes JArray for parsing dependencies, making it more efficient and less error-prone.

```cs
          while (reader.Read() && reader.TokenType != JsonToken.EndObject)
          {
              if (reader.TokenType == JsonToken.PropertyName)
              {
                  var propertyName = reader.Value.ToString();

                  switch (propertyName)
                  {
                      case JsonProperties.Dependencies:
                          // Dependencies are stored in an array
                          JArray dependenciesArray = JArray.Load(reader);
                          foreach (var dependency in dependenciesArray.Children<JObject>())
                          {
                              var id = dependency[JsonProperties.PackageId]?.ToString();
                              var version = dependency[JsonProperties.Range]?.ToString();

                              if (!string.IsNullOrWhiteSpace(id))
                              {
                                  var versionRange = string.IsNullOrWhiteSpace(version) ? null : VersionRange.Parse(version);
                                  packages.Add(new Packaging.Core.PackageDependency(id, versionRange));
                              }
                          }
                          break;

                      case JsonProperties.TargetFramework:
                          reader.Read();
                          fxName = reader.Value?.ToString();
                          break;
                  }
              }
          }
```

kartheekp-ms (dismissed):

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
kartheekp-ms: With this approach, I think we can also be deterministic about the count of packages, allowing us to pass the count when initializing the List object so that we don't have to worry about collection resizing. The only downside is the allocations caused by `JArray.Load(reader)`.

kartheekp-ms:
> This is my first time doing a perf issue, what should do to confirm this is an improvement?

@martinrrm A PerfView trace before and after the fix would help to understand the performance improvement resulting from the changes proposed in this PR.

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
martinrrm: I agree that this could make the code more readable, but the task here is to reduce allocations, that's why I'm trying to iterate by token and avoid using any `Load` function.

martinrrm:
### Before 
image


### After
image


Steps: 
1. Load Nuget.Client solution
2. Open PM UI for Solution
3. Select first installed package

I'm not 100% sure if I'm reading this right, but looks like there is an improvement 😄

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
kartheekp-ms: Can we get some performance/memory allocation metrics with this suggestion so that we can make an informed decision on which snippet to consider?

zivkan:
@martinrrm did you take that from PerfView's CPU Stacks view, or GC Heap Alloc view?

In any case, I think a BenchmarkDotNet app would be a simpler comparison, as you can microbenchmark just the converter, and the memory diagnoser will tell you exactly how much memory is allocated.

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
```diff
@@ -19,9 +18,46 @@ public class PackageDependencyGroupConverter : JsonConverter
...
+                        {
+                            string id = null;
+                            string version = null;
+                            while (reader.Read() && reader.TokenType != JsonToken.EndObject)
+                            {
+                                if (reader.TokenType.Equals(JsonToken.PropertyName) && reader.Value.Equals(JsonProperties.PackageId))
+                                {
+                                    reader.Read();
+                                    id = reader.Value.ToString();
+                                }
```
jeffkl: Should something check if `reader.Read()` returns `false`?  
```suggestion
                                if (reader.TokenType.Equals(JsonToken.PropertyName) && reader.Value.Equals(JsonProperties.PackageId) && reader.Read())
                                {
                                    id = reader.Value.ToString();
                                }
```

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
```diff
@@ -19,9 +18,46 @@ public class PackageDependencyGroupConverter : JsonConverter
...
+            string fxName = null;
+            List<Packaging.Core.PackageDependency> packages = new List<Packaging.Core.PackageDependency>();
+            while (reader.Read() && reader.TokenType != JsonToken.EndObject)
+            {
+                if (reader.TokenType.Equals(JsonToken.PropertyName))
+                {
+                    if (reader.Value.Equals(JsonProperties.Dependencies))
+                    {
+                        // Dependencies are stored in an array
+                        while (reader.Read() && reader.TokenType != JsonToken.EndArray)
```
zivkan: You could let Newtonsoft.Json handle the list itself. I can't remember the exact syntax, but something like

```
var converter = serializer.GetConverter<List<PackageDependency>>();
packages = (List<PackageDependency>)converter.ReadJson(reader, typeof(List<PackageDependency>), null, serializer);
```

We'll also need a `PackageDependencyConverter`, but the benefit is that each converter is super simple, only handling its own properties, and anything nested gets taken care by other converters.

Technically, not using more converters will have theoretically better performance, since we won't spend CPU cycles getting the correct converter. But I doubt it will have a measurable difference, especially compared to network latency.

if you don't want to delegate to different converters, I think you should at least have one method per object/array (this `ReadJson` only handles `PackageDependencyGroup`'s 2 properties, calls another method to handle the array, and that method calls another method to handle each `PackageDependency`). Keep in mind situations like [this](https://github.com/NuGet/Home/issues/13599) where long methods that throw unexpected exceptions are very difficult to fix without a repro we can debug, whereas a stack trace will pinpoint the location of an exception more precisely when methods are short.

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
zivkan: Anything using Newtonsoft.Json.Linq (so JArray, JToken, JObject) is going to be bad for allocations, we should no longer need benchmarks to tell us this, there's ample evidence from Prism, plus if you've spent just a little time debugging code that uses NJ.Linq, it's easy to logically reason about it as well.

Any time NJ's LINQ sees an object, it needs to create a `Dictionary<string, JToken>`, plus its backing array. Any object with more than 3 properties is going to cause backing array resizing, and for a dictionary, rather than a simple Array.Copy to the new backing array, it needs to re-bucket every item depending on its hash code.

This is all work that's completely unnecessary when we can use a converter that stores the object's fields as stack variables, not a dictionary, then creates the correct data type, rather than creating a `JObject` plus it's backing dictionary, plus the dictionary's backing array.

Even if the `JArray` is an array of `int` for example, `JArray` doesn't use generics, so it's going to contain a collection of objects, meaning all those `int`s have to be boxed. More likely though is that `JArray` has a collection of `JToken`, which best case has the same memory impact as boxing an `int`, but more likely is larger since the `JToken` will also need state to stay the token type is a number.

Newtonsoft.Json.Linq is great for rapid prototyping, but should be banned from any performance sensitive production code, IMO.

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
```diff
@@ -19,9 +18,46 @@ public class PackageDependencyGroupConverter : JsonConverter
...
+                        {
+                            string id = null;
+                            string version = null;
+                            while (reader.Read() && reader.TokenType != JsonToken.EndObject)
+                            {
+                                if (reader.TokenType.Equals(JsonToken.PropertyName) && reader.Value.Equals(JsonProperties.PackageId))
+                                {
+                                    reader.Read();
+                                    id = reader.Value.ToString();
```
zivkan: `reader` has a `ReadAsString()` method.

I think this code has risk of a NullReferenceException is a poorly implemented server sends json `{'id':null,}`.

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
jgonz120: At this point we're looking at a PropertyName, so if the reader returns false, that means we have a malformed json object. I forget how newtonsoft.json handles that situation 😅

zivkan: It also kind of doesn't matter, because the only time `reader.Read()` will return false is when there's no more data in the stream, which also means there's no `EndObject` token, which means the parsing should throw an exception.

If there isn't already a test for error handling when there are invalid json syntax, then there should be.

martinrrm:
Hi! I'm reopening the issue, I was finally able to work on it and run some benchmarks. Sorry for delay 😞 

Benchmarks 

| Method                   | Mean     | Error    | StdDev   | Gen0   | Gen1   | Allocated |
|------------------------- |---------:|---------:|---------:|-------:|-------:|----------:|
| ConverterWithoutLoadJson | 19.16 us | 0.134 us | 0.125 us | 0.8545 |      - |  21.18 KB |
| ConverterUsingLoadJson   | 28.62 us | 0.129 us | 0.121 us | 1.2512 | 0.0610 |  31.11 KB |

Looking at the numbers by removing the usage of LoadJson and avoid using NJ.LINQ we can improve allocations by 31.91% 

Branch with the benchmarks: dev-martinrrm-benchmarks

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyConverter.cs
```diff
@@ -0,0 +1,47 @@
...
+
+using System;
+using Newtonsoft.Json;
+using NuGet.Versioning;
+
+namespace NuGet.Protocol
+{
+    public class PackageDependencyConverter : JsonConverter
+    {
+        public override bool CanConvert(Type objectType) => objectType == typeof(Packaging.Core.PackageDependency);
```
kartheekp-ms: I think if we can inherit generic converter `JsonConverter<PackageDependency>` then we don't need this method. The primary benefit is improved type safety, with potential minor performance gains due to reduced type checking.

```cs
    public class PackageDependencyConverter : JsonConverter<PackageDependency>
    {
        public override bool CanWrite => false;

        public override PackageDependency ReadJson(JsonReader reader, Type objectType, PackageDependency existingValue, bool hasExistingValue, JsonSerializer serializer)
        {
            string id = null;
            string version = null;
            while (reader.Read() && reader.TokenType != JsonToken.EndObject)
            {
                if (reader.TokenType.Equals(JsonToken.PropertyName))
                {
                    if (reader.Value.Equals(JsonProperties.PackageId))
                    {
                        id = reader.ReadAsString();
                    }
                    else if (reader.Value.Equals(JsonProperties.Range))
                    {
                        version = id = reader.ReadAsString();
                    }
                }
            }
            if (id != null)
            {
                return new Packaging.Core.PackageDependency(id, string.IsNullOrEmpty(version) ? null : VersionRange.Parse(version));
            }
            return null;
        }

        public override void WriteJson(JsonWriter writer, PackageDependency value, JsonSerializer serializer)
        {
            throw new NotImplementedException();
        }
    }
```

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyConverter.cs
```diff
@@ -0,0 +1,47 @@
...
+        public override bool CanWrite => false;
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            string id = null;
+            string version = null;
+
+            while (reader.Read() && reader.TokenType != JsonToken.EndObject)
+            {
+                if (reader.TokenType.Equals(JsonToken.PropertyName))
```
zivkan: There's no else to this if, so if the server responds with something like `{ 'id': "Some.Id", "invalid": { } }`, this converter will exit at the wrong indentation.

Are you sure that this converter is necessary? I think there only needs to be a converter for `VersionRange` (I'm sure one already exists), and then Newtonsoft.Json can automatically figure out the `PackageDependency` data type by reflection, and do error handling correctly.

src/NuGet.Core/NuGet.Protocol/PublicAPI/net472/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,7 @@
 #nullable enable
+NuGet.Protocol.PackageDependencyConverter
```
zivkan: Let's not repeat the mistakes of the past. I dont' think there's any good reason for any of our converters to be part of the public API, so let's not make new ones public.  Please remember https://github.com/NuGet/NuGet.Client/blob/dev/docs/nuget-sdk.md#changes-to-apis every single time you see `PublicAPI.Unshipped.txt` in your own pull request, or other people's PRs. Making something new public just because something similar that already exists is public is exactly the type of problem we're trying to avoid.

kartheekp-ms:
@martinrrm - I am reviewing this code path for the first time. Hence, I think it would be great if we expand the PR description to include more details about the functionality. A few questions I have in mind are: when will this code path be executed, what would be the sample input data, and which tests can I execute locally to understand the functionality better?

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
```diff
@@ -17,11 +15,31 @@ public class PackageDependencyGroupConverter : JsonConverter
 
         public override bool CanWrite => false;
 
+        private readonly PackageDependencyConverter _converter = new PackageDependencyConverter();
+
         public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
         {
-            var set = JsonUtility.LoadJson(reader);
```
kartheekp-ms, line 22: The current implementation of [`JsonUtility.LoadJson`](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Protocol/Utility/JsonUtility.cs#L35-L38) has the following settings when loading the JSON file into a `JObject`. Can we confirm that the changes proposed in the PR have the same functionality (e.g., comments in the JSON content are ignored, etc.)?

https://github.com/NuGet/NuGet.Client/blob/7ca037b450fffbae993f65367e29b9c37da7a012/src/NuGet.Core/NuGet.Protocol/Utility/JsonUtility.cs#L15-L19

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyConverter.cs
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Newtonsoft.Json;
+using NuGet.Versioning;
+
+namespace NuGet.Protocol
+{
+    public class PackageDependencyConverter : JsonConverter
```
kartheekp-ms: I started debugging code for another work item and noticed that an exception was thrown at the following location. Then I realized that I had checked out this branch. Given that we made some changes to the deserialization of JSON in this PR, I wanted to make sure you are aware of it. I think we recently changed when we run Apex/E2E tests. It would be great if we could run those tests for this PR to ensure no regressions.

https://github.com/NuGet/NuGet.Client/blob/0ee341d7d1227b5d858d78c4a8d6879f319ee313/src/NuGet.Core/NuGet.Packaging/Core/PackageDependency.cs#L60

My repro steps are very simple:

- Started debugging the NuGet.Client repo in VS.
- Launched an existing Console application in the debugging experimental instance.
- Opened PM UI with nuget.org as the package source.
- A null reference exception was thrown at the same line of code mentioned above.

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
jgonz120: Those settings don't exist on the [JsonSerializerSettings](https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_JsonSerializerSettings.htm) class. Seem strange to mix these two implementations. Is PackageDependencyGroup known to have comments?

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyConverter.cs
jgonz120: I couldn't find a converter for it, but it seems simple enough to make. The range property name doesn't match the one on the PackageDependency object, but you can just nsj to handle that right?

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
```diff
@@ -17,11 +15,31 @@ public class PackageDependencyGroupConverter : JsonConverter
...
+                        while (reader.Read() && reader.TokenType != JsonToken.EndArray)
+                        {
+                            Packaging.Core.PackageDependency package = (Packaging.Core.PackageDependency)_converter.ReadJson(reader, typeof(Packaging.Core.PackageDependency), null, serializer);
+                            packages.Add(package);
+                        }
+                    }
+                    else if (reader.Value.Equals(JsonProperties.TargetFramework))
+                    {
+                        fxName = reader.ReadAsString();
+                    }
```
jgonz120: The same issue exists here as with the other converter. I think the best practice should be to [skip](https://www.newtonsoft.com/json/help/html/M_Newtonsoft_Json_JsonReader_Skip.htm) any properties that we don't expect.

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
zivkan: no, it's not known to have comments, but if it does, it's out of control of the customer if NuGet fails because of it.

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyConverter.cs
zivkan: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Protocol/Converters/VersionRangeConverter.cs

`[JsonPropertyName("range")]`

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyConverter.cs
zivkan: Please treat this as a bug and write a test that catches this scenario before fixing it, to ensure that it doesn't regress again. It appears to me that none of the existing tests caught this error.

martinrrm: The exception was caused because I had this on the converter when reading the range `version = id = reader.ReadAsString();`, this clearly caused an exception when packages have no version range. 

I'll add some unit tests for this converter, not sure if I can add one for that PM UI scenario, but I'm surprised this wasn't catch in the tests

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyConverter.cs
martinrrm: I tried using Newtonsoft.Json to automatically figure out `PackageDependency` by adding `JsonPropertyNames` to the class, I couldn't figure out why it wasn't working, it was complaining about the constructors not being correct (I did add the [JsonConstructor] property).

So, at the end I think I'm keeping this constructor and maybe it could be helpful in the future, it also helped me do unit tests. Also, I found that `JsonExtensions.JsonObjectSerializer` contains some of our converters, including the `VersionRangeConverter` and then just do `serializer.Deserialize<Packaging.Core.PackageDependency>(reader);`

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
martinrrm: At the end I'm using the settings we have in https://github.com/NuGet/NuGet.Client/blob/7856ac6d46ac73374ef24e6fa637283cd83ae54e/src/NuGet.Core/NuGet.Protocol/Converters/JsonExtensions.cs#L19-L35

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyConverter.cs
martinrrm: I'm closing this comment, this was my mistake, this converter is used with the registration API, and there is no scenario where the id is null https://learn.microsoft.com/en-us/nuget/api/registration-base-url-resource#package-dependency

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
```diff
@@ -19,9 +17,31 @@ public class PackageDependencyGroupConverter : JsonConverter
...
+                if (reader.TokenType.Equals(JsonToken.PropertyName))
+                {
+                    if (reader.Value.Equals(JsonProperties.Dependencies))
+                    {
+                        // Dependencies are stored in an array
+                        while (reader.Read() && reader.TokenType != JsonToken.EndArray)
+                        {
+                            Packaging.Core.PackageDependency package = serializer.Deserialize<Packaging.Core.PackageDependency>(reader);
+                            packages.Add(package);
+                        }
```
kartheekp-ms: I replaced the above code (entire while loop) with the following and the unit tests added in this PR passed.

```cs
reader.Read();
var packageDependencies = serializer.Deserialize<List<Packaging.Core.PackageDependency>>(reader);
packages.AddRange(packageDependencies);
```

I also removed the converter, and the unit tests added in this PR passed locally.
```cs
        public static readonly JsonSerializerSettings ObjectSerializationSettings = new JsonSerializerSettings
        {
            MaxDepth = JsonSerializationMaxDepth,
            NullValueHandling = NullValueHandling.Ignore,
            TypeNameHandling = TypeNameHandling.None,
            Converters = new List<JsonConverter>
            {
                new NuGetVersionConverter(),
                new VersionInfoConverter(),
                new StringEnumConverter { NamingStrategy = new CamelCaseNamingStrategy() },
                new IsoDateTimeConverter { DateTimeStyles = DateTimeStyles.AssumeUniversal },
                new FingerprintsConverter(),
                new VersionRangeConverter(),
                new PackageVulnerabilityInfoConverter()
            },
        };
```

kartheekp-ms: I guess this is what Andy was referring in another comment, is the converter really necessary?

kartheekp-ms: Forgot to mention above, I also added `JsonConstructor` attribute to the `PackageDependency` class in `NuGet.Packaging` assembly for tests in `$\test\NuGet.Core.Tests\NuGet.Protocol.Tests\JsonExtensionsTest.cs` to pass. 

        [JsonConstructor]
        public PackageDependency(string id, VersionRange versionRange)
            : this(id, versionRange, include: null, exclude: null)
        {
        }

martinrrm: Thanks for doing a deep dive into my code, I'll double check why mine wasn't working.

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
kartheekp-ms: If we can add a new unit test or modify existing ones to ensure that they pass even though the JSON content has a comment, then I guess we should be good.

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyConverter.cs
kartheekp-ms: How about running [NuGet.Client Daily Tests](https://devdiv.visualstudio.com/DevDiv/_build?definitionId=18097&_a=summary&view=runs) and [NuGet.Client-VS](https://devdiv.visualstudio.com/DevDiv/_build?definitionId=22360) for this PR branch just to make sure that all the current Apex and E2E tests pass? In the worst case, there should not be any new test failures when those pipelines are executed on this branch compared to the previous runs for those pipelines.

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
martinrrm: JSON objects cannot have a comments https://github.com/JamesNK/Newtonsoft.Json/issues/1315#issuecomment-302036211

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
martinrrm: It worked! The problem was that the `reader.Read();` was in the wrong place.

I'm decided to iterate over the json because it was throwing an error when opening the PM UI, it was complaining that I needed to save it to an JArray.

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
kartheekp-ms: > JSON objects cannot have a comments https://github.com/JamesNK/Newtonsoft.Json/issues/1315#issuecomment-302036211

That's for JObject. But James mentioned that in the same comment that `If you need to access the comment value then you will need to use the lower level JsonTextReader class to read it.`  If my understanding is correct, we are using lower-level reader here to deserialize JSON. Hence, it would be great to ensure that comments are ignored while deserializing JSON using JsonTextReader.

zivkan: Creating a console app and adding Newtonsoft.Json to it, I use this test data:

```cs
string json = @"{
""key"":""myKey"",
//  comment
""value"": ""myValue""
}";
```

Using `var jToken = JToken.Parse(json);` doesn't throw.

Using `var kvp = Newtonsoft.Json.JsonConvert.DeserializeObject<KeyValuePair<string, string>>(json);` does throw, and I couldn't figure out what `JsonSerializerSettings` to pass in to make it work, within 30 seconds.

However, this does not throw with System.Text.Json:
```cs
System.Text.Json.JsonSerializerOptions options = new()
{
    ReadCommentHandling = System.Text.Json.JsonCommentHandling.Skip
};
System.Text.Json.JsonSerializer.Deserialize<KeyValuePair<string, string>>(json, options);
```

So, if we can't figure out how to make Newtonsoft.Json's `JsonSerializer` ignore the comments, maybe we should put in the extra effort to port this to System.Test.Json instead. We're eventually going to have to anyway.

martinrrm: Sorry, I got busy with other things.

I'll try to change this to use `System.Test.Json`, but I think that is not super important, this is being used when we are deserializing the json from the registration api, I don't believe we will ever encounter a comment there

zivkan: Sorry, I forgot that the dependencies group is a child of the package metadata. We can't selectively port parts of a single deserialization from Newtonsoft.Json to System.Text.Json, so ignore that part of my previous comments. We'll have to port the entire registration page deserialization to System.Text.Json in one go.

src/NuGet.Core/NuGet.Packaging/Core/PackageDependencyGroup.cs
```diff
@@ -18,6 +19,21 @@ public class PackageDependencyGroup : IEquatable<PackageDependencyGroup>, IFrame
         private readonly NuGetFramework _targetFramework;
         private readonly IEnumerable<PackageDependency> _packages;
 
+        [JsonConstructor]
+        private PackageDependencyGroup(NuGetFramework targetFramework)
```
zivkan, line 23: I don't understand how the JSON deserialization sets the packages list, if this is the json constructor. But there are tests that pass? 😕

src/NuGet.Core/NuGet.Protocol/Converters/NuGetFrameworkConverter.cs
```diff
@@ -0,0 +1,33 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Xml.Linq;
+using Newtonsoft.Json;
+using NuGet.Frameworks;
+
+namespace NuGet.Protocol.Converters
+{
+    public class NuGetFrameworkConverter : JsonConverter
```
zivkan:
```suggestion
    internal class NuGetFrameworkConverter : JsonConverter
```

I don't think there's any good reason we need to make this public: https://github.com/NuGet/NuGet.Client/blob/dev/docs/nuget-sdk.md#changes-to-apis

src/NuGet.Core/NuGet.Protocol/Converters/NuGetFrameworkConverter.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            serializer.Serialize(writer, value.ToString());
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            var value = serializer.Deserialize<string>(reader);
```
zivkan:
```suggestion
            var value = reader.ReadString();
```

or something similar. Maybe you need to get string value, then call `Read()` to move to the next token.

Going through `JsonSerializer` will add additional CPU overhead that's not necessary (even worse if it causes additional allocations).

src/NuGet.Core/NuGet.Protocol/Converters/NuGetFrameworkConverter.cs
```diff
@@ -0,0 +1,33 @@
...
+using Newtonsoft.Json;
+using NuGet.Frameworks;
+
+namespace NuGet.Protocol.Converters
+{
+    public class NuGetFrameworkConverter : JsonConverter
+    {
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            serializer.Serialize(writer, value.ToString());
```
zivkan: nitpick: this is... complicated 😆 

The TFM value most people are used to reading, is not the value returned by `.ToString()`, but the value returned by `.GetShortFoldername()`.  And for some reason, `.nuspec` files use a 3rd format, which is what we'll probably see from package sources when deserializing dependency groups. 

But since this is a generic `NuGetFramework` converter, it's hard to say what's "correct" to use. However, I don't know of any scenario where NuGet will serialize this, so this method will probably never be used.

I wonder if we can use [this overload of `JsonConverterAttribute`](https://www.newtonsoft.com/json/help/html/M_Newtonsoft_Json_JsonConverterAttribute__ctor_1.htm) to pass data to the converter about which format type is preferred for the specific file.

zivkan (commented):
Can you please

1. Update the PR title and description to something more accurate for what this PR has become (especially since it's isolated to one specific type. it doesn't fix all usage of `LoadJson` across the entirety of NuGet.Client repo)
   * Be careful when merging the PR too. The default commit message is set by the PR's title at the time the PR was opened, not when the merge button is clicked.
2. Provide fresh benchmark results for deserializing some registration blob

thanks, this PR makes me happy!  less custom code.

src/NuGet.Core/NuGet.Packaging/Core/PackageDependencyGroup.cs
martinrrm: It's weird, since there is already a constructor that takes `targetFramework` and dependencies and null checks them I didn't want to change the behavior of it.

Previous converter would initialize an empty list for dependencies when it was null (no dependencies property in the json) thats why I added a constructor that takes only the `TargetFramework` that helps me create the empty list, this constructor is called only when deserialization happens. If we add a [JsonConstructor] it first attempts to use it and then populate the rest of the properties that have [JsonProperty]. `JsonProperty` doesn't needs a constructor to work.

martinrrm:
New benchmarks, I moved some things, so object initialization is not considered in the benchmarks. With these new changes looks like performance is better 🥳 

| Method                   | Mean      | Error     | StdDev    | Gen0   | Gen1   | Allocated |
|------------------------- |----------:|----------:|----------:|-------:|-------:|----------:|
| ConverterWithThatUsesLoadJson    | 15.660 us | 0.0328 us | 0.0274 us | 0.7019 | 0.0305 |  17.55 KB |
| RemovedPackageDependencyGroupConverter |  8.631 us | 0.0611 us | 0.0542 us | 0.3204 |      - |      8 KB |

src/NuGet.Core/NuGet.Protocol/Converters/NuGetFrameworkConverter.cs
martinrrm: I like your comment but since this is targeted for 17.2 P3 I believe I don't have enough time for this at this time, do you want me to investigate this further for next release?

src/NuGet.Core/NuGet.Packaging/Core/PackageDependencyGroup.cs
zivkan: The properties are get only, and they're not even auto-props. If Newtonsoft.Json is using reflection to set the value directly on the backing store, I don't understand how it can determine what the backing field name is. 🤷

src/NuGet.Core/NuGet.Protocol/Converters/NuGetFrameworkConverter.cs
zivkan: Honestly, I think my preference would be to throw on write, like the old converter did.  If anything needs to convert a NuGetFramework to a string in the future, then we can either investigate that "parameterized" converter, or create one converter per string format.

But I can also live with using `ToString`.

test/NuGet.Core.Tests/NuGet.Packaging.Test/PackageDependencyGroupTests.cs
```diff
@@ -218,5 +222,108 @@ public void PackageDependencyGroup_GetHashCode_SameVersionAndPackages_DifferentO
...
+            // Act
+            using var stringReader = new StringReader(PackageRegistrationDependencyGroupsJson);
+            using var jsonReader = new JsonTextReader(stringReader);
+            jsonReader.Read();
+
+            var serializer = JsonSerializer.Create(JsonExtensions.ObjectSerializationSettings);
+            PackageDependencyGroup actualPackageDependencies = serializer.Deserialize<PackageDependencyGroup>(jsonReader);
+
+            // Assert
+            Assert.Equal(packages, actualPackageDependencies.Packages);
```
zivkan:
```suggestion
            Assert.Equal(expectedPackages, actualPackageDependencies.Packages);
```

nitpick: small naming changes would help improve readability

test/NuGet.Core.Tests/NuGet.Packaging.Test/PackageDependencyGroupTests.cs
```diff
@@ -218,5 +222,108 @@ public void PackageDependencyGroup_GetHashCode_SameVersionAndPackages_DifferentO
 
             Assert.Equal(aHashCode, bHashCode);
         }
+
+        [Fact]
+        public void PackageDependencyGroup_Deserialize_ReturnsExpected()
+        {
+            // Arrange
+            const string id = "PackageA";
+            var version = new VersionRange(new NuGetVersion(1, 4, 1));
```
zivkan:
```suggestion
            var range = new VersionRange(new NuGetVersion(1, 4, 1));
```

nitpick: The range contains lower and upper version bounds, but it is not a version itself.

src/NuGet.Core/NuGet.Protocol/Converters/NuGetFrameworkConverter.cs
kartheekp-ms: Should we throw a not implemented exception here if there are going to be no callers for this method? We can address Andy's feedback in a follow up PR.

The existing code throw not implemented exception for the `WriteJson` method.

https://github.com/NuGet/NuGet.Client/pull/5872/files/f14e6f769612dbaf2e04eb9a702e7b3dcbe8b52b#diff-4e6ff8097dc206e72c42f58c234a65f2875f468c5bd9593c630a4cade52caf0aL49

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
kartheekp-ms: @martinrrm I have unresolved this conversation because you mentioned that `I don't believe we will ever encounter a comment in the registration API`. We were ignoring comments in the past. I am concerned that if this change might cause any regression if there are comments in the JSON payload.

As suggested in https://github.com/NuGet/NuGet.Client/pull/5872#discussion_r1724139234, if we can add a new unit test or modify existing ones to ensure that they pass even though the JSON content has a comment, then I guess we should be good.

kartheekp-ms:
Please consider updating the PR description to reflect the current state of the proposed code changes.

test/NuGet.Core.Tests/NuGet.Packaging.Test/PackageDependencyGroupTests.cs
```diff
@@ -218,5 +222,108 @@ public void PackageDependencyGroup_GetHashCode_SameVersionAndPackages_DifferentO
...
+            var expectedPackageDependency = new PackageDependency(id, version);
+            IEnumerable<PackageDependency> packages = new List<PackageDependency>() { expectedPackageDependency };
+
+            var targetFramework = NuGetFramework.AnyFramework;
+
+            // Act
+            using var stringReader = new StringReader(PackageRegistrationDependencyGroupsJson_NoTargetFramework);
+            using var jsonReader = new JsonTextReader(stringReader);
+            jsonReader.Read();
```
kartheekp-ms, line 264: Please help me understand why we have to invoke the Read method in the tests.

src/NuGet.Core/NuGet.Protocol/Converters/PackageDependencyGroupConverter.cs
kartheekp-ms: I see that you have already modified tests by adding comments in the JSON payload https://github.com/NuGet/NuGet.Client/pull/5872/files#diff-5d4614874e872a04258bc12705f69b82e95e76927320b5b0e8e2daa3dac1ccdcR303 I think we are good.

src/NuGet.Core/NuGet.Packaging/Core/PackageDependencyGroup.cs
```diff
@@ -18,6 +19,21 @@ public class PackageDependencyGroup : IEquatable<PackageDependencyGroup>, IFrame
...
+            if (targetFramework == null)
+            {
+                _targetFramework = NuGetFramework.AnyFramework;
+            }
+            else
+            {
+                _targetFramework = targetFramework;
+            }
+
+            _packages = new List<PackageDependency>();
```
kartheekp-ms, line 34: how about `Enumerable.Empty<PackageDependency>()` instead of creating a new List that initializes an array with the default capacity?

https://github.com/search?q=repo%3ANuGet%2FNuGet.Client%20Enumerable.Empty&type=code

zivkan (dismissed):
just small nitpicks, or nice to haves. No important comments.

src/NuGet.Core/NuGet.Protocol/Converters/NuGetFrameworkConverter.cs
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Newtonsoft.Json;
+using NuGet.Frameworks;
+
+namespace NuGet.Protocol.Converters
+{
+    internal class NuGetFrameworkConverter : JsonConverter
```
kartheekp-ms: I think we can use generic version of the JsonConverter for better type safety and may be better performance.

https://www.newtonsoft.com/json/help/html/CustomJsonConverterGeneric.htm

jgonz120 (dismissed):

test/NuGet.Core.Tests/NuGet.Packaging.Test/PackageDependencyGroupTests.cs
martinrrm: It is because the first token is null and doing the `Read()` moves the iterator to the json object, I'm not 100% sure why this happens but looks like we are doing it in other places.

kartheekp-ms (approved):

PR #5874, opened 2024-06-26
Attempt to make functional tests more reliable

jeffkl:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2501
Fixes: https://github.com/NuGet/Client.Engineering/issues/2866

## Description

The Functional tests seem to very unreliable with a few symptoms that this pull request hopefully addresses:
1. Test host process crashing
2. A particular command times out
3. A particular test or set of tests hang, causing ADO to cancel the test run

To attempt to address the issues, I've tweaked the parallelism in the test runner.  There are three settings in the `.runsettings` for parallelism:

| Setting | Purprose | Old value | New Value |
|---|---|---|---|
| `MaxCpuCount` | The number or processes that `vstest.console.exe` spawns to run tests in | 0 | 4 |
| `MaxParallelThreads` | The number of threads that the xunit runner can use at a time | 0 | 4 |
| `ParallelizeAssembly` | Whether or not the xunit test runner should run tests in assemblies in parallel | `true` | `false` |

Since we are currently using the default values for `MaxCpuCount` and `MaxParallelThreads`, as well as enabling `ParallelizeAssembly`, I suspect that `vstest.console.exe` is launching 12 process, each one then runs the tests for each assembly with 12 threads, and if one process is running multiple assemblies, they're happening in parallel.  My suspicion is that we're over parallelizing tests, leading to crashes.  By changing the values to lower, I do not see any increase in test run time and in fact, sometimes the runs are faster.  However, in my testing, I rarely see a test host process crashing anymore.

I also deleted `build\TestShared\xunit.runner.json` and moved the settings to `build\xunit.runsettings` so that everything is defined in one place.  Some projects were opting into test parallelization with the `UseParallelXunit` property which would copy the `xunit.runner.json` to the output directory, while other projects were copying it themselves.  I found it confusing to determine why some projects were allowed to run in parallel and others weren't so I changed how we run tests in parallel and made it uniform across the repo.

Finally, I've disabled the crash dump helper when running tests since all of the memory dumps it provided were always for the main `vstest.console.exe` which wasn't helpful since the tests themselves are running in a different process.  Instead, I've added the `--diag` switch to the call to `dotnet test` and now there's a diagnostic log for the main process and all of the child processes with valuable information.  These logs go to the binlog location and get uploaded when a test run fails or if you enable diagnostics on a particular pipeline run.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/build/xunit.runsettings
+++ b/build/xunit.runsettings
@@ -1,20 +1,60 @@
 <RunSettings>
-    <RunConfiguration>
-        <!-- Use as many testhosts as logical processors -->
-        <MaxCpuCount>0</MaxCpuCount>
-        <!-- Consider it an error if no tests are discovered -->
-        <TreatNoTestsAsError>true</TreatNoTestsAsError>
-    </RunConfiguration>
-    <DataCollectionRunSettings>
-        <DataCollectors>
-            <DataCollector friendlyName="blame" enabled="True">
-                <Configuration>
-                    <!-- Enables crash dump-->
-                    <CollectDump DumpType="Mini" />
-                    <!-- Enables hang dump-->
-                    <CollectDumpOnTestSessionHang TestTimeout="15min" HangDumpType="Mini" />
-                </Configuration>
-            </DataCollector>
-        </DataCollectors>
-    </DataCollectionRunSettings>
+  <!-- https://learn.microsoft.com/visualstudio/test/configure-unit-tests-by-using-a-dot-runsettings-file#runconfiguration-element -->
+  <RunConfiguration>
+    <!--
+      This setting controls the level of parallelism on process-level. Use 0 to enable the maximum process-level parallelism.
+
+      This setting determines the maximum number of test DLLs, or other test containers that can run in parallel. Each DLL runs in its own testhost process, and is isolated on the process level from the tests in other test DLLs. This setting doesn't force tests in each test DLL to run in parallel. Controlling the parallel execution within a DLL (on the thread-level) is up to the test framework such as MSTest, XUnit or NUnit.
+
+      The default value is 1, meaning that only one testhost runs at the same time. A special value 0 allows as many testhosts as you have logical processors (for example, 6, for a computer with 6 physical cores without multi-threading, or 12, for a computer with six physical cores with multi-threading).
+
+      The number of distinct DLLs in the run determines the actual number of testhosts started.
+    -->
+    <MaxCpuCount>4</MaxCpuCount>
+
+    <!-- Specify a Boolean value, which defines the exit code when no tests are discovered. If the value is true and no tests are discovered, a nonzero exit code is returned. Otherwise, zero is returned. -->
+    <TreatNoTestsAsError>true</TreatNoTestsAsError>
+
+    <EnvironmentVariables>
+      <!-- Changes the default timeout in VSTest for shutdown to 1000ms instead of 100ms which can cause intermittent failures in CI. -->
+      <VSTEST_TESTHOST_SHUTDOWN_TIMEOUT>1000</VSTEST_TESTHOST_SHUTDOWN_TIMEOUT>
+    </EnvironmentVariables>
+  </RunConfiguration>
+
+  <!-- https://xunit.net/docs/runsettings -->
+  <xUnit>
+    <!-- Set this value to true to include diagnostic information during test discovery and execution. Each runner has a unique method of presenting diagnostic messages. -->
+    <DiagnosticMessages>true</DiagnosticMessages>
+
+    <!-- Set this value to enable long-running (hung) test detection. When the runner is idle waiting for tests to finished, it will report that fact once the timeout has passed. Use a value of 0 to disable the feature, or a positive integer value to enable the feature (time in seconds). -->
+    <LongRunningTestSeconds>120</LongRunningTestSeconds>
+
+    <!-- Set this to override the maximum number of threads to be used when parallelizing tests within this assembly. Use a value of 0 or "default" to indicate that you would like the default behavior; use a value of -1 or "unlimited" to indicate that you do not wish to limit the number of threads used for parallelization. -->
+    <MaxParallelThreads>4</MaxParallelThreads>
+
+    <!-- Set this to true if this assembly is willing to participate in parallelization with other assemblies. Test runners can use this information to automatically enable parallelization across assemblies if all the assemblies agree to it. -->
+    <ParallelizeAssembly>false</ParallelizeAssembly>
+
+    <!-- Set this to true if the assembly is willing to run tests inside this assembly in parallel against each other. Tests in the same test collection will be run sequentially against each other, but tests in different test collections will be run in parallel against each other. Set this to false to disable all parallelization within this test assembly. -->
+    <ParallelizeTestCollections>false</ParallelizeTestCollections>
+
+    <!-- Set this to true to use shadow copying when running tests in separate App Domains; set to false to run tests without shadow copying. When running tests without App Domains, this value is ignored. -->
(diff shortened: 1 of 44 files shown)


build/xunit.runsettings
```diff
@@ -1,20 +1,44 @@
...
+    <DiagnosticMessages>true</DiagnosticMessages>
+
+    <!-- Set this value to enable long-running (hung) test detection. When the runner is idle waiting for tests to finished, it will report that fact once the timeout has passed. Use a value of 0 to disable the feature, or a positive integer value to enable the feature (time in seconds). -->
+    <LongRunningTestSeconds>120</LongRunningTestSeconds>
+
+    <!-- Set this to override the maximum number of threads to be used when parallelizing tests within this assembly. Use a value of 0 or "default" to indicate that you would like the default behavior; use a value of -1 or "unlimited" to indicate that you do not wish to limit the number of threads used for parallelization. -->
+    <MaxParallelThreads>4</MaxParallelThreads>
+
+    <!-- Set this to true if this assembly is willing to participate in parallelization with other assemblies. Test runners can use this information to automatically enable parallelization across assemblies if all the assemblies agree to it. -->
+    <ParallelizeAssembly>false</ParallelizeAssembly>
```
kartheekp-ms, line 36: I think this setting nullifies the `MaxcpuCount (4)` property because as per the comment it looks like `MaxcpuCount specifies the number of distinct DLLs in the run determines the actual number of testhosts started.`. Given that the multiple assemblies are not running parallel then my understanding is that only one test host process is started. I guess we should change to true so that 4 DLLs can run in parallel.

jeffkl: Since `MaxCpuCount` is set to `4` ([here](https://github.com/NuGet/NuGet.Client/pull/5874/files#diff-6b0fb5558d0f03f86a5e9cfee0a4f89ee36e018f308909358f9fe7b94e3bff34R13)), there will be 4 `vstest.console.exe` processes spawned.  I think it then launches each test assembly in those processes in parallel.  Then with `MaxParallelThreads` set to `4`, each test assembly runs with 4 threads (if there are enough CPUs).

This is really confusing since there are multiple test runners.  You can run tests under `xunit.runner.exe`, `vstest.console.exe`, etc.  I think the `ParallelizeAssembly` tells xunit specifically that it should run tests from multiple assemblies at the same time when running under certain test runners.  But since we're running the tests under `vstest.console.exe`, it seems to me that we should not be using xunit's parallelization.  I spent a lot of time read https://xunit.net/docs/runsettings but only experimenting with these numbers did I find a good balance between speed and reliability.  Do you know more about how all of this works?

kartheekp-ms: It looks like a similar issue has been discussed [here](https://github.com/xunit/visualstudio.xunit/issues/191#issuecomment-1763975518). I think we are good because it appears that xUnit parallelization is ignored by the `dotnet test` command. One interesting thing I learned from this link is that executing `dotnet test` without specifying assemblies runs test assemblies in parallel, whereas `dotnet test assembly1.dll assembly2.dll` does not run test assemblies in parallel. I am not sure if this is accurate in .NET 8, but at least that was the case earlier because the linked comment explained the different behaviors with an example.

test/TestUtilities/Test.Utility/CommandRunner.cs
```diff
@@ -53,6 +53,9 @@ public static CommandRunnerResult Run(string filename, string workingDirectory =
                     process.StartInfo.Environment["NUGET_SHOW_STACK"] = bool.TrueString;
                     process.StartInfo.Environment["NuGetTestModeEnabled"] = bool.TrueString;
                     process.StartInfo.Environment["UseSharedCompilation"] = bool.FalseString;
+                    process.StartInfo.Environment["DOTNET_SKIP_FIRST_TIME_EXPERIENCE"] = bool.TrueString;
+                    process.StartInfo.Environment["DOTNET_CLI_TELEMETRY_OPTOUT"] = bool.TrueString;
+                    process.StartInfo.Environment["SuppressNETCoreSdkPreviewMessage"] = bool.FalseString;
```
zivkan: is this a mistake?  Why wouldn't we want to suppress the .NET (Core) preview message?

eng/pipelines/pr.job.yml
```diff
@@ -78,7 +78,7 @@ jobs:
     condition: succeeded()
     inputs:
       command: test
-      arguments: --no-restore --restore:false --no-build --framework $(TestTargetFramework) --blame --logger trx --binarylogger:"$(BinlogDirectory)02-RunTests.binlog"
```
zivkan, line 81: > Finally, I've disabled the crash dump helper when running tests since all of the memory dumps it provided were always for the main vstest.console.exe which wasn't helpful since the tests themselves are running in a different process

Have you talked to the team who owns `dotnet test`?  This sounds very surprising so sounds like something that should be fixed for all customers, or if we're doing something wrong, they can help us.

build/xunit.runsettings
```diff
@@ -1,20 +1,44 @@
...
+
+    <!-- Set this to true if the assembly is willing to run tests inside this assembly in parallel against each other. Tests in the same test collection will be run sequentially against each other, but tests in different test collections will be run in parallel against each other. Set this to false to disable all parallelization within this test assembly. -->
+    <ParallelizeTestCollections>false</ParallelizeTestCollections>
```
zivkan, line 39: I compared the CI run for this PR to a different CI run, because I would have expected that setting ParallelizeTestCollections and ParallelizeAssembly both to false would make the tests run very slowly. But somehow this PR's CI run was margionally faster then the dev branch run.

It's totally not intuitive to me why, but I didn't see this mentioned in the PR description or anywhere else, so I wanted to point out that apparently the CI is not going to take a long longer with this PR merged.

zivkan (dismissed):

test/TestUtilities/Test.Utility/CommandRunner.cs
jeffkl: 🤦 Yes, thank you for catching that!

eng/pipelines/pr.job.yml
jeffkl: I see a few issues in the vstest repo:
https://github.com/microsoft/vstest/issues/3936
https://github.com/microsoft/vstest/issues/2593
https://github.com/microsoft/vstest/issues/4376
https://github.com/microsoft/vstest/issues/2952

Nothing specific to the problem I was seeing.  Supposedly the blame stuff also doesn't supply enough command-line arguments to capture other exceptions than `StackOverflowException` and `AccessViolationException`

https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/ProcDumpArgsBuilder.cs#L33

Although now there's an env var we can set to include more arguments to proc dump.  The main point in all of this is it seems like adding `--diag` to the `vstest.console.exe` execution is better.  I don't think we have that many "test crashes".  But if we feel like we need the memory dumps back, we can do so at a later time in my opinion.

build/xunit.runsettings
jeffkl: I struggle to understand it as well. Its still running tests assemblies in parallel with each other, but I guess not running tests themselves in parallel?  A machine with only 2 cores would only be able to run two tests at the same time anyway so I think the adjustments don't affect the maximum parallelization we could already achieve on CI agents.  It really just lowers the amount of parallelization that's queued up waiting for CPU.

zivkan (approved):

PR #5878, opened 2024-06-27
Catch all RestoreCommand exceptions and log them as an error

### LINKED ISSUE #13188: Multi-project restore should be resiliant to one project throwing an unhandled exception
See also https://github.com/NuGet/Home/issues/13085#issuecomment-1853072059  and https://github.com/NuGet/NuGet.Client/pull/5608#pullrequestreview-1842427425

If a solution has multiple projects, and RestoreCommand throws an unhandled exception while restoring one project, there should be some kind of graceful error handling, and allow all other projects to restore.

In VS, since restore is triggered every build, this means a restore error could occur very frequently as customers try different things to fix their build. The VS activity log only retains a fixed number of errors, so we should not spam the activity log with the same error over and over. This might mean not reporting these unhandled restore exceptions in the activity log, and recording them another way for us to investigate.

Nigusu-Allehu:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13188

## Description

* catch restore command exceptions and log them as an error to allow restore of other project in a multi-project solution

Solution structure:
     - bad.csproj : I edited our restore code to throw an exception whenever the restore project file is this file
     - good.csproj
     - multiprojectWithOneErrors.csproj

Restoring this solution now will result in an exception that interrupts the restore from restoring projects that did not cause the exception.
image
After catching the exceptions, restoring this solution looks like 
image

## PR Checklist

- [ ] PR has a linked issue.
- [ ] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Commands/PublicAPI/net472/PublicAPI.Unshipped.txt
+++ b/src/NuGet.Core/NuGet.Commands/PublicAPI/net472/PublicAPI.Unshipped.txt
@@ -1,2 +1,3 @@
 #nullable enable
 ~static NuGet.Commands.MSBuildRestoreUtility.GetRestoreAuditProperties(NuGet.Commands.IMSBuildItem specItem, System.Collections.Generic.HashSet<string> suppressionItems) -> NuGet.ProjectModel.RestoreAuditProperties
+~NuGet.Commands.RestoreSummary.RestoreSummary(bool success, System.Collections.Generic.IEnumerable<NuGet.Common.RestoreLogMessage> errors) -> void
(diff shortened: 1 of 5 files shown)


nkolev92 (changes requested):
I'm glad we're getting this fixed. 

A few things that'd be interesting to me before we commit to the particular solution. 

* How does Visual Studio behave in the same scenario? Ideally the experience is actionable in all scenarios.
* What kind of error in particular are we trying to solve? Can we add a test for that?
* Ideally for "uncontrolled" errors, we have a code and something telling our customers to file an issue. 

Some design considerations around restore changes that aren't really documented in a single space.

* For each restore, the output is an assets file, nuget.g.props and nuget.g.targets files. We try to generate that regardless of the reason for failure.
* All warnings and errors generated for a project must be part of the assets file. To achieve that, the errors/warnigns need to be generated in the RestoreCommand.
* In .NET SDK based projects, this is needed, because otherwise the warning/error will not even show up in the error list.

Nigusu-Allehu:
> I'm glad we're getting this fixed.
> 
> A few things that'd be interesting to me before we commit to the particular solution.
> 
> * How does Visual Studio behave in the same scenario? Ideally the experience is actionable in all scenarios.
> * What kind of error in particular are we trying to solve? Can we add a test for that?
> * Ideally for "uncontrolled" errors, we have a code and something telling our customers to file an issue.
> 
> Some design considerations around restore changes that aren't really documented in a single space.
> 
> * For each restore, the output is an assets file, nuget.g.props and nuget.g.targets files. We try to generate that regardless of the reason for failure.
> * All warnings and errors generated for a project must be part of the assets file. To achieve that, the errors/warnigns need to be generated in the RestoreCommand.
> * In .NET SDK based projects, this is needed, because otherwise the warning/error will not even show up in the error list.

Recreated here: https://github.com/NuGet/NuGet.Client/pull/5997

PR #5894, opened 2024-07-01
Simplify user-agent OS string

### LINKED ISSUE #13531: NuGet fails because of invalid characters in User-Agent header
### NuGet Product Used

dotnet.exe, NuGet SDK

### Product Version

9.0.100-preview.4.24267.66

### Worked before?

No

### Impact

### Repro Steps & Context

1. Install NixOS latest version (24.11 Vicuña).
2. Attempt to restore nuget packages for a project.
3. Unable to restore packages due to special character in header since the release name (Vicuña) is added to the HTTP header.

### Verbose Logs

```shell
/nix/store/lax73wx0qp7v2wpydc0bii5hszj1dys5-dotnet-core-combined/sdk/9.0.100-preview.4.24267.66/MSBuild.dll -nologo -distributedlogger:Microsoft.DotNet.Tools.MSBuild.MSBuildLogger,/nix/store/jd1kkgy9psy5km4m6k0drbzd7i3vqga6-dotnet-sdk-9.0.100-preview.4.24267.66/sdk/9.0.100-preview.4.24267.66/dotnet.dll*Microsoft.DotNet.Tools.MSBuild.MSBuildForwardingLogger,/nix/store/jd1kkgy9psy5km4m6k0drbzd7i3vqga6-dotnet-sdk-9.0.100-preview.4.24267.66/sdk/9.0.100-preview.4.24267.66/dotnet.dll -maxcpucount -target:Restore -tlp:default=auto -verbosity:m -verbosity:diag ./m3u8-dl.sln
    /home/ggg/Git/GGG-KILLER/m3u8-dl/src/m3u8-dl.csproj : error NU1301: Unable to load the service index for source https://api.nuget.org/v3/index.json.
    Failed to restore /home/ggg/Git/GGG-KILLER/m3u8-dl/src/m3u8-dl.csproj (in 5.76 sec).
    /home/ggg/Git/GGG-KILLER/m3u8-dl/test/m3u8-dl.Test.csproj : error NU1301: Unable to load the service index for source https://api.nuget.org/v3/index.json.
    /home/ggg/Git/GGG-KILLER/m3u8-dl/src/m3u8-dl.csproj : error NU1301: Unable to load the service index for source https://api.nuget.org/v3/index.json.
    Failed to restore /home/ggg/Git/GGG-KILLER/m3u8-dl/test/m3u8-dl.Test.csproj (in 11.61 sec).
```

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13531

Regression? no

## Description

This PR will change NuGet so that outgoing HTTP requests will no longer provide detailed operating system information in the User-Agent HTTP header. Going forward, it will only provide "Windows", "Linux", "OSX", and maybe "Browser". This will be invisible to customers running NuGet (except it will stop breaking for customers affected by the current bug). It is a kind of "breaking change" from the NuGet server's point of view. However, I checked with the nuget.org team, and they don't use it. Any other server that tries to parse the User-Agent header for telemetry or statistics reasons is going to be affected.

Unfortunately, I couldn't find an API that returns an `OSPlatform` or `OperatingSystem` object. So instead, we call APIs to check a bunch of platforms and return a value based on the first one to return true.

Use `RuntimeInformation.IsOSPlatform(..)` with the most common platforms (Windows, Linux, OSX) and just use the OSPlatform string in the User-Agent header. This avoids any future issues with different Linux distros using other characters that are not valid in the User-Agent HTTP header.

On .NET (Core), also test `OperatingSystem.IsBrowser()`, as I wouldn't be surprised if someone tries to use NuGet.Protocol in a Blazor app (although NuGet.Protocol writes to the filesystem, so not sure it world work). There are a bunch of other methods on `OperatingSystem`, like `IsAndroid`, `IsIOS`, `IsWatchOS`. But I don't think it's likely that developers will use NuGet.Protocol with apps on those platforms. We could add them all anyway, but then I feel like we're committing to updating the list each time .NET adds a new OS.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/XPlatUtility.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/XPlatUtility.cs
@@ -67,8 +67,7 @@ namespace NuGet.CommandLine.XPlat
         public static void SetUserAgent()
         {
 #if IS_CORECLR
-            UserAgent.SetUserAgentString(new UserAgentStringBuilder("NuGet xplat")
-                .WithOSDescription(RuntimeInformation.OSDescription));
+            UserAgent.SetUserAgentString(new UserAgentStringBuilder("NuGet xplat"));
 #else
             UserAgent.SetUserAgentString(new UserAgentStringBuilder("NuGet xplat"));
 #endif
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/XPlatUtility.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/XPlatUtility.cs
@@ -6,12 +6,6 @@
 using System.Globalization;
 using System;
 using System.IO;
-using System.Linq;
-
-#if IS_CORECLR
-using System.Runtime.InteropServices;
-#endif
-
 using NuGet.Common;
 using NuGet.Configuration;
 using NuGet.Protocol.Core.Types;
@@ -66,11 +60,7 @@ namespace NuGet.CommandLine.XPlat
 
         public static void SetUserAgent()
         {
-#if IS_CORECLR
             UserAgent.SetUserAgentString(new UserAgentStringBuilder("NuGet xplat"));
-#else
-            UserAgent.SetUserAgentString(new UserAgentStringBuilder("NuGet xplat"));
-#endif
         }
 
         internal static ISettings ProcessConfigFile(string configFile)
--- a/src/NuGet.Core/NuGet.Protocol/UserAgentStringBuilder.cs
+++ b/src/NuGet.Core/NuGet.Protocol/UserAgentStringBuilder.cs
@@ -106,6 +106,10 @@ namespace NuGet.Protocol.Core.Types
                 return OSPlatform.OSX.ToString();
             }
 #if NETCOREAPP
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.FreeBSD))
+            {
+                return OSPlatform.FreeBSD.ToString();
+            }
             else if (OperatingSystem.IsBrowser())
             {
                 return "BROWSER";
--- a/src/NuGet.Core/NuGet.Protocol/UserAgentStringBuilder.cs
+++ b/src/NuGet.Core/NuGet.Protocol/UserAgentStringBuilder.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Globalization;
+using System.Runtime.InteropServices;
 using NuGet.Packaging;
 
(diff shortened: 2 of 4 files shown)


nkolev92 (dismissed):

jeffkl:
I queued a PrivateDev manually since this PR is coming from a fork and it didn't trigger automatically.

jeffkl (dismissed):

aortiz-msft:
Would you please add a few sentences describing the impact on NuGet servers as well as the validation performed on the end-to-end customer scenario for NuGet.org?

src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/XPlatUtility.cs
```diff
@@ -67,8 +67,7 @@ public static void ConfigureProtocol()
         public static void SetUserAgent()
         {
 #if IS_CORECLR
-            UserAgent.SetUserAgentString(new UserAgentStringBuilder("NuGet xplat")
-                .WithOSDescription(RuntimeInformation.OSDescription));
+            UserAgent.SetUserAgentString(new UserAgentStringBuilder("NuGet xplat"));
 #else
             UserAgent.SetUserAgentString(new UserAgentStringBuilder("NuGet xplat"));
 #endif
```
jebriede: It looks like the compiler directive isn't needed anymore. In both cases I think it's doing the same thing now.

jebriede (dismissed):

src/NuGet.Core/NuGet.Protocol/UserAgentStringBuilder.cs
```diff
@@ -90,26 +86,35 @@ public string Build()
...
-                // When not on CoreClr and no OSDescription was provided,
-                // we will set it ourselves.
-                _osInfo = Environment.OSVersion.ToString();
-#else
-                // When on CoreClr, one should use the .WithOSDescription() method to set it.
-                _osInfo = string.Empty;
+                return "BROWSER";
+            }
 #endif
+            else
```
kartheekp-ms, line 118: It looks like [OSPlatform supports `FreeBSD`](https://learn.microsoft.com/dotnet/api/system.runtime.interopservices.osplatform?view=net-8.0) OS also, should we add support for that OS?

nkolev92 (approved):

PR #5895, opened 2024-07-01
raise warning if insecure sha-1 certificate fingerprint is passed to dotnet sign command

kartheekp-ms:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2914

## Description

Deprecate the usage of SHA-1 fingerprints in dotnet sign command especially for `--certificate-fingerprint` option. Instead, allow dotnet sign command to accept SHA-2 (SHA-256, SHA-384 and SHA-512) family fingerprints for searching a local certificate store for the certificate. 

Here is how the `dotnet sign` command works after this PR has been merged:

1. Validates the certificate fingerprint to ensure it is either SHA-1, SHA-256, SHA-384, or SHA-512 algorithm. Throws an error otherwise.
2. Raises a warning if a SHA-1 certificate fingerprint is provided. This warning will be promoted to an error in the future.
3. While searching for a certificate from the local store, if a SHA-1 hash is provided, finds the certificate using the existing API i.e. `store.Certificates.Find(X509FindType.FindByThumbprint, options.Fingerprint, validOnly);`
4. If a SHA-256, SHA-384, or SHA-512 certificate fingerprint is provided, loops through the certificates in the store to find the certificate with the matching hash.
 
## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added

- **Documentation**
  
  - [x] Documentation PR or issue filled https://github.com/NuGet/Client.Engineering/issues/2876


src/NuGet.Core/NuGet.Packaging/Signing/Utility/CertificateUtility.cs
```diff
@@ -382,5 +382,67 @@ public static IReadOnlyList<byte[]> GetRawDataForCollection(X509Certificate2Coll
...
+                    return false;
+            }
+        }
+
+        /// <summary>
+        /// Determines if the given string represents a valid hexadecimal value.
+        /// </summary>
+        /// <param name="certificateFingerprint">The string to check.</param>
+        /// <returns><c>true</c> if the string is a valid hexadecimal value; otherwise, <c>false</c>.</returns>
+        private static bool IsHex(string certificateFingerprint)
```
kartheekp-ms, line 428: Following the pattern established at: https://github.com/dotnet/sign/blob/main/src/Sign.Core/Tools/VsixSignTool/HexHelpers.cs#L17-L36

src/NuGet.Core/NuGet.Packaging/Signing/Utility/CertificateUtility.cs
```diff
@@ -382,5 +382,67 @@ public static IReadOnlyList<byte[]> GetRawDataForCollection(X509Certificate2Coll
...
             return certificatesRawData.AsReadOnly();
         }
+
+        /// <summary>
+        /// Tries to deduce the hash algorithm from the given certificate fingerprint.
+        /// </summary>
+        /// <param name="certificateFingerprint">The certificate fingerprint.</param>
+        /// <param name="hashAlgorithmName">The deduced hash algorithm name.</param>
+        /// <returns><c>true</c> if the hash algorithm was successfully deduced; otherwise, <c>false</c>.</returns>
+        public static bool TryDeduceHashAlgorithm(string certificateFingerprint, out HashAlgorithmName hashAlgorithmName)
```
kartheekp-ms, line 392: Following the pattern established at: https://github.com/dotnet/sign/blob/1df56d5c09b004fec6b5413dc6c61183553589ec/src/Sign.Cli/CertificateStoreCommand.cs#L150-L179

src/NuGet.Core/NuGet.Commands/SignCommand/CertificateProvider.cs
```diff
@@ -171,7 +171,24 @@ private static X509Certificate2Collection LoadCertificateFromStore(CertificateSo
...
+                    foreach (var cert in store.Certificates)
+                    {
+                        string actualFingerprint = CertificateUtility.GetHashString(cert, options.FingerprintHashAlgorithmName);
+
+                        if (string.Equals(actualFingerprint, options.Fingerprint, StringComparison.InvariantCultureIgnoreCase))
+                        {
+                            resultCollection.Add(cert);
+                            break;
+                        }
+                    }
```
kartheekp-ms: Following the pattern established at: https://github.com/dotnet/sign/blob/1df56d5c09b004fec6b5413dc6c61183553589ec/src/Sign.SignatureProviders.CertificateStore/CertificateStoreService.cs#L154-L177

src/NuGet.Core/NuGet.Common/HashAlgorithmName.cs
```diff
@@ -6,8 +6,9 @@ namespace NuGet.Common
     public enum HashAlgorithmName
     {
         Unknown = 0,
-        SHA256 = 1,
-        SHA384 = 2,
-        SHA512 = 3,
+        SHA1 = 1,
+        SHA256 = 2,
+        SHA384 = 3,
+        SHA512 = 4
```
zivkan: This is not a safe change. If a library is compiled against an old version of NuGet.Common, but then a new version of NuGet.Common is used, then the library will think it's using one of these enum values, but a different one will actually be used.

You must not change the existing values. add `SHA1` as `= 4`

src/NuGet.Core/NuGet.Commands/SignCommand/CertificateFindOptions.cs
```diff
@@ -42,10 +42,15 @@ internal class CertificateSourceOptions
...
         /// <summary>
-        /// The SHA-1 fingerprint of the certificate.
+        /// SHA256 or SHA384 or SHA512 fingerprint of the certificate.
         /// </summary>
         public string Fingerprint { get; set; }
 
+        /// <summary>
+        /// The SHA256 or SHA384 or SHA512 hash algorithm name used to calculate the fingerprint of the certificate.
+        /// </summary>
+        public HashAlgorithmName FingerprintHashAlgorithmName { get; set; }
```
zivkan: Considering we only support SHA algorithms, and each SHA algorithm has a different hash length, can't we deduce the algorithm from the fingerprint length?  There's a 1:1 mapping, unless we start to add support for other algorithms

src/NuGet.Core/NuGet.Commands/SignCommand/SignArgs.cs
```diff
@@ -76,6 +76,11 @@ public string PackagePath
         /// </summary>
         public string CertificateFingerprint { get; set; }
 
+        /// <summary>
+        /// Fingerprint hash algorithm for the certificate that can be used to search the local certificate store.
+        /// </summary>
+        public HashAlgorithmName CertificateHashAlgorithmName { get; set; }
```
zivkan: Same comment as elsewhere, can't we deduce the algorithm from the fingerprint length?

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XplatSignTests.cs
```diff
@@ -426,7 +427,56 @@ public void SignCommandArgParsing_ValidArgsContainsCertPathAsync_Succeeds()
                 });
         }
 
-        private void SignCommandArgs(Action<Mock<ISignCommandRunner>, CommandLineApplication, Func<LogLevel>, Func<SignArgs>> verify)
+        [Theory]
+        [InlineData("89967D1DD995010B6C66AE24FF8E66885E6E03A8")]
+        public void SignCommandArgParsing_LogsAWarningForInsecureCertificateFingerprint(string certificateFingerprint)
```
zivkan: Is this test missing more test cases?  It seems unusual to use a parameterised test, with only a single case.  Having said that, it looks like this test only checks SHA1 hashes, so there's no need for any other test cases, so a `[Fact]` is sufficient.

zivkan (changes requested):
I'm concerned that the API breaking change could lead to serious bugs.

src/NuGet.Core/NuGet.Commands/SignCommand/CertificateFindOptions.cs
kartheekp-ms: We are deducing the algorithm from the fingerprint length. https://github.com/NuGet/NuGet.Client/pull/5895/files#diff-d1848c7de42ac59aab4c7b5b1e4fcacf835ad5aed1018d6bdff8445d0d34433dR392-R422

The reason why I added this new property is because, if it is SHA-1 the following code works 
```cs 
store.Certificates.Find(X509FindType.FindByThumbprint, options.Fingerprint, validOnly);
```
Whereas for SHA-256, 384 and 512 we have to loop through the certificates to locate the certificate from the store.

https://github.com/NuGet/NuGet.Client/pull/5895/files#diff-50e182f9a08392f15c7d8229710796712bcaf61981d35eac78fbb2aa9665703cR180-R188


```cs
                    foreach (var cert in store.Certificates)
                    {
                        string actualFingerprint = CertificateUtility.GetHashString(cert, options.FingerprintHashAlgorithmName);

                        if (string.Equals(actualFingerprint, options.Fingerprint, StringComparison.InvariantCultureIgnoreCase))
                        {
                            resultCollection.Add(cert);
                            break;
                        }
                    }
```

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XplatSignTests.cs
kartheekp-ms: Good catch. VS Copilot refactored the code for me, and I forgot to edit this.

kartheekp-ms: Fixed in the latest commit.

src/NuGet.Core/NuGet.Common/HashAlgorithmName.cs
kartheekp-ms: Great feedback about API/ABI compatibility. Fixed in the latest commit.

src/NuGet.Core/NuGet.Commands/SignCommand/CertificateFindOptions.cs
kartheekp-ms: I thought it would be helpful to explain why I chose to deduce the hash algorithm based on the certificate fingerprint while parsing arguments passed to the command line.

1. If an invalid certificate fingerprint is passed, the command will exit early instead of invoking the core library. This means we have to deduce the hash algorithm while parsing the arguments.
2. Instead of deducing the hash algorithm while searching for the certificate fingerprint, I chose to add this property so that it can be used in the core library API.

src/NuGet.Clients/NuGet.CommandLine/Commands/SignCommand.cs
```diff
@@ -97,6 +97,7 @@ public SignArgs GetSignArgs()
                 CertificateStoreLocation = storeLocation,
                 CertificateSubjectName = CertificateSubjectName,
                 CertificateFingerprint = CertificateFingerprint,
+                CertificateHashAlgorithmName = HashAlgorithmName.SHA1,
```
kartheekp-ms: This is a temporary change. I will make same changes (warn for SHA-1 hash) in `nuget.exe sign` command code in a follow-up PR once this PR has been merged.

kartheekp-ms: This line of code has been removed from the latest commit since we removed the new property from the class.

src/NuGet.Core/NuGet.Commands/SignCommand/SignArgs.cs
kartheekp-ms: Removed this new property in the latest commit.

src/NuGet.Core/NuGet.Commands/SignCommand/CertificateFindOptions.cs
kartheekp-ms: Removed this new property in the latest commit.

zivkan (dismissed):

zivkan (dismissed):

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Signing/SignCommand.cs
```diff
@@ -232,6 +232,17 @@ private static StoreName ValidateAndParseStoreName(CommandOption store)
                 // Thow if the user provided a fingerprint and a subject
                 throw new ArgumentException(Strings.SignCommandMultipleCertificateException);
             }
+            else if (fingerprint.Value() != null)
```
jeffkl, line 235: What used to happen if the user specified an invalid fingerprint value?  Just curious.

jeffkl (dismissed):

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Signing/SignCommand.cs
kartheekp-ms: Previously, if the certificate fingerprint was invalid (e.g., a string with non-hex characters or an empty string), it would still pass to the Core libraries and fail later while searching for certificates. With this PR, we are failing fast if the certificate fingerprint is invalid.

zivkan (approved):

dtivel (approved):

PR #5896, opened 2024-07-02
Change NuGetAuditMode default to all for non-SDK style projects

### LINKED ISSUE #13584: Change NuGetAuditMode default to all for non-SDK style projects
### NuGet Product(s) Affected

NuGet.exe, MSBuild.exe

### Current Behavior

`NuGetAuditMode` defaults to `direct` ever since the feature was introduced in VS17.8.

### Desired Behavior

Change the default to `all`, so that customers get warned about transitive packages with known vulnerabilities, in addition to direct packages. Note that this default has already been changed for SDK style projects using the .NET 9 SDK.

**NOTE ON BREAKING CHANGE**

NuGetAudit raises warnings when packages with vulnerabilities are found. But many people use "treat errors as warnings", so it will give the appearance of a breaking change. Therefore, I've labeled this issue as "breaking change", even though it's technically not, in order to raise visibility.

### Additional Context

Just like the .NET 9 SDK NuGetAuditMode defaults change, non-SDK style customers must be able to configure their project to use `direct`, and we should respect `SdkAnalysisLevel` even for non-SDK style projects, so that custoemrs can set it in a repo level *Directory.Build.props* and have it apply to non-SDK style projects, in addition to SDK style projects.

For more information on NuGetAudit, including `TreatWarningsNotAsErrors`, see our docs on the feature: https://learn.microsoft.com/nuget/concepts/auditing-packages

zivkan:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13584

Regression? No

## Description

Change NuGet.targets' defaults for `NuGetAuditMode` to be roughly:

```
if (SdkAnalysisVersion != '')
    NuGetAuditMode = SdkAnalysisVersion >= '9.0.100' ? "all" : "direct"
else if (UsingMicrosoftNETSdk == 'true')
    // The .NET SDK only started setting SdkAnalysisVersion from 9.0.100, the same version we started changing defaults
    NugetAuditMode = "direct"
else
    NuGetAuditMode = "all"
```

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
+++ b/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
@@ -70,14 +70,20 @@ Copyright (c) .NET Foundation. All rights reserved.
   <PropertyGroup>
     <!-- Enable NuGetAudit by default -->
     <NuGetAudit Condition=" '$(NuGetAudit)' == '' ">true</NuGetAudit>
-    <!-- Report on low severity vulnerabilities, and higher. Allowed values are: low, moderate, high, critical -->
+    <!-- Report all severity vulnerabilities (low severity and higher). Allowed values are: low, moderate, high, critical -->
     <NuGetAuditLevel Condition=" '$(NuGetAuditLevel)' == '' ">low</NuGetAuditLevel>
-    <!-- Report known vulnerabilities on direct dependencies only. Allowed values are: direct, all -->
+    <!-- Report known vulnerabilities on direct and transitive dependencies, unless using .NET SDK with version lower than 9.0.100. Allowed values are: direct, all -->
     <NuGetAuditMode Condition="'$(NuGetAuditMode)' == ''
                     AND '$(NuGetExeSkipSdkAnalysisLevelCheck)' != 'true'
-                    AND $([MSBuild]::VersionGreaterThanOrEquals($([MSBuild]::ValueOrDefault('$(SdkAnalysisLevel)', '0.0')), '9.0.100'))"
-                    >all</NuGetAuditMode>
-    <NuGetAuditMode Condition=" '$(NuGetAuditMode)' == '' ">direct</NuGetAuditMode>
+                    AND '$(SdkAnalysisLevel)' != ''
+                    AND $([MSBuild]::VersionLessThan('$(SdkAnalysisLevel)', '9.0.100'))"
+                    >direct</NuGetAuditMode>
+    <NuGetAuditMode Condition="'$(NuGetAuditMode)' == ''
+                    AND '$(NuGetExeSkipSdkAnalysisLevelCheck)' != 'true'
+                    AND '$(UsingMicrosoftNETSdk)' == 'true'
+                    AND '$(SdkAnalysisLevel)' == ''"
+                    >direct</NuGetAuditMode>
+    <NuGetAuditMode Condition=" '$(NuGetAuditMode)' == '' ">all</NuGetAuditMode>
   </PropertyGroup>
 
   <PropertyGroup>
--- a/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
+++ b/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
@@ -72,7 +72,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     <NuGetAudit Condition=" '$(NuGetAudit)' == '' ">true</NuGetAudit>
     <!-- Report all severity vulnerabilities (low severity and higher). Allowed values are: low, moderate, high, critical -->
     <NuGetAuditLevel Condition=" '$(NuGetAuditLevel)' == '' ">low</NuGetAuditLevel>
-    <!-- Report known vulnerabilities on direct and transitive dependencies, unless using .NET SDK with version lower than 9.0.100. Allowed values are: direct, all -->
+    <!-- Report known vulnerabilities on direct and transitive dependencies, unless SdkAnalysisLevel (or equalivant) is lower than 9.0.100. Allowed values are: direct, all -->
     <NuGetAuditMode Condition="'$(NuGetAuditMode)' == ''
                     AND '$(NuGetExeSkipSdkAnalysisLevelCheck)' != 'true'
                     AND '$(SdkAnalysisLevel)' != ''
--- a/docs/feature-guide.md
+++ b/docs/feature-guide.md
@@ -97,6 +97,33 @@ These considerations are also useful for smaller changes that might not require 
   * World readiness considerations
 * User documentation
 
+#### Warnings and defaults
+
+The .NET 9 SDK introduces a new [`SdkAnalysisLevel` property](https://github.com/dotnet/designs/blob/main/proposed/sdk-analysis-level.md), which is intended to allow customers to avoid breaking changes while still upgrading to new versions of the build tools.
+Going forward, any in-scope change to NuGet must compare `SdkAnalysisLevel` to the version of the .NET SDK that corresponds to NuGet's dev branch is going to be inserted into.
+For example, NuGet 6.12 will ship in the .NET 9.0.100 SDK.
+Therefore, any in-scope change to NuGet in NuGet 6.12 must retain existing/previous behavior when `SdkAnalysisVersion` is lower than this version, and the new behavior applies only when it's equal or higher.
+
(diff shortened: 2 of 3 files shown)


src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
```diff
@@ -70,14 +70,20 @@ Copyright (c) .NET Foundation. All rights reserved.
...
-    <NuGetAuditMode Condition=" '$(NuGetAuditMode)' == '' ">direct</NuGetAuditMode>
+                    AND '$(SdkAnalysisLevel)' != ''
+                    AND $([MSBuild]::VersionLessThan('$(SdkAnalysisLevel)', '9.0.100'))"
+                    >direct</NuGetAuditMode>
+    <NuGetAuditMode Condition="'$(NuGetAuditMode)' == ''
+                    AND '$(NuGetExeSkipSdkAnalysisLevelCheck)' != 'true'
+                    AND '$(UsingMicrosoftNETSdk)' == 'true'
+                    AND '$(SdkAnalysisLevel)' == ''"
+                    >direct</NuGetAuditMode>
+    <NuGetAuditMode Condition=" '$(NuGetAuditMode)' == '' ">all</NuGetAuditMode>
```
nkolev92, line 86: This is changing the nuget.exe default to all as well correct?

Can we have a test for that?

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
```diff
@@ -70,14 +70,20 @@ Copyright (c) .NET Foundation. All rights reserved.
...
     <NuGetAuditLevel Condition=" '$(NuGetAuditLevel)' == '' ">low</NuGetAuditLevel>
-    <!-- Report known vulnerabilities on direct dependencies only. Allowed values are: direct, all -->
+    <!-- Report known vulnerabilities on direct and transitive dependencies, unless using .NET SDK with version lower than 9.0.100. Allowed values are: direct, all -->
     <NuGetAuditMode Condition="'$(NuGetAuditMode)' == ''
                     AND '$(NuGetExeSkipSdkAnalysisLevelCheck)' != 'true'
-                    AND $([MSBuild]::VersionGreaterThanOrEquals($([MSBuild]::ValueOrDefault('$(SdkAnalysisLevel)', '0.0')), '9.0.100'))"
-                    >all</NuGetAuditMode>
-    <NuGetAuditMode Condition=" '$(NuGetAuditMode)' == '' ">direct</NuGetAuditMode>
+                    AND '$(SdkAnalysisLevel)' != ''
+                    AND $([MSBuild]::VersionLessThan('$(SdkAnalysisLevel)', '9.0.100'))"
```
nkolev92, line 79: Was ValueOrDefault returned something that didn't evaluate correctly?

zivkan: ValueOrDefault worked fine, but now we have to use different defaults depending if it's SDK or non-SDK style.

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
zivkan: The test I already added `NuGetAuditModeDefaults_NuGetExeEmulation` checks that the property value that nuget.exe should get will have the expected value.

What kind of test were you thinking of?

zivkan: I should add, although I added the tests to `msbuild.integration.test`, if you look at the test project, it's clear that the test is really testing NuGet.targets without any project file.  I wanted to add this test to NuGet.Build.Tasks.Tests (the unit test project), however, trying to use MSBuild's `Project` or `ProjectCollection`, the test fails because MSBuild's dlls can't be found. I tried using the MSBuild locator, but then other tests failed. It was just easier to put this test in msbuild.integration.tests.

nkolev92 (dismissed):

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
```diff
@@ -70,14 +70,20 @@ Copyright (c) .NET Foundation. All rights reserved.
   <PropertyGroup>
     <!-- Enable NuGetAudit by default -->
     <NuGetAudit Condition=" '$(NuGetAudit)' == '' ">true</NuGetAudit>
-    <!-- Report on low severity vulnerabilities, and higher. Allowed values are: low, moderate, high, critical -->
+    <!-- Report all severity vulnerabilities (low severity and higher). Allowed values are: low, moderate, high, critical -->
     <NuGetAuditLevel Condition=" '$(NuGetAuditLevel)' == '' ">low</NuGetAuditLevel>
-    <!-- Report known vulnerabilities on direct dependencies only. Allowed values are: direct, all -->
+    <!-- Report known vulnerabilities on direct and transitive dependencies, unless using .NET SDK with version lower than 9.0.100. Allowed values are: direct, all -->
```
aortiz-msft:
```suggestion
    <!-- Report known vulnerabilities on direct and transitive dependencies, unless using a .NET SDK with version lower than 9.0.100 (or using an SdkAnalysisLevel of .NET 9.0.100 or lower). Allowed values are: direct, all -->
```

aortiz-msft (dismissed):

jebriede (dismissed):

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
```diff
@@ -70,14 +70,20 @@ Copyright (c) .NET Foundation. All rights reserved.
...
                     AND '$(NuGetExeSkipSdkAnalysisLevelCheck)' != 'true'
-                    AND $([MSBuild]::VersionGreaterThanOrEquals($([MSBuild]::ValueOrDefault('$(SdkAnalysisLevel)', '0.0')), '9.0.100'))"
-                    >all</NuGetAuditMode>
-    <NuGetAuditMode Condition=" '$(NuGetAuditMode)' == '' ">direct</NuGetAuditMode>
+                    AND '$(SdkAnalysisLevel)' != ''
+                    AND $([MSBuild]::VersionLessThan('$(SdkAnalysisLevel)', '9.0.100'))"
+                    >direct</NuGetAuditMode>
+    <NuGetAuditMode Condition="'$(NuGetAuditMode)' == ''
+                    AND '$(NuGetExeSkipSdkAnalysisLevelCheck)' != 'true'
+                    AND '$(UsingMicrosoftNETSdk)' == 'true'
```
kartheekp-ms, line 83: Looking at the references of [`UsingMicrosoftNETSdk`](https://github.com/search?q=repo%3ANuGet%2FNuGet.Client+UsingMicrosoftNETSdk&type=code) property in the repo, I see some casing differences. 

For example, the casing in the following line of code is different. Are `_project.GetPropertyValue("UsingMicrosoftNETSDK"); ` and `_project.GetPropertyValue("UsingMicrosoftNETSdk"); ` are functionally equivalent?

https://github.com/NuGet/NuGet.Client/blob/4a1faa2bcc771862a0ee1d241ebbfa1c196630ab/src/NuGet.Clients/NuGet.CommandLine/Commands/ProjectFactory.cs#L241

kartheekp-ms, line 83: I couldn't find documentation related to `UsingMicrosoftNETSdk` property. For my understanding, does this property need to be set in the .csproj or some props file? The .NET SDK does not set this property, correct?

zivkan: MSBuild property names (and value comparisons) are case insensitive.

zivkan: no, the .NET SDK sets it, and I wouldn't expect anything else to set it: https://github.com/dotnet/sdk/blob/7fbff0fd23c8a68879d7c93ac2e9583dfc55b80d/src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props#L22-L31

nkolev92 (approved):

jeffkl (approved):

PR #5902, opened 2024-07-03
Disable CodeQL

jeffkl:
## Description

CodeQL injects some command-line arguments that make the builds in our CI pipeline take a lot longer.  We don't need it enabled for this pipeline so this change disables it.

## PR Checklist

- [ ] PR has a linked issue.
- [ ] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/eng/pipelines/pr.yml
+++ b/eng/pipelines/pr.yml
@@ -33,6 +33,7 @@ variables:
   NUGET_EXPERIMENTAL_CHAIN_BUILD_RETRY_POLICY: 3,1000
   SkipSigning: true
   CI: true
+  Codeql.Enabled: false
 
 trigger:
   branches:


eng/pipelines/pr.yml
```diff
@@ -33,6 +33,7 @@ variables:
   NUGET_EXPERIMENTAL_CHAIN_BUILD_RETRY_POLICY: 3,1000
   SkipSigning: true
   CI: true
+  Codeql.Enabled: false
```
Nigusu-Allehu, line 36: So, we only scan in the "official" pipeline?

Nigusu-Allehu (approved):

eng/pipelines/pr.yml
jeffkl: We can scan here, but these jobs are all for running tests.  I think I'll add a new job that just builds and runs CodeQL or all SDL tooling.  We just don't need to run it for the current jobs because their intent is to run tests

PR #5903, opened 2024-07-03
Make assets file reading more permissive, allow nulls in arrays to be ignored

### LINKED ISSUE #13563: Restore may write nulls to project.assets.json
### NuGet Product Used

MSBuild.exe

### Product Version

9.0.100-preview.5.24307.3

### Worked before?

9.0.100-preview.3.24204.13

### Impact

### Repro Steps & Context

SDK task ResolvePackageAssets throws NRE while enumerating `NuGet.ProjectModel.ProjectFileDependencyGroup.Dependencies`, which returns null. 

I'm guessing that's because the `project.assets.json` contains nulls in `projectFileDependencyGroups`, which it probably shouldn't. Somehow nuget restore managed to produce these:

```
"libraries": {},
  "projectFileDependencyGroups": {
    ".NETFramework,Version=v4.7.2": [
      null,
      null,
      null,
      null,
      null,
      null,
      "Microsoft.DotNet.Build.Tasks.VisualStudio >= 9.0.0-beta.24317.3",
      "Microsoft.Net.Compilers.Toolset >= 4.11.0-2.24270.4",
      "Microsoft.VisualStudioEng.MicroBuild.Plugins.SwixBuild >= 1.1.286"
    ]
  },
```

```
C:\Program Files\dotnet\sdk\9.0.100-preview.5.24307.3\Sdks\Microsoft.NET.Sdk\targets\Microsoft.PackageDependencyResolution.targets(266,5): error MSB4018: The "ResolvePackageAssets" task failed unexpectedly.
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.NET.Build.Tasks.ProjectContext.<>c__DisplayClass31_0.<GetTopLevelDependencies>b__3(String projectFileDependency)
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at Microsoft.NET.Build.Tasks.ResolvePackageAssets.CacheWriter.ComputePackageExclusions()
   at Microsoft.NET.Build.Tasks.ResolvePackageAssets.CacheWriter..ctor(ResolvePackageAssets task)
   at Microsoft.NET.Build.Tasks.ResolvePackageAssets.CacheReader.CreateReaderFromDisk(ResolvePackageAssets task, Byte[] settingsHash)
   at Microsoft.NET.Build.Tasks.ResolvePackageAssets.CacheReader..ctor(ResolvePackageAssets task)
   at Microsoft.NET.Build.Tasks.ResolvePackageAssets.ReadItemGroups()
   at Microsoft.NET.Build.Tasks.ResolvePackageAssets.ExecuteCore()
   at Microsoft.NET.Build.Tasks.TaskBase.Execute()
   at Microsoft.Build.BackEnd.TaskExecutionHost.Execute()
   at Microsoft.Build.BackEnd.TaskBuilder.<ExecuteInstantiatedTask>d__26.MoveNext() 
[D:\R4\src\VisualStudio\Setup\Roslyn.VisualStudio.Setup.csproj]
```

### Verbose Logs

Can share full binlog. DM me on Teams.
Searching for `null,`:
image

nkolev92:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13563

## Description

This is a 2nd fix after https://github.com/NuGet/NuGet.Client/commit/ae8c3e443cce585cfc973075a8d9c7a809c549fd.

There's a discrepancy in how the arrays are read in NJ and STJ reading.
Unfortunately it's not 100% consistent in the reading overall (not a NJ vs STJ thing, but within the same implementation), so it's hard to really tell what the correct thing.

In this case, I analyzed before and after of the code paths I'm changing and I summarized the results. 

Here's the table: 

| Element | Null Safe | Reasoning |
| ----------|----------|-------------|
| Authors | not null safe | Deprecated and unused waiting to be removed, so we don't care. |
| ContentFiles | not null safe | Deprecated and unused waiting to be removed, so we don't care. |
| ConfigFilePaths | not null safe | Unlikely to have nulls at write, even through bugs. |
| FallbackFolders| not null safe | Unlikely to have nulls at write, even through bugs. |
| OriginalTargetFrameworks | not null safe | Unlikely to have nulls at write, even through bugs. |
| Owners | not null safe | Deprecated and unused waiting to be removed, so we don't care. |
| Tags | not null safe | Deprecated and unused waiting to be removed, so we don't care. |
|  runtimedescription | not null safe | could theoretically be null, but unlikely |
| TargetGraph in log messages | null safe | This change aligns it. |
| Files list in lock file target library | null safe | This change aligns it. |
| framework assembly in lock file target library | null safe | This change aligns it. |
| framework reference in lock file target library | null safe | This change aligns it. |
| dependencies in project file dependency group (the original issue)  | null safe | This change aligns it. |

The primary risk of taking this PR is that we would be potentially hiding the original issue. One would argue that's bad, as we'd rather see an issue reproing consistently than sometimes. 
This way, we're "more reliable" at the potential risk of bugs. Note that the roslyn would've seen their bug through NU1008 that was raised.

Open to opinions. Not sold 100% on either approach.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
+++ b/src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
@@ -180,9 +180,12 @@ namespace NuGet.ProjectModel
                 {
                     string value = _reader.ReadTokenAsString();
 
-                    strings ??= new List<string>();
+                    if (value != null)
+                    {
+                        strings ??= new List<string>();
 
-                    strings.Add(value);
+                        strings.Add(value);
+                    }
                 }
             }
             return strings;
@@ -250,7 +253,7 @@ namespace NuGet.ProjectModel
 
             while (Read() && _reader.TokenType != JsonTokenType.EndArray)
             {
-                string value = _reader.ReadTokenAsString();
+                string value = _reader.ReadTokenAsString(); // TODO NK - why is this one null accepting?
 
                 strings ??= new List<string>();
 
--- a/src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
+++ b/src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
@@ -253,7 +253,7 @@ namespace NuGet.ProjectModel
 
             while (Read() && _reader.TokenType != JsonTokenType.EndArray)
             {
-                string value = _reader.ReadTokenAsString(); // TODO NK - why is this one null accepting?
+                string value = _reader.ReadTokenAsString();
 
                 strings ??= new List<string>();
 
--- a/test/NuGet.Core.Tests/NuGet.ProjectModel.Test/Utf8JsonStreamReaderTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.ProjectModel.Test/Utf8JsonStreamReaderTests.cs
@@ -569,8 +569,7 @@ namespace NuGet.ProjectModel.Test
                     actualValue => Assert.Equal("a", actualValue),
                     actualValue => Assert.Equal("-2", actualValue),
                     actualValue => Assert.Equal("3.14", actualValue),
-                    actualValue => Assert.Equal("True", actualValue),
-                    actualValue => Assert.Null(actualValue));
+                    actualValue => Assert.Equal("True", actualValue));
                 Assert.Equal(JsonTokenType.EndArray, reader.TokenType);
             }
         }


src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -180,9 +180,12 @@ internal IList<string> ReadStringArrayAsIList(IList<string> strings = null)
                 {
                     string value = _reader.ReadTokenAsString();
 
-                    strings ??= new List<string>();
+                    if (value != null)
+                    {
+                        strings ??= new List<string>();
```
Nigusu-Allehu, line 186: nit: do we need the white line?

Nigusu-Allehu (dismissed):

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
nkolev92: not sure I follow what you're referring to.

Nigusu-Allehu: Line 186, but it was not added in this PR.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -180,9 +180,12 @@ internal IList<string> ReadStringArrayAsIList(IList<string> strings = null)
                 {
                     string value = _reader.ReadTokenAsString();
 
-                    strings ??= new List<string>();
+                    if (value != null)
```
zivkan, line 183: Would it be "better" to throw, rather than silently ignore what we consider a badly formed assets file?  On one hand, silently ignoring nulls will not fail builds, at least not immediately, and not in a code path that NuGet will get blamed for.  On the other hand, if we think that this should never happen, by silently ignoring it, customers are likely to get obscure build errors which will be harder to track down.  If the problem is that the assets file has a null where it shouldn't, the earlier we raise an error about it, the quicker we'll figure out the root cause.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
```diff
@@ -250,7 +253,7 @@ internal IReadOnlyList<string> ReadStringArrayAsReadOnlyListFromArrayStart()
 
             while (Read() && _reader.TokenType != JsonTokenType.EndArray)
             {
-                string value = _reader.ReadTokenAsString();
+                string value = _reader.ReadTokenAsString(); // TODO NK - why is this one null accepting?
```
zivkan: This PR isn't marked as draft, and according to our coding guidelines, we don't do TODOs in our code without a linked github issue.

zivkan (changes requested):

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
nkolev92: The concern is that we just don't know. 
Unfortunately, our reading implementation is all over the place, with it sometimes being permissive and other times not, and this is one of those where I'm not sure I want to rock the boat too much until it proves to be a bigger problem. 

That was part of the concern in the PR body.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
nkolev92: My bad, removed.

src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReader.cs
nkolev92: I'm ok with this not happening and us tabling the idea of making the parsing stricter in a future change.
With the 2 reading implementations and with both of them being inconsistent, I feel like we might be better to wait for 1 to go away before we remove the null safety checks.

zivkan (approved):

nkolev92:
Closing this in favor of a change that enables nullability and changes both the writer and reader.

PR #5904, opened 2024-07-03
Move static source analysis to CI pipeline from PrivateDev

jeffkl:
## Description

This moves the static source analysis from the PrivateDev build to the new CI pipeline

## PR Checklist

- [ ] PR has a linked issue.
- [ ] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [ ] N/A

### DIFF
--- a/eng/pipelines/templates/pipeline.yml
+++ b/eng/pipelines/templates/pipeline.yml
@@ -41,6 +41,10 @@ parameters:
   displayName: Run Mono tests on Mac
   type: boolean
   default: false
+- name: RunStaticAnalysis
+  displayName: Run static analysis
+  type: boolean
+  default: true
 
 stages:
 - stage: Initialize
@@ -74,6 +78,7 @@ stages:
   jobs:
   - job: RunStaticSourceAnalysis
     displayName: Run Static Source Analysis
+    condition: "and(succeeded(), ne(variables['RunStaticAnalysis'], 'false'))"
     timeoutInMinutes: 15
     pool:
       name: AzurePipelines-EO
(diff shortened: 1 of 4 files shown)


eng/pipelines/templates/pipeline.yml
```diff
@@ -74,6 +78,7 @@ stages:
   jobs:
   - job: RunStaticSourceAnalysis
     displayName: Run Static Source Analysis
+    condition: "and(succeeded(), ne(variables['RunStaticAnalysis'], 'false'))"
```
zivkan, line 81: Did you test this?  As far as I know, parameters != variables, so I don't think this will work.

zivkan: From https://learn.microsoft.com/en-us/azure/devops/pipelines/process/runtime-parameters?view=azure-devops&tabs=script

> Parameters are only available at template parsing time. Parameters are expanded just before the pipeline runs so that values surrounded by ${{ }} are replaced with parameter values. Use [variables](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops) if you need your values to be more widely available during your [pipeline run](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/runs?view=azure-devops).

jeffkl: Yes it works but only because we copy some parameters to variables: https://github.com/NuGet/NuGet.Client/blob/dev-jeffkl-move-static-analysis/eng/pipelines/pull_request.yml#L65

jeffkl: image

zivkan: Ah, in that case there's no benefit to adding the `RunStaticAnalysis` parameter to this pipeline.yml file, since it's an unused parameter.

zivkan: I'm not sure, but does `condition: "and(succeeded(), ne( '${{ parameter.RunStaticAnalysis }}', 'false')"` work?

It might be a way to avoid needing to declare a bunch of variables that are only used once?

jeffkl: At the moment I'm not too worried, I'm hoping in the near future to clean up PrivateDev and have it just build and get rid off all these other stages, jobs, parameters, and variables.  My primary concern is leaving things as-is so that if something is wrong we can bring back previous behavior.

zivkan (approved):

PR #5905, opened 2024-07-04
Fix race condition in EnhancedHttpRetryHelper

### LINKED ISSUE #13545: RestoreTask randomly fails after upgrading to latest version
### NuGet Product Used

dotnet.exe

### Product Version

6.11.0-rc.90

### Worked before?

NuGet in dotnet sdk 9.0-preview4

### Impact

### Repro Steps & Context

Since we updated to a dotnet 9.0-preview6 SDK we started seeing RestoreTask failing randomly in the VMR build: https://github.com/dotnet/sdk/issues/41477#issuecomment-2154989462

After the fix in https://github.com/NuGet/NuGet.Client/pull/5845 we bumped again (i.e. we're now using a very recent nuget.client) we now see these messages

> D:\a\_work\1\s\.dotnet\sdk\9.0.100-preview.6.24311.23\NuGet.targets(170,5): The feed 'dotnet-public [https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json]' lists package 'System.Text.Json.7.0.3' but multiple attempts to download the nupkg have failed. The feed is either invalid or required packages were removed while the current operation was in progress. Verify the package exists on the feed and try again. [D:\a\_work\1\s\src\nuget-client\build\restorehelper.targets]

or

> D:\a\_work\1\s\.dotnet\sdk\9.0.100-preview.6.24311.23\NuGet.targets(170,5): error : The feed 'dotnet-public [https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json]' lists package 'Castle.Core.5.1.1' but multiple attempts to download the nupkg have failed. The feed is either invalid or required packages were removed while the current operation was in progress. Verify the package exists on the feed and try again. [D:\a\_work\1\s\src\efcore\EFCore.sln]

I've attached some binlogs: [binlogs.zip](https://github.com/user-attachments/files/15821942/binlogs.zip)

It's happening for random packages and given we only started seeing this when bumping the dotnet SDK (and thus nuget) I think it's more likely a nuget bug than an issue with the AzDO feed.

Is there some way we can enable more logging?

/cc @nkolev92 @kartheekp-ms 

### Verbose Logs

_No response_

akoeplinger:
## Bug

Fixes: https://github.com/NuGet/Home/issues/13545

Regression? Last working version: 9.0-preview5

## Description

When multiple threads try to access the RetryCount property we sometimes hit a case where RetryCount returned 0 (*).
This caused the loop in `ProcessHttpSourceResultAsync()` to be skipped completely because `retry <= maxRetries`: https://github.com/NuGet/NuGet.Client/blob/6c642c2d63717acd4bf92050a42691928020eb89/src/NuGet.Core/NuGet.Protocol/Utility/FindPackagesByIdNupkgDownloader.cs#L258-L328

Fix this by using `Lazy<T>` for initializing the fields which is thread-safe.

(*) The reason is that code like `int RetryCount => _retryCount ??= 6;` gets turned into:

```csharp
int valueOrDefault = _retryCount.GetValueOrDefault();
if (!_retryCount.HasValue)
{
    valueOrDefault = 6;
    _retryCount = valueOrDefault;
    return valueOrDefault;
}
return valueOrDefault;
```

Suppose Thread A arrives first and calls `GetValueOrDefault()` (which is 0 for int) but then Thread B interjects and sets the value, now when Thread A resumes `_retryCount.HasValue` is true so we skip the if block and return valueOrDefault i.e. 0.

## PR Checklist

- [x] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [x] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/src/NuGet.Core/NuGet.Protocol/EnhancedHttpRetryHelper.cs
+++ b/src/NuGet.Core/NuGet.Protocol/EnhancedHttpRetryHelper.cs
@@ -70,17 +70,17 @@ namespace NuGet.Protocol
 
         private readonly IEnvironmentVariableReader _environmentVariableReader;
 
-        private bool? _isEnabled = null;
+        private Lazy<bool> _isEnabled;
 
-        private int? _retryCount = null;
+        private Lazy<int> _retryCount;
 
-        private int? _delayInMilliseconds = null;
+        private Lazy<int> _delayInMilliseconds;
 
-        private bool? _retry429 = null;
+        private Lazy<bool> _retry429;
 
-        private bool? _observeRetryAfter = null;
+        private Lazy<bool> _observeRetryAfter;
 
-        private TimeSpan? _maxRetyAfterDelay = null;
+        private Lazy<TimeSpan> _maxRetyAfterDelay;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="EnhancedHttpRetryHelper" /> class.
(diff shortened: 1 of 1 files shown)


donnie-msft (approved):
Thanks for the contribution! I'm running our tests pipeline now...

nkolev92 (approved):
🤝

src/NuGet.Core/NuGet.Protocol/EnhancedHttpRetryHelper.cs
```diff
@@ -90,46 +90,44 @@ internal class EnhancedHttpRetryHelper
...
         {
             _environmentVariableReader = environmentVariableReader ?? throw new ArgumentNullException(nameof(environmentVariableReader));
+            _isEnabled = new Lazy<bool>(() => GetBoolFromEnvironmentVariable(IsEnabledEnvironmentVariableName, defaultValue: DefaultEnabled, _environmentVariableReader));
+            _retryCount = new Lazy<int>(() => GetIntFromEnvironmentVariable(RetryCountEnvironmentVariableName, defaultValue: DefaultRetryCount, _environmentVariableReader));
+            _delayInMilliseconds = new Lazy<int>(() => GetIntFromEnvironmentVariable(DelayInMillisecondsEnvironmentVariableName, defaultValue: DefaultDelayMilliseconds, _environmentVariableReader));
+            _retry429 = new Lazy<bool>(() => GetBoolFromEnvironmentVariable(Retry429EnvironmentVariableName, defaultValue: true, _environmentVariableReader));
+            _observeRetryAfter = new Lazy<bool>(() => GetBoolFromEnvironmentVariable(ObserveRetryAfterEnvironmentVariableName, defaultValue: DefaultObserveRetryAfter, _environmentVariableReader));
+            _maxRetyAfterDelay = new Lazy<TimeSpan>(() =>
```
wuhkuh, line 98: `_maxRetyAfterDelay` -> `_maxRetryAfterDelay`

akoeplinger: thanks: https://github.com/NuGet/NuGet.Client/pull/6025

PR #5906, opened 2024-07-05
Update the debug helper scripts to import the props file

nkolev92:
## Bug

Fixes:

## Description

The scripts import the targets file but they were not setup to import the props file. This means things are slighlty inconsistent. Normally there's no discrepancies with the props files but https://github.com/NuGet/NuGet.Client/commit/7d8b62fcf454e5946977e8d4461d576b9b0e7e59 changed that. 

This just makes the scripts current.

## PR Checklist

- [ ] PR has a linked issue.
- [x] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [x] N/A 

- **Documentation**
  
  - **OR**
  - [x] N/A

### DIFF
--- a/scripts/nuget-debug-helpers.ps1
+++ b/scripts/nuget-debug-helpers.ps1
@@ -18,9 +18,10 @@ Function Invoke-NuGetCustom()
     $packTargetsPath = Join-Path $NuGetClientRoot "src\NuGet.Core\NuGet.Build.Tasks.Pack\NuGet.Build.Tasks.Pack.targets"
     $restoreDllPath = Join-Path $NuGetClientRoot "artifacts\NuGet.Build.Tasks\bin\$Configuration\$NETFramework\NuGet.Build.Tasks.dll"
     $nugetRestoreTargetsPath = Join-Path $NuGetClientRoot "src\NuGet.Core\NuGet.Build.Tasks\NuGet.targets"
+    $nugetPropsPath = Join-Path $NuGetClientRoot "src\NuGet.Core\NuGet.Build.Tasks\NuGet.props"
     $consoleExePath = Join-Path $NuGetClientRoot "artifacts\NuGet.Build.Tasks.Console\bin\$Configuration\$NETFramework\NuGet.Build.Tasks.Console.exe"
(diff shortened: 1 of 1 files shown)


scripts/nuget-debug-helpers.ps1
```diff
@@ -18,9 +18,10 @@ Function Invoke-NuGetCustom()
...
-    & msbuild /p:NuGetRestoreTargets=$nugetRestoreTargetsPath /p:RestoreTaskAssemblyFile=$restoreDllPath /p:NuGetBuildTasksPackTargets=$packTargetsPath /p:NuGetConsoleProcessFileName=$consoleExePath /p:ImportNuGetBuildTasksPackTargetsFromSdk=true /p:NuGetPackTaskAssemblyFile=$packDllPath $args[0..$args.Count]
+    Write-Host "msbuild /p:NuGetRestoreTargets=$nugetRestoreTargetsPath /p:NuGetPropsFile=$nugetPropsPath /p:RestoreTaskAssemblyFile=$restoreDllPath /p:NuGetBuildTasksPackTargets=$packTargetsPath /p:NuGetConsoleProcessFileName=$consoleExePath /p:ImportNuGetBuildTasksPackTargetsFromSdk=true /p:NuGetPackTaskAssemblyFile=$packDllPath $($args[0..$args.Count])"
+    & msbuild /p:NuGetRestoreTargets=$nugetRestoreTargetsPath /p:NuGetPropsFile=$nugetPropsPath /p:RestoreTaskAssemblyFile=$restoreDllPath /p:NuGetBuildTasksPackTargets=$packTargetsPath /p:NuGetConsoleProcessFileName=$consoleExePath /p:ImportNuGetBuildTasksPackTargetsFromSdk=true /p:NuGetPackTaskAssemblyFile=$packDllPath $args[0..$args.Count]
```
zivkan, line 24: This can be simplified. Firstly, NuGet.targets gets copied into the bin folder, so if you use that bin copy, `RestoreTaskAssemblyFile` can be removed, as it's already defined in the targets file. Second, if you change the NuGet.targets to use `NuGet.Build.Tasks.Console`, then you can also remove `NuGetConsoleProcessFileName`.

I'm less confident, but I expect that `NuGetPackTaskAssemblyFile` can be removed, because it should be imported correctly by the pack targets.

zivkan (approved):

scripts/nuget-debug-helpers.ps1
nkolev92: I had auto merge here. 
I'll look into your feedback. It sounds like it could make things simpler. 

I don't know how commonly this gets used fwiw. This is my preferred way of debugging restore, but I'm curious how often others use it.

PR #5911, opened 2024-07-10
Don't use obsolete X509Certificate2 constructor on net9.0

### LINKED ISSUE #13612: Use of Obsolete X509Certificate2 ctor
### NuGet Product Used

nuget-client in dotnet VMR latest build

### Repro Steps & Context

The source-build VMR build is failing in stage 2 with the following error:

`
/vmr/src/nuget-client/src/NuGet.Core/NuGet.Configuration/Settings/Items/FileClientCertItem.cs(170,66): error SYSLIB0057: 'X509Certificate2.X509Certificate2(string)' is obsolete: 'Loading certificate data through the constructor or Import is obsolete. Use X509CertificateLoader instead to load certificates.' (https://aka.ms/dotnet-warnings/SYSLIB0057) [/vmr/src/nuget-client/src/NuGet.Core/NuGet.Configuration/NuGet.Configuration.csproj]
`

This error is caused by https://github.com/dotnet/runtime/pull/104165
This issue is also described in https://github.com/dotnet/source-build/issues/4495.

akoeplinger:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13612

## Description

When building nuget.client in the VMR for source-build we're targeting net9.0. This means we get the new obsoletion of X509Certificate2 ctor: https://github.com/dotnet/docs/issues/41662.

This PR fixes all instances to use the recommended X509CertificateLoader replacement.

## PR Checklist

- [ ] Added tests


src/NuGet.Core/NuGet.Commands/SignCommand/CertificateProvider.cs
```diff
@@ -83,6 +86,9 @@ public static async Task<X509Certificate2Collection> GetCertificatesAsync(Certif
                                     options.CertificatePath)));
 
                         case CRYPT_E_NO_MATCH_HRESULT:
+#if NET9_0_OR_GREATER
+                        case CRYPT_E_BAD_DECODE:
+#endif
```
akoeplinger, line 91: @bartonjs @vcsjones FYI this was uncovered by the  [ExecuteCommandAsync_WithEmptyPkcs7File_RaisesErrorsOnceAsync](https://github.com/NuGet/NuGet.Client/blob/4a1faa2bcc771862a0ee1d241ebbfa1c196630ab/test/NuGet.Core.Tests/NuGet.Commands.Test/SignCommandRunnerTests.cs#L47) test, X509CertificateLoader returns a different hresult in this case on !Windows because we're now returning this from managed code.

Might be something we should call out in the breaking change doc.

bartonjs: HResult values from exceptions aren't guaranteed (especially when they can change because of an OS change); so I don't anticipate putting it in the breaking change documentation.

src/NuGet.Core/NuGet.Commands/SignCommand/CertificateProvider.cs
```diff
@@ -122,7 +128,12 @@ private static async Task<X509Certificate2> LoadCertificateFromFileAsync(Certifi
 
             if (!string.IsNullOrEmpty(options.CertificatePassword))
             {
-                cert = new X509Certificate2(options.CertificatePath, options.CertificatePassword); // use the password if the user provided it.
+                // use the password if the user provided it
+#if NET9_0_OR_GREATER
+                cert = X509CertificateLoader.LoadPkcs12FromFile(options.CertificatePath, options.CertificatePassword);
+#else
+                cert = new X509Certificate2(options.CertificatePath, options.CertificatePassword);
```
bartonjs, line 135: The preferred fix is to add a package reference to Microsoft.Bcl.Cryptography and always call X509CertificateLoader; but we don't have the power to mark things Obsolete in older TFMs.

If adding the package is unacceptable, then this works.  The main concern is "don't get surprised by a PFX", but this expects a PFX.  There are also some subtle differences in how PFX loading works on Windows between the two, mainly fixing issues with loading very old files or loading the same file in parallel (making Microsoft.Bcl.Cryptography/X509CertificateLoader the better choice)

src/NuGet.Core/NuGet.Commands/SignCommand/CertificateProvider.cs
akoeplinger: I'd agree with you in an ideal world, but this real world code makes me think we should at least give a heads up in the doc :)

akoeplinger:
bump

src/NuGet.Core/NuGet.Commands/SignCommand/CertificateProvider.cs
zivkan: Is `X509CertificateLoader` brand new, not even in .NET 9 Preview 6?  I had a quick look at the preview 6 package on nuget.org, but it doesn't appear to have the API.

For this reason, plus the fact that NuGet doesn't use the Arcade SDK, means I don't think NuGet can use this package for .NET 9.0.100 SDK.  Maybe after .NET 9 goes GA. For now, I think we need to take this conditional compilation change.

zivkan (dismissed):

bartonjs:
> Is X509CertificateLoader brand new, not even in .NET 9 Preview 6?

.NET 9 Preview 7, also available via the Microsoft.Bcl.Cryptography NuGet package (which won't work for .NET 9 Preview 1-6, since the package expects the type to be inbox for .NET 9)

zivkan (approved):

kartheekp-ms (approved):

PR #5913, opened 2024-07-10
Fix threading issue in OutputConsoleLogger

jeffkl:
# Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2528

## Description
The `OutputConsoleLoggerTests+Dispose.Disposes_error_list` test fails occasionally with the following exception:
```
System.ObjectDisposedException : Cannot access a disposed object.
Object name: 'Microsoft.VisualStudio.Threading.AsyncSemaphore'.
```

This change also adds synchronization to the `Dispose()` method by using an `int` and `Interlocked.CompareExchange()` to ensure that only one thread disposes of the underlying semaphore.

Before this change, I was not able to run the `OutputConsoleLoggerTests` over and over again with 100% success rate.  Now I can run them 100 times in a row and they pass every time.

## PR Checklist

- [ ] ~~Added tests~~ Existing tests provide sufficient coverage

### DIFF
--- a/src/NuGet.Clients/NuGet.VisualStudio.Common/OutputConsoleLogger.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Common/OutputConsoleLogger.cs
@@ -6,6 +6,7 @@ using System.ComponentModel.Composition;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Runtime.CompilerServices;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Threading;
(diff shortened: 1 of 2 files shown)


Nigusu-Allehu (dismissed):

src/NuGet.Clients/NuGet.VisualStudio.Common/OutputConsoleLogger.cs
```diff
@@ -31,8 +32,18 @@ public sealed class OutputConsoleLogger : INuGetUILogger, IDisposable
...
+        private readonly AsyncManualResetEvent _initialized = new AsyncManualResetEvent(initialState: false);
+
+        /// <summary>
+        /// Keeps track of whether or not Dispose() has been called.  If the value is zero, Dispose() has not been called.  If the value is not zero, Dispose() has been called.
+        /// </summary>
+        private int _isDisposed = 0;
+
         [SuppressMessage("Build", "CA2213:'OutputConsoleLogger' contains field '_semaphore' that is of IDisposable type 'ReentrantSemaphore', but it is never disposed. Change the Dispose method on 'OutputConsoleLogger' to call Close or Dispose on this field.", Justification = "Field is disposed from async task invoked from Dispose.")]
-        internal readonly ReentrantSemaphore _semaphore = ReentrantSemaphore.Create(1, NuGetUIThreadHelper.JoinableTaskFactory.Context, ReentrantSemaphore.ReentrancyMode.NotAllowed);
+        internal ReentrantSemaphore _semaphore = ReentrantSemaphore.Create(1, NuGetUIThreadHelper.JoinableTaskFactory.Context, ReentrantSemaphore.ReentrancyMode.NotAllowed);
```
nkolev92: Why remove the readonly?

src/NuGet.Clients/NuGet.VisualStudio.Common/OutputConsoleLogger.cs
```diff
@@ -89,6 +109,9 @@ public void End()
         {
             Run(async () =>
             {
+                // Wait for the async initialization in the constructor to complete
```
nkolev92: Can the initialization fail and cause this to hang?

nkolev92: I think to answer that, we'd probably need to understand what happens when it failed before your changes.

jeffkl: After @nkolev92 and I spoke offline, we decided that initialization is already synchronized so I reverted the use of the ManualResetEvent.  The change now simply uses `Interlocked.CompareExchange()` to ensure that only one thread disposes of the object and subsequent calls are a no-op.

zivkan (approved):

nkolev92 (approved):

donnie-msft (approved):

PR #5915, opened 2024-07-10
Http error: Error for non-https resource endpoints

### LINKED ISSUE #13364: Show an error when a non https source is used in a resource in a service index

Nigusu-Allehu:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13364

## Description

Here is a discussion on the implementation: https://github.com/NuGet/Client.Engineering/pull/2966
An https NuGet source could have a non https service endpoint.
```json
{
    "version": "3.0.0",
    "resources": [
        {
            "@id": "http://source",
        }]
}
```
This PR makes sure we throw an error for such endpoints unless the source is set to `AllowInsecureConnections` in the Nuget config file. In addition, it also takes SdkAnalysisLevel into consideration. I created a static class that allows restore to communicate with NuGet.Protocol whether it should throw the http error or not

### How?
* When any NuGet command requests for service endpoints, we use `GetServiceEntries` to get a list of service endpoints that are required for the specific command. 
* Before adding an entry to the list of entries, check whether it is an insecure http endpoint.
     * If it is insecure throw an exception.
* Look at the following test https://github.com/NuGet/NuGet.Client/pull/5915#discussion_r1674491329 to understand the logic
## PR Checklist

- [x] Added tests


test/NuGet.Core.Tests/NuGet.Protocol.Tests/ServiceIndexResourceV3Tests.cs
```diff
@@ -24,6 +25,52 @@ public void Constructor_InitializesProperties()
...
+            var resource = new ServiceIndexResourceV3(serviceIndex, DateTime.Now);
+
+            // Act & Assert
+            Assert.Throws<ArgumentException>(() => resource.GetServiceEntries("SearchQueryService"));
+            Assert.Throws<ArgumentException>(() => resource.GetServiceEntries("RegistrationsBaseUrl"));
+            Assert.Throws<ArgumentException>(() => resource.GetServiceEntries("LegacyGallery"));
+        }
+
+        [Fact]
+        public void GetServiceEntries_WithHttpResourceEndPointAndAllowInsecureConnections_Succeeds()
```
Nigusu-Allehu, line 142: This test shows where the Exceptions are thrown.

jgonz120 (dismissed):
I don't see any obvious problems but I'm new to the space.

src/NuGet.Core/NuGet.Protocol/Resources/ServiceIndexResourceV3.cs
```diff
@@ -105,7 +108,22 @@ private IReadOnlyList<ServiceIndexEntry> GetBestVersionMatchForType(NuGetVersion
...
-                return entries.Where(e => e.ClientVersion == bestMatch.ClientVersion).ToList();
+                List<ServiceIndexEntry> matchingEntries = new List<ServiceIndexEntry>();
+
+                foreach (var entry in entries)
+                {
+                    if (entry.ClientVersion == bestMatch.ClientVersion)
+                    {
+                        if (entry.Uri.Scheme == Uri.UriSchemeHttp && entry.Uri.Scheme != Uri.UriSchemeHttps && !_allowInsecureConnections)
+                        {
+                            throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_HttpServiceIndexUsage, entry.Uri, _sourceUri));
```
donnie-msft: Is this the best exception type? I understand it to be for method arguments. I see how it could be considered an argument, but I would think something like a ProtocolException or something else would be more accurate.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/ServiceIndexResourceV3Tests.cs
```diff
@@ -24,6 +25,52 @@ public void Constructor_InitializesProperties()
...
+            var legacyRec = resource.GetServiceEntries("LegacyGallery").FirstOrDefault().Uri.ToString();
+
+            // Assert
+            Assert.Equal(searchRec, "http://search/");
+            Assert.Equal(regRec, "http://reg/");
+            Assert.Equal(legacyRec, "http://legacy/");
+        }
+
+        [Fact]
+        public void GetServiceEntries_RequestsHttpsResourceInServiceIndexContainingOtherHttpResourcesWithoutAllowInsecureConnections_Succeeds()
```
donnie-msft, line 162: nit: could this name be shortened/simplified and also split with more underscores (_)? "Succeeds" is really "FindsSingleServiceEntry", or similar?

test/NuGet.Core.Tests/NuGet.Protocol.Tests/ServiceIndexResourceV3Tests.cs
```diff
@@ -24,6 +25,52 @@ public void Constructor_InitializesProperties()
             Assert.Equal("http://unit.test/b", resource.Entries[0].Uri.ToString());
         }
 
+        [Fact]
+        public void GetServiceEntries_WithResourceEndPoint_ThrowsException()
```
donnie-msft: The key is that insecureConnections is false and the sources are HTTP.

```suggestion
        public void GetServiceEntries_WithInsecureHttpResourceEndPoint_doNotAllowInsecureConnections_ThrowsException()
```

test/NuGet.Core.Tests/NuGet.Protocol.Tests/ServiceIndexResourceV3Tests.cs
```diff
@@ -24,6 +25,52 @@ public void Constructor_InitializesProperties()
...
+        {
+            // Arrange
+            var serviceIndex = CreateServiceIndexWithHttpResources();
+            var resource = new ServiceIndexResourceV3(serviceIndex, DateTime.Now);
+
+            // Act & Assert
+            Assert.Throws<ArgumentException>(() => resource.GetServiceEntries("SearchQueryService"));
```
donnie-msft: Could you check the exception message as well? I believe the serviceIndex.Uri will be missing from the exception message based on my other comment.

src/NuGet.Core/NuGet.Protocol/Resources/ServiceIndexResourceV3.cs
donnie-msft: `_sourceUri` could be null here, since it is a property that has to be set outside of the constructor. Should a different string be used for the message in that case?

src/NuGet.Core/NuGet.Protocol/Resources/ServiceIndexResourceV3.cs
```diff
@@ -21,6 +22,8 @@ public class ServiceIndexResourceV3 : INuGetResource
         private readonly DateTime _requestTime;
         private static readonly IReadOnlyList<ServiceIndexEntry> _emptyEntries = new List<ServiceIndexEntry>();
         private static readonly SemanticVersion _defaultVersion = new SemanticVersion(0, 0, 0);
+        internal string _sourceUri;
```
donnie-msft: Can this be made part of a new constructor. I'm not sure the property approach is ideal for this, since methods depend on this value. Perhaps it should just be an overload on GetServiceEntries? @zivkan would have good opinions on this.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetListPackageTests.cs
```diff
@@ -741,7 +739,38 @@ public async Task ListPackage_WithHttpSourceAndAllowInsecureConnections_WarnsCor
...
+
+        [PlatformFact(Platform.Windows)]
+        public async Task ListPackage_WithHttpSourceAndAllowInsecureConnectionsTrue_Succeeds()
+        {
+            // Arrange
+            using var pathContext = _fixture.CreateSimpleTestPathContext();
+            var emptyHttpCache = new Dictionary<string, string>
+                {
+                    { "NUGET_HTTP_CACHE_PATH", pathContext.HttpCacheFolder },
+                };
```
donnie-msft: nit, I think this is right indentation:

```suggestion
            {
                { "NUGET_HTTP_CACHE_PATH", pathContext.HttpCacheFolder },
            };
```

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetListPackageTests.cs
```diff
@@ -741,7 +739,38 @@ public async Task ListPackage_WithHttpSourceAndAllowInsecureConnections_WarnsCor
 
             using var mockServer = new FileSystemBackedV3MockServer(pathContext.PackageSource);
             mockServer.Start();
-            pathContext.Settings.AddSource("http-source", mockServer.ServiceIndexUri, allowInsecureConnections);
+            pathContext.Settings.AddSource("http-source", mockServer.ServiceIndexUri, allowInsecureConnectionsValue: "false");
+
+            _fixture.RunDotnetExpectSuccess(Directory.GetParent(projectA.ProjectPath).FullName, $"add package A --version 1.0.0", testOutputHelper: _testOutputHelper);
+
+            // Act & Assert
+            CommandRunnerResult listResult = _fixture.RunDotnetExpectFailure(Directory.GetParent(projectA.ProjectPath).FullName, $"list package --outdated", testOutputHelper: _testOutputHelper);
```
donnie-msft, line 748: Thoughts on checking `listResult.Errors` for your expected error? It could fail for other reasons.

donnie-msft (commented):
Messaged you offline as well about how sure we are about the runtime error experience here. Client does not error when adding a package source with an insecure service entry, so it may be jarring to receive this error at runtime. 

Checking for HTTP service entries when adding a package source would introduce overhead, though. 

Curious if you can share how this change came about, the tradeoffs discussed, etc.

src/NuGet.Core/NuGet.Protocol/Resources/ServiceIndexResourceV3.cs
```diff
@@ -105,7 +108,22 @@ private IReadOnlyList<ServiceIndexEntry> GetBestVersionMatchForType(NuGetVersion
...
             {
                 // Find all entries with the same version.
-                return entries.Where(e => e.ClientVersion == bestMatch.ClientVersion).ToList();
+                List<ServiceIndexEntry> matchingEntries = new List<ServiceIndexEntry>();
+
+                foreach (var entry in entries)
+                {
+                    if (entry.ClientVersion == bestMatch.ClientVersion)
+                    {
+                        if (entry.Uri.Scheme == Uri.UriSchemeHttp && entry.Uri.Scheme != Uri.UriSchemeHttps && !_allowInsecureConnections)
```
zivkan: I don't know if the deep link will work, since the comment is resolved in the other PR: https://github.com/NuGet/NuGet.Client/pull/5760#discussion_r1637292706

But there I suggested that rather than checking every single time the same resource is requested over and over, we can check once during initialization. But the trade off is that it means any service index entry that is unused, but pointing to an insecure HTTP URL, will prevent NuGet from working.

src/NuGet.Core/NuGet.Protocol/Resources/ServiceIndexResourceV3.cs
zivkan: Yes, I'd much prefer it to be part of the constructor. I don't love the idea of `allowInsecureConnections` being part of the `GetServiceEntries` method, because why would an app want to allow some resources to be HTTP, while enforcing others to be HTTPS?  From a public API design point of view, I think constructor makes much more sense.

Most of all, I really wish that interface and implementation was split, so that these resource implementations were all internal, and we could change freely. But that's not the code, or public APIs, that we inherited.  I think that an overloaded constructor with sensible defaults would be the best option, but the question is, if there's existing code calling the old constructor, what's a sensible default for `sourceUri`? There just isn't one, and I'm not sure the error message make sense when `sourceUri` is null.

It seems a shame to make a breaking API change just for an unlikely error message. But it's also very unlikely anyone is creating instances of this class directly, rather than through the "provider", or more likely, with `Repository.Factory.CreateV3`. So, it's probably low impact to break the API (constructor).

zivkan (commented):
I haven't yet made up my mind if I think the API design should be changed, or if it's acceptable given the constraints of the existing public APIs. So, I'm not approving, but I'm also not blocking. Maybe I'll form a stronger opinion with more time, or after reading more feedback on the PR

Nigusu-Allehu:
> Messaged you offline as well about how sure we are about the runtime error experience here. Client does not error when adding a package source with an insecure service entry, so it may be jarring to receive this error at runtime.
> 
> Checking for HTTP service entries when adding a package source would introduce overhead, though.
> 
> Curious if you can share how this change came about, the tradeoffs discussed, etc.

You are right, checking for HTTP service entries when adding a package source would have additional overhead. My implementation was designed to have a central validation point, allowing us to confirm that no HTTP resources are being used without customer consent. This approach avoids the need to implement checks on a command-by-command basis, which could lead to oversight and errors as new features are introduced. The current implementation is centralized, meaning every command that needs to make an API request will require a service entry check. Both NuGet and the dotnet CLI will log the exception as an error, providing visibility to the customer. This error logging offers a clear explanation of what went wrong. However, I am not certain about the best way to communicate these errors within Visual Studio.

src/NuGet.Core/NuGet.Protocol/Resources/ServiceIndexResourceV3.cs
Nigusu-Allehu: After reviewing the issue, I see no major problem to performing the check each time a resource is requested. The current implementation includes a loop that goes through each service entry to locate the desired version. The http check is integrated within this loop, thus the performance impact is minimal, maintaining an O(n) complexity both before and after the change. In addition, this approach doesn't restrict users from accessing a secure resource just because another listed resource may be insecure. It's possible that the source provider is in the process of upgrading their service entries to HTTPS. As long as the insecure resource isn't being utilized, I don't believe we should block

test/NuGet.Core.Tests/NuGet.Protocol.Tests/ServiceIndexResourceV3Tests.cs
```diff
@@ -24,6 +25,52 @@ public void Constructor_InitializesProperties()
...
+
+        [Fact]
+        public void GetServiceEntries_RequestsHttpsResourceInServiceIndexContainingOtherHttpResourcesWithoutAllowInsecureConnections_Succeeds()
+        {
+            // Arrange
+            var serviceIndex = CreateServiceIndexWithHttpResources();
+            var resource = new ServiceIndexResourceV3(serviceIndex, DateTime.Now);
+
+            // Act
+            var vulnRec = resource.GetServiceEntries("VulnerabilityInfo/6.7.0").FirstOrDefault().Uri.ToString();
```
aortiz-msft: This abbreviation is not clear. Is it Vulnerable Resource, Vulnerability Resource, something else? Could we just use unabbreviated words for the var?

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -277,6 +277,19 @@ await _logger.LogAsync(RestoreLogMessage.CreateWarning(NuGetLogCode.NU1803,
...
                         }
+                        else if (source.IsHttps && !source.AllowInsecureConnections)
+                        {
+                            var isErrorEnabled = SdkAnalysisLevelMinimums.IsEnabled(_request.Project.RestoreMetadata.SdkAnalysisLevel,
+                                _request.Project.RestoreMetadata.UsingMicrosoftNETSdk,
+                                SdkAnalysisLevelMinimums.HttpErrorSdkAnalysisLevelMinimumValue);
+
+                            // Enable new errors and warnings for the current SDK analysis level.
+                            Protocol.Utility.SdkAnalysisLevelSettings.EnableNewErrorsAndWarnings = isErrorEnabled;
+                            Protocol.Utility.HttpServiceEndpointLogger.LogDelegate = async (string message) =>
```
jeffkl: Using static singletons makes our tests unreliable since one test sets it and another changes it, then something fails randomly.  Is there any way to avoid this here?

src/NuGet.Core/NuGet.Protocol/Resources/ServiceIndexResourceV3.cs
```diff
@@ -109,7 +113,22 @@ private IReadOnlyList<ServiceIndexEntry> GetBestVersionMatchForType(NuGetVersion
             else
             {
                 // Find all entries with the same version.
-                return entries.Where(e => e.ClientVersion == bestMatch.ClientVersion).ToList();
+                List<ServiceIndexEntry> matchingEntries = new List<ServiceIndexEntry>();
```
jeffkl, line 116:
```suggestion
                List<ServiceIndexEntry> matchingEntries = new List<ServiceIndexEntry>(entries.Count);
```

src/NuGet.Core/NuGet.Protocol/Strings.resx
```diff
@@ -530,4 +530,9 @@ The "s" should be localized to the abbreviation for seconds.</comment>
 {1} - number of vulerabilitiy files reported
 {2} - max count allowed</comment>
   </data>
-</root>
+  <data name="Error_HttpServiceIndexUsage" xml:space="preserve">
+    <value>There is a non-HTTPS resource: '{0}', defined by your package source: '{1}'. NuGet requires HTTPS endpoints. To allow unencrypted HTTP, you must explicitly set 'allowInsecureConnections' to true in your NuGet.Config file, or talk to the administrator of your source and ask them to migrate to a secure HTTPS URL. Please refer to https://aka.ms/nuget-https-everywhere for more information.</value>
```
jeffkl, line 534:
```suggestion
    <value>There is a non-HTTPS resource, '{0}', defined by your package source: '{1}'. NuGet requires HTTPS endpoints. Update your package source to only use HTTPS resources. To allow unencrypted HTTP connections, you must explicitly set 'allowInsecureConnections' to true in your NuGet.Config file. Please refer to https://aka.ms/nuget-https-everywhere for more information.</value>
```

nkolev92:
I thought these changes were gonna wait on the telemetry first. 
In particular: https://github.com/NuGet/NuGet.Client/pull/6125.

Nigusu-Allehu:
Re opening.

PR #5917, opened 2024-07-12
Fix hang in Solution Explorer search through transitive package items of dependencies tree

### LINKED ISSUE #13619: Solution Explorer search can be broken by skipped dataflow updates
https://github.com/NuGet/NuGet.Client/pull/5508 reduced the number of tree updates by suppressing notifications when data is unchanged.

`AssetsFileDependenciesTreeSearchProvider` uses `GetLatestVersionAsync` to obtain the latest snapshot value, relative to the current data versions of the project. However suppression of updates meant that the latest version might not be present, and the call would hang indefinitely. This would break Solution Explorer search for customers who enabled "Search External Items", have SDK-style .NET projects in the solution, and make certain project changes that lead to so-called "version-only" updates.

drewnoakes:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13619
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2079818

## Description

#5508 reduced the number of dataflow updates through code that led to redundant tree updates. It did that by inhibiting updates in cases where nothing changed.

`AssetsFileDependenciesTreeSearchProvider` uses `GetLatestVersionAsync` to obtain the latest snapshot value, relative to the current data versions of the project. However suppression of updates meant that the latest version might not be present, and the call would hang indefinitely. This would break Solution Explorer search for customers who enabled "Search External Items", have SDK-style .NET projects in the solution, and make certain project changes that lead to so-called "version-only" updates.

This PR reverts #5508 and adds a downstream check for unchanged data, skipping tree updates in such cases. This means that the latest data is always available for the search component to consume, meaning there's no hang.

## PR Checklist

- [ ] Added tests

### DIFF
--- a/src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/AssetsFileDependenciesDataSource.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/AssetsFileDependenciesDataSource.cs
@@ -4,7 +4,6 @@
 #nullable enable
 
 using System;
-using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.IO;
 using System.Threading.Tasks.Dataflow;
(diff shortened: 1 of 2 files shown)


nkolev92 (approved):

zivkan (approved):

kartheekp-ms:
> This PR reverts https://github.com/NuGet/NuGet.Client/pull/5508 and adds a downstream check for unchanged data, skipping tree updates in such cases.

I interpret this comment as the perf issue https://github.com/NuGet/Home/issues/13019 will not regress even though we reverted #5508, correct?

drewnoakes:
That's right. With this change the update propagates one level further down the dataflow graph, but the code that updates the tree still avoids doing the redundant work. The fix here is that another branch of the graph (search) needs to see every version of the update, even when the snapshot is unchanged.

PR #5918, opened 2024-07-12
NuGetAuditSuppress for packages.config in VS

### LINKED ISSUE #13575: NuGetAuditSuppress for packages.config
### NuGet Product(s) Involved

NuGet.exe, MSBuild.exe

### The Elevator Pitch

NuGetAuditSuppress has already been added for projects using PackageReference. Add support for packages.config as well.

_No response_

zivkan:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13575

## Description

NuGetAuditSuppress for packages.config CLI scenarios was recently added. This PR makes it work in Visual Studio as well.

## PR Checklist

- [x] Added tests
- [x] docs: feature already documented


src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/ProjectServices/VsManagedLanguagesProjectSystemServices.cs
```diff
@@ -340,6 +340,34 @@ private bool IsCentralPackageManagementVersionsEnabled()
...
+
+            List<(string, string[])> result = items is ICollection<(string, string[])> itemCollection ? new(itemCollection.Count) : new();
+
+            do
+            {
+                result.Add(enumerator.Current);
+            } while (enumerator.MoveNext());
+
+            return result;
+
```
Nigusu-Allehu: nit: white space after return

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/ProjectServices/CpsProjectSystemReferenceReader.cs
```diff
@@ -131,5 +131,11 @@ private static async Task<IBuildDependencyProjectReferencesService> GetProjectRe
         {
             throw new NotSupportedException();
         }
+
+        public async Task<IReadOnlyList<(string id, string[] metadata)>> GetItemsAsync(string itemTypeName, params string[] metadataNames)
```
nkolev92, line 135: Does CPS implement that API? 
I remember it being added to the old project system when we wanted CPM to work.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/ProjectServices/NativeProjectSystemReferencesReader.cs
```diff
@@ -108,5 +108,11 @@ internal class NativeProjectSystemReferencesReader
         {
             throw new NotSupportedException();
         }
+
+        public async Task<IReadOnlyList<(string id, string[] metadata)>> GetItemsAsync(string itemTypeName, params string[] metadataNames)
+        {
```
nkolev92, line 113: Same here, do we know for sure this API is implemented?

src/NuGet.Clients/NuGet.SolutionRestoreManager/SolutionRestoreJob.cs
```diff
@@ -716,7 +716,10 @@ private bool AnyProjectHasVulnerablePackageWarning(IReadOnlyList<RestoreSummary>
                     {
                         using SourceCacheContext sourceCacheContext = new();
                         List<SourceRepository> sourceRepositories = _sourceRepositoryProvider.GetRepositories().AsList();
+
+                        await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(token);
```
nkolev92: Would it be cleaner to have this be an async method which switches to the UI thread?
That way you don't have to remember to switch back. 

Another optimization you can do is just make sure that there are packages.config projects before  switching to the UI thread for the remainder of the operation.

src/NuGet.Core/NuGet.PackageManagement/IDE/PackageRestoreManager.cs
```diff
@@ -214,6 +217,23 @@ public IEnumerable<PackageRestoreData> GetPackagesRestoreData(string solutionDir
             }
 
             return restoreAuditProperties;
+
+            async Task<HashSet<string>> GetSuppressionsAsync(MSBuildNuGetProject msbuildProject)
```
nkolev92, line 221: PackageRestoreManager and SolutionRestoreJob have different approaches to getting the suppressions.

PackageRestoreManager switches back and forth for every project, but SolutionRestoreJob switches once. 
Was that intentional?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/ProjectServices/CpsProjectSystemReferenceReader.cs
zivkan: I don't know. There's no code path that calls `GetItemsAsync` currently on `CpsProjectSystemReferenceReader`. It's just needed because it implements the same interface as I need for packages.config (so, software design bug).

I thought it might be better to call the same implementation as the other readers, because maybe it works, rather than throw a `NotImplementedException`.  But I can be persuaded to throw.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/ProjectServices/NativeProjectSystemReferencesReader.cs
zivkan: When debugging suppressions for a native c++ project with packages.config, it defintaely went into this codepath, and I saw in the debugger the suppressions were there. I don't know of any packages for native projects with vulnerabilities, so it's not easy to manually test. I don't remember if we have any Apex or VS E2E tests for native C++ projects. We could potentially add one, but NuGet doesn't really have huge usage in the C++ ecosystem, and it's even less likely that a native library with a known vulnerability will have an advisory on a nuget package.

src/NuGet.Core/NuGet.PackageManagement/IDE/PackageRestoreManager.cs
zivkan: Kind of, but not really. It's more about fitting within the existing software architecture.

SolutionRestoreJob is JTF aware, since the assembly is only used in our VS extension. PackageRestoreManager is in NuGet.PackageManagement, which is ILMerged into nuget.exe, and is not JTF aware. It's also not an abstract class that I could add an abstract method onto and have NuGet.PackageManagement.VisualStudio extend with JTF stuff.

Without a fairly significant refactor (plus new and/or breaking API changes in NuGet.PackageManagement), which honestly I didn't even consider/think about before creating this PR, I don't see how it would be feasible to have PackageRestoreManager avoid spamming VS's UI thread with a lot of small work. We absolutely don't want to restore packages on VS's main thread.

The only other idea which I just came up with, is to copy the pattern that we do for a couple of properties. When we create the VS class that extends NuGetProject, currently it gets a few properties and caches them in the class, so that when NuGet.PackageManagement gets the properties, it doesn't need the UI thread.  We could do the same here?  But it'd still need an API on a type that PackageReference projects implement, so they'd either need to throw InvalidOperationExceptions, or switch to the UI thread for work that no production code path uses.  The fact that NuGet.PackageManagement doesn't have different NuGetProjects for PackageReference and PackagesConfig is harming the good design on this feature.

nkolev92: > The fact that NuGet.PackageManagement doesn't have different NuGetProjects for PackageReference and PackagesConfig is harming the good design on this feature.

It does, https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.PackageManagement/Projects/BuildIntegratedNuGetProject.cs

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/ProjectServices/NativeProjectSystemReferencesReader.cs
nkolev92: > but NuGet doesn't really have huge usage in the C++ ecosystem

I think my concern was more on the side of ensuring we're not crashing the restore.

We have other types where we  ensure that the project system implements the API before we call it and that's not really being done here.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/ProjectServices/CpsProjectSystemReferenceReader.cs
nkolev92: If it's not implemented, would it crash when GetItems is called?
Unless I misread the code there.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/ProjectServices/NativeProjectSystemReferencesReader.cs
zivkan: `VsProjectAdapter.GetBuildItemInformation` returns an empty result if the project system doesn't implement `IVsBuildItemStorage`

https://github.com/NuGet/NuGet.Client/blob/46f80a3f1a83f3c5f33a20fa9fb72871e1e2d4eb/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/VsProjectAdapter.cs#L228-L229

so, no crash.

src/NuGet.Clients/NuGet.SolutionRestoreManager/SolutionRestoreJob.cs
zivkan: > Would it be cleaner to have this be an async method which switches to the UI thread?

Depends on your definition of cleaner 🤣 
I pushed a change where I made the method async, and therefore it allocates a state machine on the heap, but it doesn't need to explicitly switch back the thread pool.

> Another optimization you can do is just make sure that there are packages.config projects before switching to the UI thread for the remainder of the operation.

I spent some time looking into this, and SolutionResotreJob already does this check before calling into this method: 
https://github.com/NuGet/NuGet.Client/blob/4e9c37b88a804843dbe27dcbeb442141192e1b35/src/NuGet.Clients/NuGet.SolutionRestoreManager/SolutionRestoreJob.cs#L237-L239

To me, a lot of this class's code looks like it was written for packages.config, and then had minimial touches to add support for PackageReference.  At least, there's a bunch of code that I would refactor if I had time to do so. So, I think the code could certainly be improved to be easier to reason about, but I don't think there's a perf optimization here.

src/NuGet.Core/NuGet.PackageManagement/IDE/PackageRestoreManager.cs
zivkan: Well, something weird is going on, because `BuildIntegratedNuGetProject` doesn't extend, `MSBuildNuGetProject`, but `MSBuildNuGetProject` implements `IDependencyGraphProject` and therefore has `GetPackageSpecsAsync` and `GetPackageSpecsAndAdditionalMessagesAsync` methods. Is this due to project.json, despite project.json using xproj and not csproj (excluding UWP)?  edit: no, ProjectJsonNuGetProject extensions BuildIntegratedNuGetProject.  Maybe MSBuildNuGetProject implements package specs just because of the msbuild restore ingetration?

Anyway, point remains that the reason SolutionRestoreJob and PackageRestoreManager work a bit differently is due to NuGet's software architecture. Restore seems to go through PackageRestoreManager, and therefore has "bad" performance due to frequent thread switching, twice (once to UI thread, once to thread pool) per project.  In VS, when packages.config restore no-ops, but we want to do the audit check, then we're in SolutionRestoreJob, which is JTF aware, and we can be smarter about switching to the UI thread once.

NuGet.PackageManagement isn't JTF aware, probably because it's used by nuget.exe, so if we want to avoid the thread switching during restore, I think there are two options:

1. add an interface into NuGet.PackageManagement that "abstracts" JTF, and does nothing in outside of VS
2. Make items in MSBuildNugetProject work like properties, where we load them at project load (VsSolutionManager initialization), and then use the cached version without need for the UI thread. I don't think that currently non-SDK style projects have any way to modify NuGetAuditSuppress items without causing a project reload, but if there is, then nuGet will have a bug since it will use stale data.

Unfortunately this is one of those cases where there isn't a clear best solution, at least in my opinion.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/ProjectServices/CpsProjectSystemReferenceReader.cs
zivkan: same response as https://github.com/NuGet/NuGet.Client/pull/5918#discussion_r1692343413

zivkan: Given my misunderstanding about MSBuildNuGetProject (not) being packages.config only, if I move the `GetItemsAsync` method out of `IProjectSystemReferenceReader` and into `MSBuildNuGetProject`, then CPS won't need to implement it, but CLI restore will (it can noop/throw, since it won't be called in code path).

It's a question of what we think the best design is, as every alternative I can see as feasible has drawbacks.

src/NuGet.Clients/NuGet.SolutionRestoreManager/SolutionRestoreJob.cs
nkolev92: I think SolutionRestoreJob and all the schedulers around that came along with PackageReference. The NuGet.SolutionRestoreManager was added for perf purposes.

nkolev92 (approved):

Nigusu-Allehu (approved):

PR #5921, opened 2024-07-14
Duplicate NuGetAuditSuppress warnings use code NU1508

### LINKED ISSUE #13620: Add a log code NuGetAuditSuppress duplicate items
Duplicate item checking is supposed to raise warnings when duplicates are seen. 

When NuGetAuditSuppress was added, NU1505 was added as a warning for it. 
This should be a new log code. 

cc @zivkan

zivkan:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13620

## PR Checklist

- [x] Added tests
- [x] Link to an issue or pull request to update docs if this PR changes settings, environment variables, new feature, etc. https://github.com/NuGet/Home/issues/13621

### DIFF
--- a/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
+++ b/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
@@ -359,7 +359,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     <CheckForDuplicateNuGetItemsTask
       Items="@(NuGetAuditSuppress)"
       ItemName="NuGetAuditSuppress"
-      LogCode="NU1505"
+      LogCode="NU1508"
       MSBuildProjectFullPath="$(MSBuildProjectFullPath)"
       TreatWarningsAsErrors="$(TreatWarningsAsErrors)"
       WarningsAsErrors="$(WarningsAsErrors)"
--- a/src/NuGet.Core/NuGet.Common/Errors/NuGetLogCode.cs
+++ b/src/NuGet.Core/NuGet.Common/Errors/NuGetLogCode.cs
@@ -288,6 +288,11 @@ namespace NuGet.Common
         NU1507 = 1507,
 
         /// <summary>
+        /// Duplicate NuGetAuditSuppress found
+        /// </summary>
+        NU1508 = 1508,
+
+        /// <summary>
         /// Dependency bumped up
         /// </summary>
         NU1601 = 1601,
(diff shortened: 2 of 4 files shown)


src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
zivkan, line 1: @nkolev92 why does NuGet.exe disable duplicate checking?

https://github.com/NuGet/NuGet.Client/blob/128a5066b1438627ac69a2ffe9de564b2c09ee4d/src/NuGet.Clients/NuGet.CommandLine/MsBuildUtility.cs#L265

Nigusu-Allehu (dismissed):

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
```diff
@@ -359,7 +359,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     <CheckForDuplicateNuGetItemsTask
       Items="@(NuGetAuditSuppress)"
       ItemName="NuGetAuditSuppress"
-      LogCode="NU1505"
```
zivkan, line 362: @nkolev92 from https://github.com/NuGet/Home/issues/13620

> This should be a new log code.

Why do we use a different code for each MSBuild item type?

The docs are the same for each item type, since the resolution is the same. I feel like all this design decision does is cause us extra busy work.  I feel like one error code, where in the docs we mention that this can happen for multiple item types, would just be easier, with no downside.

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
nkolev92: Because TreatWarningsAsErrors would fail the dg spec generation and fail the operation with a message that's not very actionable. 
It wasn't worth the effort back then.

Spec has more details: https://github.com/NuGet/Home/blob/dev/accepted/2022/duplicate-nuget-item-error-handling.md#technical-explanation.

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
nkolev92: The owner of the action isn't always the same. 
In particular the SDK sets PackageDownload items, so any issues are likely problems on the SDK side. 

For example, if you search with NU1505, you end up with https://github.com/dotnet/sdk/issues/24747.
NU1504 gives you something else. 

It's duplicative sure, but I think there's an eventual limit to the number of items we'll be using :)

nkolev92 (dismissed):

nkolev92 (approved):

PR #5922, opened 2024-07-15
Allow passing assets file directly to dotnet nuget why

### LINKED ISSUE #13576: Enable `dotnet nuget why` on non-SDK style projects
### NuGet Product(s) Involved

dotnet.exe

### The Elevator Pitch

`dotnet nuget why` currently requires you to point to a project file, but the .NET SDK can only evaluate SDK style projects. Many non-SDK style Visual Studio projects try to import MSBuild props/targets files that don't exist in the .NET SDK and therefore trying to read project properties (to find the assets file) fails.

`dotnet nuget why` is particularly useful for investigating why a package with a known vulnerability is used by a project, so allowing the command to work for non-SDK style projects will help customers investigate known vulnerabilities.

`dotnet nuget why` should either allow a `project.assets.json` file to be passed as an argument instead of a project file.

We can instruct customers that they can find the assets file location using `msbuild.exe path\to\project.csproj -getProperty:ProjectAssetsFile`

_No response_

zivkan:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13576

## Description

Refactored WhyCommandRunner so the "main loop" will read the assets file, without evaluating the project via MSBuild. Then modified the code that finds all projects in a solution to handle a json file like a single project solution.

## PR Checklist

- [x] Added tests
- [x] Link to an issue or pull request to update docs if this PR changes settings, environment variables, new feature, etc: https://github.com/NuGet/Home/issues/13623

### DIFF
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
@@ -118,23 +118,13 @@ namespace NuGet.CommandLine.XPlat
                         }
 
                         string? assetsFilePath = project.GetPropertyValue(ProjectAssetsFile);
-                        if (string.IsNullOrEmpty(assetsFilePath))
-                        {
-                            logger.LogError(
-                                string.Format(
-                                    CultureInfo.CurrentCulture,
-                                    Strings.Error_ProjectAssetsFilePropertyNotFound,
-                                    project.FullPath));
-                            continue;
-                        }
-
-                        if (!File.Exists(assetsFilePath))
+                        if (string.IsNullOrEmpty(assetsFilePath) || !File.Exists(assetsFilePath))
                         {
                             logger.LogError(
                                 string.Format(
                                     CultureInfo.CurrentCulture,
                                     Strings.Error_AssetsFileNotFound,
-                                    assetsFilePath));
+                                    project.FullPath));
                             continue;
                         }
 
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
@@ -91,7 +91,7 @@ namespace NuGet.CommandLine.XPlat
         private static IEnumerable<(string assetsFilepath, string? projectPath)> FindAssetsFiles(string path, ILoggerWithColor logger)
         {
             var extension = Path.GetExtension(path);
-            if (string.Equals(".json", extension, StringComparison.OrdinalIgnoreCase))
+            if (XPlatUtility.IsJsonFile(extension))
             {
                 yield return (path, null);
                 yield break;
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
@@ -88,10 +88,9 @@ namespace NuGet.CommandLine.XPlat
             return anyErrors ? ExitCodes.Error : ExitCodes.Success;
         }
 
-        private static IEnumerable<(string assetsFilepath, string? projectPath)> FindAssetsFiles(string path, ILoggerWithColor logger)
+        private static IEnumerable<(string assetsFilePath, string? projectPath)> FindAssetsFiles(string path, ILoggerWithColor logger)
         {
-            var extension = Path.GetExtension(path);
-            if (XPlatUtility.IsJsonFile(extension))
+            if (XPlatUtility.IsJsonFile(path))
             {
                 yield return (path, null);
                 yield break;
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
@@ -30,10 +30,10 @@ namespace NuGet.CommandLine.XPlat
             }
 
             string targetPackage = whyCommandArgs.Package;
-            IEnumerable<string> projectPaths;
+            IEnumerable<(string assetsFilePath, string? projectPath)> assetsFiles;
             try
             {
-                projectPaths = MSBuildAPIUtility.GetListOfProjectsFromPathArgument(whyCommandArgs.Path);
+                assetsFiles = GetAssetsFiles(whyCommandArgs.Path, whyCommandArgs.Logger);
             }
             catch (ArgumentException ex)
             {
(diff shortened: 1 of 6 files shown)


src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -940,15 +940,23 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
     <value>Missing or invalid path '{0}'. Please provide a path to a project, solution file, or directory.</value>
     <comment>{0} - Input path argument</comment>
   </data>
+  <data name="Error_PathIsMissingOrInvalid_AllowAssetsFile" xml:space="preserve">
+    <value>Missing or invalid path '{0}'. Please provide a path to a project, solution, or NuGet assets file, or directory.</value>
```
zivkan: I'm reconsidering this message. The .NET SDK isn't designed to work with non-SDK style projects, and I think it's going to be a very uncommon scenario, so I don't think it's worthwhile complicating the message with this.  I'm also thinking of the poor translators who need to figure out how to translate this 😆 

Any feedback?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -46,57 +46,110 @@ public static int ExecuteCommand(WhyCommandArgs whyCommandArgs)
                 return ExitCodes.InvalidArguments;
             }
 
-            foreach (var projectPath in projectPaths)
+            bool anyErrors = false;
```
Nigusu-Allehu, line 49: nit: do we need this? Return `ExitCodes.Error` if asserFile is null, `ExitCodes.Success` otherwise

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -46,57 +46,110 @@ public static int ExecuteCommand(WhyCommandArgs whyCommandArgs)
...
+                            continue;
                         }
-                        else
+
+                        string assetsFilePath = project.GetPropertyValue(ProjectAssetsFile);
+                        if (string.IsNullOrEmpty(assetsFilePath))
                         {
-                            whyCommandArgs.Logger.LogMinimal(
+                            // TODO: show error message
+                            continue;
```
Nigusu-Allehu: Are we adding the error in this PR or is it a follow up?

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -184,51 +237,49 @@ private static void ValidateFrameworksOptionsExistInAssetsFile(LockFile assetsFi
...
-
-            if (!File.Exists(assetsPath))
+            if (!File.Exists(assetsFilePath))
             {
                 logger.LogError(
                     string.Format(
                         CultureInfo.CurrentCulture,
                         Strings.Error_AssetsFileNotFound,
-                        project.FullPath));
+                        assetsFilePath));
```
Nigusu-Allehu: If we are passing assersFilePath as an argument to the error instead of project.FullPath, we should update the `Error_AssersFileNotFound` to follow suit as well.

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
nkolev92: I agree with your reconsideration. 

Make the happy path .NET SDK specific.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
zivkan: Perhaps I misunderstand, but in a solution with multiple projects, we don't want to stop at the first project with an error. We can continue outputting for other projects, but the exit code should still indicate an error when at least one project had an error.

kartheekp-ms:
Once this PR has been merged, please consider updating https://github.com/NuGet/Home/blob/dev/accepted/2022/dotnet-nuget-why-proposal.md spec with the new behavior.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -30,10 +30,10 @@ public static int ExecuteCommand(WhyCommandArgs whyCommandArgs)
             }
 
             string targetPackage = whyCommandArgs.Package;
-            IEnumerable<string> projectPaths;
+            IEnumerable<(string assetsFilePath, string? projectPath)> assetsFiles;
```
kartheekp-ms, line 33: I think we should also update the CLI argument [here](https://github.com/NuGet/NuGet.Client/blob/38ab39a6691e6e5d54e29055d9e258b8b5d57220/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommand.cs#L13-L17) and [here](https://github.com/NuGet/NuGet.Client/blob/38ab39a6691e6e5d54e29055d9e258b8b5d57220/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs#L109-L113) to include assets file path.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -46,57 +46,114 @@ public static int ExecuteCommand(WhyCommandArgs whyCommandArgs)
...
 
-                        if (dependencyGraphPerFramework != null)
+        private static IEnumerable<(string assetsFilepath, string? projectPath)> FindAssetsFiles(string path, ILoggerWithColor logger)
+        {
+            var extension = Path.GetExtension(path);
+            if (string.Equals(".json", extension, StringComparison.OrdinalIgnoreCase))
+            {
+                yield return (path, null);
+                yield break;
+            }
```
kartheekp-ms:
```suggestion
            if (XPlatUtility.IsJsonFile(path))
            {
                yield return (path, null);
                yield break;
            }
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -947,8 +947,12 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
   <data name="WhyCommand_Message_NonSDKStyleProjectsAreNotSupported" xml:space="preserve">
     <value>Unable to run 'dotnet nuget why' for project '{0}'. This command only works on SDK-style projects.</value>
```
kartheekp-ms: Should we update the message here so that customers are aware that to use why command on non-sdk style projects they can pass assets file path? https://github.com/NuGet/NuGet.Client/blob/38ab39a6691e6e5d54e29055d9e258b8b5d57220/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs#L90-L93

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
zivkan: Nikolche and I disagree: https://github.com/NuGet/NuGet.Client/pull/5922#discussion_r1677586726

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -46,57 +46,114 @@ public static int ExecuteCommand(WhyCommandArgs whyCommandArgs)
...
+                                    Strings.Error_NotPRProject,
+                                    project.FullPath));
 
-                            DependencyGraphPrinter.PrintAllDependencyGraphs(dependencyGraphPerFramework, targetPackage, whyCommandArgs.Logger);
+                            continue;
                         }
-                        else
+
+                        string? assetsFilePath = project.GetPropertyValue(ProjectAssetsFile);
+                        if (string.IsNullOrEmpty(assetsFilePath))
```
zivkan: Test `NuGet.XPlat.FuncTest.XPlatWhyTests.WhyCommand_ProjectDidNotRunRestore_Fails` is failing because the wrong message is being output. Debugging, this property does not have a value, but for the life of me, I can't figure out why.

If I go to the command line, `dotnet msbuild -getProperty:ProjectAssetsFile -bl` also returns an empty string (unless I accidentally restore first, but delete the `obj` directory and it's empty again), but `dotnet build --no-restore -bl` shows in the binlog that the property is set with the correct value. If I delete the csproj that the test creates and use `dotnet new classlib`, then this weied behaviour doesn't happen. But looking at the binlogs, I just can't figure out why our test project behaves like this.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
zivkan: Good find!

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
zivkan: Good find!

nkolev92 (dismissed):

Nigusu-Allehu (dismissed):

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
kartheekp-ms: Thanks, Andy, for providing clarification offline.

nkolev92 (dismissed):

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
```diff
@@ -46,57 +46,104 @@ public static int ExecuteCommand(WhyCommandArgs whyCommandArgs)
...
+            }
 
-                        if (dependencyGraphPerFramework != null)
+            return anyErrors ? ExitCodes.Error : ExitCodes.Success;
+        }
+
+        private static IEnumerable<(string assetsFilepath, string? projectPath)> FindAssetsFiles(string path, ILoggerWithColor logger)
+        {
+            var extension = Path.GetExtension(path);
+            if (XPlatUtility.IsJsonFile(extension))
```
kartheekp-ms: `XPlatUtility.IsJsonFile()` utility method does invoke `Path.GetExtension()` to check if the file is a json file or not. I think we can directly pass the path to the utility method.

```suggestion
            if (XPlatUtility.IsJsonFile(path))
```

src/NuGet.Core/NuGet.CommandLine.XPlat/Strings.resx
```diff
@@ -942,10 +942,14 @@ Non-HTTPS access will be removed in a future version. Consider migrating to 'HTT
     <comment>{0} - Exception mssage</comment>
   </data>
   <data name="WhyCommand_Message_NonSDKStyleProjectsAreNotSupported" xml:space="preserve">
-    <value>Unable to run 'dotnet nuget why' for project '{0}'. This command only works on SDK-style projects.</value>
+    <value>Unable to run 'dotnet nuget why' for project '{0}'. See https://aka.ms/dotnet/nuget/why#older-project-format</value>
```
kartheekp-ms, line 945:
```suggestion
    <value>Unable to run 'dotnet nuget why' for project '{0}'. See https://aka.ms/dotnet-nuget-why#older-project-format</value>
```

https://aka.ms/dotnet/nuget/why returns 404 whereas https://aka.ms/dotnet-nuget-why successfully navigates to dotnet nuget why docs.

There is another occurrence of the same incorrect URL in the below entry.

zivkan: I talked to the dotnet cli team and asked them what their prefered naming convention was, and they prefer slashes.  I just deleted https://aka.ms/dotnet-nuget-why and created https://aka.ms/dotnet/nuget/why.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/WhyCommand/WhyCommandRunner.cs
zivkan: Retrying `dotnet new classlib` `dotnet msbuild -getProperty:ProjectAssetsFile` returns an empty string, so I think it is an issue with the SDK props/targets.  So, I changed the warning message to match the "you need to restore this project" message.

nkolev92 (approved):

PR #5923, opened 2024-07-15
Map SourceBranchName from sourcelink to RepositoryBranch for NuGet pack

### LINKED ISSUE #13625: Map branch name from sourcelink to RepositoryBranch for NuGet pack
### NuGet Product(s) Involved

NuGet.exe, dotnet.exe

### The Elevator Pitch

This is a companion issue for https://github.com/dotnet/sourcelink/issues/1243.

Once sourcelink emits the source branch name, NuGet should map that value to `RepositoryBranch` (and follow the existing rules around allowing the user to override).

_No response_

MattKotsenas:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13625

## Description

This is a companion PR for https://github.com/dotnet/sourcelink/pull/1248. Once that change lands and sourcelink is emitting the source branch name, this PR consumes that value and sets it to `RepositoryBranch` as part of NuGet pack.

## PR Checklist

- [X] Added tests

### DIFF
--- a/test/NuGet.Core.FuncTests/Dotnet.Integration.Test/PackCommandTests.cs
+++ b/test/NuGet.Core.FuncTests/Dotnet.Integration.Test/PackCommandTests.cs
@@ -3667,6 +3667,7 @@ namespace ClassLibrary
         <PropertyGroup>
             <SourceRevisionId>e1c65e4524cd70ee6e22abe33e6cb6ec73938cb1</SourceRevisionId>
             <PrivateRepositoryUrl>https://github.com/NuGet/NuGet.Client.git</PrivateRepositoryUrl>
+            <SourceBranchName>refs/heads/main</SourceBranchName>
         </PropertyGroup>
     </Target>
 </Project>");
@@ -3787,6 +3788,7 @@ namespace ClassLibrary
     <PropertyGroup>
       <SourceRevisionId>e1c65e4524cd70ee6e22abe33e6cb6ec73938cb1</SourceRevisionId>
       <PrivateRepositoryUrl>https://github.com/NuGet/NuGet.Client.git</PrivateRepositoryUrl>
+      <SourceBranchName>refs/heads/main</SourceBranchName>
     </PropertyGroup>
 </Target>
 </Project>";
@@ -3809,7 +3811,7 @@ namespace ClassLibrary
                     var repositoryMetadata = nuspecReader.GetRepositoryMetadata();
                     repositoryMetadata.Type.Should().Be("git");
                     repositoryMetadata.Url.Should().Be("https://github.com/NuGet/NuGet.Client.git");
-                    repositoryMetadata.Branch.Should().Be("");
+                    repositoryMetadata.Branch.Should().Be("refs/heads/main");
                     repositoryMetadata.Commit.Should().Be("e1c65e4524cd70ee6e22abe33e6cb6ec73938cb1");
                 }
             }
@@ -3836,6 +3838,7 @@ namespace ClassLibrary
                     ProjectFileUtils.AddProperty(xml, "PublishRepositoryUrl", "true");
                     ProjectFileUtils.AddProperty(xml, "RepositoryCommit", "1111111111111111111111111111111111111111");
                     ProjectFileUtils.AddProperty(xml, "RepositoryUrl", "https://github.com/Overridden");
+                    ProjectFileUtils.AddProperty(xml, "RepositoryBranch", "refs/heads/overwritten");
 
                     // mock implementation of InitializeSourceControlInformation common targets:
                     xml.Root.Add(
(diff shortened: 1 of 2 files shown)


baronfel (approved):
This is pretty much exactly what I was hoping to see 👍

MattKotsenas:
I'm going to wait to publish this until the SourceLink PR lands in case things change, specifically of interest are:

- The property name
- If the symbolic-ref prefixes are kept

Edit: Published to get eyes / feedback on this; it's still a _chance_ that I'll need to update this if the SourceLink PR change's its property name.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/PackCommandTests.cs
```diff
@@ -3809,7 +3811,7 @@ public void PackCommand_PackWithSourceControlInformation_PublishedUrl_VerifyNusp
                     var repositoryMetadata = nuspecReader.GetRepositoryMetadata();
                     repositoryMetadata.Type.Should().Be("git");
                     repositoryMetadata.Url.Should().Be("https://github.com/NuGet/NuGet.Client.git");
-                    repositoryMetadata.Branch.Should().Be("");
+                    repositoryMetadata.Branch.Should().Be("refs/heads/main");
```
nkolev92, line 3821: Do we have any idea why branch wasn't set before?
I'm asking cause there's an explicit assertion in this test.

nkolev92 (commented):
Just a question about the potential risk here.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/PackCommandTests.cs
MattKotsenas: I think it was to codify the previous SourceLink behavior. L3639 is an example: if the input property is set it shows up in the .nuspec (see https://github.com/NuGet/NuGet.Client/blob/60974f11275eabfdfb364a76130aded590cb144d/test/NuGet.Core.FuncTests/Dotnet.Integration.Test/PackCommandTests.cs#L3639). These tests we're asserting that if it wasn't explicitly set, it was empty.

Now these tests assert that if it isn't explicitly set, the value from SourceLink is used.

/cc @tmat to correct me if I'm wrong.

nkolev92 (dismissed):

nkolev92:
For the NuGet shepherd, let's wait for a review from @tmat before merging.

test/NuGet.Core.FuncTests/Dotnet.Integration.Test/PackCommandTests.cs
tmat: Yup, just validation of an existing behavior.

tmat:
Is there a potential for private info leak if we publish the branch name by default?

For Repository URL we require the project to explicitly set `PublishRepositoryUrl`.
https://github.com/NuGet/NuGet.Client/blob/60974f11275eabfdfb364a76130aded590cb144d/src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.targets#L292-L293

Branch name might potentially include code names, etc.

baronfel:
Good point - the commit sha is unique enough to not worry, but the branch could be behind the existing `'$(PublishRepositoryUrl)'` check (or a similarly named and documented check for the branch).

MattKotsenas:
@baronfel / @tmat thoughts on re-using `PublishRepositoryUrl` or creating a new flag? I defer to the experts here.

baronfel:
Happy to re-use the existing flag, mostly was bringing up the possibility of a new flag for completeness.

nkolev92:
Would pack even set the repository details if the repository url was never provided?
Like if there's no repository url, but there's a branch.

MattKotsenas:
~~Good reminder @nkolev92! I verified both [here](https://github.com/NuGet/NuGet.Client/blob/60974f11275eabfdfb364a76130aded590cb144d/src/NuGet.Core/NuGet.Build.Tasks.Pack/PackTaskLogic.cs#L177C1-L184C14) and with a quick local experiment that when `$(PublishRepositoryUrl)` is false the whole `<repository />` node is omitted from the .nuspec.~~

~~@tmat / @baronfel, I believe this means the concern is already addressed? Let me know if there's anything I missed, or if there are any other issues. Thanks!~~

Edit: This comment is incorrect; see below.

baronfel:
Perfect, that works for me. That's also the pattern we've used for the SDK Container source control information linkages.

tmat:
Sounds good. Do we have a test?

tmat:
Should also add a note to the documentation: https://github.com/dotnet/sourcelink/blob/669979bf371a587c908f732499df0ed883cd5d6c/docs/README.md?plain=1#L44

MattKotsenas:
We don't have a test covering this part, and when I add one I'm not seeing the same behavior I see from the command line. I'm looking into it, but let's hold off on merging until we understand the difference.

Edit: OK I at least understand the difference. NuGet Package Explorer doesn't show the raw nuspec XML and seems to hide the `<repository />` node if the `url` attribute is missing. Thus, my command line tests were invalid.

MattKotsenas:
I'm going to go with gating the branch name on the repository URL as suggested above and we can take a look at how that feels.

Edit: Pushed. @tmat, @nkolev92, @baronfel; mind taking another look? Thanks!

baronfel (approved):

MattKotsenas:
> Should also add a note to the documentation: https://github.com/dotnet/sourcelink/blob/669979bf371a587c908f732499df0ed883cd5d6c/docs/README.md?plain=1#L44

Available @ https://github.com/dotnet/sourcelink/pull/1252

baronfel:
Ok, we've got sourcelink merged and flowing to the dotnet/sdk again, so I think this is good to review+merge!

zivkan (approved):

PR #5925, opened 2024-07-15
Use SdkAnalysisLevel in restorecommand http-errors

### LINKED ISSUE #13546: Use `SDKAnalysisLevel` in restore "https everywhere: promote from warning to error"
Use https://github.com/NuGet/NuGet.Client/pull/5833, and avoid showing error based on the SDKAnalysisLevel value.
if it is greater than or equal to 9.0.100 > show error message.

Nigusu-Allehu:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13546

## Description
while performing restore, avoid showing error based on the SDKAnalysisLevel value.
if it is greater than or equal to 9.0.100 > show error message.

### Concerns
 Should we parse `SdkAnalysisLevel` and `UsingMicrosoftNetSdk` for the following ?
* PackageRestoreManager
     * Used to restore during scenarios where there is no solution manager
* NuGetPackageManager.cs  
     * install and update 
## PR Checklist

- [x] Added tests


build/Shared/SdkAnalysisLevelMinimums.cs
```diff
@@ -0,0 +1,10 @@
+/// <summary>
+/// Defines constants that specify the minimum SDK analysis level required to enable new features.
+/// </summary>
+namespace NuGet.Shared
+{
+    internal static class SdkAnalysisLevelMinimums
+    {
+        public const string HttpErrorSdkAnalysisLevelMinimumValue = "9.0.100";
```
nkolev92: I know it's not ready yet, but you can probably use a NuGetVersion type that way we're not reparsing it every time.

nkolev92: I'm guessing you can't use NuGetVersion because it's in shared.

Feel free to resolve this if that's the case.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -254,8 +254,14 @@ public async Task<RestoreResult> ExecuteAsync(CancellationToken token)
                         var source = remoteProvider.Source;
                         if (source.IsHttp && !source.IsHttps && !source.AllowInsecureConnections)
                         {
-                            await _logger.LogAsync(RestoreLogMessage.CreateError(NuGetLogCode.NU1302,
+                            if ((_request.Project.RestoreMetadata.SdkAnalysisLevel is not null &&
+                                _request.Project.RestoreMetadata.SdkAnalysisLevel >= new NuGetVersion(Shared.SdkAnalysisLevelMinimums.HttpErrorSdkAnalysisLevelMinimumValue)) ||
+                                (_request.Project.RestoreMetadata.UsingMicrosoftNETSdk == false && _request.Project.RestoreMetadata.SdkAnalysisLevel is null))
```
nkolev92: Do we need the `_request.Project.RestoreMetadata.UsingMicrosoftNETSdk == false` check? 

If SdkAnalysisLevel is null, should we just log the error? 
Seems easier to read and yields the same result.

test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommandTests.cs
```diff
@@ -4095,6 +4095,99 @@ public async Task Restore_WithHttpSourceAndAllowInsecureConnectionsFalse_ThrowsE
             Assert.Contains(httpSourceUrl, logMessage.Message);
         }
 
+        [Theory]
+        [InlineData("8.0.100")]
```
nkolev92: When i think of the different codepaths this theory test exercises, it's really only testing the parsing method and the compare. 
That's well tested through unit tests.
I'm not really sure we need to test all these different versions. 1 version is enough.

test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommandTests.cs
```diff
@@ -4095,6 +4095,99 @@ public async Task Restore_WithHttpSourceAndAllowInsecureConnectionsFalse_ThrowsE
...
+            // Act
+            var result = await command.ExecuteAsync();
+
+            // Assert
+            result.Success.Should().BeTrue(because: logger.ShowMessages());
+            result.LockFile.LogMessages.Should().HaveCount(0);
+        }
+
+        [Theory]
+        [InlineData("9.0.100")]
```
nkolev92: Same feedback as earlier. We're probably not gaining too much by testing many versions.

test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommandTests.cs
```diff
@@ -4095,6 +4095,99 @@ public async Task Restore_WithHttpSourceAndAllowInsecureConnectionsFalse_ThrowsE
...
+            result.LockFile.LogMessages.Should().HaveCount(1);
+            IAssetsLogMessage logMessage = result.LockFile.LogMessages[0];
+            logMessage.Code.Should().Be(NuGetLogCode.NU1302);
+            Assert.Contains(httpSourceUrl, logMessage.Message);
+        }
+
+        [Theory]
+        [InlineData("8.0.100")]
+        [InlineData("9.0.400")]
+        [InlineData("9.8.100")]
```
nkolev92: I'd just use 2, one smaller, one larger, you don't need the 3rd one.

nkolev92 (commented):
Looks good. 
Some minor feedback.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -254,8 +254,14 @@ public async Task<RestoreResult> ExecuteAsync(CancellationToken token)
...
+                            {
+                                // SdkAnalysisLevel >= 9.0.100
+                                await _logger.LogAsync(RestoreLogMessage.CreateError(NuGetLogCode.NU1302,
```
nkolev92, line 263: We should have a warning in the else statement.

build/Shared/SdkAnalysisLevelMinimums.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.Shared
+{
+    /// <summary>
+    /// Defines constants that specify the minimum SDK analysis level required to enable new features.
+    /// </summary>
+    internal static class SdkAnalysisLevelMinimums
+    {
+        public const string HttpErrorSdkAnalysisLevelMinimumValue = "9.0.100";
```
kartheekp-ms: Can we store the `NuGetVersion` type directly instead of using a string and creating a `NuGetVersion` object every time for the comparison?

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -254,8 +254,20 @@ public async Task<RestoreResult> ExecuteAsync(CancellationToken token)
...
                                 string.Format(CultureInfo.CurrentCulture, Strings.Error_HttpSource_Single, "restore", source.Source)));
+                            }
+                            else
+                            {
+                                // SdkAnalysisLevel < 9.0.100
+                                await _logger.LogAsync(RestoreLogMessage.CreateError(NuGetLogCode.NU1803,
+                                string.Format(CultureInfo.CurrentCulture, Strings.Warning_HttpServerUsage, "restore", source.Source)));
+                            }
```
kartheekp-ms: I am confused because in both cases, an error is logged with a different message. I would have expected one case to log an error, whereas the other case should log a warning. Please let me know if I am missing something.

build/Shared/SdkAnalysisLevelMinimums.cs
Nigusu-Allehu: https://github.com/NuGet/NuGet.Client/pull/5925#discussion_r1678438801

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
Nigusu-Allehu: Thanks. You are right, it should be a warning

build/Shared/SdkAnalysisLevelMinimums.cs
Nigusu-Allehu: It's because this is just a .cs file in the shared folder. It's not in any NuGet project

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
Nigusu-Allehu: I was following this logic https://github.com/NuGet/NuGet.Client/pull/5896#issue-2385635403. But since this is only going into .Net 9 only, we could remove it I guess.  @zivkan what do you think?

build/Shared/SdkAnalysisLevelMinimums.cs
zivkan: [Looking at the project graph](https://github.com/NuGet/Client.Engineering/issues/2255), there's only 4 projects that don't reference NuGet.Versioning.  Can we add `<Compile Remove="$(Shared)\SdkAnalysisLevelMinimums.cs" />` in those 4 projects?  If NuGet.Configuration ever needs to use SdkAnalysisLevel, then it might need to take a dependency on NuGet.Versioning. The other 3 I can't imagine will ever need to condition on SdkAnalysisVersion.

Something I don't like about `build/Shared/*` is that it bloats the executable size for every single binary, including the ones that don't even use the class. Personally, I would prefer a `build/KindaShared/*` where projects have to individually add `<Compile` items for only the files it cares about.  I don't know if my opinion is an unpopular one though.

zivkan: Thinking ahead to a future where we have multiple features that are gating on the same SdkAnalysisLevel version, would it be better to have one `SdkAnalysisLevelMinimums.v9_0_100` that they all use, or multiple feature-named properties with the same value?

The compiler will de-duplicate compile time constant strings, so that's not a concern.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -254,8 +254,23 @@ public async Task<RestoreResult> ExecuteAsync(CancellationToken token)
                         var source = remoteProvider.Source;
                         if (source.IsHttp && !source.IsHttps && !source.AllowInsecureConnections)
                         {
-                            await _logger.LogAsync(RestoreLogMessage.CreateError(NuGetLogCode.NU1302,
+                            var sdkLevel = _request.Project.RestoreMetadata.SdkAnalysisLevel;
+                            var minSdkLevel = new NuGetVersion(Shared.SdkAnalysisLevelMinimums.HttpErrorSdkAnalysisLevelMinimumValue);
+
+                            if (sdkLevel != null && sdkLevel >= minSdkLevel ||
+                                sdkLevel == null && _request.Project.RestoreMetadata.UsingMicrosoftNETSdk == false)
```
zivkan: We can do this in a separate PR, but I was wondering if we create a method `bool SdkAnalysisLevelMinimums.IsEnabled(NuGetVersion sdkAnalysisLevel, bool usingMicrosoftNETSdk, NuGetVersion minVersion)`, then we'll have a reusable method for all future SDK Analysis Level checks.

If we also create an overload where all parameters are strings, we should be able to call it from MSBuild, so we'll be able to reproduce what my PR did a bit simpler (although because nuget.exe is ILMerged, it will probably be difficult to get working 😭 )

test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommandTests.cs
```diff
@@ -4095,6 +4095,94 @@ public async Task Restore_WithHttpSourceAndAllowInsecureConnectionsFalse_ThrowsE
...
 
+        [Fact]
+        public async Task Restore_WithHttpSourceSdkAnalysisLevelLowerThan90100_Warns()
+        {
+            // Arrange
+            using var pathContext = new SimpleTestPathContext();
+            var packageA = new SimpleTestPackageContext("a", "1.0.0");
+            await SimpleTestPackageUtility.CreateFolderFeedV3Async(pathContext.PackageSource, packageA);
+            string httpSourceUrl = "http://api.source/index.json";
+            string httpsSourceUrl = "https://api.source/index.json";
```
zivkan: Only use `.test` TLD for tests. This is a reserved TLD that should never get registered, whereas `api.source` could potentially resolve to an IP address. Even if this test doesn't make network requests, if someone copy-pastes this into another test, they might use an API that does make network requests, and we want to eliminate the risk that our CI starts hammering some 3rd part server.

Same for all the other tests.

zivkan: I don't think the https feed plays any part in this test, right?  If that's right, it makes the test longer, and takes me longer to read/verify the test because now I'm trying to figure out why this second source is relevant to the test.

test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommandTests.cs
```diff
@@ -4095,6 +4095,94 @@ public async Task Restore_WithHttpSourceAndAllowInsecureConnectionsFalse_ThrowsE
...
+            result.Success.Should().BeFalse(because: logger.ShowMessages());
+            result.LockFile.LogMessages.Should().HaveCount(1);
+            IAssetsLogMessage logMessage = result.LockFile.LogMessages[0];
+            logMessage.Code.Should().Be(NuGetLogCode.NU1302);
+            Assert.Contains(httpSourceUrl, logMessage.Message);
+        }
+
+        [Theory]
+        [InlineData("8.0.100")]
+        [InlineData("9.8.100")]
```
zivkan: I don't know if you're using unrealistic SDK versions intentionally. There's certainly nothing stopping customers from setting it to such values. But for actual SDKs that will be released, the minor version will always be `0`. The "patch" version will be `100`, `200`, `300`, or `400`.

zivkan (commented):
I feel somewhat strongly about not using potentially valid DNS host names in tests. Otherwise I'd be happy to approve this.

I know there's too much in the coding guidelines document to remember everything, but is it worthwhile adding something in there, or create a new `testing-guidelines.md`, so that in the future we can just link to it (plus maybe someone will actually read it and remember when writing a test, leading to fewer PR iterations)

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
zivkan: It's needed, as described here: https://github.com/NuGet/NuGet.Client/blob/dev/docs/feature-guide.md#warnings-and-defaults

SDK and non-SDK projects have **different** defaults.

test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommandTests.cs
Nigusu-Allehu: It's just to make sure we are throwing errors despite having both http and https sources.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
Nigusu-Allehu: I think that's a great idea.

build/Shared/SdkAnalysisLevelMinimums.cs
Nigusu-Allehu: I prefer having feature-based constants. It gives context to people reading the code.

Nigusu-Allehu: Moved it to NuGet.Versioning, I think it makes things simpler

src/NuGet.Core/NuGet.Versioning/utility/SdkAnalysisLevelMinimums.cs
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.Versioning
```
nkolev92: I'm not 100% sure where an api like this would belong, but I don't NuGet.Versioning is the right place.

test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommandTests.cs
```diff
@@ -4095,15 +4095,103 @@ public async Task Restore_WithHttpSourceAndAllowInsecureConnectionsFalse_ThrowsE
...
         [Fact]
         public async Task Restore_WithHttpSourceAndAllowInsecureConnectionsTrue_Succeeds()
         {
             // Arrange
             using var pathContext = new SimpleTestPathContext();
             var packageA = new SimpleTestPackageContext("a", "1.0.0");
             await SimpleTestPackageUtility.CreateFolderFeedV3Async(pathContext.PackageSource, packageA);
-            string httpSourceUrl = "http://api.source/index.json";
-            string httpsSourceUrl = "https://api.source/index.json";
+            string httpSourceUrl = "http://unit.test/index.json";
```
nkolev92, line 4193: Just fyi, you might see a lot of tests referring to contoso.

Here's some history: https://learn.microsoft.com/en-us/microsoft-365/enterprise/contoso-case-study?view=o365-worldwide.

src/NuGet.Core/NuGet.Versioning/utility/SdkAnalysisLevelMinimums.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <summary>
+        /// Determines whether the feature is enabled based on the SDK analysis level.
+        /// </summary>
+        /// <param name="sdkAnalysisLevel">The project SdkAnalysisLevel value </param>
+        /// <param name="usingMicrosoftNetSdk">Is it SDK project or not</param>
+        /// <param name="minSdkVersion">The minimum version of the SDK required for the feature to be enabled.</param>
+        /// <returns>Returns true if the feature should be enabled based on the given parameters; otherwise, false.</returns>
+        public static bool IsEnabled(NuGetVersion sdkAnalysisLevel, bool usingMicrosoftNetSdk, NuGetVersion minSdkVersion)
+        {
+            if (sdkAnalysisLevel != null && sdkAnalysisLevel >= minSdkVersion ||
```
nkolev92: I think this is much easier to read than the previous implementation.

src/NuGet.Core/NuGet.Versioning/utility/SdkAnalysisLevelMinimums.cs
```diff
@@ -0,0 +1,38 @@
...
+        public static readonly NuGetVersion HttpErrorSdkAnalysisLevelMinimumValue = new("9.0.100");
+
+        /// <summary>
+        /// Determines whether the feature is enabled based on the SDK analysis level.
+        /// </summary>
+        /// <param name="sdkAnalysisLevel">The project SdkAnalysisLevel value </param>
+        /// <param name="usingMicrosoftNetSdk">Is it SDK project or not</param>
+        /// <param name="minSdkVersion">The minimum version of the SDK required for the feature to be enabled.</param>
+        /// <returns>Returns true if the feature should be enabled based on the given parameters; otherwise, false.</returns>
+        public static bool IsEnabled(NuGetVersion sdkAnalysisLevel, bool usingMicrosoftNetSdk, NuGetVersion minSdkVersion)
```
nkolev92: If this is going to be a separate API, we need tests.

nkolev92 (commented):
I like the separate API.

Maybe we keep it internal in NuGet.Commands until we have another use-case. 

3 use-cases tends to be the sweet spot for designing a good API, and currently we just have 1. An idea would be something that combines feature flags alongside the SDK analysis level, but really all of this would be easier if we had more use-cases.

src/NuGet.Core/NuGet.Versioning/utility/SdkAnalysisLevelMinimums.cs
Nigusu-Allehu: made it internal in Nuget.Commands for now. As you said, when we have more use cases in the future, we can look for better location for it if we need to or just make it public.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -254,8 +254,21 @@ public async Task<RestoreResult> ExecuteAsync(CancellationToken token)
...
+                            {
+                                await _logger.LogAsync(RestoreLogMessage.CreateError(NuGetLogCode.NU1302,
                                 string.Format(CultureInfo.CurrentCulture, Strings.Error_HttpSource_Single, "restore", source.Source)));
+                            }
+                            else
+                            {
+                                await _logger.LogAsync(RestoreLogMessage.CreateWarning(NuGetLogCode.NU1803,
+                                string.Format(CultureInfo.CurrentCulture, Strings.Warning_HttpServerUsage, "restore", source.Source)));
+                            }
+
```
jebriede: nit: remove blank line.

nkolev92 (dismissed):

test/NuGet.Core.Tests/NuGet.Commands.Test/SdkAnalysisLevelMinimumsTests.cs
```diff
@@ -0,0 +1,46 @@
...
+            var result = SdkAnalysisLevelMinimums.IsEnabled(new NuGetVersion("9.0.100"), true, new NuGetVersion("9.0.100"));
+            Assert.True(result);
+        }
+
+        [Fact]
+        public void IsEnabled_WhenSdkAnalysisLevelIsGreaterThanMinSdkVersion_ShouldReturnTrue()
+        {
+            var result = SdkAnalysisLevelMinimums.IsEnabled(new NuGetVersion("9.0.101"), true, new NuGetVersion("9.0.100"));
+            Assert.True(result);
+        }
```
jebriede, line 44: Nit: these could be combined into a data-driven test (using Theory) where you supply 4 parameters: the first three being the inputs, the fourth being the expectedValue (true or false).

zivkan (dismissed):

build/Shared/SdkAnalysisLevelMinimums.cs
zivkan: The code where it's reference is context too 😁

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -254,8 +254,20 @@ public async Task<RestoreResult> ExecuteAsync(CancellationToken token)
                         var source = remoteProvider.Source;
                         if (source.IsHttp && !source.IsHttps && !source.AllowInsecureConnections)
                         {
-                            await _logger.LogAsync(RestoreLogMessage.CreateError(NuGetLogCode.NU1302,
+                            var isErrorEnabled = SdkAnalysisLevelMinimums.IsEnabled(_request.Project.RestoreMetadata.SdkAnalysisLevel,
+                                _request.Project.RestoreMetadata.UsingMicrosoftNETSdk,
+                                SdkAnalysisLevelMinimums.HttpErrorSdkAnalysisLevelMinimumValue);
```
zivkan, line 259: We can see here how "wordy" the property name is. The C# language doesn't let you (easily) use the property without also specifying the class name, and having "SdkAnalysisLevelMinimum" in both the class and property looks redundant when used.

But this is something easily changed later, since it's not a public API.  Just something to keep in mind when writing new code.

zivkan (approved):

PR #5926, opened 2024-07-16
Display all transitive packages in PM UI

### LINKED ISSUE #13060: PM UI transitive dependencies should display all transitive dependencies, not just ones brought in through packages directly installed in a project
### NuGet Product(s) Affected

### Current Behavior

There's many facets to this: 

Example 1: 

Start with a project Library, referencing Newtonsoft.Json version 12.0.3, which is vulnerable. 
Reference that project from an application. 
Go to the application project, see you have no packages listed (despite having transitive packages)

This is incorrect, since Newtonsoft.Json is a transitive package to Application through library. 
Yes, library the project directly installing it, but in the end it is available for coding in Application in the default behavior for packages. 
```xml

  <ItemGroup>
    <ProjectReference Include="..\Library\Library.csproj" />
  </ItemGroup>

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <Nullable>enable</Nullable>
  </PropertyGroup>

</Project>
```

```xml

  <PropertyGroup>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="newtonsoft.json" Version="12.0.3" />
  </ItemGroup>

</Project>

```

[Example1.zip](https://github.com/NuGet/Home/files/13519378/Example1.zip)

Example 2: 

Building on top of example 1, say you install NuGet.Packaging version 5.11.5. 
Note that this references version 9.0.1, while the project references version 12.0.3. 

Note that the transitive package itself is shown, despite the fact that it is the project itself that brings in the transitive package. 

image

[Example2.zip](https://github.com/NuGet/Home/files/13519455/Example2.zip)

Example 3: 

Say you have the following application project: 
```xml

  <ItemGroup>
    <ProjectReference Include="..\Library\Library.csproj" />
  </ItemGroup>

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <Nullable>enable</Nullable>
  </PropertyGroup>

</Project>
```
And the following library project; 

```xml

  <PropertyGroup>
    <Nullable>enable</Nullable>
	<NoWarn>NU1605</NoWarn>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="newtonsoft.json" Version="6.0.4" />  
  </ItemGroup>
	
  <ItemGroup>
	<PackageReference Include="NuGet.Packaging" Version="5.11.5" />
  </ItemGroup>

</Project>
```

If you restore these projects in Visual Studio, you'd get: 

image

Yet, if you open the PM UI, you get: 

image

It inconsistent with what restore shows

For all these examples, there is an inconsistency. 

- dotnet list package --include-transitive shows all transitive packages
- Solution Explorer shows all transitive packages (on demand, you need to expand them).
- Inconsistent with restore

### Desired Behavior

Show all transitive packages, regardless of how they were installed.

The current behavior leads to inconsistency among Package Manager UI, Solution Explorer, dotnet list package and restore itself.
The behavior should be aligned.

### Additional Context

_No response_

martinrrm:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13060
One pager: https://github.com/NuGet/Client.Engineering/pull/2934

## Description
Currently, the PM UI is only displaying transitive packages from directly installed packages in that project. But there is another way packages can be brought to a project, by using <ProjectReference>. Since Project Reference’s are not considered installed packages, we will not bring any packages that the project has installed leading to inconsistency among PM UI, dotnet list package and restore itself.

With this PR I'm adding those packages that are brought by project references to the PM UI.

### Application.csproj
```xml
  <ItemGroup>
    <ProjectReference Include="..\Library\Library.csproj" />
  </ItemGroup>

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <Nullable>enable</Nullable>
  </PropertyGroup>
</Project>
```

### Library.csproj
```xml
	<PropertyGroup>
		<Nullable>enable</Nullable>
		<NoWarn>NU1605</NoWarn>
	</PropertyGroup>
	<ItemGroup>
		<PackageReference Include="Newtonsoft.Json" Version="6.0.4" />
	</ItemGroup>
	<ItemGroup>
		<PackageReference Include="NuGet.Packaging" Version="5.11.5" />
	</ItemGroup>
</Project>
```

### Application PM UI
image
image

## PR Checklist

- [x] Added tests

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
@@ -259,7 +259,7 @@ namespace NuGet.PackageManagement.VisualStudio
                         {
                             transitiveOrigins.TryGetValue(packageRef.PackageIdentity.Id, out TransitiveEntry cacheEntry);
                             return MergeTransitiveOrigin(packageRef, cacheEntry);
-                        });
+                        }).ToList();
 
                     lock (_transitiveOriginsLock)
                     {
@@ -347,7 +347,7 @@ namespace NuGet.PackageManagement.VisualStudio
             {
                 return targets
                     .Where(target => target.TargetFramework.Equals(targetFramework))
-                    .SelectMany(target => GetPackageReferenceUtility.UpdateTransitiveDependencies2(target.Libraries, installedPackages, transitivePackages))
+                    .SelectMany(target => GetPackageReferenceUtility.UpdateTransitiveDependencies(target.Libraries, installedPackages, transitivePackages))
                     .Select(packageIdentity => new PackageReference(packageIdentity, targetFramework))
                     .ToList();
             }
@@ -434,7 +434,7 @@ namespace NuGet.PackageManagement.VisualStudio
 
             await TaskScheduler.Default;
 
-            LockFile lockFile = LockFileUtilities.GetLockFile(assetsFilePath, NullLogger.Instance, LockFileReadFlags.PackageFolders | LockFileReadFlags.Targets | LockFileReadFlags.Libraries | LockFileReadFlags.ProjectFileDependencyGroups);
+            LockFile lockFile = LockFileUtilities.GetLockFile(assetsFilePath, NullLogger.Instance, LockFileReadFlags.PackageFolders | LockFileReadFlags.Targets);
             _packageFolders = lockFile?.PackageFolders ?? Array.Empty<LockFileItem>();
 
             return lockFile?.Targets;
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
@@ -219,27 +219,31 @@ namespace NuGet.PackageManagement.VisualStudio
                             targetsList = await GetTargetsListAsync(assetsFilePath, token);
                         }
 
-                        List<LockFileTargetLibrary> projectReferences = targetsList.SelectMany(t => t.Libraries).Where(l => l.Type == LibraryType.Project).ToList();
-                        List<PackageReference> installedProjectReferences = new List<PackageReference>(projectReferences.Count());
-
-                        // Since ProjectReferences are not in the dependencies section of the package spec (installed packages), we need to manually add them
-                        // to the installed packages list to compute transitive origins
-                        foreach (var projectReference in projectReferences)
-                        {
-                            var packageReference = new PackageReference(
-                                identity: new PackageIdentity(projectReference.Name, projectReference.Version),
-                                targetFramework: new NuGetFramework(projectReference.Framework),
-                                userInstalled: false,
-                                developmentDependency: false,
-                                requireReinstallation: false,
-                                allowedVersions: VersionRange.Parse(projectReference.Version.ToString()));
-                            installedProjectReferences.Add(packageReference);
-                        }
-
-                        installedProjectReferences.AddRange(calculatedInstalledPackages);
+                        Dictionary<string, TransitiveEntry> transitiveOriginsCache = new();
 
                         // Compute Transitive Origins
-                        transitiveOrigins = calculatedTransitivePackages.Any() ? ComputeTransitivePackageOrigins(installedProjectReferences, targetsList, token) : new Dictionary<string, TransitiveEntry>();
+                        transitiveOrigins = calculatedTransitivePackages.Any() ? ComputeTransitivePackageOrigins(calculatedInstalledPackages, targetsList, transitiveOriginsCache, token) : transitiveOriginsCache;
+
+                        // If the project has project references, we need to compute transitive origins for their packages
+                        List<LockFileTargetLibrary> projectReferences = targetsList.SelectMany(t => t.Libraries).Where(l => l.Type == LibraryType.Project).ToList();
+                        if (projectReferences.Any())
+                        {
+                            List<PackageReference> calculatedProjectReferences = new List<PackageReference>(projectReferences.Count);
+                            foreach (var projectReference in projectReferences)
+                            {
+                                var packageReference = new PackageReference(
+                                    identity: new PackageIdentity(projectReference.Name, projectReference.Version),
+                                    targetFramework: new NuGetFramework(projectReference.Framework),
+                                    userInstalled: false,
+                                    developmentDependency: false,
+                                    requireReinstallation: false,
+                                    allowedVersions: new VersionRange(projectReference.Version));
+                                calculatedProjectReferences.Add(packageReference);
+                            }
+
+                            // Compute transitive origins for project references packages
+                            transitiveOrigins = ComputeTransitivePackageOrigins(calculatedProjectReferences, targetsList, transitiveOriginsCache, token);
+                        }
                     }
                     else
                     {
@@ -360,11 +364,9 @@ namespace NuGet.PackageManagement.VisualStudio
         /// packages that depends on given transitive package</returns>
         /// <remarks>Computes all transitive origins for each Framework/Runtime-ID combiation. Runtime-ID can be <see langword="null" />.
         /// Transitive origins are calculated using a Depth First Search algorithm on all direct dependencies exhaustively</remarks>
-        internal static Dictionary<string, TransitiveEntry> ComputeTransitivePackageOrigins(List<PackageReference> installedPackages, IList<LockFileTarget> targetsList, CancellationToken ct)
+        internal static Dictionary<string, TransitiveEntry> ComputeTransitivePackageOrigins(List<PackageReference> installedPackages, IList<LockFileTarget> targetsList, Dictionary<string, TransitiveEntry> transitiveOriginsCache, CancellationToken ct)
         {
             ct.ThrowIfCancellationRequested();
-
-            Dictionary<string, TransitiveEntry> transitiveOriginsCache = new();
 
             // Find all Transitive origins and update cache
             var memoryVisited = new HashSet<PackageIdentity>();
@@ -468,29 +470,32 @@ namespace NuGet.PackageManagement.VisualStudio
             {
                 visited.Add(current.PackageIdentity); // visited
 
-                // Lookup Transitive Origins Cache
-                TransitiveEntry cachedEntry;
-                if (!transitiveOriginsCache.TryGetValue(current.PackageIdentity.Id, out cachedEntry))
+                // Set transitive origins if the dependency is a package
+                if (node.Type == LibraryType.Package.Value)
                 {
-                    cachedEntry = new Dictionary<FrameworkRuntimePair, IList<PackageReference>>
+                    // Lookup Transitive Origins Cache
+                    TransitiveEntry cachedEntry;
+                    if (!transitiveOriginsCache.TryGetValue(current.PackageIdentity.Id, out cachedEntry))
                     {
-                        [fxRidEntry] = new List<PackageReference>()
-                    };
-                }
+                        cachedEntry = new Dictionary<FrameworkRuntimePair, IList<PackageReference>>
+                        {
+                            [fxRidEntry] = new List<PackageReference>()
+                        };
+                    }
 
-                if (!cachedEntry.ContainsKey(fxRidEntry))
-                {
-                    cachedEntry[fxRidEntry] = new List<PackageReference>();
-                }
+                    if (!cachedEntry.ContainsKey(fxRidEntry))
+                    {
+                        cachedEntry[fxRidEntry] = new List<PackageReference>();
+                    }
 
-                // In order to hide Projects from the PM UI, we don't add them to the transitive origins cache
-                if (!cachedEntry[fxRidEntry].Contains(top) && node.Type == LibraryType.Package.Value) // Dictionary value is a List. If perf. is bad, change to HashSet.
-                {
-                    cachedEntry[fxRidEntry].Add(top);
-                }
+                    if (!cachedEntry[fxRidEntry].Contains(top)) // Dictionary value is a List. If perf. is bad, change to HashSet.
+                    {
+                        cachedEntry[fxRidEntry].Add(top);
+                    }
 
-                // Upsert Transitive Origins Cache
-                transitiveOriginsCache[current.PackageIdentity.Id] = cachedEntry;
+                    // Upsert Transitive Origins Cache
+                    transitiveOriginsCache[current.PackageIdentity.Id] = cachedEntry;
+                }
 
                 foreach (PackageDependency dep in node.Dependencies.ToList()) // Casting to list to prevent backing allocations
                 {
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
@@ -20,6 +20,7 @@ using NuGet.ProjectManagement;
 using NuGet.ProjectManagement.Projects;
 using NuGet.ProjectModel;
 using NuGet.Shared;
+using NuGet.Versioning;
 using NuGet.VisualStudio.Internal.Contracts;
 using TransitiveEntry = System.Collections.Generic.IDictionary<NuGet.Frameworks.FrameworkRuntimePair, System.Collections.Generic.IList<NuGet.Packaging.PackageReference>>;
 
@@ -218,7 +219,27 @@ namespace NuGet.PackageManagement.VisualStudio
                             targetsList = await GetTargetsListAsync(assetsFilePath, token);
                         }
 
-                        transitiveOrigins = calculatedTransitivePackages.Any() ? ComputeTransitivePackageOrigins(calculatedInstalledPackages, targetsList, token) : new Dictionary<string, TransitiveEntry>();
+                        List<LockFileTargetLibrary> projectReferences = targetsList.SelectMany(t => t.Libraries).Where(l => l.Type == LibraryType.Project).ToList();
+                        List<PackageReference> installedProjectReferences = new List<PackageReference>(projectReferences.Count());
+
+                        // Since ProjectReferences are not in the dependencies section of the package spec (installed packages), we need to manually add them
+                        // to the installed packages list to compute transitive origins
+                        foreach (var projectReference in projectReferences)
+                        {
+                            var packageReference = new PackageReference(
+                                identity: new PackageIdentity(projectReference.Name, projectReference.Version),
+                                targetFramework: new NuGetFramework(projectReference.Framework),
+                                userInstalled: false,
+                                developmentDependency: false,
+                                requireReinstallation: false,
+                                allowedVersions: VersionRange.Parse(projectReference.Version.ToString()));
+                            installedProjectReferences.Add(packageReference);
+                        }
+
+                        installedProjectReferences.AddRange(calculatedInstalledPackages);
+
+                        // Compute Transitive Origins
+                        transitiveOrigins = calculatedTransitivePackages.Any() ? ComputeTransitivePackageOrigins(installedProjectReferences, targetsList, token) : new Dictionary<string, TransitiveEntry>();
                     }
                     else
                     {
@@ -323,7 +344,6 @@ namespace NuGet.PackageManagement.VisualStudio
             {
                 return targets
                     .SelectMany(target => target.Libraries)
-                    .Where(library => library.Type == LibraryType.Package)
                     .SelectMany(library => GetPackageReferenceUtility.UpdateTransitiveDependencies(library, targetFramework, targets, installedPackages, transitivePackages))
                     .Select(packageIdentity => new PackageReference(packageIdentity, targetFramework))
                     .ToList();
@@ -434,7 +454,7 @@ namespace NuGet.PackageManagement.VisualStudio
             // Find first target node that matches current
             foreach (LockFileTargetLibrary lib in graph.Libraries)
             {
-                if (lib.Type == LibraryType.Package.Value
+                if ((lib.Type == LibraryType.Package.Value || lib.Type == LibraryType.Project.Value)
                     && string.Equals(lib.Name, current.PackageIdentity.Id, StringComparison.OrdinalIgnoreCase)
                     && ((current.HasAllowedVersions && current.AllowedVersions.Satisfies(lib.Version)) ||
                         (current.PackageIdentity.HasVersion && current.PackageIdentity.Version.Equals(lib.Version))))
@@ -463,7 +483,8 @@ namespace NuGet.PackageManagement.VisualStudio
                     cachedEntry[fxRidEntry] = new List<PackageReference>();
                 }
 
-                if (!cachedEntry[fxRidEntry].Contains(top)) // Dictionary value is a List. If perf. is bad, change to HashSet.
+                // In order to hide Projects from the PM UI, we don't add them to the transitive origins cache
+                if (!cachedEntry[fxRidEntry].Contains(top) && node.Type == LibraryType.Package.Value) // Dictionary value is a List. If perf. is bad, change to HashSet.
                 {
                     cachedEntry[fxRidEntry].Add(top);
                 }
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
@@ -235,10 +235,6 @@ namespace NuGet.PackageManagement.VisualStudio
                                     allowedVersions: new VersionRange(projectReference.Version));
                                 calculatedLibraryReferences.Add(packageReference);
                             }
-
-                            calculatedLibraryReferences = calculatedLibraryReferences
-                                .GroupBy(p => p.PackageIdentity)
-                                .Select(g => g.OrderBy(p => p.TargetFramework, FrameworkSorter).First()).ToList();
                         }
 
                         calculatedLibraryReferences.AddRange(calculatedInstalledPackages);
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
@@ -221,14 +221,11 @@ namespace NuGet.PackageManagement.VisualStudio
 
                         Dictionary<string, TransitiveEntry> transitiveOriginsCache = new();
 
-                        // Compute Transitive Origins
-                        transitiveOrigins = calculatedTransitivePackages.Any() ? ComputeTransitivePackageOrigins(calculatedInstalledPackages, targetsList, transitiveOriginsCache, token) : transitiveOriginsCache;
-
                         // If the project has project references, we need to compute transitive origins for their packages
                         List<LockFileTargetLibrary> projectReferences = targetsList.SelectMany(t => t.Libraries).Where(l => l.Type == LibraryType.Project).ToList();
+                        List<PackageReference> calculatedLibraryReferences = new List<PackageReference>(projectReferences.Count + calculatedInstalledPackages.Count);
                         if (projectReferences.Any())
                         {
-                            List<PackageReference> calculatedProjectReferences = new List<PackageReference>(projectReferences.Count);
                             foreach (var projectReference in projectReferences)
                             {
                                 var packageReference = new PackageReference(
@@ -238,12 +235,14 @@ namespace NuGet.PackageManagement.VisualStudio
                                     developmentDependency: false,
                                     requireReinstallation: false,
                                     allowedVersions: new VersionRange(projectReference.Version));
-                                calculatedProjectReferences.Add(packageReference);
+                                calculatedLibraryReferences.Add(packageReference);
                             }
-
-                            // Compute transitive origins for project references packages
-                            transitiveOrigins = ComputeTransitivePackageOrigins(calculatedProjectReferences, targetsList, transitiveOriginsCache, token);
                         }
+
+                        calculatedLibraryReferences.AddRange(calculatedInstalledPackages);
+
+                        // Compute Transitive Origins
+                        transitiveOrigins = calculatedTransitivePackages.Any() ? ComputeTransitivePackageOrigins(calculatedLibraryReferences, targetsList, token) : new Dictionary<string, TransitiveEntry>();
                     }
                     else
                     {
@@ -347,8 +346,8 @@ namespace NuGet.PackageManagement.VisualStudio
             else
             {
                 return targets
-                    .SelectMany(target => target.Libraries)
-                    .SelectMany(library => GetPackageReferenceUtility.UpdateTransitiveDependencies(library, targetFramework, targets, installedPackages, transitivePackages))
+                    .Where(target => target.TargetFramework.Equals(targetFramework))
+                    .SelectMany(target => GetPackageReferenceUtility.UpdateTransitiveDependencies2(target.Libraries, installedPackages, transitivePackages))
                     .Select(packageIdentity => new PackageReference(packageIdentity, targetFramework))
                     .ToList();
             }
@@ -364,9 +363,11 @@ namespace NuGet.PackageManagement.VisualStudio
         /// packages that depends on given transitive package</returns>
         /// <remarks>Computes all transitive origins for each Framework/Runtime-ID combiation. Runtime-ID can be <see langword="null" />.
         /// Transitive origins are calculated using a Depth First Search algorithm on all direct dependencies exhaustively</remarks>
-        internal static Dictionary<string, TransitiveEntry> ComputeTransitivePackageOrigins(List<PackageReference> installedPackages, IList<LockFileTarget> targetsList, Dictionary<string, TransitiveEntry> transitiveOriginsCache, CancellationToken ct)
+        internal static Dictionary<string, TransitiveEntry> ComputeTransitivePackageOrigins(List<PackageReference> installedPackages, IList<LockFileTarget> targetsList, CancellationToken ct)
         {
             ct.ThrowIfCancellationRequested();
+
+            Dictionary<string, TransitiveEntry> transitiveOriginsCache = new();
 
             // Find all Transitive origins and update cache
             var memoryVisited = new HashSet<PackageIdentity>();
@@ -433,7 +434,7 @@ namespace NuGet.PackageManagement.VisualStudio
 
             await TaskScheduler.Default;
 
-            LockFile lockFile = LockFileUtilities.GetLockFile(assetsFilePath, NullLogger.Instance, LockFileReadFlags.PackageFolders | LockFileReadFlags.Targets);
+            LockFile lockFile = LockFileUtilities.GetLockFile(assetsFilePath, NullLogger.Instance, LockFileReadFlags.PackageFolders | LockFileReadFlags.Targets | LockFileReadFlags.Libraries | LockFileReadFlags.ProjectFileDependencyGroups);
             _packageFolders = lockFile?.PackageFolders ?? Array.Empty<LockFileItem>();
 
             return lockFile?.Targets;
@@ -470,32 +471,28 @@ namespace NuGet.PackageManagement.VisualStudio
             {
                 visited.Add(current.PackageIdentity); // visited
 
-                // Set transitive origins if the dependency is a package
-                if (node.Type == LibraryType.Package.Value)
+                // Lookup Transitive Origins Cache
+                TransitiveEntry cachedEntry;
+                if (!transitiveOriginsCache.TryGetValue(current.PackageIdentity.Id, out cachedEntry))
                 {
-                    // Lookup Transitive Origins Cache
-                    TransitiveEntry cachedEntry;
-                    if (!transitiveOriginsCache.TryGetValue(current.PackageIdentity.Id, out cachedEntry))
+                    cachedEntry = new Dictionary<FrameworkRuntimePair, IList<PackageReference>>
                     {
-                        cachedEntry = new Dictionary<FrameworkRuntimePair, IList<PackageReference>>
-                        {
-                            [fxRidEntry] = new List<PackageReference>()
-                        };
-                    }
-
-                    if (!cachedEntry.ContainsKey(fxRidEntry))
-                    {
-                        cachedEntry[fxRidEntry] = new List<PackageReference>();
-                    }
-
-                    if (!cachedEntry[fxRidEntry].Contains(top)) // Dictionary value is a List. If perf. is bad, change to HashSet.
-                    {
-                        cachedEntry[fxRidEntry].Add(top);
-                    }
-
-                    // Upsert Transitive Origins Cache
-                    transitiveOriginsCache[current.PackageIdentity.Id] = cachedEntry;
+                        [fxRidEntry] = new List<PackageReference>()
+                    };
                 }
+
+                if (!cachedEntry.ContainsKey(fxRidEntry))
+                {
+                    cachedEntry[fxRidEntry] = new List<PackageReference>();
+                }
+
+                if (!cachedEntry[fxRidEntry].Contains(top)) // Dictionary value is a List. If perf. is bad, change to HashSet.
+                {
+                    cachedEntry[fxRidEntry].Add(top);
+                }
+
+                // Upsert Transitive Origins Cache
+                transitiveOriginsCache[current.PackageIdentity.Id] = cachedEntry;
 
                 foreach (PackageDependency dep in node.Dependencies.ToList()) // Casting to list to prevent backing allocations
                 {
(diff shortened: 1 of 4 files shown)


src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
```diff
@@ -218,7 +219,27 @@ internal async Task<ProjectPackages> GetInstalledAndTransitivePackagesAsync(bool
                             targetsList = await GetTargetsListAsync(assetsFilePath, token);
                         }
 
-                        transitiveOrigins = calculatedTransitivePackages.Any() ? ComputeTransitivePackageOrigins(calculatedInstalledPackages, targetsList, token) : new Dictionary<string, TransitiveEntry>();
+                        List<LockFileTargetLibrary> projectReferences = targetsList.SelectMany(t => t.Libraries).Where(l => l.Type == LibraryType.Project).ToList();
+                        List<PackageReference> installedProjectReferences = new List<PackageReference>(projectReferences.Count());
```
kartheekp-ms:
```suggestion
                        List<PackageReference> installedProjectReferences = new List<PackageReference>(projectReferences.Count);
```

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
```diff
@@ -218,7 +219,27 @@ internal async Task<ProjectPackages> GetInstalledAndTransitivePackagesAsync(bool
...
+
+                        // Since ProjectReferences are not in the dependencies section of the package spec (installed packages), we need to manually add them
+                        // to the installed packages list to compute transitive origins
+                        foreach (var projectReference in projectReferences)
+                        {
+                            var packageReference = new PackageReference(
```
jebriede: We probably should have a new type for modeling project references. Please see my comment in the one-pager for details: https://github.com/NuGet/Client.Engineering/pull/2934#discussion_r1680165077

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
```diff
@@ -218,7 +219,27 @@ internal async Task<ProjectPackages> GetInstalledAndTransitivePackagesAsync(bool
...
+                                identity: new PackageIdentity(projectReference.Name, projectReference.Version),
+                                targetFramework: new NuGetFramework(projectReference.Framework),
+                                userInstalled: false,
+                                developmentDependency: false,
+                                requireReinstallation: false,
+                                allowedVersions: VersionRange.Parse(projectReference.Version.ToString()));
+                            installedProjectReferences.Add(packageReference);
+                        }
+
+                        installedProjectReferences.AddRange(calculatedInstalledPackages);
```
jebriede: It seems strange to add calculatedInstalledPackages to a list called installedProjectReferences since these calculatedInstalledPackages are packages, not projects. We should rename the variable but this also indicates to me that we're mixing the concept of projects and packages here and treating projects as packages which may limit future changes.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
jebriede: Instantiating the list of installedProjectReferences to the number of project references would be useful to avoid array expansion if we only added the number of project references to this list but we're adding the calculatedInstalledPackages to it later. I'm not sure if the AddRange implementation considers the whole size of the new array when reallocating as to avoid multiple reallocations or not.

jebriede (changes requested):
I have concerns about treating the project reference as a package reference and how we're using the transitive origins cache but not adding the project reference to it, so that it doesn't get displayed in the PM UI. Another approach would be to build in support for Project References to set us up for future extension without losing context later. Additionally, the transitive origin cache shouldn't be what determines what is displayed in the UI. Rather, the business logic for the view should decide what it should show.

Something else to check with this implementation is how this code will handle a project that matches the name (and version) of a package dependency. For example, say I add a project called Newtsonsoft.Json to my solution and I also reference Newtonsoft.Json directly or transitively. How will this code, that treats projects as the same thing as a package, handle that?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
martinrrm: @nkolev92 Can correct me here, but my understanding is that `<ProjectReferences>` [has a duality](https://github.com/NuGet/Home/wiki/PackageReference-Specification) with `<PackageReferences>` that's why I consider it as an option.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
martinrrm: So `calculatedInstalledPackages` is calculated by looking at the the dgspec file, (calculated from the .csproj) which has different nodes for projects and packages, and then we compare this file against the assets file (which has the same nodes for projects and packages). 

That's also why I had to remove type validation in https://github.com/NuGet/NuGet.Client/pull/5926/files#diff-cf69ebf676e61580349108d38e5dab1bbff91468194c6a8d342149d264139b15L326. Removing just that line makes the PM UI start to render `ProjectReferences` as transitive packages and that is also why added a check for TransitiveOrigins cache in the UI to avoid rendering them https://github.com/NuGet/NuGet.Client/blob/38ab39a6691e6e5d54e29055d9e258b8b5d57220/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs#L1038

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
zivkan: From restore's point of view, they're similar (I guess the goal was to treat them the same, but there are some significant gaps). But basically, transitive packages (and some things like NoWarn) flow through package dependencies just the same as ProjectReferences.

I haven't looked at the code changes in this PR, so I'm just basing my comments on this thread's comments, but consider what does it mean if a project reference appears in the Package Manager UI. What will the version list contain. What if there's a package on one of the feeds with the same ID, should PM UI allow you to "install" the package (change ProjectReference to a PackageReference). Can you uninstall a ProjectReference? If so, do we need to list all solution projects somewhere so you can add project references via PM UI?

The top of the PackageReference spec says that the context is about migrating from project.json and xproj to MSBuild files. I think taking a statement about project & package duality from this document and extrapolating it to "show projects in Package Manager UI's packages list" is taking that statement out of context.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
```diff
@@ -218,7 +219,27 @@ internal async Task<ProjectPackages> GetInstalledAndTransitivePackagesAsync(bool
...
+                        // to the installed packages list to compute transitive origins
+                        foreach (var projectReference in projectReferences)
+                        {
+                            var packageReference = new PackageReference(
+                                identity: new PackageIdentity(projectReference.Name, projectReference.Version),
+                                targetFramework: new NuGetFramework(projectReference.Framework),
+                                userInstalled: false,
+                                developmentDependency: false,
+                                requireReinstallation: false,
+                                allowedVersions: VersionRange.Parse(projectReference.Version.ToString()));
```
nkolev92: Isn't that a version object? Can we just create a version range with the version object as a lower bound?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
```diff
@@ -463,7 +483,8 @@ private static void MarkTransitiveOrigin(Dictionary<string, TransitiveEntry> tra
                     cachedEntry[fxRidEntry] = new List<PackageReference>();
                 }
 
-                if (!cachedEntry[fxRidEntry].Contains(top)) // Dictionary value is a List. If perf. is bad, change to HashSet.
+                // In order to hide Projects from the PM UI, we don't add them to the transitive origins cache
```
nkolev92: Can you help me understand why the transitive origins cache is where the project refs are missing? I'm not sure how that translated to the transitive list of packages that exist. 

Another thing is that we might be mixing the model & view here, since this is really just data. If the view needs to do something different, maybe it needs to be handled there instead.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
martinrrm: `ProjectReferences` are not going to be displayed in the PM UI. 

The current implementation to gather transitive packages and their origin is to look at installed packages from the `PackageSpec` and add their dependencies. The problem here is these installed packages don't have the information for project references. That's why I'm adding them manually to a copy of the installed packages list just to gather their dependencies and calculate the origins.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
martinrrm: I'm not sure if I understand the first question, maybe explaining why we check transitive origins answers you. When we construct the transitive dependencies list, we look at the assets file, and it contains projects and packages dependencies so at the end we have a list with all packages no matter who brough it. 

The problem here is that since we cannot correlate a package to an installed package, since they were brought by a project, we don't have a transitive origin and we don't display them (https://github.com/NuGet/NuGet.Client/pull/5524/files). This caused errors between packages list and vulnerabilities indicators, https://github.com/NuGet/Home/issues/13068. So, at the end transitive origins is the way we differentiate between packages brought by packages and projects.

About mixing model and view I agree that is not the best option, but I'm afraid that if we start using a new type to represent both packages and projects it will be a huge change. Right now, the way we determine transitive packages in the UI is by filtering out `PackageReferences` without transitive origins https://github.com/NuGet/NuGet.Client/pull/5524/files

nkolev92: I think I get what it's supposed to say now. 

Why wouldn't we just try not to return the project in the list of installed packages? Are projects today included in the return type for the list? 
Note that this API is used for the get installed packages extensibility APIs as well, so we want to make sure we're not returning projects there.

martinrrm: Projects are not being added to the returned installed packages list. I'm adding projects to a copy of the installed packages so I can calculate the transitive origin for their packages.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
zivkan: It just occurred to me that this part of PM UI can be simplified.  One part of the assets file contains a flat list of packages (and projects), and another part of the assets file contains a list of direct referenced by the project. There's no need for graph walking.

Something like:
```
foreach (var package in assetsFile.Targets.Where(node => node.Type == "package"))
{
   bool isTransitive = !assetsFile.Package.Dependencies.Contains(package.Id);
}
```

Graph walking (and therefore the transitive origin cache) is only needed when calculating the transitive root, or when we eventually show a tree representing the graph.

martinrrm:
I refactored the code a bit with suggestions and separated the calculation of project dependencies.

These changes do not affect the `InstalledPackages` list, the intention just to add transitive packages from projects. `ProjectReferences` are not being added to the PM UI, at least in this iteration. 

I believe that creating a new type is necessary for this. projects are not going to be added to the PM UI, only their packages.

martinrrm (commented):
If there are still doubts/concerns, I can schedule a meeting so we can go over the changes together.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
martinrrm: Looks like we were limiting the assets file reading to the Targets section, `Libraries` is the one you are mentioning. There is even more information about Project References in `ProjectFileDependencyGroup`. 

Even if we are using Targets section, I believe we shouldn't be calculating the whole graph

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
```diff
@@ -218,7 +219,31 @@ internal async Task<ProjectPackages> GetInstalledAndTransitivePackagesAsync(bool
...
 
-                        transitiveOrigins = calculatedTransitivePackages.Any() ? ComputeTransitivePackageOrigins(calculatedInstalledPackages, targetsList, token) : new Dictionary<string, TransitiveEntry>();
+                        Dictionary<string, TransitiveEntry> transitiveOriginsCache = new();
+
+                        // Compute Transitive Origins
+                        transitiveOrigins = calculatedTransitivePackages.Any() ? ComputeTransitivePackageOrigins(calculatedInstalledPackages, targetsList, transitiveOriginsCache, token) : transitiveOriginsCache;
+
+                        // If the project has project references, we need to compute transitive origins for their packages
+                        List<LockFileTargetLibrary> projectReferences = targetsList.SelectMany(t => t.Libraries).Where(l => l.Type == LibraryType.Project).ToList();
+                        if (projectReferences.Any())
```
kartheekp-ms: I am curious to know what the user experience will be if the referenced project has another project reference which has a package reference.

Project A -> Project B -> Project C (has a package reference to Newtonsoft.Json)

Will the PM UI display `Newtonsoft.Json` as an installed package for `Project A`?

martinrrm: If you open the PM UI for Project A/B it will display `Newtonsoft.Json` as a **transitive** package. That's the only thing that will be in the PM UI packages list.

If you open the PM UI for Project C it will display `Newtonsoft.Json` as a top level package, so the current behavior.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
```diff
@@ -434,7 +456,7 @@ private static void MarkTransitiveOrigin(Dictionary<string, TransitiveEntry> tra
             // Find first target node that matches current
             foreach (LockFileTargetLibrary lib in graph.Libraries)
             {
-                if (lib.Type == LibraryType.Package.Value
+                if ((lib.Type == LibraryType.Package.Value || lib.Type == LibraryType.Project.Value)
```
nkolev92, line 458: Is there any other type?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
```diff
@@ -448,28 +470,32 @@ private static void MarkTransitiveOrigin(Dictionary<string, TransitiveEntry> tra
             {
                 visited.Add(current.PackageIdentity); // visited
 
-                // Lookup Transitive Origins Cache
-                TransitiveEntry cachedEntry;
-                if (!transitiveOriginsCache.TryGetValue(current.PackageIdentity.Id, out cachedEntry))
+                // Set transitive origins if the dependency is a package
+                if (node.Type == LibraryType.Package.Value)
```
nkolev92: Given the whole impact of transitive origins, feels like we need a description of why in particular this needs done. 

Is it somewhere in this class but I haven't found it or are we missing one?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
```diff
@@ -218,7 +219,31 @@ internal async Task<ProjectPackages> GetInstalledAndTransitivePackagesAsync(bool
                             targetsList = await GetTargetsListAsync(assetsFilePath, token);
                         }
 
-                        transitiveOrigins = calculatedTransitivePackages.Any() ? ComputeTransitivePackageOrigins(calculatedInstalledPackages, targetsList, token) : new Dictionary<string, TransitiveEntry>();
+                        Dictionary<string, TransitiveEntry> transitiveOriginsCache = new();
+
+                        // Compute Transitive Origins
+                        transitiveOrigins = calculatedTransitivePackages.Any() ? ComputeTransitivePackageOrigins(calculatedInstalledPackages, targetsList, transitiveOriginsCache, token) : transitiveOriginsCache;
+
+                        // If the project has project references, we need to compute transitive origins for their packages
```
nkolev92, line 222: Wouldn't it be more performant to calculative the transitive origins all at once? 

Right now, we're doing a 2 step and it's not clear to me why it needs to be a 2 step with potentially the same packages.

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
```diff
@@ -4053,9 +4053,36 @@ public async Task GetInstalledAndTransitivePackagesAsync_ProjectReferenceWithSin
...
-            Assert.Empty(result.TransitivePackages);
+            Assert.Equal(1, result.TransitivePackages.Count);
+        }
+
+        [Fact]
+        public async Task GetInstalledAndTransitivePackagesAsync_MutipleProjectReferencesWithSinglePackage_EmptyInstalledAndTransitivePackagesAsync()
+        {
+            // Project3 -> Project2 ->  PackageA (1.0.0)
+            //                          Project1 -> PackageB (1.0.0)
+            // PackageA (1.0.0) -> PackageB (1.0.0)
```
nkolev92: Doesn't this test the scenario with both a project and package referencing Package B?

In general, for transitive packages, we have 3 options: 
* Through package only
* Through project only
* Through project & package.

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
```diff
@@ -4053,9 +4053,36 @@ public async Task GetInstalledAndTransitivePackagesAsync_ProjectReferenceWithSin
 
             // Assert
             Assert.Empty(result.InstalledPackages); // No installed packages
-            // Foreign transitive packages in assets file don't have dependencies.
-            // Those are not considered transitive packages
-            Assert.Empty(result.TransitivePackages);
+            Assert.Equal(1, result.TransitivePackages.Count);
```
nkolev92, line 4055: In this test, there's a mention of PackageA in the test, but package A does not exist.

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
```diff
@@ -4053,9 +4053,36 @@ public async Task GetInstalledAndTransitivePackagesAsync_ProjectReferenceWithSin
...
+
+            CpsPackageReferenceProject project3 = PrepareCpsRestoredProject(prj3Spec);
+
+            // Act
+            ProjectPackages result = await project3.GetInstalledAndTransitivePackagesAsync(includeTransitiveOrigins: true, CancellationToken.None);
+
+            // Assert
+            Assert.Empty(result.InstalledPackages); // No installed packages
+            // Inner project references need to be considered as transitive packages in order to calculate transitive origins.
+            Assert.Equal(3, result.TransitivePackages.Count);
```
nkolev92: If we're calculating the transitive origins, can we verify they're correct?

nkolev92 (commented):
I feel like most of my comments are consequences of the previous implementation more so than this one. 

Some test feedback as well.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
martinrrm: Possible types: Project, ExternalProject, Package, Assembly, Reference, WinMD, Unresolved

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
martinrrm: Since we only display packages, there is no need to calculate the transitive origin of a project.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
martinrrm: @zivkan With that suggestion I was able to remove several graph iterations and removed nodes that weren't packages.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
martinrrm: Removed this line of code in new version, transitive packages list contained projects, I was able to remove them way before hitting this code path

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Utility/GetPackageReferenceUtility.cs
```diff
@@ -60,38 +60,30 @@ internal static PackageIdentity UpdateResolvedVersion(LibraryDependency projectL
...
+        /// Updates the transitive dependencies for the project.
         /// </summary>
-        /// <param name="library">Library from the project file.</param>
-        /// <param name="targetFramework">Target framework from the project file.</param>
-        /// <param name="targets">Target assets file with the package information.</param>
+        /// <param name="libraries">Libraries from the project file.</param>
         /// <param name="installedPackages">Cached installed package information</param>
         /// <param name="transitivePackages">Cached transitive package information</param>
-        internal static IReadOnlyList<PackageIdentity> UpdateTransitiveDependencies(LockFileTargetLibrary library, NuGetFramework targetFramework, IList<LockFileTarget> targets, Dictionary<string, ProjectInstalledPackage> installedPackages, Dictionary<string, ProjectInstalledPackage> transitivePackages)
+        internal static IReadOnlyList<PackageIdentity> UpdateTransitiveDependencies(IList<LockFileTargetLibrary> libraries, Dictionary<string, ProjectInstalledPackage> installedPackages, Dictionary<string, ProjectInstalledPackage> transitivePackages)
```
martinrrm, line 68: This change is key for the PR, with previous [Andy's comment](https://github.com/NuGet/NuGet.Client/pull/5926#discussion_r1681820594) I was able to remove projects from the libraries so this function return's only transitive packages.

Assets file already contains a flat list of all the dependencies for a project, there was no need to try to create a graph.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
martinrrm: This function `MarkTransitiveOrigin` iterates the dependency graph doing a DFS, adding Projects in this if statements enables considering Project's as a valid node to iterate, projects are not added, they are just considered as a valid "parent" node for package's

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
```diff
@@ -322,9 +346,8 @@ internal async Task<ProjectPackages> GetInstalledAndTransitivePackagesAsync(bool
             else
             {
                 return targets
-                    .SelectMany(target => target.Libraries)
-                    .Where(library => library.Type == LibraryType.Package)
-                    .SelectMany(library => GetPackageReferenceUtility.UpdateTransitiveDependencies(library, targetFramework, targets, installedPackages, transitivePackages))
+                    .Where(target => target.TargetFramework.Equals(targetFramework))
+                    .SelectMany(target => GetPackageReferenceUtility.UpdateTransitiveDependencies2(target.Libraries, installedPackages, transitivePackages))
```
martinrrm: We could also use `LockFile.Libraries` instead of `LockFile.Targets` to remove the need of comparing to a target framework since we don't handle tfm's in the PM UI but I will explore that idea in another PR.

nkolev92: Libraries section won't have the dependencies.

martinrrm: Truee, then I'm not sure if it's worth loading Libraries section

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
jebriede: @martinrrm Could you add a comment on why you think this comment should be resolved without being addressed? Looking forward to hearing your perspective! Thanks!

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
```diff
@@ -323,7 +344,6 @@ internal async Task<ProjectPackages> GetInstalledAndTransitivePackagesAsync(bool
             {
                 return targets
                     .SelectMany(target => target.Libraries)
-                    .Where(library => library.Type == LibraryType.Package)
```
jebriede, line 326: What types could be present here? Should this constrain it to Package and Project instead of removing the constraint?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
jebriede: This comment suggests that the transitive origins cache has a direct impact on what is shown in the UI. That should not be. I may be misremembering how it's used. I would love to understand this comment in more detail to make sure the comment is reflective of the code or that we see if there's an implicit effect we need to consider.

jebriede (changes requested):
Thanks for asking for a rereview from me, @martinrrm. I see a few comments from my previous review don't have a resolution yet. Please reach out to me if you'd like to discuss them in further detail!

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
```diff
@@ -218,7 +219,30 @@ internal async Task<ProjectPackages> GetInstalledAndTransitivePackagesAsync(bool
...
+                                    identity: new PackageIdentity(projectReference.Name, projectReference.Version),
+                                    targetFramework: new NuGetFramework(projectReference.Framework),
+                                    userInstalled: false,
+                                    developmentDependency: false,
+                                    requireReinstallation: false,
+                                    allowedVersions: new VersionRange(projectReference.Version));
+                                calculatedLibraryReferences.Add(packageReference);
+                            }
+                        }
+
```
jebriede, line 239: nit: extra blank line

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
```diff
@@ -218,7 +219,30 @@ internal async Task<ProjectPackages> GetInstalledAndTransitivePackagesAsync(bool
                             targetsList = await GetTargetsListAsync(assetsFilePath, token);
                         }
 
-                        transitiveOrigins = calculatedTransitivePackages.Any() ? ComputeTransitivePackageOrigins(calculatedInstalledPackages, targetsList, token) : new Dictionary<string, TransitiveEntry>();
+                        Dictionary<string, TransitiveEntry> transitiveOriginsCache = new();
```
jebriede: Is this being used?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Utility/GetPackageReferenceUtility.cs
```diff
@@ -60,38 +60,30 @@ internal static PackageIdentity UpdateResolvedVersion(LibraryDependency projectL
         }
 
         /// <summary>
-        /// Gets the dependencies of a top level package and caches these if the assets file has changed
+        /// Updates the transitive dependencies for the project.
```
jebriede: This comment looks less descriptive than what we had before. Consider describing the method in a way that is not already achieved by the name of the method.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
jebriede: Looking more closely, this review I did was on a previous commit. Oops! :-) We talked offline and I left a few more comments. We should still consider adding a new data type to handle project references in the future and it would be good to keep track of where we create projects as package references to make it easier to modify if needed.

jebriede (dismissed):
Approving with some comments I left as single comments and we discussed offline. Thanks!

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
```diff
@@ -218,7 +219,28 @@ internal async Task<ProjectPackages> GetInstalledAndTransitivePackagesAsync(bool
                             targetsList = await GetTargetsListAsync(assetsFilePath, token);
                         }
 
-                        transitiveOrigins = calculatedTransitivePackages.Any() ? ComputeTransitivePackageOrigins(calculatedInstalledPackages, targetsList, token) : new Dictionary<string, TransitiveEntry>();
+                        // If the project has project references, we need to compute transitive origins for their packages
+                        List<LockFileTargetLibrary> projectReferences = targetsList.SelectMany(t => t.Libraries).Where(l => l.Type == LibraryType.Project).ToList();
```
nkolev92, line 223: This is just a comment, but I'm wondering how many of our past failures in this logic stem from the fact that dependencies from different frameworks are combined together. 

It can be a separate work, but I'm wondering what would happen if project references are different.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
```diff
@@ -218,7 +219,28 @@ internal async Task<ProjectPackages> GetInstalledAndTransitivePackagesAsync(bool
...
+                        List<PackageReference> calculatedLibraryReferences = new List<PackageReference>(projectReferences.Count + calculatedInstalledPackages.Count);
+                        if (projectReferences.Any())
```
nkolev92: good practice to compare count since you have a list.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Utility/GetPackageReferenceUtility.cs
```diff
@@ -60,38 +60,30 @@ internal static PackageIdentity UpdateResolvedVersion(LibraryDependency projectL
...
         /// <param name="transitivePackages">Cached transitive package information</param>
-        internal static IReadOnlyList<PackageIdentity> UpdateTransitiveDependencies(LockFileTargetLibrary library, NuGetFramework targetFramework, IList<LockFileTarget> targets, Dictionary<string, ProjectInstalledPackage> installedPackages, Dictionary<string, ProjectInstalledPackage> transitivePackages)
+        internal static IReadOnlyList<PackageIdentity> UpdateTransitiveDependencies(IList<LockFileTargetLibrary> libraries, Dictionary<string, ProjectInstalledPackage> installedPackages, Dictionary<string, ProjectInstalledPackage> transitivePackages)
         {
             NuGetVersion resolvedVersion = default;
 
             var packageIdentities = new List<PackageIdentity>();
 
             // get the dependencies for this target framework
-            IReadOnlyList<PackageDependency> transitiveDependencies = GetTransitivePackagesForLibrary(library, targetFramework, targets);
```
nkolev92, line 77: Is this an unused method now?

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
```diff
@@ -4053,9 +4052,51 @@ public async Task GetInstalledAndTransitivePackagesAsync_ProjectReferenceWithSin
...
+            // Assert
+            Assert.Empty(result.InstalledPackages); // No installed packages
+            Assert.Equal(2, result.TransitivePackages.Count);
+
+            foreach (var transitivePackage in result.TransitivePackages)
+            {
+                var origins = transitivePackage.TransitiveOrigins.Select(transitivePackage => transitivePackage.PackageIdentity.Id);
+                if (transitivePackage.PackageIdentity.Id.Equals("PackageB"))
+                {
+                    Assert.Contains("Project1", origins);
```
nkolev92: The order matters here correct?
Can we verify that. 
A bug can easily lead to switching the order incorrectly.

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
```diff
@@ -4053,9 +4052,51 @@ public async Task GetInstalledAndTransitivePackagesAsync_ProjectReferenceWithSin
...
-            // Those are not considered transitive packages
-            Assert.Empty(result.TransitivePackages);
+            Assert.Equal(1, result.TransitivePackages.Count);
+
+            var origins = result.TransitivePackages[0].TransitiveOrigins.Select(transitivePackage => transitivePackage.PackageIdentity.Id);
+            Assert.Contains("Project1", origins);
+        }
+
+        [Fact]
+        public async Task GetInstalledAndTransitivePackagesAsync_MutipleProjectReferencesWithSamePackage_EmptyInstalledAndTransitivePackagesAsync()
```
nkolev92, line 4088: Mentioned it earlier, https://github.com/NuGet/NuGet.Client/pull/5926#discussion_r1686934615, can we add a test where a project has a dependency coming through both project and a package.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
zivkan: honestly, we should probably add `.Any()` (the overload without a predicate) to the banned API list.

Although the way that `calculatedTransitivePackages` is used elsewhere, `Any()` is probably less costly in that specific use case than materializing a collection just so `.Count > 0` can be used.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
zivkan: It wouldn't surprise me if conditional project references are a common scenario for MAUI projects. Or at least, that's where I feel like it's most likely to happen.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
```diff
@@ -235,7 +257,7 @@ internal async Task<ProjectPackages> GetInstalledAndTransitivePackagesAsync(bool
                         {
                             transitiveOrigins.TryGetValue(packageRef.PackageIdentity.Id, out TransitiveEntry cacheEntry);
                             return MergeTransitiveOrigin(packageRef, cacheEntry);
-                        });
+                        }).ToList();
```
zivkan: When `transitivePackagesWithOrigins` is used a few lines down, `.ToList()` is called there. So doing so here causes a second List to be allocated (and depending on the implementation of LINQ's ToList, if it doesn't preallocate a list of the correct size, then it will cause multiple resizes, which will be very costly).

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
nkolev92: Conditional package references are very common in MAUI.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
nkolev92: > honestly, we should probably add `.Any()` (the overload without a predicate) to the banned API list.
> 
> Although the way that `calculatedTransitivePackages` is used elsewhere, `Any()` is probably less costly in that specific use case than materializing a collection just so `.Count > 0` can be used.

I don't mind that.
Curious how much we'd have to change though. 
Not every codepath is performance sensitive.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
martinrrm: What happens when they package references are conditional? I've been trying to test with them and looks like they are not being added to the `Libraries` section of the assets file, is that something expected? I do see the project reference in the Solution Explorer

nkolev92: The assets file there doesn't load the libraries section: https://github.com/NuGet/NuGet.Client/blob/0c714c10851db160a4ca5b80e718612bf669daf3/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs#L414.

It'll only be a part of targets section for the particular framework.

nkolev92: i'm guessing you mgiht've referred to https://github.com/NuGet/NuGet.Client/blob/0c714c10851db160a4ca5b80e718612bf669daf3/src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileTarget.cs#L19. 


The solution explorer reads the assets file, so it probably should be there.

martinrrm: Sorry I meant to say Target section, are they expected to be there?

nkolev92: Yes.

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
martinrrm: I added a test for this comment, let me know if I'm not addressing your concern

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
martinrrm: Tested with conditional ProjectReferences, we will only show transitive packages from a project reference if they are the same target framework as the one the PM UI is using (I believe is the first one in the list, not sure if we order them).

What do you suggest it could be a next step? I came with another idea other than supporting tfm's in the PM UI

zivkan: > we will only show transitive packages from a project reference if they are the same target framework as the one the PM UI is using (I believe is the first one in the list, not sure if we order them)

The code this conversation thread is on is not part of PM UI, but part of the lower level `PackageReferenceProject` that all use cases of `GetInstalledPackagesAsync` uses. It should be merging the package list from all the project's TFMs.  If it's only using 1 TFM, then it's a bug.

Also, I believe that historically PM UI has always shown all (direct) packages, even conditional packages. If that stopped working when transitive packages were added to PM UI, then I believe it's a regression.

martinrrm: I don't expect a change in behavior for installed packages since this PR only changes how we calculate transitive origins. These ProjectReferences are added to a copy and used only for origins calculations.

nkolev92: We should probably check how thorough our automated/manual tests are on this topic. 

It doesn't necessarily have to be blocking this PR, but as we're getting transitives in the solution level PM UI soon, ensuring proper coverage is really important.

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
nkolev92: Looks good.

nkolev92 (approved):

PR #5928, opened 2024-07-17
Display warnings/errors on transitive package/project references

### LINKED ISSUE #13636: Bubble-up Known Vulnerability Indicators in Solution Explorer for Transitive Packages
### NuGet Product(s) Involved

### The Elevator Pitch

NuGet launched a vulnerability auditing feature last year and would like to enhance the experience further with project system help!

Today a user must directly navigate to a transitive package to see a warning produced when `NuGetAuditMode = all`

image

Ideally these warnings would bubble up the entire packages tree similar to what you see with a top-level dependency:

image

At the end of the day, if there is a transitive dependency vulnerability warning, it should show in the `Dependencies` node so the user knows how to spelunk to find the culprit. This will also complement CLI work we did in a command called [`dotnet nuget why`](https://github.com/NuGet/Home/blob/e83e772d7372288fc4475ba8b511dca97768a006/accepted/2022/dotnet-nuget-why-proposal.md) which allows you to do similar.

Many users use the solution explorer to view their dependency tree alongside the newly released [transitive dependencies in visual studio](https://devblogs.microsoft.com/nuget/introducing-transitive-dependencies-in-visual-studio/) functionality that only works for project-level today.

In developer surveys, we found that the solution explorer is one of the most desired places for people to view vulnerability information about their dependencies.

More user impact/motivation can be found in an older proposal that is related more-so to the [iconography of these indicators](https://github.com/NuGet/Home/blob/package-status-indicators/proposed/2022/package-status-indicators.md#motivation)

drewnoakes:
# Bug

Fixes:  https://github.com/NuGet/Home/issues/13636

## Description

Historically we have not shown warning/error icons on transitive package/project reference nodes in the tree. Since the introduction of NuGet Audit, it becomes more important to add these icons, and propagate them up the spine of the tree in order to guide the user from the "Dependencies" node down to the specific node, however deep within the tree.

The implementation builds a map from node to ancestors then walks from nodes with warnings up the tree, remembering the highest severity seen at each level. We then use that value on each node to select an appropriate icon for the log level.

Note that this change does not set icons on top-level dependencies. These come from design-time build data and will need to be handled separately by changes to NuGet's MSBuild tasks/targets.

### Before

image

### After

image

The actual leaf warning here is further down, for example:

image

## PR Checklist

- [x] Added tests

### DIFF
--- a/src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/Models/AssetsFileDependenciesSnapshot.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/Models/AssetsFileDependenciesSnapshot.cs
@@ -4,12 +4,14 @@
 #nullable enable
 
 using System;
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
 using Microsoft;
+using NuGet.Common;
 using NuGet.Packaging;
 using NuGet.ProjectModel;
 
@@ -172,11 +174,20 @@ namespace NuGet.VisualStudio.SolutionExplorer.Models
 
         internal static ImmutableDictionary<string, AssetsFileTargetLibrary> ParseLibraries(LockFile lockFile, LockFileTarget lockFileTarget, ImmutableArray<AssetsFileLogMessage> logMessages)
         {
+            var levelByLibrary = BuildLevelByLibrary();
+
             ImmutableDictionary<string, AssetsFileTargetLibrary>.Builder builder = ImmutableDictionary.CreateBuilder<string, AssetsFileTargetLibrary>(StringComparer.OrdinalIgnoreCase);
 
             foreach (LockFileTargetLibrary lockFileLibrary in lockFileTarget.Libraries)
             {
-                if (AssetsFileTargetLibrary.TryCreate(lockFile, lockFileLibrary, out AssetsFileTargetLibrary? library))
+                LogLevel? logLevel = null;
+
+                if (lockFileLibrary.Name is not null && levelByLibrary.TryGetValue(lockFileLibrary.Name, out var level))
+                {
+                    logLevel = level;
+                }
+
+                if (AssetsFileTargetLibrary.TryCreate(lockFile, lockFileLibrary, logLevel, out AssetsFileTargetLibrary? library))
                 {
                     builder.Add(library.Name, library);
                 }
@@ -196,6 +207,77 @@ namespace NuGet.VisualStudio.SolutionExplorer.Models
             }
 
             return builder.ToImmutable();
+
+            Dictionary<string, LogLevel> BuildLevelByLibrary()
+            {
+                Dictionary<string, HashSet<string>> ancestorsByLibrary = new(StringComparer.OrdinalIgnoreCase);
+
+                // Build a map from child-to-ancestors
+                foreach (LockFileTargetLibrary lockFileLibrary in lockFileTarget.Libraries)
+                {
+                    string? parentId = lockFileLibrary.Name;
+
+                    if (parentId is null)
+                    {
+                        continue;
+                    }
+
+                    foreach (var dep in lockFileLibrary.Dependencies)
+                    {
+                        string childId = dep.Id;
+
+                        if (!ancestorsByLibrary.TryGetValue(childId, out HashSet<string>? ancestors))
+                        {
+                            ancestors = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+                            ancestorsByLibrary.Add(childId, ancestors);
+                        }
+
+                        ancestors.Add(parentId);
+                    }
+                }
+
+                // Walk up the tree from each log message's library, propagating the highest level found.
+                Dictionary<string, LogLevel> levelByLibrary = new(StringComparer.OrdinalIgnoreCase);
+
+                foreach (AssetsFileLogMessage message in logMessages)
+                {
+                    Integrate(message.LibraryName, message.Level, visited: []);
+                }
+
+                return levelByLibrary;
+
+                void Integrate(string id, LogLevel level, HashSet<string> visited)
+                {
+                    if (!visited.Add(id))
+                    {
+                        // Avoid infinite recursion.
+                        return;
+                    }
+
+                    if (!levelByLibrary.TryGetValue(id, out LogLevel currentLevel))
+                    {
+                        // No level yet, so set it.
+                        levelByLibrary[id] = level;
+                    }
+                    else
+                    {
+                        // Higher level is more severe. If we already have a higher level, don't change it.
+                        // Also, if we already have a level, assume we have propagated it to all ancestors and can return.
+                        if (currentLevel >= level)
+                        {
+                            return;
+                        }
+                    }
+
+                    if (ancestorsByLibrary.TryGetValue(id, out HashSet<string>? ancestors))
+                    {
+                        foreach (string ancestor in ancestors)
+                        {
+                            Integrate(ancestor, level, visited);
+                        }
+                    }
+                }
+            }
         }
 
         public bool TryGetTarget(string? target, [NotNullWhen(returnValue: true)] out AssetsFileTarget? targetData)
--- a/src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/Models/AssetsFileDependenciesSnapshot.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/Models/AssetsFileDependenciesSnapshot.cs
@@ -261,8 +261,9 @@ namespace NuGet.VisualStudio.SolutionExplorer.Models
                     }
                     else
                     {
-                        // Higher level is more severe. If we already have a higher level, don't change it.
-                        // Also, if we already have a level, assume we have propagated it to all ancestors and can return.
+                        // Higher level is more severe.
+                        // - If we already have a higher level, don't change it.
+                        // - If the level matches, we will have already propagated it to ancestors, so can return.
                         if (currentLevel >= level)
                         {
                             return;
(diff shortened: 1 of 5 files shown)


drewnoakes:
@zivkan is there an existing issue for this to link against?

src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/Models/AssetsFileDependenciesSnapshot.cs
```diff
@@ -196,6 +207,77 @@ static ImmutableArray<AssetsFileLogMessage> ParseLogMessages(LockFile lockFile,
...
+                Dictionary<string, LogLevel> levelByLibrary = new(StringComparer.OrdinalIgnoreCase);
+
+                foreach (AssetsFileLogMessage message in logMessages)
+                {
+                    Integrate(message.LibraryName, message.Level, visited: []);
+                }
+
+                return levelByLibrary;
+
+                void Integrate(string id, LogLevel level, HashSet<string> visited)
```
Nigusu-Allehu, line 255: I think it would be better if we make this non-recursive so that we don't run into a stack overflow problem,

Nigusu-Allehu: We highly likely won't have a tree that deep but better safe

src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/Models/AssetsFileDependenciesSnapshot.cs
```diff
@@ -196,6 +207,77 @@ static ImmutableArray<AssetsFileLogMessage> ParseLogMessages(LockFile lockFile,
             }
 
             return builder.ToImmutable();
+
+            Dictionary<string, LogLevel> BuildLevelByLibrary()
```
nkolev92, line 217: Can we get some tests for this logic?

src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/Models/AssetsFileDependenciesSnapshot.cs
```diff
@@ -196,6 +207,77 @@ static ImmutableArray<AssetsFileLogMessage> ParseLogMessages(LockFile lockFile,
...
+                        // Also, if we already have a level, assume we have propagated it to all ancestors and can return.
+                        if (currentLevel >= level)
+                        {
+                            return;
+                        }
+                    }
+
+                    if (ancestorsByLibrary.TryGetValue(id, out HashSet<string>? ancestors))
+                    {
+                        foreach (string ancestor in ancestors)
```
Nigusu-Allehu, line 281: When we create the ancestors set,  could we have them sorted (have the largest one first) by logLevel? Just wondering if that that might allow us to select the highest loglevel first and avoid more recursion

src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/Models/AssetsFileDependenciesSnapshot.cs
drewnoakes: I looked at testing but wasn't sure how I could do so in the time I carved out for this. Any pointers on how I could spin one up without a lot of boilerplate?

src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/Models/AssetsFileDependenciesSnapshot.cs
drewnoakes: I considered this, but the graph would need to be 3000+ references deep to hit a stack overflow exception. Avoiding recursion requires heap allocation. I don't think it's a good tradeoff given the situation here. I'm already slightly concerned about the additional allocations needed here, especially for their impact on search in Solution Explorer which triggers all these snapshots to be created across the solution.

src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/Models/AssetsFileDependenciesSnapshot.cs
drewnoakes: Interesting idea. The only way I could see this making a difference is sorting the top-level log messages by descending severity, to increase the chance that later walks up the graph short-circuit. I don't expect it'd have a lot of impact, but I can try it.

src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/Models/AssetsFileDependenciesSnapshot.cs
nkolev92: We have some test cases that just add a sample project.assets.json resource, and just use that as a predetermined project setup. https://github.com/NuGet/NuGet.Client/blob/dev/test/NuGet.Core.Tests/NuGet.Build.Tasks.Pack.Test/compiler/resources/project.assets.json

There are ways to create your own within the test as well: 

https://github.com/NuGet/NuGet.Client/blob/6b9e4481b9c23665ceb3192b9964921bcbb67c30/test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/RestoreCommandTests.cs#L3220-L3235

The project declaration can also be read like: 

https://github.com/NuGet/NuGet.Client/blob/6b9e4481b9c23665ceb3192b9964921bcbb67c30/test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/RestoreCommandTests.cs#L1847-L1871

The test packages have dependencies elements so you can create a graph as you see fit.

src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/Models/AssetsFileDependenciesSnapshot.cs
```diff
@@ -196,6 +207,77 @@ static ImmutableArray<AssetsFileLogMessage> ParseLogMessages(LockFile lockFile,
...
+
+                    if (!levelByLibrary.TryGetValue(id, out LogLevel currentLevel))
+                    {
+                        // No level yet, so set it.
+                        levelByLibrary[id] = level;
+                    }
+                    else
+                    {
+                        // Higher level is more severe. If we already have a higher level, don't change it.
+                        // Also, if we already have a level, assume we have propagated it to all ancestors and can return.
```
donnie-msft: How can this assumption be made? Because the ordering of the messages in the assets file imply that ancestors come first?

donnie-msft (commented):
Will re-review once the discussion about tests is resolved.

src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/Models/AssetsFileDependenciesSnapshot.cs
drewnoakes: All nodes start with no level. Any time a level is assigned, it's pushed to all ancestors. So if, when walking, you find a node with a level, you know that it's already present on all ancestors.

src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/Models/AssetsFileDependenciesSnapshot.cs
drewnoakes: To test this I needed a unique assets file. I brought one in to the unit test directly.

Tests now added.

src/NuGet.Clients/NuGet.VisualStudio.Implementation/SolutionExplorer/Models/AssetsFileDependenciesSnapshot.cs
drewnoakes: I modified the comment. Hopefully it's clearer now.

test/NuGet.Clients.Tests/NuGet.VisualStudio.Implementation.Test/SolutionExplorer/Models/AssetsFileDependenciesSnapshotTests.cs
```diff
@@ -13,6 +13,4907 @@ namespace NuGet.VisualStudio.Implementation.Test.SolutionExplorer.Models
 {
     public class AssetsFileDependenciesSnapshotTests
     {
+        [Fact]
+        public void TransitiveDependencyWarningsPropagate()
+        {
+            #region Test data
+
+            var lockFileContent = """
```
nkolev92, line 21: The tests might be easier to read if the assets file were embedded instead of hardcoded as strings, like: https://github.com/NuGet/NuGet.Client/blob/dev/test/NuGet.Core.Tests/NuGet.Build.Tasks.Pack.Test/compiler/resources/project.assets.json.

drewnoakes: I considered that but put it here so it's immediately next to the consumption.

zivkan: If the test data was less than one or two hundred lines, I could see it being useful next to the code that uses it. But at nearly 5000 lines, my personal opinion is that it's counterproductive, especially since it's JSON. As a string in-line, VS won't give me JSON navigation tools, like being able to collapse part of the document. In VSCode, the JSON view also shows the JSON Path to the cursor's location, which also aids in navigating large JSON files.

zivkan (approved):
We can move the json test data to a resource in another PR. I'll leae this open for a few days to give other people time to review (especially since it's the weekend now), but I think it's ready to merge.

Thank you for adding a test. I think they're important to catch regressions, especially if someone else needs to start maintaining the code in the future.

PR #5930, opened 2024-07-22
Attempt to move several data structures to be immutable

### LINKED ISSUE #13647: PERF: NuGet Cloning operations are showing heavily in allocations during VS solution load
### NuGet Product Used

Other/NA

### Product Version

VS17

### Worked before?

N/A

### Impact

### Repro Steps & Context

1) Open VS, wait for everything to settle
2) Start profiler
3) Open vs-platform solution
4) Wait for things to settle, stop profiler

*** interesting bit from memory profile ***
image

### Verbose Logs

_No response_

ToddGrun:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13647
Reduces allocations during solution load in VS

## Description
Modified several data structures to be immutable, such that the PackageSpec.Clone operation can allocate significantly less. The scenario is outlined in the bug, but it's essentially cutting a 2-3% of allocations of solution load from the solutions I've tested.

This change is a bit viral, and ended up being more involved than I  had originally thought it would be, but it feels like this ends up not only being more performant, but also lends these data structures to be more sanely reasoned about.

*** before changes ***
image

*** with changes ***
image

## PR Checklist

- [x] Added tests


src/NuGet.Core/NuGet.ProjectModel/TargetFrameworkInformation.cs
```diff
@@ -13,80 +13,72 @@ namespace NuGet.ProjectModel
 {
     public class TargetFrameworkInformation : IEquatable<TargetFrameworkInformation>
```
ToddGrun, line 14: This is one of the classes made immutable

src/NuGet.Core/NuGet.LibraryModel/LibraryRange.cs
```diff
@@ -33,12 +33,20 @@ public LibraryRange(string name, VersionRange? versionRange, LibraryDependencyTa
             TypeConstraint = typeConstraint;
         }
 
-        public required string Name { get; set; }
+        [SetsRequiredMembers]
+        internal LibraryRange(LibraryRange other)
```
ToddGrun: This is one of the classes made immutable

src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
```diff
@@ -14,55 +14,46 @@ namespace NuGet.LibraryModel
 {
     public class LibraryDependency : IEquatable<LibraryDependency>
```
ToddGrun, line 13: This is one of the classes made immutable

src/NuGet.Core/NuGet.ProjectModel/PackageSpec.cs
```diff
@@ -258,12 +258,12 @@ public PackageSpec Clone()
             else
             {
                 targetFrameworks = new List<TargetFrameworkInformation>(TargetFrameworks.Count);
-                targetFrameworks.AddRange(TargetFrameworks.Select(item => item.Clone()));
+                targetFrameworks.AddRange(TargetFrameworks);
```
ToddGrun, line 261: >targetFrameworks.AddRange(TargetFrameworks); [](http://example.com/codeflow?start=16&length=44)

Here's one of the Clone calls that was heavily allocating

src/NuGet.Core/NuGet.ProjectModel/PackageSpec.cs
```diff
@@ -258,12 +258,12 @@ public PackageSpec Clone()
...
             }
 
             return new PackageSpec(
                 targetFrameworks,
-                Dependencies?.Select(item => item.Clone()).ToList(),
```
ToddGrun, line 266: >item.Clone() [](http://example.com/codeflow?start=45&length=12)

Another Clone call that was heavily allocating

src/NuGet.Core/NuGet.ProjectModel/PackageSpec.cs
```diff
@@ -258,12 +258,12 @@ public PackageSpec Clone()
...
+                Dependencies?.ToList(),
                 RuntimeGraph?.Clone(),
```
ToddGrun, line 267: >Clone [](http://example.com/codeflow?start=30&length=5)

A whole bunch of other cloning going on here, but those allocate far less, and this is already borderline too much for me to change.

src/NuGet.Core/NuGet.LibraryModel/PublicAPI.Shipped.txt
```diff
@@ -44,31 +44,18 @@ NuGet.LibraryModel.Library.this[string! key].get -> object!
 NuGet.LibraryModel.Library.this[string! key].set -> void
 NuGet.LibraryModel.LibraryDependency
 NuGet.LibraryModel.LibraryDependency.Aliases.get -> string?
-NuGet.LibraryModel.LibraryDependency.Aliases.set -> void
```
ToddGrun, line 47: So many API changes, mostly because a whole bunch of setters moved to inits (and switches to immutable data structures)

src/NuGet.Core/NuGet.ProjectModel/PackageSpecReferenceDependencyProvider.cs
```diff
@@ -207,6 +207,7 @@ private void AddLibraryProperties(Library library, PackageSpec packageSpec, NuGe
                 library[KnownLibraryProperties.FrameworkAssemblies] = frameworkAssemblies;
 
                 // Add framework references
+                // REVIEW: This is a snapshot of the framework references now, is that ok?
```
ToddGrun: >// REVIEW: This is a snapshot of the framework references now, is that ok? [](http://example.com/codeflow?start=16&length=74)

I wasn't certain about the effects of my change on code in this file.

src/NuGet.Core/NuGet.ProjectModel/CentralTransitiveDependencyGroup.cs
```diff
@@ -28,7 +28,7 @@ public CentralTransitiveDependencyGroup(NuGetFramework framework, IEnumerable<Li
 
         public string FrameworkName { get; }
 
-        public IEnumerable<LibraryDependency> TransitiveDependencies { get; }
+        public IList<LibraryDependency> TransitiveDependencies { get; }
```
nkolev92: Do we benefit from IList anywhere where this code is used? 

The reason why I'm asking is because a foreach would usually allocate on an IList the same way as IEnumerable would. 

So unless we're using it as such, I'm guessing this doesn't change the total allocations right?

I'm basically trying to figure out if we can minimize the number of breaking chnages..

src/NuGet.Core/NuGet.ProjectModel/PackageSpecReferenceDependencyProvider.cs
nkolev92: In general, the PackageSpec should be generated and never be modified during a restore operation.

This is basically the framework references declared by the project, so this should be the same.

src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
nkolev92: Because we can't enforce immutability, in the past we've left comments such as https://github.com/NuGet/NuGet.Client/blob/46f80a3f1a83f3c5f33a20fa9fb72871e1e2d4eb/src/NuGet.Core/NuGet.Packaging/RuntimeModel/RuntimeDependencySet.cs#L11-L13.

We can consider that.

src/NuGet.Core/NuGet.LibraryModel/LibraryRange.cs
```diff
@@ -170,5 +178,44 @@ public override int GetHashCode()
...
+                return this;
+            }
+
+            return new LibraryRange(this)
+            {
+                TypeConstraint = typeConstraint
+            };
+        }
+
+        public LibraryRange WithName(string name)
```
nkolev92: This seems unused.

src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
```diff
@@ -149,46 +160,116 @@ public bool Equals(LibraryDependency? other)
...
-                foreach (LibraryDependency d in packageReferences.Where(d => !d.AutoReferenced && d.LibraryRange.VersionRange == null))
+                LibraryDependency d = packageReferences[i];
+                if (!d.AutoReferenced && d.LibraryRange.VersionRange == null)
                 {
                     if (d.VersionOverride != null)
                     {
-                        d.LibraryRange.VersionRange = d.VersionOverride;
-
-                        continue;
+                        var newLibraryRange = d.LibraryRange.WithVersionRange(d.VersionOverride);
```
nkolev92:
```suggestion
                        LibraryRange newLibraryRange = d.LibraryRange.WithVersionRange(d.VersionOverride);
```

src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
```diff
@@ -149,46 +160,116 @@ public bool Equals(LibraryDependency? other)
...
             {
                 throw new ArgumentNullException(nameof(centralPackageVersions));
             }
-            if (centralPackageVersions.Count > 0)
+            if (centralPackageVersions.Count == 0)
+            {
+                return packageReferences;
+            }
+
+            LibraryDependency[] result = new LibraryDependency[packageReferences.Length];
```
nkolev92: This technically allocates more than before. 
This is only really called during the creation of the package spec. 

I think we can invert the logic. 
Read the Central versions first and then use them to build the library dependencies instead.
In that case, we don't have to copy the array.

src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
```diff
@@ -149,46 +160,116 @@ public bool Equals(LibraryDependency? other)
...
+                    }
                 }
+
+                result[i] = d;
+            }
+
+            return ImmutableCollectionsMarshal.AsImmutableArray(result);
+        }
+
+        public LibraryDependency WithIncludeType(LibraryIncludeFlags includeType)
```
nkolev92: I looked at the usages here, I'm guessing the MSBuildRestoreUtility one can be tweaked to avoid the object copying.

src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
```diff
@@ -149,46 +160,116 @@ public bool Equals(LibraryDependency? other)
...
+                return this;
             }
+
+            return new LibraryDependency(this)
+            {
+                IncludeType = includeType
+            };
+        }
+
+        public LibraryDependency WithSuppressParent(LibraryIncludeFlags suppressParent)
```
nkolev92: This is always called with the WithIncludeType as well, so it probably ends up creating 1 throwaway object than it needs to.

src/NuGet.Core/NuGet.LibraryModel/LibraryRange.cs
nkolev92: At least in non-test code.

src/NuGet.Core/NuGet.Common/MsBuildStringUtility.cs
```diff
@@ -89,28 +91,51 @@ public static bool IsTrueOrEmpty(string? value)
...
-        public static IList<NuGetLogCode> GetNuGetLogCodes(string s)
+        public static ImmutableArray<NuGetLogCode> GetNuGetLogCodes(string s)
         {
             // The Split() method already checks for an empty string and returns Array.Empty<string>().
             string[] split = MSBuildStringUtility.Split(s, ';', ',');
 
             if (split.Length == 0)
             {
-                return Array.Empty<NuGetLogCode>();
+                return [];
```
nkolev92, line 101: Does this create a new immutable array every time? 

Can we use a static instead?

src/NuGet.Core/NuGet.Common/MsBuildStringUtility.cs
```diff
@@ -89,28 +91,51 @@ public static bool IsTrueOrEmpty(string? value)
...
-                    logCodes.Add(logCode);
+                    if (logCodes == null)
+                    {
+                        logCodes = ArrayPool<NuGetLogCode>.Shared.Rent(split.Length);
+                    }
+                    else if (logCodes.Length == index)
+                    {
+                        var oldItems = logCodes;
+
+                        logCodes = ArrayPool<NuGetLogCode>.Shared.Rent(logCodes.Length * 2);
```
nkolev92: how can this ever happen? 

Wouldn't array pool return an array of at least that size?

src/NuGet.Core/NuGet.ProjectModel/PackageSpec.cs
nkolev92: A lot of the other types are almost always empty, and a bunch of them are already Obsolete in the PackageSpec.

src/NuGet.Core/NuGet.ProjectModel/TargetFrameworkInformation.cs
```diff
@@ -143,9 +135,170 @@ public bool Equals(TargetFrameworkInformation other)
                    StringComparer.OrdinalIgnoreCase.Equals(TargetAlias, other.TargetAlias);
         }
 
-        public TargetFrameworkInformation Clone()
+        /// <summary>
+        /// Returns copy of this with specified versions defined in the Central package versions management file. 
+        /// </summary>
+        public TargetFrameworkInformation WithCentralPackageVersions(ImmutableDictionary<string, CentralPackageVersion> versions)
```
nkolev92: Similar to some of the others, we probably don't need this used in more than 1 place. 
The places where we're generating the PackageSpec, such as LegacyPackageReferenceProject, VSNominationUtilities, MSBuildStaticGraphRestore and MSBuildRestoreUtility.

src/NuGet.Core/NuGet.ProjectModel/TargetFrameworkInformation.cs
```diff
@@ -143,9 +135,170 @@ public bool Equals(TargetFrameworkInformation other)
...
+            return new TargetFrameworkInformation(this)
+            {
+                CentralPackageVersions = versions
+            };
+        }
+
+        /// <summary>
+        /// Returns copy of this with specified framework references. 
+        /// </summary>
+        public TargetFrameworkInformation WithFrameworkReferences(ImmutableHashSet<FrameworkDependency> references)
```
nkolev92: We could probably get rid of this as it's only used in the codepaths that are doing a package spec construciton

src/NuGet.Core/NuGet.ProjectModel/TargetFrameworkInformation.cs
```diff
@@ -143,9 +135,170 @@ public bool Equals(TargetFrameworkInformation other)
...
+            return new TargetFrameworkInformation(this)
+            {
+                FrameworkReferences = references
+            };
+        }
+
+        /// <summary>
+        /// Returns copy of this with specified download dependencies 
+        /// </summary>
+        public TargetFrameworkInformation WithDownloadDependencies(ImmutableArray<DownloadDependency> dependencies)
```
nkolev92: Same as others.

src/NuGet.Core/NuGet.ProjectModel/TargetFrameworkInformation.cs
```diff
@@ -143,9 +135,170 @@ public bool Equals(TargetFrameworkInformation other)
...
+            return new TargetFrameworkInformation(this)
+            {
+                DownloadDependencies = dependencies
+            };
+        }
+
+        /// <summary>
+        /// Returns copy of this with specified asset target fallback
+        /// </summary>
+        public TargetFrameworkInformation WithAssetTargetFallback(bool assetTargetFallback)
```
nkolev92: Same as the others by transitivity of the types being used.

src/NuGet.Core/NuGet.ProjectModel/TargetFrameworkInformation.cs
```diff
@@ -143,9 +135,170 @@ public bool Equals(TargetFrameworkInformation other)
...
+            return new TargetFrameworkInformation(this)
+            {
+                AssetTargetFallback = assetTargetFallback
+            };
+        }
+
+        /// <summary>
+        /// Returns copy of this with specified warn
+        /// </summary>
+        public TargetFrameworkInformation WithWarn(bool warn)
```
nkolev92: I didn't go through the ones below this method, but I think similar logic applies.

nkolev92 (commented):
If public APIs were not a consideration, this would be something very easy to accept. 

We'll need to internally decide about the risk. 

I did leave some feedback on the implementation. 
Feel free to ping me offline you need any more clarifications.

src/NuGet.Core/NuGet.ProjectModel/CentralTransitiveDependencyGroup.cs
ToddGrun: One place I see is that PackTaskLogic.AddDependecies takes this in as an IList, and updates the list in place (rather than mutating the LibraryDependency directly)

src/NuGet.Core/NuGet.LibraryModel/LibraryRange.cs
ToddGrun: Was trying to be complete, but I can remove this method and change the test code if preferred.

src/NuGet.Core/NuGet.ProjectModel/CentralTransitiveDependencyGroup.cs
nkolev92: I think that's a bug in the pack code. 
Pack should be appending the central transitive groups, not editing them in place.

src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
ToddGrun: I originally started down that path, and things got complicated as there are several callers of this, and not all of the them are directly updating a targetFrameworkInformation that they had just created.

There could very likely be room for improvement in some of the immutable constructions, but it feels like those changes could be made separately.

src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
ToddGrun: Very likely something that could be optimized, just didn't show up at all in the measurements that I took, and I figured simpler/cleaner is better to start.

src/NuGet.Core/NuGet.Common/MsBuildStringUtility.cs
ToddGrun: Yeah, you're right, this can be simplified.

src/NuGet.Core/NuGet.Common/MsBuildStringUtility.cs
ToddGrun: No, this doesn't allocate. Per sharplab:

    public ImmutableArray<int> M() {
        return [];
    }

Translates to

    public ImmutableArray<int> M()
    {
        return ImmutableCollectionsMarshal.AsImmutableArray(Array.Empty<int>());
    }

src/NuGet.Core/NuGet.LibraryModel/LibraryRange.cs
nkolev92: It's a balance for sure. 
Given the original APIs weren't thoroughly reviewed, we've tended to just keep as small of an API footprint as possible, whatever gets used.

src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
nkolev92: Ideally we'd do that in the same release cycle as this particular change. 
We're making a breaking change now and introducing this new API. 
So we'd aim to make that change so it doesn't have to be a breaking change as well.

Directory.Packages.props
```diff
@@ -174,6 +175,7 @@
       <_allowBuildFromSourcePackage Include="Microsoft.Extensions.FileSystemGlobbing" />
       <_allowBuildFromSourcePackage Include="Microsoft.Web.Xdt" />
       <_allowBuildFromSourcePackage Include="Newtonsoft.Json" />
+      <_allowBuildFromSourcePackage Include="System.Collections.Immutable" />
```
zivkan, line 194: Actually, I just realised that this assembly is part of the .NET (Core) BCL. So, I think we can delete this line, and on the `<PackageReferences` add `Condition=" '$(TargetFrameworkIdentifier)' != '.NETCoreApp' "`. 

Although, several of the projects that reference the package, I think they just target `netstandard2.0`, so the packages we publish to nuget.org won't benefit from reduced dependencies for a net5+ TFM.

src/NuGet.Core/NuGet.Commands/Utility/AssetTargetFallbackUtility.cs
```diff
@@ -59,26 +58,29 @@ public static NuGetFramework GetFallbackFramework(NuGetFramework projectFramewor
...
             if (assetTargetFallback?.Any() == true)
             {
                 // AssetTargetFallback
-                targetFrameworkInfo.Imports = assetTargetFallback.AsList();
-                targetFrameworkInfo.AssetTargetFallback = true;
-                targetFrameworkInfo.Warn = true;
+                targetFrameworkInfo = targetFrameworkInfo
+                    .WithImports(assetTargetFallback.ToImmutableArray())
+                    .WithAssetTargetFallback(true)
+                    .WithWarn(true);
```
zivkan: I'm just one person, so I don't speak for the whole team, but I think this is a good example of why we should not have `.With*` methods. Each of these are copy-on-write (CoW) and therefore we're allocating 3 TargetFrameworkInfo classes on the heap, just to immeidately throw 2 away to the GC.

```
targetFrameworkInfo = new TargetFrameworkInfo(targetFrameworkInfo)
{
  Imports = assetTargetFallback.ToImmutableArray(),
  AssetTargetFallback = true,
  Warn = true
};
```
would allocate only one.

I know the `With*` methods make it neat to do this kind of chaining, or to write `value = value.WithSomething(else);` and a single line expression. But if doing `value = new Type(value) { Something = else };` (or more likely this statement over 4 lines) is "painful", then that's a feature because it will encourage us to stop unnecesary allocations via CoW, and just use stack variables followed by a single new  at the end of the method, which is a common pattern in our JSON parsing code.

Hiding the `new` allocation inside the `With*` command also might mean some people are not aware it allocates, whereas if the code is `new SomeType`, then it becomes very clear and I think people will be more careful about the number of times they do a CoW allocation.

Therefore, my recommendation is to delete all the `With*` methods. C# language features, particularly `init` setters, I think are sufficient.  This PR can use the longer syntax CoW, and we'll leave the refactoring to reduce the number of CoW allocations for another PR.  When I do a text search for `.With` on this PR, I see 84 matches (ironically, the `.With*` methods will make it easier to find the CoW allocations 🤣 )

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
```diff
@@ -289,22 +292,23 @@ internal static IEnumerable<DownloadDependency> GetPackageDownloads(IMSBuildProj
...
+        internal static ImmutableArray<LibraryDependency> GetPackageReferences(IMSBuildProject project, bool isCentralPackageVersionManagementEnabled)
         {
             // Get the distinct PackageReference items, ignoring duplicates
             List<IMSBuildItem> packageReferenceItems = GetDistinctItemsOrEmpty(project, "PackageReference").ToList();
 
-            var libraryDependencies = new List<LibraryDependency>(packageReferenceItems.Count);
+            var libraryDependencies = new LibraryDependency[packageReferenceItems.Count];
+            var libraryDependencyIndex = 0;
 
             foreach (var packageReferenceItem in packageReferenceItems)
```
zivkan: Just a personal preference, but I prefer changing these `foreach` loops to a `for` loop, because I think it's much more clear that it's a 1:1 mapping. I also think people generally have a `foreach` obsession, and then do unnecessarily complex stuff like avoiding `IReadOnly*` because of `foreach` allocations, when a simple `for` loop with index lookup would also be a zero enumeration 🤣 

```
for (int libraryDepdencyIndex = 0; libraryDependencyIndex < packageReferenceItems.Count; libraryDependencyIndex++)
{
  var packageReferenceItem = packageReferenceItems[libraryDependencyIndex];
  libraryDependencues[libraryDependencyIndex] = Transform(packageReferenceItem);
}
```

To me, using a `foreach` loop makes it feel like the index isn't important, so would be better suited with a `List<T>` and `.Add()` (moreso in a map-reduce scenario, where some items are filtered out). But given Restore is NuGet's most perf sensitive.

Anyway, I do recognise this comment is very much a personal preference. Just my 2c for writing code that might be easier to read.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/AddPackageReferenceCommandRunner.cs
```diff
@@ -305,31 +306,37 @@ public static async Task<NuGetVersion> GetLatestVersionAsync(PackageSpec origina
...
+                var orderedFrameworksWithOriginalIndex = project.TargetFrameworks
+                    .Select((frameworkInfo, originalIndex) => (frameworkInfo, originalIndex))
+                    .OrderBy(tuple => tuple.frameworkInfo.FrameworkName.ToString(), StringComparer.Ordinal);
+
+                foreach (var (frameworkInfo, originalIndex) in orderedFrameworksWithOriginalIndex)
                 {
-                    var dependency = frameworkInfo.Dependencies.First(
-                        dep => dep.Name.Equals(packageReferenceArgs.PackageId, StringComparison.OrdinalIgnoreCase));
+                    var index = frameworkInfo.Dependencies.FirstIndex(dep => dep.Name.Equals(packageReferenceArgs.PackageId, StringComparison.OrdinalIgnoreCase));
+                    var dependency = frameworkInfo.Dependencies[index];
```
zivkan, line 331: `index` isn't used more than once, so I'm not sure what what getting the index, then getting the value via the indexer gets us vs keeping `First()`.

This is code for `dotnet add package`, which is a short lived process. So reducing a single enumerator allocation isn't meaningful.

src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
zivkan: With the changes in this PR, the compiler does enforce immutability. Although it does assume the caller won't use reflection to modify things, and there's also a problem if the caller is using an older LangVersion from before the `init` keyword was added. I believe in that case the code sees it as a mutable `set`, and therefore it's true that immutability isn't enforced.  But if we assume that people using our package are going to be using recent enough LangVersion, then I don't think the XMLDoc comment adds much. Unfortunately, anyone targeting .NET Framework without explicitly setting a higher LangVersion will end up in the no-enforcement situation though.

src/NuGet.Core/NuGet.LibraryModel/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,24 @@
...
+NuGet.LibraryModel.LibraryDependency.WithLibraryRange(NuGet.LibraryModel.LibraryRange! libraryRange) -> NuGet.LibraryModel.LibraryDependency!
+NuGet.LibraryModel.LibraryDependency.WithReferenceType(NuGet.LibraryModel.LibraryDependencyReferenceType referenceType) -> NuGet.LibraryModel.LibraryDependency!
+NuGet.LibraryModel.LibraryDependency.WithSuppressParent(NuGet.LibraryModel.LibraryIncludeFlags suppressParent) -> NuGet.LibraryModel.LibraryDependency!
+NuGet.LibraryModel.LibraryDependency.WithVersionCentrallyManaged(bool versionCentrallyManaged) -> NuGet.LibraryModel.LibraryDependency!
+NuGet.LibraryModel.LibraryRange.Name.init -> void
+NuGet.LibraryModel.LibraryRange.TypeConstraint.init -> void
+NuGet.LibraryModel.LibraryRange.VersionRange.init -> void
+NuGet.LibraryModel.LibraryRange.WithName(string! name) -> NuGet.LibraryModel.LibraryRange!
+NuGet.LibraryModel.LibraryRange.WithTypeConstraint(NuGet.LibraryModel.LibraryDependencyTarget typeConstraint) -> NuGet.LibraryModel.LibraryRange!
+NuGet.LibraryModel.LibraryRange.WithVersionRange(NuGet.Versioning.VersionRange! versionRange) -> NuGet.LibraryModel.LibraryRange!
```
zivkan: As mentioned in another comment, I don't think we should add these `.With*` methods.  They were a great solution before the C# language added `init`. But given this feature now exists, and the point of this pull request is to reduce memory allocations, it's easier to further reduce allocations by using the C# language feature directly, since each of these `.With` methods cause an allocation.

Even if there wasn't an allocation concern, I just see the `.With*` methods as a duplication of the C# language feature.

src/NuGet.Core/NuGet.ProjectModel/CentralTransitiveDependencyGroup.cs
zivkan: > The reason why I'm asking is because a foreach would usually allocate on an IList

Maybe we should use for `for` loops and fewer `foreach` loops 😄

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/AddPackageReferenceCommandRunner.cs
ToddGrun: Index is also used by the SetItem call lower in the method

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
ToddGrun: Yeah, these are cleaner as for loops, as the index is needed because the data is now written to an array instead of a List.

src/NuGet.Core/NuGet.LibraryModel/LibraryRange.cs
ToddGrun: Removed all With* operations and replaced with manual calls to a copy constructor

src/NuGet.Core/NuGet.ProjectModel/TargetFrameworkInformation.cs
ToddGrun: Removed all With* operations and replaced with manual calls to a copy constructor

src/NuGet.Core/NuGet.Commands/Utility/AssetTargetFallbackUtility.cs
ToddGrun: Removed all With* operations and replaced with manual calls to a copy constructor. Note that I did refactor a bit of the code to reduce some of the allocations.

src/NuGet.Core/NuGet.LibraryModel/PublicAPI.Unshipped.txt
ToddGrun: Removed all With* operations and replaced with manual calls to a copy constructor

Directory.Packages.props
ToddGrun: I tried this and ci still failed, so I've reverted back to what I had before. The 8th commit was the one where I tried to make the change if you are interested.

src/NuGet.Core/NuGet.ProjectModel/CentralTransitiveDependencyGroup.cs
nkolev92: > > The reason why I'm asking is because a foreach would usually allocate on an IList
> 
> Maybe we should use for `for` loops and fewer `foreach` loops 😄

We can. 

The other thing about this type is that I'm *not* really worried about a few allocations in the pack codepaths.

ToddGrun: To clarify, you are worried about allocations introduced in this PR, or allocations in general?

nkolev92: My brain thought no, my fingers didn't type no. 

I'm really *not* worried about pack codepath allocations, which seem to be the ones to benefit from this change.

build/Shared/SharedExtensions.cs
```diff
@@ -126,5 +125,49 @@ public static void AppendInt(this StringBuilder sb, int value)
                 divisor /= 10;
             }
         }
+
+        public static int SingleIndex<T>(this IEnumerable<T> lst, Predicate<T> isMatch)
```
kartheekp-ms: Given that this method is so far invoked on `PackageSpec.TargetFrameworks` which is of type `List`, I think we could reduce the allocations here by allowing this method to be invoked on the `IList` and then invoke `lst.NoAllocEnumerate` in the foreach statement.

https://github.com/NuGet/NuGet.Client/blob/8e96cf5c2b5a192a3955b7e36951e482b2613915/build/Shared/NoAllocEnumerateExtensions.cs#L34-L46

build/Shared/SharedExtensions.cs
```diff
@@ -126,5 +125,49 @@ public static void AppendInt(this StringBuilder sb, int value)
...
+
+            if (foundIndex == -1)
+            {
+                throw new InvalidOperationException();
+            }
+
+            return foundIndex;
+        }
+
+        public static int FirstIndex<T>(this IEnumerable<T> lst, Predicate<T> isMatch)
```
kartheekp-ms: Similar to https://github.com/NuGet/NuGet.Client/pull/5930/files#r1697231896 comment.


Given that this method is so far invoked on `TargetFrameworkInformation.Dependencies` which is of type Immutable Array that [implements IList interface](https://learn.microsoft.com/dotnet/api/system.collections.immutable.immutablearray-1?view=net-8.0), I think we could reduce the allocations here by allowing this method to be invoked on the IList and then invoke lst.NoAllocEnumerate in the foreach statement.

ToddGrun: Thoughts:

1) If I changed this to take in an IList, I wouldn't need the enumerator, but it would reduce the number of potential callers.

2) If I keep this as taking in an IEnumerable, it seems like there is a NoAllocEnumerate extension method defined on that too. Is there a downside to using it?

Assuming those two stmts are true, would you prefer route 1 or 2?

ToddGrun: Went ahead and pushed a commit where I changed the input to an IList, and just did a for loop instead of foreach

nkolev92 (dismissed):
I don't really have any feedback here.

This is a great change and something we've wanted to do for a while. 
This should allow us to improve the no-op calculation eventually since we're primarily dealing with immutable objects (nearly all of them are I think). 

We'll want to run some manual validation and get sign off as this is a binary breaking change. 
There are ways we can reduce that footprint (by obsoleting certain method and using the new ones only), but some places, like the types becoming immutable there's simply not other way.

src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
```diff
@@ -149,46 +170,58 @@ public bool Equals(LibraryDependency? other)
...
+                    {
+                        LibraryRange = libraryRange,
+                        VersionCentrallyManaged = versionCentrallyManaged
+                    };
                 }
+
+                result[i] = d;
             }
+
+            return ImmutableCollectionsMarshal.AsImmutableArray(result);
```
kartheekp-ms:
```suggestion
            result.ToImmutableArray();
```

ToddGrun: I believe ToImmutableArray allocates a new array, whereas ImmutableCollectionsMarshal just creates a wrapper around an existing array.

nkolev92:
Team Triage: Given the recent chnages in https://github.com/NuGet/NuGet.Client/commit/e68fb0f2d96800ad4b78461fd491886e9e5084b2, we want to minimize the code churn in the restore area. 
We want to take this in the next release. 

We'd need to rebase closer to when we'd be ready for the next release.

Theghost256 (approved):

Theghost256 (approved):

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
```diff
@@ -260,7 +259,7 @@ public static bool EvaluateRuntimeDependencies(ref LibraryRange libraryRange, st
             // Look up any additional dependencies for this package
             foreach (var runtimeDependency in runtimeGraph.FindRuntimeDependencies(runtimeName, libraryRange.Name).NoAllocEnumerate())
             {
-                var libraryDependency = new LibraryDependency(noWarn: Array.Empty<NuGetLogCode>())
+                var libraryDependency = new LibraryDependency() // TODO: NK
```
zivkan: incomplete todo

src/NuGet.Core/NuGet.LibraryModel/LibraryDependency.cs
```diff
@@ -148,46 +170,59 @@ public bool Equals(LibraryDependency? other)
...
-            var clonedNoWarn = new List<NuGetLogCode>(NoWarn);
-
-            return new LibraryDependency(clonedLibraryRange, IncludeType, SuppressParent, clonedNoWarn, AutoReferenced, GeneratePathProperty, VersionCentrallyManaged, ReferenceType, Aliases, VersionOverride);
-        }
-
         /// <summary>
         /// Merge the CentralVersion information to the package reference information.
         /// </summary>
-        public static void ApplyCentralVersionInformation(IList<LibraryDependency> packageReferences, IDictionary<string, CentralPackageVersion> centralPackageVersions)
+        /// TODO NK - This probably could be removed completely by inverting the loigc.
```
zivkan: todo

src/NuGet.Core/NuGet.ProjectModel/PackageSpecOperations.cs
```diff
@@ -29,32 +30,77 @@ public static void AddOrUpdateDependency(PackageSpec spec, PackageDependency dep
...
-
+            var foundExistingDependency = false;
             var range = dependency.VersionRange;
+            var dependencyId = dependency.Id;
 
-            foreach (var existingDependency in existing)
+            for (var i = 0; i < spec.Dependencies.Count; i++)
             {
-                if (spec.RestoreMetadata?.CentralPackageVersionsEnabled == true && !existingDependency.VersionCentrallyManaged && existingDependency.VersionOverride is not null)
+                // TODO NK - The idea is to copy the version override I think.
```
zivkan: todo

Theghost256 (approved):

Theghost256 (approved):

Theghost256 (approved):

Theghost256 (approved):

Theghost256 (approved):

Theghost256 (approved):

Theghost256 (approved):

Theghost256 (approved):

Theghost256 (approved):

Theghost256 (approved):

Theghost256 (approved):
Thank you for letting me help ☺️

src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
nkolev92: It's not ready for re-review. I'll mark it as draft.

Theghost256 (approved):

nkolev92:
@NuGet/nuget-client This is ready for review.

src/NuGet.Core/NuGet.ProjectModel/PackageSpecOperations.cs
```diff
@@ -29,32 +30,74 @@ public static void AddOrUpdateDependency(PackageSpec spec, PackageDependency dep
...
 
-            var existing = GetExistingDependencies(spec, dependency.Id);
-
+            var foundExistingDependency = false;
             var range = dependency.VersionRange;
+            var dependencyId = dependency.Id;
+
+            for (var i = 0; i < spec.Dependencies.Count; i++)
+            {
+                // existingDependency.LibraryRange.VersionRange = range;
```
zivkan: This looks like old code (line 43) that was commented out during development, rather than an intentional comment. If it's an intentional comment, it needs improvement to be more understandable.

zivkan (dismissed):

src/NuGet.Core/NuGet.ProjectModel/PackageSpecOperations.cs
nkolev92: Yep, fixed now.

zivkan (approved):

PR #5931, opened 2024-07-22
Add why and config commands to dotnet nuget help

### LINKED ISSUE #13517: 'why' and 'config' command does not show up in 'dotnet nuget --help' output
### NuGet Product Used

dotnet.exe, NuGet SDK

### Product Version

.NET 9.0.100 Preview 6

### Worked before?

No

### Impact

### Repro Steps & Context

1. Run 'dotnet nuget -h'
2. See commands listed in help output: the 'why' command is missing

We are in the process of migrating our CLI commands from '_Microsoft.Extensions.CommandLineUtils.CommandLineApplication_' to '_System.Commandline.CliCommand_', and this has meant that we have a split in our command parsing here (migrated commands): https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs#L83-L85, and here (pre-migration commands): https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs#L257-L276.

All commands registered with the old code path (pre-migration) show up correctly in the help output, but migrated `_dotnet nuget *_' commands (_why_/_config_) are not being registered correctly. They still run correctly. They just do not show up in 'dotnet nuget --help' output.

### Verbose Logs

```shell
> dotnet nuget -h
```

See also: https://github.com/dotnet/sdk/issues/42084

martinrrm:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13517

## Description
Added missing commands to `dotnet nuget` help commands, since the migration of the commands is still a work in progress we need to add them to the old help option. 

This PR only register so we can display their description but the functionality is implemented in `System.Commandline.CliCommand`
## PR Checklist

- [x] Added tests

### DIFF
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs
@@ -6,7 +6,6 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
-using System.Reflection;
 using Microsoft.Extensions.CommandLineUtils;
 using NuGet.Commands;
 using NuGet.Common;
@@ -276,6 +275,8 @@ namespace NuGet.CommandLine.XPlat
                 VerifyCommand.Register(app, getHidePrefixLogger, setLogLevel, () => new VerifyCommandRunner());
                 TrustedSignersCommand.Register(app, getHidePrefixLogger, setLogLevel);
                 SignCommand.Register(app, getHidePrefixLogger, setLogLevel, () => new SignCommandRunner());
+                ConfigCommand.Register(app);
+                WhyCommand.Register(app);
             }
 
             app.FullName = Strings.App_FullName;
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs
@@ -275,6 +275,7 @@ namespace NuGet.CommandLine.XPlat
                 VerifyCommand.Register(app, getHidePrefixLogger, setLogLevel, () => new VerifyCommandRunner());
                 TrustedSignersCommand.Register(app, getHidePrefixLogger, setLogLevel);
                 SignCommand.Register(app, getHidePrefixLogger, setLogLevel, () => new SignCommandRunner());
+                // The commands below are implemented with System.CommandLine, and are here only for `dotnet nuget --help`
                 ConfigCommand.Register(app);
                 WhyCommand.Register(app);
             }
--- a/test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatHelpOutputTests.cs
+++ b/test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatHelpOutputTests.cs
@@ -12,7 +12,6 @@ using System.Text.RegularExpressions;
 
 namespace NuGet.XPlat.FuncTest
 {
-    [Collection("NuGet XPlat dotnet help test collection")]
     public class XPlatHelpOutputTests
     {
         private readonly ITestOutputHelper _testOutputHelper;
--- a/test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatHelpOutputTests.cs
+++ b/test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatHelpOutputTests.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using NuGet.CommandLine.XPlat;
 using Xunit;
 using Xunit.Abstractions;
+using System.Text.RegularExpressions;
 
 namespace NuGet.XPlat.FuncTest
 {
@@ -42,6 +43,24 @@ namespace NuGet.XPlat.FuncTest
                 yield return new object[] { new string[] { "add", "Test", "Case" } };
             }
         }
+
+        public static IEnumerable<string> HelpCommands => new List<string>
+        {
+            "add",
+            "config",
+            "delete",
+            "disable",
+            "enable",
+            "list",
+            "locals",
+            "push",
+            "remove",
+            "sign",
+            "trust",
+            "update",
+            "verify",
+            "why"
+        };
 
         [Theory]
         [MemberData(nameof(FailParsing))]
@@ -81,6 +100,33 @@ namespace NuGet.XPlat.FuncTest
             var output = consoleOutput.ToString();
             Assert.DoesNotContain("Usage", output);
             Assert.Equal(1, exitCode);
+        }
+
+        [Fact]
+        public void MainInternal_ShowsHelp()
+        {
+            // Arrange
+            var originalConsoleOut = Console.Out;
+            using var consoleOutput = new StringWriter();
+            Console.SetOut(consoleOutput);
+            var log = new TestCommandOutputLogger(_testOutputHelper);
+
+            // Act
+            var exitCode = Program.MainInternal(Array.Empty<string>(), log);
+            Console.SetOut(originalConsoleOut);
+
+            // Assert
+            var output = consoleOutput.ToString();
+
+            var commandPattern = @"^\s{2}(\w+)\s{2,}"; // Matches lines starting with two spaces, a word (command), followed by at least two spaces
+            IEnumerable<string> matches = Regex.Matches(output, commandPattern, RegexOptions.Multiline).Select(m => m.ToString().Trim());
+
+            foreach (var command in HelpCommands)
+            {
+                Assert.Contains(command, matches);
+            }
+
+            Assert.Equal(0, exitCode);
         }
     }
 }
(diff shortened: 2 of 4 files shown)


test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatHelpOutputTests.cs
```diff
@@ -82,5 +101,32 @@ public void MainInternal_OnParsingSuccess_DoesNotShowsHelp(string[] args)
...
+            // Assert
+            var output = consoleOutput.ToString();
+
+            var commandPattern = @"^\s{2}(\w+)\s{2,}"; // Matches lines starting with two spaces, a word (command), followed by at least two spaces
+            IEnumerable<string> matches = Regex.Matches(output, commandPattern, RegexOptions.Multiline).Select(m => m.ToString().Trim());
+
+            foreach (var command in HelpCommands)
+            {
+                Assert.Contains(command, matches);
+            }
```
kartheekp-ms: This seems to O(nsquare) time complexity. Is there any other API that compares the collection for equality in xUnit? Maybe `Assert.Equal(HelpCommands, matches.OrderBy(m => m));`

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatHelpOutputTests.cs
```diff
@@ -82,5 +101,32 @@ public void MainInternal_OnParsingSuccess_DoesNotShowsHelp(string[] args)
...
             Assert.Equal(1, exitCode);
         }
+
+        [Fact]
+        public void MainInternal_ShowsHelp()
+        {
+            // Arrange
+            var originalConsoleOut = Console.Out;
+            using var consoleOutput = new StringWriter();
+            Console.SetOut(consoleOutput);
```
zivkan, line 111: This can cause problems if tests run in parallel. If we end up keeping this test, then I think additional changes are necessary to ensure that this test will never run in parallel with other tests.

I think a better option would be to change `MainInternal` to take `StreamWriter` as an input, which `Main` passes `Console.Out` to, and then this test doesn't need to modify any static state.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatHelpOutputTests.cs
```diff
@@ -82,5 +101,32 @@ public void MainInternal_OnParsingSuccess_DoesNotShowsHelp(string[] args)
...
+            var log = new TestCommandOutputLogger(_testOutputHelper);
+
+            // Act
+            var exitCode = Program.MainInternal(Array.Empty<string>(), log);
+            Console.SetOut(originalConsoleOut);
+
+            // Assert
+            var output = consoleOutput.ToString();
+
+            var commandPattern = @"^\s{2}(\w+)\s{2,}"; // Matches lines starting with two spaces, a word (command), followed by at least two spaces
```
kartheekp-ms, line 121: Is this behavior always same? I mean, will each line have 2 spaces + command + 2 spaces?

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatHelpOutputTests.cs
```diff
@@ -82,5 +101,32 @@ public void MainInternal_OnParsingSuccess_DoesNotShowsHelp(string[] args)
...
+            IEnumerable<string> matches = Regex.Matches(output, commandPattern, RegexOptions.Multiline).Select(m => m.ToString().Trim());
+
+            foreach (var command in HelpCommands)
+            {
+                Assert.Contains(command, matches);
```
zivkan: I'm thinking about what kind of regressions this test will catch, and what it will miss.

It will catch someone removing an existing command from the help. For example, if one command is ported from the existing CLI parser to System.CommandLine and the person removes the comand registration from the `dotnet nuget *` code path, this test will fail, telling the developer not to do that.

This test will not catch the scenario where a brand new command is implemeneted and is special cased like config and why are, so they don't get registered through the `dotnet nuget *` code path. That's exactly the scenario for `config` and `why` commands that this PR is fixing.

Since commands don't need to implement an interface, or extend a base class, I don't have any ideas for how we could improve the test to catch the second scenario. I just wanted to raise awareness, in case someone else reviewing has an idea for an easy way to do it. My philosophy for bug fixes is that any tests added should have prevented the bug in the first place, had the test been in place before the bug was introduced. But this test unfortunately would not have, and I can't think of a nice way to do so.   Maybe this is just a temporary thing until the System.CommandLine migration is complete.

zivkan (dismissed):

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatHelpOutputTests.cs
martinrrm: For now yes, I know this is not the best way to rely, when I finish the migration to the command I will make sure there is an easier way to know which commands are implemented.

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatHelpOutputTests.cs
martinrrm: I changed the comparison a little bit, I'll check that both have the same strings in the same order and the sizes are the same.

This could me improved later when we finish the migration because there is no nice way to compare it :(

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatHelpOutputTests.cs
martinrrm: Looked at the documentation https://xunit.net/docs/running-tests-in-parallel and it says that making this class a collection avoids running them in parallel, do you think this addresses your concern?

I'm not sure if I get the recommendation to change `MainInternal` I'm not sure how can we change the function to receive a `StreamWriter`

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatHelpOutputTests.cs
kartheekp-ms: Is there an epic where you are tracking the work for migration? If yes, I suggest adding a tracking issue so that we don't forget.

src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs
```diff
@@ -276,6 +275,8 @@ private static CommandLineApplication InitializeApp(string[] args, CommandOutput
                 VerifyCommand.Register(app, getHidePrefixLogger, setLogLevel, () => new VerifyCommandRunner());
                 TrustedSignersCommand.Register(app, getHidePrefixLogger, setLogLevel);
                 SignCommand.Register(app, getHidePrefixLogger, setLogLevel, () => new SignCommandRunner());
+                ConfigCommand.Register(app);
```
zivkan, line 279:
```suggestion
                // The commands below are implemented with System.CommandLine, and are here only for `dotnet nuget --help`
                ConfigCommand.Register(app);
```

zivkan (dismissed):

test/NuGet.Core.FuncTests/NuGet.XPlat.FuncTest/XPlatHelpOutputTests.cs
```diff
@@ -8,9 +8,11 @@
 using NuGet.CommandLine.XPlat;
 using Xunit;
 using Xunit.Abstractions;
+using System.Text.RegularExpressions;
 
 namespace NuGet.XPlat.FuncTest
 {
+    [Collection("NuGet XPlat dotnet help test collection")]
```
kartheekp-ms: Follow-up for https://github.com/NuGet/NuGet.Client/pull/5931#discussion_r1688505682 conversation.

TLDR, I don't think the collection attribute is required here because multiple classes have to be decorated with the same attribute for xUnit to not run them in parallel.

[xUnit docs](https://xunit.net/docs/running-tests-in-parallel.html) state that, `If we need to indicate that multiple test classes should not be run in parallel against one another, then we place them into the same test collection. This is simply a matter of decorating each test class with an attribute that places them into the same uniquely named test collection.` In this case, the collection name is specified only for this class, so the parallel concept doesn't apply here because, by default, xUnit runs tests in a single class sequentially.

The only place where Console.Out state has been modified in the tests is NuGet.Protocol.FuncTest/PluginTests.cs.

https://github.com/NuGet/NuGet.Client/blob/46f80a3f1a83f3c5f33a20fa9fb72871e1e2d4eb/test/NuGet.Core.FuncTests/NuGet.Protocol.FuncTest/PluginTests.cs#L329-L432

Given that these tests never failed, I assume the tests in these two assemblies (NuGet.Protocol.FuncTest and NuGet.Core.FuncTests) were never executed in parallel. Given that the existing tests in `XPlatHelpOutputTests.cs` were already modifying the static state of Console, I guess it is okay. I like Andy's suggestion in the other comment I linked above to inject TextWriter into the `MainInternal` method and have the `Main` method just pass `Console.Out` so that we don't have to update the global state of Console. _I am okay with creating a tech debt issue to track this work._

martinrrm: Created tech debt issue https://github.com/NuGet/Home/issues/13655

kartheekp-ms (approved):

PR #5932, opened 2024-07-23
raise warning if insecure sha-1 certificate fingerprint is passed to mssign and reposign commands

kartheekp-ms:
# Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2939

## Description

Deprecate the usage of SHA-1 fingerprints in `mssign` and `reposign` commands especially for `CertificateFingerprint` option. Instead, allow `mssign` and `reposign` commands to accept SHA-2 (SHA-256, SHA-384 and SHA-512) family fingerprints for searching a local certificate store for the certificate.

Here is how the `mssign` and `reposign` commands after this PR has been merged:

- Validates the certificate fingerprint to ensure it is either SHA-1, SHA-256, SHA-384, or SHA-512 algorithm. Throws an error otherwise.
- Raises a warning if a SHA-1 certificate fingerprint is provided. This warning will be promoted to an error in the future.
- While searching for a certificate from the local store, if a SHA-1 hash is provided, finds the certificate using the existing API i.e. `certCollection.Find(X509FindType.FindByThumbprint, CertificateFingerprint, validOnly: false);`
- If a SHA-256, SHA-384, or SHA-512 certificate fingerprint is provided, loops through the certificates in the store to find the certificate with the matching hash.

I made similar changes to the dotnet sign command in https://github.com/NuGet/NuGet.Client/pull/5895 and nuget.exe sign command in https://github.com/NuGet/NuGet.Client/pull/5924.

## PR Checklist

- [x] Added tests
- [x] Link to an issue or pull request to update docs if this PR changes settings, environment variables, new feature, etc. https://github.com/NuGet/Client.Engineering/issues/2876

### DIFF
--- a/test/NuGet.Clients.FuncTests/NuGet.MSSigning.Extensions.FuncTest/Commands/MSSignCommandTests.cs
+++ b/test/NuGet.Clients.FuncTests/NuGet.MSSigning.Extensions.FuncTest/Commands/MSSignCommandTests.cs
@@ -25,13 +25,14 @@ namespace NuGet.MSSigning.Extensions.FuncTest.Commands
     {
         private readonly string _noTimestamperWarningCode = NuGetLogCode.NU3002.ToString();
         private readonly string _invalidCertificateFingerprintCode = NuGetLogCode.NU3043.ToString();
-        private const string _sha1Hash = "89967D1DD995010B6C66AE24FF8E66885E6E03A8";
+        private const string Sha1Hash = "89967D1DD995010B6C66AE24FF8E66885E6E03A8";
+        private const string Sha256Hash = "a591a6d40bf420404a011733cfb7b190d62c65bf0bcda32b55b046cbb7f506fb";
 
-        private TrustedTestCert<TestCertificate> _trustedTestCertWithPrivateKey;
-        private TrustedTestCert<TestCertificate> _trustedTestCertWithoutPrivateKey;
+        private readonly TrustedTestCert<TestCertificate> _trustedTestCertWithPrivateKey;
+        private readonly TrustedTestCert<TestCertificate> _trustedTestCertWithoutPrivateKey;
 
         private MSSignCommandTestFixture _testFixture;
-        private string _nugetExePath;
+        private readonly string _nugetExePath;
 
         public MSSignCommandTests(MSSignCommandTestFixture fixture)
         {
@@ -122,8 +123,10 @@ namespace NuGet.MSSigning.Extensions.FuncTest.Commands
             }
         }
 
-        [CIOnlyFact]
-        public void GetAuthorSignRequest_InvalidFingerprint()
+        [CIOnlyTheory]
+        [InlineData(Sha1Hash)]
+        [InlineData(Sha256Hash)]
+        public void GetAuthorSignRequest_InvalidFingerprint(string certificateFingerPrint)
         {
             var mockConsole = new Mock<IConsole>();
             var timestampUri = "http://timestamp.test/url";
@@ -139,7 +142,7 @@ namespace NuGet.MSSigning.Extensions.FuncTest.Commands
                     CertificateFile = test.CertificatePath,
                     CSPName = test.CertificateCSPName,
                     KeyContainer = test.CertificateKeyContainer,
-                    CertificateFingerprint = _sha1Hash,
+                    CertificateFingerprint = certificateFingerPrint,
                 };
                 signCommand.Arguments.Add(Path.Combine(dir, "package.nupkg"));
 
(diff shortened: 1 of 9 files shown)


martinrrm (dismissed):

test/NuGet.Clients.FuncTests/NuGet.MSSigning.Extensions.FuncTest/Commands/MSSignCommandTests.cs
```diff
@@ -24,6 +24,8 @@ namespace NuGet.MSSigning.Extensions.FuncTest.Commands
     public class MSSignCommandTests
     {
         private readonly string _noTimestamperWarningCode = NuGetLogCode.NU3002.ToString();
+        private readonly string _invalidCertificateFingerprintCode = NuGetLogCode.NU3043.ToString();
+        private const string _sha1Hash = "89967D1DD995010B6C66AE24FF8E66885E6E03A8";
```
dtivel: I think these break an IDE style rule that `static` and `const` fields shouldn't be `_`-prefixed.  Only instance fields should be `_`-prefixed.

`_noTimestamperWarningCode` -> `NoTimestamperWarningCode`
`_invalidCertificateFingerprintCode` -> `InvalidCertificateFingerprintCode`
`_sha1Hash ` -> `Sha1Hash`

test/NuGet.Clients.FuncTests/NuGet.MSSigning.Extensions.FuncTest/Commands/MSSignCommandTests.cs
```diff
@@ -137,7 +139,7 @@ public void GetAuthorSignRequest_InvalidFingerprint()
                     CertificateFile = test.CertificatePath,
                     CSPName = test.CertificateCSPName,
                     KeyContainer = test.CertificateKeyContainer,
-                    CertificateFingerprint = "invalid-fingerprint",
```
dtivel, line 140: An invalid fingerprint is still a good negative test case.  Instead of replacing this value, I suggest converting this test from a `Fact` into a `Theory` and test both cases, like you did [here](https://github.com/NuGet/NuGet.Client/pull/5932/files#diff-ebac56a1632255034931687d8b4bec09c58e9dfe12b6482166672639dce68ff6R181-R184).

test/NuGet.Clients.Tests/NuGet.MSSigning.Extensions.Test/NuGetMSSignCommandTest.cs
```diff
@@ -255,7 +289,7 @@ public void MSSignCommandArgParsing_ValidTimestampHashAlgorithm(string timestamp
                 var packagePath = Path.Combine(dir, "package.nupkg");
                 var timestamper = "https://timestamper.test";
                 var certFile = Path.Combine(dir, "cert.p7b");
-                var certificateFingerprint = new Guid().ToString();
+                var certificateFingerprint = _sha1Hash;
```
dtivel: When testing one validation --- in this case, timestamp algorithm --- amongst many, it's important to pass reasonably valid values for all other (non-tested) validations to isolate the validation under test.

Previously, we passed a GUID string for certificate fingerprint, but now we pass a SHA-1 fingerprint.  This PR changes fingerprint validation, specifically for SHA-1 fingerprints, so it's doubly weird that we're passing a deprecated value for a property we're not even testing.

I recommend creating a fake SHA-256 fingerprint and using it wherever we're _not_ testing certificate fingerprint validation.

test/NuGet.Clients.FuncTests/NuGet.MSSigning.Extensions.FuncTest/Commands/MSSignCommandTests.cs
kartheekp-ms: I renamed `_sha1Hash` -> `Sha1Hash`.

`_noTimestamperWarningCode` and `_invalidCertificateFingerprintCode` are read-only strings. IDE analyzers warned of the lack of the prefix _ if I remove the prefix. Hence, I didn't rename the variables.

test/NuGet.Clients.FuncTests/NuGet.MSSigning.Extensions.FuncTest/Commands/MSSignCommandTests.cs
kartheekp-ms: In this test case we are validating that if an invalid hexadecimal hash has been passed to the certificate fingerprint, then then the command will fail with `Can't find specified certificate.` exception.

Once this PR has been merged, the command will throw an exception with NU3034 error code if non-hexadecimal string is passed as fingerprint. 

As per your suggestion, I changed the test to pass invalid SHA1 and SHA256 hexadecimal hashes to ensure the certificate lookup fails with the expected exception.

test/NuGet.Clients.Tests/NuGet.MSSigning.Extensions.Test/NuGetMSSignCommandTest.cs
kartheekp-ms: Thanks for the feedback. Updated tests

dtivel (approved):

PR #5936, opened 2024-07-26
Create pipeline to run Visual Studio integration tests

zivkan:
# Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2699

## Description

Create `eng/pipelines/vs-tests.yml` for VS integration tests.  It's a pipeline that builds NuGet and the tests, creates the VS bootstrapper installer for the tests, then runs the various tests in DartLab.  This way, it doesn't require any other pipelines to run first (to generate build artifacts needed by the tests), so if we get a PR from an external contributor, we just need to queue the one pipeline for the PR branch.

Implementation details:
* Created subdirectory for templates used by new pipeline, to better organize the files
* `vs-test/apex.yml` is just `templates/Apex_Tests_On_Windows.yml` renamed and adapted.
* `vs-test/end_to_end.yml` is just `templates/EndToEnd_Tests_On_Windows.yml` renamed and adapted.
* `vs-test/build.yml` started off as a copy of `templates/Build_And_UnitTest.yml` and went through significant modification to remove stuff not needed for VS integration tests, and also some unnecessary changes I consider to be improvements.
* I removed the steps that notified github checks about the test results. If we as a team think it will be useful, we can add it back, but it'll need modifications, so I didn't want to spend the time on it unless we really want it.

I think there are other improvements that can be made to make the CI scripts easier to read, but they can be made later. This is a working pipeline, which we haven't had for a while.  Unfortunately, there are 17 failing E2E tests, so it's possible that we introduced regressions since we stopped running the tests on every build, or maybe something else in VS changed, and it's just our test code that no longer works.

## PR Checklist

- [x] ~Added tests~: Infrastructure change
- [x] ~Link to an issue or pull request to update docs if this PR changes settings, environment variables, new feature, etc.~ non-product change.


nkolev92:
I'm excited this is almost ready. The [pipeline](https://devdiv.visualstudio.com/DevDiv/_build?definitionId=22360) for the interested people. 

I don't really have any feedback on the yml. The steps are well described and from what I could tell necessary. I wish the build.yml overlapped with our regular build, but that's probably a bigger refactoring, one that may not even be possible due to the templates that we need to be using for the builds. 

What are the planned triggers? I see nothing is enabled. 

Daily Tests and this pipeline, can we keep the ymls as similar as possible? https://github.com/NuGet/NuGet.Client/blob/dev/eng/pipelines/dailytests.yml It doesn't have to be now, but curious about that as a goal? 

> This way, it doesn't require any other pipelines to run first (to generate build artifacts needed by the tests), so if we get a PR from an external contributor, we just need to queue the one pipeline for the PR branch.

I'm not sure this makes things easier. In fact, don't we need the build pipeline either way right? Wouldn't we need to trigger 2 pipelines now?

jeffkl (dismissed):

zivkan:
> I wish the build.yml overlapped with our regular build, but that's probably a bigger refactoring

I think it should be possible, and hopefully not overly difficult.  While DartLab might not support 1ES templates, and therefore we can't use DartLab in our official builds, I think it should be possible to use 1ES templates for the "build" stage of this pipeline. Then by using more template parameters and conditional sections, we should be able to use the same yml for both.

But I was just trying to get this pipeline working ASAP, so I see this as a future improvement.

> What are the planned triggers?

Honestly, I didn't think about it. First step is getting a working pipeline. I know in the past the team has discussed disabling E2E/Apex tests as part of the "every PR" requirement. I thought we pushed back and said no, knowing that VS works correctly is too important.  But I don't remember for sure.  It's also something we can configure in Azure DevOps, rather than the yaml, but we can configure it in the yml if we prefer.

> Daily Tests and this pipeline, can we keep the ymls as similar as possible?

I'm opinionated 🤣 . I don't like that the daily tests uses artifacts from the "PrivateDev" build. But I would like to make the daily pipeline align with this pipeline, building its own VS bootstrapper, and then being runnable completely isolated, in which case we can align the two, just changing which assembly actually gets run (and disabling E2E tests)

> I'm not sure this makes things easier. In fact, don't we need the build pipeline either way right? Wouldn't we need to trigger 2 pipelines now?

Short term we need to trigger two pipelines.  But there's no good reason why we need the current PrivateDev pipeline.  We just need to migrate the remaining checks it does to the public CI pipeline (localization is the only thing it can't validate, from memory, and this VS test pipeline does that), so once that's done there will be no need to trigger two pipelines, just this one. And the current PrivateDev pipeline will have no reason to exist.

nkolev92:
> Short term we need to trigger two pipelines. But there's no good reason why we need the current PrivateDev pipeline. We just need to migrate the remaining checks it does to the public CI pipeline (localization is the only thing it can't validate, from memory, and this VS test pipeline does that), so once that's done there will be no need to trigger two pipelines, just this one. And the current PrivateDev pipeline will have no reason to exist.

As long as the end goal is to have 1 pipeline longer term, I don't care what it's actually called, whether you turn the PrivateDev definition into something else or this one. 
I though that repurposing the PrivateDev pipeline, renaming it etc. would be easier, but I'm not the one doing the work for that so don't feel too strongly. 

That being said, can we make it the same triggers as PrivateDev for now?

It's minimal friction for people on the team. I'd love for things to "just work"

eng/pipelines/vs-tests.yml
```diff
@@ -1,5 +1,14 @@
 # This is a pipeline that will run NuGet.Client's Visual Studio integration tests
 
+pr: none
+trigger:
+  branches:
+    include:
+    - dev-*
+  paths:
+    exclude:
```
nkolev92, line 9: 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱

nkolev92 (approved):

PR #5939, opened 2024-07-30
Update packages used by repo

zivkan:
# Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2946

## Description

* Enable `auditSources` for repo
* make NuGetAudit warnings not errors
* did restore with `-p:NuGetAuditMode=all` because that's going to be the default when this version of NuGet ships, and fix all warnings

The new version of xunit has some changed behaviour, `ITestOutputHelper` is per-test, and if it's used after the test it was created for has finished, it throws. PackageManagerTests has tests that would set NuGet's static telemetry object, but not clear it, so I split those telemetry tests into a different class, and set that class to use a collection with parallelism disabled, which makes the tests run after all parallelisable tests have finished. I have an idea for how to improve our test infrastructure to eliminate this risk happening in other test projects, but I'll implement it in another PR.

While at it, I also split the file into a bunch of smaller files, so more of those tests can be run in parallel.  Now, `dotnet test` on my machine for this test project finishes after 1m 48s, compared to 2m 54s (and I found a test that was declared wrong, so I fixed that and now we have 2 more tests running, all in less time).

## PR Checklist

- [x] ~Added tests~  N/A
- [x] ~Link to an issue or pull request to update docs if this PR changes settings, environment variables, new feature, etc.~ N/A


test/NuGet.Core.Tests/NuGet.Build.Tasks.Pack.Test/NuGet.Build.Tasks.Pack.Test.csproj
```diff
@@ -20,6 +20,14 @@
     <PackageReference Include="Microsoft.Build.Utilities.Core" PrivateAssets="All" />
   </ItemGroup>
 
+  <ItemGroup Label="transitive package pinning">
+    <!--
+      These packages are dependencies of directly referenced PackageReferences.
+      When the above PackageReferences are upgraded to newer versions, try deleting the below PackageReferences
+      -->
+    <PackageReference Include="System.Formats.Asn1" Condition=" '$(TargetFramework)' != '$(NETFXTargetFramework)' " />
```
nkolev92: Why isn't the NuGet.Packaging bump enough?

src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.csproj
```diff
@@ -47,6 +47,14 @@
     <PackageReference Include="Microsoft.Build.Utilities.Core" ExcludeAssets="runtime" GeneratePathProperty="true" />
   </ItemGroup>
 
+  <ItemGroup Label="transitive package pinning">
+    <!--
+      These packages are dependencies of directly referenced PackageReferences.
+      When the above PackageReferences are upgraded to newer versions, try deleting the below PackageReferences
+      -->
+    <PackageReference Include="System.Security.Cryptography.Xml" GeneratePathProperty="true" Condition=" '$(TargetFramework)' != '$(NETFXTargetFramework)' " />
```
nkolev92, line 55: Where is this one coming from?

NuGet.Config
```diff
@@ -9,6 +9,10 @@
     <add key="nuget-build" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/nuget-build/nuget/v3/index.json" />
     <add key="vside" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/msft_consumption%40Local/nuget/v3/index.json" />
   </packageSources>
+  <auditSources>
+    <clear />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
```
nkolev92, line 14: Does this cause some static validations to raise false positives about nuget.org usage?

src/NuGet.Clients/NuGet.VisualStudio.Interop/NuGet.VisualStudio.Interop.csproj
```diff
@@ -17,6 +17,16 @@
     <PackageReference Include="Microsoft.VisualStudio.Shell.15.0" />
     <PackageReference Include="Newtonsoft.Json" ExcludeAssets="all" />
   </ItemGroup>
+
+  <ItemGroup Label="transitive package pinning">
+    <!--
+      These packages are dependencies of directly referenced PackageReferences.
```
nkolev92: In some places you put the name of the direct ref that brings in the package, do you think it'd be helpful to specify that?

nkolev92 (commented):
It can be laborious to audit transitives for projects that don't ship as packages or aren't executables themselves. 

We can consider limiting things to the NuGet.Core assemblies + NuGet.VisualStudio.Client, NuGet.VisualStudio, NuGet.VisualStudio.Contracts.

NuGet.Config
zivkan: nothing in the devdiv private build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=9962868&view=results

src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.csproj
zivkan:
```
> dotnet nuget why .\src\NuGet.Core\NuGet.Build.Tasks.Pack\ System.Security.Cryptography.Xml
Project 'NuGet.Build.Tasks.Pack' has the following dependency graph(s) for 'System.Security.Cryptography.Xml':

  [netstandard2.0]
   │
   ├─ Microsoft.Build.Tasks.Core (v16.8.0)
   │  └─ System.Security.Cryptography.Xml (v8.0.1)
   └─ System.Security.Cryptography.Xml (v8.0.1)

  [net472]
   │
   └─ No dependency graph(s) found for this target framework.
```

test/NuGet.Core.Tests/NuGet.Build.Tasks.Pack.Test/NuGet.Build.Tasks.Pack.Test.csproj
zivkan: it appears to be. I don't remember why I added it here as well 😕

src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.csproj
nkolev92: Ah, and we can't move it cause of our target frameworks not being compatible with newer MSBuild.

We really need https://github.com/NuGet/Home/issues/13079 so we can change the TFM of our project here.

nkolev92 (dismissed):
Nicely done. 

Excited we're using audit sources!

dtivel (approved):

nkolev92 (approved):

PR #5945, opened 2024-07-31
[CI] Re-enable Mac Mono Tests

jeffkl:
# Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2926

## Description
This re-enables the Mac Mono tests after they were disabled due to not working after a change.  I had to bring back the `xunit.runner.json` for some reason.  I spent a lot of time trying to use newer tooling but that seems to be broken.  So I really just brought them back the way there were before.  

## PR Checklist

- [ ] Added tests

### DIFF
--- a/eng/pipelines/official.yml
+++ b/eng/pipelines/official.yml
@@ -26,7 +26,7 @@ parameters:
 - name: RunMonoTestsOnMac
   displayName: Run Mono tests on Mac
   type: boolean
-  default: true
+  default: false
 
 resources:
   repositories:
--- a/eng/pipelines/official.yml
+++ b/eng/pipelines/official.yml
@@ -26,7 +26,7 @@ parameters:
 - name: RunMonoTestsOnMac
   displayName: Run Mono tests on Mac
   type: boolean
-  default: false
+  default: true
 
 resources:
   repositories:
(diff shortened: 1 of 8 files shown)


jeffkl:
Passing build: https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=9974463&view=results

nkolev92 (dismissed):

eng/pipelines/official.yml
```diff
@@ -26,7 +26,7 @@ parameters:
 - name: RunMonoTestsOnMac
   displayName: Run Mono tests on Mac
   type: boolean
-  default: false
+  default: true
```
zivkan: I'm not sure there's much benefit to running any tests on official builds. When they hang causing a timeout, the entire CI job fails, causing problems when we need to insert. Even if they don't fail, when we have to service an already released branch, it adds delays since we need to wait for slow tests to finish before we can start the insertion.

zivkan (dismissed):

eng/pipelines/official.yml
jeffkl: Agreed, originally I just reverted the commit that disabled them but I'll push a commit that leaves them off for official builds so they only run on PrivateDev

zivkan (approved):

PR #5948, opened 2024-08-02
Fix broken links in Design Review Guide

donnie-msft:
Update links and remove the Design review workflow image which I couldn't find.

### DIFF
--- a/docs/design-review-guide.md
+++ b/docs/design-review-guide.md
@@ -1,7 +1,7 @@
 # Design guide and lifecycle
 
 As an open source project, the NuGet.Client team does the large majority of its work in public GitHub repos.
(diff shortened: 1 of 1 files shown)


nkolev92 (approved):

donnie-msft:
@nkolev92 it looks like this path doesn't trigger CI. Maybe you can merge it without that since it's docs.

nkolev92:
I think we might need to revert the change for md, or tweak it.

While I can merge this, I don't this is a good long term solution. Force merging is disabled by default. 

cc @zivkan in case he has an idea: https://github.com/NuGet/NuGet.Client/blob/4d413823d0cadf1f848f4a660b81377e9506ccd3/eng/pipelines/pr.yml#L45-L47.

jeffkl:
PrivateDev was triggered either since the branch name doesn't start with `dev-`.  Feel free to get rid of the path exclusion in `pr.yml` since we can't disable the required checks by path in GitHub.

PR #5959, opened 2024-08-07
Allow override of System.Formats.Asn1 package version

mthalman:
This updates the changes from https://github.com/NuGet/NuGet.Client/pull/5956 because they were incomplete. In addition to a dependency being defined for `System.Formats.Asn1`, there also needs to be a way to override that package version via the `<package-name>PackageVersion` property pattern. This property gets set by .NET source build, which requires overriding package versions.

Without this change, the original issue described in https://github.com/NuGet/NuGet.Client/pull/5956 isn't fixed.

/cc @nkolev92, @akoeplinger

# Bug

Fixes: 

## PR Checklist

- [ ] Added tests


akoeplinger (approved):
Thanks!

Nigusu-Allehu (approved):

zivkan (approved):

PR #5963, opened 2024-08-09
Introduce optimized dependency graph resolution algorithm

### LINKED ISSUE #13692: NuGet's dependency graph resolution algorithm should scale better for large graphs
### NuGet Product(s) Affected

MSBuild.exe, dotnet.exe

### Current Behavior

NuGet's current dependency graph resolution algorithm can be rather slow for large, complex graphs.  It walks the entire tree and for each level expands out every single node. This creates more of a tree representation but also includes duplicates.  Consider a popular package, `PackageX`, that might show up as a common dependency for lots of packages.  NuGet will create a graph in memory like this:

```
ProjectA
 ├─ PackageA
 │  └─ PackageX
 │     ├─ PackageY
 │     └─ PackageZ
 ├─ PackageB
 │  └─ PackageX
 │     ├─ PackageY
 │     └─ PackageZ
 ├─ PackageC
 │  └─ PackageX
 │     ├─ PackageY
 │     └─ PackageZ
 ├─ PackageD
 │  └─ PackageX
 │     ├─ PackageY
 │     └─ PackageZ
 └─ PackageE
    └─ PackageX
       ├─ PackageY
       └─ PackageZ
```

In restores for large repos internal to Microsoft, individual projects end up with _millions_ of nodes in the graph. 

After create the graph, NuGet then walks through the graph N times looking for versions to eclipse, detecting cycles, detecting version conflicts, and detecting downgrades.  Walking millions of nodes several times causes restore times for 2,500 projects to be around 15 minutes.

### Desired Behavior

We should rewrite the dependency graph resolution algorithm to be linear by having it only walk each node once, resolving versions as it goes.  Doing this in a single pass will greatly reduce the amount of time it takes to resolve complex graphs.

### Additional Context

_No response_

jeffkl:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13692

## Description
NuGet's current dependency graph resolution algorithm can be rather slow for large, complex graphs.  It walks the entire tree and for each level expands out every single node. This creates more of a tree representation but also includes duplicates.

This change introduces a new dependency graph walk algorithm developed by @bwadswor and @brianrob  .  It is optimized to only walk nodes once, resolving the tree as it goes.  It does this with two passes:
1. Pass 1 walks each dependency once and eclipses versions when it comes across a higher one that is allowed, resulting in the "resolved graph"
2. Pass 2 walks the resolved graph and compares it to the defined graph to detect cycles, downgrades, and conflicts.  It also creates the tree representation in this pass by creating `GraphNode<T>` items with parents and children.

This new functionality is on by default but will fall back to the legacy behavior if:
1.The user sets the `RESTORE_ENABLE_NEW_RESOLVER` environment variable to `false`
2. The project is not using PackageReference (disabled for `project.json` for example)
3. The project is using [Central Transitive Pinning](https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management#transitive-pinning). I'll be working on this functionality in a future PR

**Note**: There is a behavior change in the new implementation.  The legacy code path would download all packages _before_ eclipsing versions, resulting in packages existing on disk that might not be in the final resolved graph.  The new code path resolves the full graph before downloading packages.  This means that less packages will be on disk after restore but the ramification is that during the next restore, NuGet might contact the feed again during graph resolution.  The upside of this is that NuGet won't be potentially downloaded known vulnerable packages which a lot of users want, but it also means that depending on the HTTP cache and what packages already happen to be in the global packages folder, there might be more network traffic.

## PR Checklist

- [x] ~~Added tests~~ Existing tests were used with new functionality on by default


src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
jeffkl, line 1: I just moved two local functions to public static

src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
jeffkl, line 1: This class represents the bulk of the implementation and is basically a replacement for `RemoteDependencyWalker`

src/NuGet.Core/NuGet.Commands/RestoreCommand/ProjectRestoreCommand.cs
```diff
@@ -76,20 +76,7 @@ public ProjectRestoreCommand(ProjectRestoreRequest request)
...
-                downloadDependencyResolutionTasks.Add(ResolveDownloadDependenciesAsync(
-                    context,
-                    targetFrameworkInformation,
-                    token));
-            }
-
-            var downloadDependencyResolutionResults = await Task.WhenAll(downloadDependencyResolutionTasks);
-
-            telemetryActivity.EndIntervalMeasure(telemetryPrefix + EvaluateDownloadDependenciesDuration);
+            var downloadDependencyResolutionResults = await DownloadDependenciesAsync(_request.Project, context, telemetryActivity, telemetryPrefix, token);
```
jeffkl, line 79: This functionality was abstracted out to be a resusable method so both code paths could call it.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -1301,30 +1325,142 @@ internal static Task LogDowngradeWarningsOrErrorsAsync(IEnumerable<RestoreTarget
             return allGraphs;
         }
 
-        private List<ExternalProjectReference> GetProjectReferences()
+        private async Task<IEnumerable<RestoreTargetGraph>> ExecuteRestoreAsync(
```
jeffkl, line 1328: This is where the new implementation is called

test/NuGet.Clients.Tests/NuGet.CommandLine.Test/RestoreTransitiveLoggingTests.cs
```diff
@@ -1955,7 +1955,7 @@ public async Task GivenMultipleProjectReferencesNoWarnDifferentWarningsProjectWi
 
                 // Assert
                 r.Success.Should().BeTrue();
-                GetSubstringCount(r.AllOutput, "NU1603", ignoreCase: false).Should().Be(3);
+                GetSubstringCount(r.AllOutput, "NU1603", ignoreCase: false).Should().Be(2);
```
jeffkl, line 1958: I think its a bug, the old code would log 3 warnings, 2 of which are duplicates.  The new implementation only logs 2 warnings.

build/Shared/SharedStringBuilder.cs
```diff
@@ -0,0 +1,102 @@
...
+    /// <para>
+    /// Renting and returning buffers with an <see cref="StringBuilderPool"/> can increase performance
+    /// in situations where <see cref="StringBuilder"/> instances are created and destroyed frequently,
+    /// resulting in significant memory pressure on the garbage collector.
+    /// </para>
+    /// <para>
+    /// This class is thread-safe.  All members may be used by multiple threads concurrently.
+    /// </para>
+    /// </remarks>
+    internal class SharedStringBuilder
```
zivkan, line 24: It probably needs a comment saying that the string builder must be returned, even if the code using the buffer throws an exception, otherwise all future string builder re-use stops working (well, if it happens to be the cached string builder).

This doesn't appear to be a problem in the places within NuGet.Versioning where it's curerntly used, but since it's only being used by NuGet.Versioning, I'm wondering if we're better off moving this class to that project, rather than making it a shared class?

src/NuGet.Core/NuGet.Commands/GlobalSuppressions.cs
```diff
@@ -105,7 +105,7 @@
...
 [assembly: SuppressMessage("Build", "CA1062:In externally visible method 'void PackCommandRunner.SetupCurrentDirectory(PackArgs packArgs)', validate parameter 'packArgs' is non-null before using it. If appropriate, throw an ArgumentNullException when the argument is null or add a Code Contract precondition asserting non-null argument.", Justification = "<Pending>", Scope = "member", Target = "~M:NuGet.Commands.PackCommandRunner.SetupCurrentDirectory(NuGet.Commands.PackArgs)")]
-[assembly: SuppressMessage("Build", "CA1031:Modify 'GetRuntimeGraph' to catch a more specific allowed exception type, or rethrow the exception.", Justification = "<Pending>", Scope = "member", Target = "~M:NuGet.Commands.ProjectRestoreCommand.GetRuntimeGraph(System.String)~NuGet.RuntimeModel.RuntimeGraph")]
+[assembly: SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "<Pending>", Scope = "member", Target = "~M:NuGet.Commands.ProjectRestoreCommand.GetRuntimeGraph(System.String,NuGet.Commands.RestoreCollectorLogger)~NuGet.RuntimeModel.RuntimeGraph")]
```
zivkan, line 108: I think the `pending` justification was acceptable when we first enabled the analyzers, and there were hundreds, if not thousands, of violations. It was too much effort at the time, to justify each one.  But when we have a PR that modifies a global suppression, I think it would be great to either fix the code, or add a real justification.

Personally, I'd prefer inline suppressions anyway, since this global suppression will suppress the analyzer for the entire method, meaning if someone modifies the method to add additional violations, the compiler isn't going to warn.

src/NuGet.Core/NuGet.Commands/RestoreCommand/ProjectRestoreCommand.cs
```diff
@@ -219,14 +206,34 @@ private RuntimeGraph GetRuntimeGraph(string runtimeGraphPath)
...
-            var packageDownloadTasks = targetFrameworkInformation.DownloadDependencies.Select(downloadDependency =>
-            ResolverUtility.FindPackageLibraryMatchCachedAsync(downloadDependency, context, token));
+            telemetryActivity.StartIntervalMeasure();
 
-            var packageDownloadMatches = await Task.WhenAll(packageDownloadTasks);
+            var downloadDependencyResolutionTasks = new List<Task<DownloadDependencyResolutionResult>>();
+            foreach (var targetFrameworkInformation in packageSpec.TargetFrameworks)
+            {
+                downloadDependencyResolutionTasks.Add(ResolveDownloadDependenciesAsync(
+                context,
```
zivkan, line 217:
```suggestion
                    context,
```

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -1209,6 +1229,10 @@ internal static Task LogDowngradeWarningsOrErrorsAsync(IEnumerable<RestoreTarget
                 {
                     failed = true;
                 }
+                catch (AggregateException e) when (e.InnerException is FatalProtocolException)
+                {
+                    failed = true;
```
zivkan, line 1234: This is fine if there's only a single exception wrapped in the AggregateException, but if there are multiple exceptions, and only the first one is a `FatalProtocolException`, shouldn't we do something?  I'm guessing FatalProtocolException was special cased, because it gets logged somewhere, but if we have a bug causing a NullReferenceException, ArgumentException, or similar, it will be very hard to isolate if we silently ignore the exception.

Perhaps change it to something like `when (e.InnerExceptions.All(e => e is FatalogProcoolException))`

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -1301,30 +1325,142 @@ internal static Task LogDowngradeWarningsOrErrorsAsync(IEnumerable<RestoreTarget
...
+                _success = false;
+
+                return Enumerable.Empty<RestoreTargetGraph>();
+            }
+
+            var projectRestoreRequest = new ProjectRestoreRequest(
+             _request,
+             _request.Project,
+             _request.ExistingLockFile,
+             _logger)
```
zivkan, line 1349: nit: the indentation doesn't "look right"

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -1301,30 +1325,142 @@ internal static Task LogDowngradeWarningsOrErrorsAsync(IEnumerable<RestoreTarget
...
+                    runtimes = result.Runtimes;
+                }
+                catch (FatalProtocolException)
+                {
+                    failed = true;
+                }
+                catch (AggregateException e) when (e.InnerException is FatalProtocolException)
+                {
+                    failed = true;
+                }
```
zivkan, line 1397: Similar to another comment, should we change this to `when (e.InnerExceptions.All(e => e is FatalProtocolException))`?

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -1368,7 +1504,10 @@ private List<ExternalProjectReference> GetProjectReferences()
             {
                 // We care about TFM only and null RID for compilation purposes
                 projectFrameworkRuntimePairs.Add(new FrameworkRuntimePair(framework.FrameworkName, null));
+            }
 
+            foreach (var framework in packageSpec.TargetFrameworks)
```
zivkan, line 1509: `TargetFrameworks` is an `IList<T>`, so you need to use `NoAllocEnumerate()` to (hopefully) avoid the enumerator being allocated on the heap.  Or use a `for` loop, rather than `foreach`.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreTargetGraph.cs
```diff
@@ -18,43 +17,50 @@ public class RestoreTargetGraph : IRestoreTargetGraph
...
         // TODO: Move conflicts to AnalyzeResult
         public IEnumerable<ResolverConflict> Conflicts { get; internal set; }
 
-        public AnalyzeResult<RemoteResolveResult> AnalyzeResult { get; private set; }
+        public AnalyzeResult<RemoteResolveResult> AnalyzeResult { get; internal set; }
 
-        public ISet<ResolvedDependencyKey> ResolvedDependencies { get; }
+        public ISet<ResolvedDependencyKey> ResolvedDependencies { get; internal set; }
+
+        internal RestoreTargetGraph()
```
zivkan, line 61: Adding `internal set` to all these properties makes the class mutable, but only within the assembly, and adding this internal constructor, I think is a change in the wrong direction.  It's also going to make migrating to `#nullable enable` much more difficult.

I see that in the new `DependencyGraphResolver.cs`, it does something like:
```
var newRtg = new RestoreTargetGraph();
newRtg.RuntimeIdentifier = someValue;
```

Can it collect the parameters in local variables, and then call the existing constructor?

Alternatively, make the setters all `init`, not `internal set`, and also add `#nullable enable` to this file, and fix all the compiler's nullable warnings, both in this file and in the new dependency resolver.

src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
```diff
@@ -0,0 +1,1099 @@
...
+                {
+                    downloadDependencyResolutionResults = await ProjectRestoreCommand.DownloadDependenciesAsync(_request.Project, context, _telemetryActivity, telemetryPrefix: string.Empty, token);
+
+                    _success &= await projectRestoreCommand.InstallPackagesAsync(uniquePackages, allGraphs, downloadDependencyResolutionResults, userPackageFolder, token);
+
+                    hasInstallBeenCalledAlready = true;
+                }
+
+                //Build up our new RestoreTargetGraph.
+                //This is done by making everything on RTG setable.  We should move to using a normal constructor.
```
zivkan, line 92: Comments should explain why not what. I can see from the code below, and the change to the class, that all the setters were added. I don't understand why this approach was taken, rather than just calling the existing constructor with all these empty collections.

zivkan (approved):
I didn't spend much time looking at the new `DependencyGraphResolver`, so I haven't yet understood how it works. It does seem very complex, so I hope there will be many opportunities to refactor it to be easier to read & understand, but I don't know how much is just inherit in NuGet's dependency resolver algorithm we need to maintain support for.

jeffkl:
I have filed https://github.com/NuGet/Client.Engineering/issues/2959 to track follow-up items from this change that aren't blocking and added all of Andy's suggestions.

aortiz-msft (approved):
Approved with follow-up comments.

Theghost256 (approved):

Theghost256 (approved):

PR #5966, opened 2024-08-13
Install all packages during dependency walk in new resolver

jeffkl:
## Description
The legacy dependency resolution algorithm installs all packages in the graph walk while the new one only installs packages that ended up in the resolved graph.  However, this means that for subsequent restores, more I/O is needed since NuGet will ask the feed for the package again, download it, and read the nuspec.  Especially in Visual Studio this can cause more allocations.

This change adds to the list of packages to install during the walk instead of at the end based on the resolved graph.

## PR Checklist

- [ ] Added tests

### DIFF
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
@@ -517,6 +517,13 @@ namespace NuGet.Commands
                             libraryRangeInterningTable);
 
                         findLibraryEntryCache.Add(libraryRangeOfCurrentRef, refItemResult);
+
+                        // If the package came from a remote library provider, it needs to be installed locally.
+                        var isRemote = context.RemoteLibraryProviders.Contains(refItem.Data.Match.Provider);
+                        if (isRemote)
+                        {
+                            newRTG.Install.Add(refItem.Data.Match);
+                        }
                     }
 
                     HashSet<LibraryDependencyIndex>? suppressions = default;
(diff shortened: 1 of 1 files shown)


nkolev92 (approved):

src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
```diff
@@ -517,6 +517,13 @@ public DependencyGraphResolver(RestoreCollectorLogger logger, RestoreRequest res
                             libraryRangeInterningTable);
 
                         findLibraryEntryCache.Add(libraryRangeOfCurrentRef, refItemResult);
+
+                        // If the package came from a remote library provider, it needs to be installed locally.
+                        var isRemote = context.RemoteLibraryProviders.Contains(refItem.Data.Match.Provider);
+                        if (isRemote)
+                        {
+                            newRTG.Install.Add(refItem.Data.Match);
+                        }
```
Nigusu-Allehu, line 526:
```suggestion
                        if (context.RemoteLibraryProviders.Contains(refItem.Data.Match.Provider))
                        {
                            newRTG.Install.Add(refItem.Data.Match);
                        }
```

Nigusu-Allehu (approved):

src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
jeffkl: Will consider this as part of the refactoring

PR #5969, opened 2024-08-15
dotnuet nuget why uses current directory when provided with one argument

### LINKED ISSUE #13553: dotnet nuget why shouldn't require project when there is only one in the current directory
### NuGet Product Used

dotnet.exe

### Product Version

.NET 9 SDK daily preview

### Worked before?

no, new feature

### Impact

### Repro Steps & Context

When the current directory contains only a single project file, commands like `dotnet restore`, `dotnet build`, `dotnet add package`, `dotnet list package`, and so on, all use the project file implicitly.

`dotnet nuget why` not only doesn't implicitly use the project file, but outputs an error message twice

```console
> dotnet nuget why Azure.Identity
Required argument missing for command: 'why'.

Required argument missing for command: 'why'.

Description:
  Shows the dependency graph for a particular package for a given project or solution.

Usage:
  nuget why <PROJECT|SOLUTION> <PACKAGE> [options]

Arguments:
  <PROJECT|SOLUTION>  A path to a project, solution file, or directory.
  <PACKAGE>           The package name to lookup in the dependency graph.

Options:
  -f, --framework  The target framework(s) for which dependency graphs are shown.
  -?, -h, --help   Show help and usage information
```

Compare this to `dotnet add package -h` which has:

```console
Usage:
  dotnet add [<PROJECT>] package <PACKAGE_NAME> [options]

Arguments:
  <PROJECT>       The project file to operate on. If a file is not specified, the command will search the current directory for one. [default:
                  C:\src\myrepo\]
```

### Verbose Logs

_No response_

zivkan:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13553

## Description

Big thanks to @baronfel for the advice on how to get this to work.

Change the `path` argument to use a custom parser, where we can tell System.CommandLine if we want to take zero or one tokens from the parsed argument list. When we take zero, we use the current directory as a default.

I also made why's command line parsing configuration unit testable, so we can easily and exhaustively test all scenarios we care about, and it's still super fast (compared to invoking `dotnet nuget why` in dotnet.integration.tests, waiting for the process to end, and checking the exit code).

Also moved some code and namespaces around, to make things more intuitive/how other projects organise files and namespaces.
 
## PR Checklist

- [x] Added tests
   * https://github.com/NuGet/Home/issues/13709


src/NuGet.Core/NuGet.CommandLine.XPlat/NuGet.CommandLine.XPlat.csproj
```diff
@@ -21,9 +21,9 @@
 
 <!-- Microsoft.Build.Locator is only used when debugging, and the compiler will skip copying this dependency from Release assemblies we insert because we only refer to it conditionally with the DEBUG configuration.
   Uncomment the following when debugging. Also uncomment the MSBuildLocator code from Program.cs -->
-  <!-- <ItemGroup>
+   <!--<ItemGroup>
```
nkolev92: do we need this change?

I still like it when we had this under a Choose/When condition since it was just easy to work with, but that's not the topic of this PR.

nkolev92 (dismissed):

Nigusu-Allehu (dismissed):

src/NuGet.Core/NuGet.CommandLine.XPlat/NuGet.CommandLine.XPlat.csproj
zivkan: No, we don't need that change. I made such an effort to avoid commiting this file (and uncommenting the related lines in program.cs) on each commit, but all it took was one slip-up ☹️  Then, instead of "reverting" the file, I tried using the IDE to comment out the same lines, but it didn't do it the same way as the code before.

I also prefer the choose-when for debug mode. I don't know what problem was trying to be solved by having debug and release builds work the same, but in my experience working on this bug, it makes development harder. The debug-only code was just easier to work with and had no impact on the optimized binaries that we ship to customers.

nkolev92 (approved):

PR #5979, opened 2024-08-16
Suggest using dotnet nuget why in transitive package tooltip

### LINKED ISSUE #12105: [Bug]: Extra space at start of package description in tooltip
### NuGet Product Used

### Product Version

Visual Studio 2022

### Worked before?

_No response_

### Impact

Other

### Repro Steps & Context

In the Visual Studio Package Management UI, click on the Browse tab.

Hover over a package so that the tooltip appears.

At the start of the package description in the tooltip, there is an extra space. See screenshot attached.

image

### Verbose Logs

_No response_

### LINKED ISSUE #13574: PM UI should show transitive path
### NuGet Product(s) Affected

### Current Behavior

When hovering a package in the "installed" package list, when the package is a transitve package, it will tell you a top level package that caused the package to be included:
image

### Desired Behavior

The tooltip should show the full path of all intermediate packages. This is because when dealing with transitive packages with known vulnerabilities, some customers (and my personal recommendation) is to upgrade the "closest to direct" package to a higher version where packages with known vulnerabilities are removed.

For example, consider a package path, as shown by the new `dotnet nuget why` command:

```text
  [net8.0]
   │
   └─ YesSql (v5.0.0)
      └─ YesSql.Provider.SqlServer (v5.0.0)
         ├─ Azure.Identity (v1.11.2)
         │  ├─ Microsoft.Identity.Client (v4.60.3)
         │  └─ Microsoft.Identity.Client.Extensions.Msal (v4.60.3)
         │     └─ Microsoft.Identity.Client (v4.60.3)
         └─ Microsoft.Data.SqlClient (v5.2.0)
            ├─ Azure.Identity (v1.11.2)
            │  ├─ Microsoft.Identity.Client (v4.60.3)
            │  └─ Microsoft.Identity.Client.Extensions.Msal (v4.60.3)
            │     └─ Microsoft.Identity.Client (v4.60.3)
            └─ Microsoft.Identity.Client (v4.60.3)
```

As a developer, I'd prefer to update the `YesSql` package to a newer version, but if no newer version exists, then I'd like to update `YesSql.Provider.SqlServer`, and my next preference is `Azure.identity` or `Microsoft.Data.SqlClient`, etc.  In order to investigate upgrading these packages, I need to know the package path, not just the direct/root package that brought in the dependency.

### Additional Context

_No response_

martinrrm:
# Bug

Fixes: https://github.com/NuGet/Home/issues/12105 Removing extra space in tooltip, `<Run>` should be in the same line otherwise it will add an extra space.

image

Fixes: https://github.com/NuGet/Home/issues/13574
Fixes: https://github.com/NuGet/Home/pull/13680

## Description
Add a new text to the transitive level packages tooltip that suggests the user to use the command dotnet nuget why to know more about the package transitive path.
## PR Checklist

- [x] Added tests - It's an UI change

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
@@ -1060,7 +1060,7 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
   <data name="Hyperlink_Owner" xml:space="preserve">
     <value>Owner</value>
   </data>
-  <data name="ToolTip_SeeDotnetNuGetWhy" xml:space="preserve">
+  <data name="ToolTip_TransitivePathHelp" xml:space="preserve">
     <value>To learn more about transitive paths, see</value>
   </data>
 </root>
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
@@ -1060,10 +1060,7 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
   <data name="Hyperlink_Owner" xml:space="preserve">
     <value>Owner</value>
   </data>
-  <data name="ToolTip_UseDotnetCommand" xml:space="preserve">
-    <value>To know this package transitive origin path use command</value>
-  </data>
-  <data name="ToolTip_DotnetNuGetWhy" xml:space="preserve">
-    <value>dotnet nuget why</value>
+  <data name="ToolTip_SeeDotnetNuGetWhy" xml:space="preserve">
+    <value>To learn more about transitive paths, see</value>
   </data>
 </root>
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
@@ -1060,4 +1060,10 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
   <data name="Hyperlink_Owner" xml:space="preserve">
     <value>Owner</value>
   </data>
+  <data name="ToolTip_UseDotnetCommand" xml:space="preserve">
+    <value>To know this package transitive origin path use command</value>
+  </data>
+  <data name="ToolTip_DotnetNuGetWhy" xml:space="preserve">
+    <value>dotnet nuget why</value>
+  </data>
 </root>
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.Designer.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.Designer.cs
@@ -2259,15 +2259,6 @@ namespace NuGet.PackageManagement.UI {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to dotnet nuget why.
-        /// </summary>
-        public static string ToolTip_DotnetNuGetWhy {
-            get {
-                return ResourceManager.GetString("ToolTip_DotnetNuGetWhy", resourceCulture);
-            }
-        }
-        
-        /// <summary>
         ///   Looks up a localized string similar to Install {0} version {1}..
         /// </summary>
         public static string ToolTip_InstallButton {
@@ -2331,6 +2322,15 @@ namespace NuGet.PackageManagement.UI {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to To learn more about transitive paths, see.
+        /// </summary>
+        public static string ToolTip_SeeDotnetNuGetWhy {
+            get {
+                return ResourceManager.GetString("ToolTip_SeeDotnetNuGetWhy", resourceCulture);
+            }
+        }
+        
+        /// <summary>
         ///   Looks up a localized string similar to Settings.
         /// </summary>
         public static string ToolTip_Settings {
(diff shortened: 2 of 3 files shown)


src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
```diff
@@ -1060,4 +1060,10 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
   <data name="Hyperlink_Owner" xml:space="preserve">
     <value>Owner</value>
   </data>
+  <data name="ToolTip_UseDotnetCommand" xml:space="preserve">
+    <value>To know this package transitive origin path use command</value>
```
nkolev92: Not sure transitive origin path is the best user facing name. 
I don't have a great suggestion.

```suggestion
    <value>To find the dependency graph for this package use the `dotnet nuget why` command.</value>
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
```diff
@@ -1060,4 +1060,10 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
...
+  </data>
+  <data name="ToolTip_DotnetNuGetWhy" xml:space="preserve">
+    <value>dotnet nuget why</value>
```
nkolev92: I'm worried this gets accidentally localized.
It should be done differently.

martinrrm: Yeah I was wondering the same thing, I'm just going to "hard code" the string in the XAML

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
martinrrm: @JonDouglas What do you think about this string? I like Nikolche's suggestion

kartheekp-ms: How about "To find the dependency graph for this package, use the `dotnet nuget why` commad"?

zivkan: I don't think this will be enough information for people unfamiliar with command lines in general, and they might search Visual Studio's "features" (ctrl q) for it.

Next, SDK style proejcts have can use VS's Solution Explorer (although they need to enable "search external files").

For non-SDK style projects, customers we don't currently have anything better than dotnet nuget why, but in many of those cases, they can't use `dotnet nuget why my.csproj`. They'll have to restore in VS or with msbuild.exe, and then they'll need to pass the assets file as the "project" argument.  Hence, I don't think that just mentioning dotnet nuget why is enough.  We should at least point them to https://aka.ms/dotnet/nuget/why so they might see the docs that explain the restore + assets file steps.

But I think it would be better if we had an aka.ms link for understanding transitive dependencies. For now we can point it to the nuget 2.0 blog post, and if we improve our learn docs, then we can redirect it to those future docs, so customers using SDK style projects aren't effectively told to use dotnet nuget why on the CLI for something they can do without leaving VS.

martinrrm: Since it's a tooltip we cannot add links to it or anything clickable 😢.

nkolev92: I like @kartheekp-ms's version.

JonDouglas: Not a fan of mixing IDE & CLI guidance. Maybe a generic aka.ms to audit docs or something?

> To learn more about transitive paths, see aka.ms/nuget/audit

That way it also includes solution explorer stuff and blogs

zivkan: The URL is short enough that hopefully customers can memorise and type it out. I think it's better than telling someone who doesn't know what dotnet nuget why is, or unfamiliar with the command line in general, to run the command without any additional context.

martinrrm: I think it's better to point at aka.ms/dotnet/nuget/why at this time, nuget audit docs don't mention the command

martinrrm: image

martinrrm: I added the link to the tooltip text

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
```diff
@@ -1060,4 +1060,7 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
   <data name="Hyperlink_Owner" xml:space="preserve">
     <value>Owner</value>
   </data>
-</root>
\ No newline at end of file
+  <data name="ToolTip_SeeDotnetNuGetWhy" xml:space="preserve">
```
nkolev92: I'm guessing this is supposed to be transitive path help now that we're not saying dotnet nuget why explicitly/

donnie-msft (commented):
Need a space here: 
image

nkolev92 (approved):

donnie-msft (approved):
I'm not sure if you talked to UX about this change, but I think you've done as good as can be done with that existing tooltip! 
My ideal would be a clickable link, but that requires a different style of tooltip.

Also, I'm assuming you tested the non-transitive tooltips aren't affected by the change.

zivkan (commented):
from: https://github.com/NuGet/NuGet.Client/pull/5979#discussion_r1722375569

> Since it's a tooltip we cannot add links to it or anything clickable 😢.

In the code editor, when you hover over a keyword, it shows XMLDoc as a kind of tooltip, but as long as you keep the mouse cursor over the window, it doesn't disappear. And parts of the XMLDoc are clickable. 

Can we reach out to internal people in VS, and see if this control is reusable?   Or maybe it's not unreasonably difficult to implement in WPF? (in a follow up PR, not this one)

zivkan (approved):

PR #5965, opened 2024-08-13
Implement transitive pinning in new dependency graph resolver

jeffkl:
## Description

This adds the transitive pinning feature to the new dependency resolution algorithm.

## PR Checklist

- [x] Meaningful title, helpful description and a linked NuGet/Home issue
- [ ] ~~Added tests~~ Existing tests cover scenario
- [ ] ~~Link to an issue or pull request to update docs if this PR changes settings, environment variables, new feature, etc.~~ Not needed

PR #5980, opened 2024-08-19
Address feedback from original change to introduce new dependency resolution algorithm

jeffkl:
# Bug

Tracked at https://github.com/NuGet/Client.Engineering/issues/2959

## Description
My original change to introduce a new dependency resolution algorithm was rushed and this change address the following feedback:

* See if we can get rid of the new exception handling path of AggregateException.  This was only needed because some code paths call `.GetAwaiter().GetResult()` instead of `await` but we undid some of that during the final refactoring.  (https://github.com/NuGet/NuGet.Client/pull/5963#discussion_r1713137804)
* Consider moving SharedStringBuilder to NuGet.Versioning instead of having it a "shared" compiled resource (https://github.com/NuGet/NuGet.Client/pull/5963#discussion_r1713110371)
* Use an inline suppression for ProjectRestoreCommand.GetRuntimeGraph() catching a general exception and remove it from GlobalSupressions.cs (https://github.com/NuGet/NuGet.Client/pull/5963#discussion_r1713111541)
* Fix whitespace in ProjectRestoreCommand.DownloadDependenciesAsync()  (https://github.com/NuGet/NuGet.Client/pull/5963#discussion_r1713112453)
* Fix whitespace in RestoreCommand.ExecuteRestoreAsync() (https://github.com/NuGet/NuGet.Client/pull/5963#discussion_r1713138458)
* Consider using `NoAllocEnumerator` or a `for` loop in `RestoreCommand.CreateFrameworkRuntimePairs()` (https://github.com/NuGet/NuGet.Client/pull/5963#discussion_r1713140673)
* Reconsider marking properties in `RestoreTargetGraph` as `internal` and instead use a new constructor or `init` (https://github.com/NuGet/NuGet.Client/pull/5963#discussion_r1713142543)
* Use extension method for getting the TargetFramework from the PackageSpec (https://github.com/NuGet/NuGet.Client/pull/5965/files#r1715638641)
* Use new property from PackageSpec in [PackageSpecReferenceDependencyProvider.cs](https://github.com/NuGet/NuGet.Client/pull/5965/files/fd329745e696406d6d86bfbf865fdbe0ea40b4ef#diff-b7633a3b599a74c8939203cd6ce8907c506beb2fb3d5f4c22cbe81651103d690) (https://github.com/NuGet/NuGet.Client/pull/5965#discussion_r1715668576)

I broke each one out into its own commit if that's eaiser to review.

## PR Checklist

- [ ] Added tests


src/NuGet.Core/NuGet.Commands/RestoreCommand/ProjectRestoreCommand.cs
```diff
@@ -210,13 +212,13 @@ internal static async Task<DownloadDependencyResolutionResult[]> DownloadDepende
         {
             telemetryActivity.StartIntervalMeasure();
 
-            var downloadDependencyResolutionTasks = new List<Task<DownloadDependencyResolutionResult>>();
-            foreach (var targetFrameworkInformation in packageSpec.TargetFrameworks)
+            List<Task<DownloadDependencyResolutionResult>> downloadDependencyResolutionTasks = new();
```
nkolev92:
```suggestion
            List<Task<DownloadDependencyResolutionResult>> downloadDependencyResolutionTasks = new(packageSpec.TargetFrameworks.Count);
```

Minimal gains, good practice.

src/NuGet.Core/NuGet.Commands/RestoreCommand/ProjectRestoreCommand.cs
```diff
@@ -210,13 +212,13 @@ internal static async Task<DownloadDependencyResolutionResult[]> DownloadDepende
...
+
+            foreach (TargetFrameworkInformation targetFrameworkInformation in packageSpec.TargetFrameworks)
```
nkolev92: Use NoAllocEnumerate. 

Similar as the rest, good practice, minimal gains.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -117,7 +117,7 @@ private readonly Dictionary<RestoreTargetGraph, Dictionary<string, LibraryInclud
         private const string AuditDurationOutput = "Audit.Duration.Output";
         private const string AuditDurationTotal = "Audit.Duration.Total";
 
-        private readonly bool _enableNewDependencyResolver;
+        private readonly bool _enableNewDependencyResolver = true;
```
nkolev92: I'm not a fan of this change. 
readonly values with defaults set outside of the constructor can lead to bugs. 
It's best if the property is set in a single place like the constructor.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -1224,10 +1227,6 @@ private async Task<IEnumerable<RestoreTargetGraph>> ExecuteLegacyRestoreAsync(
                 {
                     failed = true;
                 }
-                catch (AggregateException e) when (e.InnerException is FatalProtocolException)
```
nkolev92, line 1227: Can you help me understand why this is being done? 

This is regressing a previous improvement from https://github.com/NuGet/NuGet.Client/commit/52dd383ae0899680f487ff31e9dd63d0033eea32.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -1490,20 +1481,25 @@ internal static List<ExternalProjectReference> GetProjectReferences(RestoreReque
...
-        internal static IEnumerable<FrameworkRuntimePair> CreateFrameworkRuntimePairs(
-            PackageSpec packageSpec,
-            ISet<string> runtimeIds)
+        /// <summary>
+        /// Gets the list of framework/runtime pairs to restore for the project.  The list is sorted by frameworks first, then frameworks with runtimes.
+        /// </summary>
+        /// <param name="packageSpec">The <see cref="PackageSpec" /> with information about the project.</param>
+        /// <param name="runtimeIds">An <see cref="ISet{T}" /> containing the list of runtime identifiers.</param>
+        /// <returns>An <see cref="IEnumerable{T}" /> containing <see cref="FrameworkRuntimePair" /> objects with the frameworks with empty runtime identifiers followed by frameworks with the specified runtime identifiers.</returns>
+        internal static IEnumerable<FrameworkRuntimePair> CreateFrameworkRuntimePairs(PackageSpec packageSpec, ISet<string> runtimeIds)
```
nkolev92: The return type should be a list.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -1490,20 +1481,25 @@ internal static List<ExternalProjectReference> GetProjectReferences(RestoreReque
...
         {
-            var projectFrameworkRuntimePairs = new List<FrameworkRuntimePair>();
-            foreach (var framework in packageSpec.TargetFrameworks)
+            List<FrameworkRuntimePair> projectFrameworkRuntimePairs = new();
```
nkolev92: The size is knowable here. Let's use it.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreTargetGraph.cs
```diff
@@ -17,52 +17,48 @@ public class RestoreTargetGraph : IRestoreTargetGraph
         /// <summary>
         /// Gets the runtime identifier used during the restore operation on this graph
         /// </summary>
-        public string RuntimeIdentifier { get; internal set; }
+        public string RuntimeIdentifier { get; }
```
nkolev92, line 20: Given that there's so many changes in this file, can we nullable annotate this now?
It'll be super helfpul

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreTargetGraph.cs
```diff
@@ -17,52 +17,48 @@ public class RestoreTargetGraph : IRestoreTargetGraph
...
 
         /// <summary>
         /// Gets the <see cref="NuGetFramework" /> used during the restore operation on this graph
         /// </summary>
-        public NuGetFramework Framework { get; internal set; }
+        public NuGetFramework Framework { get; }
```
nkolev92, line 25: I don't see a shipped.txt change here, I'm guessing it's what's failing the build

nkolev92 (commented):
Thanks for addressing all this feedback :)

src/NuGet.Core/NuGet.Commands/RestoreCommand/ProjectRestoreCommand.cs
jeffkl: I've updated the method to used properly sized lists and NoAllocEnumerate

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
jeffkl: Okay

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
jeffkl: A previous version of the prototype had bypassed async functionality by accessing the `.Result` property which was causing exceptions to be made into `AggregateExceptions`.  This was causing tests to fail as the logging code is catching stuff like `ProtocolException`.  In this code path, it wouldn't set `failed` to `true` causing unexpected behavior.

In the final version of the implementation, I have switched to using `.GetAwaiter().GetResult()` which doesn't aggregate exceptions so I think this is not needed anymore.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
jeffkl: Done

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
jeffkl: Done

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreTargetGraph.cs
jeffkl: I think we should tackle that in a different change

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreTargetGraph.cs
jeffkl: This didn't change the public API

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreTargetGraph.cs
nkolev92: We tend to take a best effort approach towards nullability when the type is changed a lot, which I'd say in this case is a bunch of changes. 
I'd prefer if it was done in this PR, but I'm not gonna block this PR on that. 
Your call.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreTargetGraph.cs
nkolev92: Oh duh, internal :D my bad.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -146,7 +146,18 @@ public RestoreCommand(RestoreRequest request)
...
+            if (request.Project.RestoreMetadata.ProjectStyle == ProjectStyle.PackageReference
+                && !_request.Project.RestoreMetadata.CentralPackageTransitivePinningEnabled
+                && !_request.Project.RestoreMetadata.UseLegacyDependencyResolver)
+            {
+                _enableNewDependencyResolver = true;
+            }
+            else
+            {
+                _enableNewDependencyResolver = false;
+                _request.Project.RestoreMetadata.UseLegacyDependencyResolver = true;
```
nkolev92: The PackageSpec should not be edited in the restore command. It should really be treated as immutable/
If you look at the telemetry I added, it differentiates between property specified and used and legacy algorithm used because the new resolver doesn't support a certain scenario.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -1490,20 +1485,26 @@ internal static List<ExternalProjectReference> GetProjectReferences(RestoreReque
...
+        /// </summary>
+        /// <param name="packageSpec">The <see cref="PackageSpec" /> with information about the project.</param>
+        /// <param name="runtimeIds">An <see cref="ISet{T}" /> containing the list of runtime identifiers.</param>
+        /// <returns>A <see cref="List{T}" /> containing <see cref="FrameworkRuntimePair" /> objects with the frameworks with empty runtime identifiers followed by frameworks with the specified runtime identifiers.</returns>
+        internal static List<FrameworkRuntimePair> CreateFrameworkRuntimePairs(PackageSpec packageSpec, ISet<string> runtimeIds)
         {
-            var projectFrameworkRuntimePairs = new List<FrameworkRuntimePair>();
-            foreach (var framework in packageSpec.TargetFrameworks)
+            // Create a list with capacity for each framework with no runtime and each framework/runtime
+            List<FrameworkRuntimePair> projectFrameworkRuntimePairs = new(capacity: packageSpec.TargetFrameworks.Count + (packageSpec.TargetFrameworks.Count * runtimeIds.Count));
```
nkolev92: I think it's `packageSpec.TargetFrameworks.Count * (runtimeIds.Count + 1)` because of the framework, null rid combos that always exist.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
nkolev92: I think I misread this code change. 
There's a FatalProtocolException catch right above it that still exists there. So it's not really regressing anything. 

Resolving.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
jeffkl: Without this, I have to update a few dozen tests because they only set the `CentralPackageTransitivePinningEnabled` property and not the `UseLegacyDependencyResolver` property.  This seemed cleaner to me than having to update all of those tests.

When I get transitive pinning working for the new resolver, I should be able to remove this line.

nkolev92: Changing the package spec can mess with no-op of related projects.
We probably don't have tests like that, but it's how we've been able to reduce the amount of package spec cloning we do.

nkolev92: Basically https://github.com/NuGet/NuGet.Client/pull/5930 is what we'll want eventually.

nkolev92: > Without this, I have to update a few dozen tests because they only set the CentralPackageTransitivePinningEnabled property and not the UseLegacyDependencyResolver property. This seemed cleaner to me than having to update all of those tests.


Can PackageSpecReferenceDependencyProvider repeat the same check as the one here?
That should remove the need for the test code to change.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
nkolev92: I misread it. I thought you only had multiplication

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
jeffkl: Technically yes, it just places the logic in two places which was how it was before so I had changed it to read this property instead.

nkolev92: So I expect this to be temporary until we fix the transitive pinning correct?

If that's the case I'd rather duplicate some code, than risk breaking no-op and cause more restores.

zivkan: > Changing the package spec can mess with no-op of related projects.

In general, yes. In this case, it's being mutated in `RestoreCommand`'s constructor, which happens before no-op check (ExecuteAsync?). Personally, I don't care if the first restore after a VS/.NET upgrade fails no-op, only that for the same version of NuGet with the same project inputs has working no-op.

I also completely agree that PackageSpec and DGSpec should be immutable, but even if that was already implemented, this code could have done something like:

```cs
var newRestoreMetadata = new RestoreMetadata(request.Project.RestoreMetadata) { UseLegacy = true} ;
var newProject = new Project(request.Project) { RestoreMetadata = newRestoreMetadata };
_request = new RestoreRequest(request) { Project = newProject };
```

Again, this is in the constructor, before the restore actually starts. Generally agree that our code shouldn't do this, we should design Nuget to deserialize the project inputs into a PackageSpec and then never modify or clone/copy it.

I wish that PackageSpecReferenceDependencyProvider was in NuGet.Commands, not NuGet.ProjectModel (which in my mind should only has DTO objects needed for the DGSpec and maybe APIs to (de)serialize it). In which case we could have an internal helper function and call it in two places, but that's not code we have, so either we duplicate business logic (I have no hesitation in duplicate boilerplate code, but business logic is where Don't Repeat Yourself reduces bugs)

But at the same time, it's temporary until the new resolver implements the feature, and it's a simple if statement that needs to be duplicated.  I'm unconvinced that there's a best solution here. I'm just sharing my thoughts, in case anyone cares.  This PR looks good to me, and I'd be happy to approve once this conversation is resolved.

nkolev92: PackageSpecReferenceDependencyProvider has the PackageSpec so it can check the exact same condition.

jeffkl: I've removed the setting of this for now.

nkolev92 (approved):

PR #5981, opened 2024-08-20
Make NuGet.CommandLine.XPlat easier to debug

zivkan:
# Bug

Fixes:  engineering/tech debt

## Description

Working on NuGet.CommandLine.XPlat is difficult, because you need to uncomment a PackageReference in the project file, and a few lines in Program.cs, in order to debug and run locally (without patching a .NET SDK). While that's not difficult, it's frustrating to have to remember to not commit those changes every time you commit anything (and I'm used to committing all unstaged changes), which leads to frequent errors and then more commits to put back the comments.

This PR bring back what we had 2+ years ago, where by default, in debug builds, the MSBuild Locator is used to make F5 debug work "out of the box".  If there is anyone who doesn't like this behaviour for some reason, they can set a `UseMSBuildLocator` environment variable to `false` and it won't happen, and I also set it up to not use the MSBuild locator in CI builds. Hopefully that satisfies whoever didn't like the choose-when debug setup we had before.

## PR Checklist

- [x] Meaningful title, helpful description ~and a linked NuGet/Home issue~
- [x] ~Added tests~ Tests not applicable for this engineering change
- [x] ~Link to an issue or pull request to update docs if this PR changes settings, environment variables, new feature, etc.~ docs n/a

### DIFF
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs
@@ -36,7 +36,6 @@ namespace NuGet.CommandLine.XPlat
         public static int MainInternal(string[] args, CommandOutputLogger log)
         {
 #if USEMSBUILDLOCATOR
-            // Uncomment the following when debugging. Also uncomment the PackageReference for Microsoft.Build.Locator.
             try
             {
                 // .NET JIT compiles one method at a time. If this method calls `MSBuildLocator` directly, the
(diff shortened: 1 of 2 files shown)


donnie-msft (dismissed):
Should get a review from @jeffkl

nkolev92 (dismissed):

src/NuGet.Core/NuGet.CommandLine.XPlat/Program.cs
```diff
@@ -36,9 +35,9 @@ public static int Main(string[] args)
         /// </summary>
         public static int MainInternal(string[] args, CommandOutputLogger log)
         {
-#if DEBUG
+#if USEMSBUILDLOCATOR
             // Uncomment the following when debugging. Also uncomment the PackageReference for Microsoft.Build.Locator.
```
jeffkl: You can get rid of this comment now

jeffkl (dismissed):

nkolev92 (approved):

PR #5989, opened 2024-08-23
NuGet should normalize file paths when extracting

### LINKED ISSUE #13732: Package extraction on Linux when nupkg central directory uses back slashes
### NuGet Product Used

dotnet.exe

### Product Version

all

### Worked before?

_No response_

### Impact

### Repro Steps & Context

If a package author creates a package on Windows, but does not use NuGet's official tooling to create the package (for example, say they have a PowerShell script that extracts the package, makes a modification, then zips it up again), the zip file's central directory can use `\` as path separators, rather than `/`.

On Linux (and maybe Mac), `\` is not a path separator, and is a valid character in directory and file names. When I, as a package consumer, reference this package, NuGet will create a file named `path\to\file.ext` in the package directory, rather than a subdirectory named `{package directory}/path/to` and a file `file.ext`.  Therefore, when NuGet tries to do asset selection, it won't find any files named `lib/{tfm}/*.dll` and the package doesn't work.

As a package consumer on Linux, there's nothing I can do to fix this. I need the package author and nuget server to provide a different package version.

### Verbose Logs

_No response_

zivkan:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13732

## Description

There are other known issue in NuGet/Home about forward slashes vs back slashes, particularly around file globbing. This PR does not try to address any of those.  This PR only addresses when a nupkg's zip central directory contains files whose paths use `\` instead of `/`, and the package is being extracted on Linux or Mac which only use `/` as a path separator (Windows uses both).

This makes NuGet more resilient to package authors who create (or modify) their package without using official tools.

## PR Checklist

- [x] Added tests
- [x] ~Link to an issue or pull request to update docs if this PR changes settings, environment variables, new feature, etc.~ PR makes NuGet resilient to "poorly authored" packages. I don't believe it's something we want to document, certainly not encourage.

### DIFF
--- a/src/NuGet.Core/NuGet.Packaging/PackageArchiveReader.cs
+++ b/src/NuGet.Core/NuGet.Packaging/PackageArchiveReader.cs
@@ -312,9 +312,21 @@ namespace NuGet.Packaging
                     packageFileName = packageFileName.Substring(1);
                 }
 
-                // ZipArchive always has forward slashes in them. By replacing them with DirectorySeparatorChar;
-                // in windows, we get the windows-style path
-                var normalizedPath = Uri.UnescapeDataString(packageFileName.Replace('/', Path.DirectorySeparatorChar));
+                // While NuGet creates packages with forward slashes for directory separators, tools created by other people
+                // often do not normalize to forward slash when creating the zip on Windows. But Linux and Mac do not use
+                // back slashes as directory separators. In fact, back slash is a valid character in a file entry name.
+                // Therefore, normalize the path to ensure that packages created by unofficial tools are not broken
+                // on these platforms.
+                // Windows allows forward slashes as an alternate directory separator, so no normalization is needed there.
+                string normalizedPath;
+                if (RuntimeEnvironmentHelper.IsWindows)
+                {
+                    normalizedPath = Uri.UnescapeDataString(packageFileName);
+                }
+                else
+                {
+                    normalizedPath = Uri.UnescapeDataString(packageFileName.Replace('\\', Path.DirectorySeparatorChar));
+                }
 
                 destination = NormalizeDirectoryPath(destination);
                 ValidatePackageEntry(destination, normalizedPath, packageIdentity);
--- a/test/NuGet.Core.Tests/NuGet.Packaging.Test/PackageExtractorTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Packaging.Test/PackageExtractorTests.cs
@@ -5111,6 +5111,8 @@ namespace NuGet.Packaging.Test
             }
         }
 
+        // This test is likely to only catch bugs on Linux/Mac, but there's little harm in running the test
+        // on Windows as well.
         [Fact]
         public async Task InstallFromSourceAsync_NupkgWithNonNormalizedPaths_FixesPathsToExtract()
         {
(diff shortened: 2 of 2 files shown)


src/NuGet.Core/NuGet.Packaging/PackageArchiveReader.cs
```diff
@@ -312,9 +312,21 @@ public override IEnumerable<string> CopyFiles(
...
+                // While NuGet creates packages with forward slashes for directory separators, tools created by other people
+                // often do not normalize to forward slash when creating the zip on Windows. But Linux and Mac do not use
+                // back slashes as directory separators. In fact, back slash is a valid character in a file entry name.
+                // Therefore, normalize the path to ensure that packages created by unofficial tools are not broken
+                // on these platforms.
+                // Windows allows forward slashes as an alternate directory separator, so no normalization is needed there.
+                string normalizedPath;
+                if (RuntimeEnvironmentHelper.IsWindows)
+                {
+                    normalizedPath = Uri.UnescapeDataString(packageFileName);
```
Nigusu-Allehu, line 324: Suggestion: maybe we should still replace '/' with Path.DirectorySeperatorChar for windows?  That way we get the platform directory separator. However, it is of course highly likely it will always be '/' for windows.

zivkan: we've made a lot of performance gains by removing unnecessary memory allocations and therefore garbage collections. I'm hoping we could avoid the string allocations. Since this happens for every file in the package, and packages always have multiple files, it's a lot of allocations. It's less impactful on developer machines where a package is extracted into the global packages folder once, but it's very high impact on CI, where packages are typically downloaded and extracted on every build.

kartheekp-ms (approved):

test/NuGet.Core.Tests/NuGet.Packaging.Test/PackageExtractorTests.cs
```diff
@@ -5111,12 +5131,64 @@ public async Task InstallFromSourceAsync_WithPackageDownloader_LogsSourceOnNorma
...
+            }
+        }
+
         private static bool FileExistsCaseSensitively(string expectedFilePath)
         {
             var directoryPath = Path.GetDirectoryName(expectedFilePath);
-
             return Directory
-                .GetFiles(directoryPath, "*", SearchOption.TopDirectoryOnly)
+            .GetFiles(directoryPath, "*", SearchOption.TopDirectoryOnly)
```
donnie-msft, line 5191: Revert? 4 spaces is our standard as I understand it

test/NuGet.Core.Tests/NuGet.Packaging.Test/PackageExtractorTests.cs
```diff
@@ -5111,12 +5131,64 @@ public async Task InstallFromSourceAsync_WithPackageDownloader_LogsSourceOnNorma
             }
         }
 
+        // This test is likely to only catch bugs on Linux/Mac, but there's little harm in running the test
+        // on Windows as well.
+        [Fact]
+        public async Task InstallFromSourceAsync_NupkgWithNonNormalizedPaths_FixesPathsToExtract()
```
donnie-msft, line 5137: Did you try this test before your change. I assume it failed on Linux or Mac or at least you're confident it would have failed? It definitely succeeds on Windows without this change as I tested that.

PR #5990, opened 2024-08-24
Discover plugins installed using Net tools

Nigusu-Allehu:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13740

## Description

This PR makes sure NuGet discovers credential providers installed using .NET tools.
`dotnet tool install <provider package name> --global` should install the tool and add its path to the` PATH` environmental varibale. Iterate through all the paths in` PATH` and all the files in the path to find the plugins. By design, we expect these plugins to be named `nuget-plugin*`.

> #### This PR introduces a new TFM, net8.0, to `NuGet.Protocol`. The primary reason for adding net8.0 support is to take advantage of newer APIs, specifically `File.GetUnixFileMode(fileInfo.FullName)`, which simplifies handling file permissions in Unix-based systems.

### Plugin discovery method
**The discovery is done based on which environmental variable has been specified**
* If `NUGET_NETFX_PLUGIN_PATHS ` or `NUGET_NETCORE_PLUGIN_PATHS ` have been specified, use the old discovery method to look up non dotnet tools plugins in the paths added in the env variable.
*  If `NUGET_NETFX_PLUGIN_PATHS ` and `NUGET_NETCORE_PLUGIN_PATHS` have not been specified
   * if `NUGET_PLUGIN_PATHS` specified,
      * Collect the set of plugins the env variable is pointing to: They could be donet tools or non dotnet tools plugin files. 
   *  if `NUGET_PLUGIN_PATHS` not specified,
      * Discover all plugins using the default discovery method

#### Default discovery method
 * For dotnet tools
   * Traverse all the directories specified in `PATH` env variable, and search for executable `nuget-plugin-*`files
 * For non dotnet tools 
   * %UserProfile%/.nuget/plugins 

## PR Checklist

- [x] Added tests
- [x] Link to an issue or pull request to update docs if this PR changes settings, environment variables, new feature, etc. https://github.com/NuGet/Home/issues/13858

### DIFF
--- a/src/NuGet.Core/NuGet.Protocol/NuGet.Protocol.csproj
+++ b/src/NuGet.Core/NuGet.Protocol/NuGet.Protocol.csproj
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>$(TargetFrameworksLibraryForSigning)</TargetFrameworks>
+    <TargetFrameworks>$(TargetFrameworksLibraryForSigning);$(NETCoreTargetFramework)</TargetFrameworks>
     <TargetFramework />
     <NoWarn>$(NoWarn);CS1591;CS1573;CS0012;RS0041</NoWarn>
     <PackageTags>nuget protocol</PackageTags>
--- a/src/NuGet.Core/NuGet.Protocol/NuGet.Protocol.csproj
+++ b/src/NuGet.Core/NuGet.Protocol/NuGet.Protocol.csproj
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>$(TargetFrameworksLibraryForSigning);$(NETCoreTargetFramework)</TargetFrameworks>
+    <TargetFrameworks>$(TargetFrameworksLibraryForSigning)</TargetFrameworks>
     <TargetFramework />
     <NoWarn>$(NoWarn);CS1591;CS1573;CS0012;RS0041</NoWarn>
     <PackageTags>nuget protocol</PackageTags>
--- a/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
@@ -97,13 +97,31 @@ namespace NuGet.Protocol.Plugins
                     return _results;
                 }
 
-                _pluginFiles = GetPluginFiles(cancellationToken);
-
-                if (string.IsNullOrEmpty(_envVariableNetCoreAndNetFXPluginPaths))
+                if (!string.IsNullOrEmpty(_envVariableNetCoreAndNetFXPluginPaths))
                 {
-                    // Nuget plugin configuration environmental variables: NUGET_NETFX_PLUGIN_PATHS, NUGET_NETCORE_PLUGIN_PATHS
-                    // were not used to point to the credential provider plugins.
-                    _pluginFiles.AddRange(GetPluginsInNuGetPluginPathsAndPath() ?? new List<PluginFile>());
+                    // NUGET_NETFX_PLUGIN_PATHS, NUGET_NETCORE_PLUGIN_PATHS have been set.
+                    var filePaths = _envVariableNetCoreAndNetFXPluginPaths.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
+                    _pluginFiles = GetPluginFiles(filePaths, cancellationToken);
+                }
+                else if (!string.IsNullOrEmpty(_envVariableNuGetPluginPaths))
+                {
+                    // NUGET_PLUGIN_PATHS has been set
+                    _pluginFiles = GetPluginsInNuGetPluginPaths();
+                }
+                else
+                {
+                    // restore to default plugins search.
+                    // Search for plugins in %user%/.nuget/plugins
+                    var directories = new List<string> { PluginDiscoveryUtility.GetNuGetHomePluginsPath() };
+#if IS_DESKTOP
+                    // Internal plugins are only supported for .NET Framework scenarios, namely msbuild.exe
+                    directories.Add(PluginDiscoveryUtility.GetInternalPlugins());
+#endif
+                    var filePaths = PluginDiscoveryUtility.GetConventionBasedPlugins(directories);
+                    _pluginFiles = GetPluginFiles((string[])filePaths, cancellationToken);
+
+                    // Search for .Net tools plugins in PATH
+                    _pluginFiles.AddRange(GetPluginsInPATH() ?? new List<PluginFile>());
                 }
 
                 var results = new List<PluginDiscoveryResult>();
@@ -127,11 +145,9 @@ namespace NuGet.Protocol.Plugins
             return _results;
         }
 
-        private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
+        private List<PluginFile> GetPluginFiles(string[] filePaths, CancellationToken cancellationToken)
         {
             cancellationToken.ThrowIfCancellationRequested();
-
-            var filePaths = GetPluginFilePaths();
 
             var files = new List<PluginFile>();
 
@@ -155,58 +171,96 @@ namespace NuGet.Protocol.Plugins
         }
 
         /// <summary>
-        /// Retrieves authentication plugins by searching through directories specified in the `NuGET_PLUGIN_PATHS` and `PATH` 
-        /// environment variables. The method looks for files prefixed with 'nuget-plugin-' and verifies their validity for .net tools plugins.
+        /// Retrieves authentication plugins by searching through directories and files specified in the `NuGET_PLUGIN_PATHS`
+        /// environment variable. The method looks for files prefixed with 'nuget-plugin-' and verifies their validity for .net tools plugins.
         /// </summary>
         /// <returns>A list of valid <see cref="PluginFile"/> objects representing the discovered plugins.</returns>
-        internal List<PluginFile> GetPluginsInNuGetPluginPathsAndPath()
+        internal List<PluginFile> GetPluginsInNuGetPluginPaths()
         {
             var pluginFiles = new List<PluginFile>();
-            var nugetPluginPaths = _environmentVariableReader.GetEnvironmentVariable(EnvironmentVariableConstants.PluginPaths);
-            string[] paths;
-
-            if (string.IsNullOrEmpty(nugetPluginPaths))
-            {
-                nugetPluginPaths = _environmentVariableReader.GetEnvironmentVariable("PATH");
-            }
-
-            paths = nugetPluginPaths?.Split(Path.PathSeparator) ?? Array.Empty<string>();
+            string[] paths = _envVariableNuGetPluginPaths?.Split(Path.PathSeparator) ?? Array.Empty<string>();
 
             foreach (var path in paths)
             {
-                if (File.Exists(path))
+                if (PathValidator.IsValidLocalPath(path) || PathValidator.IsValidUncPath(path))
                 {
-                    FileInfo fileInfo = new FileInfo(path);
-                    if (fileInfo.Name.StartsWith("nuget-plugin-", StringComparison.CurrentCultureIgnoreCase))
+                    if (File.Exists(path))
                     {
-                        // A DotNet tool plugin
-                        if (IsValidPluginFile(fileInfo))
+                        FileInfo fileInfo = new FileInfo(path);
+                        if (fileInfo.Name.StartsWith("nuget-plugin-", StringComparison.CurrentCultureIgnoreCase))
                         {
-                            PluginFile pluginFile = new PluginFile(fileInfo.FullName, new Lazy<PluginFileState>(() => PluginFileState.Valid), isDotnetToolsPlugin: true);
-                            pluginFiles.Add(pluginFile);
+                            // A DotNet tool plugin
+                            if (IsValidPluginFile(fileInfo))
+                            {
+                                PluginFile pluginFile = new PluginFile(fileInfo.FullName, new Lazy<PluginFileState>(() => PluginFileState.Valid), isDotnetToolsPlugin: true);
+                                pluginFiles.Add(pluginFile);
+                            }
+                        }
+                        else
+                        {
+                            // A non DotNet tool plugin file
+                            var state = new Lazy<PluginFileState>(() => _verifier.IsValid(fileInfo.FullName) ? PluginFileState.Valid : PluginFileState.InvalidEmbeddedSignature);
+                            pluginFiles.Add(new PluginFile(fileInfo.FullName, state));
                         }
                     }
-                    else
+                    else if (Directory.Exists(path))
                     {
-                        // A non DotNet tool plugin file
-                        var state = new Lazy<PluginFileState>(() => _verifier.IsValid(fileInfo.FullName) ? PluginFileState.Valid : PluginFileState.InvalidEmbeddedSignature);
-                        pluginFiles.Add(new PluginFile(fileInfo.FullName, state));
+                        var directoryInfo = new DirectoryInfo(path);
+                        var files = directoryInfo.GetFiles("nuget-plugin-*");
+
+                        foreach (var file in files)
+                        {
+                            if (IsValidPluginFile(file))
+                            {
+                                // .NET SDK recently package signature verification for .NET tools, as a result we expect them to be valid.
+                                PluginFile pluginFile = new PluginFile(file.FullName, new Lazy<PluginFileState>(() => PluginFileState.Valid), isDotnetToolsPlugin: true);
+                                pluginFiles.Add(pluginFile);
+                            }
+                        }
                     }
                 }
-                else if (Directory.Exists(path))
+                else
                 {
-                    var directoryInfo = new DirectoryInfo(path);
-                    var files = directoryInfo.GetFiles("nuget-plugin-*");
+                    pluginFiles.Add(new PluginFile(path, new Lazy<PluginFileState>(() => PluginFileState.InvalidFilePath)));
+                }
+            }
 
-                    foreach (var file in files)
+            return pluginFiles;
+        }
+
+        /// <summary>
+        /// Retrieves .NET tools authentication plugins by searching through directories specified in `PATH` 
+        /// </summary>
+        /// <returns>A list of valid <see cref="PluginFile"/> objects representing the discovered plugins.</returns>
+        internal List<PluginFile> GetPluginsInPATH()
+        {
+            var pluginFiles = new List<PluginFile>();
+            var nugetPluginPaths = _environmentVariableReader.GetEnvironmentVariable("PATH");
+            string[] paths = nugetPluginPaths?.Split(Path.PathSeparator) ?? Array.Empty<string>();
+
+            foreach (var path in paths)
+            {
+                if (PathValidator.IsValidLocalPath(path) || PathValidator.IsValidUncPath(path))
+                {
+                    if (Directory.Exists(path))
                     {
-                        if (IsValidPluginFile(file))
+                        var directoryInfo = new DirectoryInfo(path);
+                        var files = directoryInfo.GetFiles("nuget-plugin-*");
+
+                        foreach (var file in files)
                         {
-                            // .NET SDK recently package signature verification for .NET tools, as a result we expect them to be valid.
-                            PluginFile pluginFile = new PluginFile(file.FullName, new Lazy<PluginFileState>(() => PluginFileState.Valid), isDotnetToolsPlugin: true);
-                            pluginFiles.Add(pluginFile);
+                            if (IsValidPluginFile(file))
+                            {
+                                // .NET SDK recently package signature verification for .NET tools, as a result we expect them to be valid.
+                                PluginFile pluginFile = new PluginFile(file.FullName, new Lazy<PluginFileState>(() => PluginFileState.Valid), isDotnetToolsPlugin: true);
+                                pluginFiles.Add(pluginFile);
+                            }
                         }
                     }
+                }
+                else
+                {
+                    pluginFiles.Add(new PluginFile(path, new Lazy<PluginFileState>(() => PluginFileState.InvalidFilePath)));
                 }
             }
 
@@ -282,20 +336,5 @@ namespace NuGet.Protocol.Plugins
 #pragma warning restore CA1031 // Do not catch general exception types
         }
 #endif
-
-        private IEnumerable<string> GetPluginFilePaths()
-        {
-            if (string.IsNullOrEmpty(_envVariableNetCoreAndNetFXPluginPaths) && string.IsNullOrEmpty(_envVariableNuGetPluginPaths))
-            {
-                var directories = new List<string> { PluginDiscoveryUtility.GetNuGetHomePluginsPath() };
-#if IS_DESKTOP
-                // Internal plugins are only supported for .NET Framework scenarios, namely msbuild.exe
-                directories.Add(PluginDiscoveryUtility.GetInternalPlugins());
-#endif
-                return PluginDiscoveryUtility.GetConventionBasedPlugins(directories);
-            }
-
-            return _envVariableNetCoreAndNetFXPluginPaths != null ? _envVariableNetCoreAndNetFXPluginPaths.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries) : new List<string>();
-        }
     }
 }
--- a/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
@@ -23,30 +23,15 @@ namespace NuGet.Protocol.Plugins
         private readonly string _nuGetPluginPaths;
         private IEnumerable<PluginDiscoveryResult> _results;
         private readonly SemaphoreSlim _semaphore;
-        private readonly EmbeddedSignatureVerifier _verifier;
         private readonly IEnvironmentVariableReader _environmentVariableReader;
 
-        public PluginDiscoverer(EmbeddedSignatureVerifier verifier)
-            : this(verifier, EnvironmentVariableWrapper.Instance)
+        public PluginDiscoverer()
+            : this(EnvironmentVariableWrapper.Instance)
         {
         }
 
-        /// <summary>
-        /// Instantiates a new <see cref="PluginDiscoverer" /> class.
-        /// </summary>
-        /// <param name="rawPluginPaths">The raw semicolon-delimited list of supposed plugin file paths.</param>
-        public PluginDiscoverer(string rawPluginPaths)
-        /// <param name="verifier">An embedded signature verifier.</param>
-        /// <param name="environmentVariableReader"> A reader for environmental varibales. </param>
-        /// <exception cref="ArgumentNullException">Thrown if <paramref name="verifier" /> is <see langword="null" />.</exception>
-        internal PluginDiscoverer(EmbeddedSignatureVerifier verifier, IEnvironmentVariableReader environmentVariableReader)
+        internal PluginDiscoverer(IEnvironmentVariableReader environmentVariableReader)
         {
-            _rawPluginPaths = rawPluginPaths;
-            if (verifier == null)
-            {
-                throw new ArgumentNullException(nameof(verifier));
-            }
-
             _environmentVariableReader = environmentVariableReader;
 #if IS_DESKTOP
             _netCoreAndNetFXPluginPaths = environmentVariableReader.GetEnvironmentVariable(EnvironmentVariableConstants.DesktopPluginPaths);
@@ -54,7 +39,6 @@ namespace NuGet.Protocol.Plugins
             _netCoreAndNetFXPluginPaths = environmentVariableReader.GetEnvironmentVariable(EnvironmentVariableConstants.CorePluginPaths);
 #endif
             _nuGetPluginPaths = _environmentVariableReader.GetEnvironmentVariable(EnvironmentVariableConstants.PluginPaths);
-            _verifier = verifier;
             _semaphore = new SemaphoreSlim(initialCount: 1, maxCount: 1);
         }
 
@@ -157,7 +141,7 @@ namespace NuGet.Protocol.Plugins
             return _results;
         }
 
-        private List<PluginFile> GetPluginFiles(string[] filePaths, CancellationToken cancellationToken)
+        private static List<PluginFile> GetPluginFiles(string[] filePaths, CancellationToken cancellationToken)
         {
             cancellationToken.ThrowIfCancellationRequested();
 
@@ -211,7 +195,7 @@ namespace NuGet.Protocol.Plugins
                         else
                         {
                             // A non DotNet tool plugin file
-                            var state = new Lazy<PluginFileState>(() => _verifier.IsValid(fileInfo.FullName) ? PluginFileState.Valid : PluginFileState.InvalidEmbeddedSignature);
+                            var state = new Lazy<PluginFileState>(() => PluginFileState.Valid);
                             pluginFiles.Add(new PluginFile(fileInfo.FullName, state));
                         }
                     }
@@ -224,8 +208,6 @@ namespace NuGet.Protocol.Plugins
                         {
                             if (IsValidPluginFile(file))
                             {
-                                // .NET SDK recently package signature verification for .NET tools, as a result we expect them to be valid.
-                                // As a result the plugin created here has PluginFileState.Valid.
                                 PluginFile pluginFile = new PluginFile(file.FullName, new Lazy<PluginFileState>(() => PluginFileState.Valid), isDotnetToolsPlugin: true);
                                 pluginFiles.Add(pluginFile);
                             }
@@ -264,7 +246,6 @@ namespace NuGet.Protocol.Plugins
                         {
                             if (IsValidPluginFile(file))
                             {
-                                // .NET SDK recently package signature verification for .NET tools, as a result we expect them to be valid.
                                 PluginFile pluginFile = new PluginFile(file.FullName, new Lazy<PluginFileState>(() => PluginFileState.Valid), isDotnetToolsPlugin: true);
                                 pluginFiles.Add(pluginFile);
                             }
--- a/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
@@ -218,8 +218,6 @@ namespace NuGet.Protocol.Plugins
 #pragma warning restore CA1031 // Do not catch general exception types
         }
 
-
-
         private IEnumerable<string> GetPluginFilePaths()
         {
             if (string.IsNullOrEmpty(_rawPluginPaths))
--- a/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
@@ -79,7 +79,6 @@ namespace NuGet.Protocol.Plugins
                 }
 
                 _pluginFiles = GetPluginFiles(cancellationToken);
-
                 var results = new List<PluginDiscoveryResult>();
 
                 for (var i = 0; i < _pluginFiles.Count; ++i)
--- a/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.IO;
 using System.Runtime.InteropServices;
 using System.Threading;
@@ -254,7 +253,7 @@ namespace NuGet.Protocol.Plugins
             try
             {
                 string output;
-                using (var process = new Process())
+                using (var process = new System.Diagnostics.Process())
                 {
                     // Use a shell command to check if the file is executable
                     process.StartInfo.FileName = "/bin/bash";
--- a/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
@@ -79,8 +80,6 @@ namespace NuGet.Protocol.Plugins
 
                 _pluginFiles = GetPluginFiles(cancellationToken);
 
-                // Get plugins added using .Net Tools
-                _pluginFiles.AddRange(GetNetToolsPluginFiles(cancellationToken));
                 var results = new List<PluginDiscoveryResult>();
 
                 for (var i = 0; i < _pluginFiles.Count; ++i)
@@ -134,19 +133,14 @@ namespace NuGet.Protocol.Plugins
         /// `PATH` environment variable.
         /// The files are also expected to have a name that starts with `nuget-plugin-`
         /// </summary>
-        /// <param name="cancellationToken"></param>
         /// <returns></returns>
-        private List<PluginFile> GetNetToolsPluginFiles(CancellationToken cancellationToken)
+        private static List<string> GetNetToolsPluginFiles()
         {
-            cancellationToken.ThrowIfCancellationRequested();
-
-            var pluginFiles = new List<PluginFile>();
+            var pluginFiles = new List<string>();
             var paths = Environment.GetEnvironmentVariable("PATH")?.Split(Path.PathSeparator) ?? Array.Empty<string>();
 
             foreach (var path in paths)
             {
-                cancellationToken.ThrowIfCancellationRequested();
-
                 if (Directory.Exists(path))
                 {
                     var directoryInfo = new DirectoryInfo(path);
@@ -156,8 +150,7 @@ namespace NuGet.Protocol.Plugins
                     {
                         if (IsValidPluginFile(file))
                         {
-                            var state = new Lazy<PluginFileState>(() => _verifier.IsValid(file.FullName) ? PluginFileState.Valid : PluginFileState.InvalidEmbeddedSignature);
-                            pluginFiles.Add(new PluginFile(file.FullName, state));
+                            pluginFiles.Add(file.FullName);
                         }
                     }
                 }
@@ -236,7 +229,12 @@ namespace NuGet.Protocol.Plugins
                 // Internal plugins are only supported for .NET Framework scenarios, namely msbuild.exe
                 directories.Add(PluginDiscoveryUtility.GetInternalPlugins());
 #endif
-                return PluginDiscoveryUtility.GetConventionBasedPlugins(directories);
+                var plugins = PluginDiscoveryUtility.GetConventionBasedPlugins(directories).ToList();
+
+                // Get plugins added using .Net Tools
+                plugins.AddRange(GetNetToolsPluginFiles());
+
+                return plugins;
             }
 
             return _rawPluginPaths.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
--- a/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
-using System.Linq;
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
--- a/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
@@ -152,7 +152,7 @@ namespace NuGet.Protocol.Plugins
         /// The files are also expected to have a name that starts with `nuget-plugin-`
         /// </summary>
         /// <returns></returns>
-        private List<PluginFile> GetNetToolsPluginFiles()
+        internal List<PluginFile> GetNetToolsPluginFiles()
         {
             var pluginFiles = new List<PluginFile>();
             string[] paths = Array.Empty<string>();
@@ -177,8 +177,8 @@ namespace NuGet.Protocol.Plugins
                     {
                         if (IsValidPluginFile(file))
                         {
-                            var state = new Lazy<PluginFileState>(() => _verifier.IsValid(file.FullName) ? PluginFileState.Valid : PluginFileState.InvalidEmbeddedSignature);
-                            PluginFile pluginFile = new PluginFile(file.FullName, state, isDotnetToolsPlugin: true);
+                            // .NET SDK recently package signature verification for .NET tools, as a result we expect them to be valid.
+                            PluginFile pluginFile = new PluginFile(file.FullName, new Lazy<PluginFileState>(() => PluginFileState.Valid), isDotnetToolsPlugin: true);
                             pluginFiles.Add(pluginFile);
                         }
                     }
@@ -195,7 +195,7 @@ namespace NuGet.Protocol.Plugins
         /// </summary>
         /// <param name="fileInfo"></param>
         /// <returns></returns>
-        private static bool IsValidPluginFile(FileInfo fileInfo)
+        internal static bool IsValidPluginFile(FileInfo fileInfo)
         {
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
@@ -224,7 +224,7 @@ namespace NuGet.Protocol.Plugins
         /// </summary>
         /// <param name="fileInfo"></param>
         /// <returns></returns>
-        private static bool IsExecutable(FileInfo fileInfo)
+        internal static bool IsExecutable(FileInfo fileInfo)
         {
 #pragma warning disable CA1031 // Do not catch general exception types
             try
--- a/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
@@ -24,11 +24,11 @@ namespace NuGet.Protocol.Plugins
         private IEnumerable<PluginDiscoveryResult> _results;
         private readonly SemaphoreSlim _semaphore;
         private readonly EmbeddedSignatureVerifier _verifier;
-        private readonly IEnvironmentVariableReader _environmentVariableReader = EnvironmentVariableWrapper.Instance;
+        private readonly IEnvironmentVariableReader _environmentVariableReader;
 
-        internal PluginDiscoverer(string rawPluginPaths, EmbeddedSignatureVerifier verifier, IEnvironmentVariableReader environmentVariableReader) : this(rawPluginPaths, verifier)
+        public PluginDiscoverer(string rawPluginPaths, EmbeddedSignatureVerifier verifier)
+            : this(rawPluginPaths, verifier, EnvironmentVariableWrapper.Instance)
         {
-            _environmentVariableReader = environmentVariableReader;
         }
 
         /// <summary>
@@ -36,9 +36,14 @@ namespace NuGet.Protocol.Plugins
         /// </summary>
         /// <param name="rawPluginPaths">The raw semicolon-delimited list of supposed plugin file paths.</param>
         public PluginDiscoverer(string rawPluginPaths)
+        /// <param name="verifier">An embedded signature verifier.</param>
+        /// <param name="environmentVariableReader"> A reader for environmental varibales. </param>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="verifier" /> is <see langword="null" />.</exception>
+        internal PluginDiscoverer(string rawPluginPaths, EmbeddedSignatureVerifier verifier, IEnvironmentVariableReader environmentVariableReader)
         {
             _rawPluginPaths = rawPluginPaths;
             _semaphore = new SemaphoreSlim(initialCount: 1, maxCount: 1);
+            _environmentVariableReader = environmentVariableReader;
         }
 
         /// <summary>
--- a/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
@@ -37,7 +37,12 @@ namespace NuGet.Protocol.Plugins
 #else
             _netCoreOrNetFXPluginPaths = environmentVariableReader.GetEnvironmentVariable(EnvironmentVariableConstants.CorePluginPaths);
 #endif
-            _nuGetPluginPaths = _environmentVariableReader.GetEnvironmentVariable(EnvironmentVariableConstants.PluginPaths);
+
+            if (string.IsNullOrEmpty(_netCoreOrNetFXPluginPaths))
+            {
+                _nuGetPluginPaths = _environmentVariableReader.GetEnvironmentVariable(EnvironmentVariableConstants.PluginPaths);
+            }
+
             _semaphore = new SemaphoreSlim(initialCount: 1, maxCount: 1);
         }
 
@@ -111,11 +116,11 @@ namespace NuGet.Protocol.Plugins
                     // Search for .Net tools plugins in PATH
                     if (_pluginFiles != null)
                     {
-                        _pluginFiles.AddRange(GetPluginsInPATH() ?? new List<PluginFile>());
+                        _pluginFiles.AddRange(GetPluginsInPath());
                     }
                     else
                     {
-                        _pluginFiles = GetPluginsInPATH() ?? new List<PluginFile>();
+                        _pluginFiles = GetPluginsInPath();
                     }
                 }
 
@@ -221,7 +226,7 @@ namespace NuGet.Protocol.Plugins
         /// Retrieves .NET tools authentication plugins by searching through directories specified in `PATH` 
         /// </summary>
         /// <returns>A list of valid <see cref="PluginFile"/> objects representing the discovered plugins.</returns>
-        internal List<PluginFile> GetPluginsInPATH()
+        internal List<PluginFile> GetPluginsInPath()
         {
             var pluginFiles = new List<PluginFile>();
             var nugetPluginPaths = _environmentVariableReader.GetEnvironmentVariable("PATH");
--- a/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
@@ -89,7 +89,7 @@ namespace NuGet.Protocol.Plugins
                 if (!string.IsNullOrEmpty(_netCoreAndNetFXPluginPaths))
                 {
                     // NUGET_NETFX_PLUGIN_PATHS, NUGET_NETCORE_PLUGIN_PATHS have been set.
-                    var filePaths = _netCoreAndNetFXPluginPaths.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
+                    var filePaths = _netCoreAndNetFXPluginPaths.Split(new[] { Path.PathSeparator }, StringSplitOptions.RemoveEmptyEntries);
                     _pluginFiles = GetPluginFiles(filePaths, cancellationToken);
                 }
                 else if (!string.IsNullOrEmpty(_nuGetPluginPaths))
@@ -201,17 +201,7 @@ namespace NuGet.Protocol.Plugins
                     }
                     else if (Directory.Exists(path))
                     {
-                        var directoryInfo = new DirectoryInfo(path);
-                        var files = directoryInfo.GetFiles("nuget-plugin-*");
-
-                        foreach (var file in files)
-                        {
-                            if (IsValidPluginFile(file))
-                            {
-                                PluginFile pluginFile = new PluginFile(file.FullName, new Lazy<PluginFileState>(() => PluginFileState.Valid), isDotnetToolsPlugin: true);
-                                pluginFiles.Add(pluginFile);
-                            }
-                        }
+                        pluginFiles.AddRange(GetNetToolsPluginsInDirectory(path) ?? new List<PluginFile>());
                     }
                 }
                 else
@@ -237,24 +227,33 @@ namespace NuGet.Protocol.Plugins
             {
                 if (PathValidator.IsValidLocalPath(path) || PathValidator.IsValidUncPath(path))
                 {
-                    if (Directory.Exists(path))
-                    {
-                        var directoryInfo = new DirectoryInfo(path);
-                        var files = directoryInfo.GetFiles("nuget-plugin-*");
-
-                        foreach (var file in files)
-                        {
-                            if (IsValidPluginFile(file))
-                            {
-                                PluginFile pluginFile = new PluginFile(file.FullName, new Lazy<PluginFileState>(() => PluginFileState.Valid), isDotnetToolsPlugin: true);
-                                pluginFiles.Add(pluginFile);
-                            }
-                        }
-                    }
+                    pluginFiles.AddRange(GetNetToolsPluginsInDirectory(path) ?? new List<PluginFile>());
                 }
                 else
                 {
                     pluginFiles.Add(new PluginFile(path, new Lazy<PluginFileState>(() => PluginFileState.InvalidFilePath)));
+                }
+            }
+
+            return pluginFiles;
+        }
+
+        private static List<PluginFile> GetNetToolsPluginsInDirectory(string directoryPath)
+        {
+            var pluginFiles = new List<PluginFile>();
+
+            if (Directory.Exists(directoryPath))
+            {
+                var directoryInfo = new DirectoryInfo(directoryPath);
+                var files = directoryInfo.GetFiles("nuget-plugin-*");
+
+                foreach (var file in files)
+                {
+                    if (IsValidPluginFile(file))
+                    {
+                        PluginFile pluginFile = new PluginFile(file.FullName, new Lazy<PluginFileState>(() => PluginFileState.Valid), isDotnetToolsPlugin: true);
+                        pluginFiles.Add(pluginFile);
+                    }
                 }
             }
 
@@ -307,17 +306,18 @@ namespace NuGet.Protocol.Plugins
                 {
                     // Use a shell command to check if the file is executable
                     process.StartInfo.FileName = "/bin/bash";
-                    process.StartInfo.Arguments = $" -c \"if [ -x {fileInfo.FullName} ]; then echo yes; else echo no; fi\"";
+                    process.StartInfo.Arguments = $" -c \"if [ -x '{fileInfo.FullName}' ]; then echo yes; else echo no; fi\"";
                     process.StartInfo.UseShellExecute = false;
                     process.StartInfo.RedirectStandardOutput = true;
 
                     process.Start();
+                    output = process.StandardOutput.ReadToEnd().Trim();
+
                     if (!process.WaitForExit(1000) || process.ExitCode != 0)
                     {
+                        process.Kill();
                         return false;
                     }
-
-                    output = process.StandardOutput.ReadToEnd().Trim();
                 }
 
                 // Check if the output is "yes"
--- a/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
@@ -25,7 +25,6 @@ namespace NuGet.Protocol.Plugins
         private readonly SemaphoreSlim _semaphore;
         private readonly EmbeddedSignatureVerifier _verifier;
         private readonly IEnvironmentVariableReader _environmentVariableReader;
-        private readonly string _pluginPaths;
 
         public PluginDiscoverer(EmbeddedSignatureVerifier verifier)
             : this(verifier, EnvironmentVariableWrapper.Instance)
--- a/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
@@ -3,8 +3,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
-using System.Linq;
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
@@ -129,6 +129,13 @@ namespace NuGet.Protocol.Plugins
             return files;
         }
 
+        /// <summary>
+        /// Gets auth plugins installed using dotnet tools. This is done by iterating through each file in directories found int eh
+        /// `PATH` environment variable.
+        /// The files are also expected to have a name that starts with `nuget-plugin-`
+        /// </summary>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
         private List<PluginFile> GetNetToolsPluginFiles(CancellationToken cancellationToken)
         {
             cancellationToken.ThrowIfCancellationRequested();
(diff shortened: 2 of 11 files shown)


src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -140,6 +144,96 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+            }
+        }
+
+        /// <summary>
+        /// Checks whether a file is executable or not in Unix.
+        /// This is done by running bash code: `if [ -x {fileInfo.FullName} ]; then echo yes; else echo no; fi`
+        /// </summary>
+        /// <param name="fileInfo"></param>
+        /// <returns></returns>
+        private static bool IsExecutable(FileInfo fileInfo)
```
zivkan: Did you consider https://learn.microsoft.com/en-us/dotnet/api/system.io.file.getunixfilemode?view=net-8.0? It would be much simpler.

Although it won't work on mono, so there's a question if we want to support the new plugin model on mono.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -140,6 +144,96 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+        {
+#pragma warning disable CA1031 // Do not catch general exception types
+            try
+            {
+                string output;
+                using (var process = new Process())
+                {
+                    // Use a shell command to check if the file is executable
+                    process.StartInfo.FileName = "sh";
+                    process.StartInfo.Arguments = $"-c \"if [ -x {fileInfo.FullName} ]; then echo yes; else echo no; fi\"";
```
zivkan: Did you test this on mac?  Apparently OSX uses a very old version of bash (the default shell is zsh?), so some things that work on modern bash on linux don't work on mac, making cross platform shell scripting more difficult (since zsh often isn't isntalled on linux distros)

Nigusu-Allehu: I tested it on mac: both using zsh and sh, and it works.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: Thanks for the suggestion. Updated code to make use of the method for net8.0

src/NuGet.Core/NuGet.Protocol/NuGet.Protocol.csproj
```diff
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>$(TargetFrameworksLibraryForSigning)</TargetFrameworks>
+    <TargetFrameworks>$(TargetFrameworksLibraryForSigning);$(NETCoreTargetFramework)</TargetFrameworks>
```
kartheekp-ms: I am curious to learn more about why we need the additional framework.

kartheekp-ms: Looks like the new tfm is used while checking if a plug is valid on non-Windows platform. https://github.com/NuGet/NuGet.Client/pull/5990/files#diff-cf421e9b689e2c0e46c47e27365571384c9fd9c7eae6522a36c26f5bd40312a4R193

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -140,6 +143,105 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+        private static bool IsValidPluginFile(FileInfo fileInfo)
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                return fileInfo.Extension.Equals(".exe", StringComparison.OrdinalIgnoreCase) ||
+                       fileInfo.Extension.Equals(".bat", StringComparison.OrdinalIgnoreCase);
+            }
+            else
+            {
+#if NET8_0
```
zivkan:
```suggestion
#if NET8_0_OR_HIGHER
```

Otherwise when the project is retarged to a higher TFM in the future, this block will no longer be used

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -140,6 +143,105 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+            }
+        }
+
+        /// <summary>
+        /// Checks whether a file is executable or not in Unix.
+        /// This is done by running bash code: `if [ -x {fileInfo.FullName} ]; then echo yes; else echo no; fi`
+        /// </summary>
+        /// <param name="fileInfo"></param>
+        /// <returns></returns>
+        private static bool IsExecutable(FileInfo fileInfo)
```
zivkan: `#if !NET8_0_OR_HIGHER`

This method is not called when the `#if` block above uses the unix file permissions API.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/Plugins/PluginDiscovererTests.cs
```diff
@@ -247,6 +248,174 @@ public async Task DiscoverAsync_IsIdempotent()
...
+        public async Task DiscoverAsync_withValidDotNetToolsPluginWindows_FindsThePlugin(string fileName)
+        {
+            using (var testDirectory = TestDirectory.Create())
+            {
+                // Arrange
+                var pluginPath = Path.Combine(testDirectory.Path, "myPlugin");
+                Directory.CreateDirectory(pluginPath);
+                var myPlugin = Path.Combine(pluginPath, fileName);
+
+                Environment.SetEnvironmentVariable("PATH", pluginPath);
```
zivkan: Environment variables are per-process, which means that `Environment.SetEnvironmentVariable` is changing it for the entire process. If there are other tests running in parallel that depend on other values, this will cause those tests to become flakey.

The general pattern we use thoughout NuGet's source code is to use an interface named something like `IEnironmentVariableHelper` in the product code, so that in tests we can pass in a mock. And in the product code where we need to use real environment variables, there's something like `EnvironmentVariableHelper.Instance`.

Basically, tests should never modify shared state.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -140,6 +150,115 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+        private List<PluginFile> GetNetToolsPluginFiles()
+        {
+            var pluginFiles = new List<PluginFile>();
+
+            if (_environmentVariableReader is null)
+            {
+                _environmentVariableReader = EnvironmentVariableWrapper.Instance;
+            }
+
+            var paths = _environmentVariableReader.GetEnvironmentVariable("PATH")?.Split(Path.PathSeparator) ?? Array.Empty<string>();
```
kartheekp-ms: In the [spec](https://github.com/NuGet/Home/blob/dev/accepted/2024/support-nuget-authentication-plugins-dotnet-tools.md?plain=1#L165-L172), we have mentioned that, if customers prefer to install NuGet plugins as a [tool-path global tool](https://learn.microsoft.com/dotnet/core/tools/global-tools-how-to-use#use-the-tool-as-a-global-tool-installed-in-a-custom-location), they can set the `NUGET_PLUGIN_PATHS` environment variable. This variable should point to the location of the .NET Tool executable.

I think we are not handling this case in this PR. Please let me know if I am missing something.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -140,6 +150,115 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+        /// </summary>
+        /// <returns></returns>
+        private List<PluginFile> GetNetToolsPluginFiles()
+        {
+            var pluginFiles = new List<PluginFile>();
+
+            if (_environmentVariableReader is null)
+            {
+                _environmentVariableReader = EnvironmentVariableWrapper.Instance;
+            }
```
kartheekp-ms: I believe if we change the `_environmentVariableReader` variable to read-only, we can skip this `if` block. In the public constructor, we can add `_environmentVariableReader = EnvironmentVariableWrapper.Instance;`, while in the internal constructor added in this PR, it is passed as a parameter.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -140,6 +150,115 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+                if (Directory.Exists(path))
+                {
+                    var directoryInfo = new DirectoryInfo(path);
+                    var files = directoryInfo.GetFiles("nuget-plugin-*");
+
+                    foreach (var file in files)
+                    {
+                        if (IsValidPluginFile(file))
+                        {
+                            var state = new Lazy<PluginFileState>(() => _verifier.IsValid(file.FullName) ? PluginFileState.Valid : PluginFileState.InvalidEmbeddedSignature);
```
kartheekp-ms: .NET SDK recently enabled package signature verification for .NET tools. Hence, I think we can skip the `_verifier.IsValid` check because .NET tools are published as NuGet packages, which can be signed by the package author.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
zivkan: we should minimize constructors that don't have empty bodies, to reduce risk of bugs. Look at this one as an example: https://github.com/NuGet/NuGet.Client/blob/785d8bb948a8b477e9160f6732e7df50a1bba5d1/src/NuGet.Clients/NuGet.VisualStudio.Common/Services/NuGetFeatureFlagService.cs#L24-L36

The public constructor just calls the internal constructor with static values, and all the code to initialize the class is in that one internal that is testable with mocks. Importantly, if changes are needed to this class in the future, there's no question where changes to a constructor happen, because there's only one where "real" stuff is happening vs needing to choose from multiple constructors and then some code paths have an unassigned variable.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: Good catch

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: Is the suggestion to not check the validity of the package and just assume it is valid?

kartheekp-ms: Yes, that is correct.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -140,6 +152,119 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+        /// </summary>
+        /// <returns></returns>
+        private List<PluginFile> GetNetToolsPluginFiles()
+        {
+            var pluginFiles = new List<PluginFile>();
+
+            string[] paths = Array.Empty<string>();
+
+            // The path to the plugins installed using dotnet tools, should be specified in the NUGET_PLUGIN_PATHS environment variable.
+            paths = _environmentVariableReader.GetEnvironmentVariable("NUGET_PLUGIN_PATHS")?.Split(Path.PathSeparator) ?? Array.Empty<string>();
```
kartheekp-ms: We can use [EnvironmentVariableConstants.PluginPaths](https://github.com/NuGet/NuGet.Client/blob/fef6562e0ebdc57c904a08f732cc917fdb4d555e/src/NuGet.Core/NuGet.Protocol/Plugins/EnvironmentVariableConstants.cs#L12) for the environment variable name.

**EDIT:** Ignore this comment for now. I will add another comment related to this and delete this one.

kartheekp-ms: If the `NUGET_PLUGIN_PATHS` environment variable is already set, the `_rawpluginpaths` parameter passed to the constructor will contain a list of paths to plugin executables that are installed as .NET tools going forward.

https://github.com/NuGet/NuGet.Client/blob/fef6562e0ebdc57c904a08f732cc917fdb4d555e/src/NuGet.Core/NuGet.Protocol/Plugins/PluginManager.cs#L320-L323

I think we should remove the logic in `PluginManager` for discovering .NET tool plugins because they have different verification logic. For example, the file name should start with `nuget-plugin-*`, the executable bit should be set for non-Windows platforms, and the file should have a .exe or .bat extension on Windows.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -140,6 +152,119 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
             return files;
         }
 
+        /// <summary>
+        /// Gets auth plugins installed using dotnet tools. This is done by iterating through each file in directories found int eh
+        /// `PATH` environment variable.
+        /// The files are also expected to have a name that starts with `nuget-plugin-`
+        /// </summary>
+        /// <returns></returns>
+        private List<PluginFile> GetNetToolsPluginFiles()
```
kartheekp-ms: [Spec](https://github.com/NuGet/Home/blob/dev/accepted/2024/support-nuget-authentication-plugins-dotnet-tools.md#plugin-discovery) suggests the following.

- `NUGET_NETFX_PLUGIN_PATHS`: Defines the plugins used by the .NET Framework-based tooling (NuGet.exe/MSBuild.exe/Visual Studio). This takes precedence over `NUGET_PLUGIN_PATHS`.
- `NUGET_NETCORE_PLUGIN_PATHS`: Defines the plugins used by the .NET Core-based tooling (dotnet.exe).
This takes precedence over `NUGET_PLUGIN_PATHS`.
- `NUGET_PLUGIN_PATHS`: Going forward this environment variable defines the plugins installed as .NET tools.

If `NUGET_NETFX_PLUGIN_PATHS` or `NUGET_NETCORE_PLUGIN_PATHS` environment variables is set then I think we should no look for plugins installed as .NET tools.

```cs
if (string.IsNullOrEmpty(_rawPluginPaths))
{
 // load .NET plugins from either `NUGET_PLUGIN_PATHS` environment variable or from `PATH` environment variable scanning
}
```

As mentioned in another comment, `_rawPluginPaths` will not be null if any of the environment variables discussed above will be set.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -140,6 +152,119 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+            {
+#if NET8_0_OR_GREATER
+                var fileMode = File.GetUnixFileMode(fileInfo.FullName);
+
+                return fileInfo.Exists &&
+                    ((fileMode & UnixFileMode.UserExecute) != 0 ||
+                    (fileMode & UnixFileMode.GroupExecute) != 0 ||
+                    (fileMode & UnixFileMode.OtherExecute) != 0);
+#else
+                return fileInfo.Exists && (fileInfo.Attributes & FileAttributes.ReparsePoint) == 0 && IsExecutable(fileInfo);
```
kartheekp-ms: Please help me understand the reasoning behind the `(fileInfo.Attributes & FileAttributes.ReparsePoint == 0)` check. My internet search suggested it is related to symbolic links, etc., but I happy to learn more.

Nigusu-Allehu: You're right. I wanted to ensure the file both exists and is a real file (not a symbolic link) to minimize the risk of running malicious code. For example, imagine a symbolic link located in one of the directories listed in the user's `PATH` environment variable. The link has a name like `nuget-plugin-*`, making it appear like a normal plugin, but it actually points to a malicious file in a different directory. If we run that symlink, we unintentionally execute the malicious file from the other location. We could say it is up to the user to ensure there is no malicious files in the PATH folders. However, I don't see a reason why we should leave that vulnerability open.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: From my understanding, isValid checks whether the file has a valid embedded signature or not. Even though we do expect the files that come from .Net tools to have a valid signature. Isn't possible some file that was not installed by .Net tools exists in the path and it does not have a valid signature. Or it could be a .Net tools package that has been tempered with and does not have a valid signature anymore.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: I removed it and now the logic is
- If `NUGET_NETFX_PLUGIN_PATHS` or  `NUGET_NETCORE_PLUGIN_PATHS` are set, use them. 
- Otherwise if  `NUGET_PLUGIN_PATHS` is set, assume it is pointing to the directory that contains the dotnet tools installed plugins. If none of the env variables are set, go a head and scan for the plugins from the directories listed in PATH.

This limits `NUGET_PLUGIN_PATHS` to be used for .NET tools plugins only. Would that not be a problem for users who use it for non .NET tools plugins previously?

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -85,6 +96,13 @@ public async Task<IEnumerable<PluginDiscoveryResult>> DiscoverAsync(Cancellation
                 }
 
                 _pluginFiles = GetPluginFiles(cancellationToken);
+
+                if (string.IsNullOrEmpty(_rawPluginPaths))
+                {
+                    // Nuget plugin configuration environmental variables were not used to discover the configuration files
+                    _pluginFiles.AddRange(GetNetToolsPluginFiles());
```
zivkan: The XMLdoc for `rawPluginPaths` says:

> The raw semicolon-delimited list of supposed plugin file paths

But I don't understand when the value will be null or not null, or where the value comes from, and therefore I'm having a hard time understanding why we look for .NET tool plugins only when it's empty.

Is `rawPluginPaths` the value of the `NUGET_PLUGIN_PATHS` or `NUGET_NETFX_PLUGIN_PATHS` environment variables when they're set?

zivkan: Today, I've been trying to create a cred provider. I used the `nuget-plugin-*.exe` file convention, but setting `NUGET_NETFX_PLUGIN_PATHS` or `NUGET_PLUGIN_PATHS` to the bin directory of my project, or setting it to the full path of the exe itself, I couldn't get NuGet to use my plugin, without adding the project's bin directory to the path (and using your test branch for this feature that does execution, in addition to discovery). However, when doing this, it also discovered other cred providers I have installed, which was making debuggign my plugin much more difficult.

It can be done in a follow up, but I think there should be a way to disable normal plugin discovery, and only discover plugins I want. The `NUGET_PLUGIN_PATHS` environment variable appears to do that for the older discovery method, so there should be a way for this new one as well.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
zivkan: The "embedded signature" is an authenticode signature, which is not related to signed packages.

It typically costs hundreds of dollars a year to buy a code signing certificate, putting it out of reach for most non-corporate NuGet plugin developers.  Even when a developer does have a certificate available for the production code, it's typically not available during development and debugging. I hit this today while trying to develop a new plugin, and I had to hardcode `Valid` state just to be able to develop the plugin.

At the very least we should add an environment variable to allow skipping the authenticode checks, so that plugin developers can actually develop. However, I'm in favour of removing the authenticode checks altogether. If someone wants to allow only signed executables to run on a Windows machine, they can configure it via group policy.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
kartheekp-ms: > Is rawPluginPaths the value of the NUGET_PLUGIN_PATHS or NUGET_NETFX_PLUGIN_PATHS environment variables when they're set?

Yes Andy, `rawPluginPaths` is the value of the `NUGET_NETFX_PLUGIN_PATHS` or `NUGET_NETCORE_PLUGIN_PATHS` or `NUGET_PLUGIN_PATHS` environment variables when they are set. There is a precedence for these environment variables which is documented [here](https://learn.microsoft.com/nuget/reference/extensibility/nuget-cross-platform-plugins#plugin-installation-and-discovery).

src/NuGet.Core/NuGet.Protocol/NuGet.Protocol.csproj
nkolev92: Adding a framework to one of our assemblies is a significant change, and one that had both me & Kartheek wondering what it's for. 
It's something that should be a part of the PR description. 


I think there's some side-effects we need to work through, some documented in https://github.com/NuGet/NuGet.Client/blob/dev/docs/updating-target-frameworks.md:

1. The previous frameworks were net472, ns2.0, net5.0 and now net8.0. 
The test frameworks are net472;net8.0;netcoreapp3.1

Each test framework will be running against the most compatible NuGet.Protocol framework. 
net472 tests net472, net8.0 tests net8.0, netcoreapp3.1 tests ns2.0. 
This leaves net5.0 untested. 

2. There's also the source build considerations. This change may lead to some duplicates in the source build work that *may* break them. 

Let's explore what options we have and make the best decision. 
Some potential ideas we can discuss (there's probably more)
- Say we won't test net5.0 from NuGet.Protocol
- Bump all the frameworks from net5.0 to net8.0 (breaking change for our packages, but I it could simplify our testing)
- make NuGet.Protocol target net8.0 only instead of net5.0

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -140,6 +158,118 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
             return files;
         }
 
+        /// <summary>
+        /// Gets auth plugins installed using dotnet tools. This is done by iterating through each file in directories found int eh
+        /// `PATH` environment variable.
+        /// The files are also expected to have a name that starts with `nuget-plugin-`
+        /// </summary>
+        /// <returns></returns>
+        private List<PluginFile> GetNetToolsPluginFiles()
```
nkolev92: This is probably a method that can be tested independently of the rest.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -140,6 +158,118 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+        }
+
+        /// <summary>
+        /// Checks whether a file is a valid plugin file for windows/Unix.
+        /// Windows: It should be either .bat or  .exe
+        /// Unix: It should be executable
+        /// </summary>
+        /// <param name="fileInfo"></param>
+        /// <returns></returns>
+        private static bool IsValidPluginFile(FileInfo fileInfo)
```
nkolev92: Similar as the previous one, I think a bunch of these can be unit tested.

That way you're not testing all combinations in integration tests.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginManager.cs
```diff
@@ -317,11 +317,6 @@ private void Initialize(IEnvironmentVariableReader reader,
 #else
             _rawPluginPaths = reader.GetEnvironmentVariable(EnvironmentVariableConstants.CorePluginPaths);
 #endif
-            if (string.IsNullOrEmpty(_rawPluginPaths))
```
nkolev92, line 320: Can you help me understand why we're changing the location of this call from PluginManager to PluginDiscoverer?

src/NuGet.Core/NuGet.Protocol/Plugins/PluginFile.cs
```diff
@@ -20,6 +20,22 @@ public sealed class PluginFile
         /// </summary>
         public Lazy<PluginFileState> State { get; }
 
+        /// <summary>
+        /// Is the plugin file, a dotnet tools plugin file?
+        /// </summary>
+        public bool IsDotnetToolsPlugin { get; }
```
nkolev92: Is this class ever used outside of this assembly?
It seems to me that we never really needed it to the public. 

We can consider not making public changes.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginManager.cs
kartheekp-ms: This change was made as per https://github.com/NuGet/NuGet.Client/pull/5990#discussion_r1759318390 conversation.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
kartheekp-ms: Great question, Nigusu. https://github.com/NuGet/Home/pull/13242#discussion_r1548145738 conversation in the spec, led us to the decision of using the existing environment variable (NUGET_PLUGIN_PATHS) for .NET tools plugins. There are few references to this environment variable in other repositories such as JetBrains, OneGet to name a few https://github.com/search?q=NUGET_PLUGIN_PATHS&type=code. The idea is to allow customers to install .NET tools as tool path tools instead of global tools. Maybe we can add a breaking change label to the issue so that it can be part of the release notes.

nkolev92: I don't follow why we have to differentiate between dotnet tool and non dotnet tool with NUGET_PLUGIN_PATHS. 
Don't we just execute the with the full path of whatever was passed in there and how it starts it not a concern?

zivkan: On .NET (core) plugins using the old discovery method are executed via `dotnet {file}.dll`, whereas the new executables are executed directly, since they're executables.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
zivkan: perhaps `_overridePath` is clearer.

Plus XMLDoc comments that explains this, because even reading the docs, it's not super clear where the value for this `_rawPluginPaths` property comes from. Even knowing the override feature, it sounds to me like this value should never be null, and when the CLI arg is not provided, it should contain the default plugin path.

(also I'm lazy, so I don't want to read the docs, if the comments explain that the CLI arg allows the customer to override normal discovery, and discover only plugins in a specific directory, we won't need to context switch to the docs, and this is assuming the person reading the code even realises that there are relevant docs)

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
zivkan: It's great that you're thinking about security!  Let's talk about this offline.  I don't know how I feel about discussing security publicly.  Maybe it's fine, and a good learning resource for people, but maybe we don't want to give bad guys an insight into our security decisions.

src/NuGet.Core/NuGet.Protocol/NuGet.Protocol.csproj
Nigusu-Allehu: Can we not add net5.0 to the list of TFMs of the test projects? That way we don't break customers who are on net 5.0

test/NuGet.Core.Tests/NuGet.Protocol.Tests/Plugins/PluginDiscovererTests.cs
```diff
@@ -247,6 +249,446 @@ public async Task DiscoverAsync_IsIdempotent()
...
+
+                            Assert.False(discovered);
+                        }
+                    }
+                }
+            }
+        }
+
+        [PlatformFact(Platform.Windows)]
+        public void GetNetToolsPluginFiles_WithNuGetPluginPaths_ReturnsPluginsInNuGetPluginPathOnly()
```
Nigusu-Allehu: Test priority of the discovery methods we have. This test makes sure we prioritize the environmental variables.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: You are right, we should allow disabling the default discovery method. The goal of this PR is actually to do that as well. I think that part was probably missing when you looked at the PR.

This PR currently prioritizes `NUGET_NETFX_PLUGIN_PATHS ` or `NUGET_NETCORE_PLUGIN_PATHS ` `NUGET_PLUGIN_PATHS ` if they are set. Then, `NUGET_PLUGIN_PATHS ` follows. And if none of these are set, we use the default plugin searching method (going through all the paths in `PATH` env variable). I have added a unit test to assert this : https://github.com/NuGet/NuGet.Client/pull/5990/files#r1791018600 . Please let me know if you have suggestions.

Nigusu-Allehu: Our current implementation of plugins also follows this prioitization rule. As per @kartheekp-ms  and I's conversation offline, I have done sandbox testing to make sure that we discover plugins based on this priority 
1. `NUGET_NETFX_PLUGIN_PATHS ` or `NUGET_NETCORE_PLUGIN_PATHS ` 
2. `NUGET_PLUGIN_PATHS `
3. default

Nigusu-Allehu: I agree, I think the name _rawPluginPaths could be a bit misleading. How about _`envVariablePluginPaths`?

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: > Great question, Nigusu. [NuGet/Home#13242 (comment)](https://github.com/NuGet/Home/pull/13242#discussion_r1548145738) conversation in the spec, led us to the decision of using the existing environment variable (NUGET_PLUGIN_PATHS) for .NET tools plugins. There are few references to this environment variable in other repositories such as JetBrains, OneGet to name a few https://github.com/search?q=NUGET_PLUGIN_PATHS&type=code. The idea is to allow customers to install .NET tools as tool path tools instead of global tools. Maybe we can add a breaking change label to the issue so that it can be part of the release notes.

How about we support both instead of making it a breaking change. We could allow this environmental varibale to point to both .net tools plugins and plugin executables (.exe or .dll just like before). So the logic would be
* If the environment variable is point to an executable, treat it as a non .net tools plugin. This helps us avoid the breaking change
* If it is pointing to a folder, start searching for .net tools plugins in that folder.

src/NuGet.Core/NuGet.Protocol/NuGet.Protocol.csproj
nkolev92: We can do that too. 

We've been having a separate discussion with @zivkan, we have a lot of frameworks in our packages right now and that does make our support matrix larger. 

net5.0 as a runtime is out of support, so I'd expect that anyone having a runtime dependency on NuGet's libraries will likely be on a newer runtime. 

We can also say, if you really only need net5.0, use an older version of our packages. 

I think it'd be great to consolidate those conversations into 1.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
zivkan: If it points to an executable, and that executable filename starts with `nuget-plugin-`, then you could treat it as a .net tool plugin as well.

src/NuGet.Core/NuGet.Protocol/NuGet.Protocol.csproj
zivkan: PR is out:
* https://github.com/NuGet/NuGet.Client/pull/6086

Let's converse over there about this

Nigusu-Allehu: I will rebase after this is merged

nkolev92:
This is probably ready to rebase now.

Nigusu-Allehu:
> This is probably ready to rebase now.

Yes. I have rebased it if you are thinking about having net8.0 as the TFM

src/NuGet.Core/NuGet.Protocol/NuGet.Protocol.csproj
nkolev92: If you rebased, you probably don't need this anymore.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -85,6 +94,14 @@ public async Task<IEnumerable<PluginDiscoveryResult>> DiscoverAsync(Cancellation
                 }
 
                 _pluginFiles = GetPluginFiles(cancellationToken);
+
+                if (string.IsNullOrEmpty(_envVariablePluginPaths))
+                {
+                    // Nuget plugin configuration environmental variables: NUGET_NETFX_PLUGIN_PATHS, NUGET_NETCORE_PLUGIN_PATHS
+                    // were not used to point to the credential provider plugins.
+                    _pluginFiles.AddRange(GetPluginsInNuGetPluginPathsAndPath());
```
kartheekp-ms: This line of code will throw a null reference exception if `GetPluginFiles()` didn't return any plugins. Maybe we can initialize `_pluginFiles = new List<PluginFile>()` and then keep add plugins files as we discover them.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -140,9 +157,138 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+        /// <summary>
+        /// Retrieves authentication plugins by searching through directories specified in the `NuGET_PLUGIN_PATHS` and `PATH` 
+        /// environment variables. The method looks for files prefixed with 'nuget-plugin-' and verifies their validity for .net tools plugins.
+        /// </summary>
+        /// <returns>A list of valid <see cref="PluginFile"/> objects representing the discovered plugins.</returns>
+        internal List<PluginFile> GetPluginsInNuGetPluginPathsAndPath()
+        {
+            var pluginFiles = new List<PluginFile>();
+            var nugetPluginPaths = _environmentVariableReader.GetEnvironmentVariable(EnvironmentVariableConstants.PluginPaths);
+            string[] paths;
```
kartheekp-ms: nit: move the variable declaration to the line of code where it is initialized. I personally find it easier to follow the code.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -140,9 +157,138 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
             return files;
         }
 
+        /// <summary>
+        /// Retrieves authentication plugins by searching through directories specified in the `NuGET_PLUGIN_PATHS` and `PATH` 
+        /// environment variables. The method looks for files prefixed with 'nuget-plugin-' and verifies their validity for .net tools plugins.
+        /// </summary>
+        /// <returns>A list of valid <see cref="PluginFile"/> objects representing the discovered plugins.</returns>
+        internal List<PluginFile> GetPluginsInNuGetPluginPathsAndPath()
```
kartheekp-ms: This method fetches plugin files from either PLUGIN_PAHS environment variable OR from the PATH environment variable but not from both as the method name implies. I think we may have to refactor this method into two methods with each method having a specific responsibility. I am happy to brainstorm ideas.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -140,9 +157,138 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+                    if (!process.WaitForExit(1000) || process.ExitCode != 0)
+                    {
+                        return false;
+                    }
+
+                    output = process.StandardOutput.ReadToEnd().Trim();
+                }
+
+                // Check if the output is "yes"
+                return output.Equals("yes", StringComparison.OrdinalIgnoreCase);
```
kartheekp-ms: Will the output be always "yes" for all the different locale settings?

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -140,9 +157,138 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+                    {
+                        // A non DotNet tool plugin file
+                        var state = new Lazy<PluginFileState>(() => _verifier.IsValid(fileInfo.FullName) ? PluginFileState.Valid : PluginFileState.InvalidEmbeddedSignature);
+                        pluginFiles.Add(new PluginFile(fileInfo.FullName, state));
+                    }
+                }
+                else if (Directory.Exists(path))
+                {
+                    var directoryInfo = new DirectoryInfo(path);
+                    var files = directoryInfo.GetFiles("nuget-plugin-*");
```
kartheekp-ms: It looks like [`Directory.GetFiles`](https://learn.microsoft.com/dotnet/api/system.io.directory.getfiles?view=net-8.0) can throw several exceptions. I think the 2 important exceptions are [UnauthorizedAccessException](https://learn.microsoft.com/dotnet/api/system.unauthorizedaccessexception?view=net-8.0) and [PathTooLongException](https://learn.microsoft.com/dotnet/api/system.io.pathtoolongexception?view=net-8.0). These 2 exceptions are also caught explicitly in the doc that guides on [How to: Enumerate directories and files](https://learn.microsoft.com/en-us/dotnet/standard/io/how-to-enumerate-directories-and-files). In our case, we should at least log to notify users that a plugin has been ignored due to an exception, at least when [plugin diagnostic logging](https://github.com/NuGet/NuGet.Client/blob/dev/docs/plugin-logging.md) is enabled.

kartheekp-ms: > In our case, we should at least log to notify users that a plugin has been ignored due to an exception, at least when [plugin diagnostic logging](https://github.com/NuGet/NuGet.Client/blob/dev/docs/plugin-logging.md) is enabled.

I am not sure if a logger is passed to this API. If there is no access to the logger, I am okay if you would like to address this part of the feedback in a follow-up PR. It would be great if the feedback related to exceptions is addressed in this PR.

Nigusu-Allehu: As you mentioned there is not loggers passed to PluginDiscoverer. This is also reflected in the current implementation : https://github.com/NuGet/NuGet.Client/blob/428d2d0013971187ac5f4ce4fc6402f191cf28e0/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoveryUtility.cs#L84
This is a similar operation that could result in an exception being thrown. I think we should definitely introduce proper exception handling and logging in PluginDiscoverer. However, that could be done as a separate effort. 

We could use https://github.com/NuGet/NuGet.Client/blob/428d2d0013971187ac5f4ce4fc6402f191cf28e0/src/NuGet.Core/NuGet.Protocol/Plugins/Logging/PluginLogger.cs#L71 and log the exceptions. However, the purpose of that logger is to log plugin messages and updates using a file stream.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: the bash code being run is `if [ -x {fileInfo.FullName} ]; then echo yes; else echo no; fi` .it outputs yes if the file is executable.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: After some digging I was able to find out that we do catch the exception at a higher level: https://github.com/NuGet/NuGet.Client/blob/428d2d0013971187ac5f4ce4fc6402f191cf28e0/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpSourceAuthenticationHandler.cs#L291.
We also do logging on it. From my debug executions, this logging was not happening because the logger passed down here is a `NullLogger`.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginManager.cs
```diff
@@ -119,7 +119,7 @@ public async Task<IEnumerable<PluginCreationResult>> CreatePluginsAsync(
             var pluginCreationResults = new List<PluginCreationResult>();
 
             // Fast path
-            if (source.PackageSource.IsHttp && IsPluginPossiblyAvailable())
+            if (source.PackageSource.IsHttp)
```
nkolev92: Doesn't this lead to us fetching the ServinceIndexResourceV3 even where there are no plugins available? 
Digging more, I'm wondering if we'd ever try to initialize a download plugin unless it's specified with an env var (which is probably a good thing tbh).

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -17,25 +19,35 @@ public sealed class PluginDiscoverer : IPluginDiscoverer
     {
         private bool _isDisposed;
         private List<PluginFile> _pluginFiles;
-        private readonly string _rawPluginPaths;
+        private readonly string _envVariableNetCoreAndNetFXPluginPaths;
```
nkolev92: I was wondering if there are better descriptive names that aren't as wordy
 Some ideas, _envVariableNetCoreAndNetFXPluginPaths could be `envVariablePlatformSpecificPluginPaths`

src/NuGet.Core/NuGet.Protocol/Plugins/PluginManager.cs
nkolev92: This logic may need reworked to ensure we're not fetching resources or starting plugins too aggressively.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -17,25 +19,35 @@ public sealed class PluginDiscoverer : IPluginDiscoverer
...
         {
             if (verifier == null)
             {
                 throw new ArgumentNullException(nameof(verifier));
             }
 
-            _rawPluginPaths = rawPluginPaths;
+            _environmentVariableReader = environmentVariableReader;
+            _envVariableNetCoreAndNetFXPluginPaths = rawPluginPaths;
+            _envVariableNuGetPluginPaths = _environmentVariableReader.GetEnvironmentVariable(EnvironmentVariableConstants.PluginPaths);
```
nkolev92: I think we moved every logic out of the PluginManager that uses the env vars, so we probably shouldn't even be fetching the env vars from there. 
We can probably make the plugin discoverer take an IEnvironmentVariableReader and call it good. 
Especially when combined with the other change that's coming in from Andy that already breaks the public API.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -84,7 +96,33 @@ public async Task<IEnumerable<PluginDiscoveryResult>> DiscoverAsync(Cancellation
...
+                    var directories = new List<string> { PluginDiscoveryUtility.GetNuGetHomePluginsPath() };
+#if IS_DESKTOP
+                    // Internal plugins are only supported for .NET Framework scenarios, namely msbuild.exe
+                    directories.Add(PluginDiscoveryUtility.GetInternalPlugins());
+#endif
+                    var filePaths = PluginDiscoveryUtility.GetConventionBasedPlugins(directories);
+                    _pluginFiles = GetPluginFiles(filePaths.ToArray(), cancellationToken);
+
+                    // Search for .Net tools plugins in PATH
+                    _pluginFiles.AddRange(GetPluginsInPATH() ?? new List<PluginFile>());
```
nkolev92: If _pluginFiles was already null you'd get an NRE on AddRange.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -140,19 +176,171 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+                    if (Directory.Exists(path))
+                    {
+                        var directoryInfo = new DirectoryInfo(path);
+                        var files = directoryInfo.GetFiles("nuget-plugin-*");
+
+                        foreach (var file in files)
+                        {
+                            if (IsValidPluginFile(file))
+                            {
+                                // .NET SDK recently package signature verification for .NET tools, as a result we expect them to be valid.
```
nkolev92: Those 2 concepts are not related.
This used to check the assembly authenticode signature, not the package.

nkolev92 (commented):
Thanks for the adding the tests!
They look very thorough. 

In the PR checklist you have: 

image


Please create an actual tracking issue for documentation, and link it here.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: The comment is for `PluginFileState.Valid` below. So that it does not look like we are using a magic value.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: I was avoiding breaking the API, but it looks like it would be a cleaner implementation if we retrieve the env variables for plugins in PluginDiscoverer than PluginManager.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
nkolev92: My feedback is that the comment is not quite right.

The comment is talking about the package itself, while this plugin validation is supposed to be for the runnable file. 

I think we can just remove the comment. Andy recently removed the authenticode stuff.

nkolev92:
Wanna update your feature branch and rebase this PR now that https://github.com/NuGet/NuGet.Client/commit/1a8ca6067347fd68340090e29464ef58a0c9ef62 has been merged?

nkolev92 (approved):

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
kartheekp-ms: > it outputs yes if the file is executable.

Thank you. Just confirming that the output will be "yes" even on platforms that use non-English locale after executing the bash code you provided in the above comment.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -17,17 +19,26 @@ public sealed class PluginDiscoverer : IPluginDiscoverer
     {
         private bool _isDisposed;
         private List<PluginFile> _pluginFiles;
-        private readonly string _rawPluginPaths;
+        private readonly string _netCoreAndNetFXPluginPaths;
```
kartheekp-ms: This variable holds either net core or netfx plugins. I am thinking the variable name can be `_netCoreOrNetFXPluginPaths`. I am curious if you have tried VS smart rename AI feature and see if it suggests any alternatives :)

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -17,17 +19,26 @@ public sealed class PluginDiscoverer : IPluginDiscoverer
...
+        internal PluginDiscoverer(IEnvironmentVariableReader environmentVariableReader)
         {
-            _rawPluginPaths = rawPluginPaths;
+            _environmentVariableReader = environmentVariableReader;
+#if IS_DESKTOP
+            _netCoreAndNetFXPluginPaths = environmentVariableReader.GetEnvironmentVariable(EnvironmentVariableConstants.DesktopPluginPaths);
+#else
+            _netCoreAndNetFXPluginPaths = environmentVariableReader.GetEnvironmentVariable(EnvironmentVariableConstants.CorePluginPaths);
+#endif
+            _nuGetPluginPaths = _environmentVariableReader.GetEnvironmentVariable(EnvironmentVariableConstants.PluginPaths);
```
kartheekp-ms: The current behavior in dev branch is that, if customers have configured net core or netfx plugins then we do not look for PluginPaths environment variable. We are changing that behavior in this feature. Is this intentional? 

> The discovery is done based on which environmental variable has been specified

> - If NUGET_NETFX_PLUGIN_PATHS or NUGET_NETFX_PLUGIN_PATHS have been specified, use the old discovery method to look up non dotnet tools plugins in the paths added in the env variable.
> - If NUGET_NETFX_PLUGIN_PATHS and NUGET_NETFX_PLUGIN_PATHS have not been specified
>   - if NUGET_PLUGIN_PATHS specified, Collect the set of plugins the env variable is pointing to: They could be donet tools or non dotnet tools plugin files.
>   - if NUGET_PLUGIN_PATHS not specified, discover all plugins using the default discovery method

I think there is a typo in the above PR description. I fixed the typos below.

> The discovery is done based on which environmental variable has been specified

> - If NUGET_NETFX_PLUGIN_PATHS or NUGET_NETCORE_PLUGIN_PATHS have been specified, use the old discovery method to look up non dotnet tools plugins in the paths added in the env variable.
> - If NUGET_NETFX_PLUGIN_PATHS and NUGET_NETCORE_PLUGIN_PATHS have not been specified
>   - if NUGET_PLUGIN_PATHS specified, Collect the set of plugins the env variable is pointing to: They could be dotnet tools or non dotnet tools plugin files.
>   - if NUGET_PLUGIN_PATHS not specified, discover all plugins using the default discovery method

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -75,7 +86,40 @@ public async Task<IEnumerable<PluginDiscoveryResult>> DiscoverAsync(Cancellation
...
+                {
+                    // restore to default plugins search.
+                    // Search for plugins in %user%/.nuget/plugins
+                    var directories = new List<string> { PluginDiscoveryUtility.GetNuGetHomePluginsPath() };
+#if IS_DESKTOP
+                    // Internal plugins are only supported for .NET Framework scenarios, namely msbuild.exe
+                    directories.Add(PluginDiscoveryUtility.GetInternalPlugins());
+#endif
+                    var filePaths = PluginDiscoveryUtility.GetConventionBasedPlugins(directories);
+                    _pluginFiles = GetPluginFiles(filePaths.ToArray(), cancellationToken);
```
kartheekp-ms: I am curious on why we invoked `.ToArray()` method on the `filePaths`. We can directly pass the enumerable or whatever is the return value of the `GetConventionBasedPlugins` method to the private static method.

Also, if `filePaths` is null for any reason then `.ToArray()` invocation will throw a null reference exception.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -75,7 +86,40 @@ public async Task<IEnumerable<PluginDiscoveryResult>> DiscoverAsync(Cancellation
                     return _results;
                 }
 
-                _pluginFiles = GetPluginFiles(cancellationToken);
+                if (!string.IsNullOrEmpty(_netCoreAndNetFXPluginPaths))
+                {
+                    // NUGET_NETFX_PLUGIN_PATHS, NUGET_NETCORE_PLUGIN_PATHS have been set.
+                    var filePaths = _netCoreAndNetFXPluginPaths.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
```
kartheekp-ms: Maybe we could use `Path.PathSeparator` instead of hardcoding semicolon character. I learnt about that API in this PR because you used it to split plugins from PATH environment variable below. Thank you.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -124,19 +166,169 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+            var pluginFiles = new List<PluginFile>();
+            string[] paths = _nuGetPluginPaths?.Split(Path.PathSeparator) ?? Array.Empty<string>();
+
+            foreach (var path in paths)
             {
-                var directories = new List<string> { PluginDiscoveryUtility.GetNuGetHomePluginsPath() };
-#if IS_DESKTOP
-                // Internal plugins are only supported for .NET Framework scenarios, namely msbuild.exe
-                directories.Add(PluginDiscoveryUtility.GetInternalPlugins());
+                if (PathValidator.IsValidLocalPath(path) || PathValidator.IsValidUncPath(path))
```
kartheekp-ms, line 190: Does the existing code in dev branch support UNC paths for plugins?

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -124,19 +166,169 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+                {
+                    if (File.Exists(path))
+                    {
+                        FileInfo fileInfo = new FileInfo(path);
+                        if (fileInfo.Name.StartsWith("nuget-plugin-", StringComparison.CurrentCultureIgnoreCase))
+                        {
+                            // A DotNet tool plugin
+                            if (IsValidPluginFile(fileInfo))
+                            {
+                                PluginFile pluginFile = new PluginFile(fileInfo.FullName, new Lazy<PluginFileState>(() => PluginFileState.Valid), isDotnetToolsPlugin: true);
```
kartheekp-ms, line 200: I think we may have used `Lazy` for plugin state previously because we used to perform Authenticode verification on plugins on Windows. 

Given https://github.com/NuGet/NuGet.Client/pull/6042 PR has been merged, I think we don't the `Lazy` initialization anymore. We can address this comment in the follow-up PR. Please create a tracking issue.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -124,19 +166,169 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
             {
-                var directories = new List<string> { PluginDiscoveryUtility.GetNuGetHomePluginsPath() };
-#if IS_DESKTOP
-                // Internal plugins are only supported for .NET Framework scenarios, namely msbuild.exe
-                directories.Add(PluginDiscoveryUtility.GetInternalPlugins());
+                if (PathValidator.IsValidLocalPath(path) || PathValidator.IsValidUncPath(path))
+                {
+                    if (File.Exists(path))
+                    {
+                        FileInfo fileInfo = new FileInfo(path);
```
kartheekp-ms, line 194: AFAIK, we have to create an instance of `FileInfo` only when the plugin is a .NET tool and for traditional plugins, we don't need to do that. I am wondering if we could create FileInfo object only when the path name starts with `nuget-plugin-`.


```cs
if (path.StartsWith("nuget-plugin-", StringComparison.CurrentCultureIgnoreCase)) 
{
 FileInfo fileInfo = new FileInfo(path);

 if (IsValidPluginFile(fileInfo)) 
  {
    //existing code
  }
}
```

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -124,19 +166,169 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+                        var files = directoryInfo.GetFiles("nuget-plugin-*");
+
+                        foreach (var file in files)
+                        {
+                            if (IsValidPluginFile(file))
+                            {
+                                PluginFile pluginFile = new PluginFile(file.FullName, new Lazy<PluginFileState>(() => PluginFileState.Valid), isDotnetToolsPlugin: true);
+                                pluginFiles.Add(pluginFile);
+                            }
+                        }
```
kartheekp-ms: I noticed same logic in `GetPluginsInNuGetPluginPaths` method. May there is some opportunity to refactor this logic into a separate method and invoke it from both the places.

https://github.com/NuGet/NuGet.Client/pull/5990/files#diff-cf421e9b689e2c0e46c47e27365571384c9fd9c7eae6522a36c26f5bd40312a4R204-R214

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -124,19 +166,169 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+        {
+#pragma warning disable CA1031 // Do not catch general exception types
+            try
+            {
+                string output;
+                using (var process = new System.Diagnostics.Process())
+                {
+                    // Use a shell command to check if the file is executable
+                    process.StartInfo.FileName = "/bin/bash";
+                    process.StartInfo.Arguments = $" -c \"if [ -x {fileInfo.FullName} ]; then echo yes; else echo no; fi\"";
```
kartheekp-ms: If the file name has special character such as space, I guess the command will throw an error. Maybe we should enclose the file path in single (') or double (") quotes to prevent the shell from interpreting special characters. Also, we can add tests to ensure that special characters in the path are escaped.

kartheekp-ms (commented):
Great progress. Left few comments on the new changes proposed.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -124,19 +166,169 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+                    process.StartInfo.UseShellExecute = false;
+                    process.StartInfo.RedirectStandardOutput = true;
+
+                    process.Start();
+                    if (!process.WaitForExit(1000) || process.ExitCode != 0)
+                    {
+                        return false;
+                    }
+
+                    output = process.StandardOutput.ReadToEnd().Trim();
```
kartheekp-ms: [Docs](https://learn.microsoft.com/dotnet/api/system.diagnostics.process.standardoutput?view=net-7.0) suggest that to avoid deadlocks, always read the output stream first and then wait for exit. This may not apply for this case, but I am not sure if there is any edge case where child process too much info to the buffer and waits for parent to read the log before the process can actually exit resulting in deadlock.

```cs
// To avoid deadlocks, always read the output stream first and then wait.  
      string output = p.StandardOutput.ReadToEnd();  
      p.WaitForExit();
```

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
kartheekp-ms: Another question, does this script work even if the file is actually being read by another dotnet nuget command process at the same time?

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -124,19 +166,169 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+                    // Use a shell command to check if the file is executable
+                    process.StartInfo.FileName = "/bin/bash";
+                    process.StartInfo.Arguments = $" -c \"if [ -x {fileInfo.FullName} ]; then echo yes; else echo no; fi\"";
+                    process.StartInfo.UseShellExecute = false;
+                    process.StartInfo.RedirectStandardOutput = true;
+
+                    process.Start();
+                    if (!process.WaitForExit(1000) || process.ExitCode != 0)
+                    {
+                        return false;
```
kartheekp-ms, line 332: I think we have to invoke [`Process.Kill()`](https://learn.microsoft.com/dotnet/api/system.diagnostics.process.kill?view=net-8.0) in this case before returning `false`. AFAIK `Process.Dispose()` does release only unmanaged resources but doesn't kill the underlying process.

Docs state that,

> Whenever you use [Start](https://learn.microsoft.com/dotnet/api/system.diagnostics.process.start?view=net-8.0) to start a process, you might need to close it or you risk losing system resources. Close processes using [CloseMainWindow](https://learn.microsoft.com/dotnet/api/system.diagnostics.process.closemainwindow?view=net-8.0) or [Kill](https://learn.microsoft.com/dotnet/api/system.diagnostics.process.kill?view=net-8.0). You can check whether a process has already been closed by using its [HasExited](https://learn.microsoft.com/dotnet/api/system.diagnostics.process.hasexited?view=net-8.0) property.

There is an API in plugins code to kill the process. I added a link to it for your reference.

https://github.com/NuGet/NuGet.Client/blob/b8efbf37ba3386e377bc4f6b0febd526f51895d9/src/NuGet.Core/NuGet.Protocol/Plugins/PluginProcess.cs#L126-L129

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: I agree the logic should be if customers have configured net core or netfx plugin env variables, use them to find plugins and not PluginPaths. What made you think the current implementation does change that behavior?

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: https://github.com/NuGet/NuGet.Client/blob/8694ecdd055f0d0453974b0b9d0ad18926996679/src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs#L112 We do validate if its a valid UNC path

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -75,7 +86,40 @@ public async Task<IEnumerable<PluginDiscoveryResult>> DiscoverAsync(Cancellation
...
+                    // Internal plugins are only supported for .NET Framework scenarios, namely msbuild.exe
+                    directories.Add(PluginDiscoveryUtility.GetInternalPlugins());
+#endif
+                    var filePaths = PluginDiscoveryUtility.GetConventionBasedPlugins(directories);
+                    _pluginFiles = GetPluginFiles(filePaths.ToArray(), cancellationToken);
+
+                    // Search for .Net tools plugins in PATH
+                    if (_pluginFiles != null)
+                    {
+                        _pluginFiles.AddRange(GetPluginsInPATH() ?? new List<PluginFile>());
```
jgonz120: GetPluginsInPATH looks like it always returns a list, is the null check necessary?

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
zivkan: `_runtimeSpecificPlatformPaths`?

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -124,19 +170,169 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+            }
+
+            return pluginFiles;
+        }
+
+        /// <summary>
+        /// Retrieves .NET tools authentication plugins by searching through directories specified in `PATH` 
+        /// </summary>
+        /// <returns>A list of valid <see cref="PluginFile"/> objects representing the discovered plugins.</returns>
+        internal List<PluginFile> GetPluginsInPATH()
```
jgonz120: nit
```suggestion
        internal List<PluginFile> GetPluginsInPath()
```

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: But does path start with "nuget-plugin-"? I think only the last token of the path which is the file name will contain "nuget-plugin-". We could do some string parsing to search for the name of the file maybe

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: I have renamed it to `_netCoreOrNetFXPluginPaths` but `_runtimeSpecificPlatformPaths` is also a good idea. But I don't see a problem with `_netCoreOrNetFXPluginPaths`

jgonz120 (approved):
This looks fine to me. There were a couple of instances where you're checking for null results from private/internal methods which don't return null.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: Yes, the output should always be either "yes" or "no". I can't imagine a scenario where it is not. They are literals being printed.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: Great catch! I have surrounded it with single quotes.

Nigusu-Allehu: I think that should be ok. This does not need to open the file. I was able to have a file open and execute `[ -x 'filename' ]` on it at the same time on my linux device.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: So we shoudl make https://github.com/NuGet/NuGet.Client/blob/8694ecdd055f0d0453974b0b9d0ad18926996679/src/NuGet.Core/NuGet.Protocol/Plugins/PluginFile.cs#L21 not `Lazy`. I will create a tracking issue for that

Nigusu-Allehu: https://github.com/NuGet/Home/issues/13865

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
kartheekp-ms: > What made you think the current implementation does change that behavior?

The current logic reads the `EnvironmentVariableConstants.PluginPaths` environment variable, regardless of whether the NETCORE or NETFX plugin environment variables are set.

As far as I know, in the development branch, if either the NETCORE or NETFX plugin environment variables are set, the PluginPaths environment variable is not used.

This should be fine because we are only retrieving the PluginPaths environment variable. However, the discovery logic ignores it if the NETCORE or NETFX environment variables are already set.

kartheekp-ms: > I think there is a typo in the above PR description. I fixed the typos below.

I have updated the PR description fixing the typo.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
kartheekp-ms: I missed the point that `path` variable contains full path not the file name only as I originally thought. I think we are good here.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -124,19 +170,169 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+                    process.StartInfo.FileName = "/bin/bash";
+                    process.StartInfo.Arguments = $" -c \"if [ -x '{fileInfo.FullName}' ]; then echo yes; else echo no; fi\"";
+                    process.StartInfo.UseShellExecute = false;
+                    process.StartInfo.RedirectStandardOutput = true;
+
+                    process.Start();
+                    output = process.StandardOutput.ReadToEnd().Trim();
 
-            return _rawPluginPaths.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
+                    if (!process.WaitForExit(1000) || process.ExitCode != 0)
```
kartheekp-ms, line 320:
```suggestion
                    if (!_process.HasExited && !_process.WaitForExit(milliseconds: 1000))
```

test/NuGet.Core.Tests/NuGet.Protocol.Tests/Plugins/PluginDiscovererTests.cs
```diff
@@ -20,15 +23,21 @@ public class PluginDiscovererTests
         [InlineData(" ")]
         public void Constructor_AcceptsAnyString(string rawPluginPaths)
         {
-            using (new PluginDiscoverer(rawPluginPaths))
+            var environmentalVariableReader = new Mock<IEnvironmentVariableReader>();
+            environmentalVariableReader.Setup(env => env.GetEnvironmentVariable(EnvironmentVariableConstants.DesktopPluginPaths)).Returns(rawPluginPaths);
+            environmentalVariableReader.Setup(env => env.GetEnvironmentVariable(EnvironmentVariableConstants.CorePluginPaths)).Returns(rawPluginPaths);
+            using (new PluginDiscoverer())
```
kartheekp-ms: Copilot suggested a better way to check if constructor raises an exception or not. The current implementation bubbles up the exception and causes the entire test suite to stop execution.

```cs
Exception exception = Record.Exception(() =>
    {
        using (new PluginDiscoverer())
        {
        }
    });

    Assert.Null(exception); // Assert that no exception was thrown
```

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: similar issue found: https://github.com/NuGet/Home/issues/12068

test/NuGet.Core.Tests/NuGet.Protocol.Tests/Plugins/PluginDiscovererTests.cs
```diff
@@ -153,5 +176,544 @@ public async Task DiscoverAsync_IsIdempotent()
...
+        public void IsExecutable_FileIsNotExecutable_ReturnsFalse()
+        {
+            // Arrange
+            using TestDirectory testDirectory = TestDirectory.Create();
+            var workingPath = testDirectory.Path;
+            var pluginFilePath = Path.Combine(workingPath, "plugin");
+            File.Create(pluginFilePath);
+
+            // Remove execute permissions
+            var process = new Process();
```
kartheekp-ms, line 673: nit: There is an opportunity to refactor here because we set/remove execute permissions for files in all the new tests. I think we can create a private utility method and invoke it from the tests as needed. Please feel free to address it a follow-up issue.

kartheekp-ms (approved):

test/NuGet.Core.Tests/NuGet.Protocol.Tests/Plugins/PluginDiscovererTests.cs
```diff
@@ -153,5 +176,544 @@ public async Task DiscoverAsync_IsIdempotent()
...
+            process.StartInfo.FileName = "/bin/bash";
+            process.StartInfo.Arguments = $"-c \"chmod -x {pluginFilePath}\"";
+            process.StartInfo.UseShellExecute = false;
+            process.StartInfo.RedirectStandardOutput = true;
+            process.Start();
+            process.WaitForExit();
```
kartheekp-ms, line 679: Given that `Process.WaitForExit` waits indefinitely for the process to exit, while addressing [this comment](https://github.com/NuGet/NuGet.Client/pull/5990/files#r1811192719) you might want to consider this code by changing it to the following:

```cs
Assert.True(process.WaitForExit(1000), userMessage: "some string that helps to troubleshoot the issue");
```

With this approach, the test fails gracefully if the process doesn't complete on time for any unknown reason.

PR #5991, opened 2024-08-25
Add support for package type queries in both local and remote sources

### LINKED ISSUE #8915: Feature ask client side -- Surface PackageType query parameter in the NuGet.org search API
https://github.com/NuGet/NuGetGallery/issues/7594

SDK team has the nuget gallery ask. And Peter Yu and @joelverhagen suggested it is better have a client side counterpart instead of directly query the web API. So all client side .NET code is in nuget client.

However, in the meanwhile SDK will continue directly using the web REST API since it is relatively simple, and we will migrate to use the client once it is available.

glopesdev:
# Bug

Fixes: https://github.com/NuGet/Home/issues/8915

## Description

nuget.org allows filtering by package type using the `packageType` query parameter since [`SearchQueryService/3.5.0`](https://learn.microsoft.com/en-us/nuget/api/search-query-service-resource#searchqueryservice350). This parameter has not been officially surfaced by the client API, although relevant infrastructure has already been introduced in the `SearchFilter` class:

https://github.com/NuGet/NuGet.Client/blob/c8d14f3c28e3af3bdc3dcd38b23a212a91d13234/src/NuGet.Core/NuGet.Protocol/SearchFilter.cs#L60

Upon inspection of relevant source code, it looks like the relevant bits had already been introduced in `PackageSearchResourceV3` but using the old query parameter name `packageTypeFilter`. The fix is simply replacing the below query parameter with `packageType`:

https://github.com/NuGet/NuGet.Client/blob/c8d14f3c28e3af3bdc3dcd38b23a212a91d13234/src/NuGet.Core/NuGet.Protocol/Resources/PackageSearchResourceV3.cs#L140-L147

For completeness, we also add package type filtering functionality to `LocalPackageSearchResource` by filtering over the package types available through `package.Nuspec.GetPackageTypes()`.

## PR Checklist

- [ ] Added tests

### DIFF
--- a/src/NuGet.Core/NuGet.Protocol/LocalRepositories/LocalPackageSearchResource.cs
+++ b/src/NuGet.Core/NuGet.Protocol/LocalRepositories/LocalPackageSearchResource.cs
@@ -59,6 +59,21 @@ namespace NuGet.Protocol
                     query = query.Where(package => ContainsAnyTerm(terms, package));
                 }
 
+                // Filter on package types
+                if (filters?.PackageTypes != null
+                    && filters.PackageTypes.Any())
+                {
+                    foreach (var packageTypeName in filters.PackageTypes)
+                    {
+                        query = query
+                            .Where(package => package.Nuspec
+                                .GetPackageTypes()
+                                .Any(packageType => StringComparer.OrdinalIgnoreCase.Equals(
+                                    packageType.Name,
+                                    packageTypeName)));
+                    }
+                }
+
                 // Collapse to the highest version per id, if necessary
                 var collapsedQuery = filters?.Filter == SearchFilterType.IsLatestVersion ||
                                      filters?.Filter == SearchFilterType.IsAbsoluteLatestVersion
--- a/src/NuGet.Core/NuGet.Protocol/LocalRepositories/LocalPackageSearchResource.cs
+++ b/src/NuGet.Core/NuGet.Protocol/LocalRepositories/LocalPackageSearchResource.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using NuGet.Common;
+using NuGet.Packaging.Core;
 using NuGet.Protocol.Core.Types;
 using NuGet.Versioning;
 
@@ -62,12 +63,7 @@ namespace NuGet.Protocol
                 // Filter on package types
                 if (!string.IsNullOrEmpty(filters.PackageType))
                 {
-                    query = query
-                        .Where(package => package.Nuspec
-                            .GetPackageTypes()
-                            .Any(packageType => StringComparer.OrdinalIgnoreCase.Equals(
-                                packageType.Name,
-                                filters.PackageType)));
+                    query = query.Where(package => MatchesPackageType(package, filters.PackageType));
                 }
 
                 // Collapse to the highest version per id, if necessary
@@ -86,6 +82,23 @@ namespace NuGet.Protocol
                     .Select(package => CreatePackageSearchResult(package, filters, log, token))
                     .ToArray();
             }, token, TaskCreationOptions.LongRunning, TaskScheduler.Default);
+        }
+
+        /// <summary>
+        /// Search package types to match specified filter. If no package types are provided,
+        /// accept package as long as the specified filter type is Dependency.
+        /// </summary>
+        private static bool MatchesPackageType(LocalPackageInfo package, string packageTypeName)
+        {
+            var packageTypes = package.Nuspec.GetPackageTypes();
+            if (packageTypes.Count == 0
+                && PackageType.PackageTypeNameComparer.Equals(packageTypeName, PackageType.Dependency.Name))
+            {
+                return true;
+            }
+
+            return packageTypes
+                .Any(packageType => PackageType.PackageTypeNameComparer.Equals(packageType.Name, packageTypeName));
         }
 
         /// <summary>
--- a/src/NuGet.Core/NuGet.Protocol/LocalRepositories/LocalPackageSearchResource.cs
+++ b/src/NuGet.Core/NuGet.Protocol/LocalRepositories/LocalPackageSearchResource.cs
@@ -60,18 +60,14 @@ namespace NuGet.Protocol
                 }
 
                 // Filter on package types
-                if (filters?.PackageTypes != null
-                    && filters.PackageTypes.Any())
+                if (!string.IsNullOrEmpty(filters.PackageType))
                 {
-                    foreach (var packageTypeName in filters.PackageTypes)
-                    {
-                        query = query
-                            .Where(package => package.Nuspec
-                                .GetPackageTypes()
-                                .Any(packageType => StringComparer.OrdinalIgnoreCase.Equals(
-                                    packageType.Name,
-                                    packageTypeName)));
-                    }
+                    query = query
+                        .Where(package => package.Nuspec
+                            .GetPackageTypes()
+                            .Any(packageType => StringComparer.OrdinalIgnoreCase.Equals(
+                                packageType.Name,
+                                filters.PackageType)));
                 }
 
                 // Collapse to the highest version per id, if necessary
--- a/src/NuGet.Core/NuGet.Protocol/Resources/PackageSearchResourceV3.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Resources/PackageSearchResourceV3.cs
@@ -142,7 +142,7 @@ namespace NuGet.Protocol
                 {
                     var types = string.Join("&",
                         filters.PackageTypes.Select(
-                            s => "packageTypeFilter=" + s));
+                            s => "packageType=" + s));
                     queryString += "&" + types;
                 }
 
--- a/src/NuGet.Core/NuGet.Protocol/Resources/PackageSearchResourceV3.cs
+++ b/src/NuGet.Core/NuGet.Protocol/Resources/PackageSearchResourceV3.cs
@@ -137,13 +137,9 @@ namespace NuGet.Protocol
                     queryString += "&" + frameworks;
                 }
 
-                if (filters.PackageTypes != null
-                    && filters.PackageTypes.Any())
+                if (!string.IsNullOrEmpty(filters.PackageType))
                 {
-                    var types = string.Join("&",
-                        filters.PackageTypes.Select(
-                            s => "packageType=" + s));
-                    queryString += "&" + types;
+                    queryString += "&packageType=" + filters.PackageType;
                 }
 
                 queryString += "&semVerLevel=2.0.0";
--- a/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/Formatters/SearchFilterFormatter.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/Formatters/SearchFilterFormatter.cs
@@ -12,7 +12,7 @@ namespace NuGet.VisualStudio.Internal.Contracts
     {
         private const string IncludePrereleasePropertyName = "includeprerelease";
         private const string IncludeDelistedPropertyName = "includedelisted";
-        private const string PackageTypesPropertyName = "packagetypes";
+        private const string PackageTypePropertyName = "packagetype";
         private const string FilterPropertyName = "filter";
         private const string OrderByPropertyName = "orderby";
         private const string SupportedFrameworksPropertyName = "supportedframeworks";
@@ -30,7 +30,7 @@ namespace NuGet.VisualStudio.Internal.Contracts
             SearchFilterType? filterType = null;
             SearchOrderBy? searchOrderBy = null;
             IEnumerable<string>? supportedFrameworks = null;
-            IEnumerable<string>? packageTypes = null;
+            string? packageType = null;
 
             int propertyCount = reader.ReadMapHeader();
             for (int propertyIndex = 0; propertyIndex < propertyCount; propertyIndex++)
@@ -43,8 +43,8 @@ namespace NuGet.VisualStudio.Internal.Contracts
                     case IncludeDelistedPropertyName:
                         includeDelisted = reader.ReadBoolean();
                         break;
-                    case PackageTypesPropertyName:
-                        packageTypes = options.Resolver.GetFormatter<IEnumerable<string>>()!.Deserialize(ref reader, options);
+                    case PackageTypePropertyName:
+                        packageType = reader.ReadString();
                         break;
                     case FilterPropertyName:
                         filterType = options.Resolver.GetFormatter<SearchFilterType?>()!.Deserialize(ref reader, options);
@@ -65,7 +65,7 @@ namespace NuGet.VisualStudio.Internal.Contracts
             {
                 SupportedFrameworks = supportedFrameworks,
                 OrderBy = searchOrderBy,
-                PackageTypes = packageTypes,
+                PackageType = packageType,
                 IncludeDelisted = includeDelisted,
             };
         }
@@ -77,8 +77,8 @@ namespace NuGet.VisualStudio.Internal.Contracts
             writer.Write(value.IncludePrerelease);
             writer.Write(IncludeDelistedPropertyName);
             writer.Write(value.IncludeDelisted);
-            writer.Write(PackageTypesPropertyName);
-            options.Resolver.GetFormatter<IEnumerable<string>>()!.Serialize(ref writer, value.PackageTypes, options);
+            writer.Write(PackageTypePropertyName);
+            writer.Write(value.PackageType);
             writer.Write(FilterPropertyName);
             options.Resolver.GetFormatter<SearchFilterType?>()!.Serialize(ref writer, value.Filter, options);
             writer.Write(OrderByPropertyName);
--- a/src/NuGet.Core/NuGet.Protocol/PublicAPI/net472/PublicAPI.Shipped.txt
+++ b/src/NuGet.Core/NuGet.Protocol/PublicAPI/net472/PublicAPI.Shipped.txt
@@ -267,8 +267,6 @@ NuGet.Protocol.Core.Types.SearchFilter.IncludeDelisted.set -> void
 NuGet.Protocol.Core.Types.SearchFilter.IncludePrerelease.get -> bool
 NuGet.Protocol.Core.Types.SearchFilter.OrderBy.get -> NuGet.Protocol.Core.Types.SearchOrderBy?
 NuGet.Protocol.Core.Types.SearchFilter.OrderBy.set -> void
-~NuGet.Protocol.Core.Types.SearchFilter.PackageTypes.get -> System.Collections.Generic.IEnumerable<string>
-~NuGet.Protocol.Core.Types.SearchFilter.PackageTypes.set -> void
 NuGet.Protocol.Core.Types.SearchFilter.SearchFilter(bool includePrerelease) -> void
 NuGet.Protocol.Core.Types.SearchFilter.SearchFilter(bool includePrerelease, NuGet.Protocol.Core.Types.SearchFilterType? filter) -> void
 ~NuGet.Protocol.Core.Types.SearchFilter.SupportedFrameworks.get -> System.Collections.Generic.IEnumerable<string>
--- a/src/NuGet.Core/NuGet.Protocol/PublicAPI/net472/PublicAPI.Unshipped.txt
+++ b/src/NuGet.Core/NuGet.Protocol/PublicAPI/net472/PublicAPI.Unshipped.txt
@@ -1,6 +1,8 @@
 #nullable enable
 NuGet.Protocol.Plugins.PluginDiscoverer.PluginDiscoverer() -> void
 virtual NuGet.Protocol.Plugins.PluginFactory.Dispose() -> void
+~NuGet.Protocol.Core.Types.SearchFilter.PackageType.get -> string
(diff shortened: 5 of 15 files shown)


src/NuGet.Core/NuGet.Protocol/Resources/PackageSearchResourceV3.cs
```diff
@@ -142,7 +142,7 @@ private async Task<T> SearchPage<T>(
                 {
                     var types = string.Join("&",
                         filters.PackageTypes.Select(
-                            s => "packageTypeFilter=" + s));
+                            s => "packageType=" + s));
                     queryString += "&" + types;
```
zivkan: @joelverhagen are multiple package types defined as `&packageType=type1&packageType=type2` or `&packageType=type1+type2`?

Additionally, are the semantics that the package must contain all package types, or just one? This is important, to make sure that local folder feed sources behave in the same way.

[the docs](https://learn.microsoft.com/en-us/nuget/api/search-query-service-resource) are not sufficiently specific about multiple values.

src/NuGet.Core/NuGet.Protocol/LocalRepositories/LocalPackageSearchResource.cs
```diff
@@ -59,6 +59,21 @@ public async override Task<IEnumerable<IPackageSearchMetadata>> SearchAsync(
...
+                    && filters.PackageTypes.Any())
+                {
+                    foreach (var packageTypeName in filters.PackageTypes)
+                    {
+                        query = query
+                            .Where(package => package.Nuspec
+                                .GetPackageTypes()
+                                .Any(packageType => StringComparer.OrdinalIgnoreCase.Equals(
+                                    packageType.Name,
+                                    packageTypeName)));
```
zivkan: As per the PR checklist, please add tests. There's already a test class named [`LocalpackageSearchResourceTests`](https://github.com/NuGet/NuGet.Client/blob/dev/test/NuGet.Core.Tests/NuGet.Protocol.Tests/LocalResourceTests/LocalPackageSearchResourceTests.cs)

src/NuGet.Core/NuGet.Protocol/Resources/PackageSearchResourceV3.cs
```diff
@@ -142,7 +142,7 @@ private async Task<T> SearchPage<T>(
                 {
                     var types = string.Join("&",
                         filters.PackageTypes.Select(
-                            s => "packageTypeFilter=" + s));
+                            s => "packageType=" + s));
```
zivkan: The fact that no tests failed, shows a gap in testing. As per the PR checklist, please add tests.

There's already a file [`PackageSearchResourceV3Tests.cs`](https://github.com/NuGet/NuGet.Client/blob/dev/test/NuGet.Core.Tests/NuGet.Protocol.Tests/PackageSearchResourceV3Tests.cs) that shows a pattern how to test with mock http responses.

src/NuGet.Core/NuGet.Protocol/LocalRepositories/LocalPackageSearchResource.cs
glopesdev: @zivkan I would love to, but I am blocked setting up a working full environment due to package restore issues:

```
Failed to download package 'Microsoft.VisualStudio.Sdk.17.10.40171' from 'https://pkgs.dev.azure.com/azure-public/3ccf6661-f8ce-4e8a-bb2e-eff943ddd3c7/_packaging/36a629e1-6c5b-4bcd-aa2e-6018802d6b99@2f2420a2-f975-407f-ad27-d1e0f85b9b15/nuget/v3/flat2/microsoft.visualstudio.sdk/17.10.40171/microsoft.visualstudio.sdk.17.10.40171.nupkg'.
  An error occurred while sending the request.
    The underlying connection was closed: An unexpected error occurred on a send.
    Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host.
    An existing connection was forcibly closed by the remote host
```

I can restore nuget packages from any other project successfully on my machine and have confirmed all TLS mitigation strategies are in place as per https://github.com/microsoft/azure-devops-tls12 and mentioned in https://github.com/NuGet/NuGetGallery/issues/9121.

I will continue investigating, but if I am missing anything obvious, please advise.

glopesdev: This is extremely weird, but solved this problem by forcing IPv4 using https://community.fabric.microsoft.com/t5/Power-Query/The-underlying-connection-was-closed-An-unexpected-error/m-p/3442332#M112277

Working on tests now, but just leaving this here as it might indicate some kind of possible TLS issue with IPv6 on some machines / connections that was hard to find / diagnose.

src/NuGet.Core/NuGet.Protocol/Resources/PackageSearchResourceV3.cs
joelverhagen: Only a single package type filter is supported. The package type parameter is plumbed through in our search service as a single string:
https://github.com/NuGet/NuGetGallery/blob/dc7abf2bd145d91596b032c19e1c9abe8b276956/src/NuGet.Services.AzureSearch/SearchService/SearchParametersBuilder.cs#L169-L172

PRs on the docs are welcome if they are unclear!

joelverhagen: Regarding multiple package types on the package, the package type parameter matches the package if ANY of the package types match, per:
> The `packageType` parameter is used to further filter the search results to only packages **that have at least one package type matching the package type name**.

Emphasis mine.

glopesdev: @joelverhagen @zivkan Thank you both for the feedback. In that case, it seems that the API surface of `SearchFilter` might be slightly inconsistent with current `SearchQueryService/3.5.0` capabilities, since `SearchFilter` allows for multiple package types to be specified.

How do you prefer to address this? Shall `PackageSearchResourceV3` and `LocalPackageSearchResource` throw if more than one package type is provided? Or shall we deprecate the entire `PackageTypes` property and introduce a new `PackageType` property with a single value?

Personally I would prefer not touching the current surface API, since this feature is very much needed and breaking the surface API might delay adoption and confuse existing API consumers (having two properties vary their names by a single letter is usually very much discouraged).

@joelverhagen Is there any expectation that multiple package type filters will ever be supported? That might inform whether the deprecation route might be preferred.

glopesdev: @zivkan One last API design question which just occurred to me: what should the behavior of `LocalPackageSearchResource` be if the package type `Dependency` is specified and the local package does not specify package types, e.g. for older packages without the `packageTypes` metadata?

In [the docs](https://learn.microsoft.com/en-us/nuget/create-packages/set-package-type?tabs=dotnet#known-package-types) it is stated that _"Packages not marked with a type, including all packages created with earlier versions of NuGet, default to the `Dependency` type."_

This would suggest that we should accept matches for all packages without `packageTypes` metadata if the specified package type filter is `Dependency`.

src/NuGet.Core/NuGet.Protocol/LocalRepositories/LocalPackageSearchResource.cs
glopesdev: @zivkan One last API design question which just occurred to me: what should the behavior of `LocalPackageSearchResource` be if the package type `Dependency` is specified and the local package does not specify package types, e.g. for older packages without the `packageTypes` metadata?

In [the docs](https://learn.microsoft.com/en-us/nuget/create-packages/set-package-type?tabs=dotnet#known-package-types) it is stated that _"Packages not marked with a type, including all packages created with earlier versions of NuGet, default to the `Dependency` type."_

This would suggest that we should accept matches for all packages without `packageTypes` metadata if the specified package type filter is `Dependency`.

Alternatively we could change the implementation of `GetPackageTypes()` to return `Dependency` instead of an empty list, if no `packageTypes` node is found in the metadata, but that might break some client code.

src/NuGet.Core/NuGet.Protocol/Resources/PackageSearchResourceV3.cs
joelverhagen: There is no expectation that NuGet.org should support multiple package type filters. If this is needed, we would need to design, implement, and document a protocol enhancement. I am not aware of any other package source that implements multiple package type filters.

IMHO, `PackageTypes` should be deprecated and `PackageType` should be introduced, to match the current state of the protocol. But I do not work on the client side much so I may be missing something.

joelverhagen: I don't know of any plans to add multiple package type support. It's not even clear to what existing scenario could benefit from such a filter.

src/NuGet.Core/NuGet.Protocol/LocalRepositories/LocalPackageSearchResource.cs
zivkan: sorry for the delay in replying. Too much work and not enough time.

> This would suggest that we should accept matches for all packages without packageTypes metadata if the specified package type filter is Dependency.

This is my preference. I'd rather `GetPackageTypes` return what the package's nuspec actually reports, just in case there are scenarios where the difference between `Dependency` not "not specified" is important elsewhere.

src/NuGet.Core/NuGet.Protocol/Resources/PackageSearchResourceV3.cs
zivkan: > Personally I would prefer not touching the current surface API

To be fair, this PR is breaking server APIs, even if it's not breaking the assembly's API. If there are any servers (that are not nuget.org) that use the current `packageTypeFilter`, this PR will break clients using those servers. [There's evidence](https://github.com/NuGet/NuGetGallery/issues/5999) that (in 2018) neither nuget.org nor myget implemented `packageTypeFilter`, but there's no way to know if some other server uses it. On the other hand, the V2 protocol isn't documented, and `packageTypeFilter` was never part of [the official V3 protocol](https://learn.microsoft.com/en-us/nuget/api/search-query-service-resource).

I'm trying to rush reviewing a bunch of old PRs that I haven't reviewed in a long time, so I haven't yet formed an opinion about breaking APIs (about changing NuGet.Protocol to remove the plural `PackageTypes` and have a singular `PackageType`). There's a case to be made that NuGet.Client should implement the official protocol spec.

fforjan:
any update on this  ?
@zivkan or @nkolev92  ?

fforjan:
the reason is that I could not find a way to filter on a package type for a search, i'm not aware of any workaround

zivkan:
Sorry again for the delay. The lead up to .NET 9 GA, and the days following, were very busy.

Anyway, if someone in the NuGet team was doing this, I would ask them for the following:

1. The protocol docs say that the server needs to publish the `SearchQueryService/3.5.0` resource in the service index for package type filtering to be available, so I think `PackageSearchResourveV3` should throw a `NotSupportedException` when a package type is supplied, and the server does not have the `SearchQueryService/3.5.0` resource.
2. Clients need to know when a server supports package type filtering, so add a boolean property so that clients can make decisions and avoid unexpected exceptions.
3. As stated [here](https://github.com/NuGet/NuGet.Client/pull/5991#discussion_r1743412106), I think we should break the existing API because the server protocol doesn't support multiple package type filters. I don't see the point in maintaining an API which never worked, and therefore is very unlikely that anyone actually uses. I used to be much more conservative in not breaking public APIs, but NuGet.Client has accumulated too much tech debt, and public APIs are one contributing factor. Alternatively throw an exception when more than one package type is provided, but this doesn't make me feel good. I don't see multi-package type filtering being added in the future.

I don't consider myself "good" at open source. After all, look at how much time passes between my comments here. I also don't know if it's reasonable to expect the same quality of contribution from external people as team members who work in the code every day. What I am confident about is that when something doesn't work as expected, NuGet will get the support requests, not the original contributor, which is why I typically don't approve PRs from external contributions unless they're close to the quality that I'd expect of a team member.  And I'm working on higher priority changes, so I don't have the time to take over this PR and implement the 3 things I listed above.

glopesdev:
> I think we should break the existing API because the server protocol doesn't support multiple package type filters.

@zivkan I was trying to do this and started getting the below analyzer error when declaring a single string property:

```c#
public string PackageType { get; set; }
```

`RS0016 All public types and members should be declared in PublicAPI.txt. This draws attention to API changes in the code reviews and source control history, and helps prevent breaking changes.`

Upon investigation there are actually three different `PublicAPI.Shipped.txt` files (`net472`, `netcoreapp5.0`, and `netstandard2.0`) where I found the currrent API declared with the syntax:
```
~NuGet.Protocol.Core.Types.SearchFilter.PackageTypes.get -> System.Collections.Generic.IEnumerable<string>
~NuGet.Protocol.Core.Types.SearchFilter.PackageTypes.set -> void
```

Trying to follow the pattern for other properties I changed it to:
```
~NuGet.Protocol.Core.Types.SearchFilter.PackageType.get -> string
~NuGet.Protocol.Core.Types.SearchFilter.PackageType.set -> void
```

As expected this now raises the same RS0016 error for the existing `PackageTypes` property, but still does not let me specify the new `PackageType` property.

Is there anything obvious I might be missing? Also, should I be adding the new API to the `Unshipped.txt` files instead? This made me wonder whether we want to mark the old property `PackageTypes` as obsolete instead of simply removing it. Is there a convention or guidelines somewhere on the recommended approach to introducing such breaking changes?

Apologies in advance for not being very clear on the use and syntax of the `PublicAPI.txt` files.

zivkan:
I also don't know the syntax for the public API files. When using an IDE, the analyzer will have a codefix (available from the "lightbulb" button), which will automatically add a syntactically correct line into the Unshipped file.

Unfortunately the analyzer (or roslyn itself) only works on one target framework at a time, so you need to repeat codefix for each target framework, as we documented here: https://github.com/NuGet/NuGet.Client/blob/dev/docs/nuget-sdk.md#development

glopesdev:
@zivkan Thanks for the feedback, no worries about any delay.

> 1. The protocol docs say that the server needs to publish the `SearchQueryService/3.5.0` resource in the service index for package type filtering to be available, so I think `PackageSearchResourveV3` should throw a `NotSupportedException` when a package type is supplied, and the server does not have the `SearchQueryService/3.5.0` resource.

This is a good point, and made me go back to double-check whether `SearchQueryService/3.5.0` was actually requested as an endpoint. It turns out that it had been left out entirely from the list of versions in `ServiceTypes`. I have now fixed this in https://github.com/NuGet/NuGet.Client/pull/5991/commits/1e4abdf44c43c11620dc7b7eedc4c36d589b3abc.

However, I am a bit at a loss at how to implement the check you requested, since the way that search query service endpoints are accessed right now is not via the "resource" abstraction. Instead, the API requests a list of URI endpoints via the `GetServiceEntryUris` extension method:

https://github.com/NuGet/NuGet.Client/blob/949031c79ea7f5c15b59f31e26d5355fb23795ca/src/NuGet.Core/NuGet.Protocol/Providers/PackageSearchResourceV3Provider.cs#L25-L26

Because `PackageSearchResourceV3` works directly on the list of `searchEndpoints`, there is no information about versions that can be accessed in the current version of the code.

My feeling is perhaps we should refactor the entire approach to use `GetServiceEntries` directly, which returns a list of `ServiceIndexEntry` objects. Inside this class there is a `ClientVersion` property which I hope corresponds to the version of the `SearchQueryService` (can you confirm this is true?).

> 2. Clients need to know when a server supports package type filtering, so add a boolean property so that clients can make decisions and avoid unexpected exceptions.

Given the above, I am going to interpret this as exposing a new boolean property in `PackageSearchResourceV3`, which will be true iff at least one of the endpoints is 3.5.0 compatible. We would then throw the exception only if none of the provided server endpoints supports 3.5.0.

Does this sound acceptable?

> 3. As stated [here](https://github.com/NuGet/NuGet.Client/pull/5991#discussion_r1743412106), I think we should break the existing API because the server protocol doesn't support multiple package type filters. I don't see the point in maintaining an API which never worked, and therefore is very unlikely that anyone actually uses. I used to be much more conservative in not breaking public APIs, but NuGet.Client has accumulated too much tech debt, and public APIs are one contributing factor. Alternatively throw an exception when more than one package type is provided, but this doesn't make me feel good. I don't see multi-package type filtering being added in the future.

Done, see https://github.com/NuGet/NuGet.Client/pull/5991/commits/cf99bb65863662b0cacd4ef688a71ef62b8b2587.

zivkan:
> However, I am a bit at a loss at how to implement the check you requested, since the way that search query service endpoints are accessed right now is not via the "resource" abstraction. Instead, the API requests a list of URI endpoints via the `GetServiceEntryUris` extension method:
> https://github.com/NuGet/NuGet.Client/blob/949031c79ea7f5c15b59f31e26d5355fb23795ca/src/NuGet.Core/NuGet.Protocol/Providers/PackageSearchResourceV3Provider.cs#L25-L26

Since `PackageSearchResourceV3Provider` gets the service entries, it can check the versions.  Then `PackageSearchResourceV3` can have a constructor parameter, or `init` property. 

> My feeling is perhaps we should refactor the entire approach to use `GetServiceEntries` directly, which returns a list of `ServiceIndexEntry` objects. Inside this class there is a `ClientVersion` property which I hope corresponds to the version of the `SearchQueryService` (can you confirm this is true?).

Something I don't like about the version strings we use is that it implies that newer client versions support all past features (making it "impossible" to deprecate features), and it implies that if a server wants to support a particular feature, it must also implement all past versions as well, even if they have features they don't want.  This makes sense for some resources and features like package metadata (registration) where older versions needed the response to be uncompressed, and newer clients support gzip and deflate.  I mean, that really should have been done via the `Accept-Encoding` header, but nuget.org's server architecture has a limitation that prevents that from being technically viable and therefore the people working on NuGet at the time decided to bake it into the service index resources instead.

Anyway, I'm getting way off topic.  What I'm trying to get at, is that going forward we probably need to treat the service index entries as feature flags, not as a client version.  `bool supportsPackageTypeFilter = entries.Contains("3.5.0")`, `bool supportsSomeFutureFeature = entries.Contains("9.8.7")`. Max resource version should not imply older features are supported.

> Given the above, I am going to interpret this as exposing a new boolean property in `PackageSearchResourceV3`
> ...
> 
> Does this sound acceptable?

Yes, that's exactly what I was thinking.

src/NuGet.Core/NuGet.Protocol/LocalRepositories/LocalPackageSearchResource.cs
glopesdev: Fixed in https://github.com/NuGet/NuGet.Client/pull/5991/commits/9b1d7696c91638e653566f66dd70a4f27a6f65af

glopesdev:
> Something I don't like about the version strings we use is that it implies that newer client versions support all past features (making it "impossible" to deprecate features), and it implies that if a server wants to support a particular feature, it must also implement all past versions as well, even if they have features they don't want.

> The protocol docs say that the server needs to publish the SearchQueryService/3.5.0 resource in the service index for package type filtering to be available, so I think PackageSearchResourveV3 should throw a NotSupportedException when a package type is supplied, and the server does not have the SearchQueryService/3.5.0 resource.

@zivkan I've been thinking more about these two points and I admit I am now conflicted on this. My personal use case is reproducing a NuGet client much like the VS code package manager. In these types of front-end, the user may select among a large number of package sources for search, including local folders, old v3 and v2 clients, or the "All" option which works as a mixed package source which can run search across all of the above simultaneously.

At that point, what should happen if the user runs a filtered search on "All" and by chance it happens that one of the sources does not support package type filtering? Should an unhandled exception bubble up to the user in this case? And what should the expected resolution be from a user point of view? Should we really disallow running package type filtering on the "All" source entirely whenever the user adds a single package source which does not support it? This would mean that to fix the situation the user would have to understand pretty deeply the NuGet service index protocol and modify by hand the package source list to get things working again.

I am really fearful this could create confusion and further error reports. Also it makes the abstraction of the `SearchFilter` class start to leak a bit. This class represents filtering options over all possible package sources for both v2 and v3 protocols. If suddenly the user of `SearchFilter` needs to worry about the exact contents of the service index entry list I feel the API will become even more cumbersome and complicated to use.

Not sure about others, but I know for example that we wouldn't use it in this case, and would simply default to having no filter and instead use our current workaround of wrapping the resource and modify query strings directly so we retain the simplicity.

What might be the negative impact of simply ignoring the filter if the resource does not support it? Maybe @joelverhagen has some thoughts on this since related decisions were impacting the SDK, e.g. https://github.com/dotnet/sdk/issues/12038?

glopesdev:
> At that point, what should happen if the user runs a filtered search on "All" and by chance it happens that one of the sources does not support package type filtering?

I guess answering myself, one option might be we leave it to the client front-end to adjust its own implementation of the "All" feed to avoid launching queries to package sources which do not support package type filtering.

zivkan:
> one option might be we leave it to the client front-end to adjust its own implementation of the "All" feed to avoid launching queries to package sources which do not support package type filtering.

Yes, this was exactly my intention.

> If suddenly the user of SearchFilter needs to worry about the exact contents of the service index entry list I feel the API will become even more cumbersome and complicated to use.

That's the point of the boolean property. The front ends check it before setting the type filter.

> What might be the negative impact of simply ignoring the filter if the resource does not support it?

Customers expecting the results to only contain packages matching the type filter, but getting packages that do not match the type filter, and then getting confused, frustrated, and/or reporting bugs to us.

PR #5993, opened 2024-08-26
Log inner exception for caught FatalProtocolExcetions in SourceRepositoryDependencyProvider

### LINKED ISSUE #12530: When a source isn't accessible, service index cannot be read issues suppress the internal message making it difficult to understand the root cause
Create a project with the following content. 

Restore it.

```csproj

  <PropertyGroup>
    <Nullable>enable</Nullable>
    <RestoreSources>https://api.nuget.org/v3/indexss.json</RestoreSources>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="nuget.common" version="6.6.0" />
  </ItemGroup>
</Project>
```

> C:\Code\ServiceIndex\ServiceIndex.csproj : error NU1301: Unable to load the service index for source https://api.nuget.org/v3/indexss.json.

The error message on detailed verbosity is not any better. It's very apparent here that some exceptions are being swallowed.

Nigusu-Allehu:
# Bug

Fixes: https://github.com/NuGet/Home/issues/12530

## Description

Makes sure inner exception is logged whene we have a FatalProtocolException. 

If a source that does not respond is added, currently, here is what our logged error messages look like
> error NU1301: Unable to load the service index for source https://api.nuget.org/v3/indexss.json.

After this PR here is what the error would be

> NU1301: Unable to load the service index for source https://api.nuget.org/v3/indexss.json.
>Response status code does not indicate success: 404 (The specified resource does not exist.).

## PR Checklist

- [x] Added tests

### DIFF
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/SourceRepositoryDependencyProvider.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/SourceRepositoryDependencyProvider.cs
@@ -682,13 +682,8 @@ namespace NuGet.Commands
                 else
                 {
                     await logger.LogAsync(RestoreLogMessage.CreateError(NuGetLogCode.NU1301, e.Message, id));
-                    Exception exception = e as Exception;
 
-                    while (exception.InnerException != null)
-                    {
-                        await logger.LogAsync(level: LogLevel.Verbose, e.InnerException.Message);
-                        exception = exception.InnerException;
-                    }
+                    logger.Log(LogLevel.Verbose, ExceptionUtilities.DisplayMessage(e));
                 }
             }
 
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/SourceRepositoryDependencyProvider.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/SourceRepositoryDependencyProvider.cs
@@ -682,9 +682,11 @@ namespace NuGet.Commands
                 {
                     await logger.LogAsync(RestoreLogMessage.CreateError(NuGetLogCode.NU1301, e.Message, id));
 
-                    if (e.InnerException != null)
+                    while (e.InnerException != null)
                     {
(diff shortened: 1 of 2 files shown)


src/NuGet.Core/NuGet.Commands/RestoreCommand/SourceRepositoryDependencyProvider.cs
```diff
@@ -682,6 +682,13 @@ private async Task LogErrorAsync(ILogger logger, string id, FatalProtocolExcepti
                 else
                 {
                     await logger.LogAsync(RestoreLogMessage.CreateError(NuGetLogCode.NU1301, e.Message, id));
+                    Exception exception = e as Exception;
+
+                    while (exception.InnerException != null)
+                    {
+                        await logger.LogAsync(level: LogLevel.Verbose, e.InnerException.Message);
```
nkolev92: Ideally you want these in a single log message. 

The rule of thumb is that each log message should have an error code. This causes additional messages to be included. If this happens for multiple projects, the fact which messages refer to which issues can be lost.

Nigusu-Allehu: We really don't know what exactly the message from the exception is; would Nu1301 be appropriate for this?

Nigusu-Allehu: just used `ExceptionUtilities.DisplayMessage(e)` instead. It makes it a single log message.

zivkan (approved):

nkolev92 (approved):

src/NuGet.Core/NuGet.Commands/RestoreCommand/SourceRepositoryDependencyProvider.cs
nkolev92: perfect.

PR #5998, opened 2024-08-27
Get the packages metadata for all packages in the solution level

martinrrm:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13758
Fixes: https://github.com/NuGet/Home/issues/13759

## Description
This PR enables the Solution PM UI to have `IPackageSearchMetadata` for all the packages, even if they are not the latest version installed, this helps us to have a list of installed and transitive's packages in the `PackageItemVireModel` to retrieve the vulnerabilities of all the versions in the Solution.

image

## PR Checklist

- ~[ ] Added tests~ I'm creating the PR to see what breaks, after that I'll update/add tests

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
@@ -273,7 +273,6 @@ namespace NuGet.PackageManagement.UI
                 if (listItemViewModels.TryGetValue(packageId, out PackageItemViewModel existingListItem))
                 {
                     existingListItem.InstalledVersions = existingListItem.InstalledVersions.Append(packageVersion);
-                    existingListItem.UpdateInstalledPackagesVulnerabilities(new PackageIdentity(packageId, packageVersion));
                 }
                 else
                 {
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
@@ -6,7 +6,6 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Globalization;
 using System.Linq;
-using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft;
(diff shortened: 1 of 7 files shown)


src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
```diff
@@ -263,85 +262,101 @@ public IEnumerable<PackageItemViewModel> GetCurrent()
...
+                    VersionRange versionOverride = null;
 
-                    if (allowedVersionsRange.Length > 0)
+                    // get the allowed version range and pass it to package item view model to choose the latest version based on that
+                    if (_packageReferences != null)
                     {
-                        allowedVersions = allowedVersionsRange[0];
+                        IEnumerable<IPackageReferenceContextInfo> matchedPackageReferences = _packageReferences.Where(r => StringComparer.OrdinalIgnoreCase.Equals(r.Identity.Id, metadataContextInfo.Identity.Id));
+                        VersionRange[] allowedVersionsRange = matchedPackageReferences.Select(r => r.AllowedVersions).Where(v => v != null).ToArray();
+                        VersionRange[] versionOverrides = matchedPackageReferences.Select(r => r.VersionOverride).Where(v => v != null).ToArray();
```
jebriede: We can refactor this to iterate over matchedPackageReferences once and get the AllowedVersions and VersionOverride.

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
```diff
@@ -263,85 +262,101 @@ public IEnumerable<PackageItemViewModel> GetCurrent()
...
+
+                        if (allowedVersionsRange.Length > 0)
+                        {
+                            allowedVersions = allowedVersionsRange[0];
+                        }
+
+                        if (versionOverrides.Length > 0)
+                        {
+                            versionOverride = versionOverrides[0];
+                        }
```
jebriede: We can use FirstOrDefault to make this code more concise. (Please verify correctness):
```suggestion
                        allowedVersions = allowedVersionsRange.FirstOrDefault() ?? VersionRange.All;                            
                        versionOverride = versionOverrides.FirstOrDefault();
```

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
```diff
@@ -263,85 +262,101 @@ public IEnumerable<PackageItemViewModel> GetCurrent()
...
+                var packageVersion = metadataContextInfo.Identity.Version;
 
-                // get the allowed version range and pass it to package item view model to choose the latest version based on that
-                if (_packageReferences != null)
+                if (listItemViewModels.TryGetValue(packageId, out PackageItemViewModel exisitngListItem))
                 {
-                    IEnumerable<IPackageReferenceContextInfo> matchedPackageReferences = _packageReferences.Where(r => StringComparer.OrdinalIgnoreCase.Equals(r.Identity.Id, metadataContextInfo.Identity.Id));
-                    VersionRange[] allowedVersionsRange = matchedPackageReferences.Select(r => r.AllowedVersions).Where(v => v != null).ToArray();
-                    VersionRange[] versionOverrides = matchedPackageReferences.Select(r => r.VersionOverride).Where(v => v != null).ToArray();
+                    exisitngListItem.InstalledVersions = exisitngListItem.InstalledVersions.Append(packageVersion);
```
jebriede: Typos
```suggestion
                    existingListItem.InstalledVersions = existingListItem.InstalledVersions.Append(packageVersion);
```

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
```diff
@@ -136,6 +136,20 @@ public string ByOwnerOrAuthor
...
+        public IEnumerable<NuGetVersion> InstalledVersions
+        {
+            get
+            {
+                return _installedVersions;
+            }
+            set
+            {
+                _installedVersions = value;
+                OnPropertyChanged(nameof(InstalledVersions));
```
jebriede, line 151: This will be triggered when the InstalledVersions variable gets set to a new value, but not when a new element is appended to the list of Installed versions as on this line: https://github.com/NuGet/NuGet.Client/pull/5998/files?diff=unified&w=0#diff-07f06084bb40184699d53dcfdf13b4535adbb5b6ee122798b2a6b2c1d5bb9e31R274

If you want the View to know when a new InstalledVersion is appended to the list and refresh, you'll need to handle that case.

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
```diff
@@ -263,85 +262,101 @@ public IEnumerable<PackageItemViewModel> GetCurrent()
...
-                {
-                    listItem.UpdateTransitivePackageStatus(metadataContextInfo.Identity.Version);
+                    listItemViewModels[packageId] = listItem;
                 }
+            }
 
-                listItemViewModels.Add(listItem);
+            foreach (var item in listItemViewModels.Values)
+            {
+                item.UpdateInstalledPackagesVulnerabilities();
```
jebriede: UpdatePackageStatus and UpdateTransitivePackageStatus will both fetch vulnerability details for the package as it gets created. This new method seems to delegate to ReloadPackageVulnerabilitiesAsync which appears to fetch vulnerability data for all installed packages. Do we need to do this in both places? Can we fetch vulnerability data in only one place?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/InstalledAndTransitivePackageFeed.cs
```diff
@@ -32,9 +32,26 @@ public override async Task<SearchResult<IPackageSearchMetadata>> ContinueSearchA
...
+            var installedPackagesGrouped = _installedPackages.GroupById();
+            var installedPackagesLatest = installedPackagesGrouped.Select(g => g.OrderByDescending(x => x.Version).First());
+
+            var latestTransitiveGrouped = transitivePkgsWithOrigins.GroupById();
+            var latestTransitivePackages = latestTransitiveGrouped.Select(g => g.OrderByDescending(x => x.Version).First());
+
+            IPackageSearchMetadata[] installedItems = await GetMetadataForPackagesAndSortAsync(PerformLookup(installedPackagesLatest, searchToken), searchToken.SearchFilter.IncludePrerelease, cancellationToken);
+            IPackageSearchMetadata[] transitiveItems = await GetMetadataForPackagesAndSortAsync(PerformLookup(latestTransitivePackages, searchToken), searchToken.SearchFilter.IncludePrerelease, cancellationToken);
+
+            // Get metedata from identity for the rest of the installed and transitive packages
```
jebriede: Typo

```suggestion
            // Get metadata from identity for the rest of the installed and transitive packages
```

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/InstalledPackageFeed.cs
```diff
@@ -61,6 +62,20 @@ internal async Task<IPackageSearchMetadata[]> GetMetadataForPackagesAndSortAsync
...
+        internal static async Task<IPackageSearchMetadata[]> GetMetadataFromIdentityForPackagesAndSortAsync<T>(T[] packages, bool includePrerelease, CancellationToken cancellationToken) where T : PackageIdentity
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            IEnumerable<IPackageSearchMetadata> items = await TaskCombinators.ThrottledAsync(
+                packages,
+                (p, t) => Task.FromResult(PackageSearchMetadataBuilder.FromIdentity(p).Build()),
+                cancellationToken);
+
+            // The packages were originally sorted which is important because we skip based on that sort
```
jebriede, line 64: Question: What does it mean that we skip based on the sort? Maybe this is obvious, but if not, maybe the comment could use a little detail about that so it's clearer when we look at this again in the future.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/InstalledPackageFeed.cs
```diff
@@ -61,6 +62,20 @@ internal async Task<IPackageSearchMetadata[]> GetMetadataForPackagesAndSortAsync
...
+            // however, the asynchronous execution has randomly reordered the set. So, we need to resort.
+            return items.OrderBy(p => p.Identity.Id).ToArray();
```
jebriede, line 66: How do we ensure this sort stays consistent with the original sorting as the code evolves? Can we call the same method to sort so we don't duplicate code and risk them being out of sync should they be modified later?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/InstalledAndTransitivePackageFeed.cs
```diff
@@ -32,9 +32,26 @@ public override async Task<SearchResult<IPackageSearchMetadata>> ContinueSearchA
...
+            var remainingInstalledItems = await GetMetadataFromIdentityForPackagesAndSortAsync(PerformLookup(installedPackagesGrouped.Select(g => g.OrderByDescending(x => x.Version).Skip(1)).SelectMany(s => s), searchToken), searchToken.SearchFilter.IncludePrerelease, cancellationToken);
+            var remainingTransitiveItems = await GetMetadataFromIdentityForPackagesAndSortAsync(PerformLookup(latestTransitiveGrouped.Select(g => g.OrderByDescending(x => x.Version).Skip(1)).SelectMany(s => s), searchToken), searchToken.SearchFilter.IncludePrerelease, cancellationToken);
+
+            // Combine the installed and transitive packages
+            IPackageSearchMetadata[] items = installedItems
+                .Concat(remainingInstalledItems)
+                .Concat(transitiveItems)
+                .Concat(remainingTransitiveItems)
+                .ToArray();
```
jebriede: We can look for opportunities to refactor this code into some helper methods to increase readability and reusability. Happy to chat about this.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
martinrrm: Actually, the line that you mentioned will trigger this OnPropertyChanged because I'm doing an `=` operation.

Looking at the code we can do an improvement, changing this to an Observable list, this could save me doing the `=` when I'm already doing an append.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/InstalledPackageFeed.cs
martinrrm: Moved this to another method and removed the "AndSort" from these functions

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/InstalledPackageFeed.cs
martinrrm: So, I believe this function was used before on other feeds and this never got updated, after looking at the code when we are working with pagination we do these skips.

I recommend visiting this later in another PR because maybe we could save this sorting if there is no pagination in installed package feeds.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/InstalledAndTransitivePackageFeed.cs
martinrrm: Refactored the code into multiple methods, let me know what you think

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
```diff
@@ -263,85 +263,101 @@ public IEnumerable<PackageItemViewModel> GetCurrent()
...
+                                allowedVersionsRange.Add(reference.AllowedVersions);
+                            }
+                            if (reference.VersionOverride != null)
+                            {
+                                versionOverrides.Add(reference.VersionOverride);
+                            }
+                        }
+
+                        allowedVersions = allowedVersionsRange.FirstOrDefault() ?? VersionRange.All;
+                        versionOverride = versionOverrides.FirstOrDefault();
```
jebriede, line 302: These can be outside the if and then you don't need the instantiation above.

jebriede: Be sure to handle the scenario where versionOverrides is null if needed.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/InstalledAndTransitivePackageFeed.cs
```diff
@@ -29,16 +29,84 @@ public override async Task<SearchResult<IPackageSearchMetadata>> ContinueSearchA
...
+        {
+            var remainingPackages = packages
+                .GroupById()
+                .Select(g => g.OrderByDescending(x => x.Version).Skip(1)) // Skip 1 to get the remaining packages, latest package is already processed
+                .SelectMany(s => s);
+
+            return await GetMetadataFromIdentityForPackagesAsync(PerformLookup(remainingPackages, searchToken), searchToken.SearchFilter.IncludePrerelease, cancellationToken);
+        }
+
+        internal static async Task<IPackageSearchMetadata[]> GetMetadataFromIdentityForPackagesAsync<T>(T[] packages, bool includePrerelease, CancellationToken cancellationToken) where T : PackageIdentity
```
jebriede:
```suggestion
        private static async Task<IPackageSearchMetadata[]> GetMetadataFromIdentityForPackagesAsync<T>(T[] packages, bool includePrerelease, CancellationToken cancellationToken) where T : PackageIdentity
```

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/InstalledAndTransitivePackageFeed.cs
```diff
@@ -29,16 +29,84 @@ public override async Task<SearchResult<IPackageSearchMetadata>> ContinueSearchA
...
+
+            IEnumerable<IPackageSearchMetadata> items = await TaskCombinators.ThrottledAsync(
+                packages,
+                (p, t) => Task.FromResult(GetMetadataFromIdentityForPackage(p, t)),
+                cancellationToken);
+
+            return SortPackagesMetadata(items);
+        }
+
+        internal static IPackageSearchMetadata GetMetadataFromIdentityForPackage<T>(T identity, CancellationToken cancellationToken) where T : PackageIdentity
```
jebriede:
```suggestion
        private static IPackageSearchMetadata GetMetadataFromIdentityForPackage<T>(T identity, CancellationToken cancellationToken) where T : PackageIdentity
```

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/InstalledAndTransitivePackageFeed.cs
jebriede: Unless there's any external references to this

```suggestion
        private static async Task<IPackageSearchMetadata[]> GetMetadataFromIdentityForPackagesAsync<T>(T[] packages, bool includePrerelease, CancellationToken cancellationToken) where T : PackageIdentity
```

jebriede (approved):
Approving with a few suggestions

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
martinrrm: I believe we cannot move this to be outside the if, `allowedVersionsRange` and `versionsOverrides` are local variables inside this if, we could move those to be outside but since they are lists I think it's better to have 2 `VersionRanges` initialized. 

Also, since `versionOverrides` is initialized as an empty, `FirstOrDefault()` should return null if there are no values.

PR #6000, opened 2024-08-29
Enable GenerateDocumentationFile

jgonz120:
# Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2845

## Description
In order to enforce certain style rules we need to enable GenerateDocumentationFile for all projects. This is a known issue, https://github.com/dotnet/roslyn/issues/41640. The code we had for enabling GenerateDocumentationFile was not working because the value was being set by C# template. This now enables GenerateDocumentationFile for all projects. It also sets "WarningsNotAsErrors" for XML style errors for projects which are not shipped. CoPilot was used to generate documentation for some classes.

This also fixes errors that come up from enabling GenerateDocumentationFile. For using statements that need to be behind a conditional compilation, I grouped them together and placed them after the other using statements. 

## PR Checklist

- [X] ~Added tests~
- [X] ~Link to an issue or pull request to update docs if this PR changes settings, environment variables, new feature, etc.~

### DIFF
--- a/build/common.project.props
+++ b/build/common.project.props
@@ -135,6 +135,12 @@
     <VsixPublishDestination>$(ArtifactRoot)$(VsixOutputDirName)\</VsixPublishDestination>
   </PropertyGroup>
 
+  <!-- Write out .XML files for projects that will be packed. -->
+  <PropertyGroup>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <DocumentationFile Condition=" '$(DocumentationFile)' == '' AND '$(IsNetCoreProject)' != 'true' ">$(OutputPath)\$(AssemblyName).xml</DocumentationFile>
+  </PropertyGroup>
+
   <!-- Set the output location for all non-test projects -->
   <!-- Test projects currently fail when the output dir is moved -->
   <PropertyGroup Condition=" '$(TestProject)' != 'true' OR '$(Shipping)' == 'true'">
--- a/build/common.project.props
+++ b/build/common.project.props
@@ -135,7 +135,7 @@
     <VsixPublishDestination>$(ArtifactRoot)$(VsixOutputDirName)\</VsixPublishDestination>
   </PropertyGroup>
 
-  <!-- Write out .XML files for projects that will be packed. -->
+  <!-- Write out .XML files for projects all projects, required to enforce IDE0005 on build. -->
   <PropertyGroup>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <DocumentationFile Condition=" '$(DocumentationFile)' == '' AND '$(IsNetCoreProject)' != 'true' ">$(OutputPath)\$(AssemblyName).xml</DocumentationFile>
--- a/src/NuGet.Core/NuGet.Credentials/PluginCredentialRequest.cs
+++ b/src/NuGet.Core/NuGet.Credentials/PluginCredentialRequest.cs
@@ -8,12 +8,24 @@ namespace NuGet.Credentials
     /// </summary>
     public class PluginCredentialRequest
     {
+        /// <summary>
+        /// Gets or sets the URI for the credential request.
+        /// </summary>
         public string Uri { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether the request is non-interactive.
+        /// </summary>
         public bool NonInteractive { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether the request is a retry.
+        /// </summary>
         public bool IsRetry { get; set; }
 
+        /// <summary>
+        /// Gets or sets the verbosity level for the request.
+        /// </summary>
         public string Verbosity { get; set; }
     }
 }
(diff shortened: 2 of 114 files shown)


build/common.project.props
```diff
@@ -135,6 +135,12 @@
     <VsixPublishDestination>$(ArtifactRoot)$(VsixOutputDirName)\</VsixPublishDestination>
   </PropertyGroup>
 
+  <!-- Write out .XML files for projects that will be packed. -->
```
zivkan: The properties below are being used for all projects, not just those that are packed.

src/NuGet.Core/NuGet.Credentials/PluginCredentialRequest.cs
```diff
@@ -8,12 +8,24 @@ namespace NuGet.Credentials
     /// </summary>
     public class PluginCredentialRequest
     {
+        /// <summary>
+        /// Gets or sets the URI for the credential request.
+        /// </summary>
         public string Uri { get; set; }
```
zivkan: @kartheekp-ms I can never remember, is this set to the current URL that NuGet is trying to download, or is this always the service index?

src/NuGet.Core/NuGet.Credentials/PluginCredentialRequest.cs
```diff
@@ -8,12 +8,24 @@ namespace NuGet.Credentials
...
 
+        /// <summary>
+        /// Gets or sets a value indicating whether the request is non-interactive.
+        /// </summary>
         public bool NonInteractive { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether the request is a retry.
+        /// </summary>
         public bool IsRetry { get; set; }
```
zivkan, line 24: @kartheekp-ms I remember you had opinions in the past about how/when NuGet sets this `IsRetry` to true.  Can you suggest how the XML comment could be changed to more accurately describe when the value will be true or false?

zivkan (dismissed):

src/NuGet.Core/NuGet.Credentials/PluginCredentialRequest.cs
jgonz120: I did use copilot to generate these comments, I only kept the ones that seemed okay.

kartheekp-ms: > I can never remember, is this set to the current URL that NuGet is trying to download, or is this always the service index?

Thanks, Andy, for asking this question. I reviewed the code again and can confirm that it is always the package source URI passed to NuGet.exe specific credential providers. [Docs](https://learn.microsoft.com/nuget/reference/extensibility/nuget-exe-credential-providers#input-parameters) also state that `Uri` refers to `The package source URI requiring credentials.`

src/NuGet.Core/NuGet.Credentials/PluginCredentialRequest.cs
kartheekp-ms: Upon looking at the code again, I learnt that this type is only used in NuGet.exe specific credential providers in `$LocalAppData/NuGet/CredentialProviders` directory. I learnt from [this doc](https://learn.microsoft.com/en-us/nuget/reference/extensibility/nuget-exe-credential-providers#nugetexe-credential-provider-discovery) that these credential providers were meant to support only nuget.exe. [Cross-platform plugins](https://learn.microsoft.com/en-us/nuget/reference/extensibility/nuget-cross-platform-authentication-plugin) replaced these nuget.exe specific credential providers. I think this is a tech debt. I am considering creating a tech debt issue to consider removing nuget.exe specific credential providers.

https://github.com/NuGet/NuGet.Client/blob/f929a0f74b92c3593521a4556d41d6f96528fb24/src/NuGet.Clients/NuGet.CommandLine/ExtensionLocator.cs#L20-L26

kartheekp-ms: Given that this type is part of tech debt, I am okay with the existing comment.

The comment for `IsRetry` property is good for `IVSCredentialProvider` interface but can be improved for `ICredentialProvider` interface.

https://github.com/NuGet/NuGet.Client/blob/f929a0f74b92c3593521a4556d41d6f96528fb24/src/NuGet.Clients/NuGet.VisualStudio/Extensibility/IVsCredentialProvider.cs#L30-L31


https://github.com/NuGet/NuGet.Client/blob/f929a0f74b92c3593521a4556d41d6f96528fb24/src/NuGet.Core/NuGet.Credentials/ICredentialProvider.cs#L12-L24

src/NuGet.Core/NuGet.Credentials/PluginCredentialRequest.cs
zivkan:
```suggestion
        /// <summary>
        /// Gets or sets the package source URI for the credential request.
        /// </summary>
        public string Uri { get; set; }
```

zivkan (dismissed):

zivkan (approved):

PR #6007, opened 2024-09-03
Address feedback from original change to introduce new dependency resolution algorithm (second attempt)

jeffkl:
## Description
This restores a previously reverted change in https://github.com/NuGet/NuGet.Client/pull/6001.  The problem with the original change is that the `PackageSpecReferenceDependencyProvider` instance used by restore needs to respect what dependency graph resolution algorithm is in use.  However, the original change looked at the `PackageSpec` for the particular project being read which is incorrect.  So this brings back the original change with a fix to still tell the `PackageSpecReferenceDependencyProvider` which dependency resolution algorithm to use for the whole restore.

Work tracked in https://github.com/NuGet/Client.Engineering/issues/2959

## PR Checklist

- [ ] Added tests


src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -1490,20 +1476,26 @@ internal static List<ExternalProjectReference> GetProjectReferences(RestoreReque
...
+        /// </summary>
+        /// <param name="packageSpec">The <see cref="PackageSpec" /> with information about the project.</param>
+        /// <param name="runtimeIds">An <see cref="ISet{T}" /> containing the list of runtime identifiers.</param>
+        /// <returns>A <see cref="List{T}" /> containing <see cref="FrameworkRuntimePair" /> objects with the frameworks with empty runtime identifiers followed by frameworks with the specified runtime identifiers.</returns>
+        internal static List<FrameworkRuntimePair> CreateFrameworkRuntimePairs(PackageSpec packageSpec, ISet<string> runtimeIds)
         {
-            var projectFrameworkRuntimePairs = new List<FrameworkRuntimePair>();
-            foreach (var framework in packageSpec.TargetFrameworks)
+            // Create a list with capacity for each framework with no runtime and each framework/runtime
+            List<FrameworkRuntimePair> projectFrameworkRuntimePairs = new(capacity: packageSpec.TargetFrameworks.Count * (runtimeIds.Count + 1));
```
donnie-msft, line 1488: Possible dereference of null `TargetFrameworks ` and maybe also `packageSpec`?

jeffkl: There was no check of `null` before so I didn't bother.  As far as I know they'll never be null

donnie-msft: 👍 That's right, foreach would have thrown

donnie-msft (approved):

Nigusu-Allehu (approved):

PR #6009, opened 2024-09-04
Attempt Loading package Icon from local folders to avoid accessing remote feeds for metadata

### LINKED ISSUE #13766: Default Package icon shown even when embedded icon file exists on disk
### NuGet Product(s) Affected

### Current Behavior

As identified in https://github.com/NuGet/Home/issues/9250, on the Browse tab, when search results come from an upstream source in an Azure DevOps feed the package icon may be inaccessible.

### Desired Behavior

If the package already exists in the global packages folder or other local folder on the machine, load the icon from there rather than show the default icon.

### Additional Context

_No response_

### LINKED ISSUE #13768: testFile.txt is not cleaned up after running Tests
After running unit tests in the NuGet.Client repo., I noticed there are non-empty folders remaining in my `.test/work/` directory.

It seems that NuGet.Test.Utility.TestDirectory attempts to delete test folders after test completion; however, `NuGetPackageFileServiceTests` do 2 things:
1. Files are locked open due to their file sharing access.
2. A readonly test file is created, so it cannot be deleted

Example of leftover test artifacts on my machine:
image

donnie-msft:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13766
Fixes: https://github.com/NuGet/Home/issues/13768

## Description

If a package icon exists in a local metadata source (global packages folder and packages folders), load it from disk rather than requesting it from the remote feed.
Remote feeds may not be able to provide icon metadata if they are, for example, an upstream Azure DevOps feed which requires authentication. Additionally, upstream sources apparently do not copy down icon (and other) metadata to the main feed.

The goal is to sometimes avoid a lookup in the remote feed, which is:
  1. More efficient (disk vs network)
  1. Reduces the occurrence of the problem where an AzDo feed doesn't have the iconUri for a package, or that Uri is inaccessible (eg, missing authentication). The default package icon was shown for nearly all packages in search results for AzDo feeds. 

### Synopsis of algorithm
- During PM UI loading of Search results, lookup and store any local icon paths in the Icon cache (using a `localIcon:` prefix as the key) when the package is found locally and its `iconUri` is non-null. 
- When icons begin loading in the PM UI, first look for the local icon path in the cache
- If there's a cache miss or the file in the cached path doesn't exist (or can't be read), only then attempt to go to the remote feed for the icon. This can result in the default package icon being shown if the icon is both not locally available and not retrievable from the remote feed.

Example (before on the left; after on the right) of how this PR impacts PM UI:

_Highlighted squares on the right are icons which do not appear prior to this PR_
image

### Secondary refactoring
- Unrelated to this change is a refactoring in this area: Stop opening files with Write permissions in `GetStream`. I'm not aware of any case where search metadata would need Write permission to a local file path using the PM UI. Similar change was made for embedded files in https://github.com/NuGet/NuGet.Client/pull/3969

## PR Checklist

- [x] Added tests
- [x] Link to an issue or pull request to update docs if this PR changes settings, environment variables, new feature, etc. N/A since this is enabling more icons to show when they currently fail to show in PM UI.

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetPackageFileService.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetPackageFileService.cs
@@ -114,7 +114,7 @@ namespace NuGet.PackageManagement.VisualStudio
                 return null;
             }
 
-            Stream? stream;
+            Stream? stream = null;
             if (IsEmbeddedUri(uri))
             {
                 stream = await GetEmbeddedFileAsync(uri, cancellationToken);
@@ -126,7 +126,8 @@ namespace NuGet.PackageManagement.VisualStudio
                 {
                     stream = await GetEmbeddedFileAsync(localUri, cancellationToken);
                 }
-                else
+
+                if (stream == null)
                 {
                     stream = await GetStream(uri);
                 }
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetPackageFileService.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetPackageFileService.cs
@@ -179,7 +179,7 @@ namespace NuGet.PackageManagement.VisualStudio
                 // use GetFullPath to normalize "..", so that zip slip attack cannot allow a user to walk up the file directory
                 if (Path.GetFullPath(extractedIconPath).StartsWith(dirPath, StringComparison.OrdinalIgnoreCase) && File.Exists(extractedIconPath))
                 {
-                    Stream fileStream = new FileStream(extractedIconPath, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
+                    Stream fileStream = File.OpenRead(extractedIconPath);
                     return fileStream;
                 }
                 else
@@ -213,7 +213,7 @@ namespace NuGet.PackageManagement.VisualStudio
             {
                 if (File.Exists(uri.LocalPath))
                 {
-                    return new FileStream(uri.LocalPath, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
+                    return File.OpenRead(uri.LocalPath);
                 }
                 else
                 {
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetPackageFileService.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetPackageFileService.cs
@@ -26,6 +26,7 @@ namespace NuGet.PackageManagement.VisualStudio
     public sealed class NuGetPackageFileService : INuGetPackageFileService, IDisposable
     {
         public static readonly string IconPrefix = "icon:";
+        public static readonly string LocalIconPrefix = "localIcon:";
         public static readonly string LicensePrefix = "license:";
 
         private ServiceActivationOptions? _options;
@@ -52,6 +53,15 @@ namespace NuGet.PackageManagement.VisualStudio
         public static void AddIconToCache(PackageIdentity packageIdentity, Uri iconUri)
         {
             string key = NuGetPackageFileService.IconPrefix + packageIdentity.ToString();
+            if (iconUri != null)
+            {
+                IdentityToUriCache.Set(key, iconUri, CacheItemPolicy);
+            }
+        }
+
+        public static void AddLocalIconToCache(PackageIdentity packageIdentity, Uri iconUri)
+        {
+            string key = NuGetPackageFileService.LocalIconPrefix + packageIdentity.ToString();
             if (iconUri != null)
             {
                 IdentityToUriCache.Set(key, iconUri, CacheItemPolicy);
@@ -92,7 +102,9 @@ namespace NuGet.PackageManagement.VisualStudio
         public async ValueTask<Stream?> GetPackageIconAsync(PackageIdentity packageIdentity, CancellationToken cancellationToken)
         {
             Assumes.NotNull(packageIdentity);
-            string key = NuGetPackageFileService.IconPrefix + packageIdentity.ToString();
+            string packageId = packageIdentity.ToString();
+            string key = NuGetPackageFileService.IconPrefix + packageId;
+
             Uri? uri = IdentityToUriCache.Get(key) as Uri;
 
             if (uri == null)
@@ -109,10 +121,30 @@ namespace NuGet.PackageManagement.VisualStudio
             }
             else
             {
-                stream = await GetStream(uri);
+                Uri? localUri = GetLocalEmbeddedIconUri(packageId);
+                if (localUri is not null)
+                {
+                    stream = await GetEmbeddedFileAsync(localUri, cancellationToken);
+                }
+                else
+                {
+                    stream = await GetStream(uri);
+                }
             }
 
             return stream;
+        }
+
+        private static Uri? GetLocalEmbeddedIconUri(string packageId)
+        {
+            string localIconKey = NuGetPackageFileService.LocalIconPrefix + packageId;
+            Uri? localUri = IdentityToUriCache.Get(localIconKey) as Uri;
+            if (localUri is not null && IsEmbeddedUri(localUri))
+            {
+                return localUri;
+            }
+
+            return null;
         }
 
         public async ValueTask<Stream?> GetEmbeddedLicenseAsync(PackageIdentity packageIdentity, CancellationToken cancellationToken)
@@ -180,7 +212,7 @@ namespace NuGet.PackageManagement.VisualStudio
             {
                 if (File.Exists(uri.LocalPath))
                 {
-                    return new FileStream(uri.LocalPath, FileMode.Open);
+                    return new FileStream(uri.LocalPath, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
                 }
                 else
                 {
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/SearchObject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/SearchObject.cs
@@ -111,7 +111,12 @@ namespace NuGet.PackageManagement.VisualStudio
             var packageSearchMetadataContextInfoCollection = new List<PackageSearchMetadataContextInfo>(mainFeedResult.Items.Count);
             foreach (IPackageSearchMetadata packageSearchMetadata in mainFeedResult.Items)
             {
-                CacheBackgroundData(packageSearchMetadata, filter.IncludePrerelease);
+                IPackageSearchMetadata? localPackageSearchMetadata = null;
+
+                // Attach local metadata in case we do not have an icon remotely, can try local metadata.
+                localPackageSearchMetadata = await _packageMetadataProvider.GetOnlyLocalPackageMetadataAsync(packageSearchMetadata.Identity, cancellationToken);
+
+                CacheBackgroundData(packageSearchMetadata, localPackageSearchMetadata, filter.IncludePrerelease);
                 var knownOwners = CreateKnownOwners(packageSearchMetadata);
                 packageSearchMetadataContextInfoCollection.Add(PackageSearchMetadataContextInfo.Create(packageSearchMetadata, knownOwners));
             }
@@ -227,6 +232,11 @@ namespace NuGet.PackageManagement.VisualStudio
 
         private void CacheBackgroundData(IPackageSearchMetadata packageSearchMetadata, bool includesPrerelease)
         {
+            CacheBackgroundData(packageSearchMetadata, localPackageSearchMetadata: null, includesPrerelease);
+        }
+
+        private void CacheBackgroundData(IPackageSearchMetadata packageSearchMetadata, IPackageSearchMetadata? localPackageSearchMetadata, bool includesPrerelease)
+        {
             if (_inMemoryObjectCache == null)
             {
                 return;
@@ -247,6 +257,10 @@ namespace NuGet.PackageManagement.VisualStudio
             }
 
             NuGetPackageFileService.AddIconToCache(packageSearchMetadata.Identity, packageSearchMetadata.IconUrl);
+            if (localPackageSearchMetadata?.IconUrl != null)
+            {
+                NuGetPackageFileService.AddLocalIconToCache(packageSearchMetadata.Identity, localPackageSearchMetadata.IconUrl);
+            }
 
             string? packagePath = (packageSearchMetadata as LocalPackageSearchMetadata)?.PackagePath ??
                     (packageSearchMetadata as ClonedPackageSearchMetadata)?.PackagePath;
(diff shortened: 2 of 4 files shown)


src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetPackageFileService.cs
```diff
@@ -58,6 +59,15 @@ public static void AddIconToCache(PackageIdentity packageIdentity, Uri iconUri)
             }
         }
 
+        public static void AddLocalIconToCache(PackageIdentity packageIdentity, Uri iconUri)
+        {
+            string key = NuGetPackageFileService.LocalIconPrefix + packageIdentity.ToString();
+            if (iconUri != null)
+            {
+                IdentityToUriCache.Set(key, iconUri, CacheItemPolicy);
```
zivkan, line 67: I've said it before, so sorry if I sound like a broken record, but the file service (and all other services defined in NuGet.VisualStudio.Internal.Contracts) were added for a VS feature that never shipped. We should consider removing the layer to remove complexity.

Personally, I think these icon and licence APIs are the most impactful to VS performance, because they cache package data that might persist in memory after PM UI is closed and therefore the metadata is no longer needed.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetPackageFileService.cs
```diff
@@ -180,7 +213,7 @@ public NuGetPackageFileService(IServiceBroker serviceBroker, INuGetTelemetryProv
             {
                 if (File.Exists(uri.LocalPath))
                 {
-                    return new FileStream(uri.LocalPath, FileMode.Open);
+                    return new FileStream(uri.LocalPath, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
```
zivkan: `FileShare.ReadWrite` means that if another file handle opens this file for writing, that it's ok.   However, if another file handle is modifying the icon while this file stream is reading, then the resulting icon data may be a mix of the old and new files, probably leading to a corrupted image.

For better or worse, there's also a simpler way to open a file for reading, the [`File.OpenRead` API](https://learn.microsoft.com/en-us/dotnet/api/system.io.file.openread?view=net-8.0#system-io-file-openread(system-string)). It doesn't have all the overloads that the `FileStream` constructor has for advanced scenarios, but unless you need to use different parameters than what `File.OpenRead` does, it's a way to avoid needing to think about FileMode, FileAccess, and FileShare values.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/SearchObject.cs
```diff
@@ -226,6 +231,11 @@ public async ValueTask<int> GetTotalCountAsync(int maxCount, SearchFilter filter
         }
 
         private void CacheBackgroundData(IPackageSearchMetadata packageSearchMetadata, bool includesPrerelease)
+        {
+            CacheBackgroundData(packageSearchMetadata, localPackageSearchMetadata: null, includesPrerelease);
+        }
+
+        private void CacheBackgroundData(IPackageSearchMetadata packageSearchMetadata, IPackageSearchMetadata? localPackageSearchMetadata, bool includesPrerelease)
         {
             if (_inMemoryObjectCache == null)
```
zivkan, line 240: again, broken record time, sorry:

so both `NuGetPackageFileService` and `SearchObject` have `MemoryCache` instances?  Are we caching the same data in both?

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Services/NuGetPackageFileServiceTests.cs
```diff
@@ -112,6 +111,47 @@ public async Task GetPackageIconAsync_EmbeddedFromFallbackFolder_HasOnlyReadAcce
...
+                File.WriteAllText(filePath, fileContents);
+
+                var packageFileService = new NuGetPackageFileService(
+                        default(ServiceActivationOptions),
+                        Mock.Of<IServiceBroker>(),
+                        new AuthorizationServiceClient(Mock.Of<IAuthorizationService>()),
+                        _telemetryProvider.Object);
+
+                var packageIdentity = new PackageIdentity("packageA", NuGetVersion.Parse("1.0.0"));
+                var remoteInaccessibleUri = new Uri("http://api.source/index.json");
```
zivkan: always use test/fake urls ending in `.test` because it's a reserved TLD, whereas in theory `api.source` could be a valid DNS name.

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Services/NuGetPackageFileServiceTests.cs
```diff
@@ -153,6 +193,45 @@ public async Task GetPackageIconAsync_EmbeddedFromFallbackFolder_DoesNotLockFile
...
+
+                var packageFileService = new NuGetPackageFileService(
+                        default(ServiceActivationOptions),
+                        Mock.Of<IServiceBroker>(),
+                        new AuthorizationServiceClient(Mock.Of<IAuthorizationService>()),
+                        _telemetryProvider.Object);
+                var packageIdentity = new PackageIdentity(nameof(GetPackageIconAsync_NonEmbedded_GetStream_DoesNotLockFile), NuGetVersion.Parse("1.0.0"));
+
+                // Do not add a fragment which is typically required for the embedded icon.
+                // Note: file:// is required for Uri to parse the Fragment (possibly a Uri bug).
```
zivkan: The docs for the single string constructor for URI says this in the remarks: https://learn.microsoft.com/en-us/dotnet/api/system.uri.-ctor?view=net-8.0#system-uri-ctor(system-string)

> This constructor assumes that the string parameter references an absolute URI and is equivalent to calling the [Uri](https://learn.microsoft.com/en-us/dotnet/api/system.uri.-ctor?view=net-8.0) constructor with [UriKind](https://learn.microsoft.com/en-us/dotnet/api/system.urikind?view=net-8.0) set to [Absolute](https://learn.microsoft.com/en-us/dotnet/api/system.urikind?view=net-8.0#system-urikind-absolute). If the string parameter passed to the constructor is a relative URI, this constructor will throw a [UriFormatException](https://learn.microsoft.com/en-us/dotnet/api/system.uriformatexception?view=net-8.0).

so, no bug, it's by design. absolute URI's must have a scheme, so if the URI is intended to point to a file, it needs the `file:` scheme.  If you don't want to provide a scheme, you can use UriKind.Relative, but you'd want to consider if relative URIs make sense in the APIs that it gets used in.

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Services/NuGetPackageFileServiceTests.cs
```diff
@@ -153,6 +193,45 @@ public async Task GetPackageIconAsync_EmbeddedFromFallbackFolder_DoesNotLockFile
...
+                var uri = new Uri("file://" + filePath);
+                NuGetPackageFileService.AddIconToCache(packageIdentity, uri);
+                FileStream fileNotLocked = File.Open(uri.LocalPath, FileMode.Open, FileAccess.ReadWrite, FileShare.Read);
```
zivkan: I don't understand the scenario for this test. When would Visual Studio need to read a file from the global packages folder that is concurrently being written to by another file handle?

NuGet doesn't handle files in either the global packages folder, or file feed sources, changing during the process (which is a problem since we have a button to clear the global packages folder, but this test isn't testing delete, it's testing write).  I imagine this test isn't intended to test reading the icon while the nupkg is still getting extracted. I'm not sure that package search would find the package in such a scenario.

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Services/NuGetPackageFileServiceTests.cs
```diff
@@ -112,6 +111,47 @@ public async Task GetPackageIconAsync_EmbeddedFromFallbackFolder_HasOnlyReadAcce
...
+                // Add a fragment to consider this an embedded icon.
+                // Note: file:// is required for Uri to parse the Fragment (possibly a Uri bug).
+                var localUri = new Uri("file://" + filePath + "#testFile.txt");
+
+                NuGetPackageFileService.AddIconToCache(packageIdentity, remoteInaccessibleUri);
+
+                // Act
+                Stream inaccessibleRemoteIconStream = await packageFileService.GetPackageIconAsync(packageIdentity, CancellationToken.None);
+                NuGetPackageFileService.AddLocalIconToCache(packageIdentity, localUri);
+                Stream foundLocalIconStream = await packageFileService.GetPackageIconAsync(packageIdentity, CancellationToken.None);
```
zivkan: Since the file being opened exists inside the `TestDirectory` created at the start of this test, but these streams are not disposed, I expect that when the `TestDirectory` is disposed due to the `using` statement, that the directory delete fails because the file is in use. The test doesn't fail, because [the code used to delete the directories](https://github.com/NuGet/NuGet.Client/blob/8285b74c46ce9f01b1111c181bac535bc26aa9e0/test/TestUtilities/Test.Utility/TestFileSystemUtility.cs#L160-L180) silently ignores exceptions.

But the end result is that the test does not clean up after itself, which arguably is a test bug.

It's also a violation of the general coding convention that disposable objects must be disposed explicitly (and the `NuGetPackageFileService` is also not being disposed). While it might be questionable which coding standard/guidelines/recommendations we want to enforce in test code vs product code, the test directories not getting cleaned up is an unwanted side effect.

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Services/NuGetPackageFileServiceTests.cs
```diff
@@ -153,6 +193,45 @@ public async Task GetPackageIconAsync_EmbeddedFromFallbackFolder_DoesNotLockFile
...
+                // Note: file:// is required for Uri to parse the Fragment (possibly a Uri bug).
+                var uri = new Uri("file://" + filePath);
+                NuGetPackageFileService.AddIconToCache(packageIdentity, uri);
+                FileStream fileNotLocked = File.Open(uri.LocalPath, FileMode.Open, FileAccess.ReadWrite, FileShare.Read);
+
+                // Act
+                // System.IO.IOException would occur in the Act if we're locking the icon file.
+                Stream iconStream = await packageFileService.GetPackageIconAsync(packageIdentity, CancellationToken.None);
+                fileNotLocked.Close();
+                FileStream fileNotLocked2 = File.Open(uri.LocalPath, FileMode.Open, FileAccess.ReadWrite, FileShare.Read);
```
zivkan: multiple streams not getting disposed here, so like my other comment, I expect that the test directory is not getting delete when the test finishes.

zivkan (commented):
Fantastic customer usability win! 👍 

A few comments about tests that will be quick to improve, but also a question about ReadWrite

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Services/NuGetPackageFileServiceTests.cs
donnie-msft: Sure.. I wondered about that but as you may guess, I found this in many other tests. I've never heard of `.source` so figured we're using it because it's safe. I'll switch this one but I won't be refactoring all tests in this PR (meaning your concern will still exist in this repo)

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Services/NuGetPackageFileServiceTests.cs
donnie-msft: As I stated in the `Secondary refactoring` section of this PR description, I also don't think VS needs to write these files, yet if you look at the linked bug - VS was _opening_ the files for Write. The test is proving that we're not opening the icon for Write and only for Read.

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Services/NuGetPackageFileServiceTests.cs
donnie-msft: I'll delete the copied the note from another test - guess nobody noticed this when the first test was written.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/SearchObject.cs
donnie-msft: There's overlap but they are caching different objects. SearchObject caches the entire metadata object and the provider for each package. 
File Service is just caching the Uri for the icon or license.
It seems valuable to me to cache those Uris independently of the model object, but let me know if you have a different thought.

donnie-msft: Also, surprisingly I couldn't find any issues about this specific question, just one issue (https://github.com/NuGet/Client.Engineering/issues/2533) about how there's redundancy within SearchObject caching the same models from multiple methods (sounds similar to your point but not quite the same).

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Services/NuGetPackageFileServiceTests.cs
donnie-msft: Yes - I had noticed a cluttered test folder, but forgot to go back and cleanup the Streams. done now

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Services/NuGetPackageFileServiceTests.cs
donnie-msft: Yes - I had noticed a cluttered test folder, but forgot to go back and cleanup the Streams. done now

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetPackageFileService.cs
donnie-msft: I kind of like separating embedded file retrieval, regardless of whether it is being used as a "service" as originally intended. 
Maybe worth a separate discussion/issue around what exactly to change here? It sort of seems like a Model at this point which isn't really doing anything extra for remoting to codespaces/etc.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetPackageFileService.cs
donnie-msft: Switched to OpenRead!

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Services/NuGetPackageFileServiceTests.cs
donnie-msft: I've changed the tests and refactored variable/test names to try to be more clear about testing that PM UI can Read a file, even if already open for Read, but will not read the file if it's open for Write.

Let me know if this looks better.

zivkan (approved):

PR #6010, opened 2024-09-04
[Transitive's in Solution] Vulnerability indicators in solution level details pane for top-level packages

martinrrm:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13760

## Description
This PR enables calculating the vulnerabilities for all the packages that we have a metadata, in the previous PR for this feature we allowed the `PackageItemViewModel` to have data for all the installed versions in the Solution and store a copy of that metadata.

Vulnerability indicators are also being added to the Solution Detail Pane so users are able to identify vulnerable installed packages more easily.

image

## PR Checklist

- [x] Added tests, missing tests


src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
```diff
@@ -152,8 +152,8 @@ public ObservableCollection<NuGetVersion> InstalledVersions
             }
         }
 
-        private IEnumerable<NuGetVersion> _vulnerableVersions = [];
-        public IEnumerable<NuGetVersion> VulnerableVersions
+        private Dictionary<NuGetVersion, int> _vulnerableVersions = [];
+        public Dictionary<NuGetVersion, int> VulnerableVersions
```
martinrrm, line 156: This were mistakenly added in the previous PR and were never used.

src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/InstalledVulnerabilitiesCountConverter.cs
```diff
@@ -0,0 +1,37 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using System.Globalization;
+using System.Windows.Data;
+
+namespace NuGet.PackageManagement.UI
+{
+    public class InstalledVulnerabilitiesCountConverter : IValueConverter
```
jebriede: We don't need to create a converter simply for formatting a string using a resource. We should be able to use a MultiBinding with a StringFormat and reference the resource string.

src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/IntToVisibilityConverter.cs
```diff
@@ -0,0 +1,36 @@
...
+
+using System;
+using System.Diagnostics;
+using System.Globalization;
+using System.Windows;
+using System.Windows.Data;
+
+namespace NuGet.PackageManagement.UI
+{
+    public class IntToVisibilityConverter : IValueConverter
```
jebriede: This looks almost identical to the DownloadCountToVisibilityConverter. Can we repurpose that, perhaps rename it to something more generalized, and use it for number to visibility conversion?

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
```diff
@@ -120,12 +121,16 @@ public string InstalledVersions
...
+            var vulnerabilitiesHash = new HashSet<NuGetVersion>();
 
             foreach (var project in _projects)
             {
                 try
                 {
                     IPackageReferenceContextInfo installedVersion = await GetInstalledPackageAsync(project.NuGetProject, Id, cancellationToken);
+                    //var installedVersion = installedAndTransitivePackages.InstalledPackages;
+                    //var transitiveVersion = installedAndTransitivePackages.TransitivePackages;
+
```
jebriede: Remove commented-out code.

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
```diff
@@ -276,6 +289,16 @@ public int InstalledVersionsCount
             }
         }
 
+        public int InstaledVulnerabilitiesCount
```
jebriede: Typo

```suggestion
        public int InstalledVulnerabilitiesCount
```

jebriede (changes requested):
Looks good so far! Requesting a few changes.

src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/InstalledVulnerabilitiesCountConverter.cs
martinrrm: The problem with multi binding is that we cannot use conditionals inside it, the only way to make it "dynamic" depending on the vulnerabilities count is to use a converter when binding.

```xaml
<TextBlock>
  <TextBlock.Text>
    <MultiBinding Converter="{StaticResource CustomConverter}">
      <Binding Source="{x:Static local:Resources.Label_InstalledVulnerabilitiesCount}" />
      <Binding Path="InstaledVulnerabilitiesCount" />
    </MultiBinding>
  </TextBlock.Text>
</TextBlock>
```

or something like this

```xaml
<MultiBinding Converter="{StaticResource StringFormatConverter}">
    <Binding Source="{x:Static local:Resources.Label_InstalledVulnerabilitiesCount}" />
    <Binding Path="InstaledVulnerabilitiesCount"
             Converter="{StaticResource CustomConverter}"/>
</MultiBinding>
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/IntToVisibilityConverter.cs
martinrrm: I'm having a hard time trying to figure out how we could repurpose the DownloadConverter, usually something grater than 0 is true, else is false, but this is considering 0 as true.

src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/InstalledVulnerabilitiesCountConverter.cs
donnie-msft: I prefer the Converter over any inline XAML syntax as this converter can be tested, and IMO, is far easier to read.

src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/IntToVisibilityConverter.cs
donnie-msft: > I'm having a hard time trying to figure out how we could repurpose the DownloadConverter, usually something grater than 0 is true, else is false, but this is considering 0 as true.

I don't follow your comment here as neither converter returns booleans, but they are slightly different.

- DownloadCount returns visible for >=0.
- You're wanting a positive non-zero value (>0), and I agree that's different.

You could create one that takes options to determine if 0 is considered visible or not.

I also think it's ok as is, but recommend renaming this new one to be more specific to the property or, maybe: `PositiveIntToVisibilityConverter` ?

Consider instead of binding XAML to integers directly to instead create a ViewModel property that's a boolean, then you can just use the `BoolToVisiblityConverter`.

src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/InstalledVulnerabilitiesCountConverter.cs
jebriede: > The problem with multi binding is that we cannot use conditionals inside it

A data trigger can handle that.

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageInstallationInfo.cs
```diff
@@ -49,6 +49,26 @@ private async ValueTask InitializeAsync(CancellationToken cancellationToken)
...
+                    _installedVersionMaxVulnerability = value;
+                    OnPropertyChanged(nameof(InstalledVersionMaxVulnerability));
+                    OnPropertyChanged(nameof(IsInstalledVersionVulnerable));
+                }
+            }
+        }
+
+        public bool IsInstalledVersionVulnerable
+        {
+            get => InstalledVersionMaxVulnerability > -1;
```
jgonz120, line 69: Why does having 0 vulnerabilities imply that the installed version is vulnerable?

martinrrm: Vulnerability Severity can be from 0 to 3, Low, Moderate, High, Critical

src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/InstalledVulnerabilitiesCountConverter.cs
martinrrm: MultiBinding doesn't have an inner Style property so if we wanted to use a DataTrigger it would look like this:

```xaml
<TextBlock>
  <TextBlock.Style>
    <Style TargetType="TextBlock">
      <Setter Property="Text">
        <Setter.Value>
          <MultiBinding Converter="{StaticResource StringFormatConverter}">
            <Binding Source="{x:Static local:Resources.Label_InstalledVulnerabilitiesCountSingle}" />
            <Binding Path="InstalledVulnerabilitiesCount" />
          </MultiBinding>
        </Setter.Value>
      </Setter>
      <Style.Triggers>
        <DataTrigger Binding="{Binding InstalledVulnerabilitiesIsMoreThanZero}" Value="True">
          <Setter Property="Text">
            <Setter.Value>
              <MultiBinding Converter="{StaticResource StringFormatConverter}">
                <Binding Source="{x:Static local:Resources.Label_InstalledVulnerabilitiesCount}" />
                <Binding Path="InstalledVulnerabilitiesCount" />
              </MultiBinding>
            </Setter.Value>
          </Setter>
        </DataTrigger>
      </Style.Triggers>
    </Style>
  </TextBlock.Style>
</TextBlock>
```

jebriede: Looks good! Instead of needing an additional property on the ViewModel for the boolean, you could do the data trigger such that it sets it to the singular form if the value is 1, otherwise it would be plural (or 0, but without double checking, I'm assuming we're handling that with visibility. If not, you can also handle the 0 case easily).

jebriede: You can also use a data template resource instead of specifying it in-line. With that, you can create a generalized Style that can be applied to any case where we need a singular / plural differentiator.

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageInstallationInfo.cs
jebriede: In other words, the 0 represents the severity of the vulnerability, not the number of vulnerabilities it has.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
```diff
@@ -869,6 +897,34 @@ private async Task ReloadPackageMetadataAsync()
...
+                        await _vulnerabilityService.GetVulnerabilityInfoAsync(packageIdentity, cancellationToken);
+
+            var maxSeverity = vulnerabilityInfoList?.FirstOrDefault()?.Severity ?? -1;
+
+            if (maxSeverity > -1)
+            {
+                VulnerableVersions.Add(packageIdentity.Version, maxSeverity);
+            }
+
+            VulnerabilityMaxSeverity = Math.Max(VulnerabilityMaxSeverity, vulnerabilityInfoList?.FirstOrDefault()?.Severity ?? -1);
```
donnie-msft:
```suggestion
            VulnerabilityMaxSeverity = Math.Max(VulnerabilityMaxSeverity, maxSeverity);
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
```diff
@@ -1060,4 +1060,10 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
   <data name="Hyperlink_Owner" xml:space="preserve">
     <value>Owner</value>
   </data>
+  <data name="Label_InstalledVulnerabilitiesCount" xml:space="preserve">
+    <value>{0} vulnerable versions installed </value>
```
donnie-msft:
```suggestion
    <value>{0} vulnerable versions installed</value>
```

donnie-msft (approved):
Nice! Thanks for doing a little refactoring to make the converter more generic and updating tests!
Left some minor points that you could address before going to dev.

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
```diff
@@ -25,6 +25,7 @@ internal class PackageSolutionDetailControlModel : DetailControlModel
         private INuGetSolutionManagerService _solutionManager;
         private string _installedVersions; // The text describing the installed versions information, such as "not installed", "multiple versions installed" etc.
         private int _installedVersionsCount; // the count of different installed versions
+        private int _installedVulnerabilitiesCount; // the count of different vulnerable installed versions
```
jebriede:
```suggestion
        private int _installedVulnerabilitiesCount; // the count of distinct vulnerable installed versions
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
```diff
@@ -120,6 +121,7 @@ public string InstalledVersions
         private async Task UpdateInstalledVersionsAsync(CancellationToken cancellationToken)
         {
             var hash = new HashSet<NuGetVersion>();
+            var vulnerabilitiesHash = new HashSet<NuGetVersion>();
```
jebriede: Or simply `vulnerabilities`. The Hash part of the HashSet is an implementation detail of the specific kind of Set being used and implies it's the vulnerabilities Hashed. What we're really interested in is a unique, unordered list of vulnerabilities, i.e. a Set.

```suggestion
            var vulnerabilitiesSet = new HashSet<NuGetVersion>();
```

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
```diff
@@ -822,32 +822,60 @@ private async System.Threading.Tasks.Task ReloadPackageVersionsAsync()
             }
         }
 
-        private async Task ReloadPackageMetadataAsync()
+        private async Task ReloadTopLevelPackageMetadataAsync()
```
jebriede, line 839: ReloadTopLevelPackageMetadataAsync and SetVulnerabilityMaxSeverityAsync both calculate and add VulnerableVersions based on the same logic. Let's isolate that logic into a single method so it isn't duplicated.

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
```diff
@@ -1060,4 +1060,10 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
   <data name="Hyperlink_Owner" xml:space="preserve">
     <value>Owner</value>
   </data>
+  <data name="Label_InstalledVulnerabilitiesCount" xml:space="preserve">
+    <value>{0} vulnerable versions installed</value>
+  </data>
+  <data name="Label_InstalledVulnerabilitiesCountSingle" xml:space="preserve">
+    <value>{0} vulnerable version installed</value>
+  </data>
```
jebriede: It might be a good idea to add a comment for the translators to know that the {0} token is a numeric value.

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
```diff
@@ -120,6 +121,7 @@ public string InstalledVersions
         private async Task UpdateInstalledVersionsAsync(CancellationToken cancellationToken)
         {
             var hash = new HashSet<NuGetVersion>();
```
jebriede: It's worth updating this variable to distinguish from the new set we created for vulnerabilities and make the name reflect its role.

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
```diff
@@ -273,6 +274,7 @@ public IEnumerable<PackageItemViewModel> GetCurrent()
                 if (listItemViewModels.TryGetValue(packageId, out PackageItemViewModel existingListItem))
                 {
                     existingListItem.InstalledVersions.Add(packageVersion);
+                    existingListItem.UpdateInstalledPackagesVulnerabilities(new PackageIdentity(packageId, packageVersion));
```
jebriede, line 283: Can we just use the `metadataContextInfo.Identity` instead of creating and instantiating a new one?

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageInstallationInfo.cs
jgonz120: got it! thanks 😊

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
```diff
@@ -869,6 +890,37 @@ private async Task ReloadPackageMetadataAsync()
...
+        private void SetVulnerabilityMaxSeverity(NuGetVersion version, int maxSeverity)
+        {
+            if (maxSeverity > -1)
+            {
+                VulnerableVersions.Add(version, maxSeverity);
+            }
+
+            VulnerabilityMaxSeverity = Math.Max(VulnerabilityMaxSeverity, maxSeverity);
+
+            OnPropertyChanged(nameof(Status));
```
jebriede: For SetVulnerabilityMaxSeverity, would we want to only calculate and set the VulnerabilityMaxSeverity and call OnPropertyChanged if the maxSeverity is > -1?

jebriede (approved):
Approving with suggestion. Just one change then 🚀!

PR #6012, opened 2024-09-05
Add package reference for Microsoft.VisualStudio.Markdown.Platform

jgonz120:
# Bug

## PR Checklist

- [X] ~~Added tests~~


donnie-msft (approved):

NuGet.Config
```diff
@@ -45,27 +46,28 @@
...
       <package pattern="nugetvalidator" />
     </packageSource>
-    <packageSource key = "vside">
+    <packageSource key="vside">
       <package pattern="envdte" />
       <package pattern="envdte100" />
       <package pattern="envdte80" />
       <package pattern="envdte90" />
       <package pattern="envdte90a" />
+      <package pattern="Microsoft.*" />
```
zivkan: no, please don't add an identical pattern that exists in another package source. We had problems in the past when trying to upgrade the VS SDK, because the vside feed had packages not on nuget.org, and I don't remember the details, but it was problematic.

It's precisely why there are so many other patterns below starting with `Microsoft.`

jebriede (approved):

NuGet.Config
```diff
@@ -45,27 +46,29 @@
...
       <package pattern="nugetvalidator" />
     </packageSource>
-    <packageSource key = "vside">
+    <packageSource key="vside">
       <package pattern="envdte" />
       <package pattern="envdte100" />
       <package pattern="envdte80" />
       <package pattern="envdte90" />
       <package pattern="envdte90a" />
+      <package pattern="Microsoft.Build.Framework"/>
```
zivkan: This package should exist on nuget.org. Why are we pulling it from vside?

NuGet.Config
```diff
@@ -45,27 +46,29 @@
...
       <package pattern="Microsoft.DataAI.NuGetRecommender.Contracts" />
       <package pattern="microsoft.internal.visualstudio.*" />
       <package pattern="Microsoft.Internal.VisualStudio.Shell.Framework" />
+      <package pattern="Microsoft.NET.StringTools" />
```
zivkan: This package also exists on nuget.org https://www.nuget.org/packages/Microsoft.NET.StringTools

why do we need to get it from vside?

jgonz120: So for both of your comments, I'm getting a NU1603 error. So the version isn't available on dotnet-public. Is there a way to add it there? 


`error NU1603: Warning As Error: Microsoft.VisualStudio.Shell.Framework 17.11.38041-preview.1 depends on Microsoft.Build.Framework (>= 17.10.0) but Microsoft.Build.Framework 17.10.0 was not found. Microsoft.Build.Framework 17.10.4 was resolved instead.`

zivkan: It's not exactly the same thing, but here's an example, that I learned about just today, of how us changing MSBuild version numbers can have unintended consequences: https://github.com/dotnet/docs/pull/42393#discussion_r1746303334

If this Microsoft.Build.Framework or Microsoft.NET.StringTools version that is not on nuget.org ends up accidentally as a dependency of any of our own packages (which we do publish to nuget.org), then any customer who uses our package will then end up with an NU1603. We've had customers (once internal, once external) complain to us about this, and I've fixed it twice, hence why I feel strongly that any packages that exist on nuget.org, we should only use from the dotnet-public feed.

The dotnet-public feed already syncs all Microsoft owned packages from nuget.org, so if it's not already on dotnet-public then it's not on nuget.org.  You could ask the package owner if they're willing to publish that specific version (17.10.0). Both of these packages happen to be owned by the MSBuild team.

Alternatively, you can ask the owner of the package (Microsoft.VisualStudio.Shell.Framework) who lists that dependency to use a version that is public. But then they'd need to publish a new version of their package to have that updated dependency version. And whatever package listed that package as a dependency needs to update, and so up, until the package that we use is updated.

A last option is to add a `PackageReference` on the packages that are getting reported in the NU1603 message. I think by making it a direct dependency of the project, that should prevent the dependency resolver from trying to download the other version that's a transitive dependency.  But if the transitive version is a higher version that what we list in our CPM versions file, then we'll get a package downgrade error. And for Microsoft.Build.Framework, upgrading to a newer version isn't an option because of that issue reported in the dotnet docs today.

Ideally, the Markdown.Platform package wouldn't have both the implementation and API contracts in the same package, and that way transitive dependencies of its implementation (MSBuild's packages) wouldn't even come up in our restore graph.  This is exactly how NuGet's VS extensibility APIs work (NuGet.VisualStudio and NuGet.VisualStudio.Contracts, they're API interfaces only, no implementation, with minimal dependencies).  Or, if they removed their specific package dependencies, and depened on Microsoft.VisualStudio.SDK instead, then that would be easier for us to override.

So, unfortunately there might not be a simple solution ☹️

jgonz120: Hmm okay I upgraded to the latest stable version of the platform tool and updated the MicrosoftBuildVersion to 17.11.4

jgonz120: Oh wait, you just said we can't upgrade Microsoft.Build.Framework 😔

jgonz120: Okay so leaving it back to that preview version and added a PackageReference to Microsoft.Build.Framework and Microsoft.NET.StringTools, so for now it works fine. 

I'll want to upgrade this package eventually, but the new version depends on `Microsoft.VisualStudio.ScriptedHost` which isn't hosted on a public feed right now. If we can get that published to a public feed that we can revisit this issue then.

zivkan: Since our dev branch is currently targeting VS 17.12 & .NET 9.0.100, according to this table we're ok with MSBuild 17.11: https://learn.microsoft.com/en-us/dotnet/core/porting/versioning-sdk-msbuild-vs#targeting-and-support-rules

But in the future we're either going to need to improve our build scripts so that `src\NuGet.Core\*` use the .NET approved version of MSBuild, even when our VS code targets a later MSBuild.

Or we're going to have to be a lot more insistent to our VS partners and get them to change their package dependencies, or explain to them why we're blocked and can't integrate their features.

zivkan (approved):

PR #6014, opened 2024-09-06
Add ReadmeFileUrl field to IPackageSearchMetadata

jgonz120:
# Bug

## Description
Adds the ReadmeFileUrl field to IPackageSearchMetadata and implements the ability to get the URL for local packages. Future PRs will include implementations for getting the URL for remote READMEs and downloading them.  

## PR Checklist

- [X] ~~Added tests~~

### DIFF
--- a/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/TransitivePackageSearchMetadata.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/TransitivePackageSearchMetadata.cs
@@ -61,7 +61,6 @@ namespace NuGet.VisualStudio.Internal.Contracts
 
         public IEnumerable<PackageVulnerabilityMetadata> Vulnerabilities => _packageSearchMetadata.Vulnerabilities;
 
-
         private readonly IPackageSearchMetadata _packageSearchMetadata;
 
         public TransitivePackageSearchMetadata(IPackageSearchMetadata package, IReadOnlyCollection<PackageIdentity> transitiveOrigins)
--- a/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/TransitivePackageSearchMetadata.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/TransitivePackageSearchMetadata.cs
@@ -33,6 +33,8 @@ namespace NuGet.VisualStudio.Internal.Contracts
 
         public Uri ReadmeUrl => _packageSearchMetadata.ReadmeUrl;
 
+        public Uri ReadmeFileUrl => _packageSearchMetadata.ReadmeFileUrl;
+
         public Uri ReportAbuseUrl => _packageSearchMetadata.ReportAbuseUrl;
 
         public Uri PackageDetailsUrl => _packageSearchMetadata.PackageDetailsUrl;
@@ -58,6 +60,7 @@ namespace NuGet.VisualStudio.Internal.Contracts
         public LicenseMetadata LicenseMetadata => _packageSearchMetadata.LicenseMetadata;
 
         public IEnumerable<PackageVulnerabilityMetadata> Vulnerabilities => _packageSearchMetadata.Vulnerabilities;
+
 
         private readonly IPackageSearchMetadata _packageSearchMetadata;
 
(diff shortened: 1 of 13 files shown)


test/TestUtilities/Test.Utility/Protocol/V3PackageSearchMetadataFixture.cs
```diff
@@ -74,6 +74,8 @@ public MockPackageSearchMetadata()
 
             public Uri ReadmeUrl => null;
 
+            public Uri ReadmeFileUrl => null;
```
donnie-msft: I was thinking there should be a few more tests that need this new property added to existing tests.

Some new tests may be needed as well, like this Local metadata for IconUri, for example: checked:https://github.com/NuGet/NuGet.Client/blob/e2f077a9a062e8190b3482e29786a758182b33fb/test/NuGet.Core.Tests/NuGet.Protocol.Tests/LocalPackageSearchMetadataTests.cs#L62

jgonz120: I was planning on adding the unit tests as part of another PR, do you think I should do it now?

jgonz120: Meh, added a unit test to match the one you linked. It looks like there's only one test class for LocalPackageSearchMetadata.

donnie-msft (approved):

src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/TransitivePackageSearchMetadata.cs
```diff
@@ -59,6 +61,7 @@ public class TransitivePackageSearchMetadata : IPackageSearchMetadata
 
         public IEnumerable<PackageVulnerabilityMetadata> Vulnerabilities => _packageSearchMetadata.Vulnerabilities;
 
+
```
zivkan: nitpick: unnecessary whitespace line

src/NuGet.Core/NuGet.Protocol/Model/PackageSearchMetadata.cs
```diff
@@ -106,6 +106,9 @@ public string Owners
         [JsonConverter(typeof(SafeUriConverter))]
         public Uri ReadmeUrl { get; private set; }
 
+        [JsonIgnore]
+        public Uri ReadmeFileUrl { get; private set; }
```
zivkan: Why ignore this from JSON?   Don't we want HTTP sources to be able to provide the raw markdown file?

The PR description also doesn't explain.

src/NuGet.Core/NuGet.Protocol/Model/IPackageSearchMetadata.cs
```diff
@@ -24,6 +24,7 @@ public interface IPackageSearchMetadata
         Uri LicenseUrl { get; }
         Uri ProjectUrl { get; }
         Uri ReadmeUrl { get; }
+        Uri ReadmeFileUrl { get; }
```
zivkan:
```suggestion
        string ReadmeFileUrl { get; }
```

It's a perf mistake to use `System.Uri`.  It's expensive to calculate, and `HttpClient` has overloads that allow you to pass the URL as a string. Maybe it parses as an Uri internally, I don't know, but it also doesn't matter. This property isn't always going to be used, so we shouldn't pay the perf cost to force it to always be parsed.

I'd love to change the other ones from `Uri` to `string`, but that would be a breaking change in a package we ship. But we can avoid making the same mistake in new APIs.

src/NuGet.Core/NuGet.Protocol/Model/PackageSearchMetadata.cs
jgonz120: I wanted to focus on the local package reading for this PR. In my head I've been keeping them separate, then in a future PR once all the remote reading stuff is added I was going to update this field.

Nigusu-Allehu (approved):

src/NuGet.Core/NuGet.Protocol/Model/PackageSearchMetadata.cs
jgonz120: Updated~

PR #6016, opened 2024-09-07
PackageManagementFormat only updates settings when it has a value

### LINKED ISSUE #13773: Saving PackageManagementFormat throws Nullable object must have a value.
### NuGet Product Used

Visual Studio Package Management UI, Visual Studio Package Manager Console

### Product Version

any

### Worked before?

no

### Impact

Other

### Repro Steps & Context

When calling `ApplyChanges` on `PackageManagementFormat` settings , if the nullable property `_showDialogValue` is not set, the attempt to save settings to disk will fail.

1. Create a PackageManagementFormat
2. Do not call a `get` method for `Enabled` property
3. Call `ApplyChanges` and observe the InvalidOperationException

### Verbose Logs

_No response_

donnie-msft:
# Bug

Fixes:  https://github.com/NuGet/Home/issues/13773

## Description

While testing VS Settings, I found that if I do not invoke getters on properties in this class, then call `ApplyChanges` two issues can occur:

- `Enabled` property: an exception was thrown due to a nullable property being unset
- `SelectedPackageManagementFormat` property: a "-1" was written to the nuget.config. This is never a valid value and should never be written to a nuget.config‼️ 

Check properties for valid values prior to updating settings.
Tests ensure that if no values are set, nothing is updated in settings, even though the call to Save still occurs. I didn't want to change any other workflow that may depend on this call to save to disk.

## PR Checklist

- [x] Added tests

### DIFF
--- a/src/NuGet.Clients/NuGet.VisualStudio.Common/PackageManagementFormat.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Common/PackageManagementFormat.cs
@@ -102,10 +102,17 @@ namespace NuGet.VisualStudio
 
         public void ApplyChanges()
         {
-            _settings.AddOrUpdate(ConfigurationConstants.PackageManagementSection,
-                new AddItem(ConfigurationConstants.DefaultPackageManagementFormatKey, _selectedPackageFormat.ToString(CultureInfo.InvariantCulture)));
-            _settings.AddOrUpdate(ConfigurationConstants.PackageManagementSection,
-                new AddItem(ConfigurationConstants.DoNotShowPackageManagementSelectionKey, _showDialogValue.Value.ToString(CultureInfo.InvariantCulture)));
+            if (_selectedPackageFormat != -1)
+            {
+                _settings.AddOrUpdate(ConfigurationConstants.PackageManagementSection,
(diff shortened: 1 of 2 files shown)


src/NuGet.Clients/NuGet.VisualStudio.Common/PackageManagementFormat.cs
```diff
@@ -102,10 +102,17 @@ public int SelectedPackageManagementFormat
 
         public void ApplyChanges()
         {
-            _settings.AddOrUpdate(ConfigurationConstants.PackageManagementSection,
-                new AddItem(ConfigurationConstants.DefaultPackageManagementFormatKey, _selectedPackageFormat.ToString(CultureInfo.InvariantCulture)));
-            _settings.AddOrUpdate(ConfigurationConstants.PackageManagementSection,
-                new AddItem(ConfigurationConstants.DoNotShowPackageManagementSelectionKey, _showDialogValue.Value.ToString(CultureInfo.InvariantCulture)));
+            if (_selectedPackageFormat != -1)
```
jebriede, line 105: To guard against any future code changes that could use another sentinel value (like -2, etc), you could make this:
```suggestion
            if (_selectedPackageFormat > -1)
```

jebriede (approved):
Good update! Approving with suggestion.

zivkan (approved):

PR #6019, opened 2024-09-09
Revert "Switch to NoWarn for XML errors"

jgonz120:
Reverts NuGet/NuGet.Client#6017

### DIFF
--- a/build/common.targets
+++ b/build/common.targets
@@ -40,7 +40,7 @@
   </PropertyGroup>
 
   <PropertyGroup Condition=" '$(PackProject)' != 'true' ">
-    <NoWarn>$(NoWarn);CS0419;CS1570;CS1572;CS1573;CS1574;CS1584;CS1587;CS1591</NoWarn>
+    <WarningsNotAsErrors>$(WarningsNotAsErrors);CS0419;CS1570;CS1572;CS1573;CS1574;CS1584;CS1587;CS1591</WarningsNotAsErrors>
   </PropertyGroup>
 
   <ItemGroup Condition=" '$(PackProject)' == 'true' ">


jeffkl:
What's the plan here?  Without this change our build logs 12,698 warnings

build/common.targets
```diff
@@ -40,7 +40,7 @@
   </PropertyGroup>
 
   <PropertyGroup Condition=" '$(PackProject)' != 'true' ">
-    <NoWarn>$(NoWarn);CS0419;CS1570;CS1572;CS1573;CS1574;CS1584;CS1587;CS1591</NoWarn>
+    <WarningsNotAsErrors>$(WarningsNotAsErrors);CS0419;CS1570;CS1572;CS1573;CS1574;CS1584;CS1587;CS1591</WarningsNotAsErrors>
```
zivkan, line 43: I asked Jonatan to reopen a PR, because the previous one was merged in less than 2 hours after opening, before I had a chance to review or comment. It was merged before 7am in my time zone, so many people here wouldn't even be awake, let alone working and reviewing PRs at that time.

These warnings are only raised on public members on public types. So, it's much less impactful than all members on all types. If people find it annoying to write XML doc, that could be considered a good thing, since they'll need to be more deliberate about public APIs, even if it's for an assembly that is not published in a package.

So, I'd like a team discussion if we consider this desirable for assemblies that do not get published as packages?

Since there're not packages, nobody can call these APIs except for other projects in this same git repository. This means we can safely refactor them, change API signature, etc, without worrying about breaking changes.

Also, if people just write low effort XMLDocs ("GetValue returns the value"), then it's just noise.

On the other hand, there's so much complexity in NuGet, and XMLDoc can be used to help us share knowledge about how APIs should work, about edge cases, about proper usage. But only if the XMLDoc is written in a way to share that knowledge.

For some reason, the API that always sticks out to me the most is `PackageReferenceProject.GetInstalledAndTransitivePackagesAsync`. This API was created initially for a PM UI feature, but it's also used now for `INuGetProjectService.GetInstalledPackagesAsync()`, which any other VS component can call (because the interface is defined in Nuget.VisualStudio.Contracts, which is a package we publish and support). I feel like people who work on `PackageReferenceProject.GetInstalledAndTransitivePackagesAsync` are usually not aware of the multiple different use-cases of the API, and make design decisions that harm performance of the non-PM UI scenarios.  I think a comment reminding us would help, and if the comment is an XMLDoc comment, then this remark can flow to other classes which inherit/override the method, plus the remark will be visible in intellisense's hover tooltip.  However, if this API is a one-off, then we could just add the XMLDoc without having the compiler raise warnings.

Other examples are pretty much everything to do with Package Manager Console. There hasn't been any significant work on it in the years I've been on the team. When needing to investigate bugs, I come across behaviour that I find surprising, and I can't tell if it's by design or a bug. I don't know if classes are a NuGet abstraction, or something that's needed by the powershell host. Or there's a long method, it can take considerable time to read and figure out what it's really doing. Hopefully the method name already says, but long methods usually have a lot of complexity that's not explained.

Obviously, keeping these messages doesn't magically solve these existing APIs that are missing XMLDoc comments. But it helps remind people writing new public methods, or modifying existing public methods, to add XMLDoc to help people in the future. And maybe be more mindful about public APIs, and make more stuff internal.

So, this is what I'd like time to think about myself, but also the rest of the team to think about.  Do warning messages about missing XMLDoc help us write better public APIs and improve knowledge about NuGet's software architecture?

zivkan, line 43: If the team decision from the other comment is that we don't think that XMLDoc will help us improve code quality, then this comment should be ignored (and this PR closed without merging). Otherwise, we can consider different wants to deal with the warnings.

12,000+ warnings is a LOT, and then everyone will have to ignore warnings because it'll be impossible to notice when creating a PR is one more is added.  On the other hand, since we use TreatWarningsAsErrors, there aren't really other scenarios where an actual warning will be raised.

Another option is to use inline suppressions to suppress all the warnings from existing violations. That is a LOT of suppressions, but it will make it really obvious when a new PR adds a public API that doesn't have XMLDoc.

Yet another option is to try editorconfig to [change the severity to suggestion](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-options#severity-level). Similar to nuget.config, [code analysis config reads from multiple locations](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-files#precedence), so we can keep these XML doc CA codes as warning for packages we publish, and downgrade to suggestion for other projects.

PR #6028, opened 2024-09-13
Properly handle missing package versions in new dependency resolver

### LINKED ISSUE #13788: New dependency resolver does not properly handle missing package versions when using CPM
### NuGet Product Used

dotnet.exe, MSBuild.exe

### Product Version

6.12

### Worked before?

Yes

### Impact

### Repro Steps & Context

If you have two projects where one references the other and you're missing a PackageVersion for CPM, then the new dependency resolver does not handle this correctly.

```
ProjectA -> ProjectB -> PackageX (missing a PackageVersion)
```

During restore of each project, we verify that CPM is configured properly.  For ProjectB, we correctly identify this.  When resolving the graph of ProjectA, none of its package references are missing any versions.  I'm not sure how the legacy dependency resolver handles this.  Maybe it has to do with ordering?

### Verbose Logs

```shell
Directory.Packages.props

<Project>
  <PropertyGroup>
    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
  </PropertyGroup>
</Project>


ClassLibrary1.csproj
```xml
  <PropertyGroup>
    <TargetFramework>net6.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <ProjectReference Include="..\ClassLibrary2\ClassLibrary2.csproj" />
  </ItemGroup>
</Project>
```

ClassLibrary2.csproj
```xml

  <PropertyGroup>
    <TargetFramework>net6.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="EnvironmentAbstractions" />
  </ItemGroup>
</Project>

```

New resolver output:
```
  ClassLibrary2.csproj : error NU1008: Projects that use central package version management should not define the version on the PackageReference items but on the PackageVersion items: EnvironmentAbstractions.
  ClassLibrary1.csproj : error NU1101: Unable to find package EnvironmentAbstractions. No packages exist with this id in source(s): nuget.org 
```

Legacy resolver output:
```
  ClassLibrary2.csproj : error NU1008: Projects that use central package version management should not define the version on the PackageReference items but on the PackageVersion items: EnvironmentAbstractions.
```
```

### PR DESCRIPTION
**Pull request by jeffkl**
# Bug

Fixes: https://github.com/NuGet/Home/issues/13788

## Description
During restore, NuGet verifies that CPM requirements are met by verifying all packages have versions.  However, when one project references another and the referenced project has a missing package, NuGet logs an error for the referenced project but restores the other.  It restores this project by skipping the dependency. 

This change updates the new dependency resolver to ignore packages with missing versions.

## PR Checklist

- [x] Added tests

### DIFF
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
@@ -205,12 +205,6 @@ namespace NuGet.Commands
                     IReadOnlyDictionary<LibraryDependencyIndex, VersionRange> currentOverrides = importRefItem.VersionOverrides!;
                     bool directPackageReferenceFromRootProject = importRefItem.IsDirectPackageReferenceFromRootProject;
 
-                    // Packages with missing versions should not be added to the graph
-                    if (currentRef.LibraryRange.VersionRange == null)
-                    {
-                        continue;
-                    }
-
                     if (!findLibraryEntryCache.TryGetValue(currentRefRangeIndex, out Task<FindLibraryEntryResult>? refItemTask))
                     {
                         Debug.Fail("This should not happen");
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs [hunk 1]
```diff
@@ -650,6 +650,12 @@ async static (state) =>
                     for (int i = 0; i < refItemResult.Item.Data.Dependencies.Count; i++)
                     {
                         var dep = refItemResult.Item.Data.Dependencies[i];
+                        // Packages with missing versions should not be added to the graph
+                        if (dep.LibraryRange.VersionRange == null)
```
nkolev92, line 654: Do we need to make sure this isn't applied to the top-level project?

jeffkl: For the top-level project, this will only be null if CPM is enabled and a version is missing.  However, before this resolver is called, an error is logged and restore doesn't happen.  This is only run if a transitive project isn't properly configured.

nkolev92 (approved):

PR #6029, opened 2024-09-13
Added the ReadmePreviewViewModel & edited NuGetPackageFileService

jgonz120:
… readme

# Bug

## Description
Adds a new ViewModel for getting the README contents, leveraging the NuGetPackageFileService to get the file from the package. 

## PR Checklist

- [X] Added tests

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
@@ -6,14 +6,16 @@ using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.VisualStudio.Threading;
-using NuGet.VisualStudio;
 using NuGet.VisualStudio.Internal.Contracts;
 
 namespace NuGet.PackageManagement.UI.ViewModels
 {
     public sealed class ReadmePreviewViewModel : ViewModelBase
     {
+        private bool _canDetermineReadmeDefined;
+        private bool _errorLoadingReadme;
         private INuGetPackageFileService _packageFileService;
+        private string _rawReadme;
 
         public ReadmePreviewViewModel(INuGetPackageFileService packageFileService)
         {
@@ -23,21 +25,18 @@ namespace NuGet.PackageManagement.UI.ViewModels
             _rawReadme = string.Empty;
         }
 
-        private bool _errorLoadingReadme;
         public bool ErrorLoadingReadme
         {
             get => _errorLoadingReadme;
             set => SetAndRaisePropertyChanged(ref _errorLoadingReadme, value);
         }
 
-        private string _rawReadme;
         public string ReadmeMarkdown
         {
             get => _rawReadme;
             set => SetAndRaisePropertyChanged(ref _rawReadme, value);
         }
 
-        private bool _canDetermineReadmeDefined;
         public bool CanDetermineReadmeDefined
         {
             get => _canDetermineReadmeDefined;
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
@@ -1,0 +1,89 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.Threading;
+using NuGet.VisualStudio;
+using NuGet.VisualStudio.Internal.Contracts;
+
+namespace NuGet.PackageManagement.UI.ViewModels
+{
+    public sealed class ReadmePreviewViewModel : ViewModelBase
+    {
+        private INuGetPackageFileService _packageFileService;
+
+        public ReadmePreviewViewModel(INuGetPackageFileService packageFileService)
+        {
+            _errorLoadingReadme = false;
+            _rawReadme = string.Empty;
+            _packageFileService = packageFileService;
+        }
+
+        private bool _errorLoadingReadme = false;
+
+        public bool ErrorLoadingReadme
+        {
+            get => _errorLoadingReadme;
+            set
+            {
+                SetAndRaisePropertyChanged(ref _errorLoadingReadme, value);
+            }
+        }
+
+        private string _rawReadme = string.Empty;
+
+        public string ReadmeMarkdown
+        {
+            get => _rawReadme;
+            set
+            {
+                SetAndRaisePropertyChanged(ref _rawReadme, value);
+            }
+        }
+
+        private bool _canDetermineReadmeDefined = true;
+
+        public bool CanDetermineReadmeDefined
+        {
+            get => _canDetermineReadmeDefined;
+            set
+            {
+                SetAndRaisePropertyChanged(ref _canDetermineReadmeDefined, value);
+            }
+        }
+
+        public async Task LoadReadmeAsync(string rawReadmeUrl, CancellationToken cancellationToken)
+        {
+            var newReadmeValue = string.Empty;
+            var isErrorWithReadme = false;
+            bool canDetermineReadmeDefined = false;
+
+            if (!string.IsNullOrWhiteSpace(rawReadmeUrl))
+            {
+
+                await TaskScheduler.Default;
+
+                var readmeStream = await _packageFileService.GetReadmeAsync(new Uri(rawReadmeUrl), cancellationToken);
+                if (readmeStream is not null)
+                {
+                    using StreamReader streamReader = new StreamReader(readmeStream);
+                    var readme = await streamReader.ReadToEndAsync();
+                    if (!string.IsNullOrWhiteSpace(readme))
+                    {
+                        isErrorWithReadme = false;
+                        canDetermineReadmeDefined = true;
+                        newReadmeValue = readme;
+                    }
+                }
+                await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(cancellationToken);
+            }
+
+            ErrorLoadingReadme = isErrorWithReadme;
+            ReadmeMarkdown = newReadmeValue;
+            CanDetermineReadmeDefined = canDetermineReadmeDefined;
+        }
+    }
+}
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
@@ -46,32 +46,31 @@ namespace NuGet.PackageManagement.UI.ViewModels
 
         public async Task LoadReadmeAsync(string rawReadmeUrl, CancellationToken cancellationToken)
         {
-            var newReadmeValue = string.Empty;
-            var isErrorWithReadme = false;
-            bool canDetermineReadmeDefined = false;
+            ReadmeMarkdown = string.Empty;
+            ErrorLoadingReadme = false;
+            CanDetermineReadmeDefined = false;
 
-            if (!string.IsNullOrWhiteSpace(rawReadmeUrl))
+            if (string.IsNullOrWhiteSpace(rawReadmeUrl))
             {
-                await TaskScheduler.Default;
-
-                var readmeStream = await _packageFileService.GetReadmeAsync(new Uri(rawReadmeUrl), cancellationToken);
-                if (readmeStream is not null)
-                {
-                    using StreamReader streamReader = new StreamReader(readmeStream);
-                    var readme = await streamReader.ReadToEndAsync();
-                    if (!string.IsNullOrWhiteSpace(readme))
-                    {
-                        isErrorWithReadme = false;
-                        canDetermineReadmeDefined = true;
-                        newReadmeValue = readme;
-                    }
-                }
-                await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(cancellationToken);
+                return;
             }
 
-            ErrorLoadingReadme = isErrorWithReadme;
-            ReadmeMarkdown = newReadmeValue;
-            CanDetermineReadmeDefined = canDetermineReadmeDefined;
+            await TaskScheduler.Default;
+
+            var readmeStream = await _packageFileService.GetReadmeAsync(new Uri(rawReadmeUrl), cancellationToken);
+            if (readmeStream is null)
+            {
+                return;
+            }
+
+            using StreamReader streamReader = new StreamReader(readmeStream);
+            var readme = await streamReader.ReadToEndAsync();
+            if (!string.IsNullOrWhiteSpace(readme))
+            {
+                ReadmeMarkdown = readme;
+                ErrorLoadingReadme = false;
+                CanDetermineReadmeDefined = true;
+            }
         }
     }
 }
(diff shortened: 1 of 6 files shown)


jgonz120:
@nkolev92 this is going to a feature branch so I don't think there's a worry about merging.

test/NuGet.Clients.Tests/NuGet.PackageManagement.UI.Test/ViewModels/ReadmePreviewViewModelTests.cs
```diff
@@ -0,0 +1,82 @@
...
+using NuGet.VisualStudio.Internal.Contracts;
+using Xunit;
+
+namespace NuGet.PackageManagement.UI.Test.ViewModels
+{
+    [Collection(MockedVS.Collection)]
+    public class ReadmePreviewViewModelTests
+    {
+        [Fact]
+        public void ReadmePreviewViewModel_WithNullPackageFileService_ThrowsArgumentNullException()
```
jebriede: Or Ctor

```suggestion
        public void Constructor_WithNullPackageFileService_ThrowsArgumentNullException()
```

test/NuGet.Clients.Tests/NuGet.PackageManagement.UI.Test/ViewModels/ReadmePreviewViewModelTests.cs
```diff
@@ -0,0 +1,82 @@
...
+            Assert.False(target.ErrorLoadingReadme);
+            Assert.Equal(string.Empty, target.ReadmeMarkdown);
+            Assert.True(target.CanDetermineReadmeDefined);
+        }
+
+        [Theory]
+        [InlineData(null)]
+        [InlineData("")]
+        [InlineData(" ")]
+        public async Task ReadmePreviewViewModel_LoadReadmeAsync_WithoutReadmeUrl_NoReadmeReturned(string readmeUrl)
```
jebriede: We can leave the ReadmePreviewViewModel part off all the tests since this test class already describes the class being tested. See https://github.com/NuGet/NuGet.Client/blob/dev/docs/coding-guidelines.md#unit-test-method-naming

```suggestion
        public async Task LoadReadmeAsync_WithoutReadmeUrl_NoReadmeReturned(string readmeUrl)
```

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        public async Task LoadReadmeAsync(string rawReadmeUrl, CancellationToken cancellationToken)
+        {
+            var newReadmeValue = string.Empty;
+            var isErrorWithReadme = false;
+            bool canDetermineReadmeDefined = false;
+
+            if (!string.IsNullOrWhiteSpace(rawReadmeUrl))
+            {
+
```
jebriede, line 56: nit: remove blank line.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
```diff
@@ -0,0 +1,90 @@
...
+        public bool CanDetermineReadmeDefined
+        {
+            get => _canDetermineReadmeDefined;
+            set
+            {
+                SetAndRaisePropertyChanged(ref _canDetermineReadmeDefined, value);
+            }
+        }
+
+        public async Task LoadReadmeAsync(string rawReadmeUrl, CancellationToken cancellationToken)
```
jebriede, line 46: Question: This isn't being used yet, except for in tests, correct?

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
```diff
@@ -0,0 +1,90 @@
...
+                    using StreamReader streamReader = new StreamReader(readmeStream);
+                    var readme = await streamReader.ReadToEndAsync();
+                    if (!string.IsNullOrWhiteSpace(readme))
+                    {
+                        isErrorWithReadme = false;
+                        canDetermineReadmeDefined = true;
+                        newReadmeValue = readme;
+                    }
+                }
+                await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(cancellationToken);
```
jebriede: We shouldn't switch to the main thread at the end of this method.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        private bool _errorLoadingReadme;
+
+        public bool ErrorLoadingReadme
+        {
+            get => _errorLoadingReadme;
+            set
+            {
+                SetAndRaisePropertyChanged(ref _errorLoadingReadme, value);
+            }
```
jebriede: Style suggestion: More concise

```suggestion
            set => SetAndRaisePropertyChanged(ref _errorLoadingReadme, value);
```

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
```diff
@@ -0,0 +1,90 @@
...
+        public ReadmePreviewViewModel(INuGetPackageFileService packageFileService)
+        {
+            _packageFileService = packageFileService ?? throw new ArgumentNullException(nameof(packageFileService));
+            _errorLoadingReadme = false;
+            _canDetermineReadmeDefined = true;
+            _rawReadme = string.Empty;
+        }
+
+        private bool _errorLoadingReadme;
+
```
jebriede: Nit: remove blank line.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            get => _errorLoadingReadme;
+            set
+            {
+                SetAndRaisePropertyChanged(ref _errorLoadingReadme, value);
+            }
+        }
+
+        private string _rawReadme;
+
```
jebriede: Nit: remove blank line.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            get => _rawReadme;
+            set
+            {
+                SetAndRaisePropertyChanged(ref _rawReadme, value);
+            }
+        }
+
+        private bool _canDetermineReadmeDefined;
+
```
jebriede: Nit: remove blank line.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        public async Task LoadReadmeAsync(string rawReadmeUrl, CancellationToken cancellationToken)
+        {
+            var newReadmeValue = string.Empty;
+            var isErrorWithReadme = false;
+            bool canDetermineReadmeDefined = false;
+
+            if (!string.IsNullOrWhiteSpace(rawReadmeUrl))
+            {
```
jebriede: To reduce nested ifs we can take basic input validation checks like this and invert it to if (string.IsNullOrWhiteSpace(rawReadmeUrl)), set some defaults, and return.

jebriede (changes requested):

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
jgonz120: Yes, the next PR will be the UX where it'll be used

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        public ReadmePreviewViewModel(INuGetPackageFileService packageFileService)
+        {
+            _packageFileService = packageFileService ?? throw new ArgumentNullException(nameof(packageFileService));
+            _errorLoadingReadme = false;
+            _canDetermineReadmeDefined = true;
+            _rawReadme = string.Empty;
+        }
+
+        private bool _errorLoadingReadme;
```
donnie-msft: Our coding guidelines suggest separating groups of fields from properties.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
donnie-msft: I don't think updating properties on a background thread will work with WPF.
Jonatan, what happens if you remove the switch to main thread? I'm guessing either no UI update or an exception?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetPackageFileService.cs
```diff
@@ -131,6 +131,23 @@ public NuGetPackageFileService(IServiceBroker serviceBroker, INuGetTelemetryProv
             return stream;
         }
 
+        public async ValueTask<Stream?> GetReadmeAsync(Uri readmeUri, CancellationToken cancellationToken)
+        {
+            Assumes.NotNull(readmeUri);
+
+            Stream? stream;
+            if (IsEmbeddedUri(readmeUri))
```
donnie-msft, line 172: I recently changed the Icon algorithm to check for a local file first, then a remote file.
Do you think this is a good opportunity to do that for README? That is, check for the md file in GPF first, and only if not found, go to the remote?

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Services/NuGetPackageFileServiceTests.cs
```diff
@@ -337,6 +337,37 @@ public async Task AddLicenseToCache_WhenMissing_IsNotFound()
...
+                var packageFileService = new NuGetPackageFileService(
+                        default(ServiceActivationOptions),
+                        Mock.Of<IServiceBroker>(),
+                        new AuthorizationServiceClient(Mock.Of<IAuthorizationService>()),
+                        _telemetryProvider.Object);
+
+                Stream readmeStream = await packageFileService.GetReadmeAsync(builder.Uri, CancellationToken.None);
+
+                Assert.NotNull(readmeStream);
+                Assert.Equal(StreamContents.Length, readmeStream.Length);
```
donnie-msft: nit: it bothers me that the file content states it is an image, even though this is a readme 😆

donnie-msft (approved):
Thanks for making smaller PRs for this effort! I would reword the title since it also updates the file service and not just the VM. Approving with suggestions.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
jgonz120: From what I remember I got an exception when I tried to set the view model properties without switching to the main thread. I can remove it and we can see what happens in the next pr, it looks like the unit test passes either way

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Services/NuGetPackageFileServiceTests.cs
jgonz120: Did you mean the comments in the CreateDummyPackage? 😅

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetPackageFileService.cs
jgonz120: I don't think that will work for the Readme, I want to decide which source it's being loaded from. We don't want to display README in the browse tab if they aren't available from the remote source.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
jebriede: A better way to handle this is to use the [Dispatcher](https://learn.microsoft.com/en-us/dotnet/api/system.windows.threading.dispatcher?view=windowsdesktop-8.0) to check if we're on the main thread or to marshal to the main thread.

Explicitly switching to the main thread is expensive and error-prone because it forces this background operation onto the main thread where it will remain, even when returning to the caller.

See: https://stackoverflow.com/questions/590590/making-%20sure-onpropertychanged-is-called-on-ui-thread-in-mvvm-wpf-app

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Services/NuGetPackageFileServiceTests.cs
donnie-msft: No, I mean StreamContents which is "hi I'm an image" or something.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetPackageFileService.cs
donnie-msft: I quickly re-read the draft spec and didn't see anything obvious indicating why READMEs should only be downloaded and not read from disk. https://github.com/NuGet/Home/pull/13200/files?short_path=6c7eb8e#diff-6c7eb8e1ec6b63a9782e440a235ee32bc0d10f4b7ec8a718f3d24a53660f679c

Could you elaborate on why "it won't work"? You can see the source that populated the GPF via the .nupkg.metadata file. 

> We don't want to display README in the browse tab if they aren't available from the remote source.
Great thing to explain the spec. Or if it's already there, please give me a pointer to that.

jebriede (approved):

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetPackageFileService.cs
jgonz120: To clarify the requirement. We only want to display README from local storage in the installed tab. That way if there isn't a remote source setup with README exposed, they won't have as much of a inconsistent README experience in the browse tab. 

So I feel like this solution won't work because it's leaving it up to the PackageFileService to determine where to pull the README from, but it doesn't have the context if we're viewing the details from the Browse or Installed tabs

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Services/NuGetPackageFileServiceTests.cs
jgonz120: whoops, fixed

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/NuGetPackageFileService.cs
donnie-msft: I'm not seeing this explicitly in the Spec. I suggest updating that. I'd like to discuss the design, and the spec would be the best place for that.

PR #6030, opened 2024-09-16
Fix null reference in PM UI for project references without framework

### LINKED ISSUE #13737: ProjectReference causing PM UI to error with "Value cannot be null. Parameter name: frameworkIdentifier"
Framework is null on projectReference resulting in ArgumentNullException: "Value cannot be null. Parameter name: frameworkIdentifier"

Reproes only on the Installed Tab (Browse tab works) on a ConsoleApp which has a ProjectReference to a ClassLibrary in the same solution.

Fixed by deleting bin/obj folders.

image

ConsoleApp2.csproj snippet:
```
 <ItemGroup>
    <PackageReference Include="Microsoft.Office.Interop.Excel">
      <Version>15.0.4795.1001</Version>
    </PackageReference>
  </ItemGroup>
  <ItemGroup>
    <ProjectReference Include="..\ClassLibrary3\ClassLibrary3.csproj">
      <Project>{63D2D8D1-16B3-4D98-9421-259E1B7A4301}</Project>
      <Name>ClassLibrary3</Name>
    </ProjectReference>
  </ItemGroup>
```

image

> NuGet.Frameworks.NuGetFramework.NuGetFramework(frameworkIdentifier = null, frameworkVersion = {System.Version}, profile = "", platform = "", platformVersion = {System.Version}) Line 59
> 	at C:\NuGet.Client\src\NuGet.Core\NuGet.Frameworks\NuGetFramework.cs(59)
> NuGet.Frameworks.NuGetFramework.NuGetFramework(frameworkIdentifier = null, frameworkVersion = {System.Version}, platform = "", platformVersion = {System.Version}) Line 53
> 	at C:\NuGet.Client\src\NuGet.Core\NuGet.Frameworks\NuGetFramework.cs(53)
> NuGet.Frameworks.NuGetFramework.NuGetFramework(framework = null, version = {System.Version}) Line 35
> 	at C:\NuGet.Client\src\NuGet.Core\NuGet.Frameworks\NuGetFramework.cs(35)
> NuGet.Frameworks.NuGetFramework.NuGetFramework(framework = null) Line 30
> 	at C:\NuGet.Client\src\NuGet.Core\NuGet.Frameworks\NuGetFramework.cs(30)
> NuGet.PackageManagement.VisualStudio.PackageReferenceProject<System.Collections.Generic.Dictionary<string, NuGet.PackageManagement.VisualStudio.Utility.ProjectInstalledPackage>, System.Collections.Generic.KeyValuePair<string, NuGet.PackageManagement.VisualStudio.Utility.ProjectInstalledPackage>>.GetInstalledAndTransitivePackagesAsync(includeTransitivePackages = true, includeTransitiveOrigins = true, token = IsCancellationRequested = false) Line 229
> 	at C:\NuGet.Client\src\NuGet.Clients\NuGet.PackageManagement.VisualStudio\Projects\PackageReferenceProject.cs(229)
> [Async Call Stack]
> [Async] NuGet.PackageManagement.VisualStudio.PackageReferenceProject<System.Collections.Generic.Dictionary<string, NuGet.PackageManagement.VisualStudio.Utility.ProjectInstalledPackage>, System.Collections.Generic.KeyValuePair<string, NuGet.PackageManagement.VisualStudio.Utility.ProjectInstalledPackage>>.GetInstalledAndTransitivePackagesAsync(includeTransitiveOrigins = true, token = IsCancellationRequested = false) Line 120
> 	at C:\NuGet.Client\src\NuGet.Clients\NuGet.PackageManagement.VisualStudio\Projects\PackageReferenceProject.cs(120)
> [External Code]
> [Async] NuGet.PackageManagement.VisualStudio.NuGetProjectManagerService.GetInstalledAndTransitivePackagesAsync(projectIds = {string[1]}, includeTransitiveOrigins = true, cancellationToken = IsCancellationRequested = false) Line 190
> 	at C:\NuGet.Client\src\NuGet.Clients\NuGet.PackageManagement.VisualStudio\Services\NuGetProjectManagerService.cs(190)
> [External Code]
> [Async] NuGet.PackageManagement.VisualStudio.IProjectContextInfoExtensions.GetInstalledAndTransitivePackagesAsync(projectContextInfo = {NuGet.VisualStudio.Internal.Contracts.ProjectContextInfo}, serviceBroker = "Process view", includeTransitiveOrigins = true, cancellationToken = IsCancellationRequested = false) Line 75
> 	at C:\NuGet.Client\src\NuGet.Clients\NuGet.PackageManagement.VisualStudio\Common\IProjectContextInfoExtensions.cs(75)
> [External Code]
> [Async] NuGet.PackageManagement.VisualStudio.PackageCollection.GetInstalledAndTransitivePackagesAsync(serviceBroker = "Process view", projectContextInfos = {NuGet.VisualStudio.Internal.Contracts.IProjectContextInfo[1]}, includeTransitiveOrigins = true, cancellationToken = IsCancellationRequested = false) Line 74
> 	at C:\NuGet.Client\src\NuGet.Clients\NuGet.PackageManagement.VisualStudio\Common\PackageCollection.cs(74)
> [Async] NuGet.PackageManagement.VisualStudio.NuGetPackageSearchService.CreatePackageFeedAsync(projectContextInfos = {NuGet.VisualStudio.Internal.Contracts.IProjectContextInfo[1]}, targetFrameworks = {string[1]}, itemFilter = Installed, isSolution = false, recommendPackages = false, sourceRepositories = Count = 1, cancellationToken = IsCancellationRequested = false) Line 453
> 	at C:\NuGet.Client\src\NuGet.Clients\NuGet.PackageManagement.VisualStudio\Services\NuGetPackageSearchService.cs(453)
> [Async] NuGet.PackageManagement.VisualStudio.NuGetPackageSearchService.SearchAsync(projectContextInfos = {NuGet.VisualStudio.Internal.Contracts.IProjectContextInfo[1]}, packageSources = {NuGet.VisualStudio.Internal.Contracts.PackageSourceContextInfo[1]}, targetFrameworks = {string[1]}, searchText = "", searchFilter = {NuGet.Protocol.Core.Types.SearchFilter}, itemFilter = Installed, isSolution = false, useRecommender = false, cancellationToken = IsCancellationRequested = false) Line 267
> 	at C:\NuGet.Client\src\NuGet.Clients\NuGet.PackageManagement.VisualStudio\Services\NuGetPackageSearchService.cs(267)
> [External Code]
> [Async] NuGet.PackageManagement.UI.PackageItemLoader.SearchAsync(cancellationToken = IsCancellationRequested = false) Line 228
> 	at C:\NuGet.Client\src\NuGet.Clients\NuGet.PackageManagement.UI\PackageItemLoader.cs(228)
>

martinrrm:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13737

## Description
Project References are now considered when getting the projects transitive packages, in some scenarios there is no framework in the project reference on the assets file, this caused a null reference in the PM UI when trying to get the framework.

This PR will ignore those project references (in the PM UI for showing transitive's) that don't have a null reference.

## PR Checklist

- [x] Added tests

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
@@ -220,29 +220,12 @@ namespace NuGet.PackageManagement.VisualStudio
                         }
 
                         // If the project has project references, we need to compute transitive origins for their packages
-                        List<LockFileTargetLibrary> projectReferences = targetsList
-                            .SelectMany(t => t.Libraries)
-                            .Where(l => l.Type == LibraryType.Project && l.Framework is not null).ToList();
-                        List<PackageReference> calculatedLibraryReferences = new List<PackageReference>(projectReferences.Count + calculatedInstalledPackages.Count);
-                        if (projectReferences.Count > 0)
-                        {
-                            foreach (var projectReference in projectReferences)
-                            {
-                                var packageReference = new PackageReference(
-                                    identity: new PackageIdentity(projectReference.Name, projectReference.Version),
-                                    targetFramework: new NuGetFramework(projectReference.Framework),
-                                    userInstalled: false,
-                                    developmentDependency: false,
-                                    requireReinstallation: false,
-                                    allowedVersions: new VersionRange(projectReference.Version));
-                                calculatedLibraryReferences.Add(packageReference);
-                            }
-                        }
+                        List<PackageReference> projectReferences = GetProjectPackageReferences(targetsList);
 
-                        calculatedLibraryReferences.AddRange(calculatedInstalledPackages);
+                        projectReferences.AddRange(calculatedInstalledPackages);
 
                         // Compute Transitive Origins
-                        transitiveOrigins = calculatedTransitivePackages.Any() ? ComputeTransitivePackageOrigins(calculatedLibraryReferences, targetsList, token) : new Dictionary<string, TransitiveEntry>();
+                        transitiveOrigins = calculatedTransitivePackages.Any() ? ComputeTransitivePackageOrigins(projectReferences, targetsList, token) : new Dictionary<string, TransitiveEntry>();
                     }
                     else
                     {
@@ -298,6 +281,34 @@ namespace NuGet.PackageManagement.VisualStudio
             IsInstalledAndTransitiveComputationNeeded = false;
 
             return new ProjectPackages(calculatedInstalledPackages, transitivePkgsResult);
+        }
+
+        private static List<PackageReference> GetProjectPackageReferences(IList<LockFileTarget> targetsList)
+        {
+            var packageReferences = new List<PackageReference>();
+
+            foreach (var target in targetsList)
+            {
+                var framework = target.TargetFramework;
+                var projectLibraries = target.Libraries.Where(library => library.Type == "project");
+
+                foreach (var library in projectLibraries)
+                {
+                    var packageReference = new PackageReference(
+                                    identity: new PackageIdentity(library.Name, library.Version),
+                                    targetFramework: new NuGetFramework(framework),
+                                    userInstalled: false,
+                                    developmentDependency: false,
+                                    requireReinstallation: false,
+                                    allowedVersions: new VersionRange(library.Version));
+                    packageReferences.Add(packageReference);
+                }
+            }
+
+            return packageReferences
+                .GroupBy(p => p.PackageIdentity)
+                .Select(g => g.OrderBy(p => p.TargetFramework, FrameworkSorter).First())
+                .ToList();
         }
 
         protected abstract IEnumerable<PackageReference> ResolvedInstalledPackagesList(IEnumerable<LibraryDependency> libraries, NuGetFramework targetFramework, IList<LockFileTarget> targets, T installedPackages);
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
@@ -220,7 +220,9 @@ namespace NuGet.PackageManagement.VisualStudio
                         }
 
                         // If the project has project references, we need to compute transitive origins for their packages
-                        List<LockFileTargetLibrary> projectReferences = targetsList.SelectMany(t => t.Libraries).Where(l => l.Type == LibraryType.Project).ToList();
+                        List<LockFileTargetLibrary> projectReferences = targetsList
+                            .SelectMany(t => t.Libraries)
+                            .Where(l => l.Type == LibraryType.Project && l.Framework is not null).ToList();
                         List<PackageReference> calculatedLibraryReferences = new List<PackageReference>(projectReferences.Count + calculatedInstalledPackages.Count);
                         if (projectReferences.Count > 0)
                         {
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
@@ -220,12 +220,18 @@ namespace NuGet.PackageManagement.VisualStudio
                         }
 
                         // If the project has project references, we need to compute transitive origins for their packages
-                        List<PackageReference> projectReferences = GetProjectPackageReferences(targetsList);
+                        List<PackageReference> projectReferences = packageSpec
+                            .TargetFrameworks
+                            .SelectMany(f => GetProjectPackageReferences(f.FrameworkName, targetsList))
+                            .GroupBy(p => p.PackageIdentity)
+                            .Select(g => g.OrderBy(p => p.TargetFramework, FrameworkSorter).First())
+                            .ToList();
 
-                        projectReferences.AddRange(calculatedInstalledPackages);
+                        List<PackageReference> calculatedLibraryReferences = new List<PackageReference>(projectReferences);
+                        calculatedLibraryReferences.AddRange(calculatedInstalledPackages);
 
                         // Compute Transitive Origins
-                        transitiveOrigins = calculatedTransitivePackages.Any() ? ComputeTransitivePackageOrigins(projectReferences, targetsList, token) : new Dictionary<string, TransitiveEntry>();
+                        transitiveOrigins = calculatedTransitivePackages.Any() ? ComputeTransitivePackageOrigins(calculatedLibraryReferences, targetsList, token) : new Dictionary<string, TransitiveEntry>();
                     }
                     else
                     {
@@ -283,32 +289,22 @@ namespace NuGet.PackageManagement.VisualStudio
             return new ProjectPackages(calculatedInstalledPackages, transitivePkgsResult);
         }
 
-        private static List<PackageReference> GetProjectPackageReferences(IList<LockFileTarget> targetsList)
+
+        private static IEnumerable<PackageReference> GetProjectPackageReferences(NuGetFramework nuGetFramework, IList<LockFileTarget> targetsList)
         {
-            var packageReferences = new List<PackageReference>();
+            var packageReferences = targetsList
+                .Where(t => t.TargetFramework.Equals(nuGetFramework))
+                .SelectMany(lib => lib.Libraries)
+                .Where(l => l.Type == "project")
+                .Select(package => new PackageReference(
+                    new PackageIdentity(package.Name, package.Version),
+                    targetFramework: nuGetFramework,
+                    userInstalled: false,
+                    developmentDependency: false,
+                    requireReinstallation: false,
+                    allowedVersions: new VersionRange(package.Version)));
 
-            foreach (var target in targetsList)
-            {
-                var framework = target.TargetFramework;
-                var projectLibraries = target.Libraries.Where(library => library.Type == "project");
-
-                foreach (var library in projectLibraries)
-                {
-                    var packageReference = new PackageReference(
-                                    identity: new PackageIdentity(library.Name, library.Version),
-                                    targetFramework: new NuGetFramework(framework),
-                                    userInstalled: false,
-                                    developmentDependency: false,
-                                    requireReinstallation: false,
-                                    allowedVersions: new VersionRange(library.Version));
-                    packageReferences.Add(packageReference);
-                }
-            }
-
-            return packageReferences
-                .GroupBy(p => p.PackageIdentity)
-                .Select(g => g.OrderBy(p => p.TargetFramework, FrameworkSorter).First())
-                .ToList();
+            return packageReferences;
         }
 
         protected abstract IEnumerable<PackageReference> ResolvedInstalledPackagesList(IEnumerable<LibraryDependency> libraries, NuGetFramework targetFramework, IList<LockFileTarget> targets, T installedPackages);
--- a/test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
+++ b/test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
@@ -4126,6 +4126,31 @@ namespace NuGet.PackageManagement.VisualStudio.Test
         }
 
         [Fact]
+        public async Task GetInstalledAndTransitivePackagesAsync_ProjectReferenceWithNullFramework_EmptyInstalledAndTransitivePackagesAsync()
+        {
+            // Project2 -> Project1 ->  PackageA (1.0.0) -> PackageB (1.0.0)
+
+            // Arrange
+            using var rootDir = new SimpleTestPathContext();
+
+            await CreatePackagesAsync(rootDir, packageAVersion: "1.0.0");
+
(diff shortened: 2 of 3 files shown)


src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
```diff
@@ -220,7 +220,9 @@ internal async Task<ProjectPackages> GetInstalledAndTransitivePackagesAsync(bool
                         }
 
                         // If the project has project references, we need to compute transitive origins for their packages
-                        List<LockFileTargetLibrary> projectReferences = targetsList.SelectMany(t => t.Libraries).Where(l => l.Type == LibraryType.Project).ToList();
+                        List<LockFileTargetLibrary> projectReferences = targetsList
```
nkolev92: Why is the framework null here?
Is this expected or a bug? 

I feel like the PR explains the symptoms of the problem, but not the root cause nor the expected impact of the change.

test/TestUtilities/Test.Utility/Commands/ProjectTestHelpers.cs
```diff
@@ -92,7 +92,11 @@ public static PackageSpec WithTestProjectReference(this PackageSpec parent, Pack
         {
             var spec = parent.Clone();
 
-            if (frameworks.Length == 0)
+            if (frameworks is null)
+            {
+                frameworks = Array.Empty<NuGetFramework>();
```
nkolev92: When is the frameworks element empty in a real project?

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
```diff
@@ -4125,6 +4125,31 @@ public async Task GetInstalledAndTransitivePackagesAsync_MutipleProjectReference
...
+
+            await RestorePackageSpecsAsync(rootDir, output: null, prj1Spec, prj2Spec);
+
+            CpsPackageReferenceProject project2 = PrepareCpsRestoredProject(prj2Spec);
+
+            // Act
+            ProjectPackages result = await project2.GetInstalledAndTransitivePackagesAsync(includeTransitiveOrigins: true, CancellationToken.None);
+
+            // Assert
+            Assert.Empty(result.InstalledPackages); // No installed packages
```
nkolev92, line 4174: So there's no project 2 packages despite the project showing them as such?

nkolev92 (changes requested):

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
martinrrm: I haven't been able to understand exactly what causes a null framework in the assets file. Donnie was able to reproduce the error but not me, he mentioned that deleting obj helped fix this, so it makes me believe that is a wrong state of the project.

This is an example of the assets file with the problem. 
image

martinrrm: Another solution is to populate the framework property using the targets. For example, in the above image targets has a dictionary of frameworks that contains a list of the libraries.

```c#
 private static List<PackageReference> GetProjectPackageReferences(IList<LockFileTarget> targetsList)
        {
            var packageReferences = new List<PackageReference>();

            foreach (var target in targetsList)
            {
                var framework = target.TargetFramework;
                var projectLibraries = target.Libraries.Where(library => library.Type == "project");

                foreach (var library in projectLibraries)
                {
                    var packageReference = new PackageReference(
                                    identity: new PackageIdentity(library.Name, library.Version),
                                    targetFramework: new NuGetFramework(framework),
                                    userInstalled: false,
                                    developmentDependency: false,
                                    requireReinstallation: false,
                                    allowedVersions: new VersionRange(library.Version));
                    packageReferences.Add(packageReference);
                }
            }

            return packageReferences
                .GroupBy(p => p.PackageIdentity)
                .Select(g => g.OrderBy(p => p.TargetFramework, FrameworkSorter).First())
                .ToList();
        }

```

donnie-msft: I don't know, either, but since I mentioned that clearing my obj/bin folders resolved it, it could be something to do with changing a TFM and then a cache/etc not updating properly? Just guessing.
IMO, fixing the method to handle a null is a good idea even if nobody understands why it happens. Screenshots prove it happened.

nkolev92: I wasn't sure before I looked at this code either. 

Unfortunately is https://github.com/NuGet/NuGet.Client/blob/2982dbfc1bac22d71234d8498af6ad43e129a49c/src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileTargetLibrary.cs#L71 not documented super well.

@martinrrm did point out the targets section has a framework, but I tihnk those are 2 different values, current project framework, vs the framework of the selected project.

Doing a quick look-up where the LockFileTargetLibrary for projects is generated (only 1 place in the code), https://github.com/NuGet/NuGet.Client/blob/2982dbfc1bac22d71234d8498af6ad43e129a49c/src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/LockFileUtils.cs#L524-L537, it seems like there's some case in which it is not going to be there. 

How is this value being used?
It seems it's supposed to mean the project framework used for the transitive project but can't think of why that'd be relevant in this case.

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
nkolev92: I'd love to understand the details beyond this test case.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
martinrrm: So, the PM UI packages list depends on the project framework. Right now, we [use a sorter](https://github.com/NuGet/NuGet.Client/blob/2982dbfc1bac22d71234d8498af6ad43e129a49c/src/NuGet.Core/NuGet.Frameworks/comparers/NuGetFrameworkSorter.cs#L14) to get a specific framework to [display the packages](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs#L181) (meaning that PM UI doesn't work with multi targeting).

Also, this framework is used to [calculate the transitive origins of a package](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs#L449), and this is important in order to display transitive packages, a package without a transitive origin is not displayed. 

Since the framework can be null, I suggest not considering those projects.

nkolev92: > So, the PM UI packages list depends on the project framework. Right now, we [use a sorter](https://github.com/NuGet/NuGet.Client/blob/2982dbfc1bac22d71234d8498af6ad43e129a49c/src/NuGet.Core/NuGet.Frameworks/comparers/NuGetFrameworkSorter.cs#L14) to get a specific framework to [display the packages](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs#L181) (meaning that PM UI doesn't work with multi targeting).

Based on this and what the code I linked seems to be doing, I think you only want the current project's framework. The framework of the project being chosen doesn't seem very relevant.

> Also, this framework is used to [calculate the transitive origins of a package](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs#L449), and this is important in order to display transitive packages, a package without a transitive origin is not displayed.

I can't tell how it's being used, but given that packages *never* set the framework value, I can't imagine the projects should be treated any differently.

> Since the framework can be null, I suggest not considering those projects.

I don't think this is right. Those projects don't seem to be invalid, there's no restore issue, but an issue with the interpretation of the value.

We'd also need to consider what ignoring those projects leads to. It'd lead to the same ghost transitive dependencies issue that we're trying to fix.

martinrrm: I updated the code to use the same target framework that we use to get the transitive and installed packages. https://github.com/NuGet/NuGet.Client/blob/bfc13ee02b8ff6128d123c820093a4110b2d6b52/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs#L180-L202

In the above code we are also comparing the package spec target frameworks against the ones in the assets file. that help us get the current project's framework.

nkolev92: There's no commit pushed.

martinrrm: Mb, just pushed now

test/TestUtilities/Test.Utility/Commands/ProjectTestHelpers.cs
nkolev92: Can you explain these changes?

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
```diff
@@ -4125,6 +4125,31 @@ public async Task GetInstalledAndTransitivePackagesAsync_MutipleProjectReference
...
+            Assert.Empty(result.TransitivePackages); // No transitive packages
```
nkolev92, line 4175: Should this be returning PackageA and PackageB?

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/CpsPackageReferenceProjectTests.cs
martinrrm: Deleting this change, I was wrong about the null framework, it's in the assets file when it's null

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
```diff
@@ -220,23 +220,14 @@ internal async Task<ProjectPackages> GetInstalledAndTransitivePackagesAsync(bool
...
-                                    allowedVersions: new VersionRange(projectReference.Version));
-                                calculatedLibraryReferences.Add(packageReference);
-                            }
-                        }
-
+                        List<PackageReference> projectReferences = packageSpec
+                            .TargetFrameworks
+                            .SelectMany(f => GetProjectPackageReferences(f.FrameworkName, targetsList))
+                            .GroupBy(p => p.PackageIdentity)
+                            .Select(g => g.OrderBy(p => p.TargetFramework, FrameworkSorter).First())
```
nkolev92, line 227: In here, you're basically deduplicating and choosing the first one. 

Can this cause issues when different frameworks of the same project have different references?
Would all of them be shown?

netstandard2.0
P1 -> P2 -> A, B
net472
P1 -> P2 -> A, C

This can be done in a follow up since it seems like this is not a new behavior, but how the old works as well.

nkolev92 (approved):

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/PackageReferenceProject.cs
martinrrm: This is definitely going to cause issues with multiple frameworks, currently the PM UI uses this `FrameworkSorter` to decide which packages to show, this behavior is also present in top-level 😭 we need to design a better solution for this problem

nkolev92: Here's my understanding. 

1. The PM UI will show *all* transitive packages. 
1. If a package is available with different versions in a different framework, the PM UI does not guarantee which  one of the versions is shown. 

My fear here was whether assumption #1 is correct.
It feels like a bigger issue.

donnie-msft: 👀I'm not following the sort & choose first TFM logic. Can a test case be added that demonstrates why the first TFM works? In Nikolche's example I think this means only net472 will be added to `projectReferences` for group `p.PackageIdentity` ?

martinrrm: > The PM UI will show all transitive packages.

You are right, it will show all top level and transitive's, but just as you mentioned in second point we do not guarantee which one is shown. There cannot be any duplicated package items in the same level (Top and transitive).

I thought the PM UI was choosing a framework before creating the list, at least we are displaying more information that I believed.

image
image

> If a package is available with different versions in a different framework, the PM UI does not guarantee which one of the versions is shown.

With the above configuration we see that the transitive origin data is using the other package version.

image

martinrrm:
@donnie-msft @nkolev92 I merged the PR since the open comment wasn't related to this PR changes. But definitely something to have in mind.

PR #6034, opened 2024-09-18
[Transitive's in Solution] Transitive's data in Solution Details Pane

martinrrm:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13698
This PR adds a new column to the details pane to show the level of the package. The installed column shows the installed version for top and transitive levels.

Vulnerability icons are also added so it's easier to users to know which ones are vulnerable.

 
## Description
image
image

## PR Checklist

- [ ] Added tests

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
@@ -118,12 +118,12 @@ namespace NuGet.PackageManagement.UI
 
         public bool IsRequestedVisible { get; private set; }
 
-        private static void UpdateProjectInstallationInfo(PackageInstallationInfo project, PackageLevel packageLevel, IPackageReferenceContextInfo installedVersion, HashSet<NuGetVersion> installedVersionsSet)
+        private static void UpdateProjectInstallationInfo(PackageInstallationInfo project, IPackageReferenceContextInfo installedVersion, HashSet<NuGetVersion> installedVersionsSet)
         {
             project.InstalledVersion = installedVersion.Identity.Version;
             installedVersionsSet.Add(installedVersion.Identity.Version);
 
-            if (packageLevel == PackageLevel.TopLevel)
+            if (project.PackageLevel == PackageLevel.TopLevel)
             {
                 project.AutoReferenced = installedVersion.IsAutoReferenced;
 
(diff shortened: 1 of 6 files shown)


src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
```diff
@@ -191,18 +212,20 @@ private async Task UpdateInstalledVersionsAsync(CancellationToken cancellationTo
         /// This method is called from several methods that are called from properties and LINQ queries
         /// It is likely not called more than once in an action.
         /// </summary>
-        private async Task<IPackageReferenceContextInfo> GetInstalledPackageAsync(
+        private async Task<(IPackageReferenceContextInfo TopLevelPackage, IPackageReferenceContextInfo TransitivePackage)> GetInstalledPackageAsync(
```
jebriede: Throughout the code, when we return just top-level packages, we have a method called GetInstalledPackagesAsync, but when we return top-level and transitive packages, we call it GetInstalledAndTransitivePackagesAsync. For consistency, we should consider doing that here now that this method returns both.

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
```diff
@@ -127,30 +143,35 @@ private async Task UpdateInstalledVersionsAsync(CancellationToken cancellationTo
...
-                    if (installedVersion != null)
+                    (IPackageReferenceContextInfo TopLevelPackage, IPackageReferenceContextInfo TransitivePackage) installedPackage = await GetInstalledPackageAsync(project.NuGetProject, Id, cancellationToken);
+                    var topLevelPackageVersion = installedPackage.TopLevelPackage;
+                    var transitivePackageVersion = installedPackage.TransitivePackage;
+
+                    project.RequestedVersion = null;
+                    project.InstalledVersion = null;
+                    project.PackageLevel = null;
+                    project.AutoReferenced = false;
+                    project.InstalledVersionMaxVulnerability = -1;
```
jebriede: Nit: I don't see a lot of gain from moving this out of the else case. It does set some defaults, but they get overridden in the 2 specified cases, so it would be fine to leave them in the else case.

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
```diff
@@ -127,30 +143,35 @@ private async Task UpdateInstalledVersionsAsync(CancellationToken cancellationTo
...
-                        {
-                            project.RequestedVersion = installedVersion?.AllowedVersions?.OriginalString;
-                        }
-
-                        if (_searchResultPackage.VulnerableVersions.TryGetValue(installedVersion.Identity.Version, out int vulnerable))
-                        {
-                            project.InstalledVersionMaxVulnerability = vulnerable;
-                            vulnerabilitiesSet.Add(installedVersion.Identity.Version);
-                        }
+                        project.PackageLevel = "Transitive";
```
jebriede: We should use the PackageLevel enum from NuGet.VisualStudio.Common.

jebriede:
Please associate the work item, move the description of your fix from the Bug/Fixes section to the Description section, and add checkmarks in the appropriate boxes at the bottom. Thanks!

jebriede:
Please be sure to test these changes in the Project-level PM UI to make sure we did not regress anything there.

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
martinrrm: One of the UX board comments was that it should be the same text as the packages list headers, and unfortunately it is "Top-level" instead of "Top-Level" and if I use the enum is going to be with capital L.

Do you want me to change the header to have capital L? Which one do you think looks better?

jebriede: It's the View's responsibility to decide what string should be displayed based on the PackageLevel type so we should use the enum in the ViewModel (which this class acts as, despite its name), and convert to a string resource in the View.

jebriede: This string also needs to be localized, along with the corresponding "Top-level" one.

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageInstallationInfo.cs
```diff
@@ -8,6 +8,7 @@
 using Microsoft.ServiceHub.Framework;
 using NuGet.PackageManagement.VisualStudio;
 using NuGet.Versioning;
+using NuGet.VisualStudio;
```
martinrrm, line 11: remove it

jebriede (approved):

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
```diff
@@ -127,30 +143,34 @@ private async Task UpdateInstalledVersionsAsync(CancellationToken cancellationTo
...
+                        project.PackageLevel = PackageLevel.TopLevel;
+                        UpdateProjectInstallationInfo(project, topLevelPackageVersion, installedVersionsSet);
                     }
                     else
                     {
-                        project.RequestedVersion = null;
                         project.InstalledVersion = null;
-                        project.AutoReferenced = false;
-                        project.InstalledVersionMaxVulnerability = -1;
+                        project.PackageLevel = null;
```
jebriede, line 167: Nit: In a previous comment I suggested moving these defaults back to the else case, but if there are some defaults that must be set for either top-level or transitive, then maybe just keep the defaults all together above. It's up to you.

PR #6036, opened 2024-09-19
Add tests for DependencyGraphResolver.LibraryRangeComparer

jeffkl:
# Bug

Tracked in https://github.com/NuGet/Client.Engineering/issues/3008

## Description
This change add tests for DependencyGraphResolver.LibraryRangeComparer as a follow-up to its implementation.

This comparer takes into account the same logic as LibraryRange.ToString() which has custom logic around what two LibraryRange objects are considered equal.

## PR Checklist

- [x] Added tests

### DIFF
--- a/test/NuGet.Core.Tests/NuGet.Commands.Test/DependencyGraphResolverTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Commands.Test/DependencyGraphResolverTests.cs
@@ -20,55 +20,15 @@ namespace NuGet.Commands.Test
             string libraryRange1String = libraryRange1.ToString();
             string libraryRange2String = libraryRange2.ToString();
 
-            var constraintString1 = string.Empty;
-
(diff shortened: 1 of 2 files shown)


test/NuGet.Core.Tests/NuGet.Commands.Test/DependencyGraphResolverTests.cs
```diff
@@ -0,0 +1,183 @@
...
+                    constraintString1 = LibraryType.ExternalProject;
+                    break;
+
+                case LibraryDependencyTarget.Project:
+                case LibraryDependencyTarget.Project | LibraryDependencyTarget.ExternalProject:
+                    constraintString1 = LibraryType.Project;
+                    break;
+            }
+
+            var constraintString2 = string.Empty;
```
nkolev92: Can we pass this in?

It feels like we're testing the test here.

nkolev92 (approved):

PR #6040, opened 2024-09-20
Show transitive package details for packages that are top-level and transitive in the solution

jebriede:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13798

## Description
To support the Solution PM UI, we are tracking information about all the package versions from each project when building the PackageItemViewModel. If a package is top-level but also has transitive origins, it means the package is top-level in at least one project, and transitive in at least one more. In that case, we want to show the transitive origins information to the user. To do that, we check if any of the package versions have transitive origins, and if so, we update the transitive info tool tip message.

Changes in the XAML include restructuring the package item tooltip code so it doesn't rely on data triggers for deciding the binding for each Run. Now, we define the section for the package description and a section for the transitive package info and we trigger visibility based on the package level and whether the transitive origins message is set. This simplifies and adds better structure to the tooltip code while preserving the same look and allows for more flexibility in the future.

## PR Checklist

- [ ] ~Added tests~ Will add test(s) in follow-up to feature branch

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
@@ -4,14 +4,12 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
-using System.Collections.ObjectModel;
 using System.Globalization;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft;
 using Microsoft.ServiceHub.Framework;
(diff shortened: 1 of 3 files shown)


jebriede:
A few example screenshots of a package in the Solution PM UI that is both top-level and transitive in the solution:

image
image

jebriede:
## Solution PM UI

- Package that is top-level in one or more projects, and transitive in one or more projects:
image

- Package that is only top level in the solution:
image

- Browse Tab (remains unchanged):
image


## Project PM UI
(Remains the same as before)

- Top level package:
image

- Transitive package:
image

martinrrm (approved):

PR #6041, opened 2024-09-20
Fix runtime dependency resolution and downgrades for VersionRange.All

### LINKED ISSUE #13794: Dotnet restore is unable to find runtime.*.Sytem packages after updating the SDK from 8.0.401 to 9.0.100-rc.1.24452.12 when using lock files
### NuGet Product Used

dotnet.exe

### Product Version

.NET SDK 9.0.100-rc.1.24452.12

### Worked before?

.NET SDK 8.0.401

### Impact

### Repro Steps & Context

We have multiple csproj files in our org that include various runtime.*.System packages as transitive dependencies, which do get resolved in packages.lock.json, similar to the following:

```
      "runtime.win.System.Diagnostics.Debug": {
        "type": "Transitive",
        "resolved": "4.3.0",
        "contentHash": "hHHP0WCStene2jjeYcuDkETozUYF/3sHVRHAEOgS3L15hlip24ssqCTnJC28Z03Wpo078oMcJd0H4egD2aJI8g=="
      },
```

When running the following after updating the SDK version:

`dotnet.exe restore --use-lock-file --no-dependencies /p:ImportProjectExtensionProps=false /p:ImportProjectExtensionTargets=false /p:PublishReadyToRun=true /p:CommandlinePlatform=x64 /nodeReuse:false /maxcpucount:1 --lock-file-path "packages.lock.json"`

It then fails with messages similar to the following (repeated for multiple runtime.*.System.* packages):

`error NU1101: Unable to find package runtime.win.System.Diagnostics.Debug. No packages exist with this id in source(s)`

The package in question exists directly on the feed being used (and is even already in the local cache due to the earlier restore call when using 8.0.401). 

If I delete the content of the lock file and re-run, restore succeeds and re-creates the exact same content. Attempting to rebuild again after that fails for restore again in the same way.  

The specific MSBuild version being used is 17.12.0-preview-24422-09+d17ec720d.  

### Verbose Logs

_No response_

jeffkl:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13794

## Description
When resolving runtime dependencies via a `packages.lock.json`, the runtime must be passed along.  The original implementation of the new resolver was not doing this and we have no tests for it.  The resolver now passes the runtime identifier for runtime dependencies.

This change also fixes a bug where the resolver considers a downgrade for a package dependency when no version is specified and instead it defaults to `VersionRange.All`.  This results in a downgrade warning and instead the code should verify if the version is satisfied by the chosen version.

## PR Checklist

- [ ] Added tests

### DIFF
--- a/test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommand_AlgorithmEquivalencyTests.cs
+++ b/test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommand_AlgorithmEquivalencyTests.cs
@@ -1289,7 +1289,7 @@ namespace NuGet.Commands.FuncTest
             // Act & Assert
             (var result, _) = await ValidateRestoreAlgorithmEquivalency(pathContext, project1);
             result.Success.Should().BeTrue();
-            result.LockFile.Targets.Should().HaveCount(1);
+            result.LockFile.Targets.Should().HaveCount(2);
             result.LockFile.Targets[0].Libraries.Should().HaveCount(2);
             result.LockFile.Targets[0].Libraries[0].Name.Should().Be("a");
             result.LockFile.Targets[0].Libraries[1].Name.Should().Be("b");
--- a/test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommand_AlgorithmEquivalencyTests.cs
+++ b/test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommand_AlgorithmEquivalencyTests.cs
@@ -1,7 +1,9 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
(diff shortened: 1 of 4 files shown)


test/NuGet.Core.FuncTests/NuGet.Commands.FuncTest/RestoreCommand_AlgorithmEquivalencyTests.cs
```diff
@@ -1131,6 +1129,173 @@ await SimpleTestPackageUtility.CreateFolderFeedV3Async(
...
+                {
+                    Dependencies = [new SimpleTestPackageContext("b", null)],
+                },
+                new SimpleTestPackageContext("b", "2.0.0")
+                );
+
+            // Setup project
+            var spec1 = @"
+                {
+                  ""runtimes"": {
```
nkolev92, line 1267: I think for this test, we could probably just remove the runtime too. 
I like having runtime tests though, so I don't mind this. 

Basically, all I'm saying is that the runtimes wasn't intentional :D

PR #6044, opened 2024-09-23
Transitive packages and vulnerability indicators in Solution PM UI

### LINKED ISSUE #13216: [Feature] Enable Transitive Dependencies for Solution-level in Visual Studio
### NuGet Product(s) Involved

### The Elevator Pitch

People love using the new transitive dependencies feature in Visual Studio for their projects! 

https://devblogs.microsoft.com/nuget/introducing-transitive-dependencies-in-visual-studio/

One common follow-up is that people would love to see this experience in the solution-level within Visual Studio.

As we know from survey data, people use projects / solution levels just about equally and we originally went with projects to experiment with this feature as it has not been done before in Visual Studio.

This issue tracks bringing this experience to the solution-level in Visual Studio.

Please 👍 or 👎 this comment to help us with the direction of this feature & leave as much feedback/questions/concerns as you'd like on this issue itself and we will get back to you shortly.

Thank You 🎉

_No response_

martinrrm:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13216

## Description
This feature brings to the Solution-level PM UI the Transitive Packages experience that we previously added to the Project-level PM UI. Additionally, we added new UX to the projects / versions grid in the Details Pane to show the package level of the selected package, as well as any vulnerabilities.

Accessibility localization PR https://github.com/NuGet/NuGet.Client/pull/6051

image

### Accessibility
image

#### Details pane
It has a minimum size so the columns will always look "good"

image

## PR Checklist

- [x] Added tests

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/GreaterThanThresholdToVisibilityConverter.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/GreaterThanThresholdToVisibilityConverter.cs
@@ -15,11 +15,14 @@ namespace NuGet.PackageManagement.UI
         {
             if (targetType == typeof(Visibility))
             {
-                if (value is not null)
+                if (value is not null && parameter is not null)
                 {
-                    long intValue = System.Convert.ToInt64(value, culture);
-                    long parameterValue = System.Convert.ToInt64(parameter, culture);
-                    if (intValue > parameterValue)
+                    bool valueConverted = TryConvertToInt64(value, culture, out long? initialValue);
+                    bool paramConverted = TryConvertToInt64(parameter, culture, out long? thresholdValue);
+
+                    if (valueConverted && paramConverted
+                        && initialValue is not null && thresholdValue is not null
+                        && initialValue > thresholdValue)
                     {
                         return Visibility.Visible;
                     }
@@ -36,6 +39,26 @@ namespace NuGet.PackageManagement.UI
             // no op
             Debug.Fail("Not Implemented");
             return null;
+        }
+
+        private static bool TryConvertToInt64(object value, CultureInfo culture, out long? convertedValue)
+        {
+            if (value is null)
+            {
+                convertedValue = null;
+                return false;
+            }
+
+            try
+            {
+                convertedValue = System.Convert.ToInt64(value, culture);
+                return true;
+            }
+            catch (Exception e) when (e is FormatException or InvalidCastException or OverflowException)
+            {
+                convertedValue = null;
+                return false;
+            }
         }
     }
 }
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/GreaterThanThresholdToVisibilityConverter.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/GreaterThanThresholdToVisibilityConverter.cs
@@ -9,7 +9,7 @@ using System.Windows.Data;
 
 namespace NuGet.PackageManagement.UI
 {
-    public class IsGreaterThanToVisibilityConverter : IValueConverter
+    public class GreaterThanThresholdToVisibilityConverter : IValueConverter
     {
         public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
         {
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/IsGreaterThanToVisibilityConverter.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/IsGreaterThanToVisibilityConverter.cs
@@ -9,16 +9,20 @@ using System.Windows.Data;
 
 namespace NuGet.PackageManagement.UI
 {
-    public class DownloadCountToVisibilityConverter : IValueConverter
+    public class IsGreaterThanToVisibilityConverter : IValueConverter
     {
         public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
         {
             if (targetType == typeof(Visibility))
             {
-                long? downloadCount = value as long?;
-                if (downloadCount >= 0)
+                if (value is not null)
                 {
-                    return Visibility.Visible;
+                    long intValue = System.Convert.ToInt64(value, culture);
+                    long parameterValue = System.Convert.ToInt64(parameter, culture);
+                    if (intValue > parameterValue)
+                    {
+                        return Visibility.Visible;
+                    }
                 }
 
                 return Visibility.Collapsed;
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
@@ -25,6 +25,7 @@ namespace NuGet.PackageManagement.UI
         private INuGetSolutionManagerService _solutionManager;
         private string _installedVersions; // The text describing the installed versions information, such as "not installed", "multiple versions installed" etc.
         private int _installedVersionsCount; // the count of different installed versions
+        private int _installedVulnerabilitiesCount; // the count of distinct vulnerable installed versions
         private bool _canUninstall;
         private bool _canInstall;
         // Indicates whether the SelectCheckBoxState is being updated in code. True means the state is being updated by code, while false means the state is changed by user clicking the checkbox.
@@ -119,7 +120,8 @@ namespace NuGet.PackageManagement.UI
 
         private async Task UpdateInstalledVersionsAsync(CancellationToken cancellationToken)
         {
-            var hash = new HashSet<NuGetVersion>();
+            var installedVersionsSet = new HashSet<NuGetVersion>();
+            var vulnerabilitiesSet = new HashSet<NuGetVersion>();
 
             foreach (var project in _projects)
             {
@@ -129,12 +131,18 @@ namespace NuGet.PackageManagement.UI
                     if (installedVersion != null)
                     {
                         project.InstalledVersion = installedVersion.Identity.Version;
-                        hash.Add(installedVersion.Identity.Version);
+                        installedVersionsSet.Add(installedVersion.Identity.Version);
                         project.AutoReferenced = installedVersion.IsAutoReferenced;
 
                         if (project.NuGetProject.ProjectStyle.Equals(ProjectStyle.PackageReference))
                         {
                             project.RequestedVersion = installedVersion?.AllowedVersions?.OriginalString;
+                        }
+
+                        if (_searchResultPackage.VulnerableVersions.TryGetValue(installedVersion.Identity.Version, out int vulnerable))
+                        {
+                            project.InstalledVersionMaxVulnerability = vulnerable;
+                            vulnerabilitiesSet.Add(installedVersion.Identity.Version);
                         }
                     }
                     else
@@ -142,6 +150,7 @@ namespace NuGet.PackageManagement.UI
                         project.RequestedVersion = null;
                         project.InstalledVersion = null;
                         project.AutoReferenced = false;
+                        project.InstalledVersionMaxVulnerability = -1;
                     }
                 }
                 catch (Exception ex)
@@ -155,16 +164,17 @@ namespace NuGet.PackageManagement.UI
                 }
             }
 
-            InstalledVersionsCount = hash.Count;
+            InstalledVersionsCount = installedVersionsSet.Count;
+            InstalledVulnerabilitiesCount = vulnerabilitiesSet.Count;
 
-            if (hash.Count == 0)
+            if (installedVersionsSet.Count == 0)
             {
                 InstalledVersions = Resources.Text_NotInstalled;
             }
-            else if (hash.Count == 1)
+            else if (installedVersionsSet.Count == 1)
             {
                 var displayVersion = new DisplayVersion(
-                    hash.First(),
+                    installedVersionsSet.First(),
                     string.Empty);
                 InstalledVersions = displayVersion.ToString();
             }
@@ -273,6 +283,16 @@ namespace NuGet.PackageManagement.UI
             {
                 _installedVersionsCount = value;
                 OnPropertyChanged(nameof(InstalledVersionsCount));
+            }
+        }
+
+        public int InstalledVulnerabilitiesCount
+        {
+            get => _installedVulnerabilitiesCount;
+            set
+            {
+                _installedVulnerabilitiesCount = value;
+                OnPropertyChanged(nameof(InstalledVulnerabilitiesCount));
             }
         }
 
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
@@ -167,7 +167,7 @@ namespace NuGet.PackageManagement.UI
                         project.PackageLevel = null;
                     }
 
-                    if (_searchResultPackage.VulnerableVersions.TryGetValue(project.InstalledVersion, out int vulnerable))
+                    if (project?.InstalledVersion is not null && _searchResultPackage.VulnerableVersions.TryGetValue(project.InstalledVersion, out int vulnerable))
                     {
                         project.InstalledVersionMaxVulnerability = vulnerable;
                         vulnerabilitiesSet.Add(project.InstalledVersion);
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
@@ -118,6 +118,22 @@ namespace NuGet.PackageManagement.UI
 
         public bool IsRequestedVisible { get; private set; }
 
+        private static void UpdateProjectInstallationInfo(PackageInstallationInfo project, IPackageReferenceContextInfo installedVersion, HashSet<NuGetVersion> installedVersionsSet)
+        {
+            project.InstalledVersion = installedVersion.Identity.Version;
+            installedVersionsSet.Add(installedVersion.Identity.Version);
+
+            if (project.PackageLevel == PackageLevel.TopLevel)
+            {
+                project.AutoReferenced = installedVersion.IsAutoReferenced;
+
+                if (project.NuGetProject.ProjectStyle.Equals(ProjectStyle.PackageReference))
+                {
+                    project.RequestedVersion = installedVersion?.AllowedVersions?.OriginalString;
+                }
+            }
+        }
+
         private async Task UpdateInstalledVersionsAsync(CancellationToken cancellationToken)
         {
             var installedVersionsSet = new HashSet<NuGetVersion>();
@@ -127,30 +143,34 @@ namespace NuGet.PackageManagement.UI
             {
                 try
                 {
-                    IPackageReferenceContextInfo installedVersion = await GetInstalledPackageAsync(project.NuGetProject, Id, cancellationToken);
-                    if (installedVersion != null)
+                    (IPackageReferenceContextInfo TopLevelPackage, IPackageReferenceContextInfo TransitivePackage) installedPackage = await GetInstalledAndTransitivePackagesAsync(project.NuGetProject, Id, cancellationToken);
+                    var topLevelPackageVersion = installedPackage.TopLevelPackage;
+                    var transitivePackageVersion = installedPackage.TransitivePackage;
+
+                    project.InstalledVersionMaxVulnerability = -1;
+                    project.RequestedVersion = null;
+                    project.AutoReferenced = false;
+
+                    if (transitivePackageVersion != null)
                     {
-                        project.InstalledVersion = installedVersion.Identity.Version;
-                        installedVersionsSet.Add(installedVersion.Identity.Version);
-                        project.AutoReferenced = installedVersion.IsAutoReferenced;
-
-                        if (project.NuGetProject.ProjectStyle.Equals(ProjectStyle.PackageReference))
-                        {
-                            project.RequestedVersion = installedVersion?.AllowedVersions?.OriginalString;
-                        }
-
-                        if (_searchResultPackage.VulnerableVersions.TryGetValue(installedVersion.Identity.Version, out int vulnerable))
-                        {
-                            project.InstalledVersionMaxVulnerability = vulnerable;
-                            vulnerabilitiesSet.Add(installedVersion.Identity.Version);
-                        }
+                        project.PackageLevel = PackageLevel.Transitive;
+                        UpdateProjectInstallationInfo(project, transitivePackageVersion, installedVersionsSet);
+                    }
+                    else if (topLevelPackageVersion != null)
+                    {
+                        project.PackageLevel = PackageLevel.TopLevel;
+                        UpdateProjectInstallationInfo(project, topLevelPackageVersion, installedVersionsSet);
                     }
                     else
                     {
-                        project.RequestedVersion = null;
                         project.InstalledVersion = null;
-                        project.AutoReferenced = false;
-                        project.InstalledVersionMaxVulnerability = -1;
+                        project.PackageLevel = null;
+                    }
+
+                    if (_searchResultPackage.VulnerableVersions.TryGetValue(project.InstalledVersion, out int vulnerable))
+                    {
+                        project.InstalledVersionMaxVulnerability = vulnerable;
+                        vulnerabilitiesSet.Add(project.InstalledVersion);
                     }
                 }
                 catch (Exception ex)
@@ -191,18 +211,20 @@ namespace NuGet.PackageManagement.UI
         /// This method is called from several methods that are called from properties and LINQ queries
         /// It is likely not called more than once in an action.
         /// </summary>
-        private async Task<IPackageReferenceContextInfo> GetInstalledPackageAsync(
+        private async Task<(IPackageReferenceContextInfo TopLevelPackage, IPackageReferenceContextInfo TransitivePackage)> GetInstalledAndTransitivePackagesAsync(
             IProjectContextInfo project,
             string packageId,
             CancellationToken cancellationToken)
         {
-            IEnumerable<IPackageReferenceContextInfo> installedPackages = await project.GetInstalledPackagesAsync(
-                ServiceBroker,
-                cancellationToken);
-            IPackageReferenceContextInfo installedPackage = installedPackages
+            IInstalledAndTransitivePackages installedAndTransitivePackages = await project.GetInstalledAndTransitivePackagesAsync(ServiceBroker, cancellationToken);
+            IPackageReferenceContextInfo installedPackages = installedAndTransitivePackages.InstalledPackages
                 .Where(p => StringComparer.OrdinalIgnoreCase.Equals(p.Identity.Id, packageId))
                 .FirstOrDefault();
-            return installedPackage;
+
+            ITransitivePackageReferenceContextInfo transitivePackages = installedAndTransitivePackages.TransitivePackages
+                .Where(p => StringComparer.OrdinalIgnoreCase.Equals(p.Identity.Id, packageId))
+                .FirstOrDefault();
+            return (installedPackages, transitivePackages);
         }
 
         protected override async Task CreateVersionsAsync(CancellationToken cancellationToken)
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/PackageInstallationInfo.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/PackageInstallationInfo.cs
@@ -8,6 +8,7 @@ using System.Threading.Tasks;
 using Microsoft.ServiceHub.Framework;
 using NuGet.PackageManagement.VisualStudio;
 using NuGet.Versioning;
+using NuGet.VisualStudio;
 using NuGet.VisualStudio.Internal.Contracts;
 
 namespace NuGet.PackageManagement.UI
@@ -76,6 +77,17 @@ namespace NuGet.PackageManagement.UI
             {
                 _versionInstalled = value;
                 OnPropertyChanged(nameof(InstalledVersion));
+            }
+        }
+
+        public PackageLevel? _packageLevel;
+        public PackageLevel? PackageLevel
+        {
+            get => _packageLevel;
+            set
+            {
+                _packageLevel = value;
+                OnPropertyChanged(nameof(PackageLevel));
             }
         }
 
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/PackageInstallationInfo.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/PackageInstallationInfo.cs
@@ -3,10 +3,12 @@
 
 using System;
 using System.ComponentModel;
+using System.Globalization;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.ServiceHub.Framework;
 using NuGet.PackageManagement.VisualStudio;
+using NuGet.Protocol;
 using NuGet.Versioning;
 using NuGet.VisualStudio;
 using NuGet.VisualStudio.Internal.Contracts;
@@ -156,7 +158,15 @@ namespace NuGet.PackageManagement.UI
 
         public override string ToString()
         {
-            return $"{ProjectName} {InstalledVersion?.Version.ToString() ?? Resources.Text_NotInstalled}";
+            var vulnerabilityString = string.Empty;
+            if (InstalledVersionMaxVulnerability != -1)
+            {
+                var converter = new IntToVulnerabilitySeverityConverter();
+                var severityString = converter.Convert(InstalledVersionMaxVulnerability, typeof(PackageVulnerabilitySeverity), null, CultureInfo.CurrentCulture) as string;
+                vulnerabilityString = string.Format(CultureInfo.CurrentCulture, Resources.Label_PackageVulnerableToolTip, severityString);
+            }
+
+            return $"{ProjectName} {InstalledVersion?.Version.ToString() ?? Resources.Text_NotInstalled} {vulnerabilityString} {PackageLevel}";
         }
 
         public event PropertyChangedEventHandler PropertyChanged;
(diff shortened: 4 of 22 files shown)


src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/IsGreaterThanToVisibilityConverter.cs
```diff
@@ -9,16 +9,20 @@
 
 namespace NuGet.PackageManagement.UI
 {
-    public class DownloadCountToVisibilityConverter : IValueConverter
+    public class IsGreaterThanToVisibilityConverter : IValueConverter
```
kartheekp-ms: For some reason, I am having a hard time understanding the name of the class :) If my understanding is correct, the visibility of some control in the UI is dependent on the count. If the count is greater than the some threshold the control will be enabled. Is that correct? Maybe we can try VS Copilot's smart rename feature.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/SolutionView.xaml
```diff
@@ -305,23 +314,115 @@
...
+            </GridViewColumn.Header>
+            <GridViewColumn.CellTemplate>
+              <DataTemplate>
+                <StackPanel
+                  VerticalAlignment="Bottom"
+                  Orientation="Horizontal">
+                  <TextBlock
+                    VerticalAlignment="Bottom"
+                    Text="{Binding InstalledVersion, Converter={StaticResource VersionToStringConverter}}" />
+                  <imaging:CrispImage
```
donnie-msft, line 362: Could the warning icon be put before the installed version? This would reduce visual clutter by aligning the icons. Also, it's more consistent with the positioning in the packages list (warning icon on the left of installed version).


_rough mockup (left is from this PR, right is my suggestion):_
image

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/SolutionView.xaml.cs
```diff
@@ -239,10 +240,21 @@ private void ListView_SizeChanged(object sender, SizeChangedEventArgs e)
             // adjust the width of the "project" column so that it takes
             // up all remaining width.
             var gridView = (GridView)_projectList.View;
-            var width = _projectList.ActualWidth - 3 * SystemParameters.VerticalScrollBarWidth;
+            var width = _projectList.ActualWidth - 8;
+
+            // If there are more than 7 projects the scrollbar will be visible
+            if (_projectList.Items.Count > 7)
```
donnie-msft, line 246: Could the screenshots from testing be added to this PR so we can see how this works with different resolutions/etc? I'm not sure if we need PLOC testing, but maybe ensuring the minimum resolution supported looks as expected would be good enough.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/InstalledAndTransitivePackageFeed.cs
```diff
@@ -29,16 +29,84 @@ public override async Task<SearchResult<IPackageSearchMetadata>> ContinueSearchA
...
+        private static IEnumerable<PackageCollectionItem> GetLatestPackages(IEnumerable<PackageCollectionItem> packages)
+        {
+            return packages
+                .GroupById()
+                .Select(g => g.OrderByDescending(x => x.Version).First());
+        }
+
+        private static async Task<IPackageSearchMetadata[]> GetRemainingPackagesMetadataAsync(IEnumerable<PackageCollectionItem> packages, FeedSearchContinuationToken searchToken, CancellationToken cancellationToken)
+        {
+            var remainingPackages = packages
```
donnie-msft, line 70: I believe the installed packages are already grouped by ID here: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Common/PackageCollection.cs#L64

src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/IsGreaterThanToVisibilityConverter.cs
jebriede: You are correct, @kartheekp-ms. The Converter checks if the value is greater than a threshold to determine visibility. Perhaps GreaterThanThresholdToVisibilityConverter would be a better name? Thoughts?

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Feeds/InstalledAndTransitivePackageFeedTests.cs
```diff
@@ -232,8 +232,13 @@ public async Task SearchAsync_WhenMultitargetingAndEmptySearch_ReturnsOnePackage
                 e => Assert.Equal(new PackageIdentity("packageA", new NuGetVersion("1.0.0")), e.Identity),
                 e => Assert.Equal(new PackageIdentity("packageB", new NuGetVersion("2.0.0")), e.Identity),
                 e => Assert.Equal(new PackageIdentity("packageC", new NuGetVersion("3.0.0")), e.Identity),
+                // Returns all the installed packages that are not the latest version
+                e => Assert.Equal(new PackageIdentity("packageB", new NuGetVersion("1.0.0")), e.Identity),
+                e => Assert.Equal(new PackageIdentity("packageC", new NuGetVersion("2.0.0")), e.Identity),
                 // Returns latest transitive package version
-                e => Assert.Equal(new PackageIdentity("transitivePackageC", new NuGetVersion("0.0.2")), e.Identity));
+                e => Assert.Equal(new PackageIdentity("transitivePackageC", new NuGetVersion("0.0.2")), e.Identity),
+                // Returns all the transtive packages that are not the latest verion
```
donnie-msft, line 240: Could the comment be expanded to explain why this assertion is made?

```suggestion
                // Returns all the transitive packages that are not the latest version because.....
```

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
```diff
@@ -136,6 +137,28 @@ public string ByOwnerOrAuthor
...
+        public Dictionary<NuGetVersion, int> VulnerableVersions
+        {
+            get
+            {
+                return _vulnerableVersions;
+            }
+            set
+            {
+                _vulnerableVersions = value;
+                OnPropertyChanged(nameof(VulnerableVersions));
```
donnie-msft: Is this necessary? I don't see where the entire dictionary is being replaced, so I don't think it's being used.

donnie-msft (commented):
Great to see this being added, nice work!
I'd like to see some accessibility testing to ensure we're not introducing a11y bugs for the new UI elements (https://github.com/NuGet/NuGet.Client/blob/dev/docs/ui-guidelines.md#accessibility-testing).

My main UI behavior concern is whether the calculations for columns and scrolling works well at different resolutions.

More unit tests to cover more of the package feeds and Services logic changes would be good as well.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/SolutionView.xaml
martinrrm: When I presented this to the UX board the warning icons where on the left* side, but their suggestion was to move to the right* side so it matches the Installed tab warning icon at the top

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/InstalledAndTransitivePackageFeed.cs
jebriede: The grouping in PackageCollection is grouping by PackageIdentity which is an object that consists of the Id (name of package) and Version. Here, we are grouping by the more specific Id only to sort the versions per package Id and skip the first one.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/SolutionView.xaml
jebriede: @martinrrm do you mean vice versa? The icons are currently on the right side and I believe when you presented this initially you had them on the left and the advice was to put the warning icons on the right side. Please correct me if I'm wrong.

martinrrm: Mb, I corrected the message, but yeah they suggested to change it so both installed "headers" have the icon at the same place, and have them mixed (rows with icon to the left and header to the right) would look weird

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/SolutionView.xaml.cs
martinrrm: image

martinrrm: Above image is how it looks with a low resolution. Details pane has a minimum width so windows size doesn't really affect the table content.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/SolutionView.xaml
donnie-msft: Ok, in that case, the Installed tab could also have probably been changed. We still have 2 different approaches, again, the packages list shows on the left.

image

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/InstalledAndTransitivePackageFeed.cs
donnie-msft: > by the more specific Id 

I'm not following what's more specific? The package ID should be unique.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/SolutionView.xaml.cs
donnie-msft: Great! I think this should be in the PR description as well.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
jebriede: Good point. I think we can delete the Setter altogether since the Dictionary is instantiated and only added to and this setter is never used. Then, the private _vulnerableVersions can be readonly.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/SolutionView.xaml.cs
martinrrm: https://github.com/NuGet/NuGet.Client/pull/6051
Realized that there was missing localization and narrator

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/SolutionView.xaml
martinrrm: Thanks Donnie, that a great suggestion, we have been getting UX approval for specific features and not considering the PM UI consistency. We should probably get back to the UX board and look at the whole PM UI. During my last UX meeting we discussed a lot of opportunities we could do improve the PM UI experience, I'll send a link to the meeting recording.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/SolutionView.xaml.cs
martinrrm: Done! Added this and accessibility testing to the PR description

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/InstalledAndTransitivePackageFeed.cs
jebriede: Grouping by PackageIdentity will group on PackageId + PackageVersion.

Here we are grouping on PackageId only so we can get a list of PackageVersions grouped by PackageId.

As an example if we have Newtonsoft.Json versions 13.0.0 and 12.0.0 installed across the solution, there will be two PackageIdentity objects like this:
PackageIdentity { Id = "Newtonsoft.Json", Version = "13.0.0" }
PackageIdentity { Id = "Newtonsoft.Json", Version = "12.0.0" }

When we group by Id here, we will group all Newtonsoft.Json Ids and have a list of the Versions [ "13.0.0", "12.0.0" ]. Then, we sort them and skip the first one per each package Id and process the rest.

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Feeds/InstalledAndTransitivePackageFeedTests.cs
martinrrm: So this PackageCollection is sorted, first we display the Installed Packages and then the TransitivePackages, and in each "group" they are sorted by latest. 

In this test I'm trying to test that they are sorted as expected

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
```diff
@@ -262,85 +263,105 @@ public IEnumerable<PackageItemViewModel> GetCurrent()
...
                     {
-                        allowedVersions = allowedVersionsRange[0];
+                        UpdateTransitiveInfo(existingListItem, metadataContextInfo);
                     }
 
-                    if (versionOverrides.Length > 0)
-                    {
-                        versionOverride = versionOverrides[0];
-                    }
+                    existingListItem.UpdateInstalledPackagesVulnerabilities(new PackageIdentity(packageId, packageVersion));
```
jebriede: @martinrrm Am I remembering correctly that we changed this to accept metadataContextInfo.Identity rather than instantiate a new PackageIdentity? I'm wondering how this made its way back here.

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
```diff
@@ -181,18 +211,20 @@ private async Task UpdateInstalledVersionsAsync(CancellationToken cancellationTo
         /// This method is called from several methods that are called from properties and LINQ queries
         /// It is likely not called more than once in an action.
         /// </summary>
-        private async Task<IPackageReferenceContextInfo> GetInstalledPackageAsync(
+        private async Task<(IPackageReferenceContextInfo TopLevelPackage, IPackageReferenceContextInfo TransitivePackage)> GetInstalledAndTransitivePackagesAsync(
```
jgonz120, line 214: Trying to make sure I'm understanding things here, this is returning the package information for a project and a specific packageId?

Is it possible for a single packageId to be both a toplevel and transitive package for a project? If not then I think this should be renamed to `GetInstalledOrTransitivePackageAsync`. 

`private async Task<(IPackageReferenceContextInfo Package, bool IsTransitive)> GetPackageContextAsync`

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageInstallationInfo.cs
```diff
@@ -49,6 +50,26 @@ private async ValueTask InitializeAsync(CancellationToken cancellationToken)
         private NuGetVersion _versionInstalled;
         private string _versionRequested;
 
+        private int _installedVersionMaxVulnerability = -1;
+        public int InstalledVersionMaxVulnerability
```
jgonz120, line 56: I think I might have already asked this, what does this mean?😅

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
```diff
@@ -854,22 +956,22 @@ public void UpdatePackageStatus(IEnumerable<PackageCollectionItem> installedPack
...
             NuGetUIThreadHelper.JoinableTaskFactory
-                .RunAsync(ReloadPackageMetadataAsync)
-                .PostOnFailure(nameof(PackageItemViewModel), nameof(ReloadPackageMetadataAsync));
+                .RunAsync(ReloadTopLevelPackageMetadataAsync)
+                .PostOnFailure(nameof(PackageItemViewModel), nameof(ReloadTopLevelPackageMetadataAsync));
 
             OnPropertyChanged(nameof(Status));
         }
 
         public void UpdateTransitivePackageStatus(NuGetVersion installedVersion)
```
jgonz120, line 960: Would it make sense for the PackageItemViewModel to keep track of whether or not the package is transitive?

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
```diff
@@ -839,6 +911,36 @@ private async Task ReloadPackageMetadataAsync()
...
         }
 
+        private async Task UpdatePackageMaxVulnerabilityAsync(PackageIdentity packageIdentity, CancellationToken cancellationToken)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            IEnumerable<PackageVulnerabilityMetadataContextInfo> vulnerabilityInfoList =
+                        await _vulnerabilityService.GetVulnerabilityInfoAsync(packageIdentity, cancellationToken);
+
+            SetVulnerabilityMaxSeverity(packageIdentity.Version, vulnerabilityInfoList?.FirstOrDefault()?.Severity ?? -1);
```
jgonz120, line 911: If you're getting a list of vulnerabilities, would it make sense to select the one with the highest severity, instead of just the first, for setting the max severity?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/InstalledAndTransitivePackageFeed.cs
```diff
@@ -29,16 +29,84 @@ public override async Task<SearchResult<IPackageSearchMetadata>> ContinueSearchA
             cancellationToken.ThrowIfCancellationRequested();
 
             // Remove transitive packages from project references. Those transitive packages don't have Transitive Origins
-            IEnumerable<PackageCollectionItem> transitivePkgsWithOrigins = _transitivePackages
-                .Where(t => t.PackageReferences.All(x => x is ITransitivePackageReferenceContextInfo y && y.TransitiveOrigins.Any()));
+            IEnumerable<PackageCollectionItem> transitivePkgsWithOrigins = GetTransitivePackagesWithOrigins(_transitivePackages);
+
+            // Get the metadata for the latest version of the installed packages and transitive packages
+            IEnumerable<PackageCollectionItem> installedPackagesLatest = GetLatestPackages(_installedPackages);
```
jgonz120, line 35: Nit: can this be updated to match the other variable name? `latestInstalledPackages`

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
jebriede: @jgonz120 It's not possible for a package to be both top-level and transitive in a single project. It's only possible for a package to be top-level and transitive in a across the solution, i.e. top-level in one project, transitive in another.

@martinrrm I think the suggested rename is a good idea because it looks like we're getting package information for a project/package combo. Could you confirm this is what we're doing and what you think of the proposed rename?

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
jebriede: It does in PackageLevel. I know, this method should not be public and the caller should not need to call the appropriate UpdatePackageStatus/UpdateTransitivePackageStatus method, and we should instead call UpdatePackageStatus and the ViewModel should do the right thing depending on whether it's top-level or transitive. However, this code existed already and at this point I want to make as few changes as possible so we can complete this work and make that enhancement later.

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
jgonz120: I put this `private async Task<(IPackageReferenceContextInfo Package, bool IsTransitive)> GetPackageContextAsync` and didn't explain it.

If only one package will ever be returned we can simplify the return with a bool to indicate if the package found is transitive or top level.

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageInstallationInfo.cs
jebriede: @jgonz120 it corresponds with:
https://github.com/NuGet/NuGet.Client/blob/1e0115a608073be2ad615201be30b93dd628300d/src/NuGet.Core/NuGet.Protocol/Model/PackageVulnerabilitySeverity.cs#L6

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
jebriede: @donnie-msft I addressed this now in the latest commit. I moved the OnPropertyChanged to a place where we add items to the VulnerableVersions dictionary, and also added an OnPropertyChanged for the VulnerableVersionsString which was previously missing. I removed the setter here because it was not being used. Thank you!

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
jebriede: I fixed this in the latest commit.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
martinrrm: So, I wasn't able to locate where we explain this, but my understanding is that they are already sorted by severity when we query them from the database so there is no need to "sort" again.

[This code was just moved into another function](https://github.com/NuGet/NuGet.Client/pull/6044/files/85f658abf961201b1ba8dea5509aa913e61b2774#diff-118d6d1c3dcb2549a9436b8dafadcf915e0ae147a2ed54f3d82767252ee94701L817), but the implementation is the same and is being used in project PM UI too.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/SolutionView.xaml
donnie-msft: Look forward to seeing that! In the meantime, if this is "ok" with UX board, then I think it could be easily changed later and not be too jarring to customers.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
jgonz120: Sounds good to me!

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
jgonz120: Or even `private async Task<(IPackageReferenceContextInfo Package, PackageLevel PackageLevel)> GetPackageContextAsync`

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/SolutionView.xaml
jgonz120: Not for this PR, but could it make sense to move the vulnerability information to a column of its own? This would help align the icons and we could also add additional icons to denote the max severity of the vulnerabilities. 

✅ - No Vul
⚠️ - Low/Moderate/High
🔴 - Critical 
❔ - Unkown

jgonz120 (dismissed):
Looks good to me!

src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/GreaterThanThresholdToVisibilityConverter.cs
```diff
@@ -9,16 +9,20 @@
...
         public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
         {
             if (targetType == typeof(Visibility))
             {
-                long? downloadCount = value as long?;
-                if (downloadCount >= 0)
+                if (value is not null)
                 {
-                    return Visibility.Visible;
+                    long intValue = System.Convert.ToInt64(value, culture);
```
kartheekp-ms: Convert.ToInt64() will throw an exception if the value is not a valid long value. The example in the [doc](https://learn.microsoft.com/dotnet/api/system.convert.toint64?view=net-8.0#system-convert-toint64(system-object)) has catch blocks for various exceptions this method can throw.

kartheekp-ms: nit: The variable name `intValue` is not clear to me. Are we referring to the initial value?

jgonz120: I'm guessing this is done so an object or a string can be converted to the correct value. Would it worth creating a `bool TryConvertToInt64(obj, formatter, out long? convertedValue)` method somewhere?

jgonz120 (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/Converters/GreaterThanThresholdToVisibilityConverter.cs
jebriede: @kartheekp-ms do you have a suggestion for another name?

jebriede: I am making it `initialValue` and `thresholdValue`

jebriede: @kartheekp-ms thanks for the feedback. @jgonz120 I like this suggestion. I've got this ready and will push changes in a moment.

jebriede: @kartheekp-ms @jgonz120 pushed an update that addresses this case and has a new test for invalid conversion scenarios.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
```diff
@@ -136,6 +138,17 @@ public string ByOwnerOrAuthor
...
+        public string VulnerableVersionsString
+        {
+            get
+            {
+                return string.Join(", ", VulnerableVersions.Keys);
+            }
+        }
+
+        private readonly Dictionary<NuGetVersion, int> _vulnerableVersions = [];
+        public Dictionary<NuGetVersion, int> VulnerableVersions => _vulnerableVersions;
```
donnie-msft, line 150: Is anything actually listening to VulnerableVersions ? Maybe I'm missing it in a binding somewhere?

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
```diff
@@ -136,6 +138,17 @@ public string ByOwnerOrAuthor
...
 
+        public string VulnerableVersionsString
+        {
+            get
+            {
+                return string.Join(", ", VulnerableVersions.Keys);
+            }
+        }
+
+        private readonly Dictionary<NuGetVersion, int> _vulnerableVersions = [];
```
donnie-msft, line 149: Is this field being used? I think it doesn't need both the field & property.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
```diff
@@ -839,6 +901,41 @@ private async Task ReloadPackageMetadataAsync()
...
+        }
+
+        private void SetVulnerabilityMaxSeverity(NuGetVersion version, int maxSeverity)
+        {
+            if (maxSeverity > -1)
+            {
+                if (VulnerableVersions.TryAdd(version, maxSeverity))
+                {
+                    OnPropertyChanged(nameof(VulnerableVersions));
+                    OnPropertyChanged(nameof(VulnerableVersionsString));
```
donnie-msft, line 921: nit: Ideally, a CollectionChanged event could be used to fire the appropriate OnPropertyChanged events when the collection gets changed.

jgonz120 (approved):

test/NuGet.Clients.Tests/NuGet.PackageManagement.VisualStudio.Test/Feeds/InstalledAndTransitivePackageFeedTests.cs
donnie-msft: If they are sorted, why is transitivePackageC 0.0.2 asserted before version 0.0.1 ?

Are you planning to fix the typos I suggested?

kartheekp-ms:
Looking at the screenshot mentioned in the PR description, I think `Uninstall` button should be invisible for transitive packages. Only `Install` button might be applicable because the install action promotes a transitive dependency to top-level dependency. I am unsure if this feedback is out of scope for this PR.

image

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
```diff
@@ -262,85 +262,105 @@ public IEnumerable<PackageItemViewModel> GetCurrent()
                 return Enumerable.Empty<PackageItemViewModel>();
             }
 
-            var listItemViewModels = new List<PackageItemViewModel>(capacity: _state.ItemsCount);
+            var listItemViewModels = new Dictionary<string, PackageItemViewModel>();
```
donnie-msft, line 265: I'm trying to understand what the dictionary is being used for. Could someone explain how the same package would exist in `_state.Results.PackageSearchItems` multiple times?  Ideally, a comment would be good here as well.

donnie-msft (approved):
Great work and can't wait to see how customers use this! 
Left some questions and nits.

jebriede:
> Looking at the screenshot mentioned in the PR description, I think `Uninstall` button should be invisible for transitive packages. Only `Install` button might be applicable because the install action promotes a transitive dependency to top-level dependency. I am unsure if this feedback is out of scope for this PR.

@kartheekp-ms We already fixed this. I thought the PR was completed but it looks like it's not merged yet. This fixes the issue: https://github.com/NuGet/NuGet.Client/pull/6050. I'll ask @martinrrm to retarget dev and submit this fix after this PR goes in.

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageItemLoader.cs
jebriede: @donnie-msft the changes made here: https://github.com/NuGet/NuGet.Client/pull/6044/files#diff-b1b2dd9d0be30f84d9309d08db86156cff27b5df8e43b54c13e90bfa5fbf72f9R38-R43 are related to this.

Let's see if I can explain this in a way that makes sense in a comment. :-) (This is actually the main reason I did the live code review meeting to explain this part). The Solution PM UI needs to know about all the versions installed across the solution, in each project, not just the latest version. For the Project PM UI, this will always only be the latest version because of how restore works. But limiting it to just the latest doesn't work in the solution because we can have the same package installed at different versions in different projects and since one of those can be vulnerable, not just the latest, we need to know all of the package versions for each package. As such, the `InstalledAndTransitivePackageFeed` will now return all the package versions. However, we want to create a single PackageItemViewModel so it shows up only once in the packages list. We keep track of the PackageItemViewModels we've created for that Package Id (the string key) and retrieve the view model if it was already created, so we can add the additional versions to it.

I hope this explanation helps! I can probably explain it better in real time so if you have questions or want a deeper dive, please set up a quick meeting and we'll go over this in more detail!

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
jebriede: It's used here: https://github.com/NuGet/NuGet.Client/pull/6044/files#diff-5c15ee840d9063ec5f345761223159c7ecde0f058e0e0190e359adcfee761290R170

It was also supposed to be used in the XAML for the VulnerableVersions.Count but there's typo. I'm fixing the XAML binding now.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
jebriede: That's a really good point. This should be an observable set / dictionary that fires CollectionChanged when it gets updated, and we listen to that and fire any additional OnPropertyChanged as needed.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
jebriede: Fixing in: https://github.com/NuGet/NuGet.Client/pull/6056

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
jebriede: @jgonz120: I created https://github.com/NuGet/NuGet.Client/pull/6065 to incorporate your feedback. Please let me know if that's what you had in mind and leave any feedback on that PR. Thanks!

PR #6047, opened 2024-09-23
Add telemetry for service entries

### LINKED ISSUE #13744: Collect telemetry for HTTPS sources with HTTP resources
Collect telemetry for HTTPS sources with HTTP resources in their service index

Nigusu-Allehu:
# Bug

Fixes:  https://github.com/NuGet/Home/issues/13744

### Telemetry Assessment: https://github.com/NuGet/Client.Engineering/pull/3034

This PR adds telemetry for service index resources that are HTTP, but their source is HTTPS.
```json
{
    "version": "3.0.0",
    "resources": [
        {
            "@id": "http://source",
        }]
}
```
The telemetry data logged is as follows

**Event name:**  &lt;`vs/nuget/RestorePackageSourceSummary`&gt; 
**Event name:**  &lt;`vs/nuget/SearchPackageSourceSummary`&gt; 

**Event Details**

Property | Description
----|----
NumHTTPSSourcesWithAnHTTPResource | The number of HTTPS NuGet sources that have at least one HTTP resource.
ServiceIndexCacheAvailable | Is the cache for the service index available. This will tell us whether `NumHTTPSSourcesWithAnHTTPResource ` value is valid or not when it is 0.

## PR Checklist

- [ ] Added tests


src/NuGet.Core/NuGet.Protocol/Model/ServiceIndexEntry.cs
```diff
@@ -26,6 +30,37 @@ public class ServiceIndexEntry
...
+                {
+                    source = _packageSource.Source;
+                }
+                catch (UriFormatException)
+                {
+                    source = _packageSource.Source;
+                }
+
+                var telemetry = new ServiceIndexEntryTelemetry(source, _packageSource.IsHttp, _packageSource.IsHttps, isResourceHTTP, isResourceHTTPs, type, "ServiceIndexEntrySummary");
+                TelemetryActivity.EmitTelemetryEvent(telemetry);
```
zivkan: Firstly, from https://github.com/NuGet/NuGet.Client/blob/dev/docs/nuget-sdk.md#changes-to-apis

I don't feel like this is a good reason to introduce a new public API.

Secondly, I have a personal preference that only one constructor should have an empty body.  So, instead of adding a new constructor that invokes the old instructor, and then does extra work, change the old constructor to call the new constructor, and move all the code from the old constructor to the new one.  This way there's only 1 constructor with code, and there's less chance of bugs when someone makes a change and breaks one of the constructors.

Finally, instances of this class are going to be created all at the same time, so it will flood VS with events all at once. NuGet.org's service index has 39 entries, by my count.

We already have a lot of telemetry being throttled and therefore lost. I imagine that sending 39 entries, just for nuget.org (plus tens more for each other package source) will make telemetry throttling much worse.  It's better to emit a single event, which contain a whole lot of properties (or one "complex property" with a lot of data) from ServiceIndexResourceV3, or ServiceIndexV3Provider. That will still emit one event per package source, so honestly, I'd prefer telemetry at an even higher level where there can be a single event with a big payload.

Look at [dotnet/sdk](https://github.com/dotnet/sdk/blob/main/NuGet.config) for example, it has 24 sources configured. and looking at the first service index, it has 11 entries. So, that's probably 264 events that this PR will currently try to emit.

Honestly though, I'd prefer it if this telemetry was either in RestoreCommand, or somewhere in `src\NuGet.Clients\*`, not in NuGet.Protocol. In order to make that work, the telemetry code will need to `sourceRepository.GetResourceAsync<ServiceIndexV3Resource>()` and then iterate the entries. But this has a problem when the service index hasn't already been fetched. If it hasn't already been fetched, then just by requesting the resource, it will fetch it, and we don't want to make unnecessary HTTP requests. Especially if the server needs authentication, and valid credentials aren't provided (and can't be obtained from a cred provider), then NuGet will cause a UI popup asking for credentials. We definately don't want to create a modal dialog for credentials just to gather some telemetry.

Perhaps we could introduce a `LazyServiceIndexV3Resource`, which, similar to the `Lazy<T>` type, has some kind of `HasValue` property, so the telemetry can get this resource, without causing a download, and then change `ServiceIndexV3Resource` to get the lazy resource and immediately get its value.  If ServiceIndexV3Resource didn't have `Entries` and `Json` properties, we could have just modified it to add a `HasEntries` property.  It's best to talk to a few people about the idea first, rather than spending a lot of time implementing something and then having different people give you conflicting feedback.

src/NuGet.Core/NuGet.Protocol/CachingSourceProvider.cs
```diff
@@ -67,7 +67,20 @@ public SourceRepository CreateRepository(PackageSource source, FeedType type)
...
+            var existingSource = _repositories.FirstOrDefault(r => r.PackageSource.Source == source.PackageSource.Source);
+
+            if (existingSource != null)
+            {
+                _repositories[_repositories.IndexOf(existingSource)] = source;
+            }
+            else
+            {
+                _repositories.Add(source);
+            }
```
Nigusu-Allehu: The reason for this issue is that we do not reuse the same `sourceProvider` in [DependencyGraphRestoreUtility.cs](https://github.com/NuGet/NuGet.Client/blob/b43ffa23bd5e6d0261a9d951c0e70483126f9c55/src/NuGet.Core/NuGet.PackageManagement/BuildIntegration/DependencyGraphRestoreUtility.cs#L420). When a new `sourceProvider` is created, the service index cache stored in the previous service index provider becomes inaccessible at higher levels in the code.

The `AddSourceRepository` method is intended to copy the source repositories from the old `sourceProvider` to the new one. However, it does not work as expected. Instead of copying the existing source repositories, the new `sourceProvider` creates its own repositories and stores them in the `_repositories` dictionary. When we attempt to copy the old repositories into the new provider, it fails because the new provider has already created repositories with the same names and values, but as distinct objects. Trying to add the old repository objects to `_repositories` will fail due to key conflicts with the already existing repositories.

In this pull request, I ensure that the `_repositories` dictionary overwrites existing entries when an item with the same key is already present. This allows for a smooth continuation of repository references.

src/NuGet.Core/NuGet.Protocol/Model/ServiceIndexEntry.cs
Nigusu-Allehu: Thank you for the review. As per our conversation offline, i have updated the PR
* to emit one event for all sources
* make use of the service index cache as a source of data

src/NuGet.Core/NuGet.Protocol/PublicAPI/net472/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,2 @@
 #nullable enable
+~NuGet.Protocol.Core.Types.SourceRepository.GetServiceIndexV3FromCache(out NuGet.Protocol.ServiceIndexResourceV3 result) -> bool
```
nkolev92: This API doesn't really seem ideal.
It's referring to a specific resource while SourceRepository tends to be a more generic resource.

src/NuGet.Clients/NuGet.VisualStudio.Common/Telemetry/PackageSourceTelemetry.cs
```diff
@@ -209,13 +213,64 @@ public async Task SendTelemetryAsync()
...
+                    {
+                        if (sourceUri.Scheme == Uri.UriSchemeHttps)
+                        {
+                            NumberOfHTTPSSources++;
+                        }
+                    }
+                }
+
+
+                if (sourceRepository.GetServiceIndexV3FromCache(out ServiceIndexResourceV3 resource))
```
nkolev92: Would the protocol diagnostics event be more appropriate instead of going through this method?

@zivkan wrote up a documented on that at some point but I can't find it now.

src/NuGet.Core/NuGet.Protocol/CachingSourceProvider.cs
nkolev92: These are some interesting findings. 

> The AddSourceRepository method is intended to copy the source repositories from the old sourceProvider to the new one.

Can you help me understand how you got to this conclusion?
CachingSourceProvider tends to be used for consolidating of source repository during a restore operation. 

> However, it does not work as expected

What are the consequences?
Any user impacting issues?

Nigusu-Allehu: > **Can you help me understand how you got to this conclusion? CachingSourceProvider tends to be used for consolidating of source repository during a restore operation.**

Yes, the `CachingSourceProvider` is used for consolidating source repositories during a restore operation. The caching source provider used by restore is created [here](https://github.com/NuGet/NuGet.Client/blob/b43ffa23bd5e6d0261a9d951c0e70483126f9c55/src/NuGet.Core/NuGet.PackageManagement/BuildIntegration/DependencyGraphRestoreUtility.cs#L420). However, when we create it, we do not reuse the package source provider that has been created at a [higher level](https://github.com/NuGet/NuGet.Client/blob/0bf00d2496ec6de66fe67b1c8f571090ac2eef87/src/NuGet.Clients/NuGet.SolutionRestoreManager/SolutionRestoreJob.cs#L123). Instead, we create a whole new provider. This new provider creates source repositories by using the sources in `context.Settings`.

The [`AddSourceRepository`](https://github.com/NuGet/NuGet.Client/blob/b43ffa23bd5e6d0261a9d951c0e70483126f9c55/src/NuGet.Core/NuGet.PackageManagement/BuildIntegration/DependencyGraphRestoreUtility.cs#L424) method is used only once in our code. It's intended to add source repositories that have been created before—for example, those in `SolutionRestoreJob._sourceRepositoryProvider`. We try to use `AddSourceRepository` to take the repositories from `_sourceRepositoryProvider` and add them to the new cache provider we've created. However, this is always a no-op if `SolutionRestoreJob._sourceRepositoryProvider` has the same sources configured as `context.Settings`. If both have the same sources configured, then both will have the same repositories configured. `AddSourceRepository` cannot add them to the cache source provider because sources with the same name (key) already exist in the cache.

> **What are the consequences? Any user impacting issues?**

The direct consequence I've observed is that the cache source provider and the source provider at a higher level end up with different objects for the source repositories. This makes the source repository cache generated during restore unavailable at higher levels like `SolutionRestoreJob`. I haven't been able to confirm any direct consequences for NuGet customers, and I don't think there are any because this is occurring at the solution level. However, if this were happening at the project level, it could result in multiple authentication requests per source.

src/NuGet.Core/NuGet.Protocol/PublicAPI/net472/PublicAPI.Unshipped.txt
Nigusu-Allehu: Maybe GetResourceFromCacheInstead? and allow any resource type that has been downloaded to be retrieved?

src/NuGet.Clients/NuGet.VisualStudio.Common/Telemetry/PackageSourceTelemetry.cs
```diff
@@ -209,13 +213,64 @@ public async Task SendTelemetryAsync()
...
 
                 if (telemetry != null)
                 {
                     TelemetryActivity.EmitTelemetryEvent(telemetry);
                 }
             }
+
+            if (sendResourceTelemetry)
+            {
+                var resourceTelemetry = new ServiceIndexHttpResourcesTelemetry(NumberOfHTTPSSources, NumberOfHTTPSSourcesWithHTTPResource, "ServiceIndexHttpResourcesSummary");
```
Nigusu-Allehu: new telemetry event.

src/NuGet.Core/NuGet.Protocol/CachingSourceProvider.cs
nkolev92: Nice sleuthing. Thanks for the explanation, that was really helpful!

I think the behavior of AddSourceRepository is more of an AddOrReplace now, so we should make sure we document that, maybe on the type. I wouldn't go in as much detail as here :D 

I'm guessing the consequences might be things like reading the ServiceIndex more than once. 
For example first time would download/write to cache, while the 2nd time would read from cache file on disk, while in theory it could just use the one that exists. 

That being said, if we're adding a behavior, it'd be great to add some tests. Seems like we're missing tests for the type altogether, so that could be improved. 

Another idea would be to add a new constructor here with the preinitialized soruce repositories so we don't bother creating source repositories needlessly. 
It'd also save on allocations. There might be other ways to save on allocations too, but this only runs once, so maybe simpler is better.

src/NuGet.Clients/NuGet.VisualStudio.Common/Telemetry/PackageSourceTelemetry.cs
nkolev92: If you want to send it in a summary event, there's also some that exist. https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Telemetry/SourceTelemetry.cs. 

It's also worth considering whether it's ok for package source diagnoistics to incldue that information. 

In the end, we'd care about how often sources have this behavior. 
Id' be shocked if it's not 0 tbh.

src/NuGet.Clients/NuGet.VisualStudio.Common/Telemetry/PackageSourceTelemetry.cs
nkolev92: Hoping @zivkan can find the doc for this discussion.
Wondering if we can use ProtocolDiagnostics instead. 

The idea is that something happens, data is reported, and we consume it use at a different level.

zivkan: I can't find a design for ProtocolDiagnostics itself, but I can find this doc on different "monitoring" APIs available to NuGet: https://github.com/NuGet/Client.Engineering/blob/main/designs/ProductMonitoring.md

The overview of how ProtocolDiagnostics works is that NuGet.Protocol have some (non-telemetry) events (so, c#'s `event` keyword, or maybe an ETW event) that get raised when potentially "interesting" things happen (http requests are made, or one of `SourceRepository`'s "resources" are used). Then the code in VS can subscribe to those events and react to them appropriately.  They way I implemented it is subscribe to the event at the start of restore, and each time the event fires, increment a counter in memory. Once restore finishes, ubsubscribe from the event, and emit telemetry about the counts that were accumulated. So, even though the events get fired many times during a restore, only a single telemetry event is sent.

I really like that Nikolche remembered protocol diagnostics and is thinking about whether it fits.  Personally, I'm not sure if it fits, but it's great to consider. To me, protocol diagnostics is really about when there are many protocol events happening, and you want to aggregate them into a single telemetry event.  Counting the number of https vs http resources in the service index doesn't feel the same, as it's a one-time count, rather than something that restore does potentially multiple times during a restore.

src/NuGet.Core/NuGet.Protocol/CachingSourceProvider.cs
```diff
@@ -36,6 +36,32 @@ public CachingSourceProvider(IPackageSourceProvider packageSourceProvider)
                 .ToList();
         }
 
+        public CachingSourceProvider(IEnumerable<SourceRepository> loadedRepositories, IPackageSourceProvider sourceProvider)
```
Nigusu-Allehu, line 39: Allows creating a cachingSourceProvider that uses already loaded repositories before creating new repositories from settings.

src/NuGet.Core/NuGet.Protocol/CachingSourceProvider.cs
Nigusu-Allehu: I have updated the constructor. We definitely need to add testing for the methods in this class.

test/NuGet.Core.Tests/NuGet.PackageManagement.Test/BuildIntegration/DependencyGraphRestoreUtilityTests.cs
```diff
@@ -724,7 +724,7 @@ public async Task PreviewRestoreAsync_PreviewNewSourceMapping_RestoreLogsError()
             // Arrange - Setup project
             var packageA = new SimpleTestPackageContext("packageA", "1.0.0");
             await SimpleTestPackageUtility.CreateFolderFeedV3Async(pathContext.PackageSource, packageA);
-            var sources = new PackageSource[] { new PackageSource(pathContext.PackageSource) };
+            var sources = new PackageSource[] { new PackageSource(pathContext.PackageSource, "source") };
```
Nigusu-Allehu, line 727: This change may appear unrelated to the purpose of this pull request, but it's necessary because a test was broken due to my modifications to [CachingSourceProvider.cs](https://github.com/NuGet/NuGet.Client/pull/6047/files#diff-78a1e090eecca2c0d70ee5dd2dffbc31c24bd04c86560eacb5c8fc6602d7330c).

The test utilizes package source mapping to prevent `packageA` from being restored, expecting an exception and therefore uses `Assert.Throws`. However, after my update to `CachingSourceProvider`, the test began to fail because the restore operation was succeeding instead of failing.

why?

In setting up the test, the source repository is created using `new PackageSource(pathContext.PackageSource)`. This constructor initializes a package source where both the name and source URI are set to `pathContext.PackageSource`. Since package source mapping is configured to exclude this source, we expect that it won't be used, leading the restore operation to fail. This behavior was consistent with the previous implementation because `CachingSourceProvider` disregarded any pre-configured `SourceRepositories`, instead creating new ones based on the sources defined in `settings`.

With my updates, `CachingSourceProvider` no longer discards existing `SourceRepositories`. It retains them and also adds any new ones configured in the settings. In the context of this test, the initially configured source repository has a name and URI equal to `pathContext.PackageSource`. Meanwhile, the `nuget.config` file generated during the test specifies a source repository with the name `"source"` and the same URI `pathContext.PackageSource`. This results in the same source being available under two different names.

Even though the test is designed to exclude the source named `"source"` through source mapping, the other source with the name `pathContext.PackageSource` remains available and is used, causing the restore operation to succeed. This unexpected success leads to the test failure.

To resolve this issue, I modified the test setup to ensure that the `SourceRepository` we configure has the same name and URI as the package source defined in the test's `nuget.config`. This alignment ensures that source mapping correctly excludes the source, causing the restore operation to fail as originally intended.

src/NuGet.Core/NuGet.Protocol/PublicAPI/netstandard2.0/PublicAPI.Unshipped.txt
```diff
@@ -1,2 +1,4 @@
 #nullable enable
+~NuGet.Protocol.CachingSourceProvider.CachingSourceProvider(System.Collections.Generic.IEnumerable<NuGet.Protocol.Core.Types.SourceRepository> loadedRepositories, NuGet.Configuration.IPackageSourceProvider sourceProvider) -> void
 ~NuGet.Protocol.Plugins.PluginDiscoverer.PluginDiscoverer(string rawPluginPaths) -> void
+~virtual NuGet.Protocol.Core.Types.SourceRepository.GetServiceIndexV3FromCache(out NuGet.Protocol.ServiceIndexResourceV3 result) -> bool
```
nkolev92, line 4: I might still be in the minority here, but I really don't think this particular method belongs here. 

Not every source repository is backed by a service index.
It feels like the other means the telemery is collected through events is better from an API design.

src/NuGet.Clients/NuGet.VisualStudio.Common/Telemetry/PackageSourceTelemetry.cs
Nigusu-Allehu: I used a similar pattern here https://github.com/NuGet/NuGet.Client/pull/6125

PR #6051, opened 2024-09-24
Update ProjectInstallationInfo ToString() on Details Pane table and improve localization of table headers

martinrrm:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13805

## Description
Narrator wasn't reading the whole table, we use the property ToString() in PackageInstallationInfo for that, now Narrator reads the vulnerability and the package level

image
image

## PR Checklist

~- [ ] Added tests~ Narrator tested

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/PackageInstallationInfo.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/PackageInstallationInfo.cs
@@ -161,7 +161,9 @@ namespace NuGet.PackageManagement.UI
             var vulnerabilityString = string.Empty;
             if (InstalledVersionMaxVulnerability != -1)
             {
-                vulnerabilityString = string.Format(CultureInfo.CurrentCulture, Resources.Label_PackageVulnerableToolTip, (PackageVulnerabilitySeverity)InstalledVersionMaxVulnerability);
+                var converter = new IntToVulnerabilitySeverityConverter();
+                var severityString = converter.Convert(InstalledVersionMaxVulnerability, typeof(PackageVulnerabilitySeverity), null, CultureInfo.CurrentCulture) as string;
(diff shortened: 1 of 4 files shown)


src/NuGet.Clients/NuGet.PackageManagement.UI/PackageInstallationInfo.cs
```diff
@@ -156,7 +158,13 @@ public override int GetHashCode()
 
         public override string ToString()
         {
-            return $"{ProjectName} {InstalledVersion?.Version.ToString() ?? Resources.Text_NotInstalled}";
+            var vulnerabilityString = string.Empty;
+            if (InstalledVersionMaxVulnerability != -1)
+            {
+                vulnerabilityString = string.Format(CultureInfo.CurrentCulture, Resources.Label_PackageVulnerableToolTip, (PackageVulnerabilitySeverity)InstalledVersionMaxVulnerability);
```
jebriede: The InstalledVersionMaxVulnerability enum values are not localized. We use IntToVulnerabilitySeverityConverter in the XAML to convert to a localized string. We should consider localization for this accessibility fix.

jebriede (changes requested):
The max severity in the table should be localized.

The title, which becomes the commit messages, should state we're updating the ToString for accessibility. It's not quite accurate that we're updating Narrator for the details pane, we're actually updating the ToString so that any screen reader (including Narrator) will read the details correctly.

src/NuGet.Clients/NuGet.PackageManagement.UI/PackageInstallationInfo.cs
martinrrm: If im not wrong when we use CultureInfo it localizes the string.

I updated the code to use the converter, not sure if this is the best way to do it. We could also copy the code from the converter into this file.

jebriede (approved):

PR #6055, opened 2024-09-25
Update ProjectSystemCache to be case insensitive for _primaryCache

jgonz120:
# Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2829

## Description
Projects information was being added/updated with different cases, causes our apex tests to fail. The primary key for the _primaryCache is projectNames.FullName which is the full path and filename of the project. Switching the dictionary to ignore case allows the project to be added to the single entry instead of creating a new one. 

When we load an empty solution, there are no projects, and [VsSolutionManager.EnsureNuGetAndVsProjectAdapterCacheAsync](https://github.com/NuGet/NuGet.Client/blob/0bf00d2496ec6de66fe67b1c8f571090ac2eef87/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSSolutionManager.cs#L746) does not initialize the cache. 

When we create a new project we call [VsSolutionManager.OnEnvDTEProjectAdded](https://github.com/NuGet/NuGet.Client/blob/0bf00d2496ec6de66fe67b1c8f571090ac2eef87/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/IDE/VSSolutionManager.cs#L693C22-L693C42) which attempts to initialize the cache again then add the project. 

EnsureNuGetAndVsProjectAdapterCacheAsync gets the project name from
https://github.com/NuGet/NuGet.Client/blob/0bf00d2496ec6de66fe67b1c8f571090ac2eef87/src/NuGet.Clients/NuGet.VisualStudio.Common/IDE/VsHierarchyUtility.cs#L37-L43

Then later in OnEnVDTEProjectedAdded we get the full name from the EnvDTE.Project full name
https://github.com/NuGet/NuGet.Client/blob/0bf00d2496ec6de66fe67b1c8f571090ac2eef87/src/NuGet.Clients/NuGet.VisualStudio.Common/ProjectSystems/ProjectNames.cs#L92

## PR Checklist

- [x] Added tests
- [x] ~~Link to an issue or pull request to update docs if this PR changes settings, environment variables, new feature, etc.~~

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/ProjectSystemCache.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/ProjectSystemCache.cs
@@ -22,7 +22,7 @@ namespace NuGet.PackageManagement.VisualStudio
         // 0 - Cache is clean
         // 1 - Cache is dirty
         private int _isCacheDirty = 0;
-        private readonly Dictionary<string, CacheEntry> _primaryCache = new Dictionary<string, CacheEntry>();
+        private readonly Dictionary<string, CacheEntry> _primaryCache = new Dictionary<string, CacheEntry>(StringComparer.OrdinalIgnoreCase);
         private readonly ReaderWriterLockSlim _readerWriterLock = new ReaderWriterLockSlim();
 
         // Secondary index. Mapping from all names to a project name structure
--- a/test/NuGet.Tests.Apex/NuGet.Tests.Apex.Daily/NuGetUITestCase.cs
+++ b/test/NuGet.Tests.Apex/NuGet.Tests.Apex.Daily/NuGetUITestCase.cs
@@ -25,7 +25,6 @@ namespace NuGet.Tests.Apex.Daily
         {
         }
 
-        [Ignore("https://github.com/NuGet/Client.Engineering/issues/2829")]
         [TestMethod]
         [Timeout(DefaultTimeout)]
         public void InstallPackageToWebSiteProjectFromUI()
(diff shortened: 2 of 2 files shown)


Nigusu-Allehu (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/ProjectSystemCache.cs
```diff
@@ -22,7 +22,7 @@ internal class ProjectSystemCache : IProjectSystemCache, IDisposable
         // 0 - Cache is clean
         // 1 - Cache is dirty
         private int _isCacheDirty = 0;
-        private readonly Dictionary<string, CacheEntry> _primaryCache = new Dictionary<string, CacheEntry>();
+        private readonly Dictionary<string, CacheEntry> _primaryCache = new Dictionary<string, CacheEntry>(StringComparer.OrdinalIgnoreCase);
```
nkolev92, line 25: Discussed offline. 

This runs on windows only. Windows is case insensitive  ve, and the other caches are case insensitive already, so it's likely not a problem. 

The concern would be, is this hiding some inconsistency that could bite us in other ways. Is it our APIs? project-system etc?

This might be easiest time to figure that out. 
Given that ProjectSystemCache is a MEF type, maybe we try importing https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.VisualStudio.Common/OutputConsoleLogger.cs and logging when we see the inconsistency. 

 In isolation, this change is likely good, but maybe there's an easy way to confirm 100%.

jgonz120, line 25: This was originally comparing ProjectNames. https://github.com/NuGet/NuGet.Client/commit/3d091dad054331cc212651ad6d88d18533d1915f#diff-5a34d358c305cf4b9fb65e2fba45c849406b2898a65aa0a75513e41370c2161aR19

[ProjectNames.Equals](https://github.com/NuGet/NuGet.Client/blob/0bf00d2496ec6de66fe67b1c8f571090ac2eef87/src/NuGet.Clients/NuGet.VisualStudio.Common/ProjectSystems/ProjectNames.cs#L243-L256) does a case insensitive check over the full name, so this change restores the dictionary to be case insensitive.

nkolev92 (approved):

PR #6011, opened 2024-09-05
Target .NET 9.0

jeffkl:
# Bug

Fixes: 

## PR Checklist

- [ ] Meaningful title, helpful description and a linked NuGet/Home issue
- [ ] Added tests
- [ ] Link to an issue or pull request to update docs if this PR changes settings, environment variables, new feature, etc.

PR #6063, opened 2024-09-27
Test .NET 9.0.1xx in the Dotnet integration tests

nkolev92:
# Bug

Related: https://github.com/NuGet/Client.Engineering/issues/3025 

## Description

Similar to https://github.com/NuGet/NuGet.Client/pull/6011.

Updating the tested .NET SDK to Net 9. Note that in the past, we'd update both the test framework and the .NET SDK being tested. I don't think this is necessary, since this is a full test where we patch to the SDK which comes with its own runtime and all. 

Note that I consolidated the places where we'd need to make changes to get these tests passing.
## PR Checklist

- [x] ~Added tests~


nkolev92:
I think this is something we should consider patching release-6.12.x with, ensuring our tests are current

zivkan (approved):

PR #6077, opened 2024-10-02
Throw an exception with a message when unable to read corrupted .nupkg.metadata

### LINKED ISSUE #13763: Report the path when unable to read corrupted .nupkg.metadata
Pick a .nupkg.metadata file from your package cache and replace the contents with some NULL bytes, or corrupt it otherwise
Use RestoreEx that reads that file

Expected: nice error message `The file 'C:\Users\user\.nuget\packages\microsoft.win32.systemevents\8.0.0\.nupkg.metadata' is corrupted. It is recommended to clear the package cache using dotnet nuget locals all --clear and then retry the restore.`

Actual: 
```
d:\Current\bin\amd64\NuGet.RestoreEx.targets(19,5): error : Found invalid data while decoding. [D:\git\move\temp-msbuild-restore\dirs.proj]
d:\Current\bin\amd64\NuGet.RestoreEx.targets(19,5): error :    at NuGet.Packaging.NupkgMetadataFileFormat.Read(TextReader reader, ILogger log, String path) [D:\git\move\temp-msbuild-restore\dirs.proj]
d:\Current\bin\amd64\NuGet.RestoreEx.targets(19,5): error :    at NuGet.Packaging.NupkgMetadataFileFormat.Read(Stream stream, ILogger log, String path) [D:\git\move\temp-msbuild-restore\dirs.proj]
d:\Current\bin\amd64\NuGet.RestoreEx.targets(19,5): error :    at NuGet.Common.FileUtility.SafeRead[T](String filePath, Func`3 read) [D:\git\move\temp-msbuild-restore\dirs.proj]
```

See also:
https://github.com/NuGet/NuGet.Client/pull/4660
https://github.com/NuGet/Home/issues/10882

Nigusu-Allehu:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13763

## Description

Previously when nuget comes across a nupkg.metadata that is corrupt, it logs a warning and rethrows an exception. The exception that is rethrown does not have a message that specifies the path of the corrupt metadata. This PR does the following
* Remove the warning being logged. We do not need to log and then throw the exception. It will result in double logging
* Adds a specific message to the exception that clarifies which file is causing the problem

## PR Checklist

- [x] Added tests

### DIFF
--- a/src/NuGet.Core/NuGet.Packaging/NupkgMetadata/NupkgMetadataFileFormat.cs
+++ b/src/NuGet.Core/NuGet.Packaging/NupkgMetadata/NupkgMetadataFileFormat.cs
@@ -53,11 +53,11 @@ namespace NuGet.Packaging
             }
             catch (Exception ex)
             {
-                log.LogWarning(string.Format(CultureInfo.CurrentCulture,
+                var message = string.Format(CultureInfo.CurrentCulture,
                     Strings.Error_LoadingHashFile,
-                    path, ex.Message));
+                    path, ex.Message);
 
-                throw;
+                throw new Exception(message);
             }
         }
 
--- a/src/NuGet.Core/NuGet.Packaging/NupkgMetadata/NupkgMetadataFileFormat.cs
+++ b/src/NuGet.Core/NuGet.Packaging/NupkgMetadata/NupkgMetadataFileFormat.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using Newtonsoft.Json;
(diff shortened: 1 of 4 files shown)


src/NuGet.Core/NuGet.Packaging/NupkgMetadata/NupkgMetadataFileFormat.cs
```diff
@@ -53,11 +53,11 @@ public static NupkgMetadataFile Read(TextReader reader, ILogger log, string path
...
             catch (Exception ex)
             {
-                log.LogWarning(string.Format(CultureInfo.CurrentCulture,
+                var message = string.Format(CultureInfo.CurrentCulture,
                     Strings.Error_LoadingHashFile,
-                    path, ex.Message));
+                    path, ex.Message);
 
-                throw;
+                throw new Exception(message);
```
zivkan:
```suggestion
                throw new Exception(message, ex);
```

It's good practise to always retain the original exception as an inner exception.

zivkan: It's also good practice to throw specific exceptions, not use the base `Exception` type.  Perhaps an `ArgumentException` can be used. From a quick look at the [SystemException docs' list of derived types](https://learn.microsoft.com/en-us/dotnet/api/system.systemexception?view=net-8.0), I don't see anything else that looks better, but I only had a very quick look.

src/NuGet.Core/NuGet.Packaging/NupkgMetadata/NupkgMetadataFileFormat.cs
```diff
@@ -53,11 +53,11 @@ public static NupkgMetadataFile Read(TextReader reader, ILogger log, string path
             }
             catch (Exception ex)
             {
-                log.LogWarning(string.Format(CultureInfo.CurrentCulture,
+                var message = string.Format(CultureInfo.CurrentCulture,
```
nkolev92: You mentioned we don't want to double log, can you please show an example of how it looked like vs how it looks like now? 

The other thing I'm curious about is how it looks with your other change about failing restore gracefully. 
Should this change be retested with that one?

donnie-msft: 👀

Nigusu-Allehu: Good question. This was more of a conclusion I reached based on the code. We have `log.logWarning` statement followed by an `throw` statement. However, after manually testing the scenario, I did not see any double logs in the console.

PR #6081, opened 2024-10-03
Migrate CacheFileFormat to System.Text.Json

### LINKED ISSUE #13059: Use System.Text.Json to read  the cache file in CacheFileFormat

Nigusu-Allehu:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13059

## Description

This PR makes sure we use System.Text.Json to read the cache file in CacheFileFormat.

## PR Checklist

- [x] Added tests


src/NuGet.Core/NuGet.ProjectModel/CacheFileFormat.cs
```diff
@@ -35,8 +34,9 @@ private static CacheFile Read(TextReader reader, ILogger log, string path)
         {
             try
             {
-                var json = JsonUtility.LoadJson(reader);
-                var cacheFile = ReadCacheFile(json);
+                string jsonString = reader.ReadToEnd();
+                var json = JsonDocument.Parse(jsonString);
```
zivkan: Unfortunately, NuGet was reading/writing json files in the worst possible way, and doing a "direct" port to System.Text.Json provides only limited benefit.  Especially, since System.Text.Json doesn't support `TextReader`, and this commit is working around that by loading the entire file into a single string, and this could actually make the perf worse, rather than better.

The problem with `JsonDocument` (similar to `JObject` from Newtonsoft.Json), is that it causes the entire json document to be loaded in memory at once, in the parser's own data structures. Since the it's first loading the entire file as a string, that's twice the file is loaded in memory, just two different data structures. And then converting it to our `CacheFile` data structure is the 3rd time it's loaded in memory.

It might not be a "huge" problem with a small file like `CacheFile`, but it would be a very significant problem for larger files.

It's much more efficient to use `JsonSerializer` instead to "stream" the json parsing, plus it's much less code on our side too.


Basically, we should follow the guidance that System.Text.Json provide: https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/deserialization

We can use `JsonSerializer.Deserialize<CacheFile>(fileStream)`  (delete the method that uses a `TextReader`), and to write use `JsonSerializer.Serialize(cacheFile, options)`, without calling `GetCacheFile`. Instead, use `[JsonPropertyName` attributes to map json and class property names when they don't match.  Then most of the code in this file can be deleted, in addition to making it have better runtime perf.

src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileFormat.cs
```diff
@@ -1027,5 +1028,84 @@ private static List<CentralTransitiveDependencyGroup> ReadProjectFileTransitiveD
             }
             return results;
         }
+
+        internal static IList<IAssetsLogMessage> ReadLogMessageArray(JsonElement json, string projectPath)
```
zivkan: the changes in this file appear unrelated to `CacheFileFormat`, from the PR's title.

src/NuGet.Core/NuGet.ProjectModel/CacheFileFormat.cs
zivkan: I ported something else to System.Text.Json to demonstrate how it can be done:
* https://github.com/NuGet/NuGet.Client/pull/6083

As you can see, there's no need to write any code that know what the json property names are. Let the serializer handle that automatically. Also, no need for the temporary anonymous object on write.

Nigusu-Allehu: That was helpful! Thank you! I made sure my code followed suit as well. I used `sonSerializer.DeserializeAsync<CacheFile>(utf8Json: stream, SerializerOptions).GetAwaiter().GetResult();`, but I noticed you were able to use `Deserialize` overload by referencing the s.t.j package

src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileFormat.cs
Nigusu-Allehu: This was used by CacheFileFormat to read logmessages. I have now moved it into the custom converter class I have added to deserialize LogMessages.

src/NuGet.Core/NuGet.ProjectModel/CacheFileFormat.cs
Nigusu-Allehu: Wondering if it could lead into  a dead lock

src/NuGet.Core/NuGet.ProjectModel/CacheFileFormat.cs
```diff
@@ -5,49 +5,172 @@
...
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+            Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping,
+            DefaultIgnoreCondition = System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingNull,
+            Converters = { new AssetsLogMessageConverter() }
+        };
+
+        /// <summary>
+        /// since Log messages property in CacheFile is an interface type, we have the following custom converter to deserialize the IAssetsLogMessage objects.
+        /// </summary>
+        private class AssetsLogMessageConverter : JsonConverter<IAssetsLogMessage>
```
nkolev92, line 30: Part of the idea of calling the LockFileFormat implementation was to avoid maintaining 2 different implementations of the assets file log message reading. 

Is there anything in that implementation we can reuse given that there's an STJ implementation?

Nigusu-Allehu: You are right. LockFileFormat has implementations for reading and writing IAssetsLogMessage. However, as you mentioned, it is [currently implemented](https://github.com/NuGet/NuGet.Client/blob/764d370df4d3c01c24340f57e6389058bce24ddf/src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileFormat.cs#L496) using NSJ. I don't think we could reuse it currently for a STJ implementation. I think, in the future, when we migrate LockFileFormat, this converter could be used by every method trying to read and write asset file log messages.

Maybe if we modify ReadLogMessage in LockFileFormat to take in a delegate that allows it to read its input, we could be able to use it to read log messages from both` NSJ JObject` and `STJ JsonElement`. The new ReadLogMessage, could have a definition that looks like 
* delegate `public delegate T JsonPropertyReader<T>(object json, string property)`
* `private static IAssetsLogMessage ReadLogMessage(
    object json, 
    string projectPath, 
    JsonPropertyReader<string> readStringProperty, 
    JsonPropertyReader<int> readIntProperty)`

My fear is we might introduce some allocation overhead from the delegates.

nkolev92: The lock file reading is implementing with STJ too: https://github.com/NuGet/NuGet.Client/commit/ef5bce328208c10d00beaf00a96a189af30cfd0b. 

It's what's used by default. 
There seems to be a converter implemented there: https://github.com/NuGet/NuGet.Client/blob/f7ae70859a8aa9c3f3fda6a3170bbfd437e47272/src/NuGet.Core/NuGet.ProjectModel/Utf8JsonReaderExtensions.cs#L19. 

I was thinking we can use that: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.ProjectModel/LockFile/Utf8JsonStreamIAssetsLogMessageConverter.cs.

zivkan: unfortunately, not really. STJ's `Utf8JsonReader` doesn't have a `Read()` or `MoveNext()` method that can read from a stream. Instead, it returns false when the end of the buffer is reached and the caller is then responsible for reading from the stream and filling the buffer, before trying again.

While STJ can internally do a reading stream with by-reflection (or source generator) serializer, if you implement their `JsonConverter`, it will buffer the entire object in memory. Since we'd need a converter for the root object, this means that STJ would need `JsonSerializer.Deserialize<AssetsFile>(stream)` to buffer the entire assets file in memory, and since that can be multiple megabytes, we don't do that.

STJ's docs recommend you implement your own streaming JsonReader, which is what we did for the assets file, and hence `IUtf8JsonStreamReaderConverter<T>` is a custom NuGet thing: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.ProjectModel/Utf8JsonStreamReaderConverter.cs#L9

This means it can't (easily?) be reused with STJ's `JsonConverter<T>` or `JsonSerializer.Deserialize<T>()`.

src/NuGet.Core/NuGet.ProjectModel/NuGet.ProjectModel.csproj
```diff
@@ -45,4 +45,8 @@
   <ItemGroup>
     <InternalsVisibleTo Include="NuGet.ProjectModel.Test" />
   </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Text.Json" />
```
zivkan: See my pull request, after it was rebased, following retargeting the projects to net8.0:
* https://github.com/NuGet/NuGet.Client/pull/6083

We shouldn't include a package reference for target frameworks where this assembly is already part of the targeting pack.

src/NuGet.Core/NuGet.ProjectModel/CacheFileFormat.cs
nkolev92: The largest part of the implementation for CacheFileFormat is the log messages. 
I know it's not the recommended way, but I'm wondering if we're better off just following that pattern with CacheFileFormat. 

I'm just trying to further minimize code duplication with serialization. We'll need 2 to implement STJ, I'd hate to have 3.

src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileFormat.cs
```diff
@@ -245,7 +244,7 @@ private static LockFile ReadLockFile(JObject cursor, string path, LockFileReadFl
                 : new List<CentralTransitiveDependencyGroup>();
 
             var logMessage = (flags & LockFileReadFlags.LogMessages) == LockFileReadFlags.LogMessages
-                ? ReadLogMessageArray(cursor[LogsProperty] as JArray, packagesSpec?.RestoreMetadata?.ProjectPath)
+                ? ReadLogMessageArray(cursor[CacheFileProperties.LogsProperty] as JArray, packagesSpec?.RestoreMetadata?.ProjectPath)
```
nkolev92: I'm not sure I see why it should belong in the CacheFileProperties instead of the LockFileFormat. It's already internal here,. No need to make changes in this file.

src/NuGet.Core/NuGet.ProjectModel/NuGet.ProjectModel.csproj
nkolev92: Please add a condition for the inclusion of the PR.

nkolev92 (commented):
There's a few comments, but these are really conversations, stylistic ideas. 

Once we're closed on this, I think this is ready to :shipit:

src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileFormat.cs
Nigusu-Allehu: I did this to avoid having the properties being defined in multiple places. It will make updates to the JSON much easier in the future. All the properties are found in `CacheFileProperties` now

nkolev92: From the perspective of CacheFile sure, but from the perspective of the LockFile, it's doing the exact opposite, now the lock file is referencing the cache file. 

I think both of our approaches are a bit opinionated, but there's a precedent here and we're replacing it with an equivalent one.

src/NuGet.Core/NuGet.ProjectModel/CacheFileFormat.cs
nkolev92: If we're not incorporating my feedback to use the existing converter, let's at least add detailed tests about the converter. 

Replaying the restore messages correctly is really important tot he experience given how often a restore no-ops.

zivkan: Sorry for not giving this feedback earlier. For some reason, I only just thought of it now 😕 

The implementation of the `Read` method can be changed to:

```cs
var converter = (JsonConverter<AssetsLogMessage>)options.GetConverter(typeof(AssetsLogMessage>));
return converter.Read(ref reader, typeof(AssetsLogMessage), options);
```

It might require annotating AssetsLogMessage to be compatible with System.Text.Json, similar to how CacheFile was annotated.  But then we can make the `Read` method only 2 or 3 lines long, not 70, and similarly the Write method can remove nearly 50 lines.


Also, I'm kind of wondering if we should break the public API and change `CacheFile`'s property to `IList<AssetsLogMessage>`, or `IReadOnlyList<AssetsLogMessage>` (or `ImmutableArray<AssetsLogMessage>`), instead of `IList<IAssetsLogMessage>`. 

Considering that `AssetsLogFile` is a DTO, there's no business logic in it, I honestly don't understand the point of `IAssetsLogMessage`, and I'm a fairly strong believer in the SOLID design principals and the port-adapter pattern.  Anyway, CacheFile could use the concrete class directly, and then we wouldn't need a 10 line converter. 
What do you think @nkolev92?

src/NuGet.Core/NuGet.ProjectModel/CacheFileProperties.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.ProjectModel
+{
+    internal class CacheFileProperties
+    {
+        internal const string VersionProperty = "version";
```
zivkan: The only place these properties are used are in `CacheFile`, in `JsonPropertyName` attributes. I think we're better off just putting the strings directly in the attribute (for example `[JsonPropertyName("version")]`), rather than having this class.

src/NuGet.Core/NuGet.ProjectModel/CacheFileFormat.cs
nkolev92: I'm guessing you're asking me with thoughts on breaking the public API for CacheFile?
I added CacheFile, and if I had known what I know now, I would've kept that in NuGet.Commands instead of NuGet.ProjectModel, and nothing would've been public. 
I don't expect anyone to be using this, so as long as we document the change, i'm ok with breaking the API if you 2 agree we want that.

Nigusu-Allehu: I like the idea of `IList<AssetsLogMessage>`. I will create another PR updating our API and then  I will update this PR

Nigusu-Allehu: https://github.com/NuGet/NuGet.Client/pull/6132

nkolev92: @Nigusu-Allehu I'll rely on @zivkan here, but I don't think you need that change. 

You can just assume the concrete class in that converter and you should be good.

src/NuGet.Core/NuGet.ProjectModel/CacheFileFormat.cs
```diff
@@ -2,52 +2,61 @@
...
+        {
+            public override IAssetsLogMessage Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+            {
+                return JsonSerializer.Deserialize<AssetsLogMessage>(ref reader, options);
```
zivkan, line 34: Regarding https://github.com/NuGet/NuGet.Client/pull/6081/files#r1825078556, we lose about 30% perf needing to go though an additional converter:

<details>
<summary>code</summary>

```csharp
using System.Text;
using System.Text.Json;
using System.Text.Json.Serialization;

[MemoryDiagnoser]
public class Program
{
    {
    }

    private static JsonSerializerOptions options1;
    private static JsonSerializerOptions options2;

    [GlobalSetup]
    public void Setup()
    {
        options1 = new JsonSerializerOptions();
        options1.Converters.Add(new ThingConverter1());

        options2 = new JsonSerializerOptions();
        options2.Converters.Add(new ThingConverter2());
    }

    private byte[] json = Encoding.UTF8.GetBytes("{\"Value\":\"something\"}");

    [Benchmark]
    public IThing? JsonSerializerDeserialize()
    {
        return JsonSerializer.Deserialize<IThing>(json, options1);
    }

    [Benchmark]
    public IThing? JsonConverterRead()
    {
        return JsonSerializer.Deserialize<IThing>(json, options2);
    }

    [Benchmark(Baseline = true)]
    public Thing? Baseline()
    {
        return JsonSerializer.Deserialize<Thing>(json);
    }

    public interface IThing
    {
        public string Value { get; }
    }

    public class Thing : IThing
    {
        public required string Value { get; init; }
    }

    private class ThingConverter1 : JsonConverter<IThing>
    {
        public override IThing? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
        {
            return JsonSerializer.Deserialize<Thing>(ref reader, options);
        }

        public override void Write(Utf8JsonWriter writer, IThing value, JsonSerializerOptions options)
        {
            throw new NotImplementedException();
        }
    }

    private class ThingConverter2 : JsonConverter<IThing>
    {
        public override IThing? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
        {
            JsonConverter<Thing> converter = (JsonConverter<Thing>)options.GetConverter(typeof(Thing));
            return converter.Read(ref reader, typeof(Thing), options);
        }

        public override void Write(Utf8JsonWriter writer, IThing value, JsonSerializerOptions options)
        {
            throw new NotImplementedException();
        }
    }
}
```

</details>

| Method                    | Mean      | Error    | StdDev   | Ratio | RatioSD | Gen0   | Allocated | Alloc Ratio |
|-------------------------- |----------:|---------:|---------:|------:|--------:|-------:|----------:|------------:|
| JsonSerializerDeserialize | 189.38 ns | 1.185 ns | 1.050 ns |  2.00 |    0.02 | 0.0076 |     128 B |        1.00 |
| JsonConverterRead         | 123.56 ns | 0.691 ns | 0.647 ns |  1.31 |    0.01 | 0.0076 |     128 B |        1.00 |
| Baseline                  |  94.51 ns | 0.602 ns | 0.563 ns |  1.00 |    0.01 | 0.0076 |     128 B |        1.00 |

We lose even more perf calling `JsonSerializer.Deserialize` from the converter, rather than my suggestion above, getting the converter directly from options, and calling the converter.

It depends on how much we want to microoptimize restore.  30 nanoseconds per project, on my machine, but nanoseconds are super small.  Is it worth the public API breaking changes for it? Probably not.  But I still think that `IAssetsLogMessage` doesn't have a reason to exist.

But I think the additional 60 nanoseconds lost by using `JsonSerializer.Deserialize` is not worth the 1 line of code it saves over getting the converter directly.

src/NuGet.Core/NuGet.ProjectModel/LockFile/AssetsLogMessage.cs
```diff
@@ -3,27 +3,43 @@
...
+        [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
         public int StartColumnNumber { get; set; }
+
+        [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
         public int EndLineNumber { get; set; }
+
+        [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
         public int EndColumnNumber { get; set; }
 
+        public AssetsLogMessage() { }
```
zivkan: I'm pretty sure that System.Text.Json can still call the constructor if it's internal. If this class had `#nullable enable`, then you'd see the problem with adding this constructor: it doesn't ensure that types the rest of our code expects to always be non-null, is in fact non-null. Therefore, we'd have to make `Message` a `string?`, and now everything that processes AssetsLogMessages has to handle when `Message` is null.

Technically, there's still the problem even if this constructor is internal or private, but at least we can suppress the nullable warning explaining why it happens.  Still, I'd prefer if it could be avoided. Can we somehow use `[JsonConstructor]` on the `LogLevel, NuGetLogCode, string, string` overload?  I don't have experience with trying to tell the JsonSerializer to set some values via the constructor, and different values via properties.

For much the same reasons, a good API is one you can't use wrong, so we don't want the rest of NuGet.Client being able to construct instances where message is null. If we need this constructor for JSON deserialization, then we should limit it's potential misuse.

src/NuGet.Core/NuGet.ProjectModel/LockFile/AssetsLogMessage.cs
```diff
@@ -54,6 +70,11 @@ public AssetsLogMessage(LogLevel logLevel, NuGetLogCode errorCode,
                     targetGraph
                 };
             }
+
+            if (logLevel == LogLevel.Warning)
+            {
+                WarningLevel = WarningLevel.Severe; // setting default to Severe as 0 implies show no warnings
```
zivkan: this `logLevel == LogLevel.Warnings` condition in order to set `WarningLevel`'s default looks like a behaviour change from the old code, and it's not intuitive to me why the condition is needed.

src/NuGet.Core/NuGet.ProjectModel/LockFile/AssetsLogMessage.cs
```diff
@@ -106,5 +127,25 @@ public override int GetHashCode()
 
             return combiner.CombinedHash;
         }
+
+        private class ToStringConverter<T> : JsonConverter<T>
```
zivkan: I believe you can use what's already built in: https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/customize-properties?pivots=dotnet-8-0#enums-as-strings

src/NuGet.Core/NuGet.ProjectModel/LockFile/AssetsLogMessage.cs
Nigusu-Allehu: https://github.com/NuGet/NuGet.Client/blob/3ee5288d335f6fa62bbb63fb168e8bea2d599dc2/src/NuGet.Core/NuGet.ProjectModel/LockFile/IAssetsLogMessage.cs#L33
I have this so that the Json serializer does not write the value of WarningLevel, when the log level is not warning. The warning level should only be serialized if the log level is warning. In addition, the warning level is useless if the log level is not warning as per the description in the interface

Nigusu-Allehu: If want to keep it the way it used to be, another option to consider is introducing another property in AssetsLogMessage that is private and 
```
 [JsonIgnore]
 public WarningLevel WarningLevel { get; set; } = WarningLevel.Severe; //setting default to Severe as 0 implies show no warnings

 [JsonInclude]
 [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
 [JsonPropertyName("warningLevel")]
 internal WarningLevel? WarningLevelJson
 {
     get => Level == LogLevel.Warning ? WarningLevel : null;
     set
     {
         if (value.HasValue)
         {
             WarningLevel = value.Value;
         }
         
```

Nigusu-Allehu:
@NuGet/nuget-client, I have removed the perf improvements. Please feel free to review

test/NuGet.Core.Tests/NuGet.ProjectModel.Test/CacheFileFormatTests.cs
```diff
@@ -19,9 +20,9 @@ public void Read_WhenVersionIs1_ReadsCorrectly()
             var logger = new TestLogger();
 
             var contents = $@"{{
-  ""version"": ""1"",
```
jeffkl: This seems wrong, shouldn't the value stay as a string?

Nigusu-Allehu: This is a test on read. We had a test that was previously reading an invalid cacheFile and passing. CacheFile.Version is an integer https://github.com/NuGet/NuGet.Client/blob/13aaaf31db2958ca85b0b98b8dea8fcf6a7961de/src/NuGet.Core/NuGet.ProjectModel/CacheFile.cs#L17 


Even in the NSJ implementation we were trying to read an int, but it never threw an exception. I am not sure if it has an intrisic type converter

Nigusu-Allehu: https://github.com/NuGet/NuGet.Client/blob/13aaaf31db2958ca85b0b98b8dea8fcf6a7961de/src/NuGet.Core/NuGet.ProjectModel/CacheFileFormat.cs#L89

src/NuGet.Core/NuGet.ProjectModel/CacheFileFormat.cs
```diff
@@ -2,52 +2,61 @@
...
+        private static JsonSerializerOptions SerializerOptions = new JsonSerializerOptions
         {
-            if (stream == null) throw new ArgumentNullException(nameof(stream));
-            if (log == null) throw new ArgumentNullException(nameof(log));
-            if (path == null) throw new ArgumentNullException(nameof(path));
+            WriteIndented = true,
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+            Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping,
+            DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
+            Converters = { new AssetsLogMessageConverter() }
```
kasperk81:
```suggestion
            NumberHandling = JsonNumberHandling.AllowReadingFromString,
            Converters = { new AssetsLogMessageConverter() }
```

this will match netwonsoft behavior, then both `"version": "1"` and `"version": 1` will desrialize to int property as before

test/NuGet.Core.Tests/NuGet.ProjectModel.Test/CacheFileFormatTests.cs
kasperk81: it requires https://github.com/NuGet/NuGet.Client/pull/6081/files#r1987744473 before reverting this line

newtonsoft has this default behavior https://dotnetfiddle.net/Jd1Wsc
stj has it opt-in with numberhandling https://dotnetfiddle.net/4axJhW

Nigusu-Allehu: I don't think it is a behavior we should support. However, since I don't want this to be a breaking change for a user who somehow edited their cachefile to have a string version value, I will add the AllowReadingFromString

kasperk81: same with https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializeroptions.allowtrailingcommas, if someone happened to have trailing comma, their workflow will break.

> I don't think it is a behavior we should support. However, since I don't want this to be a breaking change for a user who somehow edited their cachefile to have a string version value, I will add the AllowReadingFromString

you are right, if it was the day one of this config file, we could make it strict mode, but since newtonsoft has been providing the lax mode for ages and nuget has been allowing it, it's best to make this change seamless for the user.

jeffkl (approved):

kasperk81:
@Nigusu-Allehu are you planning to convert the rest https://github.com/search?q=repo%3ANuGet%2FNuGet.Client%20newtonsoft.json&type=code?

PR #6083, opened 2024-10-04
Migrate .nupkg.metadata read/write to System.Text.Json

### LINKED ISSUE #13835: Read and write .nupkg.metadata files with System.Text.Json
System.Text.Json has lower allocations and higher performance.

Anyone using the NuGet Client SDK's `NuGetMetadataFileFormat` `.Read` and `.Write` overloads with the `TextReader` or `TextWriter` parameters will have a breaking change. System.Text.Json doesn't read or write these text streams, only UTF8 byte steams.  But there's little reason for anyone to use the stream (byte or text) overloads, it's easier for most people to pass the path, and let NuGet open the file.

zivkan:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13835

## Description

Port `NupkgMetadataFileFormat` to System.Text.Json

<details>
<summary>benchmark code</summary>
<div>

```cs
using NuGet.Common;
using NuGet.Packaging;

namespace Benchmark
{
    [MemoryDiagnoser]
    public class Program
    {
        static void Main(string[] args)
        {
        }

#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider adding the 'required' modifier or declaring as nullable.
        // These properties are set by the global setup
        private string _readPath;
        private string _writePath;
        private NupkgMetadataFile _instance;
#pragma warning restore CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider adding the 'required' modifier or declaring as nullable.

        [GlobalSetup]
        public void GlobalSetup()
        {
            var gpf = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".nuget", "packages");
            foreach (var file in Directory.EnumerateFiles(gpf, ".nupkg.metadata", SearchOption.AllDirectories))
            {
                _readPath = file;
                break;
            }

            _writePath = Path.GetTempFileName();

            _instance = new NupkgMetadataFile()
            {
                Version = 2,
                ContentHash = "1234567890ABCDEFGHIJLMNOPQRSTUVWXYZ",
                Source = "https://contoso.test/v3/index.json",
            };
        }

        [Benchmark]
        [Obsolete]
        public NupkgMetadataFile NJRead()
        {
            using (var reader = new StreamReader(_readPath))
            {
                NupkgMetadataFile metadata = NupkgMetadataFileFormat.Read(reader, NullLogger.Instance, _readPath);
                return metadata;
            }
        }

        [Benchmark]
        public NupkgMetadataFile STJRead()
        {
            NupkgMetadataFile metadata = NupkgMetadataFileFormat.Read(_readPath);
            return metadata;
        }

        [Benchmark]
        [Obsolete]
        public void NJWrite()
        {
            using (var writer = new StreamWriter(_writePath))
            {
                NupkgMetadataFileFormat.Write(writer, _instance);
            }
        }

        [Benchmark]
        public void STJWrite()
        {
            NupkgMetadataFileFormat.Write(_writePath, _instance);
        }
    }
}

```

</div>
</details>

| Method   | Mean     | Error   | StdDev  | Gen0   | Allocated |
|--------- |---------:|--------:|--------:|-------:|----------:|
| NJRead   | 106.4 us | 1.70 us | 1.59 us | 0.6104 |   10680 B |
| STJRead  | 105.2 us | 1.32 us | 1.17 us |      - |     664 B |
| NJWrite  | 201.4 us | 2.30 us | 1.92 us | 0.2441 |    7416 B |
| STJWrite | 180.2 us | 2.14 us | 1.89 us | 0.2441 |    4632 B |

## PR Checklist

- [x] ~Added tests~ existing tests were updated
- [x] ~Link to an issue or pull request to update docs if this PR changes settings, environment variables, new feature, etc.~ no functional changes to product

### DIFF
--- a/src/NuGet.Core/NuGet.Packaging/NuGet.Packaging.csproj
+++ b/src/NuGet.Core/NuGet.Packaging/NuGet.Packaging.csproj
@@ -53,7 +53,7 @@
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" />
     <PackageReference Include="System.Memory" Condition="'$(TargetFramework)' == '$(NETFXTargetFramework)'" />
-    <PackageReference Include="System.Text.Json" />
+    <PackageReference Include="System.Text.Json" Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' != '$(NETFXTargetFramework)'">
--- a/src/NuGet.Core/NuGet.Packaging/NuGet.Packaging.csproj
+++ b/src/NuGet.Core/NuGet.Packaging/NuGet.Packaging.csproj
@@ -53,6 +53,7 @@
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" />
     <PackageReference Include="System.Memory" Condition="'$(TargetFramework)' == '$(NETFXTargetFramework)'" />
+    <PackageReference Include="System.Text.Json" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' != '$(NETFXTargetFramework)'">
--- a/src/NuGet.Core/NuGet.Packaging/NupkgMetadata/NupkgMetadataFileFormat.cs
+++ b/src/NuGet.Core/NuGet.Packaging/NupkgMetadata/NupkgMetadataFileFormat.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Globalization;
 using System.IO;
+using System.Text.Json;
 using Newtonsoft.Json;
 using NuGet.Common;
 
@@ -17,7 +18,16 @@ namespace NuGet.Packaging
         private const string HashProperty = "contentHash";
         private const string SourceProperty = "source";
 
-        private static readonly JsonSerializer JsonSerializer = JsonSerializer.Create(GetSerializerSettings());
+        private static readonly Newtonsoft.Json.JsonSerializer JsonSerializer = Newtonsoft.Json.JsonSerializer.Create(GetSerializerSettings());
+        private static readonly JsonSerializerOptions JsonSerializerOptions = new JsonSerializerOptions()
+        {
+            WriteIndented = true,
+            PropertyNameCaseInsensitive = true,
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+            // This file is not intended to be parsed in javascript, and the contentHash property contains a Base64 encoded string that
+            // can contain characters such as + that has not been encoded in the past.
+            Encoder = System.Text.Encodings.Web.JavaScriptEncoder.UnsafeRelaxedJsonEscaping,
+        };
 
         public static NupkgMetadataFile Read(string filePath)
         {
@@ -31,12 +41,35 @@ namespace NuGet.Packaging
 
         public static NupkgMetadataFile Read(Stream stream, ILogger log, string path)
         {
-            using (var textReader = new StreamReader(stream))
+            if (stream is null) { throw new ArgumentNullException(nameof(stream)); }
+
+            try
             {
-                return Read(textReader, log, path);
+                try
+                {
+                    NupkgMetadataFile nupkgMetadata = System.Text.Json.JsonSerializer.Deserialize<NupkgMetadataFile>(stream, JsonSerializerOptions);
+                    if (nupkgMetadata == null)
+                    {
+                        throw new InvalidDataException();
+                    }
+                    return nupkgMetadata;
+                }
+                catch (System.Text.Json.JsonException jsonException)
+                {
+                    throw new InvalidDataException(jsonException.Message, jsonException);
+                }
+            }
+            catch (Exception ex)
+            {
+                log.LogWarning(string.Format(CultureInfo.CurrentCulture,
+                    Strings.Error_LoadingHashFile,
+                    path, ex.Message));
+
+                throw;
             }
         }
 
+        [Obsolete("Use a different overload for better performance. This overload will get deleted in the future.")]
         public static NupkgMetadataFile Read(TextReader reader, ILogger log, string path)
         {
             try
@@ -75,12 +108,12 @@ namespace NuGet.Packaging
 
         public static void Write(Stream stream, NupkgMetadataFile hashFile)
         {
-            using (var textWriter = new StreamWriter(stream))
-            {
-                Write(textWriter, hashFile);
-            }
+            if (stream is null) { throw new ArgumentNullException(nameof(stream)); }
+
+            System.Text.Json.JsonSerializer.Serialize(stream, hashFile, JsonSerializerOptions);
         }
 
+        [Obsolete("Use a different overload for better performance. This overload will get deleted in the future.")]
         public static void Write(TextWriter textWriter, NupkgMetadataFile hashFile)
         {
             using (var jsonWriter = new JsonTextWriter(textWriter))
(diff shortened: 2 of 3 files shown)


src/NuGet.Core/NuGet.Packaging/NuGet.Packaging.csproj
```diff
@@ -53,6 +53,7 @@
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" />
     <PackageReference Include="System.Memory" Condition="'$(TargetFramework)' == '$(NETFXTargetFramework)'" />
+    <PackageReference Include="System.Text.Json" />
```
zivkan: I first tried with a condition that excluded .NETCoreApp TFMs, but this project builds with net5.0, and STJ in .NET 5 doesn't have a `Deserialize<T>(Stream)` overload.

src/NuGet.Core/NuGet.Packaging/NupkgMetadata/NupkgMetadataFileFormat.cs
```diff
@@ -31,12 +41,35 @@ public static NupkgMetadataFile Read(string filePath, ILogger log)
...
+                log.LogWarning(string.Format(CultureInfo.CurrentCulture,
+                    Strings.Error_LoadingHashFile,
+                    path, ex.Message));
+
+                throw;
             }
         }
 
+        [Obsolete("Use a different overload for better performance. This overload will get deleted in the future.")]
         public static NupkgMetadataFile Read(TextReader reader, ILogger log, string path)
```
zivkan, line 77: We could be more aggressive and just delete this public API, as well as the `Write(TextWriter` overload. That would allow us to delete all the Newtonsoft.Json code from this class.

test/NuGet.Core.Tests/NuGet.Packaging.Test/NupkgMetadataFileFormatTests.cs
```diff
@@ -63,58 +54,39 @@ public void Read_V2FileContents_ReturnsCorrectObject()
...
-            {
-                Assert.Throws<InvalidDataException>(() => NupkgMetadataFileFormat.Read(stringReader, logger, "from memory"));
-            }
-
-            // Assert
-            Assert.Equal(1, logger.Messages.Count);
-        }
-
-        [Fact]
-        public void Read_ContentsInvalidJson_ThrowsJsonReaderException()
```
zivkan, line 108: This test exposes a change in behaviour. Previously, it would throw a Newtonsoft.Json.JsonReaderException. But this is arguably a leaky abstraction, and it's not reasonable to maintain given we're migrating off Newtonsoft.Json.  I thought `InvalidDataException` is better, so I merged this test in with the other invalid data test.

Theghost256 (approved):

src/NuGet.Core/NuGet.Packaging/NupkgMetadata/NupkgMetadataFileFormat.cs
```diff
@@ -17,7 +18,16 @@ public static class NupkgMetadataFileFormat
...
+        private static readonly Newtonsoft.Json.JsonSerializer JsonSerializer = Newtonsoft.Json.JsonSerializer.Create(GetSerializerSettings());
+        private static readonly JsonSerializerOptions JsonSerializerOptions = new JsonSerializerOptions()
+        {
+            WriteIndented = true,
+            PropertyNameCaseInsensitive = true,
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+            // This file is not intended to be parsed in javascript, and the contentHash property contains a Base64 encoded string that
+            // can contain characters such as + that has not been encoded in the past.
+            Encoder = System.Text.Encodings.Web.JavaScriptEncoder.UnsafeRelaxedJsonEscaping,
+        };
```
Nigusu-Allehu:
```suggestion
        private static readonly JsonSerializerOptions JsonSerializerOptions = new JsonSerializerOptions()
        {
            WriteIndented = true,
            PropertyNameCaseInsensitive = true,
            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
            // This file is not intended to be parsed in javascript, and the contentHash property contains a Base64 encoded string that
            // can contain characters such as + that has not been encoded in the past.
            Encoder = System.Text.Encodings.Web.JavaScriptEncoder.UnsafeRelaxedJsonEscaping,
            DefaultIgnoreCondition = System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingNull,
        };
```
I think we should skip null values when writing

src/NuGet.Core/NuGet.Packaging/NuGet.Packaging.csproj
nkolev92: Messaged offline.

src/NuGet.Core/NuGet.Packaging/NupkgMetadata/NupkgMetadataFileFormat.cs
```diff
@@ -31,12 +41,35 @@ public static NupkgMetadataFile Read(string filePath, ILogger log)
...
+                    throw new InvalidDataException(jsonException.Message, jsonException);
+                }
+            }
+            catch (Exception ex)
+            {
+                log.LogWarning(string.Format(CultureInfo.CurrentCulture,
+                    Strings.Error_LoadingHashFile,
+                    path, ex.Message));
+
+                throw;
```
martinrrm: nit: Can we change this code to use only one try and have 2 catch conditions?

```c#
try
{
  // do something
}
catch (JsonException jsonException)
{
}
catch (Exception ex)
{
}
```

src/NuGet.Core/NuGet.Packaging/NuGet.Packaging.csproj
zivkan: I'll be able to re-add the TargetFrameworkIdentifier != .NETCoreApp condition once this gets merged:
* https://github.com/NuGet/NuGet.Client/pull/6086

martinrrm (approved):

test/NuGet.Core.Tests/NuGet.Packaging.Test/NupkgMetadataFileFormatTests.cs
```diff
@@ -2,8 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.IO;
-using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
+using System.Text;
```
nkolev92, line 5: I'm wondering if we need to add a test that ensures the reading/writing among the 2 different implementations hasn't regressed?

nkolev92 (approved):

test/NuGet.Core.Tests/NuGet.Packaging.Test/NupkgMetadataFileFormatTests.cs
zivkan: There's already one test that validates round tripping from write back to read, and another test that validates against a hard-coded json string. I think that's sufficient.

Also, our product code doesn't actually use the Newtonsoft.Json overloads. There's basically no reason for anyone else to read/write the `.nupkg.metadata` file (so I'm a little disappointed nobody encouraged me to [delete the Newtonsoft.Json overloads](https://github.com/NuGet/NuGet.Client/pull/6083#discussion_r1788386243)).

So, IMO, there isn't really a benefit, the NJ code is effectively dead code, if it wasn't a public API that someone could theoretically use.

PR #6087, opened 2024-10-08
Skip PushCommand_AllowsTimeoutToBeSpecifiedHigherThan100Seconds and PushCommand_AllowsTimeoutToBeSpecifiedLowerThan100Seconds

jeffkl:
# Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/3045

## Description
These two tests check if a user can override the default timeout, which is now 5 minutes (300 seconds).  We could update the tests but the one that verifies you can specify a longer timeout runs for over 100 seconds which seems wasteful.  The one that checks that you can specify a shorter timeout appears to hang occasionally.  I don't think either test is providing any value at this point.

## PR Checklist

- [ ] Added tests


nkolev92 (dismissed):

donnie-msft (commented):
The team policy is to use public issues for skipping tests. Client.Engineering isn't accessible to anyone outside our team reading the codebase.

zivkan (approved):

PR #6091, opened 2024-10-09
Fix invalid target framework error for Solution restore

### LINKED ISSUE #13772: "Sequence contains no matching element" when using a bad TF
### NuGet Product Used

MSBuild.exe

### Product Version

MSBuild version 17.9.8+b34f75857 for .NET Framework

### Worked before?

No

### Impact

### Repro Steps & Context

Error:
```
...\Current\bin\NuGet.targets(169,5): error : Sequence contains no matching element [<repo>\dirs.proj]
```

This happened when accidentally using:

```xml
    <TargetFramework>net8.0;net472</TargetFramework>
```

This is clearly a user error, but the error message could be more helpful. At the very least it should point to the problematic project instead of the root project for the restore.

I can only get this specific error to repro under certain conditions which I'm not able to fully understand. eg it manifests this way when restoring from the root. When restoring the project directly I helpfully get:

>NETSDK1046: The TargetFramework value 'net8.0;net472' is not valid. To multi-target, use the 'TargetFrameworks' property instead.

I would guess it maybe only manifests when a projects depends on the problematic project? Not super sure though.

Attaching a debugger shows that the exception is thrown here:
image

And the stack in case it's helpful:
image

I can provide more specific details (specific internal repo and commit) offline if desired.

### Verbose Logs

_No response_

martinrrm:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13772

## Description
There is a difference in behavior when doing a restore in Solution vs Project with an invalid target framework in the `<TargetFramework>` tag.  When doing it in the project Level, sdk will run a validation before doing restore in https://github.com/dotnet/sdk/blob/5ab26633bdf4676efa25140a31a3d6182efb02f9/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.TargetFrameworkInference.targets#L92.

The problem is that Solution level does not this validation (since it doesn't run a restore for each project) and will fail with 
```
...\Current\bin\NuGet.targets(169,5): error : Sequence contains no matching element [<repo>\dirs.proj]
```

This PR adds a validation and return the same error message as the project level. Currently, I decided to error with a NuGet error but we could also add `NETSDK1046` in the error message.

## PR Checklist

- [x] Added tests

### DIFF
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
@@ -458,6 +458,12 @@ namespace NuGet.Commands
             foreach (var item in GetItemByType(items, "TargetFrameworkInformation"))
             {
                 var frameworkString = item.GetProperty("TargetFramework");
+                if (frameworkString.Contains(';'))
+                {
+                    var error = RestoreLogMessage.CreateError(NuGetLogCode.NU1001, string.Format(CultureInfo.CurrentCulture, Strings.Error_TargetFrameworkWithSemicolon, frameworkString));
+                    throw new RestoreCommandException(error);
+                }
+
                 var targetFrameworkMoniker = item.GetProperty("TargetFrameworkMoniker");
                 var targetPlatforMoniker = item.GetProperty("TargetPlatformMoniker");
                 var targetPlatformMinVersion = item.GetProperty("TargetPlatformMinVersion");
--- a/test/NuGet.Core.Tests/NuGet.Build.Tasks.Console.Test/MSBuildStaticGraphRestoreTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Build.Tasks.Console.Test/MSBuildStaticGraphRestoreTests.cs
@@ -101,6 +101,19 @@ namespace NuGet.Build.Tasks.Console.Test
             actual.Should().BeEquivalentTo(expected);
         }
 
+        [Theory]
+        [InlineData("net40;net45")]
+        [InlineData("net40;net45 ; netstandard2.0 ")]
+        public void GetOriginalTargetFramework_WhenTargetFramworkSpecified_HasCorrectTargetFramework(string targetFrameworks)
+        {
+            var project = new MockMSBuildProject(new Dictionary<string, string>
+            {
+                ["TargetFramework"] = targetFrameworks
+            });
+
+            Assert.Throws<RestoreCommandException>(() => MSBuildStaticGraphRestore.GetTargetFrameworkStrings(project));
+        }
+
         [Fact]
         public void GetPackageDownloads_WhenDuplicatesExist_DuplicatesIgnored()
         {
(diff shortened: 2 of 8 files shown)


src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
```diff
@@ -454,6 +454,12 @@ private static IEnumerable<TargetFrameworkInformation> GetTargetFrameworkInforma
             foreach (var item in GetItemByType(items, "TargetFrameworkInformation"))
             {
                 var frameworkString = item.GetProperty("TargetFramework");
+                if (frameworkString.Contains(';'))
+                {
+                    var error = RestoreLogMessage.CreateError(NuGetLogCode.NU1001, string.Format(CultureInfo.CurrentCulture, Strings.Error_InvalidTargetFramework, frameworkString));
```
nkolev92: I know this is in draft, and we had chatted about validating the different scenarios. 
If we're going to add something like this, I think static graph might need some changes too.

martinrrm: Thanks for letting me know, I'll make sure I update static graph too

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
```diff
@@ -458,6 +458,12 @@ private static IEnumerable<TargetFrameworkInformation> GetTargetFrameworkInforma
             foreach (var item in GetItemByType(items, "TargetFrameworkInformation"))
             {
                 var frameworkString = item.GetProperty("TargetFramework");
+                if (frameworkString.Contains(';'))
```
nkolev92, line 461: Throwing from here probably leads to a bad failure in NuGet.exe?

In general, NuGet.exe expects a dg spec to create an this fails that. 

Can you double check NuGet.exe?

nkolev92 (commented):
There's some test failures that seem like actual failures.
It's also worth calling out in the description that this is only a CLI problem in the PR description and that VS errors are way better.

test/NuGet.Core.Tests/NuGet.Build.Tasks.Console.Test/MSBuildStaticGraphRestoreTests.cs
```diff
@@ -101,6 +101,19 @@ public void GetOriginalTargetFrameworks_WhenTargetFramworksSpecified_HasCorrectT
             actual.Should().BeEquivalentTo(expected);
         }
 
+        [Theory]
+        [InlineData("net40;net45")]
+        [InlineData("net40;net45 ; netstandard2.0 ")]
+        public void GetOriginalTargetFramework_WhenTargetFramworkSpecified_HasCorrectTargetFramework(string targetFrameworks)
```
jgonz120, line 107:
```suggestion
        public void GetOriginalTargetFramework_WhenMultipleTargetFramworkSpecified_ThrowRestoreCommandException(string targetFrameworks)
```

Might want to change the name of this test.

jgonz120 (approved):

PR #6092, opened 2024-10-09
[Pack] Handle empty Version tag in nuspec error

### LINKED ISSUE #7987: Nuget pack doesn't support blank <version> in .nuspec even though version is passed on the command line
## Details about Problem
When running `nuget pack` with a .nuspec file that contains an empty `version` and passing -Version on the command line, nuget will fail with the error:

> An error occured while trying to parse the value '' of property 'version' in the manifest file.
>   Value cannot be null or an empty string.
>   Parameter name: value

However, you put in a bogus version identifier in the nuspec file, the same command will run happily, with the version number passed on the command line.

NuGet product used:  NuGet.exe

NuGet version (x.x.x.xxx): 4.9.3.5777

OS version (i.e. win10 v1607 (14393.321)): Win 10 1903

## Detailed repro steps

1. Create a valid .nuspec file, but with an empty `version`, i.e. `<version></version>`

2. Run `nuget.exe pack the.nuspec -Version 1.2.3.4 `

3. See error.

4. Replace `version` in .nuspec file with `<version>0.0.0.0</version>`

5. Run same command as above

6. See success.

martinrrm:
# Bug

Fixes: https://github.com/NuGet/Home/issues/7987

## Context of the bug
When running nuget pack with a .nuspec file that contains an empty version and passing -Version on the command line, nuget will fail with the error:

```
An error occured while trying to parse the value '' of property 'version' in the manifest file.
Value cannot be null or an empty string.
Parameter name: value
``` 

However, you put in a bogus version identifier in the nuspec file, the same command will run happily, with the version number passed on the command line.

## Description
After having conversations, right now the solution is to override the nuspec version with the CLI version that the customer specified.

~This PR makes that when the Version tag is empty it creates an `0.0.0` version which will be used to pack the package.~

Currently how it works is that we read the .nuspec and validate that Version is not null, this validation is used by the `nuget spec` and `nuget pack` commands. ~Removing an error is a breaking change, so I'm not sure if this is the best solution but for me the scenario for an empty `Version` tag is not common and setting the `0.0.0` version sounds less disruptive. Also, I tried to lag a warning but looks like the logger is not close to the validation close.~

Since we run this validation when reading the file, I cannot check if we are using the `-Version` property easily, the other option for to fix this is to move that validation to another place where we can check it against pack args. 

## PR Checklist

- [x] Added tests **Waiting for team feedback before adding tests**

### DIFF
--- a/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/Manifest.cs
+++ b/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/Manifest.cs
@@ -146,7 +146,7 @@ namespace NuGet.Packaging
             var manifest = ManifestReader.ReadManifest(document);
 
             // Update manifest metadata version if version was provided by the CLI command
-            if (propertyProvider is not null)
+            if (propertyProvider is not null && propertyProvider.Target.GetType().Name.Equals("PackArgs"))
             {
                 var version = propertyProvider("version");
                 if (version is not null)
--- a/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/Manifest.cs
+++ b/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/Manifest.cs
@@ -12,6 +12,8 @@ using NuGet.Packaging.Core;
 using NuGet.Packaging.PackageCreation.Resources;
 using NuGet.Packaging.Xml;
 using static NuGet.Shared.XmlUtility;
+using NuGet.Versioning;
+
 
 #if !IS_CORECLR
 using System.Xml.Schema;
@@ -142,6 +144,16 @@ namespace NuGet.Packaging
 
             // Deserialize it
             var manifest = ManifestReader.ReadManifest(document);
+
+            // Update manifest metadata version if version was provided by the CLI command
+            if (propertyProvider is not null)
+            {
+                var version = propertyProvider("version");
+                if (version is not null)
+                {
+                    manifest.Metadata.Version = NuGetVersion.Parse(version);
+                }
+            }
 
             // Validate before returning
             Validate(manifest);
--- a/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs
+++ b/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs
@@ -88,7 +88,7 @@ namespace NuGet.Packaging
                     case "version":
                         NuGetVersion version;
                         NuGetVersion.TryParse(value, out version);
-                        manifestMetadata.Version = version ?? new NuGetVersion(0, 0, 0);
+                        manifestMetadata.Version = version;
                         break;
                     case "authors":
                         manifestMetadata.Authors = value?.Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
--- a/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs
+++ b/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs
@@ -87,8 +87,10 @@ namespace NuGet.Packaging
                         break;
                     case "version":
                         NuGetVersion version;
-                        NuGetVersion.TryParse(value, out version);
-                        manifestMetadata.Version = version;
+                        if (NuGetVersion.TryParse(value, out version))
+                        {
+                            manifestMetadata.Version = version;
+                        }
                         break;
                     case "authors":
                         manifestMetadata.Authors = value?.Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
--- a/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs
+++ b/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs
@@ -86,7 +86,9 @@ namespace NuGet.Packaging
                         manifestMetadata.Id = value;
                         break;
                     case "version":
-                        manifestMetadata.Version = NuGetVersion.Parse(value);
+                        NuGetVersion version;
+                        NuGetVersion.TryParse(value, out version);
+                        manifestMetadata.Version = version ?? new NuGetVersion(0, 0, 0);
                         break;
                     case "authors":
                         manifestMetadata.Authors = value?.Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
--- a/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs
+++ b/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs
@@ -86,8 +86,7 @@ namespace NuGet.Packaging
                         manifestMetadata.Id = value;
                         break;
                     case "version":
-                        NuGetVersion version;
-                        if (NuGetVersion.TryParse(value, out version))
+                        if (NuGetVersion.TryParse(value, out NuGetVersion version))
                         {
                             manifestMetadata.Version = version;
                         }
--- a/test/NuGet.Core.Tests/NuGet.Commands.Test/PackCommandRunnerTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Commands.Test/PackCommandRunnerTests.cs
@@ -49,6 +49,84 @@ namespace NuGet.Commands.Test
         }
 
         [Fact]
+        public void RunPackageBuild_WithInvalidNuspec_Throws()
+        {
+            using (var test = DefaultExclusionsTest.Create(validNuspec: false))
+            {
+                var args = new PackArgs()
+                {
+                    CurrentDirectory = test.CurrentDirectory.FullName,
+                    Exclude = Enumerable.Empty<string>(),
+                    Logger = NullLogger.Instance,
+                    Path = test.NuspecFile.FullName,
+                };
+                var runner = new PackCommandRunner(args, createProjectFactory: null);
+
+                try
+                {
+                    var result = runner.RunPackageBuild();
+                }
+                catch (Exception e)
+                {
+                    Assert.Equal(e.Message, "Version is required.");
+                }
+
+                var nupkgPath = Path.Combine(test.CurrentDirectory.FullName, $"DefaultExclusions.1.0.0.nupkg");
+                Assert.False(File.Exists(nupkgPath), "The output .nupkg should not exist when pack fails.");
+            }
+        }
+
+        [Fact]
+        public void RunPackageBuild_WithValidNuspec_WithVersionAsArgumentSucceds()
+        {
+            using (var test = DefaultExclusionsTest.Create(validNuspec: true))
+            {
+                var args = new PackArgs()
+                {
+                    CurrentDirectory = test.CurrentDirectory.FullName,
+                    Exclude = Enumerable.Empty<string>(),
+                    Logger = NullLogger.Instance,
+                    Path = test.NuspecFile.FullName,
+                    Version = "3.1.2"
+                };
+                var runner = new PackCommandRunner(args, createProjectFactory: null);
+
+                try
+                {
+                    var result = runner.RunPackageBuild();
+                }
+                catch (Exception e)
+                {
+                    Assert.Equal(e.Message, "Version is required.");
+                }
+
+                var nupkgPath = Path.Combine(test.CurrentDirectory.FullName, $"DefaultExclusions.1.0.0.nupkg");
+                Assert.False(File.Exists(nupkgPath), "The output .nupkg should not exist when pack fails.");
+            }
+        }
+
+        [Fact]
+        public void RunPackageBuild_WithInvalidNuspec_WithVersionAsArgumentSucceds()
+        {
+            using (var test = DefaultExclusionsTest.Create(validNuspec: false))
+            {
+                var args = new PackArgs()
+                {
+                    CurrentDirectory = test.CurrentDirectory.FullName,
+                    Exclude = Enumerable.Empty<string>(),
+                    Logger = NullLogger.Instance,
+                    Path = test.NuspecFile.FullName,
+                    Version = "3.1.2"
+                };
+                var runner = new PackCommandRunner(args, createProjectFactory: null);
+                var result = runner.RunPackageBuild();
+
+                var nupkgPath = Path.Combine(test.CurrentDirectory.FullName, $"DefaultExclusions.3.1.2.nupkg");
+                Assert.True(File.Exists(nupkgPath), "The output .nupkg should not exist when pack fails.");
+            }
+        }
+
+        [Fact]
         public void RunPackageBuild_WithGenerateNugetPackageFalse_ReturnsTrue()
         {
             using (var testDirectory = TestDirectory.Create())
@@ -112,7 +190,7 @@ namespace NuGet.Commands.Test
                 UnexpectedEntryNames = unexpectedEntryNames;
             }
 
-            internal static DefaultExclusionsTest Create()
+            internal static DefaultExclusionsTest Create(bool validNuspec = true)
             {
                 TestDirectory testDirectory = TestDirectory.Create();
 
@@ -137,21 +215,17 @@ namespace NuGet.Commands.Test
                 const string packageId = "DefaultExclusions";
                 const string packageVersion = "1.0.0";
 
-                File.WriteAllText(nuspecFile.FullName, $@"<?xml version=""1.0""?>
-<package>
-    <metadata>
-        <id>{packageId}</id>
-        <version>{packageVersion}</version>
-        <title>title</title>
-        <description>description</description>
-        <authors>author</authors>
-        <requireLicenseAcceptance>false</requireLicenseAcceptance>
-        <dependencies />
-    </metadata>
-    <files>
-        <file src=""{pattern}"" target="""" />
-    </files>   
-</package>");
+                if (validNuspec)
+                {
+                    ValidNuspec(nuspecFile.FullName, packageId, packageVersion, pattern);
+
+                }
+                else
+                {
+                    InvaidNuspecNoVersion(nuspecFile.FullName, packageId, pattern);
+
+                }
+
 
                 var nupkgFile = new FileInfo(Path.Combine(currentDirectory.FullName, $"{packageId}.{packageVersion}.nupkg"));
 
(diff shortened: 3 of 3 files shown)


src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs
```diff
@@ -86,7 +86,9 @@ private static void ReadMetadataValue(ManifestMetadata manifestMetadata, XElemen
                         manifestMetadata.Id = value;
                         break;
                     case "version":
-                        manifestMetadata.Version = NuGetVersion.Parse(value);
+                        NuGetVersion version;
```
nkolev92: How does pack behave when someone passes an actual empty or unparseable version on the commandline?

src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs
```diff
@@ -86,7 +86,9 @@ private static void ReadMetadataValue(ManifestMetadata manifestMetadata, XElemen
...
+                        NuGetVersion.TryParse(value, out version);
+                        manifestMetadata.Version = version ?? new NuGetVersion(0, 0, 0);
```
nkolev92: What if we just allowed it be null instead of a default? 
Do we ever need to differentiate between that and a 0,0,0?

Would we detect that no version has been set somewhere different?

Another idea would be, what if we didn't set a version element in the nuspec.

nkolev92 (commented):
A few questions.

src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs
martinrrm: We do a `NuGetVersion.TryParse` and if that fails we throw:

```Error NU5010: Version string specified for package reference ' ' is invalid.```
```Error NU5010: Version string specified for package reference 'sdf' is invalid.```

Empty string:
```Missing option value for: '-Version'```

File: src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs (hunk 2 above)
Comment (martinrrm) [reply src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs]: This version is used to create the nupkg name, so in this case an empty version in the nuspec will create the package as `package.0.0.0.nupkg` so most likely there is a way to detect that no version has been set later in the code.

> Another idea would be, what if we didn't set a version element in the nuspec.

I'm not sure what would this cause, for me it looks like version is very important

Comment (nkolev92) [reply src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs]: > I'm not sure what would this cause, for me it looks like version is very important

I was thinking as more of a test case actually. What would happen if there was no version specified in the nuspec when we attempt to pack and just pass it on the CLI. 

That feels like the closest analog to this test case.

Version is a mandatory element, so I'm just worried about the potential fall out of setting a default version.

How are token replacements done? Can we parse the CLI version and validate it before the we read the manifest?

File: src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs (hunk 1 above)
Comment (nkolev92) [reply src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs]: Can you link where that happens?

Comment (martinrrm) [reply src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs]: Here: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.CommandLine/Commands/PackCommand.cs#L131-L137

File: src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs (hunk 2 above)
Comment (martinrrm) [reply src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs]: > I was thinking as more of a test case actually. What would happen if there was no version specified in the nuspec when we attempt to pack and just pass it on the CLI.

Currently, if you don't have the version tag it will fail saying that the file is incorrect, if it is empty, it will fail because we try to parse that empty string and fail if the parsing returns null.

> Can we parse the CLI version and validate it before the we read the manifest?

We do validate the CLI version before reading the manifest, and after reading the manifest we check if we specified a CLI version and replace it with it. 

So I think after this conversation I think the best solution is:

1. Continue validating the CLI version
2. Don't immediately break when there is an empty Version in the nuspec
3. When we would do the "replacement", if CLI version is invalid and there is no version in nuspec break. Otherwise continue as how it is right now. (if CLI version is valid it would override the .nuspec version)

Comment (nkolev92) [reply src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs]: I was thinking, what if we validate once? 
Can we inject the CLI version into the nuspec before reading the manifest, or is that not possible?

My fear with not validating the manifest is that we're just introducing an unusual behavior that may lead to missing out a mandatory metadata.

File: src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs [hunk 3]
```diff
@@ -86,7 +86,9 @@ private static void ReadMetadataValue(ManifestMetadata manifestMetadata, XElemen
                         manifestMetadata.Id = value;
                         break;
                     case "version":
-                        manifestMetadata.Version = NuGetVersion.Parse(value);
+                        NuGetVersion version;
+                        NuGetVersion.TryParse(value, out version);
```
donnie-msft: It doesn't matter in this case, because our TryParse always resets version to `null`, but it bothers me when return values are ignored 🤷‍♂️

donnie-msft (commented):
The code looks fine to me, but since I don't know whether this is better, so just commenting for now.

src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs
martinrrm: I'm not sure if I understand your concern, `Parse` will return an error if it's unable to create the NuGetVersion, `TryParse` returns null. Are saying that `version` is ignored?

donnie-msft: > `TryParse` returns null. 

TryParse returns a non-nullable boolean. It's supposed to be used as an indication whether the parse was successful or not.

I believe this is the right reference, but I can't find any TryParse methods in our repo with a nullable return type.
https://github.com/NuGet/NuGet.Client/blob/8a5348f76ab0d2338dad0555f4abd7b59598c793/src/NuGet.Clients/NuGet.VisualStudio/LegacyTypes/SemanticVersion.cs#L156

martinrrm: I understand now, I updated the code, so I use the return value

src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs
martinrrm: Sorry for the late reply, I think I was able to address this concern, I'm validating the CLI version and updating the manifest version if it's a valid version.

test/NuGet.Core.Tests/NuGet.Commands.Test/PackCommandRunnerTests.cs
```diff
@@ -48,6 +48,84 @@ public void RunPackageBuild_WithDefaultExcludes_ExcludesDefaultExcludes()
...
+                    Assert.Equal(e.Message, "Version is required.");
+                }
+
+                var nupkgPath = Path.Combine(test.CurrentDirectory.FullName, $"DefaultExclusions.1.0.0.nupkg");
+                Assert.False(File.Exists(nupkgPath), "The output .nupkg should not exist when pack fails.");
+            }
+        }
+
+        [Fact]
+        public void RunPackageBuild_WithValidNuspec_WithVersionAsArgumentSucceds()
```
donnie-msft:
```suggestion
        public void RunPackageBuild_WithValidNuspec_WithVersionAsArgument_PackSucceeds()
```

test/NuGet.Core.Tests/NuGet.Commands.Test/PackCommandRunnerTests.cs
```diff
@@ -48,6 +48,84 @@ public void RunPackageBuild_WithDefaultExcludes_ExcludesDefaultExcludes()
...
+                    Assert.Equal(e.Message, "Version is required.");
+                }
+
+                var nupkgPath = Path.Combine(test.CurrentDirectory.FullName, $"DefaultExclusions.1.0.0.nupkg");
+                Assert.False(File.Exists(nupkgPath), "The output .nupkg should not exist when pack fails.");
+            }
+        }
+
+        [Fact]
+        public void RunPackageBuild_WithInvalidNuspec_WithVersionAsArgumentSucceds()
```
donnie-msft:
```suggestion
        public void RunPackageBuild_WithInvalidNuspec_WithVersionAsArgument_PackSucceeds()
```

test/NuGet.Core.Tests/NuGet.Commands.Test/PackCommandRunnerTests.cs
```diff
@@ -48,6 +48,84 @@ public void RunPackageBuild_WithDefaultExcludes_ExcludesDefaultExcludes()
...
+                };
+                var runner = new PackCommandRunner(args, createProjectFactory: null);
+
+                try
+                {
+                    var result = runner.RunPackageBuild();
+                }
+                catch (Exception e)
+                {
+                    Assert.Equal(e.Message, "Version is required.");
```
donnie-msft: `Assert.Throws<>(..)` would be better instead of catching to ensure the uncaught exception occurs.

But since the title of the test includes "Succeeds", I think this expected exception is supposed to be removed?

test/NuGet.Core.Tests/NuGet.Commands.Test/PackCommandRunnerTests.cs
```diff
@@ -137,21 +215,17 @@ internal static DefaultExclusionsTest Create()
...
+                if (validNuspec)
+                {
+                    ValidNuspec(nuspecFile.FullName, packageId, packageVersion, pattern);
+
+                }
+                else
+                {
+                    InvaidNuspecNoVersion(nuspecFile.FullName, packageId, pattern);
+
+                }
```
donnie-msft: blank lines

test/NuGet.Core.Tests/NuGet.Commands.Test/PackCommandRunnerTests.cs
```diff
@@ -48,6 +48,84 @@ public void RunPackageBuild_WithDefaultExcludes_ExcludesDefaultExcludes()
...
+                };
+                var runner = new PackCommandRunner(args, createProjectFactory: null);
+
+                try
+                {
+                    var result = runner.RunPackageBuild();
+                }
+                catch (Exception e)
+                {
+                    Assert.Equal(e.Message, "Version is required.");
```
donnie-msft: Assert.Throws<>(..) would be better instead of catching to ensure the uncaught exception occurs.

donnie-msft (changes requested):
Thanks for addressing my earlier feedback. The implementation looks good to me but some of the tests are unclear.

src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs
```diff
@@ -86,7 +86,11 @@ private static void ReadMetadataValue(ManifestMetadata manifestMetadata, XElemen
                         manifestMetadata.Id = value;
                         break;
                     case "version":
-                        manifestMetadata.Version = NuGetVersion.Parse(value);
+                        NuGetVersion version;
+                        if (NuGetVersion.TryParse(value, out version))
```
zivkan:
```suggestion
                        if (NuGetVersion.TryParse(value, out NuGetVersion version))
```

test/NuGet.Core.Tests/NuGet.Commands.Test/PackCommandRunnerTests.cs
```diff
@@ -48,6 +48,71 @@ public void RunPackageBuild_WithDefaultExcludes_ExcludesDefaultExcludes()
...
+                    Path = test.NuspecFile.FullName,
+                    Version = "3.1.2"
+                };
+                var runner = new PackCommandRunner(args, createProjectFactory: null);
+
+                var result = runner.RunPackageBuild();
+                Assert.True(result);
+
+                var nupkgPath = Path.Combine(test.CurrentDirectory.FullName, $"DefaultExclusions.3.1.2.nupkg");
+                Assert.True(File.Exists(nupkgPath), "nuspec file does not exist");
```
zivkan: If you used FluentAssertions, you could do `File.Exists.Should().BeTrue()`

test/NuGet.Core.Tests/NuGet.Commands.Test/PackCommandRunnerTests.cs
```diff
@@ -168,6 +227,47 @@ public void Dispose()
...
+                            <requireLicenseAcceptance>false</requireLicenseAcceptance>
+                            <dependencies />
+                        </metadata>
+                        <files>
+                            <file src=""{pattern}"" target="""" />
+                        </files>   
+                    </package>");
+            }
+
+            internal static void InvaidNuspecNoVersion(string nuspecPath, string packageId, string pattern)
```
zivkan: is this method just a special case of `ValidNuspec` where `string.Empty` is passed to the `packageVersion` argument?

test/NuGet.Core.Tests/NuGet.Commands.Test/PackCommandRunnerTests.cs
```diff
@@ -48,6 +48,71 @@ public void RunPackageBuild_WithDefaultExcludes_ExcludesDefaultExcludes()
...
+                var result = runner.RunPackageBuild();
+                Assert.True(result);
+
+                var nupkgPath = Path.Combine(test.CurrentDirectory.FullName, $"DefaultExclusions.3.1.2.nupkg");
+                Assert.True(File.Exists(nupkgPath), "nuspec file does not exist");
+            }
+        }
+
+        [Fact]
+        public void RunPackageBuild_WithInvalidNuspec_WithVersionAsArgument_PackSucceeds()
```
zivkan:
```suggestion
        public void RunPackageBuild_NuspecWithEmptyVersion_WithVersionAsArgument_PackSucceeds()
```

Describing the test scenario as "invalid nuspec" kind of suggests that using `<package />` should work

test/NuGet.Core.Tests/NuGet.Commands.Test/PackCommandRunnerTests.cs
```diff
@@ -48,6 +48,71 @@ public void RunPackageBuild_WithDefaultExcludes_ExcludesDefaultExcludes()
...
+
+                Assert.Throws<Exception>(() => runner.RunPackageBuild());
+
+                var nupkgPath = Path.Combine(test.CurrentDirectory.FullName, $"DefaultExclusions.1.0.0.nupkg");
+                Assert.False(File.Exists(nupkgPath), "The output .nupkg should not exist when pack fails.");
+            }
+        }
+
+        [Fact]
+        public void RunPackageBuild_WithValidNuspec_WithVersionAsArgument_PackSucceeds()
```
zivkan:
```suggestion
        public void RunPackageBuild_WithVersionAsArgument_NuspecVersionOverriddenByVersionArgument()
```

if the "scenario" part of the name doesn't mention inputs being invalid, I think it's safe to assume that the inputs are valid, so adding it to the name feels redundant.

Also, result "Pack Succeeds" is just saying it doesn't throw an exception (or some other kind of failure).  But the test is validating more than just no exception is thrown.

src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/Manifest.cs
```diff
@@ -143,6 +145,16 @@ public static Manifest ReadFrom(Stream stream, Func<string, string> propertyProv
             // Deserialize it
             var manifest = ManifestReader.ReadManifest(document);
 
+            // Update manifest metadata version if version was provided by the CLI command
+            if (propertyProvider is not null && propertyProvider.Target.GetType().Name.Equals("PackArgs"))
```
zivkan, line 149: I think I'd rather see a public API change, rather than adding a condition based on a delegate's type name, or other uses of reflection.

Before I spend time looking into the code paths too deeply myself, what other approaches have you considered?  I was wondering if having an additional `Action<Manifest>` delegate (or `Func<Manifest, Manifest>` delegate, so the method signature/public API wouldn't need to change if `Manifest` eventually becomes immutable) would allow this method to provide extensibility after deserialization and before validation.  But looking just a little bit into the code, that would need a bunch of public APIs on PackageBuilder as well, to pass through to here.  I still feel like it might be better than checking the delegate's type name though

Did you try not putting this condition, and always overriding the version when `propertyProvider` has a value for it? What negative side effects are there?

also, the PR description talks about setting the value to 0.0.0, which is no longer happening, so the PR description needs to be updated.

martinrrm: > Did you try not putting this condition, and always overriding the version when propertyProvider has a value for it? What negative side effects are there?
Yes, that was my first implementation but realized that some tests were failing. After doing some debugging, I realized that this function is used in multiple scenarios, the one I'm trying to solve is when the customer is using a nuspec.

The tests that were failing is in the [scenario when we are using pack and a .csproj and nuspec](https://github.com/NuGet/NuGet.Client/blob/dev/test/NuGet.Clients.Tests/NuGet.CommandLine.Test/NuGetPackCommandTest.cs#L6137). That path calls `Manifest.ReadFrom` [using another propertyProvider](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.CommandLine/Commands/ProjectFactory.cs#L1266) which represents the project file (.csproj) not the nuspec, so the information wasn't correct.

martinrrm: I looked at other approach's but given that this is very specific scenario I wanted to do minimal changes.

src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestReader.cs
donnie-msft: 👍

test/NuGet.Core.Tests/NuGet.Commands.Test/PackCommandRunnerTests.cs
```diff
@@ -137,21 +202,15 @@ internal static DefaultExclusionsTest Create()
...
-    </files>   
-</package>");
+                if (validNuspec)
+                {
+                    ValidNuspec(nuspecFile.FullName, packageId, packageVersion, pattern);
+
+                }
+                else
+                {
+                    InvaidNuspecNoVersion(nuspecFile.FullName, packageId, pattern);
```
donnie-msft:
```suggestion
                    InvalidNuspecNoVersion(nuspecFile.FullName, packageId, pattern);
```

donnie-msft (dismissed):

donnie-msft (approved):

nkolev92 (approved):

PR #6094, opened 2024-10-10
Updated PM UI to render the README inside of a tab in the README Details pane. 

jgonz120:
# Bug

## Description
Updates several views to allow for rendering the README in the PM UI.

### Screenshots
In the browse tab the README is not being rendered even though the package has been installed and exists in the GPF.
image

In the installed tab the README shows up as a new tab in the package details pane.
image

If a package is downloaded locally but no README is found, we give the user a message
image

Due to the current limitations with WebViews they cannot be allowed to move "off screen" because they always overlay. The UX was updated to remove min widths. 
ReadmeResizingGif

## PR Checklist

- [X] Added tests


src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
```diff
@@ -1076,4 +1076,22 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
...
+    <value>README</value>
+  </data>
+  <data name="Text_NoReadme" xml:space="preserve">
+    <value>Could not read package README from selected package.
+
+Only the package maintainer can add a README.
+
+If you are not the maintainer, please consider filing an issue or contacting the maintainer to request a README.
+
+For instructions on how to add a README, please visit [aka.ms/nuget/readme](https://aka.ms/nuget/readme)</value>
```
martinrrm, line 1095: We don't want the url to be localized, we could add the text later or localize that part of the string with https://ceapex.visualstudio.com/CEINTL/_wiki/wikis/CEINTL.wiki/8680/Localizability-Commenting-Overview?anchor=commonly-used-functional-commenting-instruction-(locver-rule)

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
```diff
@@ -1076,4 +1076,22 @@ Please see https://aka.ms/troubleshoot_nuget_cache for more help.</value>
   <data name="ColumnHeader_PackageLevel" xml:space="preserve">
     <value>Package Level</value>
   </data>
+  <data name="Label_PackageDetails" xml:space="preserve">
+    <value>Package Details</value>
+  </data>
+  <data name="Error_UnableToLoadReadme" xml:space="preserve">
+    <value>An error occurred while trying to load the README.</value>
```
martinrrm, line 1083: Not sure if we want to localize README too, maybe we should lock that string https://ceapex.visualstudio.com/CEINTL/_wiki/wikis/CEINTL.wiki/8680/Localizability-Commenting-Overview?anchor=commonly-used-functional-commenting-instruction-(locver-rule)

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
jgonz120: added a lock.

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
jgonz120: There are a couple of strings that contain the word "Readme" which are being localized. It looks like in Spanish it's currently being localized to Léame, which seems right? Happy to lock the read README but if we do, we should do it everywhere.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
```diff
@@ -14,12 +17,12 @@ public sealed class ReadmePreviewViewModel : ViewModelBase
     {
         private bool _canDetermineReadmeDefined;
         private bool _errorLoadingReadme;
-        private INuGetPackageFileService _packageFileService;
+        private IServiceBroker _serviceBroker;
         private string _rawReadme;
 
-        public ReadmePreviewViewModel(INuGetPackageFileService packageFileService)
+        public ReadmePreviewViewModel(IServiceBroker serviceBroker)
```
jebriede: It's preferable to pass the dependencies we need in our constructor rather than passing the entire ServiceBroker for several reasons, including but not limited to:
- It's easier to reason about this object's dependencies when they are explicitly defined in the constructor. Passing the ServiceBroker means that this object could depend on anything that the service broker can provide.
- It's easier to mock the dependencies when testing, rather than having to mock the servicebroker's return when a specific service is requested

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadmeControl.xaml.cs
```diff
@@ -0,0 +1,177 @@
...
+        {
+            if (newValue is not null &&
+                (oldValue?.Id != newValue.Id ||
+                oldValue?.Version != newValue.Version))
+            {
+                var loadCts = new CancellationTokenSource();
+                var oldCts = Interlocked.Exchange(ref _loadCts, loadCts);
+                oldCts?.Cancel();
+                oldCts?.Dispose();
+
```
jebriede: Remove extra blank line

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadmeControl.xaml.cs
```diff
@@ -0,0 +1,177 @@
...
+            if (e.OldValue is ReadmePreviewViewModel oldViewModel && oldViewModel is not null)
+            {
+                oldViewModel.PropertyChanged -= ViewModel_PropertyChangedAsync;
+            }
+            ReadmeViewModel.PropertyChanged += ViewModel_PropertyChangedAsync;
+        }
+
+#pragma warning disable VSTHRD100 // Avoid async void methods
+        private async void ViewModel_PropertyChangedAsync(object sender, System.ComponentModel.PropertyChangedEventArgs e)
+#pragma warning restore VSTHRD100 // Avoid async void methods
```
jebriede: It's best to keep the event handler synchronous and use JoinableTaskFactory.RunAsync to call an async method. An async event handler will swallow exceptions. Alternatively, we should handle exceptions in the async void method, but it's preferable to avoid async methods returning void if possible.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadmeControl.xaml.cs
```diff
@@ -0,0 +1,177 @@
...
+            {
+                if (ReadmeViewModel.CanDetermineReadmeDefined)
+                {
+                    markdown = string.IsNullOrWhiteSpace(ReadmeViewModel.ReadmeMarkdown)
+                        ? UI.Resources.Text_NoReadme
+                        : ReadmeViewModel.ReadmeMarkdown;
+                }
+            }
+
+#pragma warning disable CA1031 // Do not catch general exception types
```
jebriede: Do we know what exceptions we could be encountering from this call? It would be better to catch those explicitly.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadmeControl.xaml.cs
```diff
@@ -0,0 +1,177 @@
...
+        private async Task UpdateMarkdownAsync(string markDown)
+        {
+            if (_markdownPreview is null)
+            {
+#pragma warning disable CS0618 // Type or member is obsolete
+                _markdownPreview = new PreviewBuilder().Build();
+                descriptionMarkdownPreview.Content = _markdownPreview.VisualElement;
+#pragma warning restore CS0618 // Type or member is obsolete
+            }
+            if (DataContext is ReadmePreviewViewModel viewModel)
```
jebriede: We can remove one nested if by combining the conditional.

```suggestion
            if (markdown is not null && DataContext is ReadmePreviewViewModel viewModel)
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadmeControl.xaml.cs
```diff
@@ -0,0 +1,177 @@
...
+            }
+#pragma warning restore CA1031 // Do not catch general exception types
+        }
+
+        private async Task UpdateMarkdownAsync(string markDown)
+        {
+            if (_markdownPreview is null)
+            {
+#pragma warning disable CS0618 // Type or member is obsolete
+                _markdownPreview = new PreviewBuilder().Build();
```
jebriede: Question: Are we required to build a new markdown preview every time we want to set new content? Or can we just do this once and call UpdateContentAsync on it to update the content when it changes?

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadmeControl.xaml.cs
```diff
@@ -0,0 +1,177 @@
...
+                descriptionMarkdownPreview.Content = _markdownPreview.VisualElement;
+#pragma warning restore CS0618 // Type or member is obsolete
+            }
+            if (DataContext is ReadmePreviewViewModel viewModel)
+            {
+                if (markDown is not null)
+                {
+                    await _markdownPreview.UpdateContentAsync(markDown, ScrollHint.None);
+                    descriptionMarkdownPreview.Visibility = string.IsNullOrEmpty(markDown) ? Visibility.Collapsed : Visibility.Visible;
```
jebriede: We should handle all Visibility of this control in XAML by binding to the appropriate condition of state in the ViewModel. This applies to other places in this PR where this is set in code behind as well.

jebriede: At this point `markDown` cannot be null (but it could be empty), since we already checked if it was null before entering this code block.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataControl.xaml.cs
```diff
@@ -1,85 +1,114 @@
...
+            {
+                case PackageMetadataTab.PackageDetails:
+                    SelectedTabItem = tabPackageDetails;
+                    break;
+                case PackageMetadataTab.Readme:
+                default:
+                    SelectedTabItem = tabReadMe;
+                    break;
             }
         }
```
jebriede: I'm not sure where to best leave this comment but since it's relevant to SelectTab, and the overall design of tab switching and ViewModel and View relationship, I'll leave it here. This comment is broader and applies to the Tabs and how the select tab is determined and how the View handles the presentation of the selected tabs:

A design that adheres to the MVVM pattern would be to keep a list of possible ViewModels for each of the tabs in the parent ViewModel (or have fixed properties with each of the tab view models), and expose a SelectedTab property. Then, the view can customize how the presentation will be using data templates. We can programatically set the selected tab. I wrote some code to demonstrate this behavior that I'll include here. Please note that it's missing INotifyPropertyChanged behaviors which you'll want to have in your view models and is a simple example, but it demonstrates the pattern better than I could simply explain in words:
MainView.xaml:
```
<Window x:Class="WpfApp12.MainWindow"
        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
        xmlns:local="clr-namespace:WpfApp12"
        mc:Ignorable="d"
        Title="MainWindow" Height="450" Width="800">
    <Window.DataContext>
        <local:MainViewModel />
    </Window.DataContext>
    <Grid>
        <TabControl ItemsSource="{Binding Tabs}" SelectedItem="{Binding SelectedTab}">
            <TabControl.ItemTemplate>
                <DataTemplate>
                    <!-- Customize the tab header here -->
                    <TextBlock Text="{Binding Header}" />
                </DataTemplate>
            </TabControl.ItemTemplate>
            <TabControl.ContentTemplate>
                <DataTemplate>
                    <!-- Customize the content of the tabs here -->
                    <ContentControl Content="{Binding}">
                        <ContentControl.Resources>
                            <DataTemplate DataType="{x:Type local:SpecificTabViewModel}">
                                <StackPanel>
                                    <TextBlock Text="I'm Specific Tab!" />
                                    <Button Content="{Binding Content}" />
                                </StackPanel>
                            </DataTemplate>
                            <DataTemplate DataType="{x:Type local:AnotherTabViewModel}">
                                <StackPanel>
                                    <TextBlock Text="I'm Another Tab!" />
                                    <Label Content="{Binding Content}" />
                                </StackPanel>
                            </DataTemplate>
                        </ContentControl.Resources>
                    </ContentControl>
                </DataTemplate>
            </TabControl.ContentTemplate>
        </TabControl>
    </Grid>
</Window>
```

MainView.xaml.cs (notice there is no code behind to handle any of the interactions)
```
using System.Text;
using System.Windows;
using System.Windows.Controls;
using System.Windows.Data;
using System.Windows.Documents;
using System.Windows.Input;
using System.Windows.Media;
using System.Windows.Media.Imaging;
using System.Windows.Navigation;
using System.Windows.Shapes;

namespace WpfApp12
{
    /// <summary>
    /// Interaction logic for MainWindow.xaml
    /// </summary>
    public partial class MainWindow : Window
    {
        public MainWindow()
        {
            InitializeComponent();
        }
    }
}
```

Several View Models:
```
using System.Collections.ObjectModel;

namespace WpfApp12
{
    public class ViewModelBase
    {

    }

    public class MainViewModel : ViewModelBase
    {
        public MainViewModel()
        {
            Tabs = new ObservableCollection<TabViewModelBase>
            {
                new SpecificTabViewModel { Header = "Specific", Content = "Specific content" },
                new AnotherTabViewModel { Header = "Another", Content = "Another content" }
            };

            SelectedTab = Tabs[1];
        }

        public ObservableCollection<TabViewModelBase> Tabs { get; set; }
        public TabViewModelBase SelectedTab { get; set; }
    }

    public class TabViewModelBase : ViewModelBase
    {
        public string Header { get; set; }
        public string Content { get; set; }
    }

    public class SpecificTabViewModel : TabViewModelBase
    {
    }

    public class AnotherTabViewModel : TabViewModelBase
    {
    }
}
```

In this demonstration, I set the SelectedTab when constructing the MainViewModel. In a real-world example, you'll want this property to notify of property changes so the binding will update in XAML, and you may want to programatically set the selected tab elsewhere, instead of during object construction.

We make use of the ItemTemplates and ContentTemplates to define how the tab headers and the content of the tab will appear, and we use different data types to present different content, based on the type of ViewModel we are presenting.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadmeControl.xaml.cs
```diff
@@ -0,0 +1,177 @@
...
+
+#pragma warning disable CA1031 // Do not catch general exception types
+            try
+            {
+                await UpdateMarkdownAsync(markdown);
+            }
+            catch (Exception ex)
+            {
+                ReadmeViewModel.ErrorLoadingReadme = true;
+                descriptionMarkdownPreview.Visibility = Visibility.Collapsed;
```
jebriede: Visibility should be determined in a XAML binding. For example, when ErrorLoadingReadme is true/false.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataControl.xaml.cs
```diff
@@ -1,85 +1,114 @@
...
+            ReadmePreviewViewModel.PropertyChanged += ReadMePreviewViewModel_PropertyChanged;
+        }
 
-                NuGetUIThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+        private void ReadMePreviewViewModel_PropertyChanged(object sender, System.ComponentModel.PropertyChangedEventArgs e)
+        {
+            if (ReadmePreviewViewModel is not null
+                && e.PropertyName == nameof(ReadmePreviewViewModel.CanDetermineReadmeDefined))
+            {
+                SetReadmeTabVisibility(ReadmePreviewViewModel.CanDetermineReadmeDefined ? Visibility.Visible : Visibility.Collapsed);
```
jebriede: This ReadmePreviewViewModel is the ViewModel (and data context) of the PackageMetadataReadmeControl. The parent control should understand if it needs to render its children and should not need to rely on the state of the ViewModel of the child object. Such a design makes the ViewModel of another View also something the parent view responds to but is not its data context.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadmeControl.xaml.cs
```diff
@@ -0,0 +1,177 @@
...
+        internal CancellationTokenSource _loadCts;
+
+        private bool _renderLocalReadme = false;
+
+        public static readonly DependencyProperty PackageMetadataProperty =
+            DependencyProperty.Register(
+                nameof(PackageMetadata),
+                typeof(DetailedPackageMetadata),
+                typeof(PackageMetadataReadmeControl),
+                new PropertyMetadata(OnPropertyChanged));
```
jebriede: We can simplify how the dependency property is defined and not specify the [PropertyChangedCallback](https://learn.microsoft.com/en-us/dotnet/api/system.windows.propertymetadata.propertychangedcallback?view=windowsdesktop-8.0) but instead handle any code that needs to be invoked in the backing property's setter. That way we also avoid duplicating the call to `UpdateControl` that we currently have, once in the setter, once in the property changed. Here's an example how you can use GetValue/SetValue for the dependency property to keep the property on your object in sync:
https://learn.microsoft.com/en-us/dotnet/desktop/wpf/properties/how-to-implement-a-dependency-property?view=netdesktop-8.0

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadmeControl.xaml.cs
```diff
@@ -0,0 +1,177 @@
...
+            {
+                return;
+            }
+
+            if (disposing)
+            {
+                _loadCts.Cancel();
+                _loadCts.Dispose();
+                _markdownPreview?.Dispose();
+                var viewModel = (ReadmePreviewViewModel)DataContext;
```
jebriede: We have a property that casts this ViewModel already so we should use that.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadmeControl.xaml.cs
```diff
@@ -0,0 +1,177 @@
...
+            var markdown = string.Empty;
+            if (ReadmeViewModel is not null)
+            {
+                if (ReadmeViewModel.CanDetermineReadmeDefined)
+                {
+                    markdown = string.IsNullOrWhiteSpace(ReadmeViewModel.ReadmeMarkdown)
+                        ? UI.Resources.Text_NoReadme
+                        : ReadmeViewModel.ReadmeMarkdown;
+                }
+            }
```
jebriede:
```suggestion
            if (ReadmeViewModel?.CanDetermineReadmeDefined == true)
            {
                markdown = string.IsNullOrWhiteSpace(ReadmeViewModel.ReadmeMarkdown)
                    ? UI.Resources.Text_NoReadme
                    : ReadmeViewModel.ReadmeMarkdown;
            }
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadmeControl.xaml.cs
```diff
@@ -0,0 +1,177 @@
...
+{
+
+    /// <summary>
+    /// Interaction logic for PackageMetadataReadmeControl.xaml
+    /// </summary>
+    public partial class PackageMetadataReadmeControl : UserControl, IDisposable
+    {
+        internal CancellationTokenSource _loadCts;
+
+        private bool _renderLocalReadme = false;
```
jebriede: Whether a local readme is used is business logic that should be contained in the ViewModel, even though it's set by an operation in the View (which tab or `Filter` we're on).

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
```diff
@@ -43,33 +46,47 @@ public bool CanDetermineReadmeDefined
             set => SetAndRaisePropertyChanged(ref _canDetermineReadmeDefined, value);
         }
 
-        public async Task LoadReadmeAsync(string rawReadmeUrl, CancellationToken cancellationToken)
+        public async Task LoadReadmeAsync(DetailedPackageMetadata packageMetadata, bool renderLocalReadme, CancellationToken cancellationToken)
```
jebriede: Can we consider a design where the DetailedPackageMetadata is the Model for the ReadmePreviewViewModel and when the View tells it when to LoadReadmeAsync, which will fetch the readme contents and set a property called ReadmeMarkdown on the ViewModel that will be updated, the View will be listening for property changes for the ReadmeMarkdown property, and will tell the Markdown Preview control to update?

The ViewModel can then be responsible for wrapping the DetailedPackageMetadata and when it changes, decide how to handle the ReadmeMarkdown property it previously set for the last model. One approach would be to simply clear it because the model is now a new one. Or hypothetically, we could decide to have it cache that entry before clearing it. Or we could change it to delegate this responsibility to another class seamlessly. All this flexibility comes from handling the model in the ViewModel directly rather than just passing the model in the ViewModel's method to load the readme, and requiring the View to coordinate the rest.

jebriede (changes requested):
Thanks for breaking up the PR into smaller pieces! The design becomes much easier to understand this way. I left some comments that will likely result in some changes. Please let me know if you have any questions at all or when a new update is available to review. Thanks!

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadmeControl.xaml.cs
jebriede: Oops, I didn't notice the check for this being null before we build a new one! A suggestion would be to have a private property whose getter handles the lazy instantiation and asks if it's null, then builds a new one and returns it, otherwise just returns the instantiated one. But this works too. :-)

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
```diff
@@ -141,6 +143,11 @@ private async ValueTask InitializeAsync(PackageManagerModel model, INuGetUILogge
             _settingsKey = await GetSettingsKeyAsync(CancellationToken.None);
             UserSettings settings = LoadSettings();
             InitializeFilterList(settings);
+            var isReadmeTabEnabled = await _nugetFeatureFlagService.IsFeatureEnabledAsync(NuGetFeatureFlagConstants.RenderReadmeInPMUI);
+            _packageDetail._packageMetadataControl.InitializedReadmePreviewViewModel(Model.Context.ServiceBroker);
```
jebriede: I would recommend method names be verbs.

```suggestion
            _packageDetail._packageMetadataControl.InitializeReadmePreviewViewModel(Model.Context.ServiceBroker);
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadmeControl.xaml.cs
jgonz120: I do not, would have to check with the editor team

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadmeControl.xaml.cs
jgonz120: Okay so running a fire and forget RunAsync gives me a warning: https://github.com/Microsoft/VSSDK-Analyzers/blob/main/doc/VSSDK007.md

Running the property change synchronously could be an issue because it would block when rendering the readme, which I assume could be a problem if it takes a while to load because of the readme size. I think for now I'm going to scrap the property changed all together, since the only time the VM is updated is from the view. I know there's another comment about setting the model for the VM so this might change again

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataControl.xaml.cs
jgonz120: That makes sense, maybe you can help me with the design here. We do not want to show the README tab if there is no readme available, and we do not know if there's a readme available online. We cannot determine this until we try to render the README in the view model.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadmeControl.xaml.cs
jgonz120: So I should move this to the VM and expose a function to update the value, which the view will use?

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
donnie-msft: Maybe check with @lyndaidaii to align on localization strategy with nuget.org.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadmeControl.xaml.cs
```diff
@@ -0,0 +1,147 @@
...
+
+        private async Task UpdateMarkdownAsync()
+        {
+            var markdown = string.Empty;
+            if (ReadmeViewModel?.CanDetermineReadmeDefined == true)
+            {
+                markdown = string.IsNullOrWhiteSpace(ReadmeViewModel.ReadmeMarkdown)
+                    ? UI.Resources.Text_NoReadme
+                    : ReadmeViewModel.ReadmeMarkdown;
+            }
```
donnie-msft: If there's no Readme defined, should this be an `else` where Visibility is collapsed and ErrorLoadingReadme is false (not an error, just no readme available)?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/MultiSourcePackageMetadataProvider.cs
```diff
@@ -255,15 +256,34 @@ private async Task<IPackageSearchMetadata> GetPackageMetadataAsync(PackageIdenti
...
+                    tasks.Add(x.GetPackageMetadataFromLocalSourceAsync(identity, cancellationToken)));
+            }
+
             IEnumerable<IPackageSearchMetadata> completed = (await Task.WhenAll(tasks))
                 .Where(m => m != null);
 
-            IPackageSearchMetadata master = completed.FirstOrDefault(m => !string.IsNullOrEmpty(m.Summary))
-                ?? completed.FirstOrDefault()
-                ?? PackageSearchMetadataBuilder.FromIdentity(identity).Build();
+            PackageSearchMetadataBuilder metadataBuilder = completed.Where(m => !string.IsNullOrEmpty(m.Summary)).Select(PackageSearchMetadataBuilder.FromMetadata).FirstOrDefault()
```
donnie-msft: Curious what problem warranted this change, which as I can tell, ensures that FromMetadata is called on all `completed` metadata items.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/MultiSourcePackageMetadataProvider.cs
```diff
@@ -255,15 +256,34 @@ private async Task<IPackageSearchMetadata> GetPackageMetadataAsync(PackageIdenti
...
+                  .Build() as PackageSearchMetadataBuilder.ClonedPackageSearchMetadata;
+
+            if (string.IsNullOrWhiteSpace(clonedResult.ReadmeFileUrl))
+            {
+                clonedResult.ReadmeFileUrl = completed.Select(m => m.ReadmeFileUrl).FirstOrDefault(u => !string.IsNullOrWhiteSpace(u));
+            }
+
+            if (string.IsNullOrWhiteSpace(clonedResult.PackagePath))
+            {
+                clonedResult.PackagePath = completed.Select(m => (m as PackageSearchMetadataBuilder.ClonedPackageSearchMetadata)?.PackagePath).FirstOrDefault(u => !string.IsNullOrWhiteSpace(u));
```
donnie-msft: Even though it's just a LINQ expression, it's good to use clear variable names. Is the `u` supposed to indicate a URI string?

```suggestion
                clonedResult.PackagePath = completed.Select(m => (m as PackageSearchMetadataBuilder.ClonedPackageSearchMetadata)?.PackagePath).FirstOrDefault(uri => !string.IsNullOrWhiteSpace(uri));
```

donnie-msft: This feels like it belongs within the metadata builder, similar to the other extensions and merging methods. Can move it later if you agree that will work.

test/NuGet.Clients.Tests/NuGet.PackageManagement.UI.Test/ViewModels/ReadmePreviewViewModelTests.cs
```diff
@@ -50,9 +50,11 @@ public async Task LoadReadmeAsync_WithoutReadmeUrl_NoReadmeReturned(string readm
             //Arrange
             var mockFileService = new Mock<INuGetPackageFileService>();
             var target = new ReadmePreviewViewModel(mockFileService.Object);
+            var package = new DetailedPackageMetadata();
+            package.ReadmeFileUrl = readmeUrl;
 
             //Act
-            await target.LoadReadmeAsync(readmeUrl, CancellationToken.None);
+            await target.LoadReadmeAsync(package, false, CancellationToken.None);
```
donnie-msft:
```suggestion
            await target.LoadReadmeAsync(package, ?: false, CancellationToken.None);
```

test/NuGet.Clients.Tests/NuGet.PackageManagement.UI.Test/ViewModels/ReadmePreviewViewModelTests.cs
```diff
@@ -61,22 +63,66 @@ public async Task LoadReadmeAsync_WithoutReadmeUrl_NoReadmeReturned(string readm
...
+
 
             //Assert
             Assert.False(target.ErrorLoadingReadme);
             Assert.Equal(readmeContents, target.ReadmeMarkdown);
             Assert.True(target.CanDetermineReadmeDefined);
         }
+
+        [Fact]
+        public async Task LoadReadmeAsync_WithLocalReadmeUrl_RenderLocalReadmeFalse_ReadmeReturned()
```
donnie-msft: The test checks for no readme content. Is the test name wrong or is the assertion wrong?

```suggestion
        public async Task LoadReadmeAsync_WithLocalReadmeUrl_RenderLocalReadmeFalse_NoReadmeReturned()
```

test/NuGet.Clients.Tests/NuGet.PackageManagement.UI.Test/ViewModels/ReadmePreviewViewModelTests.cs
```diff
@@ -61,22 +63,66 @@ public async Task LoadReadmeAsync_WithoutReadmeUrl_NoReadmeReturned(string readm
...
+
+
+            //Assert
+            Assert.False(target.ErrorLoadingReadme);
+            Assert.Equal(string.Empty, target.ReadmeMarkdown);
+            Assert.False(target.CanDetermineReadmeDefined);
+        }
+
+        [Fact]
+        public async Task LoadReadmeAsync_WithLocalReadmeUrl_FileNotFound_Null()
```
donnie-msft: File Not Found to me should be an error - why is ErrorLoadingReadme false?

Name doesn't match assertion:
```suggestion
        public async Task LoadReadmeAsync_WithLocalReadmeUrl_FileNotFound_EmptyReadmeMarkdown()
```

donnie-msft (approved):

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadmeControl.xaml.cs
jgonz120: I found that the setter is never called when I place a breakpoint in it which is why I added the OnPropertyChanged.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataReadmeControl.xaml.cs
jgonz120: Yes, I updated it so we collapse the visibility of the tab if the markdown is empty

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/MultiSourcePackageMetadataProvider.cs
jgonz120: I think I just wanted them all to be a ClonedPackageSearchMetadata but seems like that wasn't necessary.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/MultiSourcePackageMetadataProvider.cs
jgonz120: I'm trying to avoid making changes to the protocol until I've made sure it's something we want to support forever.

test/NuGet.Clients.Tests/NuGet.PackageManagement.UI.Test/ViewModels/ReadmePreviewViewModelTests.cs
jgonz120: I updated the tests I think this has been resolved.

test/NuGet.Clients.Tests/NuGet.PackageManagement.UI.Test/ViewModels/ReadmePreviewViewModelTests.cs
jgonz120: The NuGetPackageFileService returns null if the file isn't found, which I think is okay. There isn't anything for the user to do if a README is defined in the nuspec but not actually in the package. Giving them the message that there is no readme and to contact the package owner is the best course of action. That said I updated the test name

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageDetailsTabViewModel.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        internal void SelectTab(PackageMetadataTab selectedPackageMetadataTab)
+        {
+            switch (selectedPackageMetadataTab)
+            {
+                case PackageMetadataTab.Readme:
+                    SelectedTab = Tabs[0];
+                    break;
+                case PackageMetadataTab.PackageDetails:
+                    SelectedTab = Tabs[1];
```
donnie-msft: Another idea is how I implemented the PMUI tabs with a Tag that contains the enum it represents.
That may help avoid indexing into the collection and support more dynamic tab layouts in the future.

https://github.com/NuGet/NuGet.Client/blob/fbb1d63c6adb94659d483ff581b99f64d9a56937/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerTopPanel.xaml.cs#L337

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageDetailsTabViewModel.cs
```diff
@@ -0,0 +1,122 @@
...
+                    break;
+            }
+        }
+
+        public ReadmePreviewViewModel ReadmePreviewViewModel { get; private set; }
+
+        public DetailControlModel DetailControlModel { get; private set; }
+
+        public ObservableCollection<TabViewModelBase> Tabs { get; private set; }
```
donnie-msft: I don't think this collection of Tabs changes, so I'd just use a simpler List that's readonly and create/initialize the list in the `Initialize`.

```suggestion
        public IReadOnlyList<TabViewModelBase> Tabs { get; private set; }
```

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/TabViewModelBase.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            get => _header;
+            set
+            {
+                SetAndRaisePropertyChanged(ref _header, value);
+            }
+        }
+
+        private bool _visible;
+        public bool Visible
```
donnie-msft: nit:
```suggestion
        public bool IsVisible
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
```diff
@@ -24,7 +24,7 @@ namespace NuGet.PackageManagement.UI
     /// The base class of PackageDetailControlModel and PackageSolutionDetailControlModel.
     /// When user selects an action, this triggers version list update.
     /// </summary>
-    public abstract class DetailControlModel : INotifyPropertyChanged, IDisposable
+    public abstract class DetailControlModel : TabViewModelBase, IDisposable
```
donnie-msft: Should this really be a Tab VM? I thought tabs were part of the detail control, not representing the entire control.

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
```diff
@@ -385,11 +389,9 @@ private async Task<IReadOnlyList<PackageDependency>> GetDependencies(IProjectCon
         // Called after package install/uninstall.
         public abstract Task RefreshAsync(CancellationToken cancellationToken);
 
-        public event PropertyChangedEventHandler PropertyChanged;
-
         protected void OnPropertyChanged(string propertyName)
         {
-            PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
+            RaisePropertyChanged(propertyName);
```
donnie-msft, line 393: nit: I agree with not changing every instance in this PR (unless you feel motivated to do so), but changing just this one to be different is probably more confusing to the reader.

```suggestion
            OnPropertyChanged(propertyName);
```

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageDetailsTabViewModel.cs
```diff
@@ -0,0 +1,122 @@
...
+        public void Initialize(DetailControlModel detailControlModel, INuGetPackageFileService nugetPackageFileService, ItemFilter currentFilter, bool readmeTabAvailable)
+        {
+            ReadmePreviewViewModel = new ReadmePreviewViewModel(nugetPackageFileService, currentFilter);
+            ReadmePreviewViewModel.Visible = readmeTabAvailable;
+            DetailControlModel = detailControlModel;
+            _readmeTabEnabled = readmeTabAvailable;
+
+            Tabs.Add(ReadmePreviewViewModel);
+            Tabs.Add(DetailControlModel);
+
```
donnie-msft, line 50: blank line

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
```diff
@@ -141,6 +145,12 @@ private async ValueTask InitializeAsync(PackageManagerModel model, INuGetUILogge
             _settingsKey = await GetSettingsKeyAsync(CancellationToken.None);
             UserSettings settings = LoadSettings();
             InitializeFilterList(settings);
+            var isReadmeTabEnabled = await _nugetFeatureFlagService.IsFeatureEnabledAsync(NuGetFeatureFlagConstants.RenderReadmeInPMUI);
+
+            _nugetPackageFileService?.Dispose();
+            _nugetPackageFileService = await _serviceBroker.GetProxyAsync<INuGetPackageFileService>(NuGetServices.PackageFileService, CancellationToken.None);
+            _packageDetail._packageDetailsTabControl.PackageDetailsTabViewModel.Initialize(_detailModel, _nugetPackageFileService, _topPanel.Filter, isReadmeTabEnabled);
+            InitializeSelectedPackageMetadataTab(settings, isReadmeTabEnabled);
```
donnie-msft: It'd probably be better to orchestrate the details pane's tabs in the DetailControlModel rather than in the View. This would also reduce how far and wide the Tab control reaches across these Views and ViewModels (eg, here in the PackageManagerControl).

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataControl.xaml
```diff
@@ -78,126 +78,160 @@
...
+        </Grid.RowDefinitions>
+        <Grid.ColumnDefinitions>
+          <ColumnDefinition
+        Width="Auto" />
+          <ColumnDefinition />
+        </Grid.ColumnDefinitions>
+
+        <!-- version -->
+        <TextBlock
+      Visibility="{Binding Path=PackageMetadata.Version,Converter={StaticResource NullToVisibilityConverter}}"
```
donnie-msft: I think some indentation got moved out of alignment.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/MultiSourcePackageMetadataProvider.cs
```diff
@@ -255,15 +256,33 @@ private async Task<IPackageSearchMetadata> GetPackageMetadataAsync(PackageIdenti
...
+            IPackageSearchMetadata packageSearchMetadataBase = completed.FirstOrDefault(m => !string.IsNullOrEmpty(m.Summary))
                 ?? completed.FirstOrDefault()
                 ?? PackageSearchMetadataBuilder.FromIdentity(identity).Build();
 
-            return master.WithVersions(
-                asyncValueFactory: () => MergeVersionsAsync(identity, completed));
+            var clonedResult = packageSearchMetadataBase
+                  .WithVersions(() => MergeVersionsAsync(identity, completed)) as PackageSearchMetadataBuilder.ClonedPackageSearchMetadata;
+
+            if (string.IsNullOrWhiteSpace(clonedResult.ReadmeFileUrl))
```
donnie-msft, line 280: clonedResult can be null

test/NuGet.Clients.Tests/NuGet.PackageManagement.UI.Test/ViewModels/ReadmePreviewViewModelTests.cs
```diff
@@ -33,50 +33,134 @@ public void Constructor_Defaults()
...
+            await target.SetPackageMetadataAsync(package, CancellationToken.None);
 
             //Assert
             Assert.False(target.ErrorLoadingReadme);
             Assert.Equal(readmeContents, target.ReadmeMarkdown);
-            Assert.True(target.CanDetermineReadmeDefined);
+        }
+
+        [Fact]
+        public async Task SetPackageMetadataAsync_WithLocalReadmeUrl_RenderLocalReadmeFalse_ReadmeReturned()
```
donnie-msft: Should it not return a readme or should it be returning the contents from the `stream` ?

```suggestion
        public async Task SetPackageMetadataAsync_WithLocalReadmeUrl_RenderLocalReadmeFalse_EmptyReadmeReturned()
```

test/NuGet.Clients.Tests/NuGet.PackageManagement.UI.Test/ViewModels/ReadmePreviewViewModelTests.cs
```diff
@@ -33,50 +33,134 @@ public void Constructor_Defaults()
...
+            //Assert
+            Assert.False(target.ErrorLoadingReadme);
+            Assert.Equal(string.Empty, target.ReadmeMarkdown);
+        }
+
+        [Theory]
+        [InlineData(ItemFilter.Installed)]
+        [InlineData(ItemFilter.Consolidate)]
+        [InlineData(ItemFilter.UpdatesAvailable)]
+        public async Task SetCurrentFilter_ItemFilterRenderingLocalReadme_ReadmeReturned(ItemFilter filter)
```
donnie-msft: 👍 Nice!

donnie-msft (approved):
Looking good, and thanks again for taking the time to split out the PRs. I'm not clear why this is titled "...local storage", so I think the title should be expanded. It's a feature branch PR so I don't want to be too picky on that, just so it's clear when reviewing commits later, this is much more than local storage.

I might not have a clear picture on how you're wanting the MVVM structure to be laid out. I think Details Pane is supposed to have its own VM still, so if that's not the case let me know. Also, a screenshot of how it's looking lately would be convenient. I didn't have time to build the feature branch.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageDetailsTabViewModel.cs
jgonz120: I do set it in the code though I don't know how often DetailControlModel changes, From the breakpoints doesn't seem like it does.

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
jgonz120:
```xml
<ContentControl Content="{Binding}">
  <ContentControl.Resources>
    <DataTemplate DataType="{x:Type nuget:DetailControlModel}">
      <nuget:PackageMetadataControl
        AutomationProperties.Name="{x:Static nuget:Resources.Accessibility_PackageMetadata}"
        DataContext="{Binding}"
        Margin="0,12,0,0"/>
    </DataTemplate>
    <DataTemplate DataType="{x:Type vm:ReadmePreviewViewModel}">
      <nuget:PackageReadmeControl
        DataContext="{Binding}"
        x:Name="_packageReadmeControl"
        AutomationProperties.Name="{x:Static nuget:Resources.Accessibility_PackageMetadata}"/>
    </DataTemplate>
  </ContentControl.Resources>
</ContentControl>
```

So from this code snippet we use the type of the VM to determine which view to render.

```xml
      <DataTemplate>
        <TextBlock Text="{Binding Header}" />
        <DataTemplate.Triggers>
          <DataTrigger Binding="{Binding Visible}" Value="False">
            <Setter Property="Visibility" Value="Collapsed" />
          </DataTrigger>
        </DataTemplate.Triggers>
      </DataTemplate>
```
In this snippet we use the `IsVisible` and `Header` elements in the TabViewModelBase class. I don't think an interface for the tab viewmodel makes sense since we would have to implement the notifyproperty change in all the implementations. 

Not sure if this works but I could add `ContentViewModel` to TabViewModelBase and maybe do this?
```xml
<ContentControl Content="{Binding ContentViewModel}">
  <ContentControl.Resources>
    <DataTemplate DataType="{x:Type nuget:DetailControlModel}">
      <nuget:PackageMetadataControl
        AutomationProperties.Name="{x:Static nuget:Resources.Accessibility_PackageMetadata}"
        DataContext="{Binding}"
        Margin="0,12,0,0"/>
    </DataTemplate>
    <DataTemplate DataType="{x:Type vm:ReadmePreviewViewModel}">
      <nuget:PackageReadmeControl
        DataContext="{Binding}"
        x:Name="_packageReadmeControl"
        AutomationProperties.Name="{x:Static nuget:Resources.Accessibility_PackageMetadata}"/>
    </DataTemplate>
  </ContentControl.Resources>
</ContentControl>
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
jgonz120: Is the suggestion to rename the RaisePropertyChanged function in ViewModelBase to OnPropertyChanged?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/MultiSourcePackageMetadataProvider.cs
jgonz120: WithVersions always returns a new ClonedPackageSearchMetadata instance. I can add the null check if though

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
jgonz120: Would this mean adding a reference to the PackageDetailsTabViewModel inside of the DetailControlModel? Then changes are all handled in that VM?

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
donnie-msft: Nope - the suggested change above is the only 1. It will then match all the other instances in this file.
Ideally, but a lot more changes required, just call RaisePropertyChanged directly in this file and get rid of OnPropertyChanged.

jgonz120: I'll do that in a separate PR since there are 49 references to OnPropertyChanged in this file.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataControl.xaml
jgonz120: They did, I'm going to do another PR to fix indentation after this merges. This makes the comparison a little easier to read rn because a whole section needs to be indented even though nothing really changed.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageMetadataControl.xaml
```diff
@@ -78,126 +78,160 @@
...
+          <RowDefinition />
+        </Grid.RowDefinitions>
+        <Grid.ColumnDefinitions>
+          <ColumnDefinition
+        Width="Auto" />
+          <ColumnDefinition />
+        </Grid.ColumnDefinitions>
+
+        <!-- version -->
+        <TextBlock
```
jebriede, line 164: We should fix indentation for newly added code.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageDetailsTabControl.xaml
```diff
@@ -0,0 +1,158 @@
...
+    </TabControl.ItemTemplate>
+    <TabControl.ContentTemplate>
+      <DataTemplate>
+        <!-- Customize the content of the tabs here -->
+        <ContentControl Content="{Binding}">
+          <ContentControl.Resources>
+            <DataTemplate DataType="{x:Type nuget:DetailControlModel}">
+              <nuget:PackageMetadataControl
+                AutomationProperties.Name="{x:Static nuget:Resources.Accessibility_PackageMetadata}"
+                DataContext="{Binding}"
```
jebriede: This shouldn't be necessary since the data context is already set.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageDetailsTabControl.xaml
```diff
@@ -0,0 +1,158 @@
...
+                Margin="0,12,0,0"/>
+            </DataTemplate>
+            <DataTemplate DataType="{x:Type vm:ReadmePreviewViewModel}">
+              <nuget:PackageReadmeControl
+                DataContext="{Binding}"
```
jebriede: This shouldn't be necessary since the data context is already set.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
```diff
@@ -460,6 +470,18 @@ private void InitializeFilterList(UserSettings settings)
             }
         }
 
+        private void InitializeSelectedPackageMetadataTab(UserSettings settings, bool isReadmeTabEnabled)
+        {
+            if (settings != null)
+            {
+                if (!isReadmeTabEnabled)
+                {
+                    settings.SelectedPackageMetadataTab = PackageMetadataTab.PackageDetails;
```
jebriede: I would recommend having the settings get set only in one place, which would probably be best when we are closing the PM UI.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageDetailsTabViewModel.cs
```diff
@@ -0,0 +1,121 @@
...
+                await ReadmePreviewViewModel.SetCurrentFilterAsync(filter);
+            }
+        }
+
+        private void ReadMePreviewViewModel_PropertyChanged(object sender, PropertyChangedEventArgs e)
+        {
+            if (ReadmePreviewViewModel is not null
+                && e.PropertyName == nameof(ReadmePreviewViewModel.ReadmeMarkdown))
+            {
+                Tabs[0].IsVisible = !string.IsNullOrWhiteSpace(ReadmePreviewViewModel.ReadmeMarkdown);
```
jebriede: The `ReadmePreviewViewModel` knows about its own `ReadmeMarkdown` property so it should be able to answer `IsVisible` on its own. In fact, I believe `IsVisible` could be read-only. It should not be set by its parent.

This will also remove one of the several places where we have to explicitly index the Tabs to know more details about it.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageDetailsTabViewModel.cs
```diff
@@ -0,0 +1,121 @@
...
+                if (!Tabs[0].IsVisible)
+                {
+                    SelectedTab = Tabs[1];
+                }
```
jebriede: To make this code more resilient to changes in the future where we may want to add a new tab and not have to modify this code, could we just say that we want the selected tab to be whatever the first visible tab is?

I also now wonder if this is going to conflict with the user settings. It would be nice to have the tab selection logic all in one place.

Disclaimer: I did not test this suggestion and wrote it quickly in the browser. Please verify correctness.

```suggestion
                foreach (var tab in Tabs) 
                {
                    if (tab.IsVisible)
                    {
                        SelectedTab = tab;
                        break;
                    }
                }
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
martinrrm: Sorry for the late reply, for example this https://docs.github.com/es/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-readmes, keeps README even in Spanish (there is one that is translated to Leame tho) but I'll ask Localization team what to do, we can do it later for all occurrences

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
jgonz120: So I discussed this a bunch with @jebriede  yesterday. I like the idea of using the VMs to determine which tabs are visible, however the DetailControlModel VM such a big VM that's handling so many different responsibilities that I think breaking it down to do this correctly would be so much work that I'm going to create a tech debt item to handle it in the future.

https://github.com/NuGet/Client.Engineering/issues/3054

jgonz120: I am going to rename the TabViewModelBase though so it doesn't imply the VM is a tab

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageDetailsTabViewModel.cs
jgonz120: I got rid of the indexing :D

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/RenderedViewModelBase.cs
```diff
@@ -0,0 +1,25 @@
+namespace NuGet.PackageManagement.UI.ViewModels
+{
+    public class RenderedViewModelBase : ViewModelBase
+    {
+        private string _header;
+        public string Header
```
jebriede:
```suggestion
        public string Title
```

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/RenderedViewModelBase.cs
```diff
@@ -0,0 +1,25 @@
+namespace NuGet.PackageManagement.UI.ViewModels
+{
+    public class RenderedViewModelBase : ViewModelBase
```
jebriede: I like the idea of the class name and its property describing the concept that this VM adds a title on top of the ViewModel base and it's a "page" that can be presented however we want: as a tab, as a section on a page, as a dialog, etc. etc. and this class becomes reusable.

```suggestion
    public class TitledPageViewModelBase : ViewModelBase
```

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageDetailsTabViewModel.cs
```diff
@@ -0,0 +1,106 @@
...
+            return SelectedTab is DetailControlModel ? PackageMetadataTab.PackageDetails : PackageMetadataTab.Readme;
+        }
+
+        public async Task InitializeAsync(DetailControlModel detailControlModel, INuGetPackageFileService nugetPackageFileService, ItemFilter currentFilter, PackageMetadataTab initialSelectedTab)
+        {
+            var nuGetFeatureFlagService = await ServiceLocator.GetComponentModelServiceAsync<INuGetFeatureFlagService>();
+            _readmeTabEnabled = await nuGetFeatureFlagService.IsFeatureEnabledAsync(NuGetFeatureFlagConstants.RenderReadmeInPMUI);
+
+            ReadmePreviewViewModel = new ReadmePreviewViewModel(nugetPackageFileService, currentFilter);
+            ReadmePreviewViewModel.IsVisible = _readmeTabEnabled;
```
jebriede: I like how IsVisible is now mostly determined by the `ReadmePreviewViewModel` object itself. This is the one remaining place where IsVisible is being set for the `ReadmePreviewViewModel`. Can we pass the state of _readmeTabEnabled to the ReadmePreviewViewModel instead and make IsVisible read-only?

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageDetailsTabViewModel.cs
```diff
@@ -0,0 +1,106 @@
...
+            DetailControlModel = detailControlModel;
+
+            Tabs.Add(ReadmePreviewViewModel);
+            Tabs.Add(DetailControlModel);
+
+            SelectedTab = _readmeTabEnabled && initialSelectedTab == PackageMetadataTab.Readme
+                ? ReadmePreviewViewModel
+                : DetailControlModel;
```
jebriede: I left this comment on a previous iteration but didn't submit my review before this latest commit so I'll leave it here:

This works fine for a 2-tab system. If we want to make sure this is resilient for future changes that has more than 2 tabs, we can iterate over the list of tabs to find the first one that's visible and set it to the SelectedTab, taking into account the `initialSelectedTab` as well.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageDetailsTabViewModel.cs
```diff
@@ -0,0 +1,106 @@
...
+        {
+            if (ReadmePreviewViewModel is not null
+                && e.PropertyName == nameof(ReadmePreviewViewModel.IsVisible))
+            {
+                if (!ReadmePreviewViewModel.IsVisible)
+                {
+                    SelectedTab = DetailControlModel;
+                }
+            }
+        }
```
jebriede: This could be rewritten to more generally handle when a tab goes invisible and make the next one in line, or first one from the start of the collection (depending on how we want it to behave), that has an IsVisible = true, the selected tab. That way we avoid some code that specifically deals with if ReadmePreviewViewModel is invisible, then select the DetailControlModel, and instead, we say if the selected tab goes invisible, select the next one. We will also need to hook this same event handler up to all the Tabs in the collection.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageDetailsTabViewModel.cs
```diff
@@ -0,0 +1,106 @@
...
+        private void DetailControlModel_PropertyChanged(object sender, PropertyChangedEventArgs e)
+        {
+            ThreadHelper.JoinableTaskFactory.Run(async () =>
+            {
+                if (_readmeTabEnabled)
+                {
+                    await ReadmePreviewViewModel.SetPackageMetadataAsync(DetailControlModel.PackageMetadata, CancellationToken.None);
+                }
+            });
+        }
```
jebriede, line 113: Another way of writing this would be to have the PackageDetailsTabViewModel let any listeners know when the PackageMetadata changed so the listener (in this case the ReadmePreviewViewModel) can update its own PackageMetadata. This removes the responsibility of this ViewModel updating the ReadmePreviewViewModel directly.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageDetailsTabViewModel.cs
```diff
@@ -0,0 +1,106 @@
...
+        public PackageDetailsTabViewModel()
+        {
+            _readmeTabEnabled = true;
+            Tabs = new ObservableCollection<RenderedViewModelBase>();
+        }
+
+        public PackageMetadataTab GetSelectedTab()
+        {
+            return SelectedTab is DetailControlModel ? PackageMetadataTab.PackageDetails : PackageMetadataTab.Readme;
+        }
```
jebriede: If we can separate the conversion from the PackageMetadataTab to the ViewModel type and back, we can use that converter when reading the selected tab and setting the selected tab. Maybe a converter class or static methods that do the conversions? That way, we can also make this code more general and rely on the conversion logic being testable and in one place.

jebriede (commented):
The latest iteration eliminated a lot of complexity with the tab switching. Great job! I've left some comments, mostly around making the code being more resilient to changes (more maintainable) and making the tab code a little more generalized. Let me know if you want to discuss!

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageDetailsTabViewModel.cs
jgonz120: Okay updated it! I check to see if the currently selected tab is visible and if it isn't then switch to the first available visible tab.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageDetailsTabViewModel.cs
jgonz120: Okay switched it a bit, lemme know what you think

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageDetailsTabViewModel.cs
```diff
@@ -0,0 +1,113 @@
...
+        private void DetailControlModel_PropertyChanged(object sender, PropertyChangedEventArgs e)
+        {
+            ThreadHelper.JoinableTaskFactory.Run(async () =>
+            {
+                if (_readmeTabEnabled)
+                {
+                    await ReadmePreviewViewModel.SetPackageMetadataAsync(DetailControlModel.PackageMetadata, CancellationToken.None);
+                }
+            });
+            IsVisible_PropertyChanged(sender, e);
```
jebriede: Visibility should be determined by the DetailControl who should invoke a property change notification if the visibility changes and any listeners do what they need to do. It's not ideal to invoke the property changed handler without the DetailControl actually firing the event for the property that changed.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageDetailsTabViewModel.cs
```diff
@@ -0,0 +1,113 @@
...
+            DetailControlModel.PropertyChanged -= DetailControlModel_PropertyChanged;
+        }
+
+        private void IsVisible_PropertyChanged(object sender, PropertyChangedEventArgs e)
+        {
+            if (e.PropertyName == nameof(TitledPageViewModelBase.IsVisible))
+            {
+                if (!SelectedTab.IsVisible)
+                {
+                    SelectedTab = Tabs.FirstOrDefault(t => t.IsVisible);
```
jebriede, line 99: This could result in null being set as the SelectedTab if there is no visible tab in the collection. If that were to occur, line 94 would throw an NRE the next time this method is invoked. We should handle this case explicitly.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageDetailsTabViewModel.cs
```diff
@@ -0,0 +1,113 @@
...
+            ReadmePreviewViewModel = new ReadmePreviewViewModel(nugetPackageFileService, currentFilter, _readmeTabEnabled);
+            DetailControlModel = detailControlModel;
+
+            Tabs.Add(ReadmePreviewViewModel);
+            Tabs.Add(DetailControlModel);
+
+            SelectedTab = Tabs.FirstOrDefault(t => t.IsVisible && ConvertFromVm(t) == initialSelectedTab) ?? Tabs.FirstOrDefault(t => t.IsVisible);
+
+            DetailControlModel.PropertyChanged += DetailControlModel_PropertyChanged;
+            ReadmePreviewViewModel.PropertyChanged += IsVisible_PropertyChanged;
```
jebriede: If we just hook up all tabs to the property changed event handler method that processes visibility, we won't have to change this code in the future if we add more tabs or change the logic so the other tabs become invisible, etc.

```suggestion
            foreach (var tab in Tabs)
            {
                tab.PropertyChanged += TabIsVisible_PropertyChanged;
            }
```

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageDetailsTabViewModel.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        private void IsVisible_PropertyChanged(object sender, PropertyChangedEventArgs e)
+        {
+            if (e.PropertyName == nameof(TitledPageViewModelBase.IsVisible))
+            {
+                if (!SelectedTab.IsVisible)
+                {
+                    SelectedTab = Tabs.FirstOrDefault(t => t.IsVisible);
+                }
+            }
```
jebriede: We may want to make sure the tab that's changing is the selected tab explicitly, otherwise we don't have to switch the tab. Something like this?

```suggestion
            if (e.PropertyName == nameof(TitledPageViewModelBase.IsVisible)
                && SelectedTab == sender
                && !SelectedTab.IsVisible)
            {
                SelectedTab = Tabs.FirstOrDefault(t => t.IsVisible);
            }
```

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageDetailsTabViewModel.cs
```diff
@@ -0,0 +1,113 @@
...
+using Microsoft.VisualStudio.Shell;
+using NuGet.VisualStudio;
+using NuGet.VisualStudio.Internal.Contracts;
+
+namespace NuGet.PackageManagement.UI.ViewModels
+{
+    public class PackageDetailsTabViewModel : ViewModelBase, IDisposable
+    {
+        private bool _disposed = false;
+
```
jebriede: Extra blank line.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageDetailsTabViewModel.cs
```diff
@@ -0,0 +1,113 @@
...
+            Tabs.Add(ReadmePreviewViewModel);
+            Tabs.Add(DetailControlModel);
+
+            SelectedTab = Tabs.FirstOrDefault(t => t.IsVisible && ConvertFromVm(t) == initialSelectedTab) ?? Tabs.FirstOrDefault(t => t.IsVisible);
+
+            DetailControlModel.PropertyChanged += DetailControlModel_PropertyChanged;
+            ReadmePreviewViewModel.PropertyChanged += IsVisible_PropertyChanged;
+        }
+
+        private static PackageMetadataTab ConvertFromVm(TitledPageViewModelBase vm)
```
jebriede:
```suggestion
        private static PackageMetadataTab ConvertFromTabType(TitledPageViewModelBase vm)
```

jebriede (approved):
Approving with suggestions

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageDetailsTabViewModel.cs
jebriede: If we do this, we need to unregister them as well, as not all these tab VMs are created by this class.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageDetailsTabViewModel.cs
```diff
@@ -85,13 +83,23 @@ public void Dispose()
...
         {
             if (e.PropertyName == nameof(TitledPageViewModelBase.IsVisible))
             {
-                if (!SelectedTab.IsVisible)
+                if (SelectedTab == null)
+                {
+                    SelectedTab = Tabs.FirstOrDefault(t => t.IsVisible);
+                }
+                else if (SelectedTab == sender
+                    && !SelectedTab.IsVisible)
```
jebriede: These could be combined, especially since they call the same exact code if true.

```suggestion
                if (SelectedTab == null || (SelectedTab == sender && !SelectedTab.IsVisible))
```

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/TabViewModelBase.cs
jebriede: Boolean properties are typically prefixed with Can, Is, Has, etc. by convention and according to Microsoft naming recommendations: https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members?redirectedfrom=MSDN#names-of-properties

jebriede (approved):

PR #6096, opened 2024-10-10
Split up functional test execution in CI

jeffkl:
# Bug

Fixes: 

## Description
This change splits the execution of functional tests into individual jobs per test assembly.  I also moved a few tests from "functional" to "unit" because they execute in under 2 minutes.  This should make the overall execution time go down a little.  Since I moved some tests to unit, there are no more non-Windows functional tests on .NET Core 3.1 so in the end our CI only uses one extra machine.

I audited the functional tests and made sure they were all represented:

|                                     | Functional net472 | Functional net8.0 | Unit net472 | Unit net8.0 |
| -| :--: | :--: | :--: | :--: |
| Dotnet.Integration.Test             |        | ✅      |        |        |
| Msbuild.Integration.Test            | ✅      |        |        |        |
| NuGet.CommandLine.FuncTest          | ✅      |        |        |        |
| NuGet.CommandLine.Test              | ✅      |        |        |        |
| NuGet.Commands.FuncTest             |        |        | ✅      | ✅      |
| NuGet.MSSigning.Extensions.FuncTest |        |        | ✅      |        |
| NuGet.Packaging.FuncTest            | ✅      | ✅      |        |        |
| NuGet.Protocol.FuncTest             |        |        | ✅      | ✅      |
| NuGet.Signing.CrossFramework.Test   |        |        | ✅      |        |
| NuGet.XPlat.FuncTest                |        |        |        | ✅      |

I also adjusted some timeout so that hung tests hopefully don't run longer than they need to.  
## PR Checklist

- [ ] Added tests

### DIFF
--- a/eng/pipelines/pr.yml
+++ b/eng/pipelines/pr.yml
@@ -153,6 +153,7 @@ stages:
         testType: Functional
         testTargetFramework: net472
         testProjectName: NuGet.CommandLine.Test
+        testVerbosity: normal
         timeoutInMinutes: 45
 
 - ${{ if eq(parameters.RunFunctionalTestsOnWindows, true) }}:
--- a/eng/pipelines/pr.yml
+++ b/eng/pipelines/pr.yml
@@ -102,7 +102,7 @@ stages:
 
 - ${{ if eq(parameters.RunFunctionalTestsOnWindows, true) }}:
   - stage:
-    displayName: 
+    displayName: Msbuild.Integration.Test on Windows (.NET Framework 4.7.2)
     dependsOn: []
     jobs:
     - template: pr.job.yml
--- a/eng/pipelines/pr.yml
+++ b/eng/pipelines/pr.yml
@@ -102,12 +102,12 @@ stages:
 
 - ${{ if eq(parameters.RunFunctionalTestsOnWindows, true) }}:
   - stage:
-    displayName: Functional Tests on Windows (.NET Framework 4.7.2)
+    displayName: Functional Tests on Windows (.NET Framework 4.7.2) Msbuild.Integration.Test
     dependsOn: []
     jobs:
     - template: pr.job.yml
       parameters:
-        displayName: Functional Tests on Windows (.NET Framework 4.7.2)
+        displayName: Functional Tests on Windows (.NET Framework 4.7.2) Msbuild.Integration.Test
         osName: Windows
         ${{ if eq(variables['System.TeamProject'], 'public') }}:
           agentPool: NetCore-Public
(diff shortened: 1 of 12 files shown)


eng/pipelines/pr.yml
```diff
@@ -116,16 +116,72 @@ stages:
...
+        testProjectName: NuGet.CommandLine.FuncTest
+
+- ${{ if eq(parameters.RunFunctionalTestsOnWindows, true) }}:
+  - stage:
+    displayName: Functional Tests on Windows (.NET Framework 4.7.2) NuGet.CommandLine.Test
+    dependsOn: []
+    jobs:
+    - template: pr.job.yml
+      parameters:
+        displayName: Functional Tests on Windows (.NET Framework 4.7.2) NuGet.CommandLine.Test
```
donnie-msft: The stage names are kind of lengthy already. Can the job name be shortened so it's easier to see the job title in the AzDo interface?

```suggestion
        displayName: NuGet.CommandLine.Test
```

jebriede:
> I also moved a few tests from "functional" to "unit" because they execute in under 2 minutes.

The classification of a unit vs functional test is not dependent on how long they run, but rather on what it is testing (https://www.softwaretestingstuff.com/functional-testing-vs-unit-testing). Are the tests we moved unit tests or are they functional tests that take under 2 minutes to run?

eng/pipelines/pr.yml
```diff
@@ -116,16 +116,72 @@ stages:
...
+        ${{ if eq(variables['System.TeamProject'], 'public') }}:
+          agentPool: NetCore-Public
+          agentDemands: ImageOverride -equals Windows.VS2022Preview.Amd64.Open
+        ${{ else }}:
+          agentPool: VSEngSS-MicroBuild2022-1ES
+        testType: Functional
+        testTargetFramework: net472
+        testProjectName: NuGet.CommandLine.Test
+        testVerbosity: normal
+        timeoutInMinutes: 45
```
donnie-msft, line 157: There's been 30 and 60 minute timeouts. Is the thinking here that 60 split 2 ways is 30 plus cushion?
Or is it a measured value?

jeffkl:
> The classification of a unit vs functional test is not dependent on how long they run, but rather on what it is testing

Our CI runs "unit" tests on one machine and "functional" tests on another.  This change splits up the functional test assemblies to run on a single machine.  Running a mixture of unit and functional tests at the same time on the same machine doesn't change their purpose, just how we run them.  I didn't rename the tests or anything, just categorized them so that they run during one phase of the build.  I could update the classification of unit vs functional test runs but that seems like more work.

donnie-msft (dismissed):

eng/pipelines/pr.yml
jeffkl: After splitting this individual test assembly out, I ran it a few times and it always finished in 30-40 minutes.  My aim is to make the timeouts just above average so that it times out sooner rather than later.

jebriede (dismissed):
Maybe we can fix the functional vs unit names later since it's being used not to group the functional vs unit tests, but instead to run tests on one machine or another. But it's out of scope for this PR.

jeffkl:
> Maybe we can fix the functional vs unit names later since it's being used not to group the functional vs unit tests, but instead to run tests on one machine or another. But it's out of scope for this PR.

👍 Excellent point though, we have some tests named as "functional" tests grouped together to run during the "unit" test run, and visa versa.  Perhaps how we label the runs just needs an update.

eng/pipelines/pr.yml
jeffkl: How about `NuGet.CommandLine.Test on Windows (.NET Framework 4.7.2)` ?  Some tests run on multiple platforms and frameworks so I think those details need to be part of the name because that's how test results are labeled.

jeffkl: Updated!

eng/pipelines/pr.yml
```diff
@@ -134,25 +190,26 @@ stages:
...
+        displayName: NuGet.Packaging.FuncTest on Windows (.NET 8.0)
         osName: Windows
         ${{ if eq(variables['System.TeamProject'], 'public') }}:
           agentPool: NetCore-Public
           agentDemands: ImageOverride -equals Windows.VS2022Preview.Amd64.Open
         ${{ else }}:
           agentPool: VSEngSS-MicroBuild2022-1ES
         testType: Functional
-        testTargetFramework: netcoreapp3.1
-        timeoutInMinutes: 60
```
donnie-msft, line 155: Just making sure this drop of the timeout is intentional. I'm guessing because net8 runs much faster.

donnie-msft (approved):

eng/pipelines/pr.yml
jeffkl: That is correct, it always finishes in under 15 minutes which is the default timeout

PR #6098, opened 2024-10-12
Modify ContentItemCollection.PopulateItemGroups to use pooling for highly allocated temporary data structures

### LINKED ISSUE #13851: ContentItemCollection.PopulateItemGroups unnecessarily allocates
### NuGet Product Used

Other/NA

### Product Version

VS17

### Worked before?

N/A

### Impact

### Repro Steps & Context

1) Start VS
2) Start profiler
3) Open vs-platform editor sln and wait for busy indicator to complete (ie, restore has completed)
4) Stop profiler

About 7.5% of allocations are under ContentItemCollection.PopulateItemGroups. About 1.1% appear to be easily removed by using pools.

### Verbose Logs

_No response_

ToddGrun:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13851

## Description
Reduce allocations in ContentItemCollection.PopulateItemGroups by using pooling

## PR Checklist

- [ ] Added tests

### DIFF
--- a/src/NuGet.Core/NuGet.Packaging/ContentModel/ContentItemCollection.cs
+++ b/src/NuGet.Core/NuGet.Packaging/ContentModel/ContentItemCollection.cs
@@ -20,6 +20,10 @@ namespace NuGet.ContentModel
 
         private List<Asset>? _assets;
         private ConcurrentDictionary<ReadOnlyMemory<char>, string?>? _assemblyRelatedExtensions;
+
+        private readonly SimplePool<List<Asset>> _listAssetPool = new(() => new List<Asset>());
+        private readonly SimplePool<Dictionary<ContentItem, List<Asset>>> _groupAssetsPool = new(() => new(GroupComparer.DefaultComparer));
+
         /// <summary>
         /// True if lib/contract exists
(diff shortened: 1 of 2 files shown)


ToddGrun:
Highlighted allocations the ones that are removed due to this change. Took two profiles of the before to verify it wasn't just a bad run:

image

image

ToddGrun:
A couple thoughts:

1) If this codepath is going to soon be removed, let's just abandon this. Not executing code is the best optimization possible.
2) If the lock usage in SimplePool is a concern, it might be good to consider using something like [Roslyn's ObjectPool](https://github.com/dotnet/roslyn/blob/main/src/Dependencies/PooledObjects/ObjectPool%601.cs).
3) If the possible increase in total memory footprint due to pooling is a concern (it hasn't been flagged as such in other VS codebases I've played in), then an option would be to clear out the pools once the work needing to use these objects is completed. Not a path I would like to go down, but thought I would call it out when listing potential concerns I could think of with this PR.

jeffkl (approved):

nkolev92 (approved):

PR #3596, opened 2020-08-20
Remove SHA-1 from cryptography use

heng-liu:
## Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/468
Regression: No  
* Last working version:   
* How are we preventing it in future:   

## Fix

Details: 
1.Removed GetSha1HashProvider, so no SHA-1 instance is created anymore
2.Validated only the hash byte array length for EssCertId, no SHA-1 hash value checking anymore
3.Changed tests accordingly
4.Removed CA5350 from GlobalSuppressions.cs
5.Removed APIs from PublicAPI.Shipped.txt : GetSha1HashProvider is removed from 3 TFMs, SignatureUtility is removed from netstandard2.0 
6.Added two tests: mismatched SHA-1 hash with same length in EssCertId , shorter hash in EssCertId
## Testing/Validation

Tests Added: Yes
Reason for not adding tests:  
Validation:

PR #6103, opened 2024-10-16
Test:  remove SHA-1 verification from test OCSP responder

dtivel:
# Bug

Fixes:  https://github.com/NuGet/Client.Engineering/issues/2996
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2221932

## Description

`CertificateAuthority` is test infrastructure that enables us to test a variety of scenarios without external dependencies.  For example, we can test revocation checking (via OCSP) using `CertificateAuthority` because it sufficiently implements an OCSP responder.  This is important for package signing test coverage and robustness.

The OCSP standard [requires](https://www.rfc-editor.org/rfc/rfc6960#section-4.1.1) that an OCSP request includes a `CertID` structure:

```
   CertID          ::=     SEQUENCE {
       hashAlgorithm       AlgorithmIdentifier,
       issuerNameHash      OCTET STRING, -- Hash of issuer's DN
       issuerKeyHash       OCTET STRING, -- Hash of issuer's public key
       serialNumber        CertificateSerialNumber }
```

`CertID` specifies a hash algorithm (e.g.:  SHA-1, SHA-256, etc.) and hashes of the issuer distinguished name and public key for the certificate for which revocation status is being checked.  These hashes are used to verify that the to-be-verified certificate is one that the responding certificate authority actually issued.  Per the OCSP specification, if a CA receives an OCSP request for a certificate which it did not issue, the OCSP responder should respond with an `Unknown` status.

Ideally, a modern timestamp authority (TSA) would support all SHA-2 algorithms, and a modern OCSP client would use SHA-2 algorithms.

We can ban SHA-1 in our test TSA; however, we can't ban SHA-1 in OCSP clients.  We don't control that.  We have tests that chain build and check revocation status and it's _the operating system_ --- in this case, Windows --- that chooses SHA-1 to identify a certificate in the OCSP request.

From https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-ocsp/22de9ccd-4347-460d-a4cc-3ea1aefa1284:

> The CertId field always uses the SHA-1 hash algorithm.

Windows follows [RFC 5019, section 2.1.1](https://www.rfc-editor.org/rfc/rfc5019.html#section-2.1.1), which says:

>    Clients MUST use SHA1 as the hashing algorithm for the
   CertID.issuerNameHash and the CertID.issuerKeyHash values.

This change:

* Gives SHA-1 `CertID`'s a pass.  Don't verify hashes; just assume they're correct.  Only verify certificate serial number.
* Improves the randomness of serial numbers.
* Implements support for SHA-2 algorithms.
* Returns an `Unknown` status for any other hash algorithm.

In product code, we have done something similar to the first point above.  See [Remove SHA-1 from cryptography use (#3596) · NuGet/NuGet.Client@3501dde (github.com)](https://github.com/NuGet/NuGet.Client/commit/3501ddedc274ac10d4b135856b7593a6bb8a72f1#diff-cdc41aa3367ec3539fb21298d72068190e4cbb92d560b53da47d5bfbc90543c3L584-R584)

## PR Checklist

- [ ] ~~Added tests~~ N/A
- [ ] ~~Link to an issue or pull request to update docs if this PR changes settings, environment variables, new feature, etc.~~ N/A

CC @kartheekp-ms

### DIFF
--- a/test/TestUtilities/Microsoft.Internal.NuGet.Testing.SignedPackages/CertificateAuthority.cs
+++ b/test/TestUtilities/Microsoft.Internal.NuGet.Testing.SignedPackages/CertificateAuthority.cs
@@ -21,8 +21,9 @@ namespace Microsoft.Internal.NuGet.Testing.SignedPackages
         private readonly Dictionary<string, X509Certificate2> _issuedCertificates;
         private readonly Dictionary<string, RevokedInfo> _revokedCertificates;
         private readonly Lazy<OcspResponder> _ocspResponder;
-        private BigInteger _nextSerialNumber;
+        private readonly HashSet<BigInteger> _assignedSerialNumbers;
         private byte[] _dnHash;
+        private byte[] _keyHash;
 
         /// <summary>
         /// This base URI is shared amongst all HTTP responders hosted by the same web host instance.
@@ -57,7 +58,7 @@ namespace Microsoft.Internal.NuGet.Testing.SignedPackages
             CertificateUri = new Uri(Url, $"{fingerprint}.cer");
             OcspResponderUri = GenerateRandomUri();
             Parent = parentCa;
-            _nextSerialNumber = new BigInteger(2);
+            _assignedSerialNumbers = new HashSet<BigInteger>();
             // The key is the certificate serial number in hexadecimal; therefore, lookups should be case insensitive.
             _issuedCertificates = new Dictionary<string, X509Certificate2>(StringComparer.InvariantCultureIgnoreCase);
             _revokedCertificates = new Dictionary<string, RevokedInfo>(StringComparer.InvariantCultureIgnoreCase);
@@ -181,10 +182,12 @@ namespace Microsoft.Internal.NuGet.Testing.SignedPackages
                             critical: true));
                 };
 
+            byte[] serialNumber = CertificateUtilities.GenerateSerialNumber();
+
             X509Certificate2 certificate = CreateCertificate(
                 options.KeyPair,
                 options.KeyPair,
-                BigInteger.One,
+                serialNumber,
                 options.SubjectName,
                 options.SubjectName,
                 options.NotBefore,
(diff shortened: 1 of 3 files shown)


test/TestUtilities/Test.Utility/Signing/CertificateAuthority.cs
```diff
@@ -195,30 +198,53 @@ internal CertStatus GetCertStatus(CertId certId, out X509Certificate2 certificat
         {
             certificate = null;
 
-            if (certId.AlgorithmIdentifier.Algorithm.Value != Oids.Sha1)
+            Oid hashAlgorithmOid = certId.AlgorithmIdentifier.Algorithm;
+
+            if (hashAlgorithmOid.Value == Oids.Sha1)
             {
-                return CertStatus.FromUnknown();
+                // SHA-1 is obsolete.  We'll rely on serial number only for matching.
```
kartheekp-ms: Should we return `CertStatus.FromUnknown();` in this case?

dtivel: Nope.  As I mentioned in the description of this PR:

> We can ban SHA-1 in our test TSA; however, we can't ban SHA-1 in OCSP clients. We don't control that. We have tests that chain build and check revocation status and it's the operating system --- in this case, Windows --- that chooses SHA-1 to identify a certificate in the OCSP request.

If we returned `CertStatus.FromUnknown()` here, OCSP revocation checking would generally fail on Windows.

test/TestUtilities/Microsoft.Internal.NuGet.Testing.SignedPackages/CertificateUtilities.cs
```diff
@@ -47,5 +51,38 @@ public static X509Certificate2 GetCertificateWithPrivateKey(X509Certificate2 cer
...
+            byte[] bytes = new byte[20];
+
+            using (RandomNumberGenerator rng = RandomNumberGenerator.Create())
+            {
+                while (true)
+                {
+                    rng.GetBytes(bytes);
+
+                    // Treat the byte array as a little-endian integer.
+                    // Ensure the number is non-negative by setting the highest bit of the first byte to 0.
```
donnie-msft: Isn't this the last byte, evaluating to `bytes[19]` ?
```suggestion
                    // Ensure the number is non-negative by setting the highest bit of the last byte to 0.
```

donnie-msft (dismissed):

test/TestUtilities/Microsoft.Internal.NuGet.Testing.SignedPackages/CertificateUtilities.cs
dtivel: It's the first byte in little endian.  Updated the comment.

donnie-msft (dismissed):

test/TestUtilities/Microsoft.Internal.NuGet.Testing.SignedPackages/CertificateUtilities.cs
```diff
@@ -47,5 +51,38 @@ public static X509Certificate2 GetCertificateWithPrivateKey(X509Certificate2 cer
...
+
+            using (RandomNumberGenerator rng = RandomNumberGenerator.Create())
+            {
+                while (true)
+                {
+                    rng.GetBytes(bytes);
+
+                    // Treat the byte array as a little-endian integer.
+                    // Ensure the number is non-negative by setting the highest bit of the first byte (little-endian) to 0.
+                    bytes[bytes.Length - 1] &= 0x7F;
```
zivkan, line 82: There are customers who run the .NET SDK on big endian machines and have reported bugs to NuGet when we didn't work on those machines, although I think Microsoft's official stance is that it's "best effort" and "community driven", rather than officially supported.

Still, if it's not too much effort, could you write this in a way that will work on both?  [BitConverter.IsLittleEndian](https://learn.microsoft.com/en-us/dotnet/api/system.bitconverter.islittleendian?view=net-8.0) exists to check at runtime.

dtivel: It does not matter.  It should be the same on big- or little-endian architectures.

1.  Fill a byte array with random bytes.
2.  Treat the byte array as little endian because `BigInteger`'s constructor takes a little-endian byte array.
3.  Use `BigInteger` to ensure uniqueness.
4.  Reverse the array to make it big endian.
5.  Return the big-endian byte array because `CertificateRequest.Create(...)` requires a number in big endian.

zivkan: Alternatively, get an int32 from the random number generator, do your bit banging (`&= ~(1 << 31)`) then let [`IPAddress.HostToNetworkOrder(Int32)`](https://learn.microsoft.com/en-us/dotnet/api/system.net.ipaddress.hosttonetworkorder?view=net-8.0#system-net-ipaddress-hosttonetworkorder(system-int32)) do the conversion when necessary, and then convert to bytes with `BitConverter.GetBytes(Int32)`

zivkan: Actually, there's a much easier way. Use `RandomNumberGenerator.GetInt32` and pass `Int32.MaxValue` as the upper range. Then you don't need to do any bit banging yourself.

zivkan: sorry, it's too early in the morning. I only just noticed you're using a 20 byte array, and `BigInteger`.  My original suggestion to use `BitConverter.IsLittleEndian` is probably the best one. You'll need to use it twice, once for bit banging here, once to check when you need to reverse the array, below.

test/TestUtilities/Microsoft.Internal.NuGet.Testing.SignedPackages/CertificateAuthority.cs
```diff
@@ -198,30 +201,53 @@ internal CertStatus GetCertStatus(CertId certId, out X509Certificate2 certificat
...
+            else
             {
                 return CertStatus.FromUnknown();
             }
 
-            string serialNumber = BitConverter.ToString(certId.SerialNumber.ToByteArray()).Replace("-", "");
+            byte[] serialNumberBytes = certId.SerialNumber.ToByteArray();
+
+            // Convert to big-endian.
+            Array.Reverse(serialNumberBytes);
```
zivkan, line 239: Here's another place it would be good to make our code big endian safe, via `BitConverter.IsLittleEndian`

zivkan: From my other comment, BigInteger always uses little endian

zivkan (approved):

jeffkl (approved):

PR #6105, opened 2024-10-17
Do not run tests in isolation

jeffkl:
# Bug

Fixes: 

## Description
When you specify environment variables in a RunSettings file, I noticed this in the [docs](https://learn.microsoft.com/visualstudio/test/configure-unit-tests-by-using-a-dot-runsettings-file?view=vs-2022#example):

> Because these environment variables should always be set when the test host is started, the tests should always run in a separate process. For this, the /InIsolation flag will be set when there are environment variables so that the test host is always invoked.

So this change sets the environment variable in the pipeline YAML instead which could improve our test run times.  Supposedly this can make tests run a little slower but I wonder if its also contributing to test runs hanging/timing out.  This mainly just removes a layer of complexity.

## PR Checklist

- [ ] Added tests


donnie-msft (approved):
Worth trying!

PR #6113, opened 2024-10-22
Execute net tools plugins

Nigusu-Allehu:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13741

## Description
This PR makes sure plugins discovered in https://github.com/NuGet/NuGet.Client/pull/5990 are executed. The plugin files are directly executed by starting a new process.

#### Unit test
The test case I wrote has a simple batch file with the following content
```bat
@echo off
echo File executed > ""{outputPath}""
```
It is passed as a plugin file to the PluginFactory. The test looks into the contents of `outputPath` to check if the plugin has been executed or not. The assert contain statement checks if `outputPath` contains ` File executed` which should be written only if the plugin file has been executed.

#### nuget-plugin-bat.bat 
If you would like to manually test the plugin I used for testing, here it is as a file. : https://github.com/Nigusu-Allehu/NuGetBatAuthPlugin/blob/main/nuget-plugin-bat.bat (Edit the file and change the value of `debugLogPath` if you would like to see what the plugin has been receiving and sending)
set env variables as follows
- NUGET_NETFX_PLUGIN_PATHS : null
- NUGET_NETCORE_PLUGIN_PATHS : null
- NUGET_PLUGIN_PATHS : path to the nuget-plugin-bat.bat file
Add a break point in `PluginManager.cs` at the line `var utilities = await PerformOneTimePluginInitializationAsync(plugin, cancellationToken);` When the execution stops at this break point take a look into the values of the plugin variable. It should have a value `plugin.Connection.State = Success`
## PR Checklist

- [x] Added tests
https://github.com/NuGet/Home/issues/13858


src/NuGet.Core/NuGet.Protocol/Plugins/PluginFactory.cs
```diff
@@ -455,5 +479,45 @@ private static void WriteCommonLogMessages(IPluginLogger logger)
             logger.Write(new ProcessLogMessage(logger.Now));
             logger.Write(new ThreadPoolLogMessage(logger.Now));
         }
+
+        public async Task<IPlugin> GetOrCreateNetToolsPluginAsync(string filePath, IEnumerable<string> arguments, IRequestHandlers requestHandlers, ConnectionOptions options, CancellationToken sessionCancellationToken)
```
nkolev92: We discussed this briefly. 

The name here talks about a net tools plugin, but it's really about an executable file, which doesn't really need to a dotnet tool, it can just be a .bat file.

We can make the names match that.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginManager.cs
```diff
@@ -212,12 +212,26 @@ private async Task<Tuple<bool, PluginCreationResult>> TryCreatePluginAsync(
...
+                        IPlugin plugin;
+
+                        if (result.PluginFile.IsDotnetToolsPlugin)
+                        {
+                            plugin = await _pluginFactory.GetOrCreateNetToolsPluginAsync(
+                            result.PluginFile.Path,
+                            PluginConstants.PluginArguments,
+                            new RequestHandlers(),
+                            _connectionOptions,
+                            cancellationToken);
```
donnie-msft: nit: please indent 4 spaces and add named parameters

src/NuGet.Core/NuGet.Protocol/Plugins/PluginFactory.cs
```diff
@@ -149,36 +149,60 @@ private async Task<IPlugin> CreatePluginAsync(
...
+                    RedirectStandardOutput = true,
+                    StandardOutputEncoding = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false),
+                    WindowStyle = ProcessWindowStyle.Hidden,
+                };
 #endif
+            }
+            else
+            {
+                // A dotnet tools plugin. Execute it directly.
+
```
donnie-msft: nit: blank line

test/NuGet.Core.Tests/NuGet.Protocol.Tests/Plugins/PluginFactoryTests.cs
```diff
@@ -135,6 +137,119 @@ public async Task GetOrCreateAsync_ThrowsIfDisposed()
             Assert.Equal(nameof(PluginFactory), exception.ObjectName);
         }
 
+        [PlatformFact(Platform.Windows)]
+        public async Task GetOrCreateNetPluginAsync_CreatesPluginAndSendsHandshakeRequest()
```
donnie-msft: nit:
```suggestion
        public async Task GetOrCreateNetPluginAsync_UsingBatchFile_CreatesPluginAndSendsHandshakeRequest()
```

donnie-msft (commented):
Just nits for now. I'll review more later today

jeffkl (approved):

donnie-msft (approved):
Just nits!

src/NuGet.Core/NuGet.Protocol/Plugins/IPluginFactory.cs
```diff
@@ -42,5 +42,36 @@ Task<IPlugin> GetOrCreateAsync(
...
+        /// The task result (<see cref="Task{TResult}.Result" />) returns a <see cref="Plugin" />
+        /// instance.</returns>
+        /// <exception cref="ArgumentException">Thrown if <paramref name="filePath" />
+        /// is either <see langword="null" /> or empty.</exception>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="arguments" />
+        /// is <see langword="null" />.</exception>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="requestHandlers" />
+        /// is <see langword="null" />.</exception>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="options" />
+        /// is <see langword="null" />.</exception>
```
zivkan: You could list the arguments in a comma separated list, rather than duplicating the entire exception element multiple times.

Also, in VS, when you hover your mouse over the method name, you can see the XMLDoc rendered. I just tried and noticed for the first time that what we write within the exception tag does not get rendered. This makes exception descriptions much less valuable then summary or remarks, since this exception doc, since when you look at the intellisense info at the place where you call the API, it won't show you this text.

src/NuGet.Core/NuGet.Protocol/Plugins/IPluginFactory.cs
```diff
@@ -42,5 +42,36 @@ Task<IPlugin> GetOrCreateAsync(
...
+        /// <exception cref="OperationCanceledException">Thrown if <paramref name="sessionCancellationToken" />
+        /// is cancelled.</exception>
+        /// <exception cref="ObjectDisposedException">Thrown if this object is disposed.</exception>
+        /// <remarks>This method is primarily used for directly executable plugin files like .bat, .exe, or other runnable scripts, making it suitable for .NET tools or custom executables.</remarks>
+        Task<IPlugin> GetOrCreateRunnablePluginAsync(
```
zivkan: A good API is one that's hard to use wrong, and since this method has a similar name, and the exact same parameter list as the existing method, so I think it's not clear. Obviously, the XML doc can help to some extent, but if the API design makes it "impossible" to call wrong, then that's even better.

However, in this case since `PluginFile`'s property saying whether or not the plugin uses the old directory layout, or if it's directly executable, is a private property, it means it's impossible for external apps to call this API correctly.

Since `PluginDiscoverer` returns a `PluginFile`, the `GetOrCreate` method can take in an instance of `PluginFile` and hide this complexity, making it easier for the callers.

src/NuGet.Core/NuGet.Protocol/Plugins/IPluginFactory.cs
```diff
@@ -42,5 +42,36 @@ Task<IPlugin> GetOrCreateAsync(
...
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="arguments" />
+        /// is <see langword="null" />.</exception>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="requestHandlers" />
+        /// is <see langword="null" />.</exception>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="options" />
+        /// is <see langword="null" />.</exception>
+        /// <exception cref="OperationCanceledException">Thrown if <paramref name="sessionCancellationToken" />
+        /// is cancelled.</exception>
+        /// <exception cref="ObjectDisposedException">Thrown if this object is disposed.</exception>
+        /// <remarks>This method is primarily used for directly executable plugin files like .bat, .exe, or other runnable scripts, making it suitable for .NET tools or custom executables.</remarks>
```
zivkan: > This method is primarily used

Saying something has a primary use suggests it also has secondary uses. However, if I call `PluginDiscoverer` and then pass all the results to this new `GetOrCreateRunnablePluginAsync` method, it's going to fail to execute any `plugin.dll` discovered in the NuGet's existing plugin directory structure. So, it's not primarily used to execute plugins found on the PATH, it's only used to execute plugins found on the PATH.

zivkan: If you feel the need to explain usage of this new method, it probably means the existing method also needs a similar clarification.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginFile.cs
```diff
@@ -21,19 +21,19 @@ public sealed class PluginFile
         public Lazy<PluginFileState> State { get; }
 
         /// <summary>
-        /// Is the plugin file, a dotnet tools plugin file?
+        /// Indicates if the plugin file is runnable, such as an executable or a script.
         /// </summary>
-        internal bool IsDotnetToolsPlugin { get; }
+        internal bool IsRunnablePluginFile { get; }
```
zivkan:
```suggestion
        internal bool FoundOnPath { get; }
```

Maybe this is a bit of a nitpick, but plugins from NuGet's plugin directory are runnable. It's just the .NET Core plugins are executed indirectly via the `dotnet` host app.   Another idea is something like `FromPluginDirectory`, but I like it less.  Yet another idea is to introduce an enum like `PluginModel` with values `Directory` and `Path`.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginFactory.cs
```diff
@@ -149,36 +149,59 @@ private async Task<IPlugin> CreatePluginAsync(
...
+                    WindowStyle = ProcessWindowStyle.Hidden,
+                };
 #endif
+            }
+            else
+            {
+                // A dotnet tools plugin. Execute it directly.
+                startInfo = new ProcessStartInfo(filePath)
+                {
+                    FileName = filePath,
```
zivkan: you're setting FileName twice, once via the constructor, and again via property assignment.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginFactory.cs
```diff
@@ -149,36 +149,59 @@ private async Task<IPlugin> CreatePluginAsync(
...
-                Arguments = args,
-                UseShellExecute = false,
-                RedirectStandardError = false,
-                RedirectStandardInput = true,
-                RedirectStandardOutput = true,
-                StandardOutputEncoding = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false),
-                WindowStyle = ProcessWindowStyle.Hidden,
-            };
+#if IS_DESKTOP
+                startInfo = new ProcessStartInfo(filePath)
```
zivkan: This `ProcessStartInfo` is identical to the one when `isRunnablePluginFile` is `true`.

Therefore, if `PluginDiscover` always sets the value to `true` on .NET Framework, then this `PluginFactory` doesn't need the conditional `#if` compilation.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginFile.cs
nkolev92: Standalone for ones that don't require specifying the dotnet host?
Self contained is another word that means a similar thing in other contexts

src/NuGet.Core/NuGet.Protocol/Plugins/IPluginFactory.cs
Nigusu-Allehu: I’ve updated the API as per your suggestion to make it more intuitive and reduce the potential for misuse. I merged the two methods into a single method:

```csharp
Task<IPlugin> GetOrCreateAsync(
    string filePath,
    IEnumerable<string> arguments,
    IRequestHandlers requestHandlers,
    ConnectionOptions options,
    bool isRunnablePluginFile,
    CancellationToken sessionCancellationToken);
```

The `isRunnablePluginFile` parameter allows callers to specify if the file is directly executable or a DLL, which addresses the ambiguity you mentioned. This update should make it clearer for external applications to use the API correctly, without needing direct access to `PluginFile`'s private properties.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginFile.cs
jgonz120: Was there a reason we wanted to move away from IsDotnetToolsPlugin? I was going to suggest `RequiresDotnetHost` but I'm not sure if we're trying to avoid including "dotnet" in the name.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/Plugins/PluginDiscovererTests.cs
```diff
@@ -525,7 +525,6 @@ public void GetPluginsInNuGetPluginPaths_NuGetPluginPathsPointsToAFileThatDoesNo
 
             // Assert
             Assert.Single(plugins);
-            Assert.False(plugins[0].IsDotnetToolsPlugin);
```
jgonz120, line 528: What does the value return now?

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -240,7 +245,7 @@ internal List<PluginFile> GetPluginsInPath()
                 }
                 else
                 {
-                    pluginFiles.Add(new PluginFile(path, new Lazy<PluginFileState>(() => PluginFileState.InvalidFilePath)));
+                    pluginFiles.Add(new PluginFile(path, new Lazy<PluginFileState>(() => PluginFileState.InvalidFilePath), isRunnablePluginFile: false));
```
jgonz120: Is there a test to validate this scenario?

jgonz120 (approved):
Looks good to me, I think a new name for IsRunnablePluginFile would be good though.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginFile.cs
zivkan: the plugin can be written in any languaege, it doesn't have to be a .NET tool

src/NuGet.Core/NuGet.Protocol/Plugins/IPluginFactory.cs
```diff
@@ -37,7 +37,7 @@ public interface IPluginFactory : IDisposable
         /// <exception cref="ObjectDisposedException">Thrown if this object is disposed.</exception>
         /// <remarks>This is intended to be called by NuGet client tools.</remarks>
         Task<IPlugin> GetOrCreateAsync(
-            string filePath,
+            PluginFile pluginFile,
```
zivkan, line 40: Considering this is a breaking API change, and the only member in the interface, I don't see the value in keeping the interface public, and therefore this method becoming a new public API.  IMO either:

1. Make the interface internal

or

2. Keep the old method, and add a new overload

[A quick search of the interface name](https://github.com/search?q=repo%3ANuGet/NuGet.Client%20IPluginFactory&type=code) seems to suggest it's only used within NuGet.Protocol, so it should be technically possible without breaking NuGet itself.

zivkan (approved):

test/NuGet.Core.Tests/NuGet.Protocol.Tests/Plugins/PluginDiscovererTests.cs
Nigusu-Allehu: The property has been renamed to `RequiresDotnetHost ` and it returns environment based value. It would be false if it is desktop and true otherwise.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: I felt we don’t need a test specifically for this scenario, as it would end up testing the behavior of `PathValidator.IsValidLocalPath` and `PathValidator.IsValidUncPath` within the if statement. Since the else case only triggers if both methods return false, testing this would indirectly validate `PathValidator`’s internal logic, which I felt isn’t necessary for this context.

PR #4010, opened 2021-04-21
Allow specifying binlog path for RestoreTaskEx

ViktorHofer:
## Bug

Fixes: https://github.com/NuGet/Home/issues/10789

Regression? Last working version:

## Description

As it's hard to set environment variables from inside MSBuild without a custom task and the nuget static graph console reads from the environment variables, allow specifying a path for the restore binlog to be written to.

## PR Checklist

- [ ] PR has a meaningful title
- [ ] PR has a linked issue.
- [ ] Described changes

- **Tests**
  - [ ] Automated tests added
  - **OR**
  
  - [ ] Test exception
  - **OR**
  - [ ] N/A 

- **Documentation**
  
  - [ ] Documentation PR or issue filled
  - **OR**
  - [ ] N/A

PR #6116, opened 2024-10-23
Allow users to specify MSBuild binary logger parameters for static graph-based restore

### LINKED ISSUE #10789: Allow specifying the msbuild binlog path when invoking static graph restore to avoid modifying the environment
## Details about Problem

NuGet product used (NuGet.exe | Visual Studio | MSBuild.exe | dotnet.exe): dotnet.exe

Product version: 6.0.100-preview.3.21202.5

Worked before? If so, with which NuGet version: 

## Repro steps and/or sample project

When using static graph restore, the environment variable `RESTORE_TASK_BINLOG_PARAMETERS` needs to be set to generate a binlog for the out-of-process evaluation of the projects. If you are already inside an msbuild instance, it's not possible to modify the existing environment variables without a custom task (either compiled or inline).

To avoid modifying the existing environment, `RestoreTaskEx` task could accept a property to specify the path to the restore binlog which is then passed in as an environment variable to the restore task console.

### LINKED ISSUE #13876: Debugging large repos with static graph-based restore is slow
### NuGet Product(s) Affected

MSBuild.exe, dotnet.exe

### Current Behavior

At the moment when debugging static graph-based restores, an MSBuild binary log is created automatically.  However, when repositories are quite big, I've noticed that the MSBuild portion of the restore takes a long time.  This is because the MSBuild binary logger is including all of the project evaluations, imports, and misc files.

### Desired Behavior

I had static graph-based restore automatically generate a binary log when debugging as a convivence when developing the feature.  However, now its probably not necessary.

### Additional Context

_No response_

jeffkl:
# Bug

Fixes: https://github.com/NuGet/Home/issues/10789
Fixes: https://github.com/NuGet/Home/issues/13876

## Description
This pull requests allows users to specify MSBuild properties to enable the MSBuild binary logger for project evaluations and builds when using static graph-based restore.

| Property Name | Description | Default |
|---|---|---|
| `RestoreStaticGraphEnableBinaryLogger` | Enables the MSBuild binary logger for project evaluations and builds. | (null) |
| `RestoreStaticGraphBinaryLoggerParameters` | Specifie the parameters for the MSBuild binary logger.| `LogFile=nuget.binlog` if `RestoreStaticGraphEnableBinaryLogger` is `true` |

If the user specifies any value for the `RESTORE_TASK_BINLOG_PARAMETERS` environment variable, the binary logger is enabled with those parameters.

This also stops automatically adding the binary logger when debugging to fix https://github.com/NuGet/Home/issues/13876

## PR Checklist

- [x] Added tests


jeffkl:
This supersedes https://github.com/NuGet/NuGet.Client/pull/4010

nkolev92 (approved):

PR #6118, opened 2024-10-24
Make CLI parser public for dotnet nuget why

zivkan:
# Bug

Fixes: engineering issue

## Description

Running `dotnet nuget why` without a directory, solution, or project file reports an error saying that a required argument is missing:
image

But you'll notice that the command ran anyway.

The reason is that the dotnet/sdk repo has a duplicate set of CLI parser definitions to enable tab-completion. There was previously a bug, which required the target to be supplied, but https://github.com/NuGet/NuGet.Client/pull/5969 fixed it by making it implicit. However, the dotnet/sdk's duplicate definition was not updated to contain the same change.

Since `dotnet nuget why` is using System.CommandLine, rather than duplicating the same change in the second repo, I intend on making dotnet/sdk use the new public API introduced by this PR, so that NuGet.Client will be a single source of truth for how to parse `dotnet nuget why`. It will also mean that why will run in the dotnet CLI's process, rather than the NuGet.CommandLine.XPlat child process.  But this is our long term plan with migrating to System.CommandLine anyway. 

So, this is step 1, create a public API that the dotnet CLI can call.  Once this change has been inserted in the dotnet/sdk repo, I can create a PR there to delete their WhyCommandParser and call this instead.

## Alternative

I investigated moving the logic of invoking NuGet.CommandLine.XPlat as a child process into NuGet.Client, and also all the NuGetCommandParser definitions from the dotnet/sdk repo here. However, it was looking to be a lot of work, and I want to get this CX bug fixed more quickly.

If the team feels strongly that we shouldn't create the "temporary" public API, I can focus more effort on this.

I did however create https://github.com/NuGet/NuGet.Client/pull/6117 as a first step towards making it easier to move all that dotnet/sdk code into NuGet.Client. Whether or not that PR gets accepted has a big impact on how to implement the rest of the move of `dotnet nuget` parser out of the dotnet/sdk repo and into here.

## PR Checklist

- [x] Meaningful title, helpful description ~and a linked NuGet/Home issue~ (no product changes, it's more of a refactoring change)
- [x] ~Added tests~ it will be covered by the existing Dotnet.Integration.Tests once the change is commit in the dotnet/sdk repo.
- [x] ~Link to an issue or pull request to update docs if this PR changes settings, environment variables, new feature, etc.~ no customer facing changes.

### DIFF
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Why/WhyCommand.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Why/WhyCommand.cs
@@ -10,7 +10,7 @@ using Microsoft.Extensions.CommandLineUtils;
 
 namespace NuGet.CommandLine.XPlat.Commands.Why
 {
-    internal static class WhyCommand
+    public static class WhyCommand
     {
         internal static void Register(CommandLineApplication app)
         {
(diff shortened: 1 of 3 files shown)


jeffkl (approved):

nkolev92 (approved):
I love this, it's what I'm hoping we can do for all of our commands in https://github.com/NuGet/Home/issues/13089. 

I think it'll make our lives easier long term!

Well done @zivkan!

martinrrm (approved):

PR #6124, opened 2024-10-29
Expand Locale parser to support three-character language codes

### LINKED ISSUE #12253: Satellite assemblies for three-letter languages are not copied from NuGet package
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/Satellite-assemblies-for-three-letter-la/10203830)._

---
Satellite assemblies for cultures with three-letter language codes, like “sma-NO” with “Sami, Southern (Norway)” provided in an NuGet package are not copied to the build output directory as two-letter language code cultures as “nb-NO” with “Norwegian Bokmål (Norway)” are. The three-letter language code satellite assembly is included in the NuGet package, and is located in the expected “sma-NO” subdirectory, next to two-letter language code satellite assemblies:

.nuget\packages\package.name\1.1.8\lib\netstandard2.0

When referencing projects within the same solution, without the use of a NuGet package, three-letter language code satellite assemblies are copied to the output directory as expected.

This issue has been observed over time with VS 17.3.x and is still not working with 17.4.1.

---

I can repro this issue.
image
image

See original VSissue for repro steps and code.

baronfel:
# Bug

Fixes: https://github.com/NuGet/Home/issues/12253

## Description

Per https://github.com/NuGet/Home/issues/12253#issuecomment-2442827723, I found that all cultures known to .NET on my Windows and Linux hosts used 2 or 3-character language codes, so I expanded the heuristic in the `Locale_Parser` to this to enable handling of resources for these locales in .NET packages. I also added comments explaining _why_ these decisions were made and a note about future fully-correct implementations if those become required.

## PR Checklist

- [x] Added tests - I would appreciate some pointers where to add tests for this

### DIFF
--- a/src/NuGet.Core/NuGet.Packaging/ContentModel/ManagedCodeConventions.cs
+++ b/src/NuGet.Core/NuGet.Packaging/ContentModel/ManagedCodeConventions.cs
@@ -171,6 +171,8 @@ namespace NuGet.Client
                 }
             }
 
+            // We use a heuristic here for common locale codes. Locale codes are often
+            // * two characters for the language: en, es, fr, de
             if (name.Length == 2)
             {
                 if (matchOnly)
@@ -179,7 +181,9 @@ namespace NuGet.Client
                 }
                 return name.ToString();
             }
-            else if (name.Length >= 4 && name.Span[2] == '-')
+
+            // * a language portion that is two or three characters followed by a '-' and a country code
+            else if (name.Length >= 4 && name.Span[2] == '-') // e.g. en-US
             {
                 if (matchOnly)
                 {
@@ -187,6 +191,15 @@ namespace NuGet.Client
                 }
                 return name.ToString();
             }
+            else if (name.Length >= 5 && name.Span[3] == '-') // e.g agq-CM
+            {
+                return name;
+            }
+
+            // there are other variations, but this heuristic doesn't cover them all. A future-proof implementation would make
+            // use of the .NET CultureInfo APIs to compare the locale against the underlying system ICU database. This would
+            // be correct, but potentially more expensive because the CultureInfo APIs are lazily-loaded and throw if an
+            // invalid/unknown locale is used.
 
             return null;
         }
--- a/src/NuGet.Core/NuGet.Packaging/ContentModel/ManagedCodeConventions.cs
+++ b/src/NuGet.Core/NuGet.Packaging/ContentModel/ManagedCodeConventions.cs
@@ -198,7 +198,7 @@ namespace NuGet.Client
                 {
                     return string.Empty;
                 }
-                return name;
+                return name.ToString();
             }
 
             // there are other variations, but this heuristic doesn't cover them all. A future-proof implementation would make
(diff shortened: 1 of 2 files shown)


src/NuGet.Core/NuGet.Packaging/ContentModel/ManagedCodeConventions.cs
```diff
@@ -159,14 +159,27 @@ private static object Locale_Parser(string name, PatternTable table)
                 }
```
nkolev92, line 171: Maybe we make the method internal and then just add some tests. 
For example, we get the list of all cultures from the runtime and ensure all are parsed correctly?

That'd tell us if we ever need an update again. Depending on the runtime for a test probably a bit of anti-pattern, but we want to support all locale options that exist.

src/NuGet.Core/NuGet.Packaging/ContentModel/ManagedCodeConventions.cs
```diff
@@ -171,22 +171,40 @@ private static object Locale_Parser(ReadOnlyMemory<char> name, PatternTable tabl
...
             }
 
-            if (name.Length == 2)
+            // We use a heuristic here for common locale codes. Locale codes are often
+            // * two characters for the language: en, es, fr, de
+            // * three characters for the language: agq
+            if (name.Length == 2 || name.Length == 3)
```
baronfel, line 177: This length=3 check wasn't discussed in the issue but is required to pass tests because all of the XXX-YYYY locales also have a XXX form.. Hopefully it's not taking thing a bit too far.

src/NuGet.Core/NuGet.Packaging/ContentModel/ManagedCodeConventions.cs
baronfel: Thanks - I've added a test that does just that!

nkolev92 (dismissed):

src/NuGet.Core/NuGet.Packaging/ContentModel/ManagedCodeConventions.cs
```diff
@@ -171,22 +171,40 @@ private static object Locale_Parser(ReadOnlyMemory<char> name, PatternTable tabl
...
                 }
                 return name.ToString();
             }
+            else if (name.Length >= 5 && name.Span[3] == '-') // e.g agq-CM
+            {
+                if (matchOnly)
+                {
+                    return string.Empty;
+                }
+                return name;
```
baronfel: the test failures are because after my rebase `name` is a ROS<char> and that doesn't coerce to `string`. I should be returning `name.ToString()` here like on the other branches.

nkolev92 (approved):

jeffkl (approved):
Thanks for the fix!

clairernovotny:
@baronfel What .NET SDK version did this make it into?

baronfel:
9.0.200+ for sure, _perhaps_ 9.0.100 but I'm not certain given the backport bar...

nkolev92:
Issue says 6.13, so that's 9.0.200

https://learn.microsoft.com/en-us/nuget/release-notes/nuget-6.13

clairernovotny:
Fantastic, thank you. I've been preparing to release Humanizer 3.0 -- an RC is currently live now -- and I added the fil locale back into the default metapackage. I'll give it a go on SDK 8 and see if/how it breaks 😅. 

What is the current story with the .NET Framework version of MSBuild w.r.t. NuGet client versions? I'm trying to understand the potential impact of this as I do not want to suddenly break people! :)

nkolev92:
17.13 and newer should have this fix.

PR #6125, opened 2024-10-29
Add telemetry for service entries

### LINKED ISSUE #13744: Collect telemetry for HTTPS sources with HTTP resources

Nigusu-Allehu:
# Bug

Fixes:  https://github.com/NuGet/Home/issues/13744

### Telemetry Assessment: https://github.com/NuGet/Client.Engineering/pull/3034

This PR adds telemetry for service index resources that are HTTP, but their source is HTTPS.
```json
{
    "version": "3.0.0",
    "resources": [
        {
            "@id": "http://source",
        }]
}
```
The telemetry data logged in the following events

**Event name:**  &lt;`vs/nuget/RestorePackageSourceSummary`&gt; 
**Event name:**  &lt;`vs/nuget/SearchPackageSourceSummary`&gt; 

The new property added to these events is the following.

Property | Description
----|----
NumHTTPSSourcesWithAnHTTPResource | The number of HTTPS NuGet sources that have at least one HTTP resource.

Currently, getting the service index information is challenging because we want to avoid making unnecessary HTTP requests. Ideally, we want to retrieve this information from a cache. The problem is that we need telemetry at a higher level to count the number of sources that contain HTTP resources. This requires us to have access to all sources and loop through them to determine how many have HTTP resources.

The following are some of the solutions discussed:

### Protocol Diagnostics:

We considered adding properties to the existing protocol diagnostics to indicate whether a resource uses HTTP or HTTPS. This approach leverages the existing telemetry infrastructure, which means we wouldn't need to make changes to the protocol itself. By building on the existing telemetry, we can monitor HTTP resources effectively. This method allows us to track the protocol used by each resource without altering the public APIs, making it a seamless integration into our current system1.

### Cache File Reading:

We also explored the idea of reading from the cache file directly. This method involves identifying the correct files for telemetry based on deterministic file names. However, we had concerns about potential deadlocks and exceptions that could arise from this approach. The implementation challenges include whether to duplicate the logic for determining file names or handle it through a lower-level API. Despite these challenges, reading from the cache file could provide detailed telemetry data.

Conclusion:

After evaluating both approaches, I have decided to go with the implementation that uses protocol diagnostics. This allows us to keep a record of resource enquiries in real-time and check if they are HTTP or HTTPS. This data is collected at a higher level and used to count the number of HTTPS sources with an HTTP

## PR Checklist

- [x] Added tests


src/NuGet.Clients/NuGet.VisualStudio.Common/Telemetry/PackageSourceTelemetry.cs
```diff
@@ -65,10 +65,22 @@ public PackageSourceTelemetry(IEnumerable<SourceRepository> sources, Guid parent
             ProtocolDiagnostics.HttpEvent += ProtocolDiagnostics_HttpEvent;
             ProtocolDiagnostics.ResourceEvent += ProtocolDiagnostics_ResourceEvent;
             ProtocolDiagnostics.NupkgCopiedEvent += ProtocolDiagnostics_NupkgCopiedEvent;
+            ProtocolDiagnostics.ServiceIndexEntryEvent += ProtocolDiagnostics_ServiceIndexEntryEvent;
```
Nigusu-Allehu, line 68: Subscribe to ServiceIndexEntryEvent

src/NuGet.Clients/NuGet.VisualStudio.Common/Telemetry/PackageSourceTelemetry.cs
```diff
@@ -65,10 +65,22 @@ public PackageSourceTelemetry(IEnumerable<SourceRepository> sources, Guid parent
...
             _parentId = parentId;
             _actionName = GetActionName(action);
         }
 
+        private void ProtocolDiagnostics_ServiceIndexEntryEvent(ProtocolDiagnosticServiceIndexEntryEvent pdEvent)
```
Nigusu-Allehu, line 73: Event listener. When there is a ServiceIndexEntryEvent, save whether the source an http resource

src/NuGet.Core/NuGet.Protocol/Model/ServiceIndexEntry.cs
```diff
@@ -26,7 +28,7 @@ public class ServiceIndexEntry
         /// </summary>
         public SemanticVersion ClientVersion { get; }
 
-        public ServiceIndexEntry(Uri serviceUri, string serviceType, SemanticVersion clientVersion)
+        internal ServiceIndexEntry(Uri serviceUri, string serviceType, SemanticVersion clientVersion, PackageSource packageSource)
```
nkolev92: If you just raise the event from the ServiceIndexResourceV3 you don't need all these changes. 
Makes the PR footprint a lot smaller.

src/NuGet.Core/NuGet.Protocol/Model/ServiceIndexEntry.cs
```diff
@@ -43,9 +45,17 @@ public ServiceIndexEntry(Uri serviceUri, string serviceType, SemanticVersion cli
                 throw new ArgumentNullException(nameof(clientVersion));
             }
 
+            if (packageSource != null)
+            {
+                var resourceIsHttp = serviceUri.Scheme == Uri.UriSchemeHttp && serviceUri.Scheme != Uri.UriSchemeHttps;
```
nkolev92: Consider where you want to raise this event every time. 
It's probably rare, so I'd probably opt-in for the more allocation friendly approach.

src/NuGet.Core/NuGet.Protocol/Resources/ServiceIndexResourceV3.cs
```diff
@@ -105,7 +111,25 @@ private IReadOnlyList<ServiceIndexEntry> GetBestVersionMatchForType(NuGetVersion
             else
             {
                 // Find all entries with the same version.
-                return entries.Where(e => e.ClientVersion == bestMatch.ClientVersion).ToList();
+                List<ServiceIndexEntry> serviceIndexEntries = new List<ServiceIndexEntry>();
+
+                foreach (var entry in entries)
```
nkolev92: Shouldn't we be doing this in MakeLookup, ensuring it's done once rather than every time a resource is requested?

src/NuGet.Core/NuGet.Protocol/Resources/ServiceIndexResourceV3.cs
```diff
@@ -160,6 +164,14 @@ private static IDictionary<string, List<ServiceIndexEntry>> MakeLookup(JObject i
                         continue;
                     }
 
+                    // Capture if the resource is http & the source is https
+                    var resourceIsHttp = uri.Scheme == Uri.UriSchemeHttp && uri.Scheme != Uri.UriSchemeHttps;
+
+                    if (packageSource != null && resourceIsHttp && packageSource.IsHttps)
+                    {
+                        ProtocolDiagnostics.RaiseEvent(new ProtocolDiagnosticServiceIndexEntryEvent(packageSource.Source, true));
```
donnie-msft: nit: name parameters

src/NuGet.Core/NuGet.Protocol/Resources/ServiceIndexResourceV3.cs
```diff
@@ -160,6 +164,14 @@ private static IDictionary<string, List<ServiceIndexEntry>> MakeLookup(JObject i
                         continue;
                     }
 
+                    // Capture if the resource is http & the source is https
+                    var resourceIsHttp = uri.Scheme == Uri.UriSchemeHttp && uri.Scheme != Uri.UriSchemeHttps;
```
donnie-msft: Isn't `Uri.UriSchemeHttp` enough? These are strings not enums, so the value can't be both UriSchemeHttp and UriSchemeHttps, anyway.

test/NuGet.Clients.Tests/NuGet.SolutionRestoreManager.Test/SourceTelemetryTests.cs
```diff
@@ -34,7 +34,7 @@ public class SourceTelemetryTests
         private const string NumHTTPSv3Feeds = "NumHTTPSv3Feeds";
 
         private static readonly Guid Parent = Guid.Parse("33411664-388A-4C48-A607-A2C554171FCE");
-        private static readonly PackageSourceTelemetry.Totals ProtocolDiagnosticTotals = new PackageSourceTelemetry.Totals(1, 2, TimeSpan.FromMilliseconds(3));
+        private static readonly PackageSourceTelemetry.Totals ProtocolDiagnosticTotals = new PackageSourceTelemetry.Totals(1, 2, TimeSpan.FromMilliseconds(3), 0);
```
donnie-msft: nit: named parameters

test/NuGet.Core.Tests/NuGet.Protocol.Tests/ServiceIndexResourceV3Tests.cs
```diff
@@ -24,6 +26,79 @@ public void Constructor_InitializesProperties()
...
+            var content = CreateServiceIndexWithFourResourceTypesTwoHTTP();
+
+            var expectedRequestTime = DateTime.UtcNow;
+            var resource = new ServiceIndexResourceV3(content, expectedRequestTime, new Configuration.PackageSource(source));
+
+            // Act
+            var result = resource.GetServiceEntries(ServiceTypes.SearchQueryService);
+
+            // Assert
+            int httpSourceCapture = 0;
```
donnie-msft: Source and resource mean very different things in this context, so I suggest being strict with the naming.

```suggestion
            int httpResourceCapture = 0;
```

src/NuGet.Clients/NuGet.VisualStudio.Common/Telemetry/PackageSourceTelemetry.cs
```diff
@@ -374,27 +390,35 @@ internal static Totals GetTotals(IReadOnlyDictionary<string, Data> data)
...
 
             public int Requests { get; }
             public long Bytes { get; }
             public TimeSpan Duration { get; }
+            public int NumberOfSourcesWithAnHttpResource { get; }
         }
 
         internal class Data
         {
+            internal bool HttpSourceHasHttpResource { get; set; }
```
donnie-msft: Am I missing a subtly or should this also be HttpsSource?

```suggestion
            internal bool HttpsSourceHasHttpResource { get; set; }
```

donnie-msft (dismissed):

nkolev92 (dismissed):

donnie-msft (approved):

src/NuGet.Core/NuGet.Protocol/Resources/ServiceIndexResourceV3.cs
```diff
@@ -160,6 +164,12 @@ private static IDictionary<string, List<ServiceIndexEntry>> MakeLookup(JObject i
                         continue;
                     }
 
+                    // Capture if the resource is http & the source is https
+                    if (packageSource != null && uri.Scheme != Uri.UriSchemeHttps && packageSource.IsHttps)
```
donnie-msft: Oops, I approved because my suggestion was marked resolved but I just noticed this is the opposite of what I suggested.

```suggestion
                    if (packageSource != null && uri.Scheme == Uri.UriSchemeHttp && packageSource.IsHttps)
```

donnie-msft (changes requested):

src/NuGet.Core/NuGet.Protocol/Events/ProtocolDiagnosticServiceIndexEntryEvent.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.Protocol.Events
+{
+    public sealed class ProtocolDiagnosticServiceIndexEntryEvent
```
jeffkl, line 9: Please add XML doc comments to any public type

jeffkl (approved):

donnie-msft (approved):

PR #6127, opened 2024-10-29
Ban usage of Environment.GetEnvironmentVariable and Environment.SetEnvironmentVariable

jeffkl:
# Bug

Fixes: 

## Description
In a recent change, I was modifying a component that was reading an environment variable. We have an interface, `NuGet.Common.IEnvironmentVariableReader` but it is not required.  With the Banned API analyzer, we can now ensure that components are using interfaces to read environment variables.  Environment variables are static so setting them in a test can affect other tests.  Any component that uses environment variables should use the interface instead.  I think tests modifying static state is contributing to overall reliability issues in our CI.

`IEnvironmentVariableReader` interface only abstracts the reading of environment variables while a few places in our code set them as well.  So any existing code path that sets an environment variable has the warning suppressed.

Public types should have a public constructor and an internal constructor so that tests can abstract away environment variable manipulation:

```c#
public class SomeClass
{
    private readonly IEnvironmentVariableReader _environmentVariableReader;

    public SomeClass()
        : this(EnvironmentVariableWrapper.Instance)
    {
    }

    internal SomeClass(IEnvironmentVariableReader environmentVariableReader)
    {
        _environmentVariableReader = environmentVariableReader ?? throw new ArgumentNullException(nameof(environmentVariableReader));
    }

    public string DetermineSomeValue()
    {
        return _environmentVariableReader.GetEnvironmentVariable("SomeVariable");
    }
}
```

Public static methods should have an internal overload that accepts an `IEnvironmentVariableReader`:
```cs
public static string DetermineSomeValue()
{
    return DetermineSomeValue(EnvironmentVariableWrapper.Instance);
}

internal static string DetermineSomeValue(IEnvironmentVariableReader environmentVariableReader)
{
    return environmentVariableReader.GetEnvironmentVariable("SomeVariable");
}
```

That way unit tests can call the internal one with the abstraction around environment variable usage.

## PR Checklist

- [ ] Added tests


src/NuGet.Core/NuGet.Common/IEnvironmentVariableReader.cs
```diff
@@ -1,8 +1,11 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+
 namespace NuGet.Common
 {
+    [Obsolete("Use IEnvironmentVariableProvider instead.")]
     public interface IEnvironmentVariableReader
```
zivkan: Considering this PR changes many instances of `IEnvironmentVariableReader` to `IEnvironmentVariableProvider`, are there any references of this interface remaining?

If every place a customer using the NuGet Client SDK has already been broken, there's no point keeping this interface around with an obsolete message

docs/coding-guidelines.md
```diff
@@ -274,6 +274,46 @@ For example the following are correct:
 
 Many of the guidelines, wherever possible, and potentially some not listed here, are enforced by an [EditorConfig](https://editorconfig.org "EditorConfig homepage") file (`.editorconfig`) at the root of the repository.
 
+### Getting or Setting Environment Variables
+Environment variables apply to the entire process and are considered static state which can cause test issues since multiple tests can be running in parallel reading or updating the same variable.  All components that use environment variables should use the `IEnvironmentVariableProvider` interface.  An instance type should have a public constructor and an internal constructor which accepts an environment variable provider:
```
zivkan: 1. markdown linters suggest that there should be a line of whitespace between headings and their content
2. I always recommend one sentence per line.

docs/coding-guidelines.md
```diff
@@ -274,6 +274,46 @@ For example the following are correct:
...
+
```
zivkan, line 286: I'd add a 1 or 2 sentence paragraph saying something like:

It's preferred that any projects that are part of the NuGet Client SDK do not use environment variables directly. Instead, prefer method parameters or class properties that developers using our packages can use themselves, and read the environment variable in the code not part of packages that we publish.

src/NuGet.Core/NuGet.Common/PathUtil/NuGetEnvironment.cs
```diff
@@ -336,12 +339,11 @@ private static string GetValueOrThrowMissingEnvVar(Func<string> getValue, string
...
         public static string GetDotNetLocation()
         {
-            var path = Environment.GetEnvironmentVariable("PATH");
+            var path = EnvironmentVariableProvider.GetEnvironmentVariable("PATH");
             var isWindows = RuntimeEnvironmentHelper.IsWindows;
-            var splitChar = isWindows ? ';' : ':';
             var executable = isWindows ? DotNetExe : DotNet;
 
-            foreach (var dir in path.Split(splitChar))
+            foreach (var dir in path.Split(Path.PathSeparator))
```
zivkan, line 345: This is not the same, and I'm a bit concerned why no test is failing because of it

src/NuGet.Core/NuGet.Common/IEnvironmentVariableReader.cs
jeffkl: There are no more references in our codebase.  I was originally thinking of trying to implement all of this in a non-breaking way but the places where `IEnvironmentVariableReader` are exposed I didn't feel like were high impact.  I'll just remove these types.

src/NuGet.Core/NuGet.Common/PathUtil/NuGetEnvironment.cs
jeffkl: This should be the same, `Path.PathSeparator` is `;` on Windows and `:` on non-Windows.  Are you thinking I used `Path.DirectorySeparator`?  That's different.

https://learn.microsoft.com/en-us/dotnet/api/system.io.path.pathseparator?view=net-8.0#remarks

src/NuGet.Core/NuGet.Common/SystemEnvironmentVariableProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+        public string? GetEnvironmentVariable(string variable, EnvironmentVariableTarget target) => Environment.GetEnvironmentVariable(variable, target);
+#pragma warning restore RS0030 // Do not use banned APIs
+
+#pragma warning disable RS0030 // Do not use banned APIs
+        public void SetEnvironmentVariable(string name, string? value) => Environment.SetEnvironmentVariable(name, value);
+#pragma warning restore RS0030 // Do not use banned APIs
+
+#pragma warning disable RS0030 // Do not use banned APIs
+        public void SetEnvironmentVariable(string name, string? value, EnvironmentVariableTarget target) => Environment.SetEnvironmentVariable(name, value, target);
+#pragma warning restore RS0030 // Do not use banned APIs
```
Nigusu-Allehu: nit
```suggestion
#pragma warning disable RS0030 // Do not use banned APIs
        public string? GetEnvironmentVariable(string variable) => Environment.GetEnvironmentVariable(variable);

        public string? GetEnvironmentVariable(string variable, EnvironmentVariableTarget target) => Environment.GetEnvironmentVariable(variable, target);

        public void SetEnvironmentVariable(string name, string? value) => Environment.SetEnvironmentVariable(name, value);
        
        public void SetEnvironmentVariable(string name, string? value, EnvironmentVariableTarget target) => Environment.SetEnvironmentVariable(name, value, target);
#pragma warning restore RS0030 // Do not use banned APIs
```

nkolev92:
I was thinking about the design here, we use `IEnvironmentVariableReader` in our APIs, and originally I couldn't think of why our code would ever be setting env vars. 
Then I looked into the changes, and I saw static graph restore, and PMC setting environment variables.
So I was like ok, I guess these are special, so why does every other place in the code need access to a set environment variable?
It feels like overexposure to me. 

I feel like it might be better if we just made the new API inherit from the old one and then just pass the "new" one only in those types where setting is actually needed. 
In my naive reasoning without attempting the change, it seems like we'd be able to avoid any public API changes whatsoever. 

Thoughts?

jeffkl:
I think it might help if I summarize the public API changes:

### NuGet.Common

```diff
-NuGet.Common.IEnvironmentVariableReader
+NuGet.Common.IEnvironmentVariableProvider

-NuGet.Common.EnvironmentVariableWrapper
+NuGet.Common.SystemEnvironmentVariableProvider
```

### NuGet.Configuration

```diff
-NuGet.Configuration.ProxyCache.ProxyCache(NuGet.Configuration.ISettings! settings, NuGet.Common.IEnvironmentVariableReader! environment) -> void
+NuGet.Configuration.ProxyCache.ProxyCache(NuGet.Configuration.ISettings! settings) -> void

-static NuGet.Configuration.SettingsUtility.GetRevocationMode(NuGet.Common.IEnvironmentVariableReader? environmentVariableReader = null) -> NuGet.Common.RevocationMode
+static NuGet.Configuration.SettingsUtility.GetRevocationMode(NuGet.Common.IEnvironmentVariableProvider? environmentVariableProvider = null) -> NuGet.Common.RevocationMode

```

### NuGet.Credentials

```diff
-~NuGet.Credentials.PluginCredentialProviderBuilder.PluginCredentialProviderBuilder(NuGet.Configuration.IExtensionLocator extensionLocator, NuGet.Configuration.ISettings settings, NuGet.Common.ILogger logger, NuGet.Common.IEnvironmentVariableReader envarReader) -> void
```
(There's already an overload to PluginCredentialProviderBuilder that does not take an environment varible abstraction)

### NuGet.PackageManagement

```diff
-~NuGet.PackageManagement.PackageRestoreConsent.PackageRestoreConsent(NuGet.Configuration.ISettings settings, NuGet.Common.IEnvironmentVariableReader environmentReader) -> void	

-~NuGet.PackageManagement.PackageRestoreConsent.PackageRestoreConsent(NuGet.Configuration.ISettings settings, NuGet.Common.IEnvironmentVariableReader environmentReader, NuGet.Configuration.ConfigurationDefaults configurationDefaults) -> void
+~NuGet.PackageManagement.PackageRestoreConsent.PackageRestoreConsent(NuGet.Configuration.ISettings settings, NuGet.Configuration.ConfigurationDefaults configurationDefaults) -> void
```

### NuGet.Packaging
```diff
-~static NuGet.Packaging.Signing.SignedPackageVerifierSettings.GetAcceptModeDefaultPolicy(NuGet.Common.IEnvironmentVariableReader environmentVariableReader = null) -> NuGet.Packaging.Signing.SignedPackageVerifierSettings	
+~static NuGet.Packaging.Signing.SignedPackageVerifierSettings.GetAcceptModeDefaultPolicy() -> NuGet.Packaging.Signing.SignedPackageVerifierSettings

-~static NuGet.Packaging.Signing.SignedPackageVerifierSettings.GetDefault(NuGet.Common.IEnvironmentVariableReader environmentVariableReader = null) -> NuGet.Packaging.Signing.SignedPackageVerifierSettings	
+~static NuGet.Packaging.Signing.SignedPackageVerifierSettings.GetDefault() -> NuGet.Packaging.Signing.SignedPackageVerifierSettings

-~static NuGet.Packaging.Signing.SignedPackageVerifierSettings.GetRequireModeDefaultPolicy(NuGet.Common.IEnvironmentVariableReader environmentVariableReader = null) -> NuGet.Packaging.Signing.SignedPackageVerifierSettings	
+~static NuGet.Packaging.Signing.SignedPackageVerifierSettings.GetRequireModeDefaultPolicy() -> NuGet.Packaging.Signing.SignedPackageVerifierSettings

-~static NuGet.Packaging.Signing.SignedPackageVerifierSettings.GetVerifyCommandDefaultPolicy(NuGet.Common.IEnvironmentVariableReader environmentVariableReader = null) -> NuGet.Packaging.Signing.SignedPackageVerifierSettings
+~static NuGet.Packaging.Signing.SignedPackageVerifierSettings.GetVerifyCommandDefaultPolicy() -> NuGet.Packaging.Signing.SignedPackageVerifierSettings
```

### NuGet.Protocol

```diff
-~NuGet.Protocol.Plugins.PluginManager.EnvironmentVariableReader.get -> NuGet.Common.IEnvironmentVariableReader
-~NuGet.Protocol.Plugins.PluginManager.PluginManager(NuGet.Common.IEnvironmentVariableReader reader, System.Lazy<NuGet.Protocol.Plugins.IPluginDiscoverer> pluginDiscoverer, System.Func<System.TimeSpan, NuGet.Protocol.Plugins.IPluginFactory> pluginFactoryCreator, System.Lazy<string> pluginsCacheDirectoryPath) -> void

-~static NuGet.Protocol.Plugins.ConnectionOptions.CreateDefault(NuGet.Common.IEnvironmentVariableReader reader = null) -> NuGet.Protocol.Plugins.ConnectionOptions
+~static NuGet.Protocol.Plugins.ConnectionOptions.CreateDefault() -> NuGet.Protocol.Plugins.ConnectionOptions
```

In my opinion, I don't see any actual affect on developers with these changes.  If `IEnvironmentVariableReader` usage was more widespread, I would be hesitant to touch it.  Does anyone have a compelling reason for any of those changes to not be made?  I imagine the fact that our abstraction ever creeped into the public APIs was a mistake since removing them did not cause me to have to add InternalsVisibleTo anywhere and everything still works.  As long as the interface is in `NuGet.Common` and public (even though no external calls need to know about it) then our code can pass it around as `internal` implementation details.

The overall design of my change makes sense to me.  Abstractions are used to cover the functionality of a component.  For example, a file system abstraction is for all things with files; reading, writing, checking existence, updating attributes, etc.  We wouldn't want an abstraction just for getting file attributes.  Environment variables can be read and written so it makes sense for any abstraction around environment variables to include both pieces of functionality.  Ideally, our abstraction would have had this from the beginning.  It's very odd to me that we only have an abstraction for reading environment variables.  Our abstraction doesn't even include reading machine-wide or user-wide environment variables.  Its very, very specific to just reading process-wide environment variables.

For me, if all of this was internal anyways, it would be a no-brainer to make this change.  The fact that our abstraction somehow became part of our public APIs is the only sticking point and I don't see any public API changes that would actually impact anything.

src/NuGet.Core/NuGet.Common/SystemEnvironmentVariableProvider.cs
jeffkl: I had it both ways when I was developing this.  The only problem with your suggestion is that it lets anyone add new code between the `#prama` that gets to use the suppression.  I know all of these suppressions are ugly in code but it makes sure that anyone adding to this class has to think about any code analysis warnings they might have introduced.  That said, I'm fine with either one, this is just what I personally settled on.  What does everyone else think?

nkolev92:
No piece of code in NuGet.Configuration ever needs to set an env var, so it doesn't need to go from IEnvironmentVariableReader to IEnvironmentVariableProvider, since it never provides. 

I know Environment is a static, but we can choose how much "capabilities" we provide for it, so until it needs to "provide" environment variables, I don't see why it should stop using a Reader and go to a provider.

src/NuGet.Core/NuGet.Common/PathUtil/NuGetEnvironment.cs
zivkan: 🤦

How do I make the 🤦 bigger, because it's a big 🤦 moment.

Yes, I was thinking DirectorySeparator 🤦

zivkan:
For what it's worth, I also felt surprised by this PR because I didn't expect NuGet to set environment variables anywhere at all.

When I first reviewed the PR I hoped it was only setting variables in one place, so I could comment to just make it internal to that assembly and not make public API changes. Unfortunately, there's two.  My personal preference would still be to create internal interfaces in those two places (or the class could take an `Action<string, string>` delegate in an internal overload, to make it testable).

I didn't mention this in my first review, because I'm trying not to let my idealism and my own subjective preferences block other peoples work. But just as a data point, I had similar feelings to what Nikolche wrote. I generally don't want Nuget setting environment variables, so I don't think it should be part of any common interface/API we have.

I really like adding Environment.GetEnvironmentVariable to the banned API analyzer file though. I've commented on several peoples PRs telling them to use it, so it's great to automate that.

jeffkl:
I've updated this to just ban Set and Get environment variables with no public API changes

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildFeatureFlags.cs
```diff
@@ -2,7 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-
+#pragma warning disable RS0030 // Do not used banned APIs.  This class is allowed to get and set environment variables.
```
zivkan: It's possible to set environment variable in `StartProcessInfo`, so `RestoreTaskEx` could set the variables for the child process, and then this console app won't need to modify its own environment variables.

docs/coding-guidelines.md
```diff
@@ -274,6 +274,64 @@ For example the following are correct:
 
 Many of the guidelines, wherever possible, and potentially some not listed here, are enforced by an [EditorConfig](https://editorconfig.org "EditorConfig homepage") file (`.editorconfig`) at the root of the repository.
 
+### Getting or Setting Environment Variables
```
donnie-msft, line 277: Nice addition!

build/BannedSymbols.txt
```diff
@@ -21,3 +21,9 @@ M:System.IO.Path.GetTempPath(); Use NuGetEnvironment.GetFolderPath(NuGetFolderPa
 
 M:NuGet.Protocol.Core.Types.SourceRepository.GetResource`1(); Pass a cancellation token
 M:NuGet.Protocol.Core.Types.SourceRepository.GetResourceAsync`1(); Pass a cancellation token
+
+M:System.Environment.GetEnvironmentVariable(System.String); Use NuGet.Common.IEnvironmentVariableReader
+M:System.Environment.GetEnvironmentVariable(System.String,System.EnvironmentVariableTarget); Use NuGet.Common.IEnvironmentVariableReader
+
+M:System.Environment.SetEnvironmentVariable(System.String,System.String); NuGet should not need to set environment variables. If you need to set one, you must seek an exception.
```
donnie-msft: nit: just curious if we should mention to find these guidelines in our repo, or even provide a link to the markdown?

src/NuGet.Clients/NuGet.PackageManagement.PowerShellCmdlets/PowerShellHost.cs
```diff
@@ -536,7 +538,9 @@ private static void AddPathToEnvironment(string path)
             if (currentPaths.Add(path))
             {
                 var newPath = currentPath + Path.PathSeparator + path;
-                Environment.SetEnvironmentVariable("path", newPath, EnvironmentVariableTarget.Process);
+#pragma warning disable RS0030 // Do not use banned APIs
+                Environment.SetEnvironmentVariable("path", newPath);
```
donnie-msft, line 542: A reason isn't given here as the new coding guideline requires 😄

src/NuGet.Clients/NuGet.PackageManagement.PowerShellCmdlets/Providers/PowerShellHostProvider.cs
```diff
@@ -27,16 +28,20 @@ internal class PowerShellHostProvider : IHostProvider
         public const string PowerConsoleHostName = "Package Manager Host";
 
         private readonly IRestoreEvents _restoreEvents;
+        private readonly IEnvironmentVariableReader _environmentVariableReader;
+
```
donnie-msft, line 32: nit: extra line

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildFeatureFlags.cs
donnie-msft: nit:
```suggestion
#pragma warning disable RS0030 // Do not used banned APIs.  This class is intended to get and set MSBuild environment variables.
```

donnie-msft (commented):
Looks great! I'm thinking I can refactor some tests which don't even consider the host machine having an environment variable set, to use an envvar provider parameter like you're doing here (eg, https://github.com/NuGet/Home/issues/13180).

Left a few nits. I'll approve once the 1 missing exception reasoning is added.

zivkan (dismissed):
I forgot to submit this yesterday 🤦

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildFeatureFlags.cs
donnie-msft: 👀

jgonz120 (dismissed):

build/BannedSymbols.txt
jeffkl: I'll add a link to the message

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildFeatureFlags.cs
jeffkl: Definitely a good idea, I'll consider that for another change though

src/NuGet.Clients/NuGet.PackageManagement.PowerShellCmdlets/PowerShellHost.cs
```diff
@@ -536,7 +538,9 @@ private static void AddPathToEnvironment(string path)
             if (currentPaths.Add(path))
             {
                 var newPath = currentPath + Path.PathSeparator + path;
-                Environment.SetEnvironmentVariable("path", newPath, EnvironmentVariableTarget.Process);
+#pragma warning disable RS0030 // Do not use banned APIs (This add a path to the PATH environment variable just for the PowerShell console session)
```
zivkan, line 541: "just" for the PowerShell console session, and also every single other VS component running in devenv.exe 🤣

zivkan (approved):

Nigusu-Allehu (approved):

PR #6131, opened 2024-10-31
Support cancellation during command-line based NuGet audit

### LINKED ISSUE #13644: Fetching Vulnerability Resources doesn't respect cancellation
### NuGet Product Used

dotnet.exe

### Product Version

.NET 9 preview 7

### Worked before?

_No response_

### Impact

### Repro Steps & Context

This requires a vulnerability information provider to have a network problem, either simulated or real. For me, the Azure outage this Thursday evening presented such that when I cancelled a `dotnet build` of the dotnet/sdk repo MSBuild attempted to cancel the build but the Restore continued - I got error messages like the following after the 90s timeout finished:

```
E:\Code\dotnet-sdk\src\Layout\toolset-tasks\toolset-tasks.csproj : error NU1900: Warning As Error: Error occurred while getting package vulnerability data: Unable to load the service index for source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet6/nuget/v3/index.json.
```
and
```
E:\Code\dotnet-sdk\artifacts\bin\redist\Debug\dotnet\sdk\9.0.100-dev\NuGet.targets(180,5): warning MSB4220: Waiting for the currently executing task "RestoreTask" to cancel.
```

To me this felt like a classic case of not passing a CTok, or using CTok.None. I dug into the RestoreTask and found that 

* RestoreDoes set up and [pass a CTok](https://github.com/NuGet/NuGet.Client/blob/46f80a3f1a83f3c5f33a20fa9fb72871e1e2d4eb/src/NuGet.Core/NuGet.Build.Tasks/RestoreTask.cs#L163)
* Restores are [run for each request](https://github.com/NuGet/NuGet.Client/blob/46f80a3f1a83f3c5f33a20fa9fb72871e1e2d4eb/src/NuGet.Core/NuGet.Build.Tasks/BuildTasksUtility.cs#L251)
* Also run [here](https://github.com/NuGet/NuGet.Client/blob/46f80a3f1a83f3c5f33a20fa9fb72871e1e2d4eb/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreRunner.cs#L251)
* Audit [is performed](https://github.com/NuGet/NuGet.Client/blob/46f80a3f1a83f3c5f33a20fa9fb72871e1e2d4eb/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs#L336-L337), which requires
* packages vulnerabilities to [be checked](https://github.com/NuGet/NuGet.Client/blob/46f80a3f1a83f3c5f33a20fa9fb72871e1e2d4eb/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs#L513), which requires
* [fetching vulnerability data](https://github.com/NuGet/NuGet.Client/blob/46f80a3f1a83f3c5f33a20fa9fb72871e1e2d4eb/src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs#L90), which requires
* querying vulnerability data [from the configured providers](https://github.com/NuGet/NuGet.Client/blob/46f80a3f1a83f3c5f33a20fa9fb72871e1e2d4eb/src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs#L332)

Through this whole chain the CTok is flowed in a way that makes sense to me. However, the Vulnerability provider interface seems to accept a CTok, but [not use it](https://github.com/NuGet/NuGet.Client/blob/46f80a3f1a83f3c5f33a20fa9fb72871e1e2d4eb/src/NuGet.Core/NuGet.Commands/RestoreCommand/VulnerabilityInformationProvider.cs#L34), perhaps for caching purposes?

I think if this was changed to accept the flowed-in CTok in some way that Restore would be properly cancellable again.
 

### Verbose Logs

```shell
I don't have standalone logs, but I did include some error messages above.
```

jeffkl:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13644

## Description
Normally, an async method like `VulnerabilityInformationProvider.GetVulnerabilityInformationAsync()` would accept a `CancellationToken` in order to support canceling the task.  However, this method uses an `AsyncLazy<T>` to ensure its only called once per restore.  This makes passing in a `CancellationToken` problematic calls might just get a cached value.

This change introduces a constructor to `VulnerabilityInformationProvider` which stores a `CancellationToken` as a private field to be passed along when `IVulnerabilityInfoResource.GetResourceAsync()` and `IVulnerabilityInfoResource.GetVulnerabilityInfoAsync()` are called.  So while the `GetVulnerabilityInfoAsync()` call is unorthodox, the calls inside the method follow the normal pattern.

I had to add a new public constructor to `RestoreCommandProvidersCache` but I left the existing one in place so it won't be a breaking change.

## PR Checklist

- [x] Added tests


src/NuGet.Core/NuGet.Commands/RestoreCommand/VulnerabilityInformationProvider.cs
```diff
@@ -17,19 +17,19 @@ internal sealed class VulnerabilityInformationProvider : IVulnerabilityInformati
...
 
-        public VulnerabilityInformationProvider(SourceRepository source, ILogger logger, bool isAuditSource)
+        public VulnerabilityInformationProvider(SourceRepository source, ILogger logger, bool isAuditSource, CancellationToken cancellationToken)
         {
             _source = source;
             _logger = logger;
             IsAuditSource = isAuditSource;
-
+            _cancellationToken = cancellationToken;
             _vulnerabilityInfo = new AsyncLazy<GetVulnerabilityInfoResult?>(GetVulnerabilityInfoAsync);
```
Nigusu-Allehu: Is it possible we do something like
```suggestion
            _vulnerabilityInfo = new AsyncLazy<GetVulnerabilityInfoResult?>(
                () => GetVulnerabilityInfoAsync(cancellationToken));
```

jeffkl: When the method took an argument, specifying the delegate made sense but when I removed the argument, VS recommended simplifying the lamda.

Nigusu-Allehu: I meant we could do
`_vulnerabilityInfo = new AsyncLazy<GetVulnerabilityInfoResult?>(
                () => GetVulnerabilityInfoAsync(cancellationToken));
`
and not make the _cancelation token part of the class. In addition, put the defintion of `GetVulnerabilityInfoAsync(cancellationToken)` back to one that takes a cancelation token as an argument. 

I was thinking if we make the cancelation token part of the class, it could result in all ongoing tasks stopping regardless of their individual needs. However, I don't think we create multiple calls to `GetVulnerabilityInfoAsync(cancellationToken)` with different canclationTokens so I guess it should not be a problem for this specific scenario.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommandProviders.cs
```diff
@@ -42,7 +43,7 @@ private static IReadOnlyList<IVulnerabilityInformationProvider> CreateVulnerabil
             {
                 if (remoteProviders[i].SourceRepository != null)
                 {
-                    providers.Add(new VulnerabilityInformationProvider(remoteProviders[i].SourceRepository, NullLogger.Instance, isAuditSource: false));
+                    providers.Add(new VulnerabilityInformationProvider(remoteProviders[i].SourceRepository, NullLogger.Instance, isAuditSource: false, CancellationToken.None));
```
zivkan: I was going to comment that this means the request can't be cancelled, but this class (well, all public constructors) is marked as obsolete, so this line is never executed

src/NuGet.Core/NuGet.Commands/RestoreCommand/VulnerabilityInformationProvider.cs
```diff
@@ -17,19 +17,19 @@ internal sealed class VulnerabilityInformationProvider : IVulnerabilityInformati
...
         }
 
-        public async Task<GetVulnerabilityInfoResult?> GetVulnerabilityInformationAsync(CancellationToken cancellationToken)
+        public async Task<GetVulnerabilityInfoResult?> GetVulnerabilityInformationAsync()
```
zivkan: You mentioned it's unusual in the PR description, but I just don't like moving the cancellation token into a constructor.  Operations should be cancelled, not the state of a class. I'm not aware of anything in the BCL that takes a CancellationToken in a constructor.

Another idea is for the class to use a `CancellationTokenSource`, which `GetVulnerabilityInfoAsync` uses for cancellation, and then in this method we create a second task that gets cancelled when the cancellation token is requested, and we react to it by requesting cancellation in the cancellation token source. Something like:

```csharp
if (_vulnerabilityInfo.HasValue)
{
    return _vulnerabilityInfo.Value;
}

Task cancellationTask = new Task(() => throw new NotImplementedException(), cancellationToken);
Task<GetVulnerabilityInfoResult?> resultTask = _vulnerabilityInfo;
var completedTask = await Task.WhenAny(cancellationTask, resultTask);

if (completedTask == cancellationTask)
{
  _cancellationTokenSource.Cancel();
  await resultTask;
}

return resultTask.Result;
```

In the future, this is extensible for more advanced logic, like ref counting how many callers are waiting for the result, and only cancelling the cancellation token source when the ref counter becomes zero. This PR's approach of moving the cancellation token to the constructor would require yet another public API break if we end up going down that path.

I also feel like it's a leaky abstraction, making a public API change to move the cancellation token to a constructor, that Client SDK callers shouldn't need to know about, and as mentioned, could change again if we make certain improvements to this code in the future.  I'd rather hide it so we don't need to think about public API changes.

src/NuGet.Core/NuGet.Commands/RestoreCommand/VulnerabilityInformationProvider.cs
```diff
@@ -17,19 +17,19 @@ internal sealed class VulnerabilityInformationProvider : IVulnerabilityInformati
     {
         private readonly SourceRepository _source;
         private readonly ILogger _logger;
-
+        private readonly CancellationToken _cancellationToken;
         private readonly AsyncLazy<GetVulnerabilityInfoResult?> _vulnerabilityInfo;
 
-        public VulnerabilityInformationProvider(SourceRepository source, ILogger logger, bool isAuditSource)
+        public VulnerabilityInformationProvider(SourceRepository source, ILogger logger, bool isAuditSource, CancellationToken cancellationToken)
```
nkolev92: A cancellation token is usually passed for a particular operation. This is tying a cancellation token to the lifetime of an object. 

The VulnerabilityInformationProvider would ideally have a different lifetime in VS, and this would make that more challenging.

src/NuGet.Core/NuGet.Commands/RestoreCommand/VulnerabilityInformationProvider.cs
nkolev92: Why not just change where the lazy is initialized and pass the first CancellationToken that comes in the GetVulnerabilityInformationAsync method.

zivkan: It's hard to guarantee that the lazy would only be initialized once. Well, with "good perf" anyway. If we do a `lock (_object) { if (_vulnerabilityInfo == null) { _vulnerabilityInfo = new(...); } }`, I think it should be set once, but then there's a bit of overhead, and if multiple projects get to this at the same time, there will be lock contention and I don't think `lock` does a spinlock, which means the OS will sleep the thread, and then it'll need to be rescheduled once the lock is released, causing additional context switching perf issues, which is precisely what async is supposed to avoid.

nkolev92: I'm suggesting locking around the lazy assignment.
The lock contention cost + likelihood around that is negligible.

```cs
public async Task<GetVulnerabilityInfoResult?> GetVulnerabilityInformationAsync(CancellationToken cancellationToken)
{
    if (!_initialized && _vulnerabilityInfo != null)
    {
        lock (_lockObject)
        {
            _vulnerabilityInfo = _vulnerabilityInfo  ?? new AsyncLazy<GetVulnerabilityInfoResult?>(() => GetVulnerabilityInfoAsync(cancellationToken));
            _initialized = true;
        }
    }

    GetVulnerabilityInfoResult? result = await _vulnerabilityInfo!;
    return result;
}
```

src/NuGet.Core/NuGet.Commands/RestoreCommand/VulnerabilityInformationProvider.cs
jeffkl: All threads in restore use the same operation since the result is cached.  The only other fix I can think of is to stop using `AsyncLazy<T>` and use a semaphore.  All threads could pass in their own cancellation token and they'd wait for the operation.  That could add lock contention, although technically there's already lock contention.

In this case, all threads are passing in the same cancellation token since only one is created during restore.  So I don't know if its worth making this abnormal class look more normal?

nkolev92: I have a suggestion in: https://github.com/NuGet/NuGet.Client/pull/6131#discussion_r1826355701. 
That pattern already exists with SourceRepositoryDependencyProvider for example: https://github.com/NuGet/NuGet.Client/blob/a34df25e2651ea48e40c75b0273ddbb8398639db/src/NuGet.Core/NuGet.Commands/RestoreCommand/SourceRepositoryDependencyProvider.cs, in there we're just caching the task and using your TaskResultCache.

src/NuGet.Core/NuGet.Commands/RestoreCommand/VulnerabilityInformationProvider.cs
jeffkl: If multiple threads call `GetVulnerabilityInformationAsync(CancellationToken token)`, but the method itself is just calling `GetVulnerabilityInfoAsync()` once and returning a cached value to all the other threads, which `CancellationToken` should be respected?

My thought was the creation of this class happens once during restore and at that time there's only a single `CancellationToken` for the whole restore operation.  Multiple threads are then running to restore individual projects and they call into this provider to get the vulnerability information, but this class only does the operation once. So I'm not sure if this individual method needs to accept a cancellation token, checking vulnerability information is a single operation as part of restore.

src/NuGet.Core/NuGet.Commands/RestoreCommand/VulnerabilityInformationProvider.cs
jeffkl: `lock(object)` won't respect cancellation, but `SemaphoreSlim.WaitAsync()` will.  You also cannot `await` inside of a `lock` operation.  

This code will respect multiple unique cancellation tokens:
```cs
internal sealed class Operation
{
    private readonly SemaphoreSlim _semaphore = new SemaphoreSlim(initialCount: 1, maxCount: 1);

    private object? _result;
        
    public async Task<object?> PerformOperationAsync(CancellationToken cancellationToken)
    {
        try
        {
            await _semaphore.WaitAsync(cancellationToken);
        }
        catch (OperationCanceledException)
        {
            return null;
        }

        try
        {
            if (_result != null)
            {
                return _result;
            }

            try
            {
                await Task.Delay(TimeSpan.FromSeconds(10), cancellationToken);
            }
            catch (OperationCanceledException)
            {
                return null;
            }

            _result = new object();

            return _result;
        }
        finally
        {
            _semaphore.Release();
        }
    }
}
```

nkolev92: You can do that too. It's more correct since it supports multiple cancellations.

The suggestion I had wasn't await inside a lock. It was just creating that async lazy with the correct token passed on the method. It was just awaiting the first cancellation similar to what the current iteration is doing but without breaking the public API, or passing the token to a class that has indeterminate lifetime.

jeffkl: I've reworked this to use a `SemphoreSlim` instead of an `AsyncLazy<T>`

src/NuGet.Core/NuGet.Commands/RestoreCommand/VulnerabilityInformationProvider.cs
```diff
@@ -13,38 +13,53 @@
...
-            GetVulnerabilityInfoResult? result = await _vulnerabilityInfo;
-            return result;
+            await _semaphore.WaitAsync(cancellationToken);
+
+            try
+            {
+                if (_cachedResult is not null)
+                {
+                    return _cachedResult;
+                }
```
zivkan, line 45: You could duplicate this block before the semaphore wait, so that once it's cached, there's no longer work needed to get and release the semaphore.  In a large solution restore, this early exist will happen for the majority of projects.

zivkan (dismissed):

nkolev92 (approved):

Nigusu-Allehu (approved):

jgonz120 (approved):

jeffkl:
@zivkan are you okay with this approach?  I swapped out the `AsyncLazy<T>` for a `SemaphoreSlim` since the latter supports cancellation when a thread is awaiting another thread that's doing the actual work.

zivkan (approved):
LGTM.  Thanks for the improvement following our feedback!

PR #6136, opened 2024-11-05
Add ReadmeUriTemplateResource for download README from remote sources.

jgonz120:
# Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/2918

## Description
Created a new resource type for generating the README download URI. `ReadmeUriTemplate/6.13.0` is the resource we will be looking for in the service index. We use the URL set in the service index as the template for generating the URI for downloading the README. 

Instead of directly exposing this resource type, the PackageMetadataResourceV3 class has been updated to set the ReadmeFileUrl with the URI. 

Since the README will start pulling from remote sources there may be a delay while we download the README. I've added a loading message which will be displayed in the README control. 
image

## PR Checklist

- [x] Added tests

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
@@ -60,7 +60,6 @@ namespace NuGet.PackageManagement.UI.ViewModels
         {
             if (packageMetadata != null && (!string.Equals(packageMetadata.Id, _packageMetadata?.Id) || packageMetadata.Version != _packageMetadata?.Version))
             {
-                ReadmeMarkdown = Resources.Text_Loading;
                 _packageMetadata = packageMetadata;
                 await LoadReadmeAsync(cancellationToken);
             }
@@ -73,6 +72,7 @@ namespace NuGet.PackageManagement.UI.ViewModels
 
         private async Task LoadReadmeAsync(CancellationToken cancellationToken)
         {
+            ReadmeMarkdown = Resources.Text_Loading;
             if (string.IsNullOrWhiteSpace(_packageMetadata.ReadmeFileUrl))
             {
                 ReadmeMarkdown = _canRenderLocalReadme && !string.IsNullOrWhiteSpace(_packageMetadata.PackagePath) ? Resources.Text_NoReadme : string.Empty;
@@ -91,20 +91,22 @@ namespace NuGet.PackageManagement.UI.ViewModels
             }
 
             var readme = Resources.Text_NoReadme;
-            await ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+            try
             {
-                await TaskScheduler.Default;
-                using var readmeStream = await _nugetPackageFileService.GetReadmeAsync(readmeUrl, cancellationToken);
-                if (readmeStream is null)
+                await ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
                 {
-                    return;
-                }
+                    await TaskScheduler.Default;
+                    using var readmeStream = await _nugetPackageFileService.GetReadmeAsync(readmeUrl, cancellationToken);
+                    if (readmeStream is null)
+                    {
+                        return;
+                    }
 
-                using StreamReader streamReader = new StreamReader(readmeStream);
-                readme = await streamReader.ReadToEndAsync();
-            });
-
-            if (!cancellationToken.IsCancellationRequested)
+                    using StreamReader streamReader = new StreamReader(readmeStream);
+                    readme = await streamReader.ReadToEndAsync();
+                });
+            }
+            finally
             {
                 ReadmeMarkdown = readme;
                 IsVisible = !string.IsNullOrWhiteSpace(readme);
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
@@ -18,6 +18,7 @@ namespace NuGet.PackageManagement.UI.ViewModels
         private string _rawReadme;
         private DetailedPackageMetadata _packageMetadata;
         private bool _canRenderLocalReadme;
+        private bool _isBusy;
 
         public ReadmePreviewViewModel(INuGetPackageFileService packageFileService, ItemFilter itemFilter, bool isReadmeFeatureEnabled)
         {
@@ -25,16 +26,33 @@ namespace NuGet.PackageManagement.UI.ViewModels
             _canRenderLocalReadme = CanRenderLocalReadme(itemFilter);
             _nugetPackageFileService = packageFileService;
             _errorLoadingReadme = false;
+            _isBusy = false;
             _rawReadme = string.Empty;
             _packageMetadata = null;
             Title = Resources.Label_Readme_Tab;
             IsVisible = isReadmeFeatureEnabled;
         }
 
+        public bool IsReadmeReady { get => !IsBusy && !ErrorLoadingReadme; }
+
         public bool ErrorLoadingReadme
         {
             get => _errorLoadingReadme;
-            set => SetAndRaisePropertyChanged(ref _errorLoadingReadme, value);
+            set
+            {
+                SetAndRaisePropertyChanged(ref _errorLoadingReadme, value);
+                RaisePropertyChanged(nameof(IsReadmeReady));
+            }
+        }
+
+        public bool IsBusy
+        {
+            get => _isBusy;
+            set
+            {
+                SetAndRaisePropertyChanged(ref _isBusy, value);
+                RaisePropertyChanged(nameof(IsReadmeReady));
+            }
         }
 
         public string ReadmeMarkdown
@@ -72,12 +90,13 @@ namespace NuGet.PackageManagement.UI.ViewModels
 
         private async Task LoadReadmeAsync(CancellationToken cancellationToken)
         {
-            ReadmeMarkdown = Resources.Text_Loading;
+            IsBusy = true;
             if (string.IsNullOrWhiteSpace(_packageMetadata.ReadmeFileUrl))
             {
                 ReadmeMarkdown = _canRenderLocalReadme && !string.IsNullOrWhiteSpace(_packageMetadata.PackagePath) ? Resources.Text_NoReadme : string.Empty;
                 IsVisible = !string.IsNullOrWhiteSpace(ReadmeMarkdown);
                 ErrorLoadingReadme = false;
+                IsBusy = false;
                 return;
             }
 
@@ -87,6 +106,7 @@ namespace NuGet.PackageManagement.UI.ViewModels
                 ReadmeMarkdown = string.Empty;
                 IsVisible = false;
                 ErrorLoadingReadme = false;
+                IsBusy = false;
                 return;
             }
 
@@ -111,6 +131,7 @@ namespace NuGet.PackageManagement.UI.ViewModels
                 ReadmeMarkdown = readme;
                 IsVisible = !string.IsNullOrWhiteSpace(readme);
                 ErrorLoadingReadme = false;
+                IsBusy = false;
             }
         }
     }
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
@@ -60,6 +60,7 @@ namespace NuGet.PackageManagement.UI.ViewModels
         {
             if (packageMetadata != null && (!string.Equals(packageMetadata.Id, _packageMetadata?.Id) || packageMetadata.Version != _packageMetadata?.Version))
             {
+                ReadmeMarkdown = Resources.Text_Loading;
                 _packageMetadata = packageMetadata;
                 await LoadReadmeAsync(cancellationToken);
             }
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
@@ -13,7 +13,7 @@ namespace NuGet.PackageManagement.UI.ViewModels
 {
     public sealed class ReadmePreviewViewModel : TitledPageViewModelBase
     {
-        private bool _errorLoadingReadme;
+        private bool _errorWithReadme;
         private INuGetPackageFileService _nugetPackageFileService;
         private string _rawReadme;
         private DetailedPackageMetadata _packageMetadata;
@@ -25,7 +25,7 @@ namespace NuGet.PackageManagement.UI.ViewModels
             _nugetPackageFileService = packageFileService ?? throw new ArgumentNullException(nameof(packageFileService));
             _canRenderLocalReadme = CanRenderLocalReadme(itemFilter);
             _nugetPackageFileService = packageFileService;
-            _errorLoadingReadme = false;
+            _errorWithReadme = false;
             _isBusy = false;
             _rawReadme = string.Empty;
             _packageMetadata = null;
@@ -33,14 +33,14 @@ namespace NuGet.PackageManagement.UI.ViewModels
             IsVisible = isReadmeFeatureEnabled;
         }
 
-        public bool IsReadmeReady { get => !IsBusy && !ErrorLoadingReadme; }
+        public bool IsReadmeReady { get => !IsBusy && !ErrorWithReadme; }
 
-        public bool ErrorLoadingReadme
+        public bool ErrorWithReadme
         {
-            get => _errorLoadingReadme;
+            get => _errorWithReadme;
             set
             {
-                SetAndRaisePropertyChanged(ref _errorLoadingReadme, value);
+                SetAndRaisePropertyChanged(ref _errorWithReadme, value);
                 RaisePropertyChanged(nameof(IsReadmeReady));
             }
         }
@@ -94,7 +94,7 @@ namespace NuGet.PackageManagement.UI.ViewModels
             {
                 ReadmeMarkdown = _canRenderLocalReadme && !string.IsNullOrWhiteSpace(_packageMetadata.PackagePath) ? Resources.Text_NoReadme : string.Empty;
                 IsVisible = !string.IsNullOrWhiteSpace(ReadmeMarkdown);
-                ErrorLoadingReadme = false;
+                ErrorWithReadme = false;
                 return;
             }
 
@@ -103,7 +103,7 @@ namespace NuGet.PackageManagement.UI.ViewModels
             {
                 ReadmeMarkdown = string.Empty;
                 IsVisible = false;
-                ErrorLoadingReadme = false;
+                ErrorWithReadme = false;
                 return;
             }
 
@@ -111,7 +111,7 @@ namespace NuGet.PackageManagement.UI.ViewModels
             try
             {
                 IsBusy = true;
-                ErrorLoadingReadme = false;
+                ErrorWithReadme = false;
                 await ThreadHelper.JoinableTaskFactory.RunAsync(async () =>
                 {
                     await TaskScheduler.Default;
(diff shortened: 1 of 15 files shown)


src/NuGet.Core/NuGet.Protocol/Resources/ReadmeUriTemplateResource.cs
```diff
@@ -0,0 +1,47 @@
...
+    {
+        private readonly string _uriTemplate;
+
+        public ReadmeUriTemplateResource(string uriTemplate)
+        {
+            _uriTemplate = uriTemplate;
+        }
+
+        /// <summary>
+        /// Gets a URL for reporting package abuse. The URL will not be verified to exist.
```
jeffkl: This text is wrong...

jeffkl (dismissed):

zivkan (commented):
The tech spec hasn't been approved/merged yet: https://github.com/NuGet/Home/pull/13200/files

Shouldn't we get agreement on the technical approach, before implementation?

test/NuGet.Core.Tests/NuGet.Protocol.Tests/Providers/ReadmeUriTemplateResourceProviderTests.cs
```diff
@@ -0,0 +1,114 @@
...
+            };
+            var sourceRepository = new SourceRepository(_packageSource, resourceProviders);
+
+            Tuple<bool, INuGetResource> result = await _target.TryCreate(sourceRepository, CancellationToken.None);
+
+            Assert.True(result.Item1);
+            Assert.IsType<ReadmeUriTemplateResource>(result.Item2);
+            Assert.Equal(
+                "https://unit.test/packages/mypackage/1.0.0/readme",
+                ((ReadmeUriTemplateResource)result.Item2).GetReadmeUrl("MyPackage", NuGetVersion.Parse("1.0.0")));
```
donnie-msft: nit: This feels like an Act to me, and also would be good to include `GetReadmeUrl` in the test title somewhere. eg, `TryCreate_WhenResourceExists_GetReadmeUrlIsCorrect` or `GetReadmeUrl_WhenReadmeUriTemplateResourceExists_ReturnsCorrrectUrl`

src/NuGet.Core/NuGet.Protocol/Resources/ReadmeUriTemplateResource.cs
```diff
@@ -0,0 +1,47 @@
...
+        public string GetReadmeUrl(string id, NuGetVersion version)
+        {
+            if (_uriTemplate == null)
+            {
+                return string.Empty;
+            }
+
+            var uriString = _uriTemplate
+#if NETCOREAPP
+               .Replace("{lower_id}", id.ToLowerInvariant(), StringComparison.OrdinalIgnoreCase)
```
donnie-msft: Could you help me understand the lower-casing of ID here? I don't think protocol expects these to be lowercase based on https://learn.microsoft.com/en-us/nuget/api/package-details-template-resource#url-placeholders. I might be misunderstanding that doc.

jgonz120: Here's an example of one that excepts it to be lower. https://learn.microsoft.com/en-us/nuget/api/package-base-address-resource. @joelverhagen it doesn't seem like the case is important for downloading the README from nuget.org.

donnie-msft: Look forward to hearing Joel's clarification. To restate my concern, I wouldn't want us putting something in Client that's specific to nuget.org. Since ID is globally unique, passing lowercase values in the query string sounds like a server implementation detail to me.

jgonz120: > Here's an example of one that excepts it to be lower. https://learn.microsoft.com/en-us/nuget/api/package-base-address-resource. @joelverhagen it doesn't seem like the case is important for downloading the README from nuget.org.

Hmm never mind, it does look like the NuGet.org URL is case sensitive. 
I had originally suggested we provide options for lower_id and upper_id. We don't do that for any of the other APIs though. [Package metadata](https://learn.microsoft.com/en-us/nuget/api/registration-base-url-resource) and [package content](https://learn.microsoft.com/en-us/nuget/api/package-base-address-resource) both use the lower_id.

I think passing a lower ID is likely the safest choice. I imagine for the most part servers that implement these endpoints make them case insensitive. If they are going to implement something that is case sensitive then it really only makes sense to do upper or lower, since package IDs don't follow a specific casing, https://www.nuget.org/packages/NuGet.Frameworks, for example.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/Providers/ReadmeUriTemplateResourceProviderTests.cs
jgonz120: Hmm I'm going to delete that line all together. ReadmeUriTemplateResourceTests already tests that method.

src/NuGet.Core/NuGet.Protocol/Resources/ReadmeUriTemplateResource.cs
joelverhagen: Given this is new protocol we can define it however we want. The NuGet.org implementation needs lowercase.

In other places as @jgonz120 mentioned, V3 protocol documents lower ID and lower normalized version to be used. This should be the same.

Generally speaking URLs are not case insensitive but it is also common for web services to choose to be case insensitive. This is easy to do when you have a web app under your control handling the request. It's really a server to service decision. We can say "lower case IDs before putting them in the URL" for NuGet protocol. This gives predictability for case sensitive underlying tech (such as blob storage) but of course also works for case insensitive implementations.

jeffkl (dismissed):

src/NuGet.Core/NuGet.Protocol/Resources/ReadmeUriTemplateResource.cs
jgonz120: I'm trying to think if I've ever seen a web api that's case sensitive that requires everything to be upper case. I would feel like everything is yelling at me.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
```diff
@@ -60,6 +60,7 @@ public async Task SetPackageMetadataAsync(DetailedPackageMetadata packageMetadat
         {
             if (packageMetadata != null && (!string.Equals(packageMetadata.Id, _packageMetadata?.Id) || packageMetadata.Version != _packageMetadata?.Version))
             {
+                ReadmeMarkdown = Resources.Text_Loading;
```
donnie-msft: Please include a screenshot of how this looks.

src/NuGet.Core/NuGet.Protocol/Providers/ReadmeUriTemplateResourceProvider.cs
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
```
donnie-msft, line 4:
```suggestion
#nullable enable

```

src/NuGet.Core/NuGet.Protocol/Providers/ReadmeUriTemplateResourceProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+    internal class ReadmeUriTemplateResourceProvider : ResourceProvider
+    {
+        public ReadmeUriTemplateResourceProvider()
+            : base(typeof(ReadmeUriTemplateResource),
+                  nameof(ReadmeUriTemplateResource),
+                  NuGetResourceProviderPositions.Last)
+        {
+        }
+
+        public override async Task<Tuple<bool, INuGetResource>> TryCreate(SourceRepository source, CancellationToken token)
```
donnie-msft: Missing inheritdoc comment like in other providers.

src/NuGet.Core/NuGet.Protocol/Providers/ReadmeUriTemplateResourceProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using NuGet.Protocol.Core.Types;
+
+namespace NuGet.Protocol
+{
+    internal class ReadmeUriTemplateResourceProvider : ResourceProvider
```
donnie-msft, line 14: Missing summary/remarks comment blocks like in other providers.

src/NuGet.Core/NuGet.Protocol/Resources/ReadmeUriTemplateResource.cs
```diff
@@ -0,0 +1,47 @@
...
+using NuGet.Protocol.Core.Types;
+using NuGet.Versioning;
+
+#if NETCOREAPP
+using System;
+#endif
+
+namespace NuGet.Protocol
+{
+    internal class ReadmeUriTemplateResource : INuGetResource
```
donnie-msft, line 16: Shouldn't this be part of the public API?
Perhaps this...
```suggestion
    public sealed class ReadmeUriTemplateResource : INuGetResource
```

donnie-msft (changes requested):
Looks mostly good to go, just a few minor cleanup suggestions/questions!

src/NuGet.Core/NuGet.Protocol/Resources/ReadmeUriTemplateResource.cs
jgonz120: My thought was to keep it internal while this feature was in development and switch it to public once we decide we want to commit to supporting this implementation.

donnie-msft: Ok, but I was thinking if the resource is already added to nuget.org, and we're supporting it for nuget.org, we were past that decision point already.

donnie-msft (dismissed):

src/NuGet.Core/NuGet.Protocol/Providers/ReadmeUriTemplateResourceProvider.cs
jgonz120: Looks like a bunch of the providers are missing this. Should I create a separate PR get add it to all of them?

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
jebriede: The state that indicates that the loading is happening should be set in the LoadReadmeAsync before we start the operation and unset in the finally block of the try/catch in LoadReadmeAsync. Which also brings me to the point that we probably want to have a try/finally block in LoadReadmeAsync so we can make sure to unset the loading state or indicator.

jebriede: The specific way in which a loading indicator shows up on the screen is a presentation concern. Ideally, the ViewModel would have store state on whether there is an operation in progress or not. Something like IsBusy or IsLoading and then the View would decide how to render that. That provides the flexibility of making it a text or spinner or indeterminate progress bar without the ViewModel needing to change.

src/NuGet.Core/NuGet.Protocol/Model/PackageSearchMetadata.cs
```diff
@@ -107,7 +107,7 @@ public string Owners
         public Uri ReadmeUrl { get; private set; }
 
         [JsonProperty(PropertyName = JsonProperties.ReadmeFileUrl)]
-        public string ReadmeFileUrl { get; private set; }
+        public string ReadmeFileUrl { get; internal set; }
```
jebriede: The ReadmeFileUrl should either come from the Json payload or be set, but if we look at the pattern established in this class, we either get it from the Json and have a private setter, or we have a public setter, but not both. Are we expecting this property from the json? If not, we can use the JsonIgnore attribute.

src/NuGet.Core/NuGet.Protocol/Repository.cs
```diff
@@ -53,6 +53,7 @@ public virtual IEnumerable<Lazy<INuGetResourceProvider>> GetCoreV3()
                 yield return new Lazy<INuGetResourceProvider>(() => new RegistrationResourceV3Provider());
                 yield return new Lazy<INuGetResourceProvider>(() => new SymbolPackageUpdateResourceV3Provider());
                 yield return new Lazy<INuGetResourceProvider>(() => new ReportAbuseResourceV3Provider());
+                yield return new Lazy<INuGetResourceProvider>(() => new ReadmeUriTemplateResourceProvider());
```
jebriede, line 56: Nit: I'm not certain, but for convention do we want to have V3 here? Maybe worth checking on this.

src/NuGet.Core/NuGet.Protocol/Resources/ReadmeUriTemplateResource.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        private readonly string _uriTemplate;
+
+        public ReadmeUriTemplateResource(string uriTemplate)
+        {
+            _uriTemplate = uriTemplate;
+        }
+
+        /// <summary>
+        /// Get  the URL for downloading the readme file.
```
jebriede: Remove extra space

```suggestion
        /// Get the URL for downloading the readme file.
```

src/NuGet.Core/NuGet.Protocol/Resources/ReadmeUriTemplateResource.cs
```diff
@@ -0,0 +1,50 @@
...
+                return string.Empty;
+            }
+
+            var uriString = _uriTemplate
+#if NETCOREAPP
+               .Replace("{lower_id}", id.ToLowerInvariant(), StringComparison.OrdinalIgnoreCase)
+               .Replace("{lower_version}", version.ToNormalizedString().ToLowerInvariant(), StringComparison.OrdinalIgnoreCase);
+#else
+               .Replace("{lower_id}", id.ToLowerInvariant())
+               .Replace("{lower_version}", version.ToNormalizedString().ToLowerInvariant());
```
jebriede: It would be good to make the tokens constants so we don't duplicate them and in case of future changes we make sure they're changed in only one place.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
jebriede: Additionally, if any code needs to check on whether the ViewModel is still in the loading state, it's preferable for it to check the IsBusy/IsLoading property instead of checking if the ReadmeMarkdown == Resources.Text_Loading in the appropriate language, which would also break if a valid readme actually had that text.

jgonz120: I tried implementing something with IsBusy and was getting some unexpected results with the README being rendered. I'm going to move the setting the loading message to LoadReadmeAsync, so it always sets the loading message. I'll create an issue to change how the loading message is set and start on that once this PR is merged

src/NuGet.Core/NuGet.Protocol/Resources/ReadmeUriTemplateResource.cs
jgonz120: Yea, but it'll be easier to change our mind in the future about how it's implemented if we keep it internal.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
```diff
@@ -73,6 +72,7 @@ private static bool CanRenderLocalReadme(ItemFilter filter)
 
         private async Task LoadReadmeAsync(CancellationToken cancellationToken)
         {
+            ReadmeMarkdown = Resources.Text_Loading;
```
jebriede: As mentioned in my previous review's comment, I think we should represent loading or busy state in the ViewModel and decide how that should be presented in the View.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
```diff
@@ -72,12 +90,13 @@ private static bool CanRenderLocalReadme(ItemFilter filter)
 
         private async Task LoadReadmeAsync(CancellationToken cancellationToken)
         {
-            ReadmeMarkdown = Resources.Text_Loading;
+            IsBusy = true;
```
jebriede: We shouldn't need the IsBusy flag to be set here. We can simply set it before doing our expensive operation. Until we get to the async code, all these are very quick O(1) operations. If we move this IsBusy to just before the expensive operation, we can also avoid a lot of the code that's setting IsBusy back to false if the conditions aren't met.

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
```diff
@@ -18,23 +18,41 @@ public sealed class ReadmePreviewViewModel : TitledPageViewModelBase
...
+            _isBusy = false;
             _rawReadme = string.Empty;
             _packageMetadata = null;
             Title = Resources.Label_Readme_Tab;
             IsVisible = isReadmeFeatureEnabled;
         }
 
+        public bool IsReadmeReady { get => !IsBusy && !ErrorLoadingReadme; }
+
         public bool ErrorLoadingReadme
```
jebriede: ErrorLoadingReadme should not be public and called from an outside component. The View should not be setting a state property in the ViewModel when a UI component fails to set the content correctly. This violates the responsibilities of the View and ViewModel in MVVM.

Loading the Readme and Rendering the readme are two separate concepts. The ViewModel loads the Readme, the View renders it. The initial concern is with the name of ErrorLoadingReadme in this context. This property is actually dealing with an error when the control fails to update the markdown due to an argument exception or invalid operation, not when the readme fails to load. The second concern is how the View then sets a property on the ViewModel which then is bound back to the View, for an error that occurred on the View side.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml
```diff
@@ -28,8 +30,13 @@
       x:Uid="descriptionMarkdownPreview"
       ClipToBounds="True"
       Focusable="False"
-      Visibility="{Binding Path=ErrorLoadingReadme, Mode=OneWay, Converter={StaticResource NegatedBooleanToVisibilityConverter}}"
+      Visibility="{Binding Path=IsReadmeReady, Mode=OneWay, Converter={StaticResource BooleanToVisibilityConverter}}"
```
jebriede, line 33: There are 4 states for the Readme operations and results:

1. The Readme has not been loaded yet
2. The Readme is currently loading
3. The Readme has finished loading and it updated successfully
4. The Readme has finished loading and it failed

During state 2, the Loading indicator should be shown. Based on this, I don't think it's doing the proper thing to determine if the loading indicator should be based on the computed property IsReadmeReady, which is in turn based on ErrorLoadingReadme. ErrorLoadingReadme will only be set after the loading is complete and should be cleared (set to false) at the beginning of the next loading (if we were to implement a refresh or retry, for example).

I'm not seeing the need for the IsReadmeReady property or its use here but please help me understand the scenario where it's needed. Note, I have feedback in another comment about how ErrorLoadingReadme might not be appropriate for the ViewModel considering its use is to determine if the markdown View component failed during updating.

jebriede (approved):

Nigusu-Allehu (approved):

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/ReadmePreviewViewModel.cs
jgonz120: For now I'm renaming ErrorLoadingReadme to ErrorWithReadme, so that the name better reflects it's usage. I'll look into what it would take to handle the errors properly so that the View isn't updating the ViewModel when there's an error trying to render the README.

jeffkl (approved):

PR #6138, opened 2024-11-06
[Feature] Implement Support for NuGet Authentication Plugins as .NET Tools

### LINKED ISSUE #12567: Support for credential providers deployed via .NET tools
### NuGet Product(s) Affected

NuGet.exe, MSBuild.exe, dotnet.exe, NuGet SDK

### Current Behavior

Currently, credential providers support \*.exe for .NET Framework or \*.dll for .NET Core, and are usually split between two folders under the NuGet credential provider base path. For .NET Framework the providers are discovered by searching credentialprovider*.exe and executing that program, and for .NET Core they are discovered by searching credentialprovider*.dll and executing ```dotnet <credentialprovider>.dll```. Prior to .NET Core 2.0 this split was required as .NET Core only supported platform agnostic dll's (with the side affect of excluding platform dependent credential providers not based on .NET). With the latest versions of .NET however, this split is no longer necessary and requires multiple versions to be supported and deployed. This is also reinforced by the two plugin path environment variables ```NUGET_NETFX_PLUGIN_PATHS``` and ```NUGET_NETCORE_PLUGIN_PATHS``` which need to be set per framework version.

A deployment solution for dotnet is [.NET tools](https://learn.microsoft.com/en-us/dotnet/core/tools/global-tools) which provide a seamless .NET install and management experience for NuGet packages. Using .NET tools as a deployment mechanism has been a recurring feature ask for users of the .NET ecosystem which need to authenticate with private repositories. The ideal workflow would be (for private repositories like Azure DevOps):
1. Customers have dotnet CLI tools installed
2. Run ```dotnet tool install -g Microsoft.CredentialProviders```
3. Run ```dotnet restore``` with a private endpoint and it 'just works' (e.g. credential providers from step 2 are used during credential acquisition and used to authenticate against private endpoints)

This almost works today, but for Windows .NET Framework only. The goal is to support this cross-platform for all supported .NET runtimes.

### Problems
This work for Windows and .NET Framework today with the following caveats:
1. You must use the "tool path" option when installing to get the tool in a path known by NuGet. By default .NET tools are installed to %USERPROFILE%\\.dotnet\\tools which is a location not examined by NuGet today. This can be specified explicitly, but can be error prone and doesn't take into account any environment customization:
    ```
    dotnet tool install -g Microsoft.CredentialProviders --tool-path %USERPROFILE%\.nuget\plugins\netfx\CredentialProvider.Microsoft
    ```
2. .NET tools will wrap up the .NET application and layout with a single *.exe and hide all the details of the NuGet package and associated files in a .store directory as an implementation detail. This works for .NET Framework which exclusively checks for *.exe, but not for .NET Core.

Deployment via a .NET tool doesn't work for .NET Core or cross platform for the following reasons:
1. For Windows .NET tools will always use a *.exe extension. This doesn't work for .NET Core as that requires a *.dll to be invoked from dotnet, even if the .NET tool exe would otherwise work for both versions.
2. For Linux/OSX the executable will not have any extension, but instead just have an executable with the execute bit (+x) set on it. E.g.:
    ```
    ~$ dotnet tool install -g my-dotnet-tool
    You can invoke the tool using the following command: my-dotnet-tool
    Tool 'my-dotnet-tool' (version '0.0.1') was successfully installed.

    ~$ ls -l -a ~/.dotnet/tools
    total 120
    drwxrwxrwx  3 user user   4096 Mar 25  2019 .
    drwxrwxrwx 13 user user   4096 Mar 30 17:34 ..
    drwxrwxrwx  4 user user   4096 Mar 25  2019 .store
    -rwxr-xr-x  1 user user 106744 Mar 25  2019 my-dotnet-tool
    ```
This doesn't work as NuGet is exclusively looking for *.exe or *.dll

### Desired Behavior

The following changes in behavior would be required to make the proposal work:

### 1. Support platform appropriate executables
For Windows: support *.exe for both .NET Framework and .NET Core.
For Linux/OSX: support extensionless executables (+x bit set).

### 2. Support single plugin path environment variable
In order to ship a single credential provider that works for both .NET Framework and .NET Core that doesn't conflict with the existing semantics of the ```NUGET_NETFX_PLUGIN_PATHS``` and ```NUGET_NETCORE_PLUGIN_PATHS``` environment variables, a new environment variable should be introduced that supports the new semantics for # 1 above. It's possible that ```NUGET_PLUGIN_PATHS``` can be repurposed here, but that currently has lower precedence that the framework versions.

### 3. Support the .NET tool default installation location for credential providers
The default .NET tool installation directory should be included when discovering credential providers. This will simplify the installation and should reduce cases where tools are installed to a location that NuGet does not consider, reducing troubleshooting and support cases.

### Additional Context

The overall context and vision for the credential providers that support Azure DevOps (nuget, npm, python, etc.) is that they should be supported and easily installable in the ecosystems they are used in. For .NET using the dotnet CLI that means installable using dotnet tool install, for npm that means npm install, etc. Currently, various installation scripts are required, per platform, which already assume a set of tools and configuration about the users environment (PowerShell/pwsh, bash/sh, etc.).

### LINKED ISSUE #13742: [Epic] Support for credential providers deployed via .NET tools
* [x] [NuGet discovers credentials providers installed using .NET tools](https://github.com/NuGet/Home/issues/13740)
* [x] [NuGet executes credentials providers installed using .NET tools](https://github.com/NuGet/Home/issues/13741)

Nigusu-Allehu:
# Bug

Fixes: https://github.com/NuGet/Home/issues/12567
Fixes: https://github.com/NuGet/Home/issues/13742

## Description
This PR implements support for deploying and discovering NuGet authentication plugins as .NET tools, as outlined in [ the design](https://github.com/NuGet/Home/pull/13242). With this change, plugin authors can package their tools as global .NET tools by naming them with the prefix nuget-plugin-*, enabling NuGet to discover and execute them by scanning the PATH. On Windows, .exe and .bat extensions are recognized, while Linux/Mac platforms identify plugins via the executable bit. 

This is a feature branch PR.
Here are the individual commits into the feature branch
https://github.com/NuGet/NuGet.Client/pull/5990
https://github.com/NuGet/NuGet.Client/pull/6113

## PR Checklist

- [x] Added tests
- [x] Link to an issue or pull request to update docs if this PR changes settings, environment variables, new feature, etc. https://github.com/NuGet/Home/issues/13858


Nigusu-Allehu:
I will address the two comments regarding improving `PluginDiscovererTests` in https://github.com/NuGet/NuGet.Client/pull/5990, in a follow up PR

jeffkl (dismissed):

jeffkl (approved):

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -117,26 +179,180 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+        }
+
+        private static List<PluginFile> GetNetToolsPluginsInDirectory(string directoryPath)
+        {
+            var pluginFiles = new List<PluginFile>();
+
+            if (Directory.Exists(directoryPath))
+            {
+                var directoryInfo = new DirectoryInfo(directoryPath);
+                var files = directoryInfo.GetFiles("nuget-plugin-*");
```
dtivel, line 268: I believe the case sensitivity of this match will be platform-dependent, meaning on Windows it will be commonly ([not guaranteed](https://devblogs.microsoft.com/commandline/per-directory-case-sensitivity-and-wsl/)) case insensitive, but on Linux it will be case sensitive.

However, the file name comparison [here](https://github.com/NuGet/NuGet.Client/pull/6138/files#diff-cf421e9b689e2c0e46c47e27365571384c9fd9c7eae6522a36c26f5bd40312a4R206) is always case-insensitive.

Can you explain your reasoning?

dtivel, line 268: `DirectoryInfo.GetFiles(...)` can throw.  Is that OK?

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
```diff
@@ -117,26 +179,180 @@ private List<PluginFile> GetPluginFiles(CancellationToken cancellationToken)
...
+        internal static bool IsExecutable(FileInfo fileInfo)
+        {
+#pragma warning disable CA1031 // Do not catch general exception types
+            try
+            {
+                string output;
+                using (var process = new System.Diagnostics.Process())
+                {
+                    // Use a shell command to check if the file is executable
+                    process.StartInfo.FileName = "/bin/bash";
```
dtivel, line 328: This code could run on Alpine Linux, where [/bin/bash does not exist by default](https://wiki.alpinelinux.org/wiki/How_to_get_regular_stuff_working#Bash_shell).

Is there an even more portable way of checking?  I don't know, but maybe `ls -l <path>`?

src/NuGet.Core/NuGet.Protocol/Plugins/PluginManager.cs
```diff
@@ -213,12 +212,14 @@ private async Task<Tuple<bool, PluginCreationResult>> TryCreatePluginAsync(
...
                     {
-                        var plugin = await _pluginFactory.GetOrCreateAsync(
-                            result.PluginFile.Path,
-                            PluginConstants.PluginArguments,
-                            new RequestHandlers(),
-                            _connectionOptions,
-                            cancellationToken);
+                        IPlugin plugin;
+
+                        plugin = await _pluginFactory.GetOrCreateAsync(
```
dtivel, line 217:
```suggestion
                        IPlugin plugin = await _pluginFactory.GetOrCreateAsync(
```

test/NuGet.Core.Tests/NuGet.Protocol.Tests/Plugins/PluginFactoryTests.cs
```diff
@@ -135,6 +149,34 @@ public async Task GetOrCreateAsync_ThrowsIfDisposed()
...
+        echo File executed > ""{outputPath}""
+    ";
+
+            File.WriteAllText(pluginPath, batFileContent);
+
+            var args = PluginConstants.PluginArguments;
+            var reqHandler = new RequestHandlers();
+            var options = ConnectionOptions.CreateDefault();
+
+            var pluginFactory = new PluginFactory(Timeout.InfiniteTimeSpan);
```
dtivel, line 170: This is disposable and should be disposed.

test/NuGet.Core.Tests/NuGet.Credentials.Test/PluginManagerMock.cs
```diff
@@ -66,9 +66,21 @@ internal TestExpectation(
 
     internal sealed class PluginManagerMock : IDisposable
     {
+        public static bool IsDesktop
```
dtivel, line 69: I think all of these `IsDesktop` properties would be better off named `RequiresDotNetHost`, since that's how it's used.  `IsDesktop` is a little more difficult to understand than `RequiresDotNetHost`.  It would simplify call sites too.  Instead of `requiresDotnetHost: !IsDesktop`, you could simply pass `RequiresDotNetHost`.

Your call though.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/Plugins/PluginDiscovererTests.cs
```diff
@@ -20,15 +23,27 @@ public class PluginDiscovererTests
...
+
+            Exception exception = Record.Exception(() =>
             {
-            }
+                using (new PluginDiscoverer())
+                {
+                }
+            });
+
+            Assert.Null(exception);
```
dtivel, line 37: Why not simply?

```C#
   using (new PluginDiscoverer())
   {
   }
```

An exception would fail the test.

src/NuGet.Core/NuGet.Protocol/Plugins/IPluginFactory.cs
```diff
@@ -11,20 +11,20 @@ namespace NuGet.Protocol.Plugins
     /// <summary>
     /// A plugin factory.
     /// </summary>
-    public interface IPluginFactory : IDisposable
+    internal interface IPluginFactory : IDisposable
```
dtivel, line 14: Can you explain this change?

jgonz120 (approved):

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: That was also a point raised here https://github.com/NuGet/NuGet.Client/pull/5990#discussion_r1796393325. It will be caught at a higher level

test/NuGet.Core.Tests/NuGet.Protocol.Tests/Plugins/PluginDiscovererTests.cs
Nigusu-Allehu: https://github.com/NuGet/NuGet.Client/pull/5990#discussion_r1809632533

src/NuGet.Core/NuGet.Protocol/Plugins/IPluginFactory.cs
Nigusu-Allehu: We had to update the API for IPluginFactory so that it is able to create plugins for dotnet tools plugins. Since we needed to break the API, I used it as an opportunity to un-expose the interface. The reason for making it internal is that the interface only has one implementation and there is no need for it to be a public API https://github.com/NuGet/NuGet.Client/pull/6113#discussion_r1830480428

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: You are right, that's inconsistent. I will address this in a follow up PR.

Nigusu-Allehu:
Thank you for the review @dtivel! I will address the comments in a follow up PR.

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
baronfel: On unix systems, `FileInfo` (and DirectoryInfo) has a `UnixFileMode` property that is a flags enum with the typical unix properties. You can use HasFlag or your favorite enum comparisons to see if a given file has any executable flags:

```fsharp
$ dotnet fsi

Microsoft (R) F# Interactive version 12.8.102.0 for F# 8.0
Copyright (c) Microsoft Corporation. All Rights Reserved.

For help type #help;;

> open System.IO;;
> let f = new FileInfo("/usr/bin/curl");;
val f: FileInfo = /usr/bin/curl

> f.UnixFileMode;;
val it: UnixFileMode =
  OtherExecute, OtherRead, GroupExecute, GroupRead, UserExecute, UserWrite, UserRead
```

zivkan: This code is in a `#if !NET8_0_OR_GREATER` code block because the APIs that @baronfel mentioned are not available before .NET 8.  When @Nigusu-Allehu created the original PR to add this code to the feature branch (this PR is merging the feature branch into dev) I asked him to use the .NET 8 APIs already, which is being done on line 303.

I also suggested to @Nigusu-Allehu that we could consider skipping this PATH scanning for plugins for .NET Framework on Linux and Mac, since that only means Mono, [which we don't officially support](https://github.com/NuGet/docs.microsoft.com-nuget/pull/3335), but I don't remember any replies to that comment, and clearly it wasn't actioned either.

But I think it's also ok to keep here, because I really can't imagine anyone using Alpine wanting to install Mono and use NuGet.exe. And even if they do, the whole thing is wrapped in a try-catch block.  The plugin won't be discovered, but it won't break restore either (apart from not being able to authenticate if a nuget-plugin-* app is installed)

src/NuGet.Core/NuGet.Protocol/Plugins/PluginDiscoverer.cs
Nigusu-Allehu: https://github.com/NuGet/NuGet.Client/pull/6255

PR #6139, opened 2024-11-07
Do not post fault telemetry for cancellation exceptions

jeffkl:
# Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/3013

## Description
Our code is posting fault telemetry for cancellations and it should not be.  The `PackageItemViewModel` was assuming that HTTP timeouts were throwing a `TaskCanceledException` but that is not the case.  In `TimeoutUtility` we detect a timeout and throw a `TimeoutException` instead:

https://github.com/NuGet/NuGet.Client/blob/38729596a6549df3cd1eed37b93d51df1bf62bff/src/NuGet.Core/NuGet.Protocol/Utility/TimeoutUtility.cs#L46

This change updates the view model based on my findings and blocks any `OperationCancelledException` from posting telemetry.  Any code in the future where we want to post fault telemetry but its not really a cancellation should wrap the exception as something else like a `TimeoutException`.

## PR Checklist

- [ ] Added tests

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
@@ -814,13 +814,13 @@ namespace NuGet.PackageManagement.UI
                 LatestVersion = result;
                 Status = GetPackageStatus(LatestVersion, InstalledVersion, AutoReferenced);
             }
-            catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+            catch (OperationCanceledException)
             {
                 // UI requested cancellation
             }
-            catch (TaskCanceledException)
+            catch (TimeoutException)
             {
-                // HttpClient throws TaskCanceledExceptions for HTTP timeouts
+                // Our code throws a TimeoutException for HTTP timeouts
                 try
                 {
                     await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(cancellationToken);
@@ -851,13 +851,13 @@ namespace NuGet.PackageManagement.UI
 
                 SetVulnerabilityMaxSeverity(identity.Version, packageMetadata?.Vulnerabilities?.FirstOrDefault()?.Severity ?? -1);
             }
-            catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+            catch (OperationCanceledException)
             {
                 // UI requested cancellation.
             }
-            catch (TaskCanceledException)
+            catch (TimeoutException)
             {
-                // HttpClient throws TaskCanceledExceptions for HTTP timeouts
+                // Our code throws a TimeoutException for HTTP timeouts
                 try
                 {
                     await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(cancellationToken);
@@ -882,13 +882,13 @@ namespace NuGet.PackageManagement.UI
                     await UpdatePackageMaxVulnerabilityAsync(identity, cancellationToken);
                 }
             }
-            catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+            catch (OperationCanceledException)
             {
                 // UI requested cancellation.
             }
-            catch (TaskCanceledException)
+            catch (TimeoutException)
             {
-                // HttpClient throws TaskCanceledExceptions for HTTP timeouts
+                // Our code throws a TimeoutException for HTTP timeouts
                 try
                 {
                     await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(cancellationToken);
--- a/src/NuGet.Clients/NuGet.VisualStudio.Common/Telemetry/TelemetryUtility.cs
+++ b/src/NuGet.Clients/NuGet.VisualStudio.Common/Telemetry/TelemetryUtility.cs
@@ -34,6 +34,13 @@ namespace NuGet.VisualStudio.Telemetry
                 throw new ArgumentNullException(nameof(e));
             }
 
+            if (e is OperationCanceledException)
+            {
+                // Don't post telemetry for cancellation exceptions since these should only happen for user initiated actions.
+                // Any other cancellation exceptions like a timeout should be a TimeoutException instead.
+                return;
+            }
+
             var caller = $"{callerClassName}.{callerMemberName}";
             var description = $"{e.GetType().Name} - {e.Message}";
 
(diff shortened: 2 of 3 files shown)


dtivel (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
```diff
@@ -814,7 +814,7 @@ private async System.Threading.Tasks.Task ReloadPackageVersionsAsync()
                 LatestVersion = result;
                 Status = GetPackageStatus(LatestVersion, InstalledVersion, AutoReferenced);
             }
-            catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+            catch (OperationCanceledException e) when (e.CancellationToken.IsCancellationRequested || cancellationToken.IsCancellationRequested)
```
donnie-msft: Isn't `e.CancellationToken.IsCancellationRequested` always true? How can it be false?
Would I have to `throw new CancellationException(tokenWhichIsNotCancelled)` ?  Not understanding how this would happen.

Should the `when` just be removed? What cases are we NOT handling cancellation?

jeffkl: I think in most cases it is, but when we're swapping tokens it seemed like it wouldn't always be the case.

There are two cancellations here, user cancellation and timeout.  I guess we still want to know when a timeout occurred but don't care about user cancellations.  This `catch` is trying to ignore the second one there.  From telemetry, the majority of reported faults are coming from the Semaphore throwing due to cancellation

image

Under the debugger, this change did the trick.

zivkan: `e.CancellationToken` will always be the cancellation token that triggered the cancellation. So, when it's an HTTP timeout, not NuGet requesting cancellation on its own cancellation token source, then `e.CancellationToken` will be HttpClient's internal cancellation token.  So, I agree with Donnie, this is effectively `when (true)`, or a catch without a `when` clause

src/NuGet.Core/NuGet.Protocol/Providers/ServiceIndexResourceV3Provider.cs
```diff
@@ -62,15 +62,12 @@ public override async Task<Tuple<bool, INuGetResource>> TryCreate(SourceReposito
...
-                    // Track if the semaphore needs to be released
-                    var release = false;
-                    try
-                    {
-                        await _semaphore.WaitAsync(token);
-                        release = true;
+                    await _semaphore.WaitAsync(token);
 
-                        token.ThrowIfCancellationRequested();
+                    token.ThrowIfCancellationRequested();
```
zivkan: I don't feel like this `token.ThrowIfCancellationRequested` is useful right after an async operation that uses the same token. I mean, it's probably not computationally expensive, so I doubt that we can measure any harm in checking the token, but if the cancellation was requested, the semaphore would have thrown.

src/NuGet.Core/NuGet.Protocol/Providers/ServiceIndexResourceV3Provider.cs
```diff
@@ -131,6 +125,8 @@ private async Task<ServiceIndexResourceV3> GetServiceIndexResourceV3(
 
             for (var retry = 1; retry <= maxRetries; retry++)
             {
+                token.ThrowIfCancellationRequested();
```
zivkan: There are only two methods that might take some time, `client.GetAsync` and `Task.Delay`, but the cancellation token is already passed in both places.  So, I don't imagine that this `token.ThrowIf..` will ever prevent save any significant CPU time

src/NuGet.Clients/NuGet.PackageManagement.UI/ViewModels/PackageItemViewModel.cs
zivkan: Also, the point of line 800 is to capture the `_cancellationTokenSource.Token` before `_cancellationTokenSource` gets swapped (it's done before calling `GetVersionsAsync`), so I would expect this to always be the pre-swapped token.  Is that not working as expected?

Since `GetVersionsAsync` doesn't take a `CancellationToken` (from memory it does something weird with a Lazy, or something), maybe there's some kind of timing issue with this method using a different `_cancellationTokenSource` instance than `GetVersionsAsync`?   

Maybe we can change `GetVersionsAsync` to do something similar for what you did with the vulnerability info resource provider recently, so that we make sure `GetVersionsAsync` uses the same cancellation token.  But in this case we have to be more careful, to make sure there's no timing bug in an HTTP response coming in just at the same time as a new search query causing the wrong results to be displayed.

jgonz120: When does the swap happen in the VM? It looks like the VM creates its own source on construction and the only time it's cancelled is when the object is disposed.

jeffkl: I've reworked all of this based on my findings while debugging the code.  Our code does not throw an `OperationCanceledException` and instead wraps this outcome as a `TimeoutException`.

donnie-msft (approved):

dtivel (approved):

nkolev92 (approved):

src/NuGet.Clients/NuGet.VisualStudio.Common/Telemetry/TelemetryUtility.cs
```diff
@@ -34,6 +34,13 @@ public static async Task PostFaultAsync(Exception e, string callerClassName, [Ca
                 throw new ArgumentNullException(nameof(e));
             }
 
+            if (e is OperationCanceledException)
+            {
+                // Don't post telemetry for cancellation exceptions since these should only happen for user initiated actions.
+                // Any other cancellation exceptions like a timeout should be a TimeoutException instead.
```
zivkan, line 40: This is where I thought it would be most valuable to write that NuGet rethrows HttpClient timeouts as TimeoutException, because someone who sees this code, but not `PackageItemViewModel`, might assume this is buggy, since HttpClient is "well known" to use OperationCancelledException for timeouts.

I think the second sentence is trying to address that, but IMO it's not very clear. To me, it reads like "any timeout NuGet implements should throw TimeoutException", rather than acknowledging HttpClient

zivkan (approved):

Nigusu-Allehu (approved):

jeffkl:
I had to rebase, re-requesting reviews...

Nigusu-Allehu (approved):

zivkan (approved):

PR #6142, opened 2024-11-12
Add support for PrunePackageReference in the new resolver

### LINKED ISSUE #7344: Support "Supplied by Platform" scenario in restore
With .NET Core 3.0, we will be using [targeting packs](https://github.com/dotnet/cli/issues/10006) to deliver reference assemblies for the framework.  These targeting packs will not be represented as NuGet packages in the package graph in the way that Microsoft.NETCore.App, Microsoft.AspNetCore.App, and NETStandard.Library currently are.

When a .NET Core 3 project depends on a package that has a dependency on one of these older "platform" packages, we want that platform package and its dependencies to be ignored in the NuGet graph, as those APIs are supplied by the platform via other mechanisms (targeting packs at build time and runtime packs for self-contained deployments).

We've thought of two possible ways to exclude the platform packages and dependencies from the graph.  This issue is for one of them.  The other one is #7343 

To exclude these packages, we could pass an additional list of packages to the restore operation that should be ignored in the graph walk.  We might call these "Supplied by Platform" packages.  When walking the graph, if the restore operation encountered a dependency on any of these package (with a version less than or equal to the version marked as supplied by the platform), then it would ignore that dependency (and would not walk its dependencies further).

Spec: https://github.com/NuGet/Home/blob/dev/accepted/2024/prune-package-reference.md

Project-System changes: https://github.com/dotnet/project-system/pull/9610

nkolev92:
# Bug

Fixes: https://github.com/NuGet/Home/issues/7344

## Description

For reference, the design is at https://github.com/NuGet/Home/blob/dev/accepted/2024/prune-package-reference.md
This PR implements the full scope of the pruning in the new resolver. 
Primary functions are *not downloading* and skipping packages to be pruned, and *not* representing them as dependencies in the targets section of the framework they were skipped for.
This PR *does not* add a list of pruned packages anywhere, and that's intentional at this point. The design lists that as a future possibility too. 

Some pointers: 

- Meat of the changes is in `DependecyGraphResolver`. As we walk the graph, if a graph item is eligible (package and lower min version than the one specified in the PrunePackageReference, we skip it). We do that by tracking the indices of the packages we skip. This is the safe way, since it preserves the index calculation for the items in the graph. 
At flattening time, we then *skip* those dependencies and we "recreate" the RemoteResolveResult *dependencies* section to ensure the assets file contains the correct information.
- LockFileUtils.CreateLockFileTargetLibrary now contains the dependencies count in the hashing to avoid issues where things were pruned in one framework but not in the other.
- [SdkAnalysisLevelMinimums.cs](https://github.com/NuGet/NuGet.Client/pull/6142/files#diff-da63ad102af316a0c509f299a34f7ee7c5ae6f0b6fd28cda0f9e158a0ddc2d51) adds warnings for project pruning and direct package reference pruning only in .NET 10 SDK Analysis level scenarios. The feature is still disabled by default for all frameworks anyways. 
- The tests [RestoreCommandTests.cs](https://github.com/NuGet/NuGet.Client/pull/6142/files#diff-5e6f2fe813945cf26d2d437dc5bccfa9c63f632ee21285f11f678c7feb528d4c) are exhaustive and they contain all scenarios called out in the spec & combinations of pruning normal and runtime dependencies, across different or same frameworks.

## PR Checklist

- [x] Added tests
- [x] Link to an issue or pull request to update docs if this PR changes settings, environment variables, new feature, etc. - https://github.com/NuGet/Home/issues/14002

### DIFF
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
@@ -91,7 +91,7 @@ namespace NuGet.Commands
                     hasInstallBeenCalledAlready = true;
                 }
 
-                TargetFrameworkInformation? projectTargetFramework = _request.Project.GetTargetFramework(pair.Framework);
+                TargetFrameworkInformation projectTargetFramework = _request.Project.GetTargetFramework(pair.Framework)!;
 
                 var unresolvedPackages = new HashSet<LibraryRange>();
 
@@ -104,7 +104,7 @@ namespace NuGet.Commands
                     // This is guaranteed to be computed before any graph with a RID, so we can assume this will return a value.
 
                     // PCL Projects with Supports have a runtime graph but no matching framework.
-                    var runtimeGraphPath = projectTargetFramework?.RuntimeIdentifierGraphPath;
+                    var runtimeGraphPath = projectTargetFramework.RuntimeIdentifierGraphPath;
 
                     RuntimeGraph? projectProviderRuntimeGraph = default;
                     if (runtimeGraphPath != null)
@@ -635,9 +635,20 @@ namespace NuGet.Commands
                         suppressions = currentSuppressions;
                     }
 
+                    List<int>? prunedPackageIndices = null;
                     for (int i = 0; i < refItemResult.Item.Data.Dependencies.Count; i++)
                     {
                         LibraryDependency dep = refItemResult.Item.Data.Dependencies[i];
+                        bool isPackage = dep.LibraryRange.TypeConstraintAllows(LibraryDependencyTarget.Package);
+                        bool isDirectPackageReferenceFromRootProject = (currentRefRangeIndex == rootProjectRefItem.LibraryRangeIndex) && isPackage;
+
+                        if (ShouldPrunePackage(projectTargetFramework!, refItemResult, dep, isPackage, isDirectPackageReferenceFromRootProject))
+                        {
+                            prunedPackageIndices ??= [];
+                            prunedPackageIndices.Add(i);
+                            continue;
+                        }
+
                         LibraryDependencyIndex depIndex = refItemResult.GetDependencyIndexForDependency(i);
 
                         // Skip this node if the VersionRange is null or if its not transitively pinned and PrivateAssets=All
@@ -647,9 +658,6 @@ namespace NuGet.Commands
                         }
 
                         VersionRange? pinnedVersionRange = null;
-
-                        bool isPackage = dep.LibraryRange.TypeConstraintAllows(LibraryDependencyTarget.Package);
-                        bool isDirectPackageReferenceFromRootProject = (currentRefRangeIndex == rootProjectRefItem.LibraryRangeIndex) && isPackage;
 
                         if (!isDirectPackageReferenceFromRootProject && directPackageReferences?.Contains(depIndex) == true)
                         {
@@ -752,6 +760,14 @@ namespace NuGet.Commands
                         {
                             foreach (var dep in runtimeDependencies)
                             {
+                                if (ShouldPrunePackage(projectTargetFramework!, refItemResult, dep, isPackage: true, isDirectPackageReferenceFromRootProject: false))
+                                {
+                                    prunedPackageIndices ??= [];
+                                    prunedPackageIndices.Add(runtimeDependencyIndex);
+                                    runtimeDependencyIndex++;
+                                    continue;
+                                }
+
                                 DependencyGraphItem runtimeDependencyGraphItem = new()
                                 {
                                     LibraryDependency = dep,
@@ -785,6 +801,12 @@ namespace NuGet.Commands
                                 runtimeDependencyIndex++;
                             }
                         }
+                    }
+
+                    // If the latest item was chosen, keep track of the pruned dependency indices.
+                    if (chosenResolvedItems.TryGetValue(currentRefDependencyIndex, out ResolvedDependencyGraphItem? resolvedGraphItem))
+                    {
+                        resolvedGraphItem.PrunedDependencyIndices = prunedPackageIndices;
                     }
                 }
 
@@ -829,18 +851,20 @@ namespace NuGet.Commands
                     LibraryRangeIndex[] pathToChosenRef = foundItem.Path;
                     bool directPackageReferenceFromRootProject = foundItem.IsDirectPackageReferenceFromRootProject;
                     List<HashSet<LibraryDependencyIndex>> chosenSuppressions = foundItem.Suppressions;
-
                     if (findLibraryEntryCache.TryGetValue(chosenRefRangeIndex, out Task<FindLibraryEntryResult>? nodeTask))
                     {
                         FindLibraryEntryResult node = await nodeTask;
-
-                        flattenedGraphItems.Add(node.Item);
 
                         for (int i = 0; i < node.Item.Data.Dependencies.Count; i++)
                         {
                             var dep = node.Item.Data.Dependencies[i];
 
                             if (dep.LibraryRange.VersionRange == null)
+                            {
+                                continue;
+                            }
+
+                            if (foundItem.PrunedDependencyIndices != null && foundItem.PrunedDependencyIndices.Contains(i))
                             {
                                 continue;
                             }
@@ -1034,6 +1058,31 @@ namespace NuGet.Commands
                                 range: newGraphNode.Key.VersionRange,
                                 child: newGraphNode.Item.Key));
                         }
+
+                        if (foundItem.PrunedDependencyIndices != null && foundItem.PrunedDependencyIndices!.Count > 0)
+                        {
+                            int dependencyCount = node.Item.Data.Dependencies.Count - foundItem.PrunedDependencyIndices.Count;
+
+                            List<LibraryDependency> dependencies = dependencyCount > 0 ? new(dependencyCount) : [];
+
+                            for (int i = 0; dependencyCount > 0 && i < node.Item.Data.Dependencies.Count; i++)
+                            {
+                                if (!foundItem.PrunedDependencyIndices.Contains(i))
+                                {
+                                    dependencies.Add(node.Item.Data.Dependencies[i]);
+                                }
+                            }
+
+                            RemoteResolveResult remoteResolveResult = new RemoteResolveResult()
+                            {
+                                Match = node.Item.Data.Match,
+                                Dependencies = dependencies,
+                            };
+
+                            node.Item.Data = remoteResolveResult;
+                        }
+
+                        flattenedGraphItems.Add(node.Item);
                     }
                 }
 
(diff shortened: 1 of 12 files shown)


src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
```diff
@@ -845,6 +864,11 @@ async static (state) =>
                                 continue;
                             }
 
+                            if (foundItem.PrunedDependencyIndices != null && foundItem.PrunedDependencyIndices.Contains(i))
```
jeffkl:
```suggestion
                            if (foundItem.PrunedDependencyIndices?.Contains(i))
```

src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
```diff
@@ -1034,6 +1058,31 @@ async static (state) =>
                                 range: newGraphNode.Key.VersionRange,
                                 child: newGraphNode.Item.Key));
                         }
+
+                        if (foundItem.PrunedDependencyIndices != null && foundItem.PrunedDependencyIndices!.Count > 0)
```
jeffkl:
```suggestion
                        if (foundItem.PrunedDependencyIndices?.Count > 0)
```

src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
```diff
@@ -1295,6 +1380,8 @@ private class ResolvedDependencyGraphItem
             public required LibraryRangeIndex[] Path { get; set; }
 
             public required List<HashSet<LibraryDependencyIndex>> Suppressions { get; set; }
+
+            public List<int>? PrunedDependencyIndices { get; set; }
```
jeffkl:
```suggestion
            public HashSet<VersionRangeIndex>? PrunedDependencies { get; set; }
```

Let's talk about how maybe using `VersionRangeIndex` will be more performant.

src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
```diff
@@ -91,7 +91,7 @@ public async Task<ValueTuple<bool, List<RestoreTargetGraph>, RuntimeGraph>> Reso
                     hasInstallBeenCalledAlready = true;
                 }
 
-                TargetFrameworkInformation? projectTargetFramework = _request.Project.GetTargetFramework(pair.Framework);
+                TargetFrameworkInformation projectTargetFramework = _request.Project.GetTargetFramework(pair.Framework)!;
```
jeffkl: Let's revert this but clean it up in a future refactoring where we don't need to call `GetTargetFramework()` because it will just be part of the `FrameworkRuntimePair`

src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
```diff
@@ -1204,6 +1253,42 @@ async static (state) =>
...
 
+        private bool ShouldPrunePackage(TargetFrameworkInformation projectTargetFramework, FindLibraryEntryResult refItemResult, LibraryDependency dep, bool isPackage, bool isDirectPackageReferenceFromRootProject)
+        {
+            if (projectTargetFramework!.PackagesToPrune.TryGetValue(dep.Name, out PrunePackageReference? prunableVersion))
+            {
+                if (dep.LibraryRange!.VersionRange!.Satisfies(prunableVersion.VersionRange!.MaxVersion!))
+                {
+                    if (!isPackage)
+                    {
+                        if (SdkAnalysisLevelMinimums.IsEnabled(
```
jeffkl, line 1291: I think this whole feature should be off by default but when enabled, we should consider a mode like "strict" where these errors are logged.  I'm thinking of the scenario someday when I update to the latest .NET SDK and now I can't restore.  If we make this on by default but just silently ignore misconfigurations, that would be a fine experience.  If I want to opt into getting these errors, that would at least be a change I made.  I think good documentation would help users know that they can opt into the more strict mode.

src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
nkolev92: I think the comment is in the wrong location, but I understand the feedback, so I'll resolve this in the next commit.

src/NuGet.Core/NuGet.Commands/Utility/SdkAnalysisLevelMinimums.cs
```diff
@@ -15,6 +15,11 @@ internal static class SdkAnalysisLevelMinimums
         /// </summary>
         internal static readonly NuGetVersion HttpErrorSdkAnalysisLevelMinimumValue = new("9.0.100");
 
+        /// <summary>
+        /// Minimum SDK Analysis Level required for warning for packages and projects that cannot be pruned.
+        /// </summary>
+        internal static readonly NuGetVersion PruningWarnings = new("10.0.100");
```
zivkan, line 21: Given my experience with NuGetAuditMode, I'm now a huge opponent of SdkAnalysisLevel.  It doesn't work the way a lot of customers expect it to, and it doesn't work in the way that our division's leadership want us to provide backwards compatibility to customers. I think SdkAnalysisLevel doesn't work and should be considered a failed experiment we no longer use.

Unfortunately, I also don't have a suggestion on how we'll ever turn on new features by default. It's something we're going to have to have discussions about (and when we have a proposal, take it to leadership) early next year.

src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
nkolev92: Discussed briefly offline. 

I'll add this to the backlog for PrunePackageReference and work on this as we progress towards the official roll-out: 
https://github.com/NuGet/Client.Engineering/issues/3111

src/NuGet.Core/NuGet.Commands/Utility/SdkAnalysisLevelMinimums.cs
nkolev92: I don't know if we should give up on SDKAnalysisLevel after a single release, but we should certainly be looking for something that allows to enable defaults for *every* scenario.

Your feedback is similar to Jeff's and I have created a follow-up that I'll make part of the backlog of PrunePackageReference and something we want to figure out before the official roll-out enabled by default in whatever scenario we deem we want it: https://github.com/NuGet/Client.Engineering/issues/3111

src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
```diff
@@ -1204,6 +1256,60 @@ async static (state) =>
...
+                {
+                    LibraryDependencyIndex depIndex = libraryDependencyInterningTable.Intern(item.Value);
+                    prunedPackageVersions[depIndex] = item.Value.VersionRange;
+                }
+            }
+
+            return prunedPackageVersions;
+        }
+
+        private bool ShouldPrunePackage(IReadOnlyDictionary<LibraryDependencyIndex, VersionRange>? packagesToPrune, FindLibraryEntryResult refItemResult, LibraryDependency dep, LibraryDependencyIndex libraryDependencyIndex, bool isPackage, bool isDirectPackageReferenceFromRootProject)
```
zivkan: When a method has this many parameters, can we put them on per line?  The line is so long it's hard to read.

src/NuGet.Core/NuGet.Commands/RestoreCommand/LockFileBuilderCache.cs
```diff
@@ -29,7 +29,7 @@ public class LockFileBuilderCache
         private readonly ConcurrentDictionary<CriteriaKey, List<(List<SelectionCriteria>, bool)>> _criteriaSets =
             new();
 
-        private readonly ConcurrentDictionary<(CriteriaKey, string path, string aliases, LibraryIncludeFlags), Lazy<(LockFileTargetLibrary, bool)>> _lockFileTargetLibraryCache =
+        private readonly ConcurrentDictionary<(CriteriaKey, string path, string aliases, LibraryIncludeFlags, int), Lazy<(LockFileTargetLibrary, bool)>> _lockFileTargetLibraryCache =
```
zivkan: The two strings in the value tuple, whose purpose can't be guessed by type alone, have names. I think the new `int` should as well.

src/NuGet.Core/NuGet.Commands/Strings.resx
```diff
@@ -1127,4 +1127,16 @@ NuGet requires HTTPS sources. Refer to https://aka.ms/nuget-https-everywhere for
...
     <comment>{0} is the source name</comment>
   </data>
-</root>
\ No newline at end of file
+  <data name="RestoreDebugPruningPackageReference" xml:space="preserve">
+    <value>Pruning the package '{0}' as a dependency of '{1}'. The maximum prunable version is '{2}'</value>
+    <comment>0 - package id and version, 1 - version</comment>
+  </data>
+  <data name="Error_RestorePruningDirectPackageReference" xml:space="preserve">
+    <value>A direct PackageReference cannot be pruned, {0}. Consider removing this package from your dependencies, as it is likely unnecessary.</value>
```
zivkan:
```suggestion
    <value>PackageReference {0} will not be pruned. Consider removing this package from your dependencies, as it is likely unnecessary.</value>
```

src/NuGet.Core/NuGet.Common/Errors/NuGetLogCode.cs
```diff
@@ -297,6 +297,16 @@ public enum NuGetLogCode
         /// </summary>
         NU1509 = 1509,
 
+        /// <summary>
+        /// Direct reference to a package that cannot be pruned.
```
zivkan:
```suggestion
        /// Direct reference to a package that will not be pruned.
```

src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
```diff
@@ -1204,6 +1256,60 @@ async static (state) =>
...
+                    {
+                        if (SdkAnalysisLevelMinimums.IsEnabled(
+                            _request.Project!.RestoreMetadata!.SdkAnalysisLevel,
+                            _request.Project.RestoreMetadata.UsingMicrosoftNETSdk,
+                            SdkAnalysisLevelMinimums.PruningWarnings))
+                        {
+                            _logger.Log(RestoreLogMessage.CreateWarning(NuGetLogCode.NU1511, string.Format(CultureInfo.CurrentCulture, Strings.Error_RestorePruningProjectReference, dep.Name)));
+                        }
+                    }
+                    else if (isDirectPackageReferenceFromRootProject)
```
zivkan, line 1299: I see the design spec says that direct package references of transitive projects will be pruned, but reviewing this PR makes me wonder if this will have unexpected/non-intuitive behaviour for customers?

The BCL rarely has breaking changes, but if I customer really does want to reference an older package, they will now need to reference it in every web/console/gui app, as well as every test project. Rather than just once in the class library project that uses the package.

zivkan (dismissed):

src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
nkolev92: > The BCL rarely has breaking changes, but if I customer really does want to reference an older package, they will now need to reference it in every web/console/gui app, as well as every test project. Rather than just once in the class library project that uses the package.

Even when that's the case, conflict resolution will make the latest assembly win.

jeffkl (approved):

PR #6146, opened 2024-11-15
Handle negative value for SolutionView width

### LINKED ISSUE #13928: Small Solution PM UI size can cause a System.ArgumentException SolutionView.ListView_SizeChanged
More details about the error
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1253059?src=WorkItemMention&src-action=artifact_link

martinrrm:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13928

## Description
If the user has a scaling/font size that affects the PM UI size and make it really small, this affects when we calculate the width of SolutionView, which with the new column it can make it a negative value

## PR Checklist

- ~[] Added tests~ UI change, tests width different scaling and font size

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/SolutionView.xaml.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/SolutionView.xaml.cs
@@ -259,7 +259,10 @@ namespace NuGet.PackageManagement.UI
             }
 
             var newWidth = _projectColumnHeader.ActualWidth + width;
-            _projectColumn.Width = newWidth;
+            if (newWidth > 0)
+            {
+                _projectColumn.SetValue(Canvas.WidthProperty, newWidth);
+            }
 
             // this width adjustment is only done once.
             _projectList.SizeChanged -= ListView_SizeChanged;


src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/SolutionView.xaml.cs
```diff
@@ -259,7 +259,10 @@ private void ListView_SizeChanged(object sender, SizeChangedEventArgs e)
             }
 
             var newWidth = _projectColumnHeader.ActualWidth + width;
-            _projectColumn.Width = newWidth;
+            if (newWidth > 0)
```
donnie-msft, line 262: This is probably the most failsafe change to catch the problem before setting the property.
 I wonder if there is another change in the calculations above that should be done. If ActualWidth is <8, then `width` is negative upon initialization. So, all remaining logic will just make the value more negative. See a few lines above: https://github.com/NuGet/NuGet.Client/pull/6146/files#diff-16de17305f46c11c521245379e35ac238c259620d6bd1894ede8fa2ca66a8216R243

donnie-msft (approved):
I appreciate you looking at this so quickly and talking with me about it, as we found the root cause quickly as one involving accessibility.

As I point out in the review comment, there may be more to change. I'm also wondering if we can reduce or eliminate some of this calculation.

My advice for a plan here is:
1. Talk to Vendors to see if DPI scaling is part of any testing. If so, we can ask them to help us validate that they can repro before this change, and do not repro after this change. /cc @nkolev92 for ideas there.

2. I'll follow-up by searching for similar accessibility bugs around ListView width and DPI scaling. We could engage the Accessibility team if we need more insight, but I'm not sure we do at this time.

3. Attempt to reduce the logic further, add tests, anything to simplify and guard against this going forward.

4. Get this change (or any addendums in step 3) inserted into VS.

donnie-msft:
Merging per @aortiz-msft

PR #6149, opened 2024-11-16
Fix new resolver when using transitive pinning to resolve subgraphs correctly

### LINKED ISSUE #13938: New dependency resolver erroneously reports NU1605 (downgrade) when using transitive pinning a direct dependency and a downgrade exists in a package graph
### NuGet Product Used

dotnet.exe, MSBuild.exe, NuGet.exe

### Product Version

6.12

### Worked before?

Yes

### Impact

### Repro Steps & Context

If you have a package with a downgrade in its graph, you have a direct package reference, and you have transitive pinning enabled, the new dependency resolution algorithm does not properly treat it as "pinned" and instead logs a downgrade warning.  The legacy dependency resolution algorithm treats pinned dependencies as special, even if they are direct dependencies and will use higher versions, working around the downgrade.  This results in erroneous NU1605 warnings being logged when they weren't before.

### Verbose Logs

```shell
NU1605: Warning As Error: Detected package downgrade: PackageC from 2.0.0 to 1.0.0. Reference the package directly from the project to select a different version.
 Project -> PackageA 1.0.0 -> PackageB 1.0.0 -> PackageC (>= 1.0.0)
 Project -> PackageA 1.0.0 -> PackageC (>= 2.0.0)
```

### Workaround

Users can workaround the issue by setting an MSBuild property in their `Directory.Build.props` or an environment variable:
```xml
<Project>
  <PropertyGroup>
    <RestoreUseLegacyDependencyResolver>true</RestoreUseLegacyDependencyResolver>
  </PropertyGroup>
</Project>
```

nkolev92:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13938

## Description

This fixes the new dependency resolver when transitive pinning is enabled and a package subgraph happens to contain versions that should or should not be eclipsed.

The way eclipsing works is for each subgraph, the first one wins:

```
Project1
└── A 1.0.0
    ├── B 1.0.0  
    └── C 1.0.0
        └── B 2.0.0
```
In this case, `B 2.0.0` is part of the graph of A which already depends on `B 1.0.0` so `B 1.0.0` is chosen.

However, if a higher version comes in through a different subgraph, a different version of `B` could win:

```
Project1
├── A 1.0.0
│   ├── B 1.0.0  
│   └── C 1.0.0
│       └── B 2.0.0
└── F 1.0.0
    └── B 3.0.0
```
In this case, `B 3.0.0` is not in the subgraph of `A` and its version eclipses `B 1.0.0`.

Transitive pinning is supposed to be syntactic sugar for a direct dependency and so if you pinned `C 1.0.0` in the first example, the resolver treats the graph as:

```
Project1
├── A 1.0.0
│   └── B 1.0.0  
└── C 1.0.0
    └── B 2.0.0
```

Now `B 2.0.0` is not part of the subgraph of `A` and the resolver should pick `B 2.0.0`.  However, the current implementation of the new resolver only looks to see if the proposed `B 2.0.0` and the already chosen `B 1.0.0` have the common ancestors.  This leads to the new resolver resolving the wrong graph and logging downgrade warnings as described in https://github.com/NuGet/Home/issues/13938

The fix is to modify the paths for each item if it is pinned to correctly assess if a package should eclipse another.  But when the paths are modified, the resolver must also keep track of the original parent in order to correctly create downgrade warnings if one exists. 

## PR Checklist

- [x] Added tests

### DIFF
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
@@ -489,7 +489,7 @@ namespace NuGet.Commands
                             }
 
                             //If the one we already have chosen is pure, then we can skip this one.  Processing it wont bring any new info
-                            if (chosenSuppressions.Count == 1 && chosenSuppressions[0].Count == 0)
+                            if (chosenSuppressions.Count == 1 && chosenSuppressions[0].Count == 0 && HasCommonAncestor(chosenResolvedItem.Path, pathToCurrentRef))
                             {
                                 continue;
                             }
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
@@ -161,6 +161,8 @@ namespace NuGet.Commands
                     IsDirectPackageReferenceFromRootProject = false,
                 };
 
+                LibraryRangeIndex[] rootedDependencyPath = new[] { rootProjectRefItem.LibraryRangeIndex };
+
                 _ = findLibraryEntryCache.GetOrAddAsync(
                     rootProjectRefItem.LibraryRangeIndex,
                     async static (state) =>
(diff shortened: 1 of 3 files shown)


jeffkl (dismissed):

jgonz120 (dismissed):
Seems fine! I definitely don't know much about this area, that method is massive though. A refactor would definitely help make it easier to understand.

src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
```diff
@@ -161,6 +161,8 @@ public async Task<ValueTuple<bool, List<RestoreTargetGraph>, RuntimeGraph>> Reso
                     IsDirectPackageReferenceFromRootProject = false,
                 };
 
+                LibraryRangeIndex[] rootedDependencyPath = new[] { rootProjectRefItem.LibraryRangeIndex };
```
nkolev92, line 164: Can we move this to where it's used?
I know it's being reused for pinning right now, but we don't need this array when we're not using transitive pinning which is by far the more common scenairo.

nkolev92 (commented):
The changes in DependencyGraphResolver look correct. 

Still not 100% convinced RestoreCommand_WithPackageDrivenDowngradeAndTransitivePinning_ElevatesTransitiveToDirect_AndDoesNotRaiseWarning is the best solution, despite what the old algorithm did. 

I did want to do add some IncludeAssets/ExcludeAssets tests to make sure they're not broken with this change.

src/NuGet.Core/NuGet.Commands/RestoreCommand/DependencyGraphResolver.cs
jeffkl: All direct package references have a path of `[ 0 ]` so I thought it would be a decent optimization to not create a one item array for every direct package reference.

jeffkl (approved):

jgonz120 (approved):

PR #6154, opened 2024-11-19
Add RID specific packages to `dotnet nuget why`

### LINKED ISSUE #13718: dotnet nuget why should check RID specific packages
### NuGet Product(s) Affected

dotnet.exe

### Current Behavior

Add any package which eventually adds `System.Runtime` as a depenedncy. For example, `System.Collections.Specialized`. Restore will warn you about `System.Private.Uri` having a known vulnerability (when NuGetAuditMode is set to `all`, which is default from .NET 9).  However, `dotnet nuget why my.csproj System.Private.Uri` will tell you that the package is not used by the project

### Desired Behavior

`dotnet nuget why` should tell me why this package is being restored.

### Additional Context

Looking at the assets file, the package is being included due to RID-specific (Runtime IDentifier) dependencies. Therefore, `dotnet nuget why` appears to be looking only at the "RIDless" package graph in the assets file.

Nigusu-Allehu:
# Bug

Fixes:  https://github.com/NuGet/Home/issues/13718

## Description

Originally `dotnet nuget why` only showed a path in the dependency graph, only for RIDless packages only. This PR makes sure we show RID specific packages in the tree as well. This is done by creating a tree for each RID and framework combination. Each combination will have its own tree. For example, if a user has `net9.0` framework and `win-x64` RID, we will have a tree for 
- `net 9.0 / win-x64`
- `net 9.0`
image

## PR Checklist

- [x] Added tests
- [x] Link to an issue or pull request to update docs if this PR changes settings, environment variables, new feature, etc. https://github.com/NuGet/Home/issues/13944

### DIFF
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Why/DependencyGraphFinder.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Why/DependencyGraphFinder.cs
@@ -8,6 +8,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using NuGet.ProjectModel;
+using NuGet.RuntimeModel;
 
 namespace NuGet.CommandLine.XPlat.Commands.Why
 {
@@ -36,6 +37,29 @@ namespace NuGet.CommandLine.XPlat.Commands.Why
 
             if (topLevelReferencesByFramework.Count > 0)
             {
+                foreach (KeyValuePair<string, RuntimeDescription> keyValuePair in assetsFile.PackageSpec.RuntimeGraph.Runtimes)
+                {
+                    foreach (var (targetFrameworkAlias, topLevelReferences) in topLevelReferencesByFramework)
+                    {
+                        LockFileTarget? target = assetsFile.GetTarget(targetFrameworkAlias, runtimeIdentifier: keyValuePair.Key);
+
+                        // get all package libraries for the framework
+                        IList<LockFileTargetLibrary>? packageLibraries = target?.Libraries;
+
+                        // if the project has a dependency on the target package, get the dependency graph
+                        if (packageLibraries?.Any(l => l?.Name?.Equals(targetPackage, StringComparison.OrdinalIgnoreCase) == true) == true)
+                        {
+                            doesProjectHaveDependencyOnPackage = true;
+                            dependencyGraphPerFramework.Add($"{targetFrameworkAlias} / {keyValuePair.Key}",
+                                                            GetDependencyGraphForTargetPerFramework(topLevelReferences, packageLibraries, targetPackage));
+                        }
+                        else
+                        {
+                            dependencyGraphPerFramework.Add(targetFrameworkAlias, null);
+                        }
+                    }
+                }
+
                 foreach (var (targetFrameworkAlias, topLevelReferences) in topLevelReferencesByFramework)
                 {
                     LockFileTarget? target = assetsFile.GetTarget(targetFrameworkAlias, runtimeIdentifier: null);
--- a/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Why/DependencyGraphFinder.cs
+++ b/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Why/DependencyGraphFinder.cs
@@ -53,16 +53,9 @@ namespace NuGet.CommandLine.XPlat.Commands.Why
                         if (packageLibraries?.Any(l => l?.Name?.Equals(targetPackage, StringComparison.OrdinalIgnoreCase) == true) == true)
                         {
                             doesProjectHaveDependencyOnPackage = true;
-                            if (runtimeIdentifier != null)
-                            {
-                                dependencyGraphPerFramework.Add($"{targetFrameworkAlias}/{runtimeIdentifier}",
-                                                            GetDependencyGraphForTargetPerFramework(topLevelReferences, packageLibraries, targetPackage));
-                            }
-                            else
-                            {
-                                dependencyGraphPerFramework.Add(targetFrameworkAlias,
-                                                            GetDependencyGraphForTargetPerFramework(topLevelReferences, packageLibraries, targetPackage));
-                            }
+                            var targetFrameworkDisplayName = runtimeIdentifier == null ? targetFrameworkAlias : $"{targetFrameworkAlias}/{runtimeIdentifier}";
+                            dependencyGraphPerFramework.Add(targetFrameworkDisplayName,
+                                GetDependencyGraphForTargetPerFramework(topLevelReferences, packageLibraries, targetPackage));
                         }
                         else
                         {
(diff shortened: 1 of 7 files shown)


jeffkl (dismissed):

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Why/DependencyGraphFinder.cs
```diff
@@ -31,28 +31,43 @@ internal static class DependencyGraphFinder
...
+                            if (runtimeIdentifier != null)
+                            {
+                                dependencyGraphPerFramework.Add($"{targetFrameworkAlias}/{runtimeIdentifier}",
+                                                            GetDependencyGraphForTargetPerFramework(topLevelReferences, packageLibraries, targetPackage));
+                            }
+                            else
+                            {
+                                dependencyGraphPerFramework.Add(targetFrameworkAlias,
+                                                            GetDependencyGraphForTargetPerFramework(topLevelReferences, packageLibraries, targetPackage));
+                            }
```
jgonz120:
```suggestion
                            doesProjectHaveDependencyOnPackage = true;
                            var targetFrameworkDisplayName = runtimeIdentifier == null ? targetFrameworkAlias : $"{targetFrameworkAlias}/{runtimeIdentifier}";
                            dependencyGraphPerFramework.Add(targetFrameworkDisplayName,
                                GetDependencyGraphForTargetPerFramework(topLevelReferences, packageLibraries, targetPackage));
```

jgonz120 (dismissed):
lgtm. GetAllDependencyGraphsForTarget is getting a little long though.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Why/DependencyGraphFinder.cs
```diff
@@ -31,28 +31,36 @@ internal static class DependencyGraphFinder
...
-                    {
-                        doesProjectHaveDependencyOnPackage = true;
-                        dependencyGraphPerFramework.Add(targetFrameworkAlias,
-                                                        GetDependencyGraphForTargetPerFramework(topLevelReferences, packageLibraries, targetPackage));
-                    }
-                    else
+                    foreach (var runtimeIdentifier in runtimeIdentifiers)
                     {
-                        dependencyGraphPerFramework.Add(targetFrameworkAlias, null);
+                        LockFileTarget? target = assetsFile.GetTarget(targetFrameworkAlias, runtimeIdentifier: runtimeIdentifier);
```
nkolev92: This is preexisting code so you don't have to change it, but just some general thoughts that I'll create an issue for. 
I don't think this code is doing a very good job of using nullability. 
It's basically treating everything as nullable, but there are situations where the results not be null. 

For example, we should check whether target is null, and once it's not null, things like Libraries shouldn't be null.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Why/DependencyGraphFinder.cs
```diff
@@ -31,28 +31,36 @@ internal static class DependencyGraphFinder
...
+
+                        // get all package libraries for the framework
+                        IList<LockFileTargetLibrary>? packageLibraries = target?.Libraries;
```
nkolev92: This should not be null at this point.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Why/DependencyGraphFinder.cs
```diff
@@ -31,28 +31,36 @@ internal static class DependencyGraphFinder
...
+                        if (packageLibraries?.Any(l => l?.Name?.Equals(targetPackage, StringComparison.OrdinalIgnoreCase) == true) == true)
+                        {
+                            doesProjectHaveDependencyOnPackage = true;
+                            var targetFrameworkDisplayName = runtimeIdentifier == null ? targetFrameworkAlias : $"{targetFrameworkAlias}/{runtimeIdentifier}";
+                            dependencyGraphPerFramework.Add(targetFrameworkDisplayName,
+                                GetDependencyGraphForTargetPerFramework(topLevelReferences, packageLibraries, targetPackage));
+                        }
+                        else
+                        {
+                            dependencyGraphPerFramework.Add(targetFrameworkAlias, null);
```
nkolev92: This is in a for loop with multiple rids, I think it can fail if the package does not exist in any of the rids.

test/NuGet.Core.Tests/NuGet.CommandLine.Xplat.Tests/Commands/Why/DependencyGraphFinderTests.cs
```diff
@@ -206,6 +206,32 @@ public void DifferentCaseUsedForTargetPackageId_MatchesAreCaseInsensitive_AllPat
             Assert.Contains(dependencyGraphs["net472"].First(dep => dep.Id == "DotnetNuGetWhyPackage").Children, dep => (dep.Id == "System.Text.Json") && (dep.Version == "8.0.0"));
         }
 
+        [Fact]
+        public void GetAllDependencyGraphsForTarget_WithWinX64Runtime_FindsRIDSpecificPackages()
```
nkolev92, line 210: I think we're not testing enough scenarios here. 

You're introducing the search in graphs with rids and you're only adding a success case where the package exists in a rid. 

Some other scenarios that should be tested: 

1. Ridless graph has 1 version, ridfull graph has another version.
2. multi targetted project, at least 2x frameworks, at least 2x rids, combinations of package being found in 0, 1 or both graphs with rids. 
3. The package is not available at all in a multi-rid project. I think this catches the bug I pointed out above. 
4. Casing tests. If I change line 53 in DependencyGraphFinder, no tests likely fail.

nkolev92 (changes requested):
Exciting to see this work being done. 
I do think we need more exhaustive testing. I've provided some examples in my other comments.

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Why/DependencyGraphFinder.cs
Nigusu-Allehu: I am not sure if I am clear on what you meant by it can fail. Are you suggesting a package that is not in any of the rids libraries could end up being not discovered? If so, I added a null value to the list of rids that are being requested for targets.
```csharp
var runtimeIdentifiers = assetsFile.PackageSpec.RuntimeGraph.Runtimes.Keys
                .Append(null)
                .ToList();
``` 

in line 36.
The idea is a package that is not in any of the rids will be discovered when the loop reaches `runtieIdentifier = null` from the above list?

nkolev92: Add doesn't allow duplicates. 

If you have 3 rids, and the package is missing in all 3 rids, you'll hit this codepath 3 times and it'll throw when you try to add `targetFrameworkAlias` only instead of `targetFrameworkAlias/rid`

Nigusu-Allehu: oh yeah that should be targetFrameworkDisplayName instead

src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Why/DependencyGraphFinder.cs
Nigusu-Allehu: Yes, I did not change much of the code and there is a lot of objects that are being treated as nulls

jeffkl (approved):

nkolev92 (approved):

PR #6159, opened 2024-11-22
Add support for reading PrunePackageReference in all restore codepaths

nkolev92:
# Bug

Related: https://github.com/NuGet/Home/issues/7344 

## Description

This covers the `wiring up` part of the supplied by platform work as designed in https://github.com/NuGet/Home/pull/13634. 

This change adds the PrunePackageReference concept to the PackageSpec, ensures it's read correctly in all scenarios, well tested for package spec reading and writing and all the no-op bits and disables everything by  default. 
There is no interpretation of these references. So if someone were to enable the feature specify the references, all that will happen is that those package ids will appear package spec section of the assets file with no effect. 

This handles calling `CollectPrunePackageReferences` and the reading of:
```xml
<ItemGroup>
    <PrunePackageReference Include="System.Text.Json" Version="8.0.0" />
</ItemGroup>

<PropertyGroup>
    <RestoreEnablePackagePruning>true</RestoreEnablePackagePruning>
</PropertyGroup>
```

Some more pointers to make the review easier:

The new type is called `PrunePackageReference` matching the item name. Note that the `Version` gets parsed in `(, Version]` Version Range. 
Pruning is only expected to be max version.
It's a dictionary from id to an immutable type in TargetFrameworkInformation, which means it's supported per framework. The dictionary is helpful because that's how restore would use it, similarly to how we added them in CPM.
Note that while we do read the property to enable/disable, we don't really persist it in the package spec. 
The existing of items to be pruned indicates the feature is enabled. 

For PackageReference, there's 4 places where the items are read, CPS nomination code, Legacy PR, Static Graph and Standard Restore.  I added tests in each location.

## PR Checklist

- [x] Added tests


src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
```diff
@@ -159,6 +159,34 @@ private CentralPackageVersion ToCentralPackageVersion(string packageId, string v
...
+
+            IEnumerable<(string PackageId, string Version)> packageVersions =
+                        _vsProjectAdapter.GetBuildItemInformation(ProjectItems.PrunePackageReference, ProjectBuildProperties.Version)
+                        .Select(item => (PackageId: item.ItemId, Version: item.ItemMetadata.FirstOrDefault()));
+
+            return packageVersions
+                .Select(item => ToPrunePackageReference(item.PackageId, item.Version))
+                .ToDictionary(cpv => cpv.Name, StringComparer.OrdinalIgnoreCase);
+
+            static PrunePackageReference ToPrunePackageReference(string id, string versionString)
```
jeffkl: There are a few different places where the constructor to `PrunePackageReference` is called with a string and `VersionRange` via `NuGetVersion.Parse()` which sets `includeMaxVersion`.  I think it would be better to just have a constructor to `PrunePackageReference` or static method that takes two strings and handles all of the validation.

```cs
// All xml docs here calling out exceptions
public PrunePackageReference(string id, string version)
{
    if (string.IsNullOrEmpty(id))
    {
      throw new ArgumentNullException(nameof(id));
    }

    if (string.IsNullOrEmpty(versionString))
    {
      throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Strings.Error_PrunePackageReferenceMissingVersion, id));
    }
  
    Name = name;
    VersionRange = new VersionRange(maxVersion: NuGetVersion.Parse(version), includeMaxVersion: true);
}
```

Then all of the places that need the PrunePackageReference could just call that constructor, and this static method isn't needed.

https://github.com/NuGet/NuGet.Client/pull/6159/files#diff-9d32a7ae6b417fc1b23cef544c58cea9aac005c8539bdee7d430c4004a9ab7a4R617
https://github.com/NuGet/NuGet.Client/pull/6159/files#diff-a1e0e948fb5d1439e0368357d8587e83fafe5d5bb1a4f3fad40b040b3b9bf022R392
https://github.com/NuGet/NuGet.Client/pull/6159/files#diff-69bcc4cc18dd020ef093e419c357cd2a3452094ec27be239a03c665b32de863bR788

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
```diff
@@ -716,6 +736,7 @@ internal static List<TargetFrameworkInformation> GetTargetFrameworkInfos(IReadOn
                 }
 
                 var dependencies = GetPackageReferences(msBuildProjectInstance, isCpvmEnabled, centralPackageVersions);
+                var prunedReferences = msBuildProjectInstance.IsPropertyTrue("RestoreEnablePackagePruning") ? GetPrunePackageReferences(msBuildProjectInstance) : [];
```
jeffkl, line 734: Is there a string const for "RestoreEnablePackagePruning" ?

src/NuGet.Core/NuGet.Build.Tasks/GetRestorePrunedPackageReferencesTask.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// </summary>
+        [Output]
+        public ITaskItem[] RestoreGraphItems { get; set; }
+
+        public override bool Execute()
+        {
+            if (PrunePackageReferences.Length == 0) return true;
+
+            var entries = new List<ITaskItem>(PrunePackageReferences.Length);
+            var seenIds = new HashSet<string>(PrunePackageReferences.Length, StringComparer.Ordinal);
```
jeffkl, line 40: Does this need to use a case insensitive comparer?

src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets
```diff
@@ -86,6 +86,10 @@ Copyright (c) .NET Foundation. All rights reserved.
     <NuGetAuditMode Condition=" '$(NuGetAuditMode)' == '' ">all</NuGetAuditMode>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <RestoreEnablePackagePruning Condition="$(RestoreEnablePackagePruning) == '' ">false</RestoreEnablePackagePruning>
```
jeffkl, line 80: I go back and forth on this, but its my personal opinion that we should never declare a property for a boolean that we control the default value of.  Basically, if its declared by a user that's fine, but if its not declared then we respect a default.  Every property that's declared is just one more string that every project will allocate.  If this property is going to default to false for now and later on we'll default it to true, then we can remove the property declaration here.  Just my two cents.

nkolev92: We'll make it true eventually. I had it true in early parts of my implementation. 

We'll definitely make it key off SDKAnalysisLevel and TFM targetted (enabled for .NET 10 for example).

nkolev92: I know that declaring is required for the project-system properties page, but this won't really go there.

src/NuGet.Core/NuGet.Build.Tasks/GetRestorePrunedPackageReferencesTask.cs
nkolev92: Technically yes, but fortunately there's another check that "hid" this bug.

I had some test in the "full implementation" that validated that, which worked because `MSBuildRestoreUtility` itself handles the correct casing. 

I'll make a change and push some tests.

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
nkolev92: Not one available here. 
We don't really have a properties const for this in the CLI codepaths. 
We probably should, but in this case I followed the established pattern.

We can create a tracking issue but I doubt it's something we need tracked. It's one of those, refactor when someone has time. 

lmk what you think we should do.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
nkolev92: That's a good idea. I'll look into and re-request review when ready.

src/NuGet.Core/NuGet.Build.Tasks/GetRestorePrunedPackageReferencesTask.cs
nkolev92: Scratch this. I was actually wrong.
I forgot that this needs to remain Ordinal in order to do the duplicate items check and warn.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Strings.Designer.cs
```diff
@@ -248,7 +248,7 @@ public static string Error_MultipleFrameworks {
                 return ResourceManager.GetString("Error_MultipleFrameworks", resourceCulture);
             }
         }
-        
+
```
Nigusu-Allehu: nit: we can remove this to reduce the number of files changed

src/NuGet.Core/NuGet.Build.Tasks/Strings.Designer.cs
```diff
@@ -140,7 +140,7 @@ public static string Error_PackagesConfigParseError {
                 return ResourceManager.GetString("Error_PackagesConfigParseError", resourceCulture);
             }
         }
-        
+
```
Nigusu-Allehu: nit: we can remove this

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
```diff
@@ -724,6 +725,66 @@ private static void AddPackageReferences(PackageSpec spec, IEnumerable<IMSBuildI
...
+                var versionString = item.GetProperty("VersionRange");
+                HashSet<string> tfms = GetFrameworks(item);
+
+                if (tfms.Count > 0)
+                {
+                    foreach (var targetAlias in tfms)
+                    {
+                        if (isPruningEnabled[targetAlias])
+                        {
+                            var frameworkInfo = PrunePackageReferences[targetAlias];
```
Nigusu-Allehu: is the `targetAlias` key always available in these dictionaries?

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Strings.Designer.cs
nkolev92: good catch. 
Isn't even a nit tbh, no need to change something that doesn't have any functional changes :)

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/MSBuildRestoreUtility.cs
nkolev92: Yes, on line 734 it is initialized to all values. 
If it is not available, then we have a bug.

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
```diff
@@ -159,6 +159,19 @@ private CentralPackageVersion ToCentralPackageVersion(string packageId, string v
...
+        {
+            ThreadHelper.ThrowIfNotOnUIThread();
+
+            IEnumerable<(string PackageId, string Version)> packageVersions =
+                        _vsProjectAdapter.GetBuildItemInformation(ProjectItems.PrunePackageReference, ProjectBuildProperties.Version)
+                        .Select(item => (PackageId: item.ItemId, Version: item.ItemMetadata.FirstOrDefault()));
+
+            return packageVersions
+                .Select(item => PrunePackageReference.Create(item.PackageId, item.Version))
+                .ToDictionary(cpv => cpv.Name, StringComparer.OrdinalIgnoreCase);
```
jeffkl:
```suggestion
                .ToDictionary(i => i.Name, StringComparer.OrdinalIgnoreCase);
```

Looks like this was copy/pasted from `CentralPackageVersions`...

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
```diff
@@ -445,11 +458,15 @@ private async Task<PackageSpec> GetPackageSpecAsync(ISettings settings)
             IReadOnlyDictionary<string, CentralPackageVersion> centralPackageVersions = null;
             if (isCpvmEnabled)
             {
-                // Add the central version information and merge the information to the package reference dependencies
+                // Add the central versionString information and merge the information to the package reference dependencies
                 centralPackageVersions = GetCentralPackageVersions();
                 packageReferences = ApplyCentralVersionInformation(packageReferences, centralPackageVersions);
             }
 
+            IReadOnlyDictionary<string, PrunePackageReference> packagesToPrune = MSBuildStringUtility.IsTrue(_vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.RestoreEnablePackagePruning)) ?
```
jeffkl: For some reason I can just make the suggestion here but I think this format is easier to read for the ternary operator with the `?` and `:` at the beginning of the line instead of the end:

```cs
IReadOnlyDictionary<string, PrunePackageReference> packagesToPrune = MSBuildStringUtility.IsTrue(_vsProjectAdapter.BuildProperties.GetPropertyValue(ProjectBuildProperties.RestoreEnablePackagePruning))
                ? GetPackagesToPrune()
                : ImmutableDictionary<string, PrunePackageReference>.Empty;
```

Just my opinion.

src/NuGet.Core/NuGet.Build.Tasks.Console/MSBuildStaticGraphRestore.cs
jeffkl: Okay just wanted to make sure, if we have a const use it, if not, I don't think they are always necessary.

src/NuGet.Core/NuGet.Build.Tasks/GetRestorePrunedPackageReferencesTask.cs
jeffkl: I think you should add a comment explaining, just in case someone runs a code analysis some day and tries to be too smart.

src/NuGet.Core/NuGet.LibraryModel/PrunePackageReference.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        public PrunePackageReference(
+            string name,
+            VersionRange versionRange)
+        {
+            Name = name ?? throw new ArgumentNullException(nameof(name));
+            VersionRange = versionRange ?? throw new ArgumentNullException(nameof(versionRange));
+        }
+
+        public static PrunePackageReference Create(string name, string version)
```
jeffkl, line 35: This looks much better having the validation in only one place

jeffkl (commented):
Overall looks good, just a couple comments

src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Projects/LegacyPackageReferenceProject.cs
nkolev92: Yep it was :D 
I need the same method calls D:

jeffkl (approved):
🚀

PR #6169, opened 2024-11-26
Update README loading text

jgonz120:
# Bug

Fixes: 

## Description
Update the message displayed to the user when the README is loading, per the recommendation from the UX board. Ensure screen readers announce when the README is loading. 
image

## PR Checklist

- [x] ~Added tests~

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
@@ -283,7 +283,7 @@
     <value>Loading README...</value>
   </data>
   <data name="Text_ReadmeLoaded" xml:space="preserve">
-    <value>readme loaded</value>
+    <value>README loaded</value>
   </data>
   <data name="Text_NoPackagesFound" xml:space="preserve">
     <value>No packages found</value>
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
@@ -279,6 +279,12 @@
   <data name="Text_Loading" xml:space="preserve">
     <value>Loading...</value>
   </data>
+  <data name="Text_LoadingReadme" xml:space="preserve">
+    <value>Loading README...</value>
+  </data>
+  <data name="Text_ReadmeLoaded" xml:space="preserve">
+    <value>readme loaded</value>
+  </data>
   <data name="Text_NoPackagesFound" xml:space="preserve">
     <value>No packages found</value>
   </data>
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml
@@ -58,7 +58,20 @@
         Visibility="Collapsed"
         IsFrequencyLimited="True"
         Focusable="False"
-        AutomationProperties.IsOffscreenBehavior="Offscreen" />
+        AutomationProperties.IsOffscreenBehavior="Offscreen">
+        <vs:LiveTextBlock.Resources>
+          <Style TargetType="vs:LiveTextBlock">
+            <Style.Triggers>
+              <DataTrigger Binding="{Binding Path=IsBusy}" Value="True">
+                <Setter Property="Text" Value="{x:Static nuget:Resources.Text_LoadingReadme}"/>
+              </DataTrigger>
+              <DataTrigger Binding="{Binding Path=IsBusy}" Value="False">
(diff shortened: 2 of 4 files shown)


src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
```diff
@@ -279,6 +279,12 @@
   <data name="Text_Loading" xml:space="preserve">
     <value>Loading...</value>
   </data>
+  <data name="Text_LoadingReadme" xml:space="preserve">
+    <value>Loading README...</value>
+  </data>
+  <data name="Text_ReadmeLoaded" xml:space="preserve">
+    <value>readme loaded</value>
```
donnie-msft: Why is the casing different here?

donnie-msft (dismissed):
Looks like a nice improvement! Is there any other testing pending?

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
jgonz120: Since Text_ReadmeLoaded isn't visible I figured I would set it all to lower case.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -40,6 +41,16 @@ private void ReadmeViewModel_PropertyChanged(object sender, PropertyChangedEvent
...
+            if (e.PropertyName == nameof(ReadmePreviewViewModel.IsBusy))
+            {
+                var loadingStatusText = ReadmeViewModel.IsBusy
+                    ? Resx.Text_LoadingReadme
+                    : Resx.Text_ReadmeLoaded;
+                if (!string.Equals(_ltbLoading.Text, loadingStatusText, StringComparison.OrdinalIgnoreCase))
+                {
+                    _ltbLoading.Text = loadingStatusText;
+                }
+            }
```
jebriede: This should be handled in XAML with binding and a data trigger, for example.

src/NuGet.Clients/NuGet.PackageManagement.UI/Resources.resx
donnie-msft: As we discussed offline, it's not visible, but it does affect how screen readers read it. Thanks for updating it!

donnie-msft (dismissed):
Thanks for creating a Binding and Trigger!
I suggest re-testing the accessibility of this, since we've never used a LiveTextBlock with a trigger, would be good to just ensure it works properly with screen readers.

Nigusu-Allehu (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml
```diff
@@ -31,10 +32,47 @@
...
+        Padding="10,0,0,0"
+        Grid.Column="1"
+        TextWrapping="Wrap"
+        Text="{x:Static nuget:Resources.Text_LoadingReadme}" />
+      <!--
+        Special TextBlock to report loading status to assistive technologies (narrator).
+        Code behind changes to the Text property will trigger a narrator event.
+      -->
+      <vs:LiveTextBlock
+        x:Name="_ltbLoading"
```
jebriede: The name is not needed unless you want to reference it in codebehind, which we try to avoid as much as possible.

jebriede: The name is not needed unless you want to reference it in codebehind, which we try to avoid as much as possible.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -71,6 +70,7 @@ private async Task UpdateMarkdownAsync()
         {
             try
             {
+                descriptionMarkdownPreview.Content = descriptionMarkdownPreview.Content ?? _markdownPreview.VisualElement;
```
jebriede, line 73: Question: Is this related to this particular change or does this address something else?

jgonz120: Sort of related. The loading dialog was shown large on initial load, it seemed to be related to when the content for descriptionMarkdownPreview was being set. The datacontext for the view isn't set in the construction of the view. So I moved where I sent the content to the update function since that's the only time it needs to be set anyway, and I made sure to only set it once.

jebriede (dismissed):

jebriede (approved):

Nigusu-Allehu (approved):

PR #6173, opened 2024-11-27
Dotnet tools plugin discovery: Ignore invalid paths in PATH env variable

Nigusu-Allehu:
# Bug

Fixes: 

## Description

There’s no need to notify users when an invalid path is found in the `PATH` environment variable. Instead, such paths should simply be ignored.

While investigating the cause of insertion failures, the only issue identified was the handling of invalid paths in `PATH`. Previously, the logic attempted to report these invalid paths to users, but this approach inadvertently caused exceptions to be thrown when the `PATH` environment variable contained invalid entries. This change resolves the issue by ignoring invalid paths, ensuring the process continues without disruption.

Insertion Build passes: https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/595426
## PR Checklist

- [x] Added tests

### DIFF
--- a/test/NuGet.Core.Tests/NuGet.Protocol.Tests/Plugins/PluginDiscovererTests.cs
+++ b/test/NuGet.Core.Tests/NuGet.Protocol.Tests/Plugins/PluginDiscovererTests.cs
@@ -484,6 +484,30 @@ namespace NuGet.Protocol.Plugins.Tests
         }
 
         [PlatformFact(Platform.Windows)]
+        public void GetPluginsInPATH_WithOneValueInPathEmpty_ReturnsPlugin()
+        {
+            // Arrange
+            using TestDirectory pluginPathDirectory = TestDirectory.Create();
+            using TestDirectory pathDirectory = TestDirectory.Create();
+            var pluginInNuGetPluginPathDirectoryFilePath = Path.Combine(pluginPathDirectory.Path, "nuget-plugin-auth.exe");
+            var pluginInPathDirectoryFilePath = Path.Combine(pathDirectory.Path, "nuget-plugin-in-path-directory.exe");
+            File.Create(pluginInNuGetPluginPathDirectoryFilePath);
+            File.Create(pluginInPathDirectoryFilePath);
+            Mock<IEnvironmentVariableReader> environmentalVariableReader = new Mock<IEnvironmentVariableReader>();
+            environmentalVariableReader.Setup(env => env.GetEnvironmentVariable(EnvironmentVariableConstants.PluginPaths)).Returns(Directory.GetParent(pluginInNuGetPluginPathDirectoryFilePath).FullName);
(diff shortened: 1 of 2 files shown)


test/NuGet.Core.Tests/NuGet.Protocol.Tests/Plugins/PluginDiscovererTests.cs
```diff
@@ -483,6 +483,30 @@ public void GetPluginsInPATH_WithPATHSet_ReturnsPlugin()
...
+        {
+            // Arrange
+            using TestDirectory pluginPathDirectory = TestDirectory.Create();
+            using TestDirectory pathDirectory = TestDirectory.Create();
+            var pluginInNuGetPluginPathDirectoryFilePath = Path.Combine(pluginPathDirectory.Path, "nuget-plugin-auth.exe");
+            var pluginInPathDirectoryFilePath = Path.Combine(pathDirectory.Path, "nuget-plugin-in-path-directory.exe");
+            File.Create(pluginInNuGetPluginPathDirectoryFilePath);
+            File.Create(pluginInPathDirectoryFilePath);
+            Mock<IEnvironmentVariableReader> environmentalVariableReader = new Mock<IEnvironmentVariableReader>();
+            environmentalVariableReader.Setup(env => env.GetEnvironmentVariable(EnvironmentVariableConstants.PluginPaths)).Returns(Directory.GetParent(pluginInNuGetPluginPathDirectoryFilePath).FullName);
```
zivkan:
```suggestion
            environmentalVariableReader.Setup(env => env.GetEnvironmentVariable(EnvironmentVariableConstants.PluginPaths)).Returns(pluginPathDirectory.Path);
```

Also, FWIW, I think `Path.GetDirectory(path)` is shorter than `Directory.GetParent(path).FullName`, and doesn't require the allocation of an additional `DirectoryInfo` object.

test/NuGet.Core.Tests/NuGet.Protocol.Tests/Plugins/PluginDiscovererTests.cs
```diff
@@ -483,6 +483,30 @@ public void GetPluginsInPATH_WithPATHSet_ReturnsPlugin()
...
+            environmentalVariableReader.Setup(env => env.GetEnvironmentVariable("PATH")).Returns($"{Directory.GetParent(pluginInPathDirectoryFilePath).FullName}{Path.PathSeparator}");
```
zivkan:
```suggestion
            environmentalVariableReader.Setup(env => env.GetEnvironmentVariable("PATH")).Returns($"{pathDirectory.Path}{Path.PathSeparator}");
```

Also, I think it's worthwhile adding a whitespace line before this line, and adding a comment saying something along the lines "path separator must be last, to ensure that there's an empty path entry".  When looking quickly at this test, I found it kind of difficult to figure out where it's setting up the scenario described by the test name.

zivkan (approved):

PR #6174, opened 2024-11-27
[release-6.12.x] Handle negative value for SolutionView width (#6146)

donnie-msft:
Cherry-pick from dev to release-6.12.x: 1fd453890624fd3bf58b0aeba7f5c422e414ac35
Bug: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1253059?src=WorkItemMention&src-action=artifact_link


zivkan (dismissed):

nkolev92:
Do we need to bump the version to 6.12.3?
@zivkan yell if you think we're wrong. 

6.12.2 has the audit revert and it will ship earlier than this change will.

nkolev92 (dismissed):

donnie-msft:
> Do we need to bump the version to 6.12.3? @zivkan yell if you think we're wrong.
> 
> 6.12.2 has the audit revert and it will ship earlier than this change will.

I went ahead and bumped to 6.12.3 until someone says we shouldn't.

zivkan (approved):

PR #6179, opened 2024-12-02
Don't select transitive packages in PM UI when updating a package

### LINKED ISSUE #13893: Nuget Packages Manager screen for Solution automatically selects transitive dependencies
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/Nuget-Packages-Manager-screen-for-Soluti/10776764)._

---
In the Nuget Packages Manager screen for solution, in the "Updates" list, when selecting a package to update, all the projects when the package is top-level AND transitive are automatically selected.

Clicking the Update button (underneath the Uninstall button) will update the package reference for the top-level registration, but also add a package reference in all the projects where the package is used as a transitive dependency. Which is not the intended use of transitive dependencies.

I would expect to only have the project where the dependency is a top-level one to be selected automatically.

The solution I'm working with a .NET Framework 4.8 solution, using PackageReference and with SDK-style projects.

_PS: when updating multiple packages at once, by selecting them in the package list of the Update tab, and clicking the "Update" button located on top of that list, only the projects where the dependencies are top-level are updated, which is the intended behavior._

---
#### Feedback Bot on 28/10/2024, 00:10 PM: 

<p>We have directed your feedback to the appropriate engineering team for further evaluation. The team will review the feedback and notify you about the next steps.</p>

martinrrm:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13893

## Description
Transitive packages are automatically selected when you are in the Updates tab and clicking the Update button will update the package reference for the top-level packages but also add a package reference in all the projects where the package is used as a transitive dependency, promoting them to top-level, which is not the desired behavior in most of the cases.

Promoting a package from transitive to top-level is not a very common scenario, most likely only done when fixing a vulnerability or resolving a dependency conflict. 

image

## PR Checklist

- [ ] Added tests ~I'm not adding since it's a UI change in a code path with no tests, I could try to add a unit test if team wants~

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
@@ -612,7 +612,7 @@ namespace NuGet.PackageManagement.UI
             {
                 foreach (PackageInstallationInfo project in Projects)
                 {
-                    project.IsSelected = project.InstalledVersion != null;
+                    project.IsSelected = project.InstalledVersion != null && project.PackageLevel.Equals(PackageLevel.TopLevel);
                 }
             }
             finally


jeffkl (approved):

nkolev92:
Can you ask the vendors to add a manual test if you can't add an automated one?

Nigusu-Allehu (approved):

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/PackageSolutionDetailControlModel.cs
```diff
@@ -612,7 +612,7 @@ private void BatchUpdateIsSelectedBasedOnInstalledVersion()
             {
                 foreach (PackageInstallationInfo project in Projects)
                 {
-                    project.IsSelected = project.InstalledVersion != null;
+                    project.IsSelected = project.InstalledVersion != null && project.PackageLevel.Equals(PackageLevel.TopLevel);
```
jebriede, line 615: Protect against null PackageLevel.
```suggestion
                    project.IsSelected = project.InstalledVersion != null && PackageLevel.TopLevel.Equals(project.PackageLevel);
```

```suggestion
                    project.IsSelected = project.InstalledVersion != null && project.PackageLevel?.Equals(PackageLevel.TopLevel) == true;
```

jebriede (approved):

PR #6180, opened 2024-12-04
Add PackageDownload vulnerability telemetry

Nigusu-Allehu:
# Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/3112

## Description
Telemetry design https://github.com/NuGet/Client.Engineering/blob/main/designs/telemetry/telemetry-design-PackageDownload-vulnerabilities.md
This PR introduces telemetry collection for vulnerabilities in `<PackageDownload>` packages:

#### Summary of Changes
- Collects the list of vulnerable package IDs at the end of the restore operation.

#### Goals
This telemetry will enable us to:
- Determine how many restore operations generate warnings due to vulnerable `<PackageDownload>` packages.
- Use package names to estimate the proportion of warnings attributable to SDK-managed packages.

## PR Checklist

- [x] Added tests


src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -78,10 +82,99 @@ public AuditUtility(
...
+
+            if (packagesWithKnownVulnerabilities == null)
+            {
+                return;
+            }
+
+            PackageDownloadPackagesWithAdvisory = new(capacity: packagesWithKnownVulnerabilities.Values.Count);
+
+            // no-op checks DGSpec hash, which means the order of everything must be deterministic.
+            foreach ((PackageIdentity package, PackageAuditInfo auditInfo) in packagesWithKnownVulnerabilities.OrderBy(p => p.Key.Id))
```
zivkan: The DGSpec is project restore inputs, but audit affects outputs. So, the comment doesn't make sense in this context.

I can't remember if we sort for PackageReference warnings, but I can't think of any benefit to sorting for telemetry. And audit runs in restore, which is our most perf-critical path. All LINQ methods cause a enumerator allocation on the heap, then since it returns an enumerable, the foreach allocates another. And finally, OrderBy will have to create an internal list (which might need resizing) in order to sort the results.

So, the easiest is if we don't need to sort, we can avoid LINQ and all the memory allocations it needs.  If we do need to sort, then we should re-think the `FindDownloadDependenciesWithKnownVulnerabilities` method's data structures and find a better solution that won't require so many heap allocations.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -78,10 +82,99 @@ public AuditUtility(
...
+                PackageDownloadPackagesWithAdvisory.Add(package.Id);
+
+                foreach (var advisory in auditInfo.GraphsPerVulnerability.Keys)
+                {
+                    PackageVulnerabilitySeverity severity = advisory.Severity;
+                    if (severity == PackageVulnerabilitySeverity.Low) { Sev0DirectMatches++; }
+                    else if (severity == PackageVulnerabilitySeverity.Moderate) { Sev1DirectMatches++; }
+                    else if (severity == PackageVulnerabilitySeverity.High) { Sev2DirectMatches++; }
+                    else if (severity == PackageVulnerabilitySeverity.Critical) { Sev3DirectMatches++; }
+                    else { InvalidSevDirectMatches++; }
```
zivkan: I'm still processing this, but my gut reaction is since we're not warning customers about PackageDownload vulnerabilities, we won't want to increment the same counters as PackageReference (where we are warning), because it makes it harder for us to understand if customers are taking action on the warnings we do provide.

Even when we start warning customers about PackageDownloads with known vulnerabilities, my gut feeling is that we should have a different set of counters. I expect most of the PackageDownloads with vulnerabilities will be from the SDK, and the way customers need to take action is radically different to upgrading a PackageReference. Even if a customer has a custom PackageDownload, they can't use any of our tooling (in VS, this is mainly the Package manager UI) to upgrade the version.

So, I think it helps us make better product decisions if the PackageDownload severity counts are different to direct PackageReference for the same reason that it was useful to know direct vs transitive.

test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/Utility/AuditUtilityTests.cs
```diff
@@ -179,6 +179,9 @@ public async Task Check_ProjectReferencingPackageWithVulnerability_WarningLogged
 
         context.PackagesDependencyProvider.Package("pkga", "1.0.0").DependsOn("pkgb", "1.0.0");
         context.PackagesDependencyProvider.Package("pkgb", "1.0.0");
+        context.PackageDownloadPackages.Add(new DownloadDependency("pkgDownload", new VersionRange(new NuGetVersion("1.0.0"), maxVersion: new NuGetVersion("1.0.0"))));
```
zivkan: more of a nitpick:  these package download package lines look a lot less nice than adding the package reference packages on the two lines above, since they're using a builder pattern, but this package download is not.

I wonder if it's worthwhile creating a `DownloadDependencyProvider` so tests that use it are easier to read.

Having said that, half the line length could be reduced if the `VersionRange` was changed to `VersionRange.Parse("[1.0.0])`.  I wouldn't be surprised if I did the `new VersionRange` thing in the past, since I know it will execute faster than Parse. I keep changing my mind between "OMG, NuGet's tests are so slow, we need to make them faster, if only by a little bit" and also "these tests need to be easier to read" 🤣

test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/Utility/AuditUtilityTests.cs
```diff
@@ -427,6 +437,7 @@ await createGraphTasks[1]
 
         // Act
         var audit = new AuditUtility(
+            downloadDependencies: new List<DownloadDependency>(),
```
zivkan:
```suggestion
            downloadDependencies: Enumerable.Empty<DownloadDependency>(),
```

I don't remember if there is an `Enumerable.Empty`, but considering `AuditUtility` takes in an `IEnumerable`, there are several zero alloc options, like `Array.Empty<T>()`.  I wouldn't be surprised if simply putting `[]` lets the compiler auto-detect its own choice for a zero alloc zero-length collection.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -535,6 +536,7 @@ private async Task<bool> PerformAuditAsync(IEnumerable<RestoreTargetGraph> graph
         {
             telemetry.StartIntervalMeasure();
             var audit = new AuditUtility(
+                _request.Project.TargetFrameworks.SelectMany(t => t.DownloadDependencies),
```
zivkan: LINQ always causes heap allocations. Firstly in the delegate method it needs to pass, secondly the enumerator that it internally uses to iterate all the items in `TargetFrameworks`, and finally there will be another allocation when the result of the `SelectMany` is enumerated. The return type of SelectMany may even be a fourth allocation.  It is a fixed number of allocations per call, but it will happen for each project in the restore.

So, I'm thinking that passing through the `TargetFrameworks` collection, and then having audit iterate that (using `.NoAllocEnumerate()`, or whatever that API is called), and only read the `DownloadDependencies` property will allow us to check PackageDownloads without any allocations, and therefore fewer garbage collections when there are a huge number of projects.

Unfortunately, it will make setting up tests harder, so it's a perf trade-off. audit runs in restore, which is our most perf-critical path.  I hope other people in the team speak up if they think I'm being too nit-picky about 3 or 4 allocations.

test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/Utility/AuditUtilityTests.cs
nkolev92: I don't think we ever sort the log messages. 
The order in which they're raised is preserved, which is why we've never had issues. 
They're never relevant in the input/output calculation, since they're normally just be replayed and there's per log code sorting in things like error list and msbuild.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
nkolev92: We never sort them because they're not used in any hashing. 
The order in which they're raised is preserved, so that part is deterministic which is all we need.

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
nkolev92: The "equivalent" way of doing the audit to PR is to actually audit the downloaded packages as well. 

In this case it's not as relevant, but if that was represented in these types, maybe we wouldn't have needed to pass anything new. 
Given that's a different refactoring that I might need to do as part of https://github.com/NuGet/Home/issues/7722, and this PR would carry more urgency than that, I think it's ok if that refactoring is not done here but something to consider.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -23,7 +23,8 @@ namespace NuGet.Commands.Restore.Utility
 {
     internal class AuditUtility
     {
-        private readonly ProjectModel.RestoreAuditProperties? _restoreAuditProperties;
+        private readonly IEnumerable<DownloadDependency> _downloadDependencies;
```
nkolev92: It's a good practice to add new types at the bottom of the field declarations unless the new type is clearly the most important one in this class.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -52,12 +54,14 @@ internal class AuditUtility
         internal int TotalWarningsSuppressedCount { get; private set; }
 
         public AuditUtility(
-            ProjectModel.RestoreAuditProperties? restoreAuditProperties,
+            IEnumerable<DownloadDependency> downloadDependencies,
```
nkolev92: Similar feedback, new args at the end. It's often ok if it things like cancellation token or loggers are last, so you can add it before that, or maybe even after graphs.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
nkolev92: 1. I agree we shouldn't mix them, so we should change that at minimum.
2. Having data about the different severities feels like something interesting to know, but I don't know if it truly drives our behavior as @zivkan said, so it's ok to keep one counter, although I wouldn't be completely opposed to per severity ones.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -78,10 +82,99 @@ public AuditUtility(
...
+        {
+            Dictionary<PackageIdentity, PackageAuditInfo>? result = null;
+
+            foreach (var downloadDependency in _downloadDependencies)
+            {
+                List<PackageVulnerabilityInfo>? knownVulnerabilitiesForPackage = GetKnownVulnerabilities(downloadDependency.Name, downloadDependency.VersionRange.MaxVersion!, knownVulnerabilities);
+
+                if (knownVulnerabilitiesForPackage?.Count > 0)
+                {
+                    PackageIdentity packageIdentity = new(downloadDependency.Name, downloadDependency.VersionRange.MinVersion);
```
nkolev92: Do we not have this package identity anywhere? 

Can we use the DownloadDependency type? PackageDownload dependency tend to be duplicated among projects, so we should reduce allocations in these codepaths as much as possible.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -78,10 +82,99 @@ public AuditUtility(
...
+                            continue;
+                        }
+
+                        if (SuppressedAdvisories?.TryGetValue(knownVulnerability.Url.OriginalString, out bool advisoryUsed) == true)
+                        {
+                            TotalWarningsSuppressedCount++;
+
+                            if (!advisoryUsed)
+                            {
+                                SuppressedAdvisories[knownVulnerability.Url.OriginalString] = true;
```
nkolev92: @zivkan Do we think we'd want to differentiate between PD and PR suppressions?
I feel like we would, mostly because PD could be indicative of out of date SDK and that's a totally different problem.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -78,10 +82,99 @@ public AuditUtility(
...
+            }
+            return result;
+        }
+
+
         public async Task<bool> CheckPackageVulnerabilitiesAsync(CancellationToken cancellationToken)
         {
             // Performance: Early exit if restore graph does not contain any packages.
-            if (!HasPackages())
+            if (!HasPackages() && (_downloadDependencies == null || !_downloadDependencies.Any()))
```
nkolev92: Use Count.

nkolev92 (commented):
Consider using a telemetry template issue: https://github.com/NuGet/Client.Engineering/issues/new?assignees=&labels=Type%3AEngineering%2CPriority%3A2&projects=&template=telemetry.md. 

I think we've forgotten to use it reliable (includes all of us :) )

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
jeffkl: I agree with Andy that making a copy of these is not a good idea.  Passing along the collection of TargetFrameworks isn't ideal either, maybe we should just pass the entire `PackageSpec` or perhaps an array of the lists?

```csharp
var downloadDependencyLists = new ImmutableArray<DownloadDependency>[_request.Project.TargetFrameworks.Count];

for (int i = 0; i < _request.Project.TargetFrameworks.Count; i++)
{
  downloadDependencyLists[i] = _request.Project.TargetFrameworks[i].DownloadDependencies;
}
```

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -78,10 +82,99 @@ public AuditUtility(
             }
         }
 
+        private void CheckDownloadDependenciesVulnerabilities(IReadOnlyList<IReadOnlyDictionary<string, IReadOnlyList<PackageVulnerabilityInfo>>> knownVulnerabilities)
```
jeffkl: There's a mixture of naming between `DownloadDependencies` and `PackageDownloads`.  Since in the code its already called `DownloadDependencies`, we should probably just stick with that?  The MSBuild items of course are `<PackageDownload />` 🤦

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
Nigusu-Allehu: I have removed this so that we don't mix the counts for PD and PR. In addition, this count does not help with the questions we are trying to answer: how many restores could have a warning from PD vulnerabilities

test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/Utility/AuditUtilityTests.cs
Nigusu-Allehu: Thanks for pointing this out. Personally, I am indifferent on this. But `VersionRange.Parse()` looks cleaner to me

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
Nigusu-Allehu: Thanks all for the ideas. I think @nkolev92 is right, refactoring might have been a good idea. But since we want to merge this ASAP, maybe we could do it as part of https://github.com/NuGet/Home/issues/7722.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
Nigusu-Allehu: I have removed this telemetry as well. It does not help with the question we are trying to answer

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -115,6 +115,7 @@ private readonly Dictionary<RestoreTargetGraph, Dictionary<string, LibraryInclud
         private const string AuditTransitiveVulnerabilitySev2 = "Audit.Vulnerability.Transitive.Severity2";
         private const string AuditTransitiveVulnerabilitySev3 = "Audit.Vulnerability.Transitive.Severity3";
         private const string AuditTransitiveVulnerabilitySevInvalid = "Audit.Vulnerability.Transitive.SeverityInvalid";
+        private const string AuditPackageDownloadVulnerabilitiesPackages = "Audit.Vulnerability.PackageDownload.Packages";
```
nkolev92: Did you and @zivkan have an extra conversation that wasn't included here?

Curious why we're not adding the counts for package downloads with vulnerabilities?

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -78,10 +82,27 @@ public AuditUtility(
...
+                    }
+                }
+            }
+        }
+
         public async Task<bool> CheckPackageVulnerabilitiesAsync(CancellationToken cancellationToken)
         {
             // Performance: Early exit if restore graph does not contain any packages.
-            if (!HasPackages())
+            if (!HasPackages() && (_targetFrameworks.Count() == 0))
```
zivkan: never use use Linq's `.Count()` when the type has a `.Length` or `.Count` property.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -78,10 +82,27 @@ public AuditUtility(
             }
         }
 
+        private void CheckPackageDownloadVulnerabilities(IReadOnlyList<IReadOnlyDictionary<string, IReadOnlyList<PackageVulnerabilityInfo>>> knownVulnerabilities)
+        {
+            foreach (var targetFramework in _targetFrameworks)
```
zivkan: Since `_targetFrameworks` is `IList`, this `foreach` loop will cause an allocation. And since audit runs in restore, which is our most perf-critical path, we need to do everything we can to minimize perf impacts.  So, either change this to a `for` loop and use the indexer to get the value from the list, or use `NoAllocEnumerate` (or whatever that extension method/class is called).

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
zivkan: Not that I can remember 🤣 

I think `Audit.Vulnerability.PackageDownload.*` properties make sense and will make it easier for us to tell the difference between older versions of NuGet/VS that don't support this telemetry, and restores that have zero PackageDownloads with known vulnerabilities.  We could do it by devenv ExeVersion, but I prefer to filter out null values.

Nigusu-Allehu: I was thinking we could do the counting from the `AuditPackageDownloadVulnerabilitiesPackages` list.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -78,10 +90,100 @@ public AuditUtility(
...
+
+            return result;
+        }
+
+
         public async Task<bool> CheckPackageVulnerabilitiesAsync(CancellationToken cancellationToken)
         {
             // Performance: Early exit if restore graph does not contain any packages.
-            if (!HasPackages())
+            if (!HasPackages() && (_targetFrameworks.Count == 0))
```
zivkan: I don't think it's possible for there to ever be `_targetFrameworks.Count == 0`. Any project with a valid `<TargetFramework>` is going to have a non-empty collection, even if it doesn't have any PackageReference or PackageDownload items.

There is a test that attempts to check for this scenario: https://github.com/NuGet/NuGet.Client/blob/949031c79ea7f5c15b59f31e26d5355fb23795ca/test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/Utility/AuditUtilityTests.cs#L109

However, due to how this class and the test helper methods were modified, it allows the test to use data that is impossible in production, and therefore fail to catch this regression.

There might be better ideas, but my first idea is to have code that validates that the list of `RestoreTargetGraph[]` TFMs match the `IList<TargetFrameworkInformation>` TFMs.  Both in count and actually validate that the two collections use identical TFMs. Since restore perf is critical, this test should be done in the test helper. While most tests in the class use the "context" class, Check_MultiTargetingProjectFile_WarningsHaveExpectedProperties does not use it, so it would be great if you could find a way to implement this validation that works for that test as well (or improve the "context" class so that it can be used by this test)

test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/Utility/AuditUtilityTests.cs
```diff
@@ -431,6 +453,7 @@ await createGraphTasks[1]
             "/path/proj.csproj",
             graphs,
             vulnerabilityProviders,
+            targetFrameworks: Enumerable.Empty<TargetFrameworkInformation>().ToList(),
```
zivkan:
```suggestion
            targetFrameworks: new List<TargetFrameworkInformation>(0),
```

The purpose of `Enumerable.Empty<T>()` is to get an empty enumerable that doesn't cause any heap allocations (after the first time it's called). However, using `ToList()` defeats that purpose, as it has to allocate a `List<T>`. Using LINQ's `ToList` is just a less efficient way of creating the list directly, so might as well make the line shorter as well.

However, as mentioned in a comment in the production code, this is invalid from a business point of view. This list cannot ever be empty when the restore `graphs` are non empty. They must have the same TFMs, after all the restore graphs are calculated from the input target frameworks.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
Nigusu-Allehu: I have updated the if statement to actually go through the TFMs and count `PackageDownload`s. This should not have perf regressions. Since we already have another loop that goes through all the `PackageDownload`s, adding one more loop for counting purposes should be okay as long as we don't do an allocation.

I have updated the test context to have at least one TFM.

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -78,10 +90,110 @@ public AuditUtility(
...
+
+            foreach (var targetFramework in _targetFrameworks.NoAllocEnumerate())
+            {
+                foreach (var downloadDependency in targetFramework.DownloadDependencies)
+                {
+                    List<PackageVulnerabilityInfo>? knownVulnerabilitiesForPackage = GetKnownVulnerabilities(downloadDependency.Name, downloadDependency.VersionRange.MaxVersion!, knownVulnerabilities);
+
+                    if (knownVulnerabilitiesForPackage?.Count > 0)
+                    {
+                        DownloadDependency downloadDependencyPackage = new(downloadDependency.Name, downloadDependency.VersionRange);
```
zivkan: this variable doesn't appear to be used anywhere

src/NuGet.Core/NuGet.Commands/RestoreCommand/Utility/AuditUtility.cs
```diff
@@ -78,10 +90,110 @@ public AuditUtility(
...
+
+            return result;
+        }
+
+
         public async Task<bool> CheckPackageVulnerabilitiesAsync(CancellationToken cancellationToken)
         {
             // Performance: Early exit if restore graph does not contain any packages.
-            if (!HasPackages())
+            if (!HasPackages() && (CountPackageDownloads() == 0))
```
zivkan: nitpick: The existing method `HasPackages` doesn't sound specific to PackageReference, and in fact `CountPackageDownloads() == 0` sounds like it belongs within `HasPackages`. I don't think it should be a different method, just a new couple of lines within `HasPackages()`

test/NuGet.Core.Tests/NuGet.Commands.Test/RestoreCommandTests/Utility/AuditUtilityTests.cs
```diff
@@ -431,6 +453,7 @@ await createGraphTasks[1]
             "/path/proj.csproj",
             graphs,
             vulnerabilityProviders,
+            targetFrameworks: new List<TargetFrameworkInformation>(0),
```
zivkan: This test has a restore graph with two TFMs, so creating this collection with none means the production code is being tested with inputs that are impossible from a customer's perspective.  While I don't believe that this is causing any problems in this PR, the point is that an earlier iteration of this PR introduced a regression, even though all tests passed.

If we don't improve the test infrastructure, maybe the next person who works on `AuditUtility` will introduce a similar regression that tests won't catch, and whoever reviews that PR won't notice either.

Whereas, if we write our code in a way that prevents us from making these mistakes, then it's less likely that human mistakes will cause bugs in what we ship to customers.

zivkan (dismissed):
I'm confident that this PR now achieves its stated goal, without introducing regressions.

I still think we can make improvements to reduce the risk of future regressions though.

zivkan (approved):

PR #6185, opened 2024-12-05
reduce memory in ManifestMetadata

SimonCropp:
# Bug

Fixes: 

## Description

reduce memory in ManifestMetadata by using collection expressions that will result in those `IEnumerable` members being compiled to a `.Empty`.

let me know if u want to use explicit `Enumerable.Empty`s

## PR Checklist

- [ ] Added tests

### DIFF
--- a/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestMetadata.cs
+++ b/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestMetadata.cs
@@ -187,7 +187,7 @@ namespace NuGet.Packaging
 
         public RepositoryMetadata Repository { get; set; }
 
-        private IEnumerable<PackageDependencyGroup> _dependencyGroups = new List<PackageDependencyGroup>();
+        private IEnumerable<PackageDependencyGroup> _dependencyGroups = [];
         public IEnumerable<PackageDependencyGroup> DependencyGroups
         {
             get
@@ -200,11 +200,11 @@ namespace NuGet.Packaging
             }
         }
 
-        public IEnumerable<FrameworkReferenceGroup> FrameworkReferenceGroups { get; set; } = new List<FrameworkReferenceGroup>();
+        public IEnumerable<FrameworkReferenceGroup> FrameworkReferenceGroups { get; set; } = [];
 
-        public IEnumerable<FrameworkAssemblyReference> FrameworkReferences { get; set; } = new List<FrameworkAssemblyReference>();
+        public IEnumerable<FrameworkAssemblyReference> FrameworkReferences { get; set; } = [];
 
-        private IEnumerable<PackageReferenceSet> _packageAssemblyReferences = new List<PackageReferenceSet>();
+        private IEnumerable<PackageReferenceSet> _packageAssemblyReferences = [];
 
         [ManifestVersion(2)]
         public IEnumerable<PackageReferenceSet> PackageAssemblyReferences


src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestMetadata.cs
```diff
@@ -187,7 +187,7 @@ public Uri ProjectUrl
 
         public RepositoryMetadata Repository { get; set; }
 
-        private IEnumerable<PackageDependencyGroup> _dependencyGroups = new List<PackageDependencyGroup>();
+        private IEnumerable<PackageDependencyGroup> _dependencyGroups = [];
```
Nigusu-Allehu, line 190: nit: we could make the intent clear by
```suggestion
        private IEnumerable<PackageDependencyGroup> _dependencyGroups = Enumerable.Empty<PackageDependencyGroup>();
```

Nigusu-Allehu (approved):

src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestMetadata.cs
nkolev92: Under the hood, they're the same thing in .NET Core.

nkolev92 (approved):
Thanks for the contribution @SimonCropp 

Something to consider. This change is not really in a hot path. ManifestMetadata is only done at pack time and given that pack runs in a separate process, the small gains here aren't winning us much. Small refactoring like this do have an impact on `git blame`, and in some cases, make things a bit more challenging to debug/understand intent behind certain changes, so when it comes cost/benefit, the hotter paths are better suited for these types of changes

src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestMetadata.cs
SimonCropp: @Nigusu-Allehu there are a significant number of places in this code base where the intent is clear, but is results in sub-optimal code. the `[]` under the hood is smart enough to chose the best type for the majority of scenarios. IMO we should let it choose.

but it is your code base so happy to be overruled

SimonCropp:
@nkolev92 

> Something to consider. This change is not really in a hot path. ManifestMetadata is only done at pack time and given that pack runs in a separate process, the small gains here aren't winning us much. Small refactoring like this do have an impact on git blame, and in some cases, make things a bit more challenging to debug/understand intent behind certain changes, so when it comes cost/benefit, the hotter paths are better suited for these types of changes

i guess there are 3 options

 * continue with small targeted PRS
 * do large PRs on the whole code base 
 * leave the slow code in place.

as for "This change is not really in a hot path". true. but it is more efficient and less code, given that it seems a shame to not apply the change no matter the code path

src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestMetadata.cs
Nigusu-Allehu: Either is fine with me as well. We can mark this as resolved

nkolev92:
Hey @SimonCropp, 

I just wanted to again thank you for all the contributions lately, I see you've raised a bunch over the last weekend as well. 

With all the history and compatibility promises, we think there's benefit in preserving history to allow us to use `git blame` effectively.
We don't really have a blanket policy for refactoring changes, since it's all about the cost/benefit of each change, and with the amount of features in this repo and the different velocities of each, the comfort level for these changes will be different.

So a few things we'll consider:

- When changing coding patterns that improve performance for example, we prefer fixing and enabling a rule to never introduce the issue again. Seeing the amount of changes to enable the rule would allow to decide between more targeted/wholesale fixes.
- Non-functional refactoring PRs are not likely to be something we accept, especially in codepaths that are not very active, and haven't had changes recently. If there's a rule that can be enable to ensure the pattern is respected, we might considered it more strongly.

We want to be appreciative of your efforts, so before you spend a lot of effort on certain changes, please follow https://github.com/NuGet/NuGet.Client/blob/dev/CONTRIBUTING.md#contributing-step-by-step and file an issue to get a buy-off.

PR #6186, opened 2024-12-05
Add indicator for deprecated and vulnerable packages to Package Details tab

### LINKED ISSUE #13974: Add indicator for deprecated and vulnerable packages to Package Details tab header.
The vulnerability and deprecated warnings were moved to inside the package details tab. We want to make sure users are aware of the any issues with the packages they have selected so we want to add a warning icon similar to the one on the "Installed" tab in the PM UI.

jgonz120:
# Bug

Fixes: https://github.com/NuGet/Home/issues/13974

## Description
Add warning icon to the Package Details tab to indicate to users when the selected package has a vulnerability or is deprecated.
image
image
image
image
image

## PR Checklist

- [x] ~Added tests~

### DIFF
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
@@ -444,7 +444,6 @@ namespace NuGet.PackageManagement.UI
                 OnPropertyChanged(nameof(PackageVulnerabilities));
                 OnPropertyChanged(nameof(PackageVulnerabilityMaxSeverity));
                 OnPropertyChanged(nameof(IsPackageVulnerable));
-                OnPropertyChanged(nameof(IsPackageVulnerableOrDeprecated));
                 OnPropertyChanged(nameof(PackageVulnerabilityCount));
             }
         }
@@ -535,9 +534,9 @@ namespace NuGet.PackageManagement.UI
                     OnPropertyChanged(nameof(PackageMetadata));
                     OnPropertyChanged(nameof(IsPackageDeprecated));
                     OnPropertyChanged(nameof(IsPackageVulnerable));
-                    OnPropertyChanged(nameof(IsPackageVulnerableOrDeprecated));
                     OnPropertyChanged(nameof(PackageVulnerabilityCount));
                     OnPropertyChanged(nameof(PackageVulnerabilities));
+                    OnPropertyChanged(nameof(IsPackageVulnerableOrDeprecated));
                     OnPropertyChanged(nameof(PackageVulnerabilityMaxSeverity));
                 }
             }
--- a/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
+++ b/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
@@ -444,6 +444,7 @@ namespace NuGet.PackageManagement.UI
                 OnPropertyChanged(nameof(PackageVulnerabilities));
                 OnPropertyChanged(nameof(PackageVulnerabilityMaxSeverity));
                 OnPropertyChanged(nameof(IsPackageVulnerable));
+                OnPropertyChanged(nameof(IsPackageVulnerableOrDeprecated));
                 OnPropertyChanged(nameof(PackageVulnerabilityCount));
             }
         }
@@ -456,6 +457,11 @@ namespace NuGet.PackageManagement.UI
         public bool IsPackageVulnerable
         {
             get => PackageVulnerabilities?.Count > 0;
+        }
+
+        public bool IsPackageVulnerableOrDeprecated
+        {
+            get => IsPackageVulnerable || IsPackageDeprecated;
         }
 
         public int PackageVulnerabilityCount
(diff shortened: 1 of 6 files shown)


src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageDetailsTabControl.xaml
```diff
@@ -115,12 +116,30 @@
...
+        <StackPanel Orientation="Horizontal" >
+          <TextBlock Text="{Binding Title}" />
+          <imaging:CrispImage
+            Margin="3,0"
+            Visibility="{Binding Path=IsPackageVulnerableOrDeprecated, FallbackValue=Collapsed, Mode=OneWay, Converter={StaticResource BooleanToVisibilityConverter}}"
+            Moniker="{x:Static catalog:KnownMonikers.StatusWarning}">
+            <imaging:CrispImage.ToolTip>
+              <StackPanel>
+                <TextBlock
+                  Visibility="{Binding IsPackageVulnerable,Converter={StaticResource BooleanToVisibilityConverter}}">
```
jebriede: nit: for consistency, add space after comma.

```suggestion
                  Visibility="{Binding IsPackageVulnerable, Converter={StaticResource BooleanToVisibilityConverter}}">
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
```diff
@@ -529,6 +535,7 @@ public DetailedPackageMetadata PackageMetadata
                     OnPropertyChanged(nameof(PackageMetadata));
                     OnPropertyChanged(nameof(IsPackageDeprecated));
                     OnPropertyChanged(nameof(IsPackageVulnerable));
+                    OnPropertyChanged(nameof(IsPackageVulnerableOrDeprecated));
```
jebriede, line 536: It looks like we're doing the OnPropertyChanged call for this property, and some of the other ones, more than one time. The setter for PackageVulnerabilities calls OnPropertyChanged with the appropriate fields that depend on it. Since this setter for PackageMetadata sets the PackageVulnerabilities, and then again calls OnPropertyChanged for those fields, we're calling it twice. We should make sure the OnPropertyChanged events get fired only once for each of these properties. Separate which ones are dependent on PackageVulnerabilities and which ones are dependent on PackageMetadata, considering how the PackageMetadata setter also sets PackageVulnerabilities.

jgonz120: Valid, I think I should be able to get away with only firing it in the PackageMetadata setter. It looks like a bunch of them are being set twice at the moment though.

Nigusu-Allehu (dismissed):

martinrrm:
@jgonz120 We just added new vulnerability warning icons to the Details Pane in the Solution PM UI in the projects table, which use the same icon. I think for project level this fine, but maybe we need to add deprecated data the icons in the Solution level otherwise they will display different information. 

I'm not saying that we should do that in this PR but maybe is needed for this VS version.

jeffkl (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs
```diff
@@ -535,9 +534,9 @@ public DetailedPackageMetadata PackageMetadata
                     OnPropertyChanged(nameof(PackageMetadata));
                     OnPropertyChanged(nameof(IsPackageDeprecated));
                     OnPropertyChanged(nameof(IsPackageVulnerable));
-                    OnPropertyChanged(nameof(IsPackageVulnerableOrDeprecated));
                     OnPropertyChanged(nameof(PackageVulnerabilityCount));
                     OnPropertyChanged(nameof(PackageVulnerabilities));
+                    OnPropertyChanged(nameof(IsPackageVulnerableOrDeprecated));
```
jebriede, line 536: If another code path sets PackageVulnerabilities, that code path would also need to make sure to update IsPackageVulnerable, PackageVulnerabilityCount, PackageVulnerabilities, IsPackageVulnerableOrDeprecated. I think these could all be updated only in the setter for PackageVulnerabilities, instead of here. Thoughts?

jebriede: Also, the issue I mentioned in my previous comment has not been resolved here. For example, take the IsPackageVulnerable property. In this code path, the PackageVulnerabilities will be set which will invoke `OnPropertyChanged(nameof(IsPackageVulnerable));` then after setting PackageVulnerabilities in this method, we once again call `OnPropertyChanged(nameof(IsPackageVulnerable));` therefore calling it twice. We should remove that redundancy.

jgonz120: I agree, I'm going to move most of these items to the setter for PackageVulnerabilities. That said I want to keep IsPackageVulnerableOrDeprecated in this location since it needs to be updated if the Deprecation information changes. Since that isn't set anywhere else right now, I want to keep here. If I move it to PackageVulnerabilities then we'll fire it twice.

jebriede (approved):

jeffkl (approved):

Nigusu-Allehu (approved):

PR #6190, opened 2024-12-12
Update MarkdownPreview to use Daytona. 

jgonz120:
# Bug

Fixes: https://github.com/NuGet/Client.Engineering/issues/3109
Fixes: https://github.com/NuGet/Client.Engineering/issues/3104
## Description
Update the Microsoft.VisualStudio.Markdown.Platform package to version 17.13.161. Set editor options in the PreviewBuilder to so MarkdownPreview renders content in Daytona. 

## PR Checklist

- [ ] ~Added tests~
- [ ] ~Link to an issue or pull request to update docs if this PR changes settings, environment variables, new feature, etc.~


NuGet.Config
```diff
@@ -59,9 +59,11 @@
       <package pattern="envdte80" />
       <package pattern="envdte90" />
       <package pattern="envdte90a" />
+      <package pattern="Microsoft.Build.Framework" />
       <package pattern="Microsoft.DataAI.NuGetRecommender.Contracts" />
       <package pattern="microsoft.internal.visualstudio.*" />
       <package pattern="Microsoft.Internal.VisualStudio.Shell.Framework" />
+      <package pattern="Microsoft.NET.StringTools" />
```
jgonz120, line 68: I recall there being concern about getting these items from the vside feed. 

`Microsoft.VisualStudio.Markdown.Platform` depends on a version of the SDK which is using the 17.13.x versions of both `Microsoft.NET.StringTools` and `Microsoft.Build.Framework`. Since these aren't released yet I would need to pull these from the VSIDE feed. 

I have another way of accomplishing this, where I created a new NuGet.config for the `NuGet.PackageManagement.UI` and set the PackageSourceMapping to * for the VSIDE source. It required a bunch of other changes

jeffkl (dismissed):
Can't wait to see it in action!

NuGet.Config
nkolev92: One concern is that now by specifying the full id, you're mapping these packages to this feed only.
If we need something temporarily, I think we should double map instead since we own both feeds.
We should move away from it eventually when the versions are released. 

> I have another way of accomplishing this, where I created a new NuGet.config for the NuGet.PackageManagement.UI and set the PackageSourceMapping to * for the VSIDE source

This won't be enough since you need all the parents of NuGet.PackageManagement.UI to share the same config as well. 

It's a best practice to use a single config and everything that works with that. 
Final thing is that nested configs don't really work in VS.

Nigusu-Allehu (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -87,7 +85,14 @@ private void PackageReadmeControl_Unloaded(object sender, RoutedEventArgs e)
 
         private void PackageReadmeControl_Loaded(object sender, RoutedEventArgs e)
         {
-            NuGetUIThreadHelper.JoinableTaskFactory.Run(UpdateMarkdownAsync);
+            NuGetUIThreadHelper.JoinableTaskFactory.Run(async () =>
+            {
+                var componentModel = await AsyncServiceProvider.GlobalProvider.GetComponentModelAsync();
+                var markdownPreviewSingleton = componentModel.GetService<MarkdownPreviewSingleton>();
+                _markdownPreview = markdownPreviewSingleton.GetInstance();
```
donnie-msft: The singleton could be null, right?

jgonz120: We don't consistently check for null for objects retrieved using `componentModel.GetService`. My gut says it shouldn't happen though since MarkdownPreviewSingleton is set up using MEF. I can make a change to hade the tab if markdownPreviewSingleton is null if you think it's necessary.

zivkan: I haven't seen it happen in a while, but until a year or two ago, each year I'd see at least 3 or 4 customers per year report that NuGet in VS was broken. Turned out that the MEF cache was broken and repairing VS fixed it, but since NuGet was the first component in VS that needed MEF, so NuGet was the first thing the customer noticed was broken.

So, on one hand, PM UI probably won't even load if the MEF cache is completely busted, so by the time it gets to this code, the MEF cache will probably successfully resolve this component.

On the other hand, there's been enough MEF problems over the years that I feel like it's only a matter of time until it gets a null (even if it's only in the tiniest percentage of customers), then line 92 causes a null ref exception.  Since this code is in a WPF callback, we know it's running on the UI thread, and an unhandled exception on the UI thread is going to cause a crash to desktop.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -87,7 +85,14 @@ private void PackageReadmeControl_Unloaded(object sender, RoutedEventArgs e)
 
         private void PackageReadmeControl_Loaded(object sender, RoutedEventArgs e)
         {
-            NuGetUIThreadHelper.JoinableTaskFactory.Run(UpdateMarkdownAsync);
+            NuGetUIThreadHelper.JoinableTaskFactory.Run(async () =>
+            {
+                var componentModel = await AsyncServiceProvider.GlobalProvider.GetComponentModelAsync();
+                var markdownPreviewSingleton = componentModel.GetService<MarkdownPreviewSingleton>();
```
zivkan: @davkean Is obtaining a MEF instance on the UI thread acceptable from a perf (I'm thinking UI delay) persepctive?

If `MarkdownPreviewSingleton` hasn't yet been initialized, and it requires assembly load, I'm guessing it could take non-trivial time, so this should be done on a background thread?  On the other hand, when it's already initialized, then getting the WPF callback on the UI thread, switching to a background thread, then switching back to the UI thread, is going to add latency.

The MEF related entries in the DevDiv wiki's perf guidelines doesn't explain if it's ok to `GetService`

Also @jgonz120 , I'm not sure if getting the singleton counts as "one time intialization", which the perf guidelines explicitly say not to do in the Loaded event: https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/27530/WPF?anchor=%E2%9D%8C-**do-not**-use-wpf%27s-loaded-event-for-%27one-time-initialization%27-kind-of-work.

davkean: This will hang the UI thread, both for the retrieval of IComponentModel if you are the first to retrieve it (especially if out of date, it could take >15 seconds), and on the retrieval of MarkdownPreviewSingleton, this work needs to be done off the UI thread.

jgonz120: @martinrrm you should check this out too, since this is done in the PackageRestoreBar.xaml as well.

jgonz120: > Also @jgonz120 , I'm not sure if getting the singleton counts as "one time intialization", which the perf guidelines explicitly say not to do in the Loaded event: https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/27530/WPF?anchor=%E2%9D%8C-**do-not**-use-wpf%27s-loaded-event-for-%27one-time-initialization%27-kind-of-work.

I'm not ure if this qualifies as a "one time initialization". Currently whenever we tab away and go back the view is recreated so all those values need to be set.

jgonz120: Okay I just pushed a change. We do a bunch of loading already in the PackageManagerControl, so I just threw the loading in there so it's in one place. I made sure to get off of the UI thread when getting the component model and the Markdown preview singleton. 

I also updated the tab control so the PackageReadmeControl isn't recreated, that helped reduce the size of the view load function

zivkan: Since singletons get initialized only once, it certainly feels like a one time initialization 😁 

As the above comments mentioned, if the markdown renderer hasn't been loaded yet, then it'll also cause assemblies, which is a one time intiailzation, even if you didn't intend or know about it.  Perf is complicated.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -22,16 +25,47 @@ public partial class PackageReadmeControl : UserControl
...
             InitializeComponent();
-            _markdownPreview = MarkdownPreviewSingleton.GetInstance();
-            descriptionMarkdownPreview.Content = _markdownPreview.VisualElement;
         }
 
         public ReadmePreviewViewModel ReadmeViewModel { get => (ReadmePreviewViewModel)DataContext; }
 
+        private async Task InitializeAsync()
+        {
+            if (_markdownPreview is null && !_isInitializing)
```
davkean: This isn't thread safe which is a requirement for async methods. Wrap this in an AsyncLazy<T> or a AsyncLazyInitializer, to ensure its only called once.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -75,19 +109,22 @@ private void UserControl_DataContextChanged(object sender, DependencyPropertyCha
...
-                ReadmeViewModel.PropertyChanged -= ReadmeViewModel_PropertyChanged;
             }
         }
 
         private void PackageReadmeControl_Loaded(object sender, RoutedEventArgs e)
         {
-            NuGetUIThreadHelper.JoinableTaskFactory.Run(UpdateMarkdownAsync);
+            _ = NuGetUIThreadHelper.JoinableTaskFactory.RunAsync(async () =>
+            {
+                await InitializeAsync();
```
davkean: ~You should have a cancellation token representing the lifetime of this JoinableTaskFactory:~

~[❌ DO NOT call JoinableTaskFactory.RunAsync without passing a cancellation token based on its owner's lifetime](
https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/24027/Threading?anchor=%E2%9D%8C-**do-not**-call-%60joinabletaskfactory.runasync%60-without-passing-a-cancellation-token-based-on-its-owner%27s-lifetime)~

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -22,16 +25,47 @@ public partial class PackageReadmeControl : UserControl
...
+                await TaskScheduler.Default;
+                IEditorOptionsFactoryService editorOptionsFactoryService;
+                try
+                {
+                    var componentModel = await AsyncServiceProvider.GlobalProvider.GetComponentModelAsync();
+                    editorOptionsFactoryService = componentModel?.GetService<IEditorOptionsFactoryService>();
+                }
+                finally
+                {
+                    await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
```
davkean: By the time this resumes, your UI could have been closed and disposed:

To demo this this add the following after the `TaskScheduler.Default`:

``` C#
await Task.Delay(15000)
```

Then close the UI before that 15 seconds times out, you will see that you crash Visual Studio when you resume.

What you should be doing is passing a CancellationToken that represents your "UI control/pages lifetime" and then you will be guaranteed to not resume on the UI thread should that cancellation token be disposed on the UI thread in Dispose.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -61,9 +95,9 @@ private void UserControl_DataContextChanged(object sender, DependencyPropertyCha
         {
             if (e.OldValue is ReadmePreviewViewModel oldMetadata)
             {
-                ThreadHelper.JoinableTaskFactory.Run(async () =>
+                _ = NuGetUIThreadHelper.JoinableTaskFactory.RunAsync(async () =>
```
davkean: If I'm reading code correctly `NuGetUIThreadHelper.JoinableTaskFactory` is wrapping `ThreadHelper.JoinableTaskFactory`, if so, then:

[❌ AVOID calling JoinableTaskFactory.RunAsync from the default factory on fire and forget tasks](https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/24027/Threading?anchor=%E2%9D%8C-**avoid**-calling-%60joinabletaskfactory.runasync%60-from-the-default-factory-on-fire-and-forget-tasks).

Usually you would call JoinableTaskFactory.RunAsync on a factory that lives for the lifetime of the thing that owns the task, in this case the UI that owns this tasks, but that's not the case here.

JoinableTaskFactory serves two purposes;

1) It lets you synchronously and safely block on a task that needs the UI thread. In this case, no one ever waits on it (as its fire and forget), so that will never occur.
2) It lets you "track" the task in the collection so that you can wait for it to finish via [JoinableTaskCollection.WaitTilEmpty](https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.threading.joinabletaskcollection.jointillemptyasync?view=visualstudiosdk-2022). The default factory from `ThreadHelper.JoinableTaskFactory` doesn't track tasks and is never waited on, so this is unneeded.

If this was a `AsyncPackage.JoinableTaskFactory`, then this would be a different story.

The equivalent of this without JoinableTaskFactory.

``` C#
FooAsync().FileAndForget(...);
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
davkean: Turns out this is the default factory, so this rule doesn't apply but do see cancellation token comment around InitializeAsync.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -22,16 +25,47 @@ public partial class PackageReadmeControl : UserControl
...
+        private async Task InitializeAsync()
+        {
+            if (_markdownPreview is null && !_isInitializing)
+            {
+                _isInitializing = true;
+                await TaskScheduler.Default;
+                IEditorOptionsFactoryService editorOptionsFactoryService;
+                try
+                {
+                    var componentModel = await AsyncServiceProvider.GlobalProvider.GetComponentModelAsync();
```
davkean: Please plumb through a cancellation token - this can block VS on possible very long running call if you change the `JoinableTaskFactory` to something other than `ThreadHelper.JoinableTaskFactory`. This will be likely in the future as I'm going to write up a correct pattern in our docs for UI components on how to use `JoinableTaskFactory` correctly and I hope you folks move to it. I'm can confusion around here and elsewhere in the VS tree that shows we're missing the right documentation around tghis.

~I don't know where GetComponentModelAsync comes from~, but I added a `CancellationToken` overload to `GetServiceAsync` in 17.11.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -22,16 +25,47 @@ public partial class PackageReadmeControl : UserControl
...
+                {
+                    await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+                }
+                if (editorOptionsFactoryService is not null)
+                {
+#pragma warning disable CS0618 // Type or member is obsolete
+                    var previewBuilder = new PreviewBuilder();
+#pragma warning restore CS0618 // Type or member is obsolete
+                    previewBuilder.EditorOptions = editorOptionsFactoryService.GlobalOptions;
+                    _markdownPreview = previewBuilder.Build();
```
davkean: Should markdown be built off the UI thread? I believe Amadeusz Wieczorek mentioned this can take a while [in some cases](https://developercommunity.visualstudio.com/t/Markdown-parser-stuck-in-infinite-loop/10791956).

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -22,16 +25,47 @@ public partial class PackageReadmeControl : UserControl
...
+                    var componentModel = await AsyncServiceProvider.GlobalProvider.GetComponentModelAsync();
+                    editorOptionsFactoryService = componentModel?.GetService<IEditorOptionsFactoryService>();
+                }
+                finally
+                {
+                    await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+                }
+                if (editorOptionsFactoryService is not null)
+                {
+#pragma warning disable CS0618 // Type or member is obsolete
```
davkean, line 69: A comment as to why this usage is okay probably needs warranted?

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
davkean: Note, if you do this, you are inevitably going to run into this [method](https://github.com/NuGet/NuGet.Client/blob/4e10b1165316e09b6570e51b4316d88b493889e9/src/NuGet.Clients/NuGet.VisualStudio.Common/ServiceProviderExtensions.cs#L38).

What is states is no longer true in 17.11+ and it can be deleted. GetServiceAsync is now free-threaded if the service does not need the UI thread. From now on, you should only be calling [ServiceExtensions.GetServiceAsync](https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.serviceextensions.getserviceasync?view=visualstudiosdk-2022) overloads, they do the right thing and the docs are completely up-to-date. @nkolev92 

This can clearly be deleted in a different PR and is unrelated to above.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
jgonz120: I think that would happen when we call `_markdownPreview.UpdateContentAsync`, I can check with him if it's possible to call that off of the UI thread.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
jgonz120: Yea I think it makes sense, they want to avoid initialization stuff to happen every time the control is loaded in the UI. I moved the init out of there.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
jgonz120: I changed it so it's no longer async, the component is loaded from the PackageManagerControl and passed into the init.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
jgonz120: Updated so it's no longer async

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
jgonz120: I switched it back to "Run" for now

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
jgonz120: I just tried it and UpdateContentAsync fails, since it eventually updates a webview/daytona with the results from Markdig I imagine it expects to be on the UI thread.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
jgonz120: I switched this to use https://github.com/NuGet/NuGet.Client/blob/c7a7a1ec509590a4eb2530d72fe266a2feb5008a/src/NuGet.Clients/NuGet.VisualStudio.Common/ServiceLocator.cs?plain=1#L85C1-L85C108. 

At least now we can just update one location once the change is made to load the component model

nkolev92: I'll do the clean-up of the rest of the methods. 

https://github.com/NuGet/Home/issues/14007

nkolev92: I'll clean-up the extra method. https://github.com/NuGet/Home/issues/14007

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
jgonz120: Okay turns out I forgot that I was accessing the VM. I made a change to allow it to run UpdateContentAsync  in the background. UpdateContentAsync doesn't take a cancellation token so unfortunately once I make the request to load the markdown it update the UI when it's complete.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
```diff
@@ -146,8 +147,15 @@ private async ValueTask InitializeAsync(PackageManagerModel model, INuGetUILogge
 
             _nugetPackageFileService?.Dispose();
             _nugetPackageFileService = await _serviceBroker.GetProxyAsync<INuGetPackageFileService>(NuGetServices.PackageFileService, CancellationToken.None);
+            var nuGetFeatureFlagService = await ServiceLocator.GetComponentModelServiceAsync<INuGetFeatureFlagService>();
+            var readmeTabEnabled = await nuGetFeatureFlagService.IsFeatureEnabledAsync(NuGetFeatureFlagConstants.RenderReadmeInPMUI);
+            if (readmeTabEnabled)
+            {
+                var editorOptionsFactoryService = await ServiceLocator.GetComponentModelServiceAsync<IEditorOptionsFactoryService>();
```
davkean, line 154: This line will increase the number of times this method hangs as it will force types and assemblies to be faulted in on the UI thread. It is [already is hanging]((https://prism.vsdata.io/events/?eventType=UIDelay&query=c%3DPackageManagerControl.InitializeAsync), this will increase it.

Retrieve the `IEditorOptionsFactoryService` off the UI thread;

``` C#
await TaskScheduler.Default;

var editorOptionsFactoryService = await ServiceLocator.GetComponentModelServiceAsync<IEditorOptionsFactoryService>();

await SwitchToMainThreadAsync(cancellationToken);
```
Note you need a cancellation token, otherwise the comment I said yesterday will apply; it will crash VS on resume.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -17,54 +19,123 @@ namespace NuGet.PackageManagement.UI
...
+            }
+            return true;
+        }
+
         private void ReadmeViewModel_PropertyChanged(object sender, PropertyChangedEventArgs e)
         {
             if (e.PropertyName == nameof(ReadmePreviewViewModel.ReadmeMarkdown))
             {
-                NuGetUIThreadHelper.JoinableTaskFactory.RunAsync(UpdateMarkdownAsync).PostOnFailure(nameof(PackageReadmeControl), nameof(ReadmeViewModel_PropertyChanged));
+                NuGetUIThreadHelper.JoinableTaskFactory.RunAsync(async () =>
```
davkean: Give this was already there, don't feel like you need to fix this for this PR but this is another fire and forget that adds JTF overhead for no reason, see: https://github.com/NuGet/NuGet.Client/pull/6190#discussion_r1889485483.

@nkolev92 @zivkan

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -17,54 +19,123 @@ namespace NuGet.PackageManagement.UI
...
+                    ReadmeViewModel.ErrorWithReadme = true;
+                    ReadmeViewModel.ReadmeMarkdown = string.Empty;
+                    await TelemetryUtility.PostFaultAsync(ex, nameof(ReadmePreviewViewModel));
+                }
+            }
+            finally
+            {
+                if (!_cancellationTokenSource.IsCancellationRequested)
+                {
+                    await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
```
davkean: This can resume after the UI has disappeared and crash VS, this needs to respect a cancellation token triggered in dispose.

While you already check here, there is a race between checking and you switching to the UI thread, the token must be passed to SwitchToMainThreadAsync.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -17,54 +19,123 @@ namespace NuGet.PackageManagement.UI
...
                 }
             }
             catch (Exception ex) when (ex is ArgumentException || ex is InvalidOperationException)
             {
-                ReadmeViewModel.ErrorWithReadme = true;
-                ReadmeViewModel.ReadmeMarkdown = string.Empty;
-                await TelemetryUtility.PostFaultAsync(ex, nameof(ReadmePreviewViewModel));
+                if (_cancellationTokenSource.IsCancellationRequested)
+                {
+                    await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
```
davkean: This can resume after the UI has disappeared and crash VS, this needs to respect a cancellation token triggered in dispose.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -17,54 +19,123 @@ namespace NuGet.PackageManagement.UI
...
+            PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(nameof(IsReadmeReady)));
+            PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(nameof(IsBusy)));
+        }
+
         private void UserControl_DataContextChanged(object sender, DependencyPropertyChangedEventArgs e)
         {
             if (e.OldValue is ReadmePreviewViewModel oldMetadata)
             {
-                ThreadHelper.JoinableTaskFactory.Run(async () =>
+                NuGetUIThreadHelper.JoinableTaskFactory.RunAsync(async () =>
```
davkean: Another fire and forget that adds JTF overhead for no reason, see: https://github.com/NuGet/NuGet.Client/pull/6190#discussion_r1889485483.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -73,21 +144,26 @@ private void UserControl_DataContextChanged(object sender, DependencyPropertyCha
             }
         }
 
-        private void PackageReadmeControl_Unloaded(object sender, RoutedEventArgs e)
+        private void PackageReadmeControl_Loaded(object sender, RoutedEventArgs e)
         {
-            if (ReadmeViewModel is not null)
+            NuGetUIThreadHelper.JoinableTaskFactory.RunAsync(async () =>
```
davkean: Another fire and forget that adds JTF overhead for no reason, see: https://github.com/NuGet/NuGet.Client/pull/6190#discussion_r1889485483.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
```diff
@@ -146,8 +147,15 @@ private async ValueTask InitializeAsync(PackageManagerModel model, INuGetUILogge
 
             _nugetPackageFileService?.Dispose();
             _nugetPackageFileService = await _serviceBroker.GetProxyAsync<INuGetPackageFileService>(NuGetServices.PackageFileService, CancellationToken.None);
+            var nuGetFeatureFlagService = await ServiceLocator.GetComponentModelServiceAsync<INuGetFeatureFlagService>();
```
davkean: The lines that pull components from MEF will increase the number of times this method hangs as it will force types and assemblies to be faulted in on the UI thread. It is [already is hanging](https://prism.vsdata.io/events/?eventType=UIDelay&query=c%3DPackageManagerControl.InitializeAsync), this will increase those hangs.

Retrieve the `INuGetFeatureFlagService`  and `IEditorOptionsFactoryService` off the UI thread;

``` C#
await TaskScheduler.Default;

var nuGetFeatureFlagService = await ServiceLocator.GetComponentModelServiceAsync<INuGetFeatureFlagService>();
....

var editorOptionsFactoryService = await ServiceLocator.GetComponentModelServiceAsync<IEditorOptionsFactoryService>();

await SwitchToMainThreadAsync(cancellationToken);
```
Note you need a cancellation token triggered by the UI from the Dispose, otherwise the comment I said yesterday will apply; it will crash VS on resume if the UI has been closed.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -73,21 +144,26 @@ private void UserControl_DataContextChanged(object sender, DependencyPropertyCha
...
-                ThreadHelper.JoinableTaskFactory.Run(async () =>
+                if (!string.IsNullOrWhiteSpace(ReadmeViewModel.ReadmeMarkdown))
                 {
-                    await _markdownPreview.UpdateContentAsync("", ScrollHint.None);
-                });
-                ReadmeViewModel.PropertyChanged -= ReadmeViewModel_PropertyChanged;
-            }
+                    await UpdateMarkdownAsync(ReadmeViewModel.ReadmeMarkdown);
+                }
+            }).PostOnFailure(nameof(PackageReadmeControl), nameof(PackageReadmeControl_Loaded));
```
davkean: @nkolev92 It seems this extension method that is attached to JoinableTask is forcing folks to `RunAsync` for fire and forget methods when it does nothing but overhead to these paths. That does not seem correct, see https://github.com/NuGet/NuGet.Client/pull/6190#discussion_r1889485483.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
jgonz120: Assuming I don't need to check the cancellation request since  SwitchToMainThreadAsync should throw an exception.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
jgonz120: I used the token from NuGetPackage. This function is called before the window is created so if I close the IDE while waiting for the service locators to run the dispose isn't called.

davkean: You have to worry about the reverse; what happens if you close the UI before this method finishes? it should be easy to test; add a Task.Delay(15000) within here and close the UI without closing the IDE, what happens? I see it touch UI when it resumes on the UI thread, so I don't see how it doesn't crash/fail. You should be combining the tokens at least.

jgonz120: Which UI? The pm UI doesn't open until after this initialization completes.

davkean: There is UI, this is a method on PackageManagerControl. You are saying, it's not visible at this point? 

So walk me through this, let's say the MEF cache is out of date and needs to be rebuilt. This takes at minimum of 15 seconds in a very fast machine in the lab, probably upwards of a minute on slower machines. You are saying the user goes to open the Package Manager and:

1) They are confused as nothing appears to happen while that cache is being rebuilt in the background.

-or-

2) They decide not to wait and instead close VS (let's say they need to update Setup), the package now blocks on that cache rebuild, VS UI disappears but this stops VS process from exiting, and setup complains that VS is still open?

These aren't hypertheticals, I constantly see 2) in VS hang dumps. This should be super easy to test; simply add a Task.Delay here for 60 seconds, and do above, what happens does the VS process exit (watch Task Manager) straight away?

jgonz120: So the method here is called before the PackageManagerControl is put in a window and that window is shown. Which is why I was saying that the PM UI is not visible during this process. The only option for the user would be to close VS completely. 

When closing VS completely it doesn't appear to be hanging. I also tried debugging placing a breakpoint after the 60 second delay and the debugger disconnected when I closed VS instead of hitting the breakpoint.

zivkan: it sounds like you're describing number 1 from Dave's reply.  If on someone's slow machine the markdown renderer control takes 15 seconds to initialize, the customer will have a 15 second delay between when they click the button to open PM UI to when it actually displays. This is bad CX.

Also, if the markdown control fails to initialize and throws an exception, then the customer will be unable to manage their packages at all, even if they don't want to read the package's readme. I see customers report "PM UI doesn't open" bug reports 1-5 times a year and in my opinion we "need" to refactor the PM UI open code so that it can't fail in this way (when it fails to open, an error message needs to be shown to the customer). But every time I've tried, it turned out to be a much bigger job than I planned for, so I keep giving up.

I think the original PR with a fire-and-forget JTF.RunAsync was better, because you could put a loading spinner in place of the readme control while waiting for it to load. If it throws an exception, replace the spinner with an error message. If it works, replace the spinner with the control.

If it was changed because of https://github.com/NuGet/NuGet.Client/pull/6190#discussion_r1889485483, I think there was a misunderstanding what Dave's comment was about. Dave's saying that NuGet's `NuGetUIThreadHelper` class violates the perf team's guidelines. It wasn't saying avoid async work.  I'd consider that work outside the scope of this PR, because NuGet uses NuGetUIThreadHelper in tons of places, so we need a significant architectural refactoring to remove it, and to pass around the `AsyncPackage`'s JTF instead.

jgonz120: Well, I moved these to avoid the one-time initializations in the component loaded event. I thought it would make sense to put these all in the same place so when we need to refactor, we don't need to search the code for the different locations. 

So, is it the case that each MEF load can add an additional 15-60s? There are 5 MEF objects being loaded in this area, 2 I added, so is it the case that we could be hanging up to 5 minutes because of these loads? If that's the case Is 5 minutes significantly worse than 3? At this point at least the UI would be responsive since I've moved the load off of the UI thread.

I'm happy to move these loads to another location but I would need one to occur when the control that loads the tab is being loaded, which does not have a loading icon, because it needs the INuGetFeatureFlagService to determine if the README tab is even available. 

So would this mean I should put these loads in the constructor of both of these user controls?

davkean: The first person to retrieve IComponentModel pays the cost of the load. The retrieval of services after that is still costly and we want off the UI thread, but will be faster than that.

1) The first retrieval of IComponentModel costs _at least_ between 400 - 520ms when the cache is up-to-date in the lab:

{F99BC6BF-17EA-4E52-93E7-3E7C72D78CA7}

2) When it is out of date (~5-10% of sessions right now), it can be many factors longer than that. While this out of dateness isn't always a rebuild, full rebuilds do occur and it takes 15 seconds in the lab:

{7D979431-23EA-406B-892D-0C76CE087DB8}

The lab machines are very fast machines, on slower machines or when the machine is busy, it will be much longer than this. The correct approach for this async "package manager window" is use an async tool window so that users know that something is loading, but failing that you can do after the UI has shown.

zivkan: I just remembered that NuGet has a second `AsyncPackage`, `RestoreManagerPackage`, that autoloads when any solution is open, so that it can handle packages.config (and legacy csproj PackageReference) restore on build events. It uses MEF.

So, if the 15 second delay is caused by the first thing that uses MEF, then MEF will already be loaded by the time the customer tries to open PM UI. 

The dotnet/project-system (or CPS?) communicates with NuGet by importing a type that NuGet exports. So SDK style project loads also need MEF to be loaded at solution open.

That means the only way this PR makes perf worse is by waiting for the markdown control to be loaded before displaying PM UI. But the 15 second is an impossible hypothetical unless there are significant rewrites of other parts of NuGet (and other components that need NuGet?) first. As such, I think it's way out of scope for this work. But it's good to increase our team's knowledge of ways that NuGet's implementation has design flaws that would be nice to improve when we have time.

davkean: It doesn't matter that there's another unrelated NuGet package that loads MEF, unless its waited on or blocked on by this path. What matters is the registered package and the path that causes this tool window be instantiated, it's the only thing guaranteed to have finished loading before the tool window is opened. That appears [to be this package](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.VisualStudio.OnlineEnvironment.Client/NuGetClientPackage.cs#L30), but TBH, it was a little hard to follow as NuGet instantiates this window very differently to other folks around the VS tree. 

I would not assume this path is fast, I can see, just on the UI thread (doesn't include time spent on background thread) along this path had about [13,000 UI freezes over 1 second in the past 21 days](https://prism.vsdata.io/events/?eventType=UIDelay&query=c%3DPackageManagerControl.InitializeAsync&showFG=false&failureType=none). I'll let you folks figure this one out, but I would not block on MEF on any path on the UI thread, or give no indication anything is happening while you wait for it.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
```diff
@@ -77,27 +78,34 @@ private PackageManagerControl()
...
+            await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(cancellationToken);
             _sinceLastRefresh = Stopwatch.StartNew();
 
             _installedTabTelemetryData = new PackageManagerInstalledTabData();
 
             Model = model;
             _uiLogger = uiLogger;
-            Settings = await ServiceLocator.GetComponentModelServiceAsync<ISettings>();
 
+            await TaskScheduler.Default;
```
davkean, line 101: Does CreateDocWindowForSolutionAsync end up waiting on this right? 

Re-reading how this code works trying to understand it for the MEF conversation, I think we might have introduced a race here; previously CreateDocWindowForSolutionAsync assumed that FindExistingSolutionWindowFrameAsync had already checked if the window was open but that check is only relevant while you are on the UI thread (someone can open the window behind your back when you get off UI thread).

I presume previously despite being marked as async, any awaits on this path weren't really yielded and stayed on the UI thread, effectively making it a synchronous path. 

Now, we're actually hitting real "yields" here (TaskScheduler.Default, any yields in service retrieval, etc) there's a race; if you invoke the command twice in succession very quickly, this path I think will be hit twice and the window will get opened twice or break state (I've not confirmed or looked at it).

You can simulate what this be like on slow machines or high thread pool latency by putting a Task.Delay(15000) here, and then invoke the method command twice in a row; what happens?

Other VS tool windows don't have this problem, because they say on the UI thread between the check and the creation of the window, or use the built-in async tool window logic for asynchronously loading the tool window.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
jgonz120: Is this in scope for this PR though? There are 3 MEF components already in this method, 1 from a downstream method that I moved here so it can be done off of the UI path. At this point I've added 1 additional MEF component load, is there a significant impact to performance by adding that additional MEF component?

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
jgonz120: I thought I responded to this. I did the following:
1. Set a 15000 delay after the line commented on. 
2. Started debugging
3. Right clicked solution and opened PM UI
4. Repeated step 3
5. Waited

When it completed, I only had one PM UI window open and it was working fine. I tried the same thing for the project level PM UI. 

I didn't get any exceptions in the debugger either. When I stepped through the debugger it wasn't obvious to me why a new window wasn't created.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
```diff
@@ -170,7 +181,7 @@ private async ValueTask InitializeAsync(PackageManagerModel model, INuGetUILogge
                 controller.PackageManagerControl = this;
             }
 
-            ISourceRepositoryProvider sourceRepositoryProvider = await ServiceLocator.GetComponentModelServiceAsync<ISourceRepositoryProvider>();
+
```
zivkan: nitpick, remove double empty lines

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -17,54 +20,111 @@ namespace NuGet.PackageManagement.UI
...
+        }
+
         private void ReadmeViewModel_PropertyChanged(object sender, PropertyChangedEventArgs e)
         {
             if (e.PropertyName == nameof(ReadmePreviewViewModel.ReadmeMarkdown))
             {
-                NuGetUIThreadHelper.JoinableTaskFactory.RunAsync(UpdateMarkdownAsync).PostOnFailure(nameof(PackageReadmeControl), nameof(ReadmeViewModel_PropertyChanged));
+                if (!string.IsNullOrWhiteSpace(ReadmeViewModel.ReadmeMarkdown))
+                {
+                    UpdateMarkdownAsync(ReadmeViewModel.ReadmeMarkdown, _cancellationTokenSource.Token).Forget();
```
zivkan: Does this mean when I'm viewing a package with a readme, but then I switch to a different version (or a different package) that does not have a readme, that the old package-version's readme is incorrectly going to be displayed?

zivkan: Is `.Forget()` being used, rather than `.PostOfFailure` because PostOnFailure only works on a JoinableTask?

If so, I think we should add an overflow for "normal" `Task`.  If unexpected exceptions start being thrown, we should get telemetry so we know we need to fix the customer experience (even if that's just showing a meaningful error message)

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -17,54 +20,111 @@ namespace NuGet.PackageManagement.UI
...
+                    await TaskScheduler.Default;
+                    await _markdownPreview.UpdateContentAsync(markdown, ScrollHint.None);
                 }
             }
             catch (Exception ex) when (ex is ArgumentException || ex is InvalidOperationException)
             {
+                await NuGetUIThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync(token);
                 ReadmeViewModel.ErrorWithReadme = true;
                 ReadmeViewModel.ReadmeMarkdown = string.Empty;
                 await TelemetryUtility.PostFaultAsync(ex, nameof(ReadmePreviewViewModel));
```
zivkan: It wasn't added in this PR, and it's related to my other comment regarding `Forget` vs `PostOnFailure`, but why is fault telemetry being added only when the exception is `ArgumentException` or `InvalidOperationException`.  If there's some other exception type being thrown preventing anything meaningful shown to the customer, why wouldn't we want to know about it?

I think we should remove the `PostFaultAsync` and use a fire-and-forget that automatically catches failed tasks, like our `JoinableTask` extension method.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -17,54 +20,111 @@ namespace NuGet.PackageManagement.UI
...
 
         private void UserControl_DataContextChanged(object sender, DependencyPropertyChangedEventArgs e)
         {
             if (e.OldValue is ReadmePreviewViewModel oldMetadata)
             {
-                ThreadHelper.JoinableTaskFactory.Run(async () =>
-                {
-                    await _markdownPreview.UpdateContentAsync("", ScrollHint.None);
-                });
+                UpdateMarkdownAsync("", _cancellationTokenSource.Token).Forget();
```
zivkan: similar to another comment, I find `.Forget()` surprising, rather than using `PostOnFailure` or a new overload of it.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -73,21 +133,23 @@ private void UserControl_DataContextChanged(object sender, DependencyPropertyCha
...
         {
-            if (ReadmeViewModel is not null)
+            if (!string.IsNullOrWhiteSpace(ReadmeViewModel.ReadmeMarkdown))
             {
-                ThreadHelper.JoinableTaskFactory.Run(async () =>
-                {
-                    await _markdownPreview.UpdateContentAsync("", ScrollHint.None);
-                });
-                ReadmeViewModel.PropertyChanged -= ReadmeViewModel_PropertyChanged;
+                UpdateMarkdownAsync(ReadmeViewModel.ReadmeMarkdown, _cancellationTokenSource.Token).Forget();
```
zivkan: Another `.Forget()` I find surprising.

src/NuGet.Clients/NuGet.Tools/NuGetPackage.cs
```diff
@@ -85,6 +85,8 @@ public sealed partial class NuGetPackage : AsyncPackage, IVsPackageExtensionProv
         private const string F1KeywordValuePmUI = "VS.NuGet.PackageManager.UI";
 
         private AsyncLazy<IVsMonitorSelection> _vsMonitorSelection;
+        private CancellationToken _cancellationToken;
```
zivkan: cancellation tokens happen on actions, so it doesn't seem right to keep them as state. Especially on the `NuGetPackage` state. I see it being set only in an initialize, which means it probably never gets cancellation requested (unless VS is closed before startup completes).

We already use `VsShellUtilities.ShutdownToken` elsewhere, so that's a better candidate to use whereever this field is currently being used.

src/NuGet.Clients/NuGet.VisualStudio.OnlineEnvironment.Client/SolutionExplorer/PackageManagerUICommandHandler.cs
```diff
@@ -292,7 +292,7 @@ private async ValueTask<IVsWindowFrame> CreateToolWindowAsync(WorkspaceVisualNod
                 INuGetUI uiController = await UIFactory.Value.CreateAsync(serviceBroker, projectContextInfo);
                 // This model takes ownership of --- and Dispose() responsibility for --- the INuGetUI instance.
                 var model = new PackageManagerModel(uiController, isSolution: false, editorFactoryGuid: GuidList.NuGetEditorType);
-                var control = await PackageManagerControl.CreateAsync(model, OutputConsoleLogger.Value);
+                var control = await PackageManagerControl.CreateAsync(model, OutputConsoleLogger.Value, CancellationToken.None);
```
zivkan: Like another comment, it should use the VS shutdown token

test/NuGet.Clients.Tests/NuGet.PackageManagement.UI.Test/NuGet.PackageManagement.UI.Test.csproj
```diff
@@ -18,10 +18,8 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.VisualStudio.Sdk.TestFramework.Xunit" />
-    <PackageReference Include="Microsoft.Build.Framework" />
-    <PackageReference Include="Microsoft.NET.StringTools" />
   </ItemGroup>
-
+
```
zivkan: nitpick: trailing spaces should be removed

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
jgonz120: No, if there is no readme available, we display a text letting the user know. If it's trying to render from a local package and  it isn't in the GPF the README tab is completely hidden.

jgonz120: Yea, based off of the comments like this one https://github.com/NuGet/NuGet.Client/pull/6190#discussion_r1891213554

I think FileAndForget is the equivalent but that's in the banned symbols list.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -17,54 +20,103 @@ namespace NuGet.PackageManagement.UI
...
+
         private void UserControl_DataContextChanged(object sender, DependencyPropertyChangedEventArgs e)
         {
             if (e.OldValue is ReadmePreviewViewModel oldMetadata)
             {
-                ThreadHelper.JoinableTaskFactory.Run(async () =>
-                {
-                    await _markdownPreview.UpdateContentAsync("", ScrollHint.None);
-                });
+                UpdateMarkdownAsync("", _cancellationTokenSource.Token).PostOnFailure(nameof(PackageReadmeControl)); ;
```
zivkan: async method (or at least, method claiming to be async) isn't awaited, and doesn't use `PostOnFailure`.

Also has unnecessary extra `;` on the line.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -17,54 +20,103 @@ namespace NuGet.PackageManagement.UI
...
     /// Interaction logic for PackageReadmeControl.xaml
     /// </summary>
-    public partial class PackageReadmeControl : UserControl
+    public partial class PackageReadmeControl : UserControl, IDisposable, INotifyPropertyChanged
     {
 #pragma warning disable CS0618 // Type or member is obsolete
         private IMarkdownPreview _markdownPreview;
 #pragma warning restore CS0618 // Type or member is obsolete
+        private bool _isDisposed = false;
+        private CancellationTokenSource _cancellationTokenSource;
```
zivkan: The type already tells us it's a cancellation token source. It should have a name that describes it's usage, like `_isDisposed` is not named `_bool`.

Perhaps something like `_controlDisposingTokenSource`

src/NuGet.Clients/NuGet.VisualStudio.Common/GlobalSuppressions.cs
```diff
@@ -32,3 +32,5 @@
...
 [assembly: SuppressMessage("Performance", "CA1822:Mark members as static", Justification = "<Pending>", Scope = "member", Target = "~M:NuGet.VisualStudio.Common.NuGetFeedbackDiagnosticFileProvider.HashSourceData(NuGet.ProjectModel.DependencyGraphSpec)")]
 [assembly: SuppressMessage("Globalization", "CA1305:Specify IFormatProvider", Justification = "<Pending>", Scope = "member", Target = "~M:NuGet.VisualStudio.Common.NuGetFeedbackDiagnosticFileProvider.GetOutputFile~System.IO.FileStream")]
+[assembly: SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "<Pending>", Scope = "member", Target = "~M:NuGet.VisualStudio.Telemetry.TaskExtensions.PostOnFailure(System.Threading.Tasks.Task,System.String,System.String)")]
+[assembly: SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "<Pending>", Scope = "member", Target = "~M:NuGet.VisualStudio.Telemetry.TaskExtensions.PostOnFailureAsync(System.Threading.Tasks.Task,System.String,System.String)~System.Threading.Tasks.Task{System.Boolean}")]
```
zivkan: I don't know how the rest of the team feels, but I really don't like GlobalSuppressions.cs. I'd rather have inline suppressions so I can see the code being suppressed, and make a decision if we should reduce tech debt by following the analyzer's advice.

But if you're going to use the global suppressions file, can we at least provide a justification?

src/NuGet.Clients/NuGet.VisualStudio.Common/Telemetry/TaskExtensions.cs
```diff
@@ -0,0 +1,55 @@
...
+                    await TelemetryUtility.PostFaultAsync(e, callerClassName, callerMemberName);
+                }
+            });
+        }
+
+        /// <summary> Records error information when the given task faults. </summary>
+        /// <param name="task"> Joinable task to execute. </param>
+        /// <param name="callerClassName"> Caller class name. </param>
+        /// <param name="callerMemberName"> Caller member name. </param>
+        /// <returns>True, if the task finishes successfully</returns>
```
zivkan:
```suggestion
        /// <returns>True, if the task did now throw an unhandled exception</returns>
```

Success can mean many things, depending on context A task that handled an error internally, is arguably not successful, but this method would return `true`.

zivkan (dismissed):
just nitpicks

edit: forgot the first one was not a nitpick

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
jgonz120: Removed the extra `;`, it has the post on failure though

src/NuGet.Clients/NuGet.VisualStudio.Common/GlobalSuppressions.cs
jgonz120: I prefer inline too

zivkan (dismissed):

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
jgonz120: @davkean I opened https://github.com/NuGet/Home/issues/14027 to capture the work that is needed for this. We are no longer blocking on the UI thread for the MEF components.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
jgonz120: I created an override for Task that allows for PostOnFailure.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -17,54 +20,103 @@ namespace NuGet.PackageManagement.UI
...
+
+        public bool IsBusy
+        {
+            get => _isBusy || ReadmeViewModel.IsBusy;
+        }
+
+        public bool IsReadmeReady
+        {
+            get => !_isBusy && ReadmeViewModel.IsReadmeReady;
+        }
```
donnie-msft: I'd expect properties like these to added to `ReadmePreviewViewModel`. Is there a reason that can be added to the description/etc as to why it's better here?

src/NuGet.Clients/NuGet.VisualStudio.Common/Telemetry/TaskExtensions.cs
donnie-msft, line 1: Should we just use JoinableTasks and then we won't need a generic TaskExtensions class?

src/NuGet.Clients/NuGet.VisualStudio.Common/Telemetry/TaskExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+using System;
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+
+namespace NuGet.VisualStudio.Telemetry
+{
+    public static class TaskExtensions
+    {
+        /// <summary> Records error information when the given task faults. </summary>
+        /// <param name="task"> Joinable task to execute. </param>
```
donnie-msft: nit: Task, not a Joinable task?

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
```diff
@@ -146,17 +154,20 @@ private async ValueTask InitializeAsync(PackageManagerModel model, INuGetUILogge
 
             _nugetPackageFileService?.Dispose();
             _nugetPackageFileService = await _serviceBroker.GetProxyAsync<INuGetPackageFileService>(NuGetServices.PackageFileService, CancellationToken.None);
+            var readmeTabEnabled = await nuGetFeatureFlagService.IsFeatureEnabledAsync(NuGetFeatureFlagConstants.RenderReadmeInPMUI);
+            if (readmeTabEnabled)
+            {
+                readmeTabEnabled = _packageDetail._packageDetailsTabControl.PackageReadmeControl.Initialize(editorOptionsFactoryService);
+            }
```
donnie-msft: Since PackageManagerControl acts as its own ViewModel, please refactor this such that it can be added to `IPackageManagerControlViewModel`.
The benefits are:
1. Other controls/models can leverage this code to know if the Readme is relevant in the current state.
2. Testing can be done (similar to 1)
3. When we separate this ViewModel into a separate class, the refactoring will be simpler.

src/NuGet.Clients/NuGet.VisualStudio.Common/Telemetry/TaskExtensions.cs
jgonz120: We were told to avoid using JoinableTasks for fire and forget methods
https://github.com/NuGet/NuGet.Client/pull/6190#discussion_r1889485483

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
jgonz120: I added an extra layer on top of it for when the markdown is being processed and rendered. Since this is happening in the view I added this "IsBusy" bool

jgonz120: Added comment

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
jgonz120: I think I did what you asked, though right now I don't see any packages using this viewmodel.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -17,54 +20,108 @@ namespace NuGet.PackageManagement.UI
...
             InitializeComponent();
-            _markdownPreview = MarkdownPreviewSingleton.GetInstance();
-            descriptionMarkdownPreview.Content = _markdownPreview.VisualElement;
+            _controlDisposingTokenSource = new CancellationTokenSource();
+        }
+
+        public event PropertyChangedEventHandler PropertyChanged;
+
+        /// <summary>
+        /// Added to allow the view to display the loading spinner while the markdown is being rendered
```
donnie-msft: I don't think the comment is necessary. It's clear to me what the property is used for, but I'm unclear why it's here and not in the ViewModel.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs
donnie-msft: > I think I did what you asked, though right now I don't see any packages using this viewmodel.

What do you mean by packages?

It's used in the NuGetUI class and its corresponding tests (https://github.com/search?q=repo%3ANuGet%2FNuGet.Client%20IPackageManagerControlViewModel&type=code).

jgonz120: Packages on the brain, I don't see any other classes using IsReadmeTabEnabled right now.

donnie-msft: >  I don't see any other classes using IsReadmeTabEnabled right now.

I wouldn't expect there to be since this PR is adding the property.
However, not having other references doesn't mean it shouldn't be in the VM. Most VM properties are only referenced by one View, yet there are still benefits. Also, the NuGetUI tests could be updated in this PR to use IsReadmeTabEnabled.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -17,54 +20,110 @@ namespace NuGet.PackageManagement.UI
...
-        private async Task UpdateMarkdownAsync()
+        private void ReadmeViewModel_PropertyChanged(object sender, PropertyChangedEventArgs e)
         {
-            try
+            if (e.PropertyName == nameof(ReadmePreviewViewModel.ReadmeMarkdown))
             {
                 if (!string.IsNullOrWhiteSpace(ReadmeViewModel.ReadmeMarkdown))
                 {
-                    await _markdownPreview.UpdateContentAsync(ReadmeViewModel.ReadmeMarkdown, ScrollHint.None);
+                    UpdateMarkdownAsync(ReadmeViewModel.ReadmeMarkdown, _controlDisposingTokenSource.Token).PostOnFailure(nameof(PackageReadmeControl)); ;
```
donnie-msft:
```suggestion
                    UpdateMarkdownAsync(ReadmeViewModel.ReadmeMarkdown, _controlDisposingTokenSource.Token).PostOnFailure(nameof(PackageReadmeControl));
```

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml.cs
```diff
@@ -17,54 +20,110 @@ namespace NuGet.PackageManagement.UI
...
+#pragma warning restore CS0618 // Type or member is obsolete
+                previewBuilder.EditorOptions = options.GlobalOptions;
+                _markdownPreview = previewBuilder.Build();
+                descriptionMarkdownPreview.Content = _markdownPreview?.VisualElement;
+            }
+            return true;
         }
 
-        private async Task UpdateMarkdownAsync()
+        private void ReadmeViewModel_PropertyChanged(object sender, PropertyChangedEventArgs e)
```
donnie-msft, line 79: We talked offline:
Eventually, I'd like to see this entire method go away in favor of setting the ViewModel `IsBusy` property instead.
Consider filing a techDebt issue describing the changes we discussed.

src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageReadmeControl.xaml
```diff
@@ -30,11 +30,11 @@
       x:Name="descriptionMarkdownPreview"
       ClipToBounds="True"
       Focusable="False"
-      Visibility="{Binding Path=IsReadmeReady, Mode=OneWay, FallbackValue=Collapsed, Converter={StaticResource BooleanToVisibilityConverter}}" />
+      Visibility="{Binding ElementName=PackageReadmeUserControl, Path=IsReadmeReady, Mode=OneWay, FallbackValue=Collapsed, Converter={StaticResource BooleanToVisibilityConverter}}" />
```
donnie-msft, line 33: We discussed offline:
This will also go away once a all the properties can be combined into a single ViewModel.

donnie-msft (dismissed):
Looks good!
The few design decisions that I think can be improved we discussed offline, and we agreed these can be handled in a future iteration.

Consider a bug bash early next week to help ensure there's no uncaught issues before GA.

donnie-msft (approved):

PR #6193, opened 2024-12-14
inline out variables

SimonCropp:
# Bug

Fixes: 

## PR Checklist

- [ ] Added tests

### DIFF
--- a/src/NuGet.Clients/NuGet.CommandLine/Commands/ProjectFactory.cs
+++ b/src/NuGet.Clients/NuGet.CommandLine/Commands/ProjectFactory.cs
@@ -432,8 +432,7 @@ namespace NuGet.CommandLine
 
         public string GetPropertyValue(string propertyName)
         {
-            string value;
-            if (!_properties.TryGetValue(propertyName, out value) &&
+            if (!_properties.TryGetValue(propertyName, out var value) &&
                 !ProjectProperties.TryGetValue(propertyName, out value))
             {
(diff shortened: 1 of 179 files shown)


src/NuGet.Clients/NuGet.CommandLine/Commands/ProjectFactory.cs
```diff
@@ -432,8 +432,7 @@ public string InitializeProperties(Packaging.IPackageMetadata metadata)
 
         public string GetPropertyValue(string propertyName)
         {
-            string value;
-            if (!_properties.TryGetValue(propertyName, out value) &&
+            if (!_properties.TryGetValue(propertyName, out var value) &&
```
zivkan, line 435: This is a violation of our repo's coding guideline number 10: https://github.com/NuGet/NuGet.Client/blob/dev/docs/coding-guidelines.md

it's not obvious what `value`'s type is, so it shouldn't be `var`. While I'm not a big fan of this coding guideline (if I really care, I'll open VS and use intellisense), it's what the team voted for (to make reviewing PRs easiler, maybe?).

Searching GitHub's PR page for `var out`, there are over 300 instances, and I expect most of them will be "non-obvious" and therefore should be changed if we follow the coding guidelines.

SimonCropp: @zivkan given that convention. can we merge this one first https://github.com/NuGet/NuGet.Client/pull/6202

PR #6194, opened 2024-12-14
cleanup ProjectRestoreInfo3Adapter

SimonCropp:
# Bug

Fixes: 

## PR Checklist

- [ ] Added tests

### DIFF
--- a/src/NuGet.Clients/NuGet.SolutionRestoreManager/ProjectRestoreInfo3Adapter.cs
+++ b/src/NuGet.Clients/NuGet.SolutionRestoreManager/ProjectRestoreInfo3Adapter.cs
@@ -72,12 +72,11 @@ namespace NuGet.SolutionRestoreManager
         {
             if (reference is null) { return null; }
 
-            var a = new VsReferenceItem2Adapter
+            return new VsReferenceItem2Adapter
             {
                 Name = reference.Name,
                 Metadata = ToPropertiesDictionary(reference.Properties),
             };
-            return a;
         }
 
         private static IReadOnlyList<IVsTargetFrameworkInfo4> ToTargetFrameworkInfoList(IVsTargetFrameworks targetFrameworks)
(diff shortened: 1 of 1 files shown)


src/NuGet.Clients/NuGet.SolutionRestoreManager/ProjectRestoreInfo3Adapter.cs
```diff
@@ -72,12 +72,11 @@ public static IVsProjectRestoreInfo3 Create(IVsProjectRestoreInfo2 projectRestor
         {
             if (reference is null) { return null; }
 
-            var a = new VsReferenceItem2Adapter
+            return new VsReferenceItem2Adapter
```
zivkan, line 75: I'm not a fan of returning new instances on the same line they're allocated.

When debugging, there are times I want to inspect the newly created instance. When the allocation and return are on separate lines, I can put a breakpoint on the return, and that's it.

After this PR, I'll need to put a breakpoint on the line, then once the breakpoint is hit, do a step out so I can see the instance in the calling method, assuming it saves the result in a variable. If the caller calls this method in a property initializer pattern, or a nested method call, then it gets much more difficult to inspect the instance.

The variable name `a` isn't very good, it probably should have been `adapter`. But otherwise, I don't understand what benefit there is to remove the variable.

src/NuGet.Clients/NuGet.SolutionRestoreManager/ProjectRestoreInfo3Adapter.cs
```diff
@@ -176,36 +171,30 @@ private static IReadOnlyDictionary<string, IReadOnlyList<IVsReferenceItem2>> ToI
...
-                IReadOnlyList<IVsReferenceItem2> projectReferences = ToReferenceItemList(targetFrameworkInfo.ProjectReferences);
-                result[ProjectItems.ProjectReference] = projectReferences;
+                result[ProjectItems.ProjectReference] = ToReferenceItemList(targetFrameworkInfo.ProjectReferences);
             }
 
             if (targetFrameworkInfo.PackageReferences?.Count > 0)
             {
-                IReadOnlyList<IVsReferenceItem2>? packageReferences = ToReferenceItemList(targetFrameworkInfo.PackageReferences);
-                result[ProjectItems.PackageReference] = packageReferences;
+                result[ProjectItems.PackageReference] = ToReferenceItemList(targetFrameworkInfo.PackageReferences);
```
zivkan, line 179: Similar to my other comment, variables are free, and this makes it easier to inspect while debugging.  You didn't add any details to the PR description, which is a violation of our process, and in particular the description part of the PR template would have helped me understand why you want to do this.

What benefit are you intending to gain from this?

nkolev92:
This is purely a stylistic change. 

If the code was written originally like this, I'd approve it. If I had reviewed the original change PR, I'd probably approve it too (with some exception for examples, there's a variable named `a`. 

We won't take this PR at this point. 

Thanks for the effort!

PR #6197, opened 2024-12-16
avoid type allocation and use nameof instead

SimonCropp:
# Bug

Fixes: 

## PR Checklist

- [ ] Added tests

### DIFF
--- a/src/NuGet.Clients/NuGet.CommandLine/Common/ConsoleCredentialProvider.cs
+++ b/src/NuGet.Clients/NuGet.CommandLine/Common/ConsoleCredentialProvider.cs
@@ -13,7 +13,7 @@ namespace NuGet.CommandLine
         public ConsoleCredentialProvider(IConsole console)
         {
             Console = console;
-            Id = $"{typeof(ConsoleCredentialProvider).Name}_{Guid.NewGuid()}";
+            Id = $"{nameof(ConsoleCredentialProvider)}_{Guid.NewGuid()}";
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


src/NuGet.Clients/NuGet.CommandLine/Common/ConsoleCredentialProvider.cs
```diff
@@ -13,7 +13,7 @@ public class ConsoleCredentialProvider : Credentials.ICredentialProvider
         public ConsoleCredentialProvider(IConsole console)
         {
             Console = console;
-            Id = $"{typeof(ConsoleCredentialProvider).Name}_{Guid.NewGuid()}";
+            Id = $"{nameof(ConsoleCredentialProvider)}_{Guid.NewGuid()}";
```
nkolev92, line 16: Is there a rule we can enable for this?

nkolev92 (commented):
Can we get an issue created so it can be captured as a fixed issue in our release notes?

SimonCropp:
> Can we get an issue created so it can be captured as a fixed issue in our release notes?

can you change your release notes process so it can include a PR without an issue. given a PR has all the features of an issue, just that the changeset is included in it. requiring every simple improvement to need a PR+issue pair is unnecessary bureaucracy

src/NuGet.Clients/NuGet.CommandLine/Common/ConsoleCredentialProvider.cs
SimonCropp: there is, but only for Rider and R#. the editorconfig key is `resharper_use_name_of_instead_of_type_of_highlighting`. want me to add it? it will light up for r# and rider, and be ignored by everything else

nkolev92: I think it's fine to leave it out in that case. 
We'd basically want it to be raised at build time.

PR #2873, opened 2019-06-05
Style:  silence "var" code style rules

dtivel:

PR #6202, opened 2024-12-20
fix var settings

SimonCropp:
# Bug

Fixes: 

## Description

dont use var except where type is apparent. as per https://github.com/NuGet/NuGet.Client/blob/dev/docs/coding-guidelines.md#c-coding-style

## PR Checklist

- [ ] Added tests

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -59,15 +59,15 @@ dotnet_style_explicit_tuple_names = true:suggestion
 dotnet_style_coalesce_expression = true:suggestion
 dotnet_style_null_propagation = true:suggestion
 
-; Licence header
+; License header
 file_header_template = Copyright (c) .NET Foundation. All rights reserved.\nLicensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 ; CSharp code style settings
 [*.cs]
 ; IDE0007 'var' preferences
-csharp_style_var_for_built_in_types = true:none
-csharp_style_var_when_type_is_apparent = true:none
-csharp_style_var_elsewhere = false:none
+csharp_style_var_for_built_in_types = false:error
+csharp_style_var_when_type_is_apparent = true:error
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -59,15 +59,15 @@ dotnet_style_explicit_tuple_names = true:suggestion
...
 
 ; CSharp code style settings
 [*.cs]
 ; IDE0007 'var' preferences
-csharp_style_var_for_built_in_types = true:none
-csharp_style_var_when_type_is_apparent = true:none
-csharp_style_var_elsewhere = false:none
+csharp_style_var_for_built_in_types = false:error
+csharp_style_var_when_type_is_apparent = true:error
+csharp_style_var_elsewhere = false:error
```
zivkan: If you use git blame to look through the file history, these rules were disabled in https://github.com/NuGet/NuGet.Client/pull/2873

Unfortunately, it didn't have context, so I just posted a comment with context.

So, while this PR is well intentioned, as I just posted in that PR, the benefit isn't worth the impact to us.

Just to verify, I checked out your branch, opened VS and opened a bunch of files I often work with, and that I will need to work with soon.  A lot of them had a whole lot of red squiggly lines, which is exactly what we wanted to avoid when we disabled these style rules in 2019. It was unfortunate that we didn't have context publicly available before.

SimonCropp: if u dont care about consitency of var usage, then why are you blocking PRs based on var usage https://github.com/NuGet/NuGet.Client/pull/6193

zivkan: Some people in the team cared enough to bring it up at a team meeting and get enough agreement to add it to our coding guidelines.

SimonCropp: seems strange to care enought to add it to the coding guidelines, but not care enough to make the code match it

but i have downgraded the setting to warning in this PR. is that ok

PR #6205, opened 2024-12-20
Add SDKAnalysisLevel and UsingMicrosoftNETSdk to the restore telemetry

nkolev92:
# Bug

Related: https://github.com/NuGet/Client.Engineering/issues/3075

## Description

This will allow to make inferences about the SDKAnalysisLevel usage (a bit of guess work with the default SDK version) and different between legacy and SDK projects in the projectrestoreinformation event.

## PR Checklist

- [x] Added tests

### DIFF
--- a/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
+++ b/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
@@ -59,6 +59,8 @@ namespace NuGet.Commands
         private const string TargetFrameworksCount = nameof(TargetFrameworksCount);
         private const string RuntimeIdentifiersCount = nameof(RuntimeIdentifiersCount);
         private const string TreatWarningsAsErrors = nameof(TreatWarningsAsErrors);
+        private const string SDKAnalysisLevel = nameof(SDKAnalysisLevel);
+        private const string UsingMicrosoftNETSdk = nameof(UsingMicrosoftNETSdk);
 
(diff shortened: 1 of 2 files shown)


Nigusu-Allehu (approved):

src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreCommand.cs
```diff
@@ -59,6 +59,8 @@ private readonly Dictionary<RestoreTargetGraph, Dictionary<string, LibraryInclud
         private const string TargetFrameworksCount = nameof(TargetFrameworksCount);
         private const string RuntimeIdentifiersCount = nameof(RuntimeIdentifiersCount);
         private const string TreatWarningsAsErrors = nameof(TreatWarningsAsErrors);
+        private const string SDKAnalysisLevel = nameof(SDKAnalysisLevel);
+        private const string UsingMicrosoftNETSdk = nameof(UsingMicrosoftNETSdk);
```
zivkan, line 63: We should also include `NETCoreSdkVersion`, otherwise we have no way guess if the customer is using the default `SDKAnalysisLevel` or if they explicitly set a different value.  Without being able to estimate explicit SDKAnalysisLevel usage, I don't know what the point of collecting the telemetry is.

nkolev92: Talked offline. 

I'll add that as well.

